PR #36, opened 2014-10-25
Organize platform detection code to avoid "Duplication of Roles".

muojp:
- Unified similar code in PlatformApis and Libuv.
 - Switched to better platform constants than IsWindows/IsDarwin/...
 - Adopt KestrelEngine to new style.


DamianEdwards:
Apologies for sitting on this for so long. Since this was sent, we've added an API `IRuntimeEnvironment` in DNX itself which can be used to provide details of the host OS.

Issue #9, opened 2014-06-28
Shutting the server down causes a hang


davidfowl:
I ran the sample on osx using k run and then hit enter to quit the app after making a single request, it hangs forever

Tragetaschen:
#60 fixes this

PR #47, opened 2014-12-26
Fixed shutdown hanging

davidfowl:
- Added Stopping event to KestrelThread so that
incoming connections get a chance to gracefully close
before disposing the event loop
- Call _post.Dispose instead of UnReference. Sometimes
Dispose would happen on the finalizer thread and _queueCloseHandle
would be null.

#9


borgdylan:
ping

PR #60, opened 2015-02-08
P/Invoke cleanup

Tragetaschen:
This PR cleans up large portions of the P/Invoke handling of Kestrel to libuv. I know these are a lot of changes, but I have kept the single commits for easier following.

The highlights are:
- ~~Use `DllImport` and Mono's `DllMap` feature instead of manually loading and binding a platform-specific library~~
- The various `SafeHandle`s now do all the required allocation work in their constructors
- No more `unsafe`
- Working shutdown

I have tested on both Windows and Linux, but could not do so on a Mac simply because I don't have access to one. My test setup included a web application with an active SignalR connection. I was not able to reproduce any of the shutdown issues again (yet?)


msotclas:
Hi __@Tragetaschen__, I'm your friendly neighborhood Microsoft Open Technologies, Inc. Pull Request Bot (You can call me MSOTBOT). Thanks for your contribution!
    <span>You've already signed the contribution license agreement. Thanks!</span>
        <p>The agreement was validated by Microsoft Open Technologies, Inc. and real humans are currently evaluating your PR.</p>

TTYL, MSOTBOT;

lodejard:
Wow, that's a lot of work! The individual commits are very helpful and they sound like good ideas based on the descriptions. We'll take a closer look soon. Thank you very much!

Tragetaschen:
As a final step (well, let's see about that):

- `UvHandles` are not `SafeHandle`s anymore

Since they have to clean both loop resource as well as the unmanaged memory, the SafeHandle pattern does not really work: By the time the Finalizer runs, the loop might have been long brought down and posting to it won't work. The new base class implements a working UvHandle pattern and warns when a finalizer does the cleanup (I don't get the warning).

Tragetaschen:
Indirectly, I know that OSx won't work, but it's close. I have modified my Linux environment so it looks similar to how OSx will behave (basically removed libuv* from /usr/lib64 and placed the files in various folders).

The reason it worked on Linux so far is that mono is quite smart with `DllImport`s. "libuv.dll" is mapped to "libuv.so" and then searched for in the standard folders. Because I have a link `libuv.so -> libuv.so.1.0.0` in `/usr/lib64`, everything Just Works (tm). On OSx, a compatible file is not in the standard folders, though.

I'll revert to the manual binding :-/

src/Microsoft.AspNet.Server.Kestrel/Networking/Libuv.cs
```diff
@@ -2,413 +2,44 @@
...
-        public Libuv()
+        public static bool IsWindows
         {
-            IsWindows = PlatformApis.IsWindows();
-            if (!IsWindows)
+            get
             {
-                IsDarwin = PlatformApis.IsDarwin();
+#if ASPNETCORE50
+                return true;
```
Eilon: While this is true now, it won't be true in the future... what would be the correct code once Core CLR runs on non-Windows?

src/Microsoft.AspNet.Server.Kestrel/App.config
```diff
@@ -0,0 +1,4 @@
+﻿<configuration>
+  <dllmap dll="libuv.dll" target="libuv.so.1" os="linux"/>
+  <dllmap dll="libuv.dll" target="native/darwin/universal/libuv.dylib" os="osx"/>
```
Eilon: How does the `app.config` get used? Normally I thought only EXE's can have a config file, no?

desdesdes: This is a special file for library mapping in mono. See http://www.mono-project.com/docs/advanced/pinvoke/#library-names

Eilon: Does the file need to get included in the NuGet package? Does that happen automatically in this build system? (Sorry for the n00b questions on this!)

markrendle: Has anyone done relative performance testing on Kestrel before and after these changes? Happy to do so tomorrow on Linux and Windows, can't do Mac.

davidfowl: This file can't really be used in general with the KRE. It won't work because it won't be found

Tragetaschen: And it's gone with the recent commit.

src/Microsoft.AspNet.Server.Kestrel/Http/SocketOutput.cs
```diff
@@ -46,21 +44,26 @@ public void Write(ArraySegment<byte> buffer, Action<Exception, object> callback,
...
             private static readonly Action<UvWriteReq, int, Exception, object> _writeCallback = WriteCallback;
+
             private static void WriteCallback(UvWriteReq req, int status, Exception error, object state)
             {
                 ((ThisWriteReq)state).OnWrite(req, status, error);
             }
 
             SocketOutput _self;
-            ArraySegment<byte> _buffer;
+            byte[] _buffer;
```
davidfowl: Why?

src/Microsoft.AspNet.Server.Kestrel/Infrastructure/KestrelThread.cs
```diff
@@ -98,67 +79,53 @@ public Task PostAsync(Action<object> callback, object state)
...
-                if (_stopImmediate)
+                try
                 {
-                    // thread-abort form of exit, resources will be leaked
-                    return;
+                    _post.Reference();
+                    _loop.Run();
+
+                    _loop.Validate();
+                    UnsafeNativeMethods.uv_walk(
```
davidfowl: Why do we still need this after the cleanup? Shutdown should work fine without this logic right?

src/Microsoft.AspNet.Server.Kestrel/Networking/Libuv.cs
Tragetaschen: Good question.

This code is the same as before, I just took it as is. However, I spent a few thoughts and decided I don't know enough to fix it.

src/Microsoft.AspNet.Server.Kestrel/Infrastructure/KestrelThread.cs
Tragetaschen: There might still be active connections (like SignalR). These need to be closed manually.

src/Microsoft.AspNet.Server.Kestrel/App.config
lodejard: Just checked with @GrabYourPitchforks and it sounds like you wouldn't expect a difference between DllImport P/Invoke and GetProcAddress delegate P/Invoke, as long as the same delegate instances are used repeatedly.


________________________________

src/Microsoft.AspNet.Server.Kestrel/Infrastructure/KestrelThread.cs
davidfowl: This definitely isn't the way to kill signalr connections. I'm ok with this being a last resort. SignalR itself listens to the ApplicationStopping token and releases connections gracefully when that happens. 

I'm just wondering under what circumstances with the new code does this code need to run.

src/Microsoft.AspNet.Server.Kestrel/Http/SocketOutput.cs
Tragetaschen: The only code path into the Contextualize currently creates a copy of the incoming ArraySegment. I figured I can simplify the code down to the P/Invoke.

src/Microsoft.AspNet.Server.Kestrel/Infrastructure/KestrelThread.cs
Tragetaschen: Hmm, you bring up a good point.

If I put a break point in the Dispose logic while there is an active SignalR connection, I still get `UvAsyncHandle.Send`-Requests that come in through `FrameResponseStream.Write`. I didn't look into that, but now that you brought it up I actually looked a the stack trace when this happens.

```
   at System.Environment.GetStackTrace(Exception e, Boolean needFileInfo)
   at System.Environment.get_StackTrace()
   at Microsoft.AspNet.Server.Kestrel.Networking.UvAsyncHandle.Send()
   at Microsoft.AspNet.Server.Kestrel.KestrelThread.Post(Action`1 callback, Object state)
   at Microsoft.AspNet.Server.Kestrel.Http.SocketOutput.Write(ArraySegment`1 buffer, Action`2 callback, Object state)
   at Microsoft.AspNet.Server.Kestrel.Http.Frame.Write(ArraySegment`1 data, Action`2 callback, Object state)
   at Microsoft.AspNet.Server.Kestrel.Http.FrameResponseStream.WriteAsync(Byte[] buffer, Int32 offset, Int32 count, CancellationToken cancellationToken)
   at System.IO.Stream.WriteAsync(Byte[] buffer, Int32 offset, Int32 count)
   at Microsoft.AspNet.Server.Kestrel.Http.FrameResponseStream.Write(Byte[] buffer, Int32 offset, Int32 count)
   at Microsoft.AspNet.SignalR.ResponseExtensions.Write(HttpResponse response, ArraySegment`1 data)
   at Microsoft.AspNet.SignalR.Transports.ServerSentEventsTransport.PerformKeepAlive(Object state)
   at Microsoft.AspNet.SignalR.Transports.ServerSentEventsTransport.<>c__DisplayClass0.<KeepAlive>b__1(Object state)
   at Microsoft.AspNet.SignalR.Infrastructure.TaskQueue.InvokeNext(Func`2 next, Object nextState, Object queueState)
   at Microsoft.AspNet.SignalR.Infrastructure.TaskQueue.<>c__DisplayClass0.<Enqueue>b__1(Func`2 n, Object ns, TaskQueue q)
   at Microsoft.AspNet.SignalR.TaskAsyncHelper.FromMethod[T1,T2,T3](Func`4 func, T1 arg1, T2 arg2, T3 arg3)
   at Microsoft.AspNet.SignalR.TaskAsyncHelper.Then[T1,T2,T3](Task task, Func`4 successor, T1 arg1, T2 arg2, T3 arg3)
   at Microsoft.AspNet.SignalR.Infrastructure.TaskQueue.Enqueue(Func`2 taskFunc, Object state)
   at Microsoft.AspNet.SignalR.Transports.TransportDisconnectBase.EnqueueOperation(Func`2 writeAsync, Object state)
   at Microsoft.AspNet.SignalR.Transports.ServerSentEventsTransport.KeepAlive()
   at Microsoft.AspNet.SignalR.Transports.TransportHeartbeat.CheckTimeoutAndKeepAlive(ConnectionMetadata metadata)
   at Microsoft.AspNet.SignalR.Transports.TransportHeartbeat.Beat(Object state)
[...]
   at System.Threading.TimerQueue.AppDomainTimerCallback()"
```

Looks like the heartbeat does not stop.

src/Microsoft.AspNet.Server.Kestrel/Networking/Libuv.cs
Eilon: Ah OK we'll figure this out in the future, then.

Tragetaschen:
- No more manual writing in unmanaged memory
- Clean shutdown logic (no more `uv_stop`)

src/Microsoft.AspNet.Server.Kestrel/Http/SocketOutput.cs
lodejard: That part of the server hasn't been optimized at all really (which is why it's copying the array segment every time) - so I wouldn't worry too much about fixing that code path before it's been fine-tuned.

Tragetaschen:
I will now stop shuffeling code around :)

Eilon:
@lodejard can you try this out locally and merge?

Tragetaschen:
I'm close to fixing most of the bugs I found so far (see the [-beta2 branch in my fork](https://github.com/Tragetaschen/KestrelHttpServer/tree/pinvoke-cleanup-beta2)). For stability reasons, I'd prefer that additional work to be included

Tragetaschen:
- Connections are now closed cleanly on shutdown as opposed through the Finalizer
- Keep-Alive working
- Connections closed by the client working

I'm not quite convinced with the last commit. Not that I couldn't defend its practical correctness, but some code paths down the `MessageBody.Consume` implementations are not that clean in their handling of `RemoteIntakeFin` and `Buffer.Count == 0`

The leaking SignalR connections currently trigger a `Console` warning during shutdown.

Tragetaschen:
I've done a fair amount of testing including `ab` stress testing with half a million requests and I don't expect to have introduced any regressions while fixing a couple of issues.

Tragetaschen:
There is a race between closing the socket client-wise on the one hand and still writing from the application's end. The write request could throw an `ObjectDisposedException` on its on-loop-part that would bring down the loop.

Instead of catching this exception, I made the `Write` call `WriteAsync`. Now, the exception still occurs, but is handled by whomever is currently handling the request.

Tragetaschen:
When a connection with an active write request is closed, the write request's callback gets -UV_ECANCELED as error code. This is now handled as an expected error.

src/Microsoft.AspNet.Server.Kestrel/Infrastructure/KestrelThread.cs
```diff
@@ -67,167 +53,121 @@ public void Stop(TimeSpan timeout)
...
-                {
-                    Trace.WriteLine("KestrelThread.DoPostCloseHandle " + ex.ToString());
-                }
+                work();
             }
+            finishedBatch.Clear();
         }
 
-        private struct Work
+        private Queue<Action> finishCurrentBatch()
```
davidfowl: Lowercase?

src/Microsoft.AspNet.Server.Kestrel/Infrastructure/KestrelThread.cs
```diff
@@ -67,167 +53,121 @@ public void Stop(TimeSpan timeout)
...
+            // However, the world isn't perfect and there are currently ways
+            //  that handles are left that would make the loop run forever.
+            // Right now from the loop's point of view, _post is still active.
+            // So we skip _post when we go through the handles that are still active
+            //  and close them manually.
+            UnsafeNativeMethods.uv_walk(
+                _loop,
+                (ptr, arg) =>
+                {
+                    if (ptr != postHandle)
```
davidfowl: We don't do self closing ifs. Use braces.

src/Microsoft.AspNet.Server.Kestrel/Infrastructure/KestrelThread.cs
```diff
@@ -50,16 +42,10 @@ public Task StartAsync()
...
             if (!_thread.Join((int)timeout.TotalMilliseconds))
             {
-                Post(OnStopImmediate, null);
-                if (!_thread.Join((int)timeout.TotalMilliseconds))
-                {
-#if ASPNET50
-                    _thread.Abort();
-#endif
-                }
+                throw new TimeoutException("Loop did not close");
```
davidfowl: Why aren't we aborting the thread?

Tragetaschen: There are two reasons:

First, aborting a thread doesn't work while it's in a native call. This means while in the loop and waiting to invoke callbacks, `Abort` only schedules the `ThreadAbortException` to be called once the loop enters managed code (in either direction).

On the other hand, the documentation for Libuv is clear in that the loop will only run while there are active handles. `OnStop` makes sure that those are forcibly closed, so the loop will exit naturally.

davidfowl:
This needs a rebase on top of dev

Tragetaschen:
Sure

src/Microsoft.AspNet.Server.Kestrel/Http/Connection.cs
```diff
@@ -35,78 +35,80 @@ public interface IConnectionControl
...
-            {
-                KestrelTrace.Log.ConnectionRead(_connectionId, status);
-            }
-            else if (normalDone || errorDone)
+            if (!normalRead)
             {
                 KestrelTrace.Log.ConnectionReadFin(_connectionId);
                 SocketInput.RemoteIntakeFin = true;
-                _socket.ReadStop();
+                if (status != -4095)
```
Tratcher: Can you define a descriptive constant for this? The same goes for the ones above.

Tragetaschen:
Watching [last week's Community Standup](https://www.youtube.com/watch?v=bFXseBPGAyQ&list=PL0M0zPgJ3HSftTAAHttA3JQU4vOjXFquF&index=1) made me really, really sad :(

PR #68, opened 2015-03-02
Implemented k kestrel should display the address and port that is being listened on

ziyasal:
Implemented k kestrel should display the address and port that is being
listened on


msotclas:
Hi __@ziyasal__, I'm your friendly neighborhood Microsoft Open Technologies, Inc. Pull Request Bot (You can call me MSOTBOT). Thanks for your contribution!
    <span>
        In order for us to evaluate and accept your PR, we ask that you sign a contribution license agreement. It's all electronic and will take just minutes. I promise there's no faxing. https://cla2.msopentech.com.
    </span>

TTYL, MSOTBOT;

src/Microsoft.AspNet.Server.Kestrel/KestrelEngine.cs
```diff
@@ -21,6 +23,10 @@ public KestrelEngine(ILibraryManager libraryManager)
             Memory = new MemoryPool();
             Libuv = new Libuv();
 
+            var loggerFactory = new LoggerFactory();
```
Tragetaschen: Use DI to get the "real" logger factory instead of creating a local one

src/Microsoft.AspNet.Server.Kestrel/KestrelEngine.cs
```diff
@@ -21,6 +23,10 @@ public KestrelEngine(ILibraryManager libraryManager)
...
+            loggerFactory.AddConsole();
+            Logger = loggerFactory.Create(typeof(KestrelEngine).FullName);
```
Tragetaschen: Why not use the generic `Create<>` extension method?

src/Microsoft.AspNet.Server.Kestrel/project.json
```diff
@@ -2,7 +2,9 @@
     "version": "1.0.0-*",
     "description": "ASP.NET 5 cross platform development web server.",
     "dependencies": {
-        "Microsoft.Framework.Runtime.Interfaces": "1.0.0-*"
+		"Microsoft.Framework.Runtime.Interfaces": "1.0.0-*",
```
Tragetaschen: Fix indentation

src/Microsoft.AspNet.Server.Kestrel/KestrelEngine.cs
ziyasal: Updated to get logger from via DI.

src/Microsoft.AspNet.Server.Kestrel/KestrelEngine.cs
ziyasal: Updated to use generic ```Create<>``` extension method.

src/Kestrel/project.json
```diff
@@ -3,7 +3,8 @@
     "description": "ASP.NET 5 cross platform development web server.",
     "dependencies": {
         "Microsoft.AspNet.Hosting": "1.0.0-*",
-        "Microsoft.AspNet.Server.Kestrel": "1.0.0-*"
+		"Microsoft.AspNet.Server.Kestrel": "1.0.0-*",
```
Eilon: Tabs --> spaces

src/Kestrel/ServerFactory.cs
```diff
@@ -36,7 +40,8 @@ public IDisposable Start(IServerInformation serverInformation, Func<IFeatureColl
         {
             var disposables = new List<IDisposable>();
             var information = (ServerInformation)serverInformation;
-            var engine = new KestrelEngine(_libraryManager);
+            var loggerFactory = _serviceProvider.GetRequiredService<ILoggerFactory>();
```
Eilon: I'm actually not certain if this'll work, but instead of using the service locator pattern (i.e. calling the service provider directly), can the `ILoggerFactory` be a parameter to the `ServerFactory` constructor and perhaps automatically resolved from DI?

src/Microsoft.AspNet.Server.Kestrel/KestrelEngine.cs
```diff
@@ -8,19 +8,25 @@
 using Microsoft.AspNet.Server.Kestrel.Http;
 using Microsoft.Framework.Runtime;
 using System.IO;
+using Microsoft.Framework.Logging;
+using Microsoft.Framework.Logging.Console;
+using Microsoft.Framework.Runtime.Infrastructure;
```
Eilon: Formatting: sort namespaces

src/Microsoft.AspNet.Server.Kestrel/KestrelEngine.cs
```diff
@@ -8,19 +8,25 @@
...
 
 namespace Microsoft.AspNet.Server.Kestrel
 {
     public class KestrelEngine : IDisposable
     {
 
-        public KestrelEngine(ILibraryManager libraryManager)
+        public KestrelEngine(ILibraryManager libraryManager,ILoggerFactory loggerFactory)
```
Eilon: Formatting: space after the `,`

src/Microsoft.AspNet.Server.Kestrel/KestrelEngine.cs
```diff
@@ -66,6 +72,7 @@ public KestrelEngine(ILibraryManager libraryManager)
         public List<KestrelThread> Threads { get; private set; }
         public List<Listener> Listeners { get; private set; }
 
+        public ILogger Logger { get; private set; }
```
Eilon: Newline after this property

src/Microsoft.AspNet.Server.Kestrel/project.json
```diff
@@ -1,9 +1,11 @@
 {
-    "version": "1.0.0-*",
-    "description": "ASP.NET 5 cross platform development web server.",
-    "dependencies": {
-        "Microsoft.Framework.Runtime.Interfaces": "1.0.0-*"
-    },
+	"version": "1.0.0-*",
```
Eilon: Tabs :frowning: :panda_face:

test/Microsoft.AspNet.Server.KestrelTests/EngineTests.cs
```diff
@@ -54,6 +55,14 @@ ILibraryManager LibraryManager
             }
         }
 
+        ILoggerFactory LoggerFactory
+        {
+            get
+            {
+               return new LoggerFactory();
```
Eilon: A property that always returns a new instance seems like an anti-pattern. Could the individual unit tests just create their own logger factories inline?

test/Microsoft.AspNet.Server.KestrelTests/NetworkingTests.cs
```diff
@@ -44,6 +45,14 @@ ILibraryManager LibraryManager
             }
         }
 
+        ILoggerFactory LoggerFactory
+        {
+            get
+            {
+                return new LoggerFactory();
```
Eilon: See earlier comment about this.

test/Microsoft.AspNet.Server.KestrelTests/TestServer.cs
```diff
@@ -40,9 +41,17 @@ ILibraryManager LibraryManager
             }
         }
 
+        ILoggerFactory LoggerFactory
+        {
+            get
+            {
+                return new LoggerFactory();
```
Eilon: And this one.

test/Microsoft.AspNet.Server.KestrelTests/TestServer.cs
```diff
@@ -4,6 +4,7 @@
 using Microsoft.Framework.Runtime.Infrastructure;
 using System;
 using System.Threading.Tasks;
+using Microsoft.Framework.Logging;
```
Eilon: Sort namespaces.

test/Microsoft.AspNet.Server.KestrelTests/project.json
```diff
@@ -2,7 +2,8 @@
     "version": "1.0.0-*",
     "dependencies": {
         "xunit.runner.kre": "1.0.0-*",
-        "Microsoft.AspNet.Server.Kestrel": "1.0.0-*"
+		"Microsoft.AspNet.Server.Kestrel": "1.0.0-*",
```
Eilon: Tabs

msotclas:
__@ziyasal__, Thanks for signing the contribution license agreement so quickly! Actual humans will now validate the agreement and then evaluate the PR.
<br />Thanks, MSOTBOT;

src/Kestrel/ServerFactory.cs
ziyasal: It works well. But, I agree with you, automatically resolve from DI more efficient way.

src/Microsoft.AspNet.Server.Kestrel/project.json
Eilon: This still has tabs.

test/Microsoft.AspNet.Server.KestrelTests/EngineTests.cs
```diff
@@ -54,6 +56,14 @@ ILibraryManager LibraryManager
             }
         }
 
+        ILoggerFactory LoggerFactory
+        {
+            get
+            {
+               return new TestLoggerFactory();
```
Eilon: This is still returning a new instance every time - that is highly unusual for properties because it means if code uses this property twice, it gets a different value. Recommend directly using either `NullLogger`/`Factory` or `TestLogger`/`Factory` from https://github.com/aspnet/Testing/tree/dev/src/Microsoft.AspNet.Testing.Logging instead anyway.

src/Microsoft.AspNet.Server.Kestrel/project.json
Eilon: To be clear, we never use tabs, only four spaces per indentation instead.

src/Kestrel/ServerFactory.cs
```diff
@@ -1,15 +1,16 @@
...
 using Microsoft.AspNet.Builder;
 using Microsoft.AspNet.FeatureModel;
 using Microsoft.AspNet.Hosting.Server;
 using Microsoft.AspNet.Server.Kestrel;
 using Microsoft.Framework.ConfigurationModel;
+using Microsoft.Framework.Logging;
 using Microsoft.Framework.Runtime;
+using System;
+using System.Collections.Generic;
+using System.Threading.Tasks;
```
davidfowl: This is the wrong sort order for usings.

Eilon: Right, we put System.* first, then everything else sorted. There's a VS option for that.

src/Microsoft.AspNet.Server.Kestrel/project.json
ziyasal: Thanks @Eilon , my editor settings causes this problem so I just set it to keep spaces.

test/Microsoft.AspNet.Server.KestrelTests/EngineTests.cs
```diff
@@ -54,6 +56,14 @@ ILibraryManager LibraryManager
             }
         }
 
+        ILoggerFactory LoggerFactory
+        {
+            get
```
Eilon: Sorry if I wasn't clearer in my earlier comment: I'm recommending that these helper properties be completely removed because they don't really add much value. They don't have state and their content is trivial. I recommend referencing `NullLoggerFactory.Instance` directly in the individual unit tests like this:

```c#
var engine = new KestrelEngine(LibraryManager, NullLoggerFactory.Instance);
```

src/Microsoft.AspNet.Server.Kestrel/project.json
```diff
@@ -1,9 +1,11 @@
 {
     "version": "1.0.0-*",
-    "description": "ASP.NET 5 cross platform development web server.",
-    "dependencies": {
-        "Microsoft.Framework.Runtime.Interfaces": "1.0.0-*"
-    },
+	"description": "ASP.NET 5 cross platform development web server.",
```
Eilon: Both the JSON files still have tabs :frowning:

test/Microsoft.AspNet.Server.KestrelTests/EngineTests.cs
ziyasal: I agree with you, use it directly makes sense. Thanks.

src/Microsoft.AspNet.Server.Kestrel/KestrelEngine.cs
```diff
@@ -1,26 +1,31 @@
...
 
-        public KestrelEngine(ILibraryManager libraryManager)
+        public KestrelEngine(ILibraryManager libraryManager, ILoggerFactory loggerFactory)
         {
             Threads = new List<KestrelThread>();
             Listeners = new List<Listener>();
             Memory = new MemoryPool();
             Libuv = new Libuv();
 
+            loggerFactory.AddConsole();
```
Eilon: @lodejard before I merge this, do you have any thoughts on this? Is it OK for Kestrel to register the Console logger? Presumably an app might register the Console logger as well and if it does I think it'll print every message twice (once from each console logger). Maybe that's not really a problem because apps have to ultimately be aware of which hosts they're running on, so they have to know that under Kestrel to not add the console logger.

Thoughts?

davidfowl: Nah this isn't right. The app does this not kestrel. Kestrel can log but it shouldn't be registering providers.

Eilon: So Kestrel out of the box prints nothing to the console? Weaksauce?

davidfowl: Yes. Libraries shouldn't so stuff like this. What happens when the user adds their own console logger. Stuff showed up twice?

Eilon: Not saying Kestrel has to use logging. Should Kestrel just do `Console.WriteLine` instead?

davidfowl: I think that Hosting should log tbh.

Eilon: Sure, but when I think about it now, I think it makes sense for Kestrel to always write to the console (not through a logger). The usability of Kestrel is otherwise fairly difficult especially to a beginner.

Eilon: (And Hosting should probably log a lot more stuff, I agree.)

ziyasal: Just use Console.WriteLine instead of using logger, right?

Eilon: @ziyasal perhaps... waiting to see what @davidfowl or @lodejard think. That would certainly be a lot simpler.

Tratcher: Yeah, Kestrel should not call `AddConsole()`. WebListener deals with this by using Logger.WriteInformation and then we put `AddConsole()` in the `Startup` samples.

I'd recommend against doing Console.WriteLine's, together with or in place of Logger. You want all the information to go to the logger where you can configure the output more directly (to console, disk, remote, etc.). Having this one piece of information take a different rout is just going to make it get lost.

Eilon: Not sure I agree (but I could possibly be convinced). I think everyone running Kestrel will want to see what port it's listening to by default. Sure, maybe there could be a `--quiet` switch or something to shut it up.

src/Microsoft.AspNet.Server.Kestrel/KestrelEngine.cs
```diff
@@ -95,6 +101,7 @@ public IDisposable CreateServer(string scheme, string host, int port, Func<Frame
             {
                 var listener = new Listener(Memory);
                 listener.StartAsync(scheme, host, port, thread, application).Wait();
+                Logger.WriteInformation(string.Format("Server listening on {0}://{1}:{2}", scheme, host, port));
```
Tragetaschen: This console output is a lie, because [the `host` parameter is not used by the listener](https://github.com/aspnet/KestrelHttpServer/blob/dev/src/Microsoft.AspNet.Server.Kestrel/Http/Listener.cs#L58). In the current implementation the server is always listening on `IPAddress.Any`.

davidfowl:
We've decided that we need to centralize this logic in hosting. See https://github.com/aspnet/Hosting/issues/331.

Thanks for the contribution.

Issue #67, opened 2015-02-28
RemoteIpAddress is not set by Kestrel


borgdylan:
Currently, getting the IP address for a request when running on Kestrel yields nothing instead of the remote client's IP.

borgdylan:
I have been working on thsi and will open a PR. @loudej @lodejard

troydai:
The change is nearly done. But we need more tests on various platform to ensure the quality. Move to 1.0.0-rc2.

troydai:
Finished in e9a60610236c53e9518fe80ccf75f14b7e5d02f8

PR #77, opened 2015-03-07
Initial implementation of connection feature

borgdylan:
Signed-off-by: Dylan Borg <borgdylan@hotmail.com>

#67

### DIFF
--- a/src/Kestrel/ServerRequest.cs
+++ b/src/Kestrel/ServerRequest.cs
@@ -5,6 +5,7 @@ using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
+using System.Net;
 using System.Threading.Tasks;
 using Microsoft.AspNet.FeatureModel;
 using Microsoft.AspNet.Http;
@@ -12,13 +13,13 @@ using Microsoft.AspNet.Server.Kestrel.Http;
 
 namespace Kestrel
 {
-    public class ServerRequest : IHttpRequestFeature, IHttpResponseFeature, IHttpUpgradeFeature
+    public class ServerRequest : IHttpRequestFeature, IHttpResponseFeature, IHttpUpgradeFeature, IHttpConnectionFeature
     {
         Frame _frame;
         string _scheme;
         string _pathBase;
         private FeatureCollection _features;
-
+        
         public ServerRequest(Frame frame)
         {
             _frame = frame;
@@ -31,6 +32,7 @@ namespace Kestrel
             _features.Add(typeof(IHttpRequestFeature), this);
             _features.Add(typeof(IHttpResponseFeature), this);
             _features.Add(typeof(IHttpUpgradeFeature), this);
+            _features.Add(typeof(IHttpConnectionFeature), this);
         }
 
         internal IFeatureCollection Features
@@ -217,7 +219,7 @@ namespace Kestrel
             }
         }
 
-        async Task<Stream> IHttpUpgradeFeature.UpgradeAsync()
+        Task<Stream> IHttpUpgradeFeature.UpgradeAsync()
         {
             _frame.StatusCode = 101;
             _frame.ReasonPhrase = "Switching Protocols";
@@ -231,7 +233,65 @@ namespace Kestrel
                 }
             }
             _frame.ProduceStart();
-            return _frame.DuplexStream;
+            return Task.FromResult<Stream>(_frame.DuplexStream);
+        }
+        
+        IPAddress IHttpConnectionFeature.RemoteIpAddress
+        {
+        	get
+        	{
+        		try
+        		{
+	        		return IPAddress.Parse(_frame.RemoteIpAddress);
+	        	}
+	        	catch
+	        	{
+	        		return IPAddress.Any;
+	        	}
+        	}
+        	set
+        	{
+        	}
+        }
+        
+        IPAddress IHttpConnectionFeature.LocalIpAddress
+        {
+        	get
+        	{
+        		try
+        		{
+	        		return IPAddress.Parse(_frame.LocalIpAddress);
+	        	}
+	        	catch
+	        	{
+	        		return IPAddress.Any;
+	        	}
+        	}
+        	set
+        	{
+        	}
+        }
+        
+        int IHttpConnectionFeature.RemotePort { get; set; }
+        
+        int IHttpConnectionFeature.LocalPort { get; set; }
+        
+        bool IHttpConnectionFeature.IsLocal
+        {
+        	get
+        	{
+        		try
+        		{
+	        		return IPAddress.IsLoopback(IPAddress.Parse(_frame.RemoteIpAddress));
+	        	}
+	        	catch
+	        	{
+	        		return false;
+	        	}
+        	}
+        	set
+        	{
+        	}
         }
     }
 }
--- a/src/Kestrel/ServerRequest.cs
+++ b/src/Kestrel/ServerRequest.cs
@@ -19,7 +19,7 @@ namespace Kestrel
         string _scheme;
         string _pathBase;
         private FeatureCollection _features;
-        
+
         public ServerRequest(Frame frame)
         {
             _frame = frame;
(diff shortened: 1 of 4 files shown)


msotclas:
Hi __@borgdylan__, I'm your friendly neighborhood Microsoft Open Technologies, Inc. Pull Request Bot (You can call me MSOTBOT). Thanks for your contribution!
    <span>You've already signed the contribution license agreement. Thanks!</span>
        <p>The agreement was validated by Microsoft Open Technologies, Inc. and real humans are currently evaluating your PR.</p>

TTYL, MSOTBOT;

MartinJohns:
Hey @borgdylan,

A few points I've noticed in your pull request:

- You're mixing spaces and tabs, leading to poorly formatted code. According to the guidelines you should be using 4 spaces for indentation.
- It would be better to use `IPAddress.TryParse()` instead of `IPAddress.Parse()` wrapped in a try-catch block.
- You added empty setters. Was this intentionally? Wouldn't removing them or throwing an `NotSupportedException` better?
- You're using explicit types. The guidelines say to use `var` as the compiler will allow.
- You're adding trailing whitespaces in some locations.
- In Libuv.cs line 342 you added an empty line, whereas all other occurrences of this pattern have none.
- Why are you storing the ips as strings, and not as `IPAddress`? I think parsing them is more expensive than calling `ToString()` on an `IPAddress`.
- Usually comments have a space after the `//` and before the comment.

borgdylan:
They have to be stored as strings and loaded from the frame. This is because the connection can move between different frames as is the case for keep-alive.

borgdylan:
The empty setters are needed to implement the interface. Obviously the operation makes no sense to have in the first place.

borgdylan:
I fixed all that could be fixed without breaking anything.

MartinJohns:
Another minor style issue: https://github.com/aspnet/KestrelHttpServer/pull/77/files#diff-dd545329c8aad9bb5cdb5a93e7fcafd0R90

Here you have two times the opening bracket in the some line as the condition.

src/Kestrel/ServerRequest.cs
```diff
@@ -231,7 +233,53 @@ bool IHttpUpgradeFeature.IsUpgradableRequest
...
+        {
+            get
+            {
+                IPAddress ip;
+                var result = IPAddress.TryParse(_frame.RemoteIpAddress, out ip);
+                return result ? ip : IPAddress.Any;
+            }
+            set
+            {
+                throw new NotSupportedException();
```
Tratcher: We actually do want these to be settable. We want middleware to be able to apply changes from alternate sources like x-forward headers.

src/Kestrel/ServerRequest.cs
```diff
@@ -231,7 +233,53 @@ bool IHttpUpgradeFeature.IsUpgradableRequest
...
+
+        int IHttpConnectionFeature.LocalPort { get; set; }
+
+        bool IHttpConnectionFeature.IsLocal
+        {
+            get
+            {
+                IPAddress ip;
+                var result = IPAddress.TryParse(_frame.RemoteIpAddress, out ip);
+                return result ? IPAddress.IsLoopback(ip) : false;
```
Tratcher: You can non-loopback local connections. Elsewhere we've used `LocalIpAddress.Equals(RemoteIpAddress)`

src/Microsoft.AspNet.Server.Kestrel/Http/Connection.cs
```diff
@@ -63,13 +65,53 @@ public Connection(ListenerContext context, UvStreamHandle socket) : base(context
...
+            var localString = new StringBuilder(50);
+            // storage for errors during decoding processes
+            Exception peerException;
+            Exception localException;
+
+            // get local and peer IPs in binary form
+            _socket.Libuv.tcp_getsockname((UvTcpHandle)_socket, out local, ref localLength);
+            _socket.Libuv.tcp_getpeername((UvTcpHandle)_socket, out peer, ref peerLength);
+            // decode as IPv4 addresses
+            _socket.Libuv.ip4_name(ref local, localString, 50, out localException);
```
Tratcher: It's possible (and preferable) to go directly from a sockaddr to an IPAddress, you don't have to convert to the intermediate string format and then re-parse it first. For example (though the format may not be exactly the same here)
https://github.com/Microsoft/referencesource/blob/c697a4b9782dc8c85c02344a847cb68163702aa7/System/net/System/Net/SocketAddress.cs#L195

src/Microsoft.AspNet.Server.Kestrel/Http/Connection.cs
```diff
@@ -104,9 +146,9 @@ private void OnRead(UvStreamHandle handle, int status, Exception error)
                 }
             }
 
-
             try
             {
+                ExtractIP();
```
Tratcher: Why is ExtractIP called twice?
(I'll still getting familiar with Kestrel's control flow)

src/Kestrel/ServerRequest.cs
borgdylan: You cannot set them since these are bound when a connection is made. It is up to Libuv to select them when the connection is made.

src/Microsoft.AspNet.Server.Kestrel/Http/Connection.cs
borgdylan: Libuv only gives a string for IP addresses. It is not reliable to use its binary version since you will not know if IPv6 or Ipv4 were used while binding the port.

src/Microsoft.AspNet.Server.Kestrel/Http/Connection.cs
borgdylan: When the connection is re-instated, the remote port or local NIC chosen may change.

src/Microsoft.AspNet.Server.Kestrel/Http/Connection.cs
Tratcher: interesting. sockaddr on windows has a built in field telling you if it's IPv4 or IPv6. According to this (http://stackoverflow.com/questions/13157151/isnt-struct-sockadr-in-supposed-to-work-for-both-ipv4-and-ipv6) Linux has a similar built in field.

Besides, the rest of your code is hard-coded to IPv4 and will need to be retrofitted for IPv6 later.

src/Kestrel/ServerRequest.cs
Tratcher: They don't need to be settable back to the underlying connection, they just need to be set locally on ServerRequest which is scoped per request.

src/Microsoft.AspNet.Server.Kestrel/Http/Connection.cs
borgdylan: The sockaddr used is the one provided by Libuv, not the OS ones.

Tragetaschen: The connection is built from a listening socket and this is either IPv4 or IPv6. Why not cache that information and call the appropriate function based on that?

Side note: In my patched version of Kestrel, I'm using a listening socket on `IPAdress.IPv6Any` instead of `IPAddress.Any`, which is a dual-stack socket per default. In that scenario, using the IPv6 version is still the correct option and IPv4-adresses are formatted like `::ffff:192.168.1.2` as expected. I haven't found anything in Kestrel that's not ready for IPv6, libuv [certainly is](http://docs.libuv.org/en/latest/migration_010_100.html#unification-of-ipv4-ipv6-apis).

src/Microsoft.AspNet.Server.Kestrel/Http/Connection.cs
Tragetaschen: Huh? KeepAlive means the underlying TCP connection is not closed between two requests. This means both the remote port and the server IP address cannot change, because then this would be a different TCP connection.

KeepAlive as implemented on `dev` is currently broken. Maybe you were chasing those problems by accident?

src/Kestrel/ServerRequest.cs
Tratcher: I think we're talking past each other.  The I*Feature interfaces represent accessing and storing per-request data for the web app.  While the data (like paths, IPs, etc.) initially come from underlying sources (Frame, Connection, etc.), the web application needs to be able to update all of the fields while processing the request. These updates do not effect the original data source, they are only for use by the higher level application.

For example: My server is hosted behind a reverse proxy. The RemoteIpAddress as reported by Kestrel is always the IP of the reverse proxy.  However, the proxy forwards me the real remote IP in an x-forwarded header. I've written a middleware that will take the value from the x-forwarded header and set it to the RemoteIpAddress field. This way only my middleware needs to be aware of the x-forwarded header, while the rest of the application can just look at the RemoteIpAddress field.  The underlying Connection is never modified, ServerRequest just stores the modified value internally for the duration of the request.

src/Microsoft.AspNet.Server.Kestrel/Http/Connection.cs
Tratcher: Good to hear. Can Libuv (& Linux) use one socket to listen to both IPv6Any and Any? If so we should change to that.

Tragetaschen: Yes, that's the whole point of dual-stack sockets. In libuv, the OS's defaults are harmonized and you have to explicitly disable (!) ([see uv_tcp_bind](http://docs.libuv.org/en/latest/tcp.html#c.uv_tcp_bind)) dual-stack support

Here is an excerpt from my Azure virtual machine running Ubuntu. You can see that port 5004 is only opened as `tcp6`, but the two HEAD requests using IPv4 and IPv6 respectively both succeed:
```
$ netstat -polta
(Not all processes could be identified, non-owned process info
 will not be shown, you would have to be root to see it all.)
Active Internet connections (servers and established)
Proto Recv-Q Send-Q Local Address           Foreign Address         State       PID/Program name Timer
tcp        0      0 *:ssh                   *:*                     LISTEN      -                off (0.00/0/0)
[redacted]
tcp        0    556 [redacted]:ssh          [redacted]              ESTABLISHED -                on (0.06/0/0)
tcp6       0      0 [::]:5004               [::]:*                  LISTEN      26133/mono       off (0.00/0/0)
tcp6       0      0 [::]:ssh                [::]:*                  LISTEN      -                off (0.00/0/0)
$ curl --head http://127.0.0.1:5004/
HTTP/1.1 200 OK
Content-Type: text/html; charset=utf-8
Connection: close

$ curl --head http://[::1]:5004/
HTTP/1.1 200 OK
Content-Type: text/html; charset=utf-8
Connection: close

$
```

davidfowl:
I'm guessing this needs a rebase and a redo /cc @halter73 @lodejard

DamianEdwards:
@borgdylan could you try bringing this PR up to date with the latest source and addressing the comments stated thus far? If you're not going to be able to do this in the next few days, could you let us know as we need to get this feature in for our next release. Thanks!

borgdylan:
I might not be able to do so until the weekend and I do not possess the relevant Libuv knowledge to add the IPv6 stuff. It may be time to pass this to someone else who knows Libuv,IPv6 and sockets more than me. The work I did in this PR was just to make kestrel work for me while Nowin was broken. Since I got Nowin to work I turned away from Kestrel completely to be able to write portable apps without the Libuv dependency.

davidfowl:
We'll take this on then. Thanks for the interest.

PR #89, opened 2015-04-07
Allow Kestrel to boot when on CoreCLR and *NIX

ellismg:
With cross platform .NET Core support coming online, we need to update
our IsWindows check to not assume running on .NET Core means running on
Windows.  Since CoreFX doesn't yet expose a method for doing this (they
are working on adding it back), we'll just call Uname and if that
returns an empty string assume that we are on Windows.

### DIFF
--- a/src/Microsoft.AspNet.Server.Kestrel/Networking/PlatformApis.cs
+++ b/src/Microsoft.AspNet.Server.Kestrel/Networking/PlatformApis.cs
@@ -13,7 +13,9 @@ namespace Microsoft.AspNet.Server.Kestrel.Networking
         public static bool IsWindows()
         {
 #if DNXCORE50
-            return true;
+            // Until Environment.OSVersion.Platform is exposed on .NET Core, we
+            // try to call uname and if that fails we assume we are on Windows.
+            return GetUname() == string.Empty;
 #else
             var p = (int)Environment.OSVersion.Platform;
             return (p != 4) && (p != 6) && (p != 128);


msotclas:
Hi __@ellismg__, I'm your friendly neighborhood Microsoft Open Technologies, Inc. Pull Request Bot (You can call me MSOTBOT). Thanks for your contribution!
    <p>
        It looks like you're a Microsoft contributor (Matt Ellis). If you're full-time, we DON'T require a Contribution License Agreement. If you are a vendor, please DO sign the electronic Contribution License Agreement. It will take 2 minutes and there's no faxing! https://cla2.msopentech.com.
    </p>

TTYL, MSOTBOT;

ellismg:
/cc @davidfowl @stephentoub

The Travis CI build failed, but it looks like it has been failing for the same reason on previous changes.  I don't think it's related to my change?

davidfowl:
:shipit:

davidfowl:
@ellismg don't merge this yet.

ellismg:
@davidfowl Okay.  Just let me know when.

ellismg:
@davidfowl FYI, I booted Kestrel on Windows/CoreCLR using this method (pulled down the PR, used global.json to build it from source and booted HelloWeb).  I even have a "uname" command on my path from cygwin.

I see in the debugger we correctly try to call uname from libc, which fails because it can't be found:

```
0:000> !pe
Exception object: 000000397e238320
Exception type:   System.DllNotFoundException
Message:          Unable to load DLL 'libc': The specified module could not be found. (Exception from HRESULT: 0x8007007E)
InnerException:   <none>
StackTrace (generated):
<none>
StackTraceString: <none>
HResult: 80131524
```

Uname correctly returns the empty string in this case and the upstack check works.

Let me know what else you want me to run before merging.

davidfowl:
:shipit:

PR #118, opened 2015-06-30
Pre-create Continue bytes in Frame

markrendle:
The `HTTP/1.1 100 Continue` header bytes can be `static readonly`, and can be pre-created as an `ArraySegment` instead of creating it for each response.


markrendle:
I'm pretty sure those build fails are not mine. I ran the tests on my machine against Mono/beta6 and all passed.

halter73:
I merged this PR into dev: https://github.com/aspnet/KestrelHttpServer/commit/98995132de817fcec4c02536e6494a32eaf13cb7

Thanks again!

PR #120, opened 2015-07-01
Server address patch

ardalis:
Moving parsing to ServerAddress where it can be tested and doesn't bloat ServerInformation.

Not sure why the project.json.lock files are included in this...


davidfowl:
Can you fix the commit? The lock file and other changes shouldn't be included. Maybe you did a merge before.

ardalis:
I accidentally started on master with this PR. Delete and use PR #121 instead.

PR #122, opened 2015-07-05
Create an unsafe native methods type

Tragetaschen:
This is a first batch of changes from #60 that untangles the type that is `Libuv`.

The general theme is to move all the native methods into `UnsafeNativeMethods` and reduce the number of indirections from the handles into the functions. Basically libuv's functions are now called directly from the handles.

I have cherry-pick'ed and adapted the commits from my branch and didn't squash yet. There are some intermediate commits that don't build.


davidfowl:
What's the point of this change? Is it mostly a style change?

Tragetaschen:
Yes it's a style change. I'm trying to get to something where the remaining work of #60 can be based upon without starting from scratch.

halter73:
I don't think we are ready to accept a PR full of mostly style changes unless it's something that clearly improves maintainability.

Removing the `Libuv.cs` methods that wrap native calls arguably hurts maintainability. Calls to `Check` and `Validate` which used to be mostly contained in `Libuv.cs` are now strewn throughout the rest of the code. This change might also make adding unit tests harder in the future since mocking `Libuv` won't be as useful when native static delegates are being invoked directly.

I see that there are other changes in this PR, a lot of which seem unrelated. With 19 commits, it's difficult to tease out specific changes we might want.

If any of the changes fix any bugs or improve functionality someway, it's probably best to file issues describing the bug you intend to fix or the functional change you plan to implement. If you submit PRs that only address a single issue it should be much easier to review and accept your changes.

We are also looking at improving the performance of Kestrel. So if you have any changes that improve perf measurably, we're interested. Again, the more limited the scope of the PR, the easier it is to accept.

Tragetaschen:
Kestrel has a number of problems as it is implemented now. I have done a considerable amount of work to fix those and as a result, I'm probably the only one in the world who can use it with ASP.NET 5 on Linux without any crashes / hangs / whatever.

I have sit patiently for someone to turn to Kestrel from the team, but it was clear that there hasn't been anyone looking into this for over a year now. What you are asking is for me now is to bring in the necessary architectural changes piece by piece. I can only say, 

I don't have the time for this. It's too much to ask.

The code behind #60 exists and is a working web server for ASP.NET 5 and I will continue to use that with the occasional API changes from upstream. I'm creating a (my) fork and don't plan to actively bring back any of the changes in there. Feel free to take whatever you want.

I'm out of here.

Issue #115, opened 2015-06-29
Get NullReferenceException when running kestrel for a while


cnblogs-dudu:
```
# dnx . kestrel
Started

Unhandled Exception: System.NullReferenceException: 
Object reference not set to an instance of an object.
at Microsoft.AspNet.Server.Kestrel.Networking.UvShutdownReq.UvShutdownCb(IntPtr ptr, Int32 status)

# dnx . kestrel
Started
```

Tragetaschen:
#63

shhsu:
I have actually gotten a similar issue

lodejard:
It sounds like the period of time that the async shutdown is "active" needs to be guarded by a gchandle as well. Same as the uvwritereq - handle created when operation begins and freed in the callback.

shhsu:
The reproduction of this problem is simple.
* Build and run HelloMvc app with coreclr on linux (I used Ubuntu)
* Hit the index page with some load test (I used WCat)
Kestrel should crash pretty fast (10~20 seconds)

Note: I seem to be getting similar trouble with Kestrel on coreclr on Windows as well. After hitting HelloMvc on Kestrel on Windows the dnx process would become unresponsive. I will be debugging it.

PR #128, opened 2015-07-14
Fix NPE issue where UvShutdownReq being garbage collected before the shutdown callback executed

shhsu:
Related to issue https://github.com/aspnet/KestrelHttpServer/issues/115


shhsu:
@halter73 @loudej

PR #130, opened 2015-07-15
Surface fatal exceptions that stop the event loop

halter73:
- Request an app Shutdown so KestrelEngine gets disposed
- Ensure Listener.Dispose doesn't deadlock before the engine
  can be disposed
- Rely on the existing logic to rethrow in the fatal error when
  the engine gets disposed


halter73:
@DamianEdwards @lodejard

lodejard:
:shipit:

Issue #117, opened 2015-06-29
Thoughts for SocketOutput fine-tuning


lodejard:
Imagine the state of the SocketOutput had a "buffers pending" list and a "writes sending" count. 

A buffer in this case would be an object that has an IntPtr to native memory, the actual length of that native memory, the offset of active data in the buffer, and the length of active data in the buffer.

The actual IntPtr and length of native memory would come from a pool. That pool would VirtualAlloc a larger contiguous space at once, and hand out references to buffers which are aligned and sized on page boundaries (4k).

When a Write comes into the SocketOutput, if there are "buffers pending" and the final buffer has trailing space between the end of active data and end of native memory, it would first copy as much of the outgoing data (possibly all) into the existing buffer. If there are more bytes to send which have not been copied, a new 4k buffer is acquired from the pool and the process is repeated until all bytes to send are copied to "buffers pending"

If at the end of this process the "writes sending" count is less than three, increment it, and Post a callback on the uv thread. In that callback, all buffers are removed from the "buffers pending" list and sent to uv. When the uv write callback is invoked then the buffers which were written are returned to the pool. Then, if there are "buffers pending" remove them all and send to uv. In either, if a "buffers pending" check reveals that list is empty, decrement the "writes sending" count.

The SocketOutput will need a simple lock object at both of these points to ensure the data structures aren't compromised.

Tragetaschen:
The following is a quick brain dump. It's short, not rude 😊.

- What's the advantage of pooling allocated memory in a garbage collected environment?
- Why use native memory instead of the blittable `byte[]` and do a copy?
- Isn't there already a lot of buffering going on in the upper layers? (buffer bloat)

benaadams:
Some ideas...

Create the buffer pools per Kestrel thread and take memory from and give back to the pool on the thread associated with the socket; which should encourage CPU cache and NUMA node locality.

For [something like](https://github.com/aspnet/KestrelHttpServer/blob/e5144e31396fce3d092f3a97ead4a4769488a9df/src/Microsoft.AspNet.Server.Kestrel/Http/SocketOutput.cs#L42), keep the code the same; but allocate these pooled buffers in chunks larger than 85k so they are allocated in the LoH and outside of regular GC work. Allocate buffers in a x^2-1 batch

For the pool itself you could slice the buffers with array segments and put them onto and remove from a ```ConcurrentQueue```.

[...]

PR #131, opened 2015-07-15
Write behind response buffering

halter73:
- Reduce calls to uv_write by calling it with multiple buffers when possible
- Complete WriteAsync Tasks early when there are less than 64KB buffered
- Add sample that can produce large responses

TODO:
- Determine why this causes kestrel to progressively slow down on mono as more requests are handled.
- Use Marshal.AllocCoTaskMem to allocate buffers.
- Pool buffers.

https://github.com/aspnet/KestrelHttpServer/issues/117


halter73:
@lodejard

lodejard:
:shipit:

Issue #43, opened 2014-12-15
Returns 200 for unhandled exceptions


Tratcher:
When the application throws an exception and it is caught by the server, Kestrel returns the current response as is. See Frame.ExecuteAsync(). It should send a 500 error response instead.

Expected behavior:
- If the headers have been sent:
 - Abort the connection
- If the headers have not been sent:
 - Set the status code to 500
 - Clear the response headers
 - Do not fire the OnSendingHeaders event
 - Send the response

davidfowl:
@muratg I moved this bug to beta6.

PR #132, opened 2015-07-17
Gracefully handle uncaught exceptions thrown from user code when possible

halter73:


halter73:
@Tratcher @lodejard

src/Microsoft.AspNet.Server.Kestrel/Http/Frame.cs
```diff
@@ -317,6 +317,30 @@ public void ProduceStart()
...
+                if (_resultStarted)
+                {
+                    // We can no longer respond with a 500, so we simply close the connection.
+                    ConnectionControl.End(ProduceEndType.SocketDisconnect);
+                    return;
+                }
+                else
+                {
+                    StatusCode = 500;
+
```
Tratcher: You should also clear/set ReasonPhrase.

test/Microsoft.AspNet.Server.KestrelTests/EngineTests.cs
```diff
@@ -57,6 +57,20 @@ ILibraryManager LibraryManager
...
+        {
+            // Anything added to the ResponseHeaders dictionary is ignored
+            frame.ResponseHeaders["Content-Length"] = new[] { "11" };
+            throw new Exception();
+        }
+
+        private async Task AppThatThrowsAfterWrite(Frame frame)
+        {
+            frame.ResponseHeaders["Content-Length"] = new[] { "11" };
+            await frame.ResponseBody.WriteAsync(Encoding.UTF8.GetBytes("Hello World"), 0, 11);
```
Tratcher: Because you specified Content-Length and then wrote that many bytes, the response is technically finished after this line. You should have an additional test that throws after having written only half of the data (and/or a response without Content-Length).

Tratcher:
:shipit:

halter73:
Fixes #43

src/Microsoft.AspNet.Server.Kestrel/Http/Frame.cs
```diff
@@ -317,6 +317,31 @@ public void ProduceStart()
...
+                    ReasonPhrase = null;
+
+                    // If OnStarting hasn't been triggered yet, we don't want to trigger it now that
+                    // the app func has failed. https://github.com/aspnet/KestrelHttpServer/issues/43
+                    _onStarting = null;
+
+                    ResponseHeaders.Clear();
+                    if (_keepAlive)
+                    {
+                        ResponseHeaders["Content-Length"] = new[] { "0" };
```
lodejard: Would it hurt to set content-length all of the time for response consistency?

lodejard:
:shipit:

src/Microsoft.AspNet.Server.Kestrel/Http/Frame.cs
halter73: I thought that would prevent "Connection: close" from being automatically sent, but that's not the case. I'll change this to be written all the time.

BTW Do you think we should store `new[] { "0" }` in a static field?

halter73:
Fixed by 4a9515d2e00ac11fcc0fe62530bc495042aa944b and 6da15a65b2f7a304e58e178dea48f2aa402ebcd8

PR #134, opened 2015-07-19
Fix sample perf (Option 2)

benaadams:
Don't cause the await to marshal back to the current SynchronizationContext. [ref](http://blogs.msdn.com/b/pfxteam/archive/2012/04/13/10293638.aspx)


benaadams:
Also see #135

halter73:
Closing now that Option 1 is merged into dev.

PR #135, opened 2015-07-19
Fix sample perf (Option 1)

benaadams:
No need to await just return Task


benaadams:
Also see #134

halter73:
As I'm sure you're already aware, this sample isn't really meant to be fast with the call to `Console.WriteLine` and all. Nevertheless, I don't see anything wrong with this change, so I'm going to merge this and close #134.

Issue #25, opened 2014-07-24
Exception in OnSendingHeaders not handled resulting in request hangs


Praburaj:
Currently exceptions in `OnSendingHeaders` are not handled in `Kestrel` resulting in request hangs when an exception is thrown. 

I ran into this issue with https://github.com/aspnet/Security/issues/28. Kestrel in this case is resulting in a request hang.

halter73:
@muratg I just submitted a PR out that fixes this, but I don't think it's worth pushing through ask mode to get this in beta 6. Does punting this to beta 7 sound OK?

muratg:
Yup, moving.

glennc:
@halter73 Is this PR still good to go?

halter73:
@glennc Yeah. I just opened a new PR. I didn't notice that deleting the throw-500 branch automatically closed the old PR. Thanks for the heads up.

PR #137, opened 2015-07-20
Gracefully handle exceptions thrown from OnStarting callbacks

halter73:
- If OnStarting is being called after the app func has completed, return a 500.
- If Onstarting is being called due to a call to write, throw from write.


halter73:
Fixes #25

PR #145, opened 2015-07-24
Multiple threads processing connections

lodejard:
Uses named pipes to dispatch connected sockets to multiple thread loops

Currently uses the number of processors from Environment to create threads. Should be a server setting in the future.


src/Microsoft.AspNet.Server.Kestrel/Networking/Libuv.cs
```diff
@@ -58,6 +58,8 @@ public void Load(string dllToLoad)
 
         public int Check(int statusCode)
         {
+            var x = Marshal.GetLastWin32Error();
```
Tragetaschen: What's that doing?

src/Microsoft.AspNet.Server.Kestrel/Networking/Libuv.cs
```diff
@@ -85,7 +87,7 @@ public int Check(int statusCode, out Exception error)
 
         [UnmanagedFunctionPointer(CallingConvention.Cdecl)]
         delegate int uv_loop_init(UvLoopHandle a0);
-        uv_loop_init _uv_loop_init;
+        uv_loop_init _uv_loop_init = default(uv_loop_init);
```
Tragetaschen: I had just suppressed the warning for that file…

src/Microsoft.AspNet.Server.Kestrel/Http/Listener.cs
```diff
@@ -53,7 +53,7 @@ public Listener(IMemoryPool memory)
                     ListenSocket = new UvTcpHandle();
                     ListenSocket.Init(Thread.Loop, Thread.QueueCloseHandle);
                     ListenSocket.Bind(new IPEndPoint(IPAddress.Any, port));
-                    ListenSocket.Listen(10, _connectionCallback, this);
+                    ListenSocket.Listen(128, _connectionCallback, this);
```
Tragetaschen: The number of listeners is based on the processor count. Is there anything better than a magic number that's also repeated a couple of times later?

src/Microsoft.AspNet.Server.Kestrel/Http/ListenerSecondary.cs
```diff
@@ -0,0 +1,117 @@
...
+                            try
+                            {
+                                var ptr = Marshal.AllocHGlobal(16);
+                                var buf = Thread.Loop.Libuv.buf_init(ptr, 16);
+
+                                DispatchPipe.ReadStart(
+                                    (_1, _2, _3) => buf,
+                                    (_1, status2, error2, state2) =>
+                                    {
+                                        if (status2 <= 0)
```
Tragetaschen: Right now, the stream's read implementation will only ever give 0 (for any kind of error) or >0 for successful reads

src/Microsoft.AspNet.Server.Kestrel/Networking/UvWriteReq.cs
```diff
@@ -81,6 +81,54 @@ public void Init(UvLoopHandle loop)
             }
         }
 
+        public unsafe void Write2(
```
Tragetaschen: So much copy + paste for just a '2'

test/Microsoft.AspNet.Server.KestrelTests/MultipleLoopTests.cs
```diff
@@ -0,0 +1,251 @@
...
+        {
+
+            var loop = new UvLoopHandle();
+            loop.Init(_uv);
+
+            var serverConnectionPipe = default(UvPipeHandle);
+
+            var serverListenPipe = new UvPipeHandle();
+            serverListenPipe.Init(loop, false);
+            serverListenPipe.Bind(@"\\.\pipe\ServerPipeListenForConnections");
```
Tragetaschen: "Wrong" name for the pipe

src/Microsoft.AspNet.Server.Kestrel/Networking/Libuv.cs
lodejard: Ah - leftover code - wasn't meant to be pushed

src/Microsoft.AspNet.Server.Kestrel/Http/Listener.cs
lodejard: That's the socket listen backlog number, what would be better? an integer constant somewhere?

src/Microsoft.AspNet.Server.Kestrel/Networking/Libuv.cs
lodejard: either works :)

kind of nice to see warnings if other code in the file references an uninitialized thing

src/Microsoft.AspNet.Server.Kestrel/Http/ListenerSecondary.cs
```diff
@@ -0,0 +1,117 @@
...
+    public class ListenerSecondary : ListenerContext, IDisposable
+    {
+        UvPipeHandle DispatchPipe { get; set; }
+
+        public ListenerSecondary(IMemoryPool memory)
+        {
+            Memory = memory;
+        }
+
+        public Task StartAsync(
```
Tragetaschen: I think I have convinced myself, that this code does not have a resource leak in the exceptional cases, but it's really, really hard to get to that conclusion. I know it's asynchronous code, but is there any way to streamline the control flow?

src/Microsoft.AspNet.Server.Kestrel/Http/Listener.cs
```diff
@@ -64,13 +64,18 @@ public Listener(IMemoryPool memory)
             return tcs.Task;
         }
 
-        private void OnConnection(UvStreamHandle listenSocket, int status)
+        protected void OnConnection(UvStreamHandle listenSocket, int status)
```
Tragetaschen: I don't think the primary or secondary listener uses that.

test/Microsoft.AspNet.Server.KestrelTests/Program.cs
```diff
@@ -8,9 +11,21 @@ namespace Microsoft.AspNet.Server.KestrelTests
...
+        {
+            this.env = env;
+            this.sp = sp;
+        }
+
+        public int Main()
         {
-            new EngineTests().DisconnectingClient().Wait();
+            return new Xunit.Runner.AspNet.Program(env, sp).Main(new string[] {
+                "-class",
```
halter73: Might be nice to take an argument for the test class name since this seems pretty useful in general.

src/Microsoft.AspNet.Server.Kestrel/Networking/UvWriteReq.cs
lodejard: yeah, but didn't want to have one code path if it meant passing a null safehandle for the non-'2' case

plus this'll probably change a bit in the future as buffer stuff is profiled

src/Microsoft.AspNet.Server.Kestrel/Http/Listener.cs
Tragetaschen: That would be a good start :)

Also there could be a comment as to why that's a good value.

src/Microsoft.AspNet.Server.Kestrel/Http/ListenerPrimary.cs
```diff
@@ -0,0 +1,88 @@
...
+        {
+            var index = _dispatchIndex++ % (_dispatchPipes.Count + 1);
+            if (index == _dispatchPipes.Count)
+            {
+                base.DispatchConnection(socket);
+            }
+            else
+            {
+                var dispatchPipe = _dispatchPipes[index];
+                var write = new UvWriteReq();
```
halter73: Should we dispose this?

Tragetaschen: That's done in the callback below.

halter73: I now see that dispose is already being called in the callback. I leave this here as a reminder to calle write2.Dispose() to avoid the closure allocation.

src/Microsoft.AspNet.Server.Kestrel/Http/ListenerSecondary.cs
lodejard: I'm sure there is - I'll file that as another issue in the interest of getting this PR merged.

src/Microsoft.AspNet.Server.Kestrel/Networking/UvPipeHandle.cs
```diff
@@ -0,0 +1,40 @@
...
+
+            _uv.pipe_init(loop, this, false);
+        }
+
+        public void Bind(string name)
+        {
+            _uv.pipe_bind(this, name);
+        }
+
+        //public void Open(IntPtr hSocket)
```
halter73: Remove commented code

src/Microsoft.AspNet.Server.Kestrel/project.json
```diff
@@ -6,7 +6,7 @@
         "url": "git://github.com/aspnet/kestrelhttpserver"
     },
     "dependencies": {
-        "Microsoft.Framework.Runtime.Abstractions": "1.0.0-*"
+        "Microsoft.Framework.Runtime.Abstractions": "1.0.0-beta7-*"
```
halter73: Why do we need to specify beta7 here?

src/Microsoft.AspNet.Server.Kestrel/KestrelEngine.cs
```diff
@@ -91,16 +91,37 @@ public void Dispose()
 
         public IDisposable CreateServer(string scheme, string host, int port, Func<Frame, Task> application)
         {
-            var listeners = new List<Listener>();
+            var listeners = new List<IDisposable>();
 
             try
             {
+                var pipeName = (Libuv.IsWindows ? @"\\.\pipe\kestrel_" : "/tmp/kestrel_") + Guid.NewGuid().ToString("n");
```
halter73: I've verified that both pipeName's work on OS X. I think the logic where it defaults to "/tmp/..." for non-Windows OSs is good.

PR #150, opened 2015-07-30
Install Libuv on Travis

BrennanConroy:
@halter73

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,12 @@
 language: csharp
 sudo: false
+install:
+  - curl -sSL https://github.com/libuv/libuv/archive/v1.4.2.tar.gz | tar zxfv - -C /tmp && cd /tmp/libuv-1.4.2/
+  - sh autogen.sh
+  - ./configure --prefix=$HOME/libuvinstall
+  - make
+  - make install
+  - export LD_LIBRARY_PATH="$HOME/libuvinstall/lib"
+  - cd $OLDPWD
 script:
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -1,4 +1,12 @@
 language: csharp
-sudo: false
+sudo: required
+install:
+  - sudo apt-get install automake libtool
+  - curl -sSL https://github.com/libuv/libuv/archive/v1.4.2.tar.gz | tar zxfv - -C /tmp && cd /tmp/libuv-1.4.2/
```
halter73: So I assume libuv-1.4.2 is available via apt-get?

.travis.yml
```diff
@@ -1,4 +1,12 @@
 language: csharp
-sudo: false
+sudo: required
```
halter73: This seems to prevent travis from [using containers](http://docs.travis-ci.com/user/workers/container-based-infrastructure/) which might slow down our builds.

Is there any reason that libuv needs to be installed globally? Couldn't we just install it anywhere in the path such as the working directory?

akoeplinger: +1. We're pushing folks towards the container infrastructure as it's much faster and more reliable. Installing libuv locally without sudo should work fine.

BrennanConroy:
Ok, no more sudo!

halter73:
:shipit:

BrennanConroy:
be06eec7ba3b4c971fe0121faeea379d566fded5

Issue #138, opened 2015-07-21
Automatically add "Server" header to responses


halter73:
This is required by [TechEmpower benchmarks](https://www.techempower.com/benchmarks/#section=code).

Daniel15:
Is it worth adding something like `X-Powered-By` too? The server header is usually overwritten by reverse proxies (eg. Nginx).

benaadams:
Also ```X-AspNet-Version``` however, both ```X-Powered-By``` and ```X-AspNet-Version``` should be able to be switched off. This is probably a good reference to be aware of: https://scotthelme.co.uk/hardening-your-http-response-headers/

Daniel15:
I like X-Powered_By  because it helps in estimating how many people are using a particular technology, but I do agree that it should be possible to turn it off.

Daniel15:
I like `X-Powered-By` and `Server` headers because they help in estimating how many people are using a particular technology, but I do agree that it should be possible to turn it off. Maybe have three options:
1.  Full name + version (eg. `Server: Kestrel/1.2.3456` and `X-Powered-By: ASP.NET 5.0.xxxxx`)
2. Name without version (eg. `Server: Kestrel` and `X-Powered-By: ASP.NET`)
3. Nothing at all

benaadams:
However; you'd want the fast path to be with them disabled as that's what you'd want to test on ```plaintext``` as it would push the header to payload ratio to heavy disadvantage. Equally you'd want them to default to on as I don't think anyone would go out of their way to switch them on.

So at that point I assume its more of a PR/Marketing question around what approach you want for internet server stat measurers

benaadams:
Maybe versions/powered by could be middleware like [CORS](https://github.com/aspnet/CORS) and be added by default on file->new like static files; but not added on New->Empty?

halter73:
I guess marked marked this done too soon:

- I think adding an `X-Powered-By` header makes sense if the `Server` header is usually overridden by proxies. It can be removed/edited like the other headers by modifying the header dictionary. Do we think `X-Powered-By: ASP.NET 5.0.xxxxx` is the best default?
- I think we all agree `Server` header should have a version by default.
- I think `X-AspNet-Version` should be added by either the Hosting, MVC or some other middleware like @benaadams suggests. This is something we want for Helios and WebListener as well.

This issue is assigned to the beta7 milestone meaning it should be finished by this Friday. I think we should create another issue to track anything that we don't decide on by then.

[...]

PR #151, opened 2015-07-30
Include Server and Date in the initial response header dictionary

halter73:
Should we do the same thing for the Date header required by [TechEmpower benchmarks](https://www.techempower.com/benchmarks/#section=code)? (Edit: Done)

Addresses https://github.com/aspnet/KestrelHttpServer/issues/138


benaadams:
[RFC 2616 http/1.1 14.18 Date](http://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html)

> Origin servers MUST include a Date header field in all responses, except in these cases:

>      1. If the response status code is 100 (Continue) or 101 (Switching
         Protocols), the response MAY include a Date header field, at
         the server's option.
      2. If the response status code conveys a server error, e.g. 500
         (Internal Server Error) or 503 (Service Unavailable), and it is
         inconvenient or impossible to generate a valid Date.
      3. If the server does not have a clock that can provide a
         reasonable approximation of the current time, its responses
         MUST NOT include a Date header field. In this case, the rules
         in section 14.18.1 MUST be followed.
> A received message that does not have a Date header field MUST be assigned one by the recipient if the message will be cached by that recipient or gatewayed via a protocol which requires a Date. An HTTP implementation without a clock MUST NOT cache responses without revalidating them on every use. An HTTP cache, especially a shared cache, SHOULD use a mechanism, such as NTP [28], to synchronize its clock with a reliable external standard.

halter73:
@benaadams Thanks for the comment. I didn't realize the Date header was required by the HTTP spec.

Issue #156, opened 2015-08-09
Support UNIX sockets


Daniel15:
The common use-case for Kestrel in production will be behind a reverse proxy such as Nginx. In cases where the reverse proxy is located on the same machine as the application, connecting via a UNIX socket is more efficient than a TCP socket, as it avoids going through the network layer. Accessing `127.0.0.1` through TCP still needs to initiate a TCP connection and perform handshaking, checksumming, etc, all of which is avoided by using UNIX sockets.

Please add support for UNIX sockets to Kestrel. Current solutions for running ASP.NET on Linux (Mono's FastCGI server) already support UNIX sockets so it would be good to get this into Kestrel too. As far as I know, libuv does support UNIX sockets as Node.js supports them.

More references:

> UNIX domain sockets should offer better performance than TCP sockets over loopback interface (less copying of data, fewer context switches).

(http://unix.stackexchange.com/questions/91774/performance-of-unix-sockets-vs-tcp-ports)

> The Unix domain sockets are always more efficient and the efficiency benefit is in the 2-3x range. 

(https://nicisdigital.wordpress.com/2014/03/03/unix-domain-sockets-vs-loopback-tcp-sockets/)

markrendle:
+1

On Sun, Aug 9, 2015 at 09:54, aspnet/KestrelHttpServer
<reply@reply.github.com>
wrote:
The common use-case for Kestrel in production will be behind a reverse proxy
such as Nginx. In cases where the reverse proxy is located on the same machine
as the application, connecting via a UNIX socket is more efficient than a TCP
socket, as it avoids going through the network layer. Accessing 127.0.0.1 through TCP still needs to initiate a TCP connection and perform handshaking,
checksumming, etc, all of which is avoided by using UNIX sockets.

Please add support for UNIX sockets to Kestrel. Current solutions for running
ASP.NET on Linux (Mono's FastCGI server) already support UNIX sockets so it
would be good to get this into Kestrel too. As far as I know, libuv does support
UNIX sockets as Node.js supports them.

More references:

UNIX domain sockets should offer better performance than TCP sockets over
loopback interface (less copying of data, fewer context switches).

( 
http://unix.stackexchange.com/questions/91774/performance-of-unix-sockets-vs-tcp-ports
[http://unix.stackexchange.com/questions/91774/performance-of-unix-sockets-vs-tcp-ports] )

The Unix domain sockets are always more efficient and the efficiency benefit is
in the 2-3x range.

( 
[...]

PR #157, opened 2015-08-10
Implement support for UNIX sockets

Daniel15:
The common use-case for Kestrel in production will be behind a reverse proxy such as Nginx. In cases where the reverse proxy is located on the same machine as the application, connecting via a UNIX socket is more efficient than a TCP socket, as it avoids going through the network layer. Accessing 127.0.0.1 through TCP still needs to initiate a TCP connection and perform handshaking, checksumming, etc, all of which is avoided by using UNIX sockets.

 - Moved TCP-specific stuff from Listener into new TcpListener class (same with ListenerPrimary and ListenerSecondary)
 - Made Listener abstract
 - Created new PipeListener. Note that while the use case is for UNIX sockets, this is called "Pipe" in uv, so I've called this "PipeListener" so the terminology is consistant
 - Uses "unix" URL scheme to determine whether to use socket. "http://127.0.0.1:5000" is for listening via TCP while "unix:///var/run/kestrel-test.sock" is for listening via UNIX socket

Example Nginx configuration for proxying to a UNIX socket:
```nginx
server {
  server_name kestreltest.localhost;
  root /var/www/dan.cx/net5test/wwwroot;

  location / {
    # Serve static files directly; proxy everything else to Kestrel
    try_files $uri @aspnet;
  }

  location @aspnet {
    proxy_pass http://unix:/tmp/kestrel-test.sock;
    proxy_set_header Host $host;
  }
}
```

#156

### DIFF
--- a/src/Microsoft.AspNet.Server.Kestrel/Http/Listener.cs
+++ b/src/Microsoft.AspNet.Server.Kestrel/Http/Listener.cs
@@ -1,25 +1,22 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
-using Microsoft.AspNet.Server.Kestrel.Infrastructure;
 using Microsoft.AspNet.Server.Kestrel.Networking;
 using System;
 using System.Diagnostics;
-using System.Net;
 using System.Threading.Tasks;
 
 namespace Microsoft.AspNet.Server.Kestrel.Http
 {
     /// <summary>
-    /// Summary description for Accept
+    /// Base class for listeners in Kestrel. Listens for incoming connections
     /// </summary>
-    public class Listener : ListenerContext, IDisposable
+    /// <typeparam name="T">Type of socket used by this listener</typeparam>
+    public abstract class Listener<T> : ListenerContext, IListener where T : UvStreamHandle
     {
-        private static readonly Action<UvStreamHandle, int, Exception, object> _connectionCallback = ConnectionCallback;
+        protected T ListenSocket { get; private set; }
 
-        UvTcpHandle ListenSocket { get; set; }
-
-        private static void ConnectionCallback(UvStreamHandle stream, int status, Exception error, object state)
+        protected static void ConnectionCallback(UvStreamHandle stream, int status, Exception error, object state)
         {
             if (error != null)
             {
@@ -27,11 +24,11 @@ namespace Microsoft.AspNet.Server.Kestrel.Http
             }
             else
             {
-                ((Listener)state).OnConnection(stream, status);
+                ((Listener<T>)state).OnConnection((T)stream, status);
             }
         }
 
-        public Listener(IMemoryPool memory)
+        protected Listener(IMemoryPool memory)
         {
             Memory = memory;
         }
@@ -51,10 +48,7 @@ namespace Microsoft.AspNet.Server.Kestrel.Http
             {
                 try
                 {
-                    ListenSocket = new UvTcpHandle();
-                    ListenSocket.Init(Thread.Loop, Thread.QueueCloseHandle);
-                    ListenSocket.Bind(new IPEndPoint(IPAddress.Any, port));
-                    ListenSocket.Listen(Constants.ListenBacklog, _connectionCallback, this);
+                    ListenSocket = CreateListenSocket(host, port);
                     tcs.SetResult(0);
                 }
                 catch (Exception ex)
@@ -65,16 +59,19 @@ namespace Microsoft.AspNet.Server.Kestrel.Http
             return tcs.Task;
         }
 
-        private void OnConnection(UvStreamHandle listenSocket, int status)
-        {
-            var acceptSocket = new UvTcpHandle();
-            acceptSocket.Init(Thread.Loop, Thread.QueueCloseHandle);
-            listenSocket.Accept(acceptSocket);
+        /// <summary>
+        /// Creates the socket used to listen for incoming connections
+        /// </summary>
+        protected abstract T CreateListenSocket(string host, int port);
 
-            DispatchConnection(acceptSocket);
-        }
+        /// <summary>
+        /// Handles an incoming connection
+        /// </summary>
+        /// <param name="listenSocket">Socket being used to listen on</param>
+        /// <param name="status">Connection status</param>
+        protected abstract void OnConnection(T listenSocket, int status);
 
-        protected virtual void DispatchConnection(UvTcpHandle socket)
+        protected virtual void DispatchConnection(T socket)
         {
             var connection = new Connection(this, socket);
             connection.Start();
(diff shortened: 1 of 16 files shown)


halter73:
Thanks for the PR. Have you tested Nginx proxying to a UNIX socket with `IKestrelServerInformation.ThreadCount > 1`? It looks like you probably made the necessary changes for this to work, but I want to make sure you tried it out. You can get at the `IKestrelServerInformation` object in your app startup code by casting `IApplicationBuilder.Server`.

Testing this should ensure that our logic to load balance across multiple libuv event loops works with UNIX sockets.

src/Microsoft.AspNet.Server.Kestrel/Http/Listener.cs
```diff
@@ -1,25 +1,22 @@
...
     /// <summary>
-    /// Summary description for Accept
+    /// Base class for listeners in Kestrel. Listens for incoming connections
     /// </summary>
-    public class Listener : ListenerContext, IDisposable
+    /// <typeparam name="T">Type of socket used by this listener</typeparam>
+    public abstract class Listener<T> : ListenerContext, IListener where T : UvStreamHandle
     {
-        private static readonly Action<UvStreamHandle, int, Exception, object> _connectionCallback = ConnectionCallback;
+        protected T ListenSocket { get; private set; }
```
halter73: Any reason this couldn't just be a UvStreamHandle instead of being generic?

Daniel15:
> Have you tested Nginx proxying to a UNIX socket with IKestrelServerInformation.ThreadCount > 1

I believe I did test this, I'll test it out again tonight and confirm though. Unfortunately I forgot to include my test plan in my commit message.

src/Microsoft.AspNet.Server.Kestrel/Http/Listener.cs
Daniel15: It was just in case I needed to call a method specific to the particular implementation. I don't think I ended up needing to do that though, so it could probably just be `UvStreamHandle`.

markrendle:
Have you done a performance comparison between running an Nginx reverse-proxy over the Unix socket vs a TCP socket? If not I'm happy to do it myself and share the results; I just don't want to spend time if you've got numbers already.

Daniel15:
@markrendle I haven't done any perf testing.

Daniel15:
@halter73 - I just tested it again on Debian Linux + Mono 4.0.2 and it worked fine. These were the changes I made when testing:
```diff
diff --git a/samples/SampleApp/Startup.cs b/samples/SampleApp/Startup.cs
index 5f81cbe..fddd1c9 100644
--- a/samples/SampleApp/Startup.cs
+++ b/samples/SampleApp/Startup.cs
@@ -2,6 +2,8 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using Kestrel;
 using Microsoft.AspNet.Builder;
 using Microsoft.AspNet.Http;
 
@@ -11,6 +13,10 @@ public class Startup
     {
         public void Configure(IApplicationBuilder app)
         {
+	        var server = (IKestrelServerInformation) app.Server;
+	        server.ThreadCount = 10;
+			Console.WriteLine("Threads: {0}", server.ThreadCount);
+
             app.Run(async context =>
             {
                 Console.WriteLine("{0} {1}{2}{3}",

```

I also tested on Mac OS, where it seems broken even without my changes (I get "Unknown system error"). I'll submit a separate issue for that.

Daniel15:
I just rebased and rebuilt and now it works on Mac OS X for me. Strange.

halter73:
@markrendle Are you still planning to do the performance comparison between this and proxying via a TCP socket? It would be nice to know since the main benefit of using a UNIX socket seems to be efficiency. Thanks!

davidfowl:
The changes look weird now. I see a bunch of unrelated commits. Can you clean up the PR and make sure the diff only shows your changes?

Daniel15:
Oops, looks like something weird happened when I was merging master into my branch. This commit looks like the right one: https://github.com/Daniel15/KestrelHttpServer/commit/0bb837c4019d861b41daa3f092bf53815673aee3. I'm not very experienced with Git merging so I'll see if I can figure out how to clean it up tomorrow.

Daniel15:
@davidfowl  - It should be fixed now, let me know if it still looks weird for you :)

markrendle:
@halter73 I am still planning to do the perf testing. I'm literally just finishing up on porting a monolithic ASP.NET MVC 5 application to Docker-hosted DNX MVC 6 microservices, then I'll have some time to set up a test environment with Nginx proxying to Kestrel over both TCP and a Socket. I'll post the results when I have them.

src/Microsoft.AspNet.Server.Kestrel/ServerFactory.cs
```diff
@@ -43,7 +43,8 @@ public IDisposable Start(IServerInformation serverInformation, Func<IFeatureColl
             {
                 disposables.Add(engine.CreateServer(
                     address.Scheme,
-                    address.Host,
+                    // Unix sockets use a file path, not a hostname.
+                    address.Scheme == Constants.UnixScheme ? address.Path : address.Host,
```
halter73: How do you specify the address since the host is ignored. Do you use triple forward slash? e.g:

    unix:///tmp/kestrel-test.sock/

Daniel15: Yeah, triple slash. I based that on the `file` URL scheme which uses an empty host section to represent localhost, and on other apps that use `unix:///` as the prefix for UNIX socket paths.

halter73: Can we include the scheme (e.g. http, https, fastcgi) in the address? Sorta like Nginix does in its proxy_pass directive, i.e. http://unix:/tmp/kestrel-test.sock.

lodejard: Yeah, looking at how `unix:` is used it appears it's treated as the "host" information rather than the "scheme". Which kind of makes sense - it determines where to connect/listen rather than how you frame data on the stream after the connection.

Also, if fastcgi is added as a way to frame data - that would likely appear to kestrel as a different scheme which would work on either tcp network sockets or unix domain sockets.

So all of the below would be valid:
* `http://127.0.0.1:8080/`
* `http://unix:/tmp/my-server/:/`
* `fastcgi://127.0.0.1:9090/`
* `fastcgi://unix:/tmp/my-other-server/:/`

So that would involve two changes... Kestrel needs to understand `unix:*:` is a valid host format, allowing everything after the second : to become the path, and then you would look for `unix:` leading characters in the host to tell that it's not a tcp connection.

Daniel15: Is `:` a valid character in the hostname section of a URI? Would the parser allow `http://unix:/tmp/foo` as a valid URI?

Tratcher: No, but we don't use the System.Uri parser for any of this. We have to support wildcards anyways.

halter73: @Daniel15 @lodejard 

[Here is a change to ServerAddress](https://github.com/aspnet/KestrelHttpServer/commit/6b0dfda0aea65a00421c5cab6e278f67aa30385b) that supports your suggested URI syntax.

@lodejard Does this look like what you want? If so, I can start adding badly needed URI parsing tests for this.

Daniel15:
Thanks! :smile:

PR #158, opened 2015-08-11
Fix ListenerSecondary so it can reliably accept new connections on Windows

halter73:
Calling uv_read_start on a named pipe with a larger than necessary buffer
would cause pieces of the next uv_ipc_frame_uv_stream struct to be read into
the uv_read_start buffer when a lot of tcp handles were passed quickly over
the pipe.

This prevented the struct from properly being queued for the next call to
uv_accept to consume. The empty queue caused the call to uv_accept in
ListenerSecondary to fail and return WSAEWOULDBLOCK leaving the connection
in a zombie state.


lodejard:
:shipit: :clap:

Issue #40, opened 2014-12-03
Add keep-alive


borgdylan:
I have tried kestrel and noticed that connections are closed immediately after returning data by using "Connection: close". Would it be possible to add keep-alive functionality to kestrel which would allow sockets to eb kept open and reused instead of doing expensive creates and closes everytime?

DamianEdwards:
Covered by #97

PR #161, opened 2015-08-13
Initial non-optimized support for automatically chunking responses

halter73:
#40

### DIFF
--- a/src/Microsoft.AspNet.Server.Kestrel/Http/Frame.cs
+++ b/src/Microsoft.AspNet.Server.Kestrel/Http/Frame.cs
@@ -27,6 +27,7 @@ namespace Microsoft.AspNet.Server.Kestrel.Http
         Mode _mode;
         private bool _responseStarted;
         private bool _keepAlive;
+        private bool _autoChunk;
         private readonly FrameRequestHeaders _requestHeaders = new FrameRequestHeaders();
         private readonly FrameResponseHeaders _responseHeaders = new FrameResponseHeaders();
 
@@ -253,6 +254,12 @@ namespace Microsoft.AspNet.Server.Kestrel.Http
                 {
                     FireOnStarting();
                 }
+
+                if (_autoChunk)
+                {
+                    WriteChunkPrefix(numOctets: 0);
+                    WriteChunkSuffix();
+                }
             }
             catch (Exception ex)
             {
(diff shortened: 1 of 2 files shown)


src/Microsoft.AspNet.Server.Kestrel/Http/Frame.cs
```diff
@@ -235,6 +236,12 @@ private void FireOnCompleted()
             try
             {
                 await Application.Invoke(this).ConfigureAwait(false);
+
+                if (_autoChunk)
+                {
+                    WriteChunkPrefix(numOctets: 0);
```
lodejard: The prefix and suffix combo can probably just be a constant.

src/Microsoft.AspNet.Server.Kestrel/Http/Frame.cs
```diff
@@ -247,11 +254,53 @@ private void FireOnCompleted()
             }
         }
 
-
         public void Write(ArraySegment<byte> data, Action<Exception, object> callback, object state)
         {
             ProduceStart(immediate: false);
-            SocketOutput.Write(data, callback, state);
+
+            if (_autoChunk)
```
lodejard: Watch out for zero-length writes - when chunking is active they can look like an end-of-stream in the middle of the response.

lodejard:
:shipit: - with two notes. only the one about zero-length writes is critical

davidfowl:
Where are the tests?

lodejard:
oh snap

halter73:
@davidfowl I'll be enabling [this ServerTest](https://github.com/aspnet/ServerTests/blob/e48cf0a91bd9915261301565cd8d99ab5b0f5961/test/ServerComparison.FunctionalTests/ResponseTests.cs#L65) once this is merged. I'll add an EngineTest too. Including one with a zero length write.

src/Microsoft.AspNet.Server.Kestrel/Http/Frame.cs
```diff
@@ -265,11 +271,74 @@ private void FireOnCompleted()
...
+            {
+                if (data.Count == 0)
+                {
+                    callback(null, state);
```
davidfowl: named parameters..........

src/Microsoft.AspNet.Server.Kestrel/Http/Frame.cs
```diff
@@ -265,11 +271,74 @@ private void FireOnCompleted()
...
+        private void WriteChunkPrefix(int numOctets)
+        {
+            var numOctetBytes = CreateAsciiByteArraySegment(numOctets.ToString("x") + "\r\n");
+
+            SocketOutput.Write(numOctetBytes,
+                    (error, _) =>
+                    {
+                        if (error != null)
+                        {
+                            Trace.WriteLine("WriteChunkPrefix" + error.ToString());
```
davidfowl: :disappointed:

davidfowl:
:shipit:

PR #167, opened 2015-08-19
[DESIGN] Damianedwards/date header perf

DamianEdwards:
Design review please.

@halter73 @lodejard @davidfowl @Tratcher

### DIFF
--- a/src/Microsoft.AspNet.Server.Kestrel/Http/DateHeaderValueManager.cs
+++ b/src/Microsoft.AspNet.Server.Kestrel/Http/DateHeaderValueManager.cs
@@ -3,14 +3,19 @@
 
 using System;
 using System.Threading;
+using Microsoft.AspNet.Server.Kestrel.Infrastructure;
 
 namespace Microsoft.AspNet.Server.Kestrel.Http
 {
+    /// <summary>
+    /// Manages the generation of the date header value.
+    /// </summary>
     public class DateHeaderValueManager : IDisposable
     {
         private readonly TimeSpan _timeWithoutRequestsUntilIdle = TimeSpan.FromSeconds(10);
         private readonly TimeSpan _timerInterval = TimeSpan.FromSeconds(1);
         private readonly uint _timerTicksWithoutRequestsUntilIdle;
+        private readonly ISystemClock _systemClock;
 
         private volatile string _dateValue;
         private bool _isDisposed = false;
@@ -20,7 +25,14 @@ namespace Microsoft.AspNet.Server.Kestrel.Http
         private int _timerTicksSinceLastRequest;
 
         public DateHeaderValueManager()
+            : this(new SystemClock())
         {
+
+        }
+
+        public DateHeaderValueManager(ISystemClock systemClock)
+        {
+            _systemClock = systemClock;
             _timerTicksWithoutRequestsUntilIdle = (uint)(_timeWithoutRequestsUntilIdle.TotalMilliseconds / _timerInterval.TotalMilliseconds);
         }
 
@@ -28,6 +40,8 @@ namespace Microsoft.AspNet.Server.Kestrel.Http
         {
             PumpTimer();
 
+            // The null-coalesce here is to protect against this getting called after Dispose() is called, at which point
+            // _dateValue will be null forever more.
             return _dateValue ?? DateTime.UtcNow.ToString("r");
         }
         
@@ -47,6 +61,9 @@ namespace Microsoft.AspNet.Server.Kestrel.Http
         {
             _hadRequestsSinceLastTimerTick = true;
 
+            // If we're already disposed we don't care about starting the timer again. This avoids us having to worry
+            // about requests in flight during dispose (not that that should actually happen) as those will just get
+            // DateTime.UtcNow (aka "the slow way").
             if (!_isDisposed && _dateValueTimer == null)
             {
                 lock (_timerLocker)
@@ -61,19 +78,21 @@ namespace Microsoft.AspNet.Server.Kestrel.Http
             }
         }
 
+        // Called by the Timer (background) thread
         private void UpdateDateValue(object state)
         {
+            // See http://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.18 for required format of Date header
             _dateValue = DateTime.UtcNow.ToString("r");
 
             if (_hadRequestsSinceLastTimerTick)
             {
-                // We served requests since the last tick, just return
+                // We served requests since the last tick, just return as we're still active
                 _hadRequestsSinceLastTimerTick = false;
                 _timerTicksSinceLastRequest = 0;
                 return;
             }
 
-            // No requests since the last timer tick
+            // No requests since the last timer tick, we need to check if we're beyond the idle threshold
             _timerTicksSinceLastRequest++;
             if (_timerTicksSinceLastRequest == _timerTicksWithoutRequestsUntilIdle)
             {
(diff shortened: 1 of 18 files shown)


src/Microsoft.AspNet.Server.Kestrel/Http/DateHeaderValueManager.cs
```diff
@@ -0,0 +1,101 @@
...
+
+            if (!_isDisposed && _dateValueTimer == null)
+            {
+                lock (_timerLocker)
+                {
+                    if (!_isDisposed && _dateValueTimer == null)
+                    {
+                        // Immediately assign the date value and start the timer again
+                        _dateValue = DateTime.UtcNow.ToString("r");
+                        _dateValueTimer = new Timer(UpdateDateValue, state: null, dueTime: _timerInterval, period: _timerInterval);
```
Tragetaschen: Why not create the timer in the constructor and just do `Change()` here? Then, `_dateValueTimer != null` is an invariant and wouldn't need to be checked all the time.

Also: Why not use `dueTime: TimeSpan.Zero` to run the handler immediately, which sets the `_dateValue` immediately.

benaadams: > Why not create the timer in the constructor and just do Change() here? Then, _dateValueTimer != null is an invariant and wouldn't need to be checked all the time.

```_dateValueTimer == null``` is the thing determining whether it needs to be switched on again. Could use boolean with ```_dateValueTimer.Change()``` but that may make the logic more complicated?

>   Why not use dueTime: TimeSpan.Zero to run the handler immediately

That would queue the callback to run immediately; but doesn't call it as part of the ```Timer``` constructor so the initialisation of the value would be in a race with the function return; which it would probably always loose. Though that would be fixed by the null coalescing in ```GetDateHeaderValue``` as it would create its own value.

However, not sure switching it off during low request periods vs creating a string every second with a single timer has much benefit as the timer isn't necessarily a real "individual" timer? http://referencesource.microsoft.com/#mscorlib/system/threading/timer.cs,29

> We use a single native timer, supplied by the VM, to schedule all managed timers in the AppDomain.

So in the spirit of
> Why not create the timer in the constructor

Just switch the timer on in constructor and leave it ticking away until shut down? The side effect would be the GC has to clean up a string every second when the server isn't doing anything; but its happy doing that when the server is fully loaded; and the logic becomes much simpler in ```UpdateDateValue``` and you can drop ```PumpTimer```altogether?

Comment (Tragetaschen) [inline src/Microsoft.AspNet.Server.Kestrel/Http/DateHeaderValueManager.cs]: Here is a different thought for the entire implementation that completely removes any need for synchronization:

http://docs.libuv.org/en/latest/timer.html

There would be one timer per Kestrel thread, but on the other hand the DateTime string wouldn't be shared between processors and every update is already in the "correct" processor's cache.

Comment (benaadams) [inline src/Microsoft.AspNet.Server.Kestrel/Http/DateHeaderValueManager.cs]: The libuv timers are implemented in more or less the same way on [linux](https://github.com/libuv/libuv/blob/bb2632b339d86f69f7dd9aa7168e4b46bf4126a5/src/unix/timer.c#L154) and [win](https://github.com/libuv/libuv/blob/bb2632b339d86f69f7dd9aa7168e4b46bf4126a5/src/win/timer.c#L192) as the .net timer; but you'd pay the added overhead of interop'ing the callback and the code for it would be less clear.

As Kestrel uses multiple (maybe?) libuv threads for socket IO then as you say, this would update in different cpu caches. However, as its a write once per second, read many times the processor cache would only have to refresh once per second (or once every 100k? executions) so it wouldn't be that significant; and it would put extra book keeping on the socket IO threads.

**Comment by benaadams** [comment]
An additional micro-optimization would be to encode and store the header values as byte[] when added to the header dictionary { set }; encoding back to string if read by user code { get }; and using the byte[] when output to the socket (alternate IEnumerable interface?). 

As strings are immutable this wouldn't break user code; to change the header the value would need to be set again. This would mean the server header would only need to be encoded once for the lifetime and date only encoded once per second, rather than both on every response.

However, while this DateTime.UtcNow PR removes a kernel call per response; the encoding cost isn't probably that significant.

File: src/Microsoft.AspNet.Server.Kestrel/Http/DateHeaderValueManager.cs (hunk 1 above)
Comment (DamianEdwards) [inline src/Microsoft.AspNet.Server.Kestrel/Http/DateHeaderValueManager.cs]: There was concern over having any code running when the process is idle as it may interfere with mechanisms that perform operations on or in relation to idle processes. This design allows us to run the timer only when there are requests.

Comment (benaadams) [inline src/Microsoft.AspNet.Server.Kestrel/Http/DateHeaderValueManager.cs]: Process suspension type things? If is a requirement LGTM

**Comment by benaadams** [comment]
Added extra bit to: #169

File: src/Microsoft.AspNet.Server.Kestrel/Http/DateHeaderValueManager.cs (hunk 1 above)
Comment (DamianEdwards) [inline src/Microsoft.AspNet.Server.Kestrel/Http/DateHeaderValueManager.cs]: Yep, that was one of the scenarios.

Comment (benaadams) [inline src/Microsoft.AspNet.Server.Kestrel/Http/DateHeaderValueManager.cs]: That requirement makes this solution rather elegant :+1:

**Comment by halter73** [comment]
I like the design :+1:

**Comment by muratg** [comment]
@DamianEdwards Merge or close?

File: src/Microsoft.AspNet.Server.Kestrel/Http/FrameHeaders.Generated.cs [hunk 2]
```diff
@@ -6,10 +6,10 @@ namespace Microsoft.AspNet.Server.Kestrel.Http
 {
     public partial class FrameResponseHeaders
     {
-        public FrameResponseHeaders()
+        public FrameResponseHeaders(string dateHeaderValue)
         {
             _Server = new[] { "Kestrel" };
-            _Date = new[] { DateTime.UtcNow.ToString("r") };
+            _Date = new[] { dateHeaderValue };
```
Tratcher: This needs to be set at the end of the response rather than the beginning. Things like the error handling middleware will clear the response headers.

DamianEdwards: Yeah, we'll do that as a separate change.

DamianEdwards:
@muratg I'll close it and resubmit after rebasing and adding tests

muratg:
:+1:

muratg:
:watch:

halter73:
https://github.com/aspnet/KestrelHttpServer/pull/207 will conflict with this.

lodejard:
If they're set at the end the app can't see or change them... Should we change the middleware to preserve the server and re-apply the current time?

Tratcher:
Date is required and you can't trust misc middleware to handle it correctly. You can honor their value if they set it, but the server needs to set it at the end if it's missing.

Server is optional, we don't really care if it gets cleared.

src/Microsoft.AspNet.Server.Kestrel/Infrastructure/ISystemClock.cs
```diff
@@ -0,0 +1,18 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+
+namespace Microsoft.AspNet.Server.Kestrel.Infrastructure
+{
+    /// <summary>
+    /// Abstracts the system clock to facilitate testing.
+    /// </summary>
+    public interface ISystemClock
```
davidfowl: internal

DamianEdwards: Closed this one, look at #220 instead

PR #170, opened 2015-08-21
Prevent pipes from being closed prematurely on OS X and Linux

halter73:
This change fixes the in-process marshalling of TCP handles on Mac and Linux
that is used to support having multiple threads accepting connections from
multiple loops.

On these two platforms, the ReadStart callback somtimes gets called with a
status and pipe_pending_count equal to zero. Now when the status is zero
just exit the callback without closing the pipe.

This change more closely follows the example at
https://nikhilm.github.io/uvbook/processes.html#sending-file-descriptors-over-pipes

### DIFF
--- a/src/Microsoft.AspNet.Server.Kestrel/Http/ListenerSecondary.cs
+++ b/src/Microsoft.AspNet.Server.Kestrel/Http/ListenerSecondary.cs
@@ -1,11 +1,12 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
-using Microsoft.AspNet.Server.Kestrel.Networking;
 using System;
 using System.Diagnostics;
 using System.Runtime.InteropServices;
 using System.Threading.Tasks;
+using Microsoft.AspNet.Server.Kestrel.Infrastructure;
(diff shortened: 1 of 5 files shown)


davidfowl:
@halter73 whats the deal with this PR?

halter73:
@davidfowl  I want to get this merged. I've tested it on windows, mac and linux.But no one has reviewed it yet.

src/Microsoft.AspNet.Server.Kestrel/Http/ListenerSecondary.cs
```diff
@@ -61,13 +62,26 @@ protected ListenerSecondary(IMemoryPool memory)
...
+                                                // TODO: Replace Trace.WriteLine with real logging
+                                                Trace.WriteLine("DispatchPipe.ReadStart " + ex.Message);
+                                            }
+
                                             DispatchPipe.Dispose();
                                             Marshal.FreeHGlobal(ptr);
                                             return;
                                         }
 
+                                        if (Thread.Loop.Libuv.pipe_pending_count(DispatchPipe) == 0)
```
lodejard: Add `.PendingCount` to UvPipeHandle

lodejard:
:shipit:

PR #178, opened 2015-08-27
Use ILogger to log message

drussilla:
- KestrelTrace uses ILogger to log messages
- Trace.WriteLine replaced with ILogger (throug KestrelTrace)
- Enable Console logger in SampleApp to test logging through Microsoft.Framework.Logging framework
- Tests modified to support new logging mechanism

### DIFF
--- a/src/Microsoft.AspNet.Server.Kestrel/Http/Connection.cs
+++ b/src/Microsoft.AspNet.Server.Kestrel/Http/Connection.cs
@@ -4,6 +4,7 @@
 using System;
 using Microsoft.AspNet.Server.Kestrel.Networking;
 using System.Diagnostics;
+using Microsoft.Framework.Logging;
 
 namespace Microsoft.AspNet.Server.Kestrel.Http
 {
@@ -75,18 +76,17 @@ namespace Microsoft.AspNet.Server.Kestrel.Http
 
                 if (errorDone && error != null)
                 {
-                    Trace.WriteLine("Connection.OnRead " + error.ToString());
+                    KestrelTrace.Log.LogError("Connection.OnRead " + error, error);
                 }
             }
-
-
+            
             try
             {
                 _frame.Consume();
             }
             catch (Exception ex)
             {
-                Trace.WriteLine("Connection._frame.Consume " + ex.ToString());
+                KestrelTrace.Log.LogError("Connection._frame.Consume " + ex, ex);
             }
         }
 
--- a/src/Microsoft.AspNet.Server.Kestrel/Infrastructure/KestrelTrace.cs
+++ b/src/Microsoft.AspNet.Server.Kestrel/Infrastructure/KestrelTrace.cs
@@ -1,85 +1,117 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
-//using System.Diagnostics.Tracing;
+using System;
+using Microsoft.Framework.Logging;
 
 namespace Microsoft.AspNet.Server.Kestrel
 {
     /// <summary>
     /// Summary description for KestrelTrace
     /// </summary>
-    public class KestrelTrace //: EventSource
+    public class KestrelTrace : ILogger
     {
-        public static KestrelTrace Log = new KestrelTrace();
-  //      static EventTask Connection = (EventTask)1;
-    //    static EventTask Frame = (EventTask)1;
+        public readonly ILogger _logger; 
 
+        public KestrelTrace(ILogger logger)
+        {
+            _logger = logger;
+            Log = this;
+        }
 
-      //  [Event(13, Level = EventLevel.Informational, Message = "Id {0}")]
+        public static KestrelTrace Log;
+  
         public void ConnectionStart(long connectionId)
         {
-         //   WriteEvent(13, connectionId);
+            this.LogDebug(13, $"{nameof(ConnectionStart)}: Id: {connectionId}");
         }
 
       //  [Event(14, Level = EventLevel.Informational, Message = "Id {0}")]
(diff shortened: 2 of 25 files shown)


src/Microsoft.AspNet.Server.Kestrel/Infrastructure/KestrelTrace.cs
```diff
@@ -1,85 +1,109 @@
...
+
+                return _instance;
+            }
+        }
 
-      //  [Event(13, Level = EventLevel.Informational, Message = "Id {0}")]
         public void ConnectionStart(long connectionId)
         {
-         //   WriteEvent(13, connectionId);
+            this.LogDebug(13, $"{nameof(ConnectionStart)} -> Id: {connectionId}");
```
Tratcher: remove `this.`

src/Microsoft.AspNet.Server.Kestrel/project.json
```diff
@@ -7,7 +7,8 @@
     },
     "dependencies": {
         "Microsoft.AspNet.Hosting": "1.0.0-*",
-        "Microsoft.Dnx.Runtime.Abstractions": "1.0.0-*"
+        "Microsoft.Dnx.Runtime.Abstractions": "1.0.0-*",
+        "Microsoft.Framework.Logging": "1.0.0-*"
```
Tratcher: Logging.Abstractions?

src/Microsoft.AspNet.Server.Kestrel/Infrastructure/KestrelTrace.cs
drussilla: Done. Thanks

src/Microsoft.AspNet.Server.Kestrel/project.json
drussilla: Done. Thanks

glennc:
@drussilla are you able to sign the CLA? See https://github.com/aspnet/Home/blob/dev/CONTRIBUTING.md and then go here: https://cla2.dotnetfoundation.org/

drussilla:
@glennc Done!

glennc:
Thanks @drussilla. @lodejard is going to look at this and help get it merged.

halter73:
@drussilla Hi. I just pushed a [commit to dev](https://github.com/aspnet/KestrelHttpServer/commit/f10c989d9034e8cc830db7a4a18008be3b3f84ab) that adds a ServiceContext object that is flowed from KestrelEngine down to ListenerContext and KestrelThread.

It should be possible to add an instance of KestrelTrace to the ServiceContext and copy it into a  new ListenerContext property. Since ListenerContext is ultimately the base class of all the Listeners, Connection and Frame, this should make KestrelTrace accessible pretty much everywhere we need.

The only thing left to that would be to pass the logger when we initialize our Handles.

@drussilla Do you mind updating this PR to avoid the use of the static property?

src/Microsoft.AspNet.Server.Kestrel/Http/Connection.cs
```diff
@@ -75,18 +76,17 @@ private void OnRead(UvStreamHandle handle, int status, Exception error)
 
                 if (errorDone && error != null)
                 {
-                    Trace.WriteLine("Connection.OnRead " + error.ToString());
+                    KestrelTrace.Log.LogError("Connection.OnRead " + error, error);
```
halter73: I think loggers are expected to trace the exception themselves if one is passed as an argument, so I don't think it should be necessary to concatenate the exception to the message string anymore.

Ex: https://github.com/aspnet/Logging/blob/9aa1ab4fc4e1445e1af259de1294e9328fe039ca/src/Microsoft.Framework.Logging.Console/ConsoleLogger.cs#L51

src/Microsoft.AspNet.Server.Kestrel/Infrastructure/KestrelTrace.cs
```diff
@@ -1,85 +1,109 @@
...
         }
 
- //       [Event(12, Message = "Id {0} Status {1}")]
         internal void ConnectionWriteCallback(long connectionId, int status)
         {
-   //         WriteEvent(12, connectionId, status);
+            _logger.LogDebug(12, $"{nameof(ConnectionWriteCallback)} -> Id: {connectionId}, Status: {status}");
+        }
+
+        void ILogger.Log(LogLevel logLevel, int eventId, object state, Exception exception, Func<object, Exception, string> formatter)
```
halter73: Might as well make this public for consistency once we get rid of the static property.

src/Microsoft.AspNet.Server.Kestrel/Infrastructure/KestrelTrace.cs
```diff
@@ -1,85 +1,109 @@
...
-       //     WriteEvent(14, connectionId);
+            _logger.LogDebug(14, $"{nameof(ConnectionStop)} -> Id: {connectionId}");
         }
 
-
-   //     [Event(4, Message = "Id {0} Status {1}")]
         internal void ConnectionRead(long connectionId, int status)
         {
-     //       WriteEvent(4, connectionId, status);
+            _logger.LogDebug(4, $"{nameof(ConnectionRead)} -> Id: {connectionId}, Status: {status}");
```
halter73: This might be a little verbose, even for debug logging.

@loudej Do you have any opinions on this? Should we just not log this at all by default? Is there anything else we shouldn't log by default like like ConnectionReadStart?

drussilla:
@halter73 Hi, Thanks for feedback,
Will work on this PR today.

drussilla:
@halter73 I've messed up with my fork, so I've recreated PR (https://github.com/aspnet/KestrelHttpServer/pull/194). I will delete unnecessary logging if needed after @loudej clarification.

Issue #184, opened 2015-09-01
Style: Kestrel cleanup


glennc:
- [x] Members should have appropriate visibility
- [x] One class per file
- [x] Nested classes or enums at bottom of file
- [x] Constructors come before properties
- [x] Remove all commented code
- [x] Sort namespace
- [x] move declarations to the top of the file
- [x] x is a bad variable name

glennc:
@lodejard Given that all these are ticked does that mean you are done?

DamianEdwards:
@lodejard ping

PR #186, opened 2015-09-03
Code cleanup for issue #184

lodejard:


src/Microsoft.AspNet.Server.Kestrel/Networking/Libuv.cs
```diff
@@ -395,16 +395,19 @@ public uv_buf_t buf_init(IntPtr memory, int len)
...
-            long x0;
-            long x1;
-            long x2;
-            long x3;
+            public sockaddr(long ignored) { x3 = x0 = x1 = x2 = x3 = 0; }
         }
 
         public struct uv_buf_t
         {
+            public IntPtr x0;
```
davidfowl: Fix these variable names

src/Microsoft.AspNet.Server.Kestrel/Networking/Libuv.cs
```diff
@@ -395,16 +395,19 @@ public uv_buf_t buf_init(IntPtr memory, int len)
 
         public struct sockaddr
         {
-            public sockaddr(long ignored) { x3 = x0 = x1 = x2 = x3 = 0; }
+            private long x0;
```
davidfowl: Fix these variable names

src/Microsoft.AspNet.Server.Kestrel/Infrastructure/KestrelThread.cs
```diff
@@ -261,7 +299,8 @@ private void DoPostCloseHandle()
 
         private struct Work
         {
-            public Action<object> Callback;
+            public Action<object, object> Callback1;
```
davidfowl: What is Callback1 and Callback2? Can we get better names?

src/Microsoft.AspNet.Server.Kestrel/Infrastructure/KestrelThread.cs
```diff
@@ -16,16 +16,17 @@ namespace Microsoft.AspNet.Server.Kestrel
...
-        UvLoopHandle _loop;
-        UvAsyncHandle _post;
-        Queue<Work> _workAdding = new Queue<Work>();
-        Queue<Work> _workRunning = new Queue<Work>();
-        Queue<CloseHandle> _closeHandleAdding = new Queue<CloseHandle>();
-        Queue<CloseHandle> _closeHandleRunning = new Queue<CloseHandle>();
-        object _workSync = new Object();
-        bool _stopImmediate = false;
+        private static Action<object, object> _objectCallback = (cb, obj) => ((Action<object>)cb).Invoke(obj);
+        private KestrelEngine _engine;
```
davidfowl: Should be readonly

src/Microsoft.AspNet.Server.Kestrel/Infrastructure/KestrelThread.cs
```diff
@@ -16,16 +16,17 @@ namespace Microsoft.AspNet.Server.Kestrel
...
+        private Thread _thread;
+        private UvLoopHandle _loop;
```
davidfowl: This too. Basically anything assigned and never reassigned.

src/Microsoft.AspNet.Server.Kestrel/Http/MessageBodyExchanger.cs
```diff
@@ -17,11 +17,11 @@ public class MessageBodyExchanger
         private static readonly WaitCallback _completePending = CompletePending;
         protected readonly FrameContext _context;
 
-        object _sync = new Object();
+        private object _sync = new Object();
 
-        ArraySegment<byte> _buffer;
-        Queue<ReadOperation> _reads = new Queue<ReadOperation>();
-        bool _send100Continue = true;
+        private ArraySegment<byte> _buffer;
```
davidfowl: Should be readonly

src/Microsoft.AspNet.Server.Kestrel/Http/MemoryPool.cs
```diff
@@ -121,5 +92,34 @@ public void FreeSegment(ArraySegment<byte> segment)
         {
             FreeByte(segment.Array);
         }
+
+        class Pool<T>
```
davidfowl: Where's the visibility here? And why didn't standards police complain?

src/Microsoft.AspNet.Server.Kestrel/Http/MessageBodyExchanger.cs
lodejard: readonly is a tougher rule to write

src/Microsoft.AspNet.Server.Kestrel/Http/Frame.cs
```diff
@@ -16,32 +16,22 @@ namespace Microsoft.AspNet.Server.Kestrel.Http
...
     public class Frame : FrameContext, IFrameControl
     {
-        enum Mode
-        {
-            StartLine,
-            MessageHeader,
-            MessageBody,
-            Terminated,
-        }
+        private static Encoding _ascii = Encoding.ASCII;
```
davidfowl: readonly

src/Microsoft.AspNet.Server.Kestrel/Networking/Libuv.cs
lodejard: They are 16 bytes with an undefined nature. What's a better name? long0 through long3?

src/Microsoft.AspNet.Server.Kestrel/Http/MemoryPool.cs
lodejard: the police are looking for readonly on field declarations, not nested types...

davidfowl:
@lodejard you need to sign the CLA

src/Microsoft.AspNet.Server.Kestrel/Http/ListenerPrimary.cs
```diff
@@ -15,16 +15,19 @@ namespace Microsoft.AspNet.Server.Kestrel.Http
...
-        ArraySegment<ArraySegment<byte>> _1234 = new ArraySegment<ArraySegment<byte>>(new[] { new ArraySegment<byte>(new byte[] { 1, 2, 3, 4 }) });
+        // this message is passed to write2 because it must be non-zero-length, 
+        // but it has no other functional significance
+        private readonly ArraySegment<ArraySegment<byte>> _dummyMessage = new ArraySegment<ArraySegment<byte>>(new[] { new ArraySegment<byte>(new byte[] { 1, 2, 3, 4 }) });
 
         protected ListenerPrimary(IMemoryPool memory) : base(memory)
         {
         }
 
+        UvPipeHandle ListenPipe { get; set; }
```
davidfowl: Visibility? Public? Internal? Private?

src/Microsoft.AspNet.Server.Kestrel/Http/ListenerSecondary.cs
```diff
@@ -15,13 +15,13 @@ namespace Microsoft.AspNet.Server.Kestrel.Http
...
     public abstract class ListenerSecondary : ListenerContext, IDisposable
     {
-        UvPipeHandle DispatchPipe { get; set; }
-
         protected ListenerSecondary(IMemoryPool memory)
         {
             Memory = memory;
         }
 
+        UvPipeHandle DispatchPipe { get; set; }
```
davidfowl: Same comment, what's the visibility

src/Microsoft.AspNet.Server.Kestrel/Http/MemoryPool.cs
davidfowl: Need to update that thing to just look at all declarations

src/Microsoft.AspNet.Server.Kestrel/Http/SocketOutput.cs
```diff
@@ -216,14 +216,14 @@ private class CallbackContext
 
         private class WriteContext
         {
+            public SocketOutput Self;
```
davidfowl: readonly

global.json
```diff
@@ -1,5 +1,6 @@
 {
     "projects": [
-        "src"
+        "src",
+        "tools"
```
halter73: Will this cause NuGet packages to get created for the tools projects? The reason I created the tools directory in the first place was to avoid that.

src/Microsoft.AspNet.Server.Kestrel/Http/FrameRequestHeaders.cs
```diff
@@ -0,0 +1,57 @@
...
+                _current = default(KeyValuePair<string, StringValues>);
+                _hasUnknown = collection.MaybeUnknown != null;
+                _unknownEnumerator = _hasUnknown
+                    ? collection.MaybeUnknown.GetEnumerator()
+                    : default(Dictionary<string, StringValues>.Enumerator);
+            }
+
+            public KeyValuePair<string, StringValues> Current => _current;
+
+            object IEnumerator.Current => _current;
```
halter73: Why use an explicit interface implementation for only this property?

src/Microsoft.AspNet.Server.Kestrel/Http/FrameResponseHeaders.cs
```diff
@@ -0,0 +1,57 @@
...
+        }
+
+        protected override IEnumerator<KeyValuePair<string, StringValues>> GetEnumeratorFast()
+        {
+            return GetEnumerator();
+        }
+
+        public partial struct Enumerator : IEnumerator<KeyValuePair<string, StringValues>>
+        {
+            private FrameResponseHeaders _collection;
```
halter73: readonly?

src/Microsoft.AspNet.Server.Kestrel/Http/ListenerSecondary.cs
halter73: Do we know why the module isn't catching this? Because it's an abstract class?

src/Microsoft.AspNet.Server.Kestrel/compiler/preprocess/SPCM.cs
```diff
@@ -0,0 +1,6 @@
+﻿namespace Microsoft.AspNet.Server.Kestrel
+{
+    public class StandardsPoliceCompileModule : Microsoft.StandardsPolice.StandardsPoliceCompileModule
```
halter73: Does this have to be public and in the primary namespace?

src/Microsoft.AspNet.Server.Kestrel/compiler/preprocess/SPCM.cs
```diff
@@ -0,0 +1,6 @@
+﻿namespace Microsoft.AspNet.Server.Kestrel
```
halter73: Shouldn't we just name the file StandardsPoliceCompileModule.cs?

tools/Microsoft.StandardsPolice/StandardsPoliceCompileModule.cs
```diff
@@ -0,0 +1,288 @@
...
+            ScanCompilation(context.Diagnostics, context.Compilation);
+        }
+
+        internal static void ScanCompilation(IList<Diagnostic> diagnostics, CSharpCompilation compilation)
+        {
+            ScanNamespace(diagnostics, compilation.GlobalNamespace);
+
+            foreach (var st in compilation.SyntaxTrees)
+            {
+                if (!st.FilePath.EndsWith(".Generated.cs"))
```
halter73: Why skip generated code? I notice you've changed GeneratedCode/KnownHeaders.cs anyway.

tools/Microsoft.StandardsPolice/StandardsPoliceCompileModule.cs
```diff
@@ -0,0 +1,288 @@
...
+                if (memberZone >= ClassZone.OtherThings)
+                {
+                    continue;
+                }
+                if (memberZone < currentZone)
+                {
+                    if (member.Locations.Count() == 1)
+                    {
+                        diagnostics.Add(Diagnostic.Create(
+                            "SP1003", "StandardsPolice", $"{memberZone} like {typeSymbol.Name}::{member.Name} shouldn't be after {currentZone} like {currentZoneExample.Name}",
```
halter73: Nice. This is getting pretty strict :+1:

global.json
halter73: I just checked. This doesn't cause a new package to get created. So you can disregard the above comment.

tools/Microsoft.StandardsPolice/StandardsPoliceCompileModule.cs
lodejard: Because there are more than one type in that file, and it would be harder to change the generation to write multiple files

src/Microsoft.AspNet.Server.Kestrel/Http/FrameResponseHeaders.cs
lodejard: moving readonly to a different issue

src/Microsoft.AspNet.Server.Kestrel/compiler/preprocess/SPCM.cs
lodejard: namespace doesn't matter, probably does need to be public, but doesn't get added to the actual output assembly

Issue #102, opened 2015-05-20
ServerFactory.Start does not cleanup properly if exception after first engine.CreateServer


smbecker:
If multiple addresses are being used in ServerFactory and an exception is thrown in engine.CreateServer, then the proceeding disposables are not cleaned up properly. Consider changing the method implementation to something similar to:

```c#
public IDisposable Start(IServerInformation serverInformation, Func<IFeatureCollection, Task> application)
{
    var disposables = new List<IDisposable>();
    var dispose = new Disposable(() =>
    {
        foreach (var disposable in disposables)
        {
            disposable.Dispose();
        }
    });
    var information = (ServerInformation)serverInformation;
    var engine = new KestrelEngine(_libraryManager);
    engine.Start(1);
    try
    {
        foreach (var address in information.Addresses)
        {
            disposables.Add(engine.CreateServer(
                address.Scheme,
                address.Host,
                address.Port,
                async frame =>
                {
                    var request = new ServerRequest(frame);
                    await application.Invoke(request.Features);
                }));
        }
        disposables.Add(engine);
        return dispose;
    }
    catch (Exception)
    {
        dispose.Dispose();
        throw;
    }
}
```

PR #189, opened 2015-09-03
Clean up if an exception is thrown in the middle of ServerFactory.Start

halter73:
#102

### DIFF
--- a/src/Microsoft.AspNet.Server.Kestrel/ServerFactory.cs
+++ b/src/Microsoft.AspNet.Server.Kestrel/ServerFactory.cs
@@ -36,30 +36,44 @@ namespace Microsoft.AspNet.Server.Kestrel
 
         public IDisposable Start(IFeatureCollection serverFeatures, Func<IFeatureCollection, Task> application)
         {
-            var disposables = new List<IDisposable>();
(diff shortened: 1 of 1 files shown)


src/Microsoft.AspNet.Server.Kestrel/ServerFactory.cs
```diff
@@ -36,30 +36,44 @@ public IFeatureCollection Initialize(IConfiguration configuration)
...
-                    address.Port,
-                    async frame =>
-                    {
-                        var request = new ServerRequest(frame);
-                        await application.Invoke(request.Features).ConfigureAwait(false);
-                    }));
-            }
-            disposables.Add(engine);
-            return new Disposable(() =>
+            var disposables = new Stack<IDisposable>();
```
Tratcher: Why a stack?

Tratcher:
:shipit:

src/Microsoft.AspNet.Server.Kestrel/ServerFactory.cs
halter73: Just to make sure the engine gets disposed last.

lodejard:
:shipit:

PR #190, opened 2015-09-04
Enabled tests on OSX

davidfowl:
Seems like things are failing on OSX. We should work on this branch until we get it green then merge it in.

/cc @halter73 @lodejard


muratg:
@davidfowl more work planned on this?  It's stale :)

Issue #191, opened 2015-09-05
[Beta8] Can't write into HttpResponse.Body (Stream) on Kestrel.


MJomaa:
It was possible to write into that stream in beta7 but fails in beta8 (15564).
The stream is not writeable.
However it works on WebListener [beta8] and IIS [beta8].

`Kestrel:`
![image](https://cloud.githubusercontent.com/assets/9029497/9698742/1cde1fbe-53c4-11e5-98f3-1b5824dca5cb.png)

![image](https://cloud.githubusercontent.com/assets/9029497/9698753/63ebed78-53c4-11e5-8278-e56bf308ffcc.png)

`WebListener:`
![image](https://cloud.githubusercontent.com/assets/9029497/9698757/a777b9f0-53c4-11e5-81ce-8b83b6e66d17.png)

`IIS:`
![image](https://cloud.githubusercontent.com/assets/9029497/9698763/cde1e11a-53c4-11e5-8bc0-7766f7662923.png)

**Repo to reproduce:**
https://github.com/MJomaa/KestrelResponseStreamBug

---
This affects some "CopyToAsync"-related code, for example:
https://github.com/aspnet/Mvc/search?utf8=%E2%9C%93&q=CopyToAsync

For a temp workaround, you can use this:
https://github.com/aspnet/StaticFiles/blob/864ee7f3477c26c9e2a447355f872e1751194220/src/Microsoft.AspNet.StaticFiles/StreamCopyOperation.cs

MJomaa:
Tested the latest build (v1.0.0-beta8-15585) and it's fixed. Thank you!

PR #192, opened 2015-09-05
Fix bug #191 introduced by b25d2d9

Tragetaschen:
CanWrite [was true](https://github.com/aspnet/KestrelHttpServer/commit/b25d2d97721aca7e6137de3619cdc875e37eac77#diff-4d4d276981c056bacc1a401293b08a93L107) before


davidfowl:
Hah @lodejard 's refactoring broke it :smile: . We need a test for this. Can you add one @Tragetaschen ?

davidfowl:
Actually, we'll add that test. Thanks for the PR!

PR #195, opened 2015-09-09
Log message style changes

halter73:
- Increment connection id
- Disabled some log messages for possibly being too verbose.
- Don't log ConnectionReadFin when there is an error.

### DIFF
--- a/src/Microsoft.AspNet.Server.Kestrel/Infrastructure/KestrelTrace.cs
+++ b/src/Microsoft.AspNet.Server.Kestrel/Infrastructure/KestrelTrace.cs
@@ -32,7 +32,7 @@ namespace Microsoft.AspNet.Server.Kestrel
         public void ConnectionRead(long connectionId, int count)
         {
             // Don't log for now since this could be *too* verbose.
-            //_logger.LogVerbose(3, @"Connection id ""{ConnectionId}"" received ""{Count}"" bytes.", connectionId, count);
+            // Reserved: Event ID 3
         }
 
         public void ConnectionPause(long connectionId)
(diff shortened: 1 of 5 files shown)


halter73:
@drussilla Here are some additional logging changes we'd like to make.

drussilla:
@halter73 Looks reasonable.

src/Microsoft.AspNet.Server.Kestrel/Infrastructure/KestrelTrace.cs
```diff
@@ -21,57 +21,65 @@ public KestrelTrace(ILogger logger)
...
         {
-            _logger.LogDebug(10, $"{nameof(ConnectionDisconnect)} -> Id: {connectionId}");
+            _logger.LogDebug(10, @"Connection id ""{ConnectionId}"" disconnected.", connectionId);
         }
 
         public void ConnectionWrite(long connectionId, int count)
         {
-            _logger.LogDebug(11, $"{nameof(ConnectionWrite)} -> Id: {connectionId}, Count: {count}");
+            // Don't log for now since this could be *too* verbose.
+            //_logger.LogVerbose(11, @"Connection id ""{ConnectionId}"" sent ""{Count}"" bytes.", connectionId, count);
```
davidfowl: Don't leave in the commented out code. Just remove it and leave the comment.

halter73: @davidfowl This was to make it a little easier to do some debugging when working on Kestrel itself. I discussed this with @lodejard and he was fine with it. Do you feel strongly about this?

davidfowl: Yes. Dead code must die.

halter73:
Thanks again @drussilla

PR #197, opened 2015-09-09
Display the address and port that is being listened on.

drussilla:
Close aspnet/KestrelHttpServer#32


Tratcher:
We're actually considering handling this generically with https://github.com/aspnet/Hosting/issues/328

drussilla:
Should I close this PR then?

davidfowl:
Yep

PR #198, opened 2015-09-10
Negative thread count fix

benaadams:
Someone will be silly and complain...


davidfowl:
Rather throw an exception so people know they did something stupid.

benaadams:
Now throws

halter73:
Thanks

PR #199, opened 2015-09-10
React to `xunit.runner.aspnet` package updates

dougbu:
- aspnet/aspnet.xunit@5a12e89


dougbu:
FYI only since this fixes a build break.

test/Microsoft.AspNet.Server.KestrelTests/Microsoft.AspNet.Server.KestrelTests.xproj
```diff
@@ -13,5 +13,8 @@
   <PropertyGroup>
     <SchemaVersion>2.0</SchemaVersion>
   </PropertyGroup>
+  <ItemGroup>
+    <Service Include="{82a7f48d-3b50-4b1e-b82e-3ada8210c358}" />
+  </ItemGroup>
```
dougbu: Just let VS do its thing.

dougbu:
fea510f

PR #201, opened 2015-09-10
Don't swallow too many exceptions

halter73:
- Swallowing too many exceptions can end up hiding issues in Kestrel itself. It's better to let the process die.
- If we want to handle certain exceptions we should be as specific as possible with our try/catch blocks.
- Catch and log uv_accept errors

### DIFF
--- a/src/Microsoft.AspNet.Server.Kestrel/Http/Connection.cs
+++ b/src/Microsoft.AspNet.Server.Kestrel/Http/Connection.cs
@@ -93,6 +93,7 @@ namespace Microsoft.AspNet.Server.Kestrel.Http
             catch (Exception ex)
             {
                 Log.LogError("Connection._frame.Consume ", ex);
+                throw;
             }
         }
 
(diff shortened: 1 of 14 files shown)


davidfowl:
Before saying yes to this PR, I'd like us to make sure we know what happens when each of those throws gets hit. Does the server crash hard or is it left in an invalid state?

halter73:
@davidfowl I manually added `throw new Exception();` in each of those try blocks to see what would happen. The server crashes hard in each case after logging the error.

I also see the following assertion failure:

    ThreadId is incorrect
       at Microsoft.AspNet.Server.Kestrel.Networking.UvMemory.Validate(Boolean closed)
       at Microsoft.AspNet.Server.Kestrel.Networking.Libuv.loop_close(UvLoopHandle handle)
       at Microsoft.AspNet.Server.Kestrel.Networking.UvLoopHandle.ReleaseHandle()
       at System.Runtime.InteropServices.SafeHandle.Dispose(Boolean disposing)

This opens an assertion dialog briefly that disappears after about a second.

I'm not sure how important it is to try to shutdown gracefully when facing these kind of unexpected exceptions. The idea is that if an exception is thrown in these places, its probably a bug in Kestrel itself.

davidfowl:
@halter73 That looks pretty bad. Is it an assert?

davidfowl:
:shipit:

Issue #183, opened 2015-09-01
Style: Refactor flow control


glennc:
Use Task values instead of callback plus state managers. Monitor perf as the changes are made.

- Exceptions from read and write become eternal. User mode calls to read and write on the request bodies will be re-thrown.

PR #206, opened 2015-09-17
Updating input/output flow control

lodejard:
See #183 

* Eliminates callbacks and buffer copies
* Utilizes c# async to flow completion and exception
* Introduces large object heap buffer sub-allocations


lodejard:
github is buggy

PR #207, opened 2015-09-17
Updating flow control

lodejard:
See #183 

* Eliminates callbacks and buffer copies
* Utilizes c# async to flow completion and exception
* Introduces large object heap buffer sub-allocations


src/Microsoft.AspNet.Server.Kestrel/Http/SocketOutput.cs
```diff
@@ -211,6 +209,54 @@ private void TriggerCallback(CallbackContext context)
             }, context);
         }
 
+        void ISocketOutput.Write(ArraySegment<byte> buffer, bool immediate)
+        {
+            ((ISocketOutput)this).WriteAsync(buffer, immediate).Wait();
```
davidfowl: `GetAwaiter().GetResult()`

src/Microsoft.AspNet.Server.Kestrel/Http/MessageBodyExchanger.cs
```diff
@@ -119,6 +119,31 @@ public Task<int> ReadAsync(ArraySegment<byte> buffer)
...
+            if (send100Continue)
+            {
+                _context.FrameControl.ProduceContinue();
+            }
+            return result;
+        }
+
+        public virtual Task<int> ReadAsyncImplementation(ArraySegment<byte> buffer, CancellationToken cancellationToken)
+        {
+            throw new NotImplementedException("TODO");
```
davidfowl: What?

src/Microsoft.AspNet.Server.Kestrel/Http/Frame.cs
```diff
@@ -6,9 +6,13 @@
 using System.IO;
 using System.Linq;
 using System.Text;
+using System.Threading;
 using System.Threading.Tasks;
+using Microsoft.AspNet.Server.Kestrel.Infrastructure;
 using Microsoft.Framework.Logging;
 using Microsoft.Framework.Primitives;
+using System.Numerics;
+using Microsoft.AspNet.Hosting.Builder;
```
davidfowl: Sort these usings

src/Microsoft.AspNet.Server.Kestrel/Http/Connection.cs
```diff
@@ -11,13 +12,17 @@ namespace Microsoft.AspNet.Server.Kestrel.Http
 {
     public class Connection : ConnectionContext, IConnectionControl
     {
-        private static readonly Action<UvStreamHandle, int, Exception, object> _readCallback = ReadCallback;
+        private const int EOF = -4095;
+        private const int ECONNRESET = -4077;
```
halter73: Don't need these consts anymore. They're now in the Constants class.

samples/SampleApp/Startup.cs
```diff
@@ -24,9 +24,11 @@ public void Configure(IApplicationBuilder app, ILoggerFactory loggerFactory)
                     context.Request.Path,
                     context.Request.QueryString);
 
+                await context.Request.Body.CopyToAsync(Console.OpenStandardOutput());
```
halter73: Are we keeping this change to the sample?

src/Microsoft.AspNet.Server.Kestrel/Http/Listener.cs
```diff
@@ -13,8 +13,10 @@ namespace Microsoft.AspNet.Server.Kestrel.Http
     /// </summary>
     public abstract class Listener : ListenerContext, IDisposable
     {
-        protected Listener(ServiceContext serviceContext) : base(serviceContext)
+        protected Listener(ServiceContext serviceContext) 
+            : base(serviceContext)
         {
+            Memory2 = new MemoryPool2();
```
davidfowl: Rename this

davidfowl:
Rename Memory*2 to something else and rename the IMemoryPool to something that says "used for output"

src/Microsoft.AspNet.Server.Kestrel/Http/Listener.cs
lodejard: I want @halter73 's PR to rebase before causing larger code churn like that. That's why the original memory pool and this one are temporarily distinct from each other.

src/Microsoft.AspNet.Server.Kestrel/Http/Connection.cs
```diff
@@ -11,13 +12,14 @@ namespace Microsoft.AspNet.Server.Kestrel.Http
...
     {
-        private static readonly Action<UvStreamHandle, int, Exception, object> _readCallback = ReadCallback;
+        private static readonly Action<UvStreamHandle, int, int, Exception, object> _readCallback = ReadCallback;
         private static readonly Func<UvStreamHandle, int, object, Libuv.uv_buf_t> _allocCallback = AllocCallback;
 
         private static long _lastConnectionId;
 
         private readonly UvStreamHandle _socket;
         private Frame _frame;
+        private Task _frameTask;
```
davidfowl: Remove this now since it's never assigned.

src/Microsoft.AspNet.Server.Kestrel/Http/Listener.cs
davidfowl: So this replaces the IMemoryPool in general? They seem unrelated. InputPool and MemoryPool?

src/Microsoft.AspNet.Server.Kestrel/Http/Connection.cs
lodejard: D'oh!

src/Microsoft.AspNet.Server.Kestrel/Http/ListenerSecondary.cs
```diff
@@ -59,7 +59,7 @@ protected ListenerSecondary(ServiceContext serviceContext) : base(serviceContext
 
                                 DispatchPipe.ReadStart(
                                     (_1, _2, _3) => buf,
-                                    (_1, status2, error2, state2) =>
+                                    (_1, status2, errCode, error2, state2) =>
```
davidfowl: wtf is _1, status2, errCode, erro2, state2? I know we have nested lambdas by maybe this is a case for a method...

src/Microsoft.AspNet.Server.Kestrel/Http/MemoryPool2.cs
```diff
@@ -0,0 +1,107 @@
+﻿using System;
+using System.Collections.Concurrent;
+
+namespace Microsoft.AspNet.Server.Kestrel.Http
+{
+    public class MemoryPool2 : IDisposable
```
davidfowl: A brief description of what the algorithm used would be awesome. Bonus points if you add a why

src/Microsoft.AspNet.Server.Kestrel/Http/MemoryPool2.cs
```diff
@@ -0,0 +1,107 @@
...
+    {
+        private const int blockStride = 4096;
```
davidfowl: Reasons for choosing these would be great

src/Microsoft.AspNet.Server.Kestrel/Http/ListenerSecondary.cs
lodejard: _1 _2 _3 are all different ignored parameters, because they can't all be called _

There is another bug to track fixing that particular mess

src/Microsoft.AspNet.Server.Kestrel/Http/Listener.cs
lodejard: The byte/char pool is used for the text writer. All signs point to that thing's strategy being changed for performance. I don't think the original IMemoryPool and MemoryPool will still be around after that.

src/Microsoft.AspNet.Server.Kestrel/Http/MemoryPool2.cs
```diff
@@ -0,0 +1,107 @@
...
+                // N/A: free unmanaged resources (unmanaged objects) and override a finalizer below.
+
+                // N/A: set large fields to null.
+
+                disposedValue = true;
+            }
+        }
+
+        // N/A: override a finalizer only if Dispose(bool disposing) above has code to free unmanaged resources.
+        // ~MemoryPool2() {
```
davidfowl: Is this important?

src/Microsoft.AspNet.Server.Kestrel/Http/MemoryPool2.cs
```diff
@@ -0,0 +1,107 @@
...
+            }
+        }
+
+        public void Return(MemoryPoolBlock2 block)
+        {
+            block.Reset();
+            _blocks.Push(block);
+        }
+
+        #region IDisposable Support
```
davidfowl: We don't region bro.

src/Microsoft.AspNet.Server.Kestrel/Http/MemoryPoolBlock2.cs
```diff
@@ -0,0 +1,551 @@
...
+            }
+
+            public int Seek(int char0, int char1)
+            {
+                if (IsDefault)
+                {
+                    return -1;
+                }
+
+                var byte0 = (byte)char0;
```
davidfowl: I think an explanation is in order for how this is using SIMD to seek through bytes.

src/Microsoft.AspNet.Server.Kestrel/Http/MemoryPoolBlock2.cs
```diff
@@ -0,0 +1,551 @@
...
+                var length = 0;
+                while (true)
+                {
+                    if (block == end._block)
+                    {
+                        return length + end._index - index;
+                    }
+                    else if (block.Next == null)
+                    {
+                        throw new Exception("end did not follow iterator");
```
davidfowl: Never throw Exception. Use InvalidOperationException. When happens when this throws? Does the connection die?

src/Microsoft.AspNet.Server.Kestrel/Http/MemoryPoolBlock2.cs
```diff
@@ -0,0 +1,551 @@
...
+using System.Collections.Concurrent;
+using System.Diagnostics;
+using System.Linq;
+using System.Numerics;
+using System.Runtime.InteropServices;
+using System.Text;
+
+namespace Microsoft.AspNet.Server.Kestrel.Http
+{
+    public class MemoryPoolBlock2
```
davidfowl: This class needs some **UBER** unit testing. The enumerating over blocks should be easy to verify. All the edge cases :smile:

src/Microsoft.AspNet.Server.Kestrel/Http/MessageBody.cs
```diff
@@ -3,29 +3,41 @@
...
 using System;
 using System.Collections.Generic;
+using System.Threading;
+using System.Threading.Tasks;
 using Microsoft.Framework.Primitives;
 
 namespace Microsoft.AspNet.Server.Kestrel.Http
 {
-    public abstract class MessageBody : MessageBodyExchanger
+    public abstract class MessageBody
```
davidfowl: This class shouldn't have passed the standards police. This needs to be broken up into different files.

src/Microsoft.AspNet.Server.Kestrel/Http/MessageBody.cs
```diff
@@ -111,29 +132,43 @@ public override void Consume()
...
+                    var end = begin.CopyTo(buffer.Array, buffer.Offset, limit, out actual);
+                    _inputLength -= actual;
+                    input.ConsumingComplete(end, end);
+                    if (actual != 0)
+                    {
+                        return actual;
+                    }
+                    if (input.RemoteIntakeFin)
+                    {
+                        throw new Exception("Unexpected end of request content");
```
davidfowl: InvalidOperationException

src/Microsoft.AspNet.Server.Kestrel/Http/MemoryPoolBlock2.cs
Tragetaschen: I like that something called 'char' is of type 'int' and cast to 'byte' :)

src/Microsoft.AspNet.Server.Kestrel/Http/MemoryPoolBlock2.cs
```diff
@@ -0,0 +1,551 @@
...
+                                _block = block;
+                                _index = index;
+                                return char0;
+                            }
+                        }
+                    }
+                }
+            }
+
+            public int Seek(int char0, int char1)
```
halter73: :+1: for comments. In this method for example. Are we seeking to char0 and/or char1?

src/Microsoft.AspNet.Server.Kestrel/Http/Frame.cs
```diff
@@ -62,95 +64,125 @@ public bool HasResponseStarted
...
+        }
+
+        public void ResetResponseHeaders()
+        {
+            _responseHeaders.Reset();
+            _responseHeaders.HeaderServer = "Kestrel";
+            _responseHeaders.HeaderDate = DateTime.UtcNow.ToString("r");
+        }
+
+        public async void ProcessFraming()
```
halter73: async void!? Without any casting, Task.Run expects a `Func<Task>` here so this fails to build:

    Error: Connection.cs(43,22): DNX,Version=v4.5.1 error CS0407: 'void Frame.ProcessFraming()' has the wrong return type

src/Microsoft.AspNet.Server.Kestrel/Http/MemoryPoolBlock2.cs
lodejard: Yes - now everything that happens inside the frame's asyc processing function (which includes framing, application invocation, and all calls back into stream read/write) will ultimately land at the by the processing function's catch/finally (if they're not caught by the app or error handling middleware).

The processing function's ultimate catch/finally will disconnect the socket, and everything will teardown.

src/Microsoft.AspNet.Server.Kestrel/Http/MessageBody.cs
lodejard: Can do, but nested classes are allowed, and they do come after all other MessageBody members

src/Microsoft.AspNet.Server.Kestrel/Http/MessageBody.cs
lodejard: I changed this to InvalidDataException. This one will throw back from Request.Body.Read/ReadAsync. Then it is caught where it's caught, worst-case disconnecting the connection when the frame's async processing function does it's finally {} clause

src/Microsoft.AspNet.Server.Kestrel/Http/MemoryPoolBlock2.cs
lodejard: Added comments to all the other mem pool types... I'll move the iterator to it's own file and add comments. Probably rename it to Cursor instead of iterator as it goes...

But yes, this is seeking to char0 or char1, whichever comes first, and returns char0, char1, or -1 based on what it found (-1 if it found neither)

src/Microsoft.AspNet.Server.Kestrel/Http/Frame.cs
lodejard: That's inconvenient... I fixed it with `Task.Run((Action)_frame.ProcessFraming);` Dunno why it couldn't take an action automatically, but it could take the func.

src/Microsoft.AspNet.Server.Kestrel/Http/MemoryPool2.cs
lodejard: It's the boiler-plate commented code left by the "implement IDisposable" thing. I left it intact, in case it turned out that this class did need to have something in that code-path eventually. Ya know... Trusting that uncommenting the correct code would be better than typing it from memory, if that turned out to be the case.

src/Microsoft.AspNet.Server.Kestrel/Http/MemoryPool2.cs
lodejard: Same reason as above, really - this was added by the "implement IDisposable the only way that works" and I thought it might be better to leave the one-true-and-correct boilerplate intact. We can de-boilerplate it if you think that's best

src/Microsoft.AspNet.Server.Kestrel/Http/MemoryPoolBlock2.cs
lodejard: yeah :) they actually were byte arguments at one point, but then the caller needed to have `Seek((byte)'\r)` which was nasty... Turns out that '\r' will implicitly become an `int`, so it can `Seek('\r')` now.

The return value being and int is also valuable to be able to return -1, which is outside the set of valid char/byte values. A minor concern, but seemed nicer all things considered...

src/Microsoft.AspNet.Server.Kestrel/Http/SocketOutput.cs
```diff
@@ -222,13 +268,89 @@ private class CallbackContext
...
+                {
+                    _writeReq.Dispose();
+                    var _this = (WriteContext)state;
+                    _this.WriteStatus = status;
+                    _this.WriteError = error;
+                    StageTwo();
+                }, this);
+            }
+
+            public void StageTwo()
```
davidfowl: Rename

src/Microsoft.AspNet.Server.Kestrel/Http/SocketOutput.cs
```diff
@@ -222,13 +268,89 @@ private class CallbackContext
...
+                    var _this = (WriteContext)state;
+                    _this.ShutdownSendStatus = status;
+
+                    Self._log.ConnectionWroteFin(Self._connectionId, status);
+
+                    StageThree();
+                }, this);
+            }
+
+            public void StageThree()
```
davidfowl: Rename

src/Microsoft.AspNet.Server.Kestrel/Infrastructure/MemoryPool2.cs
```diff
@@ -0,0 +1,169 @@
...
+        /// the caller may write to the block's entire size rather than being limited to the minumumSize requested.</param>
+        /// <returns>The block that is reserved for the called. It must be passed to Return when it is no longer being used.</returns>
+        public MemoryPoolBlock2 Lease(int minimumSize)
+        {
+            if (minimumSize > blockLength)
+            {
+                // The requested minimumSize is actually larger then the usable memory of a single block.
+                // Because this is the degenerate case, a one-time-use byte[] array and tracking object are allocated.
+                // When this block tracking object is returned it is not added to the pool - instead it will be 
+                // allowed to be garbace collected normally.
```
davidfowl: grabace -> garbage

src/Microsoft.AspNet.Server.Kestrel/Infrastructure/MemoryPool2.cs
```diff
@@ -0,0 +1,169 @@
...
+        /// leaving "dead zones" in the slab due to lost block tracking objects.
+        /// </summary>
+        /// <param name="block">The block to return. It must have been acquired by calling Lease on the same memory pool instance.</param>
+        public void Return(MemoryPoolBlock2 block)
+        {
+            block.Reset();
+            _blocks.Push(block);
+        }
+
+        #region IDisposable Support
```
davidfowl: remove regions

src/Microsoft.AspNet.Server.Kestrel/Http/Frame.cs
benaadams: Shudder...

Return a cached completed Task

```static Task _completedTask = Task.FromResult(0);```

async void makes the gods angry

Comment (benaadams) [inline src/Microsoft.AspNet.Server.Kestrel/Http/Frame.cs]: Actually its a common||gardern async function? Just change ```void``` to ```Task```

File: src/Microsoft.AspNet.Server.Kestrel/Http/Connection.cs [hunk 24]
```diff
@@ -29,15 +30,17 @@ public Connection(ListenerContext context, UvStreamHandle socket) : base(context
...
         }
 
         public void Start()
         {
             Log.ConnectionStart(_connectionId);
 
-            SocketInput = new SocketInput(Memory);
-            SocketOutput = new SocketOutput(Thread, _socket, _connectionId, Log);
-            _frame = new Frame(this);
+            Task.Run((Action)_frame.ProcessFraming);
```
benaadams: Remove cast and change ProcessFraming to Task returning; don't need to change function body

test/Microsoft.AspNet.Server.KestrelTests/EngineTests.cs
```diff
@@ -12,6 +12,7 @@
 using Microsoft.Dnx.Runtime;
 using Microsoft.Dnx.Runtime.Infrastructure;
 using Xunit;
+using System.Linq;
```
davidfowl: sort usings

test/Microsoft.AspNet.Server.KestrelTests/EngineTests.cs
```diff
@@ -23,7 +24,7 @@ public class EngineTests
         private async Task App(Frame frame)
```
davidfowl: Remove Console.WriteLines from the test

src/Microsoft.AspNet.Server.Kestrel/Http/Connection.cs
benaadams: Just checking something for this

benaadams: Does it matter that ProcessFraming could complete before OR after ReadStart?

benaadams: Ok, measured a few permutations and stackwalked them; Task.Run does a bunch of stuff but ends up being ```ThreadPool.QueueUserWorkItem``` with extras; so cut to the chase here as you don't care about the extras.

Change ```ProcessFraming``` from ```async void``` to ```async Task``` as though it probably won't do any harm in this scenario, ```async void``` like the ```goto``` of the async world and people will make fun of you.

Then just jump straightish to ```QueueUserWorkItem``` as follows:

```csharp
public void Start()
{
    Log.ConnectionStart(_connectionId);

    ThreadPool.QueueUserWorkItem(Process, _frame.ProcessFraming);
    _socket.ReadStart(_allocCallback, _readCallback, this);
}

private void Process(object state)
{
    ((Func<Task>)state)();
}
```

Still question on ordering of ```_socket.ReadStart``` though?

src/Microsoft.AspNet.Server.Kestrel/Http/SocketOutput.cs
lodejard: Done

src/Microsoft.AspNet.Server.Kestrel/Http/SocketOutput.cs
lodejard: Done

src/Microsoft.AspNet.Server.Kestrel/Infrastructure/MemoryPool2.cs
lodejard: grabace -> garbace :)

src/Microsoft.AspNet.Server.Kestrel/Infrastructure/MemoryPool2.cs
lodejard: done

test/Microsoft.AspNet.Server.KestrelTests/EngineTests.cs
lodejard: done

src/Microsoft.AspNet.Server.Kestrel/Http/Connection.cs
lodejard: The timing of ReadStart doesn't matter in this case. It initiates the first read with libuv, but because this function is running on the uv loop it won't fire any callbacks until it returns.

Even if it did, it would still be safe because the read callback and the frame consumption are mediated by the SocketInput class.

benaadams: K, makes sense; then I'd just change ```Task.Run``` -> ```ThreadPool.QueueUserWorkItem``` as per snippet you are ignoring the Task results and catching all the Exceptions in ```ProcessFraming``` and ```Task.Run``` allocates more to allow for StackTrace unwinding in whatever awaits it; and eventually just calls ```ThreadPool.QueueUserWorkItem```

lodejard: Actually - was making that change and realized that the Task being returned will be important for graceful shutdown and finishing requests in progress. I reworked the code to support that case, and moved the details into From.Start/.Stop

benaadams: Path is:
[Task.Run](https://github.com/dotnet/coreclr/blob/4cf8a6b082d9bb1789facd996d8265d3908757b2/src/mscorlib/src/System/Threading/Tasks/Task.cs#L5659)
* ```StackCrawlMark stackMark = StackCrawlMark.LookForMyCaller```
* [TaskFactory.StartNew](https://github.com/dotnet/coreclr/blob/4cf8a6b082d9bb1789facd996d8265d3908757b2/src/mscorlib/src/System/Threading/Tasks/TaskFactory.cs#L569)
* InternalStartNew
 * ```new Task```
 * ```Task.PossiblyCaptureContext```
 * ```Task.ScheduleAndStart```
  * ...
  * [```TaskScheduler.InternalQueueTask```](https://github.com/dotnet/coreclr/blob/4cf8a6b082d9bb1789facd996d8265d3908757b2/src/mscorlib/src/System/Threading/Tasks/TaskScheduler.cs#L265)
    * [```TaskScheduler.QueueTask```](https://github.com/dotnet/coreclr/blob/4cf8a6b082d9bb1789facd996d8265d3908757b2/src/mscorlib/src/System/Threading/Tasks/TaskScheduler.cs#L632)
    * [SynchronizationContext.Post](https://github.com/dotnet/coreclr/blob/4cf8a6b082d9bb1789facd996d8265d3908757b2/src/mscorlib/src/System/Threading/SynchronizationContext.cs#L126)
      * ThreadPool.QueueUserWorkItem

src/Microsoft.AspNet.Server.Kestrel/Http/SocketOutput.cs
```diff
@@ -222,13 +268,111 @@ private class CallbackContext
...
+                    buffers[i++] = buffer;
+                }
+
+                var writeReq = new UvWriteReq(Self._log);
+                writeReq.Init(Self._thread.Loop);
+                writeReq.Write(Self._socket, new ArraySegment<ArraySegment<byte>>(buffers), (_writeReq, status, error, state) =>
+                {
+                    if (error != null)
+                    {
+                        var x = 5;
```
halter73: :laughing:

benaadams:
It will be important to verify this (beyond regular unit tests) on different hardware x64 with SSE, AVX and AVX2

e.g. where 
* Vector<byte>.Count == 16
* Vector<byte>.Count == 32
* Vector<byte>.Count == 64

As that's the amount of bytes the Vector<byte> will process at once; and to max sure you are ignoring data passed the end of the input and not reading passed the end of the allocated memory chunk. I don't think it will go passed 64 bytes for some time as that's the processor cache line width.

(Probably only need to verify with 64 and the earlier ones will be covered)

benaadams:
You are probably disregarding beyond read buffer already; so might just be necessary to ensure allocated buffer is large enough; which you may already be doing for cache line safety; so the AVX2 case may already be covered.

samples/SampleApp/Startup.cs
```diff
@@ -26,7 +30,7 @@ public void Configure(IApplicationBuilder app, ILoggerFactory loggerFactory)
 
                 context.Response.ContentLength = 11;
                 context.Response.ContentType = "text/plain";
-                return context.Response.WriteAsync("Hello world");
```
halter73: Any particular reason behind this change?

samples/SampleApp/project.json
```diff
@@ -13,6 +13,7 @@
         }
     },
     "commands": {
+        "web": "Microsoft.AspNet.Hosting --server Microsoft.AspNet.Server.Kestrel",
```
halter73: Do we need web, run and kestrel commands that all do the same thing?

src/Microsoft.AspNet.Server.Kestrel/Http/Frame.cs
```diff
@@ -16,29 +18,34 @@ namespace Microsoft.AspNet.Server.Kestrel.Http
...
         private static readonly ArraySegment<byte> _endChunkedResponseBytes = CreateAsciiByteArraySegment("0\r\n\r\n");
         private static readonly ArraySegment<byte> _continueBytes = CreateAsciiByteArraySegment("HTTP/1.1 100 Continue\r\n\r\n");
+        private static readonly ArraySegment<byte> _emptyData = new ArraySegment<byte>(new byte[0]);
+        private static readonly byte[] _hex = Encoding.ASCII.GetBytes("0123456789abcdef");
 
-        private Mode _mode;
-        private bool _responseStarted;
-        private bool _keepAlive;
-        private bool _autoChunk;
+        private readonly object _onStartingSync = new Object();
```
halter73: nit: make "Object" all lower case

tools/Microsoft.AspNet.Server.Kestrel.GeneratedCode/KnownHeaders.cs
```diff
@@ -190,6 +178,28 @@ public partial class {loop.ClassName}
         {Each(loop.Headers, header => @"
         private StringValues _" + header.Identifier + ";")}
 
+        {Each(loop.Headers, header => $@"
+        public StringValues Header{header.Identifier}
```
halter73: Why are these all prefixed with "Header"?

src/Microsoft.AspNet.Server.Kestrel/Http/Frame.cs
```diff
@@ -62,95 +69,158 @@ public bool HasResponseStarted
...
-                        Execute();
-                        break;
-
-                    case Mode.MessageBody:
-                        if (MessageBody.LocalIntakeFin)
+                        finally
                         {
-                            // NOTE: stop reading and resume on keepalive?
-                            return;
+                            FireOnCompleted();
```
halter73: So if the user's oncompleted callback throws, the connection dies?

halter73: Maybe we can fix this up when we make OnStarted/OnCompleted callback invocation async.

src/Microsoft.AspNet.Server.Kestrel/Http/Frame.cs
```diff
@@ -218,33 +288,37 @@ private void FireOnCompleted()
...
+        {
+            ProduceStart(immediate: false);
 
-                // Trigger FireOnStarting if ProduceStart hasn't been called yet.
-                // We call it here, so it can go through our normal error handling
-                // and respond with a 500 if an OnStarting callback throws.
-                if (!_responseStarted)
+            if (_autoChunk)
+            {
+                if (data.Count == 0)
```
halter73: I know this is the way it was before, but would it make sense to exit early even if we aren't chunking for zero length writes?

src/Microsoft.AspNet.Server.Kestrel/Http/Frame.cs
```diff
@@ -355,18 +419,8 @@ public void ProduceStart(bool immediate = true, bool appCompleted = false)
...
-                {
-                    if (error != null)
-                    {
-                        Log.LogError("ProduceStart ", error);
-                    }
-                    ((IDisposable)state).Dispose();
-                },
-                responseHeader.Item2,
-                immediate: immediate);
+            SocketOutput.Write(responseHeader.Item1, immediate: immediate);
```
halter73: I don't think ProduceStart should ever block waiting for writes to complete.  I know that most of the time `immediate = false`, but that's not always true.

I think we're going to have to make this async eventually anyway when FireOnStarting becomes async.

tools/Microsoft.AspNet.Server.Kestrel.GeneratedCode/KnownHeaders.cs
```diff
@@ -190,6 +178,28 @@ public partial class {loop.ClassName}
...
+        {{
+            get
+            {{
+                if ({header.TestBit()})
```
halter73: Since `ClearFast` now actually sets all the properties to `StringValues.Empty`, I think we can remove the if/else and just `return _{header.Identifier};` in the getter.

test/Microsoft.AspNet.Server.KestrelTests/TestLogger.cs
```diff
@@ -4,20 +4,44 @@
...
+        public TestKestrelTrace() : base(new TestLogger())
         {
+
         }
 
-        public bool IsEnabled(LogLevel logLevel)
+        public override void ConnectionRead(long connectionId, int count)
         {
-            return false;
+            //_logger.LogDebug(1, @"Connection id ""{ConnectionId}"" recv {count} bytes.", connectionId, count);
```
halter73: @davidfowl  Why no comment yet?

tools/Microsoft.StandardsPolice/StandardsPoliceCompileModule.cs
```diff
@@ -71,8 +71,8 @@ internal static void ScanSyntaxTree(IList<Diagnostic> diagnostics, SyntaxTree sy
                     acceptableOrder = true;
                 }
                 if (!acceptableOrder &&
-                    priorUsingDirective.Name.ToString().StartsWith("System.") &&
-                    !usingDirective.Name.ToString().StartsWith("System."))
+                    priorUsingDirective.Name.ToString().StartsWith("System") &&
```
halter73: This avoids some false negatives.

    (priorUsingDirective.Name.ToString() == "System" ||
    priorUsingDirective.Name.ToString().StartsWith("System.")) &&
    !usingDirective.Name.ToString().StartsWith("System."))

which flags:

    using System2;
    using MyNamespace;

src/Microsoft.AspNet.Server.Kestrel/Infrastructure/MemoryPool2.cs
```diff
@@ -0,0 +1,166 @@
...
+                // N/A: free unmanaged resources (unmanaged objects) and override a finalizer below.
+
+                // N/A: set large fields to null.
+
+                disposedValue = true;
+            }
+        }
+
+        // N/A: override a finalizer only if Dispose(bool disposing) above has code to free unmanaged resources.
+        // ~MemoryPool2() {
```
halter73: Remove commented code.

src/Microsoft.AspNet.Server.Kestrel/Infrastructure/MemoryPool2.cs
```diff
@@ -0,0 +1,166 @@
+﻿using System;
+using System.Collections.Concurrent;
+
+namespace Microsoft.AspNet.Server.Kestrel.Infrastructure
+{
+    /// <summary>
+    /// Used to allocate and distribute re-usable blocks of memory.
+    /// </summary>
+    public class MemoryPool2 : IDisposable
```
halter73: Will we entirely replace the existing memory pool with this?

src/Microsoft.AspNet.Server.Kestrel/Infrastructure/Disposable.cs
```diff
@@ -41,6 +39,5 @@ public void Dispose()
             Dispose(true);
             GC.SuppressFinalize(this);
```
halter73: Doesn't this patter only matter if this class (or one derived from it) has a finalizer?

src/Microsoft.AspNet.Server.Kestrel/Http/SocketOutput.cs
```diff
@@ -222,13 +325,107 @@ private class CallbackContext
...
+            {
+                DoWriteIfNeeded();
+            }
+
+            /// <summary>
+            /// First step: initiate async write if needed, otherwise go to next step
+            /// </summary>
+            public void DoWriteIfNeeded()
+            {
+                if (Buffers.Count == 0 || Self._socket.IsClosed)
```
halter73: :+1: Glad we're now checking if the socket is closed.

test/Microsoft.AspNet.Server.KestrelTests/MemoryPoolExtensions.cs
```diff
@@ -0,0 +1,13 @@
+﻿using Microsoft.AspNet.Server.Kestrel.Infrastructure;
+
+namespace Microsoft.AspNet.Server.KestrelTests
+{
+    public static class MemoryPoolExtensions
+    {
+        public static MemoryPoolIterator2 Add(this MemoryPoolIterator2 iterator, int count)
+        {
+            int actual;
+            return iterator.CopyTo(new byte[count], 0, count,  out actual);
```
halter73: nit: ctrl+k+d

src/Microsoft.AspNet.Server.Kestrel/Infrastructure/MemoryPoolIterator2.cs
```diff
@@ -0,0 +1,452 @@
...
+                return new ArraySegment<byte>(_block.Array, _index, end._index - _index);
+            }
+
+            var length = GetLength(end);
+            var array = new byte[length];
+            CopyTo(array, 0, length, out length);
+            return new ArraySegment<byte>(array, 0, length);
+        }
+
+        public MemoryPoolIterator2 CopyTo(byte[] array, int offset, int count, out int actual)
```
halter73: c# compiler is pretty smart to determine that `actual` is always set before returning. :smile:

src/Microsoft.AspNet.Server.Kestrel/Infrastructure/MemoryPoolBlock2.cs
```diff
@@ -0,0 +1,184 @@
...
+
+        /// <summary>
+        /// Reference to the next block of data when the overall "active" bytes spans multiple blocks. At the point when the block is
+        /// leased Next is guaranteed to be null. Start, End, and Next are used together in order to create a linked-list of discontiguous 
+        /// working memory. The "active" memory is grown when bytes are copied in, End is increased, and Next is assigned. The "active" 
+        /// memory is shrunk when bytes are consumed, Start is increased, and blocks are returned to the pool.
+        /// </summary>
+        public MemoryPoolBlock2 Next { get; set; }
+
+        ~MemoryPoolBlock2()
```
halter73: Should we use the dispose pattern for this?

I like that there is a finalizer, since consuming code theoretically might not return or dispose the block. Still, it would be nice if the pool tried disposing these when the pool itself gets disposed to prevent this from running on the finallizer thread.

src/Microsoft.AspNet.Server.Kestrel/Infrastructure/MemoryPoolBlock2.cs
```diff
@@ -0,0 +1,184 @@
...
+                Start = data.Offset,
+                End = data.Offset,
+            };
+        }
+
+        /// <summary>
+        /// called when the block is returned to the pool. mutable values are re-assigned to their guaranteed initialized state.
+        /// </summary>
+        public void Reset()
+        {
```
halter73: `Debug.Assert(!_pinHandle.IsAllocated);` might be nice.

src/Microsoft.AspNet.Server.Kestrel/Infrastructure/MemoryPoolIterator2.cs
```diff
@@ -0,0 +1,452 @@
...
+            {
+                return default(ArraySegment<byte>);
+            }
+            if (end._block == _block)
+            {
+                return new ArraySegment<byte>(_block.Array, _index, end._index - _index);
+            }
+
+            var length = GetLength(end);
+            var array = new byte[length];
```
halter73: Use the old memory pool for this?

test/Microsoft.AspNet.Server.KestrelTests/MemoryPoolExtensions.cs
halter73: Wait is this being used from anywhere other than the tests?

src/Microsoft.AspNet.Server.Kestrel/Infrastructure/MemoryPoolIterator2.cs
```diff
@@ -0,0 +1,452 @@
...
+            }
+            if (end._block == _block)
+            {
+                return _utf8.GetString(_block.Array, _index, end._index - _index);
+            }
+
+            var decoder = _utf8.GetDecoder();
+
+            var length = GetLength(end);
+            var charLength = length * 2;
```
halter73: Why double the number of bytes for the charLength?

src/Microsoft.AspNet.Server.Kestrel/Infrastructure/MemoryPoolIterator2.cs
```diff
@@ -0,0 +1,452 @@
...
+            var chars = new char[charLength];
```
halter73: Should we pool these char arrays like we did in the old memory pool?

src/Microsoft.AspNet.Server.Kestrel/Infrastructure/MemoryPool2.cs
```diff
@@ -0,0 +1,166 @@
...
+    {
+        /// <summary>
+        /// The gap between blocks' starting address. 4096 is chosen because most operating systems are 4k pages in size and alignment.
+        /// </summary>
+        private const int blockStride = 4096;
```
halter73: The other instances of `private const` in Kestrel are still prefixed with "_" (e.g. `MemoryPoolTextWriter._textLength`)

src/Microsoft.AspNet.Server.Kestrel/Infrastructure/MemoryPool2.cs
```diff
@@ -0,0 +1,166 @@
...
+        /// <summary>
+        /// Thread-safe collection of slabs which have been allocated by this pool. As long as a slab is in this collection and slab.IsActive, 
+        /// the blocks will be added to _blocks when returned.
+        /// </summary>
+        private readonly ConcurrentStack<MemoryPoolSlab2> _slabs = new ConcurrentStack<MemoryPoolSlab2>();
+
+        /// <summary>
+        /// This is part of implementing the IDisposable pattern.
+        /// </summary>
+        private bool disposedValue = false; // To detect redundant calls
```
halter73: _

src/Microsoft.AspNet.Server.Kestrel/Http/SocketInput.cs
```diff
@@ -35,80 +56,172 @@ public ArraySegment<byte> Take(int count)
...
+                        DataPtr = dataPtr,
+                    };
+                }
             }
+
+            _pinned = _memory.Lease(minimumSize);
+            return new IncomingBuffer
+            {
+                Data = _pinned.Data,
+                DataPtr = _pinned.Pin()
```
halter73: Isn't this memory already pinned in practice. Is there a penalty for pinning twice?

src/Microsoft.AspNet.Server.Kestrel/Http/SocketInput.cs
```diff
@@ -35,80 +56,172 @@ public ArraySegment<byte> Take(int count)
...
+                _awaitableIsNotCompleted);
 
-            Buffer = new ArraySegment<byte>(Buffer.Array, Buffer.Offset, Buffer.Count + count);
+            if (awaitableState == _awaitableIsNotCompleted)
+            {
+                return;
+            }
+            else if (awaitableState == _awaitableIsCompleted)
+            {
+                Task.Run(continuation);
```
halter73: Should we log any errors in the continuation in case the calling code doesn't properly handle it?

src/Microsoft.AspNet.Server.Kestrel/Http/SocketInput.cs
```diff
@@ -35,80 +56,172 @@ public ArraySegment<byte> Take(int count)
...
+            }
+            else
+            {
+                // THIS IS AN ERROR STATE - ONLY ONE WAITER CAN WAIT
```
halter73: Trace.Assert?

src/Microsoft.AspNet.Server.Kestrel/Infrastructure/MemoryPoolBlock2.cs
benaadams: Dispose and log "You have been bad" in finaliazer?

src/Microsoft.AspNet.Server.Kestrel/Infrastructure/MemoryPoolSlab2.cs
```diff
@@ -0,0 +1,96 @@
...
+            Dispose(false);
+        }
+
+        // This code added to correctly implement the disposable pattern.
+        public void Dispose()
+        {
+            // Do not change this code. Put cleanup code in Dispose(bool disposing) above.
+            Dispose(true);
+            // uncomment the following line if the finalizer is overridden above.
+            GC.SuppressFinalize(this);
```
benaadams: Log that the finalizer has run in some way?

davidfowl:
@lodejard get this in for beta8

benaadams:
Not sure Vectors support AVX2 ( Vector.Count == 64 ) currently; so just AVX ( Vector.Count == 32 ) and SSE ( Vector.Count == 16 )

This is quite an exciting PR!

lodejard:
@benaadams the Seek operation uses Vector.Count to increment, and falls back to byte iteration when less than that length, so it'll never Vector past valid data. We can look at vectoring past valid data if the fallback on the tail is slow, but that'll be a perf thing for the future

samples/SampleApp/project.json
lodejard: dunno, but not a change for this PR :)

tools/Microsoft.AspNet.Server.Kestrel.GeneratedCode/KnownHeaders.cs
lodejard: To avoid accidentally conflicting any other class members, because the header names are all over the map

src/Microsoft.AspNet.Server.Kestrel/Http/Frame.cs
lodejard: Probably

src/Microsoft.AspNet.Server.Kestrel/Http/Frame.cs
lodejard: The concern there would be that zero length writes would no longer act like a flush for the purposes of getting the response header out the door. It would be too subtle to say "any write, except those of zero length, will begin the response body".

src/Microsoft.AspNet.Server.Kestrel/Http/Frame.cs
lodejard: It'll only block in the case that the response data in write-behind is over 64k. That should be the less likely case, but in that case it's more important to stop allocating unboundedly.

tools/Microsoft.AspNet.Server.Kestrel.GeneratedCode/KnownHeaders.cs
lodejard: Done - needed set to empty in Remove also

test/Microsoft.AspNet.Server.KestrelTests/MemoryPoolExtensions.cs
lodejard: nope, but it's in the test project

src/Microsoft.AspNet.Server.Kestrel/Infrastructure/MemoryPoolBlock2.cs
lodejard: didn't want to implement dispose pattern because finalizer isn't actually disposing, and i didn't want the public api hint that blocks are IDisposable (which they technically aren't)... Also - was concerned that in the finalizer the world is in such an unstable shape - calling log on the finalizer thread from a dying object could end up calling into many object that are already on the road to re-collection

src/Microsoft.AspNet.Server.Kestrel/Infrastructure/MemoryPoolIterator2.cs
lodejard: old memory pool probably not going to survive for long... this is the degenerate case also - request headers are expected to very, very rarely span across multiple blocks

src/Microsoft.AspNet.Server.Kestrel/Infrastructure/MemoryPoolIterator2.cs
lodejard: because people don't realize utf8 is actually invalid in a header, and it's not worth doing two passes to figure out what the actual char length was

src/Microsoft.AspNet.Server.Kestrel/Infrastructure/MemoryPoolIterator2.cs
lodejard: yes, almost certainly!

src/Microsoft.AspNet.Server.Kestrel/Infrastructure/MemoryPool2.cs
lodejard: Done

src/Microsoft.AspNet.Server.Kestrel/Infrastructure/MemoryPool2.cs
lodejard: :ok:

src/Microsoft.AspNet.Server.Kestrel/Http/SocketInput.cs
lodejard: It's pinned only when the block wasn't provided as a slab segment. e.g. in the case you require more than ~4k linear bytes.

src/Microsoft.AspNet.Server.Kestrel/Http/SocketInput.cs
lodejard: This continuation should re-enter an async function, which should handle all of the exception cases without throwing back to the continuation caller

src/Microsoft.AspNet.Server.Kestrel/Http/SocketInput.cs
lodejard: I've actually been a lot more concerned about Trace.Assert's, given that they have had a tradition of turning unexpected-but-recoverable situations into upfront process failures. It multiplies the perception of instability.

src/Microsoft.AspNet.Server.Kestrel/Infrastructure/MemoryPoolSlab2.cs
lodejard: Same remark as earlier - logging in finalizer is likely not safe...

src/Microsoft.AspNet.Server.Kestrel/Http/Frame.cs
halter73: I don't think this is a big deal, but it's even more subtle to say "any write, except those of zero length without the Content-Length header aleready being set, will begin the response body".

src/Microsoft.AspNet.Server.Kestrel/Infrastructure/MemoryPoolBlock2.cs
halter73: So after in-person discussion, we decided we could suppress finalization for this in MemoryPool2. We also said we could put a Debug.Assert here so we can verify this doesn't actually run.

PR #210, opened 2015-09-18
Fast path default header collection type

benaadams:
Don't use interface's enumerable as it allocates

### DIFF
--- a/src/Microsoft.AspNet.Server.Kestrel/Http/Frame.cs
+++ b/src/Microsoft.AspNet.Server.Kestrel/Http/Frame.cs
@@ -354,7 +354,7 @@ namespace Microsoft.AspNet.Server.Kestrel.Http
 
             var status = ReasonPhrases.ToStatus(StatusCode, ReasonPhrase);
 
-            var responseHeader = CreateResponseHeader(status, appCompleted, ResponseHeaders);
+            var responseHeader = CreateResponseHeader(status, appCompleted);
             SocketOutput.Write(
                 responseHeader.Item1,
                 (error, state) =>
@@ -413,8 +413,7 @@ namespace Microsoft.AspNet.Server.Kestrel.Http
 
         private Tuple<ArraySegment<byte>, IDisposable> CreateResponseHeader(
             string status,
-            bool appCompleted,
-            IEnumerable<KeyValuePair<string, StringValues>> headers)
+            bool appCompleted)
         {
             var writer = new MemoryPoolTextWriter(Memory);
             writer.Write(HttpVersion);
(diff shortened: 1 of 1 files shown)


Tratcher:
@lodejard  I'd recommend kestrel always perform operations on the original FrameResponseHeaders. Headers are replaceable in the IHttpResponseFeature for the sake of wrapping, but I think servers should only ever try to operate on the original. Compare to the response Stream. That should eliminate the need for fast & slow paths like this.

lodejard:
Yep, it should always use the known-typed field instead of the idictionary. Properties have been added to the class for known headers also, so those should be used directly internally instead of try get.

benaadams:
You'll need to throw a NotImplementedExecption for RequestHeaders set then

benaadams:
Actually at that point you should probably change ```IHttpResponseFeature``` from:

```IDictionary<string, StringValues> Headers { get; set; }```

to:

```IDictionary<string, StringValues> Headers { get; }```

Else you can only determine if the server supports changing headers by catching an execption which wouldn't be great?

**Comment by Tratcher** [comment]
IHttpResponseFeature stores data during the execution of a request, it does not replace the server's underlying implementations. There are scenarios where you would wrap or replace the response headers dictionary with your own (e.g. adding filters, etc..) and store it back in the feature for others to use, but you're responsible for writing back to the original collection. The server is not expected to consume the wrapper.

**Comment by benaadams** [comment]
Ah sort of:

```Response.Headers = new Filter(Response.Headers)```

Makes sense, like you'd do with the response stream

**Comment by lodejard** [comment]
Yeah, exactly so

**Comment by benaadams** [comment]
Makes it simpler then :D

**Comment by benaadams** [comment]
> The server is not expected to consume the wrapper.

This breaks the tests

**Comment by Tratcher** [comment]
Which ones?

**Comment by benaadams** [comment]
ThrowingInOnStartingResultsIn500Response

> Assert.StartsWith() Failure:
Expected: Date:
Actual:   Conte...

If I change https://github.com/aspnet/KestrelHttpServer/blob/dev/src/Microsoft.AspNet.Server.Kestrel/Http/Frame.cs#L391 to clearing the headers rather than creating a new set and add a Date header; then Kestrel produces the wrong output.

Testing using coreclr

**Comment by benaadams** [comment]
The previous change was better; and I preferred, which followed the:

> The server is not expected to consume the wrapper.

So I wouldn't commit this; but something goes wrong in that circumstance

**Comment by benaadams** [comment]
> The previous change was better

e.g. its a one line change vs a multi line change :P

**Comment by benaadams** [comment]
> Which ones?

Specifically https://travis-ci.org/aspnet/KestrelHttpServer/builds/81145705#L1785

**Comment by benaadams** [comment]
Reverting to the good code change; will raise failure as issue if I can workout how to express it...

File: src/Microsoft.AspNet.Server.Kestrel/Http/Frame.cs [hunk 1]
```diff
@@ -426,40 +425,38 @@ public void ProduceEnd(Exception ex)
             var hasConnection = false;
             var hasTransferEncoding = false;
             var hasContentLength = false;
-            if (headers != null)
+
+            foreach (var header in _responseHeaders)
```
halter73: Using _responseHeaders here is an issue because we set the ResponseHeaders property in [the ProduceEnd method on line 391](https://github.com/aspnet/KestrelHttpServer/pull/210/files#diff-3ad18666c41931c02b7bd1f3226cb709R391).

I  think this PR can wait until @lodejard's [flow control PR](https://github.com/aspnet/KestrelHttpServer/pull/207) gets merged which [changes this](https://github.com/aspnet/KestrelHttpServer/pull/207/files#diff-3ad18666c41931c02b7bd1f3226cb709R445).

benaadams: I tried changing that and it went very strange. Had an alternative approach that did work but was ugly; so will wait till this works a bit better.

benaadams:
Updated for recent changes in PR #224

PR #215, opened 2015-09-20
Pawelka/libuv

moozzyk:
@DamianEdwards @halter73 @lodejard @davidfowl 
Early preview. Issues:
- nano does not work - ~need to talk to someone from the nano team to figure out where to pull missing APIs from (not a regression - what we currently have would not work on nano either)
- libuv does not contain @lodejard fix. As a result tests are failing. When I hacked libuv to include the fix all tests passed


moozzyk:
@davidfowl: why? I talked to a few people and no-one had concerns. Are we going to fork libuv?

davidfowl:
@moozzyk I though that's what we were doing. Last i recall we were going to make another repo and consume libuv as a package.

moozzyk:
No. Libuv.dll is part of Kestrel package and not a separate package. We don't want to be in the business of distributing libuv.dll as a autonomic NuGet package because people will take dependency on this and we will have to maintain it. @halter73 was to talk to @lodejard to submit a fix to libuv repo so we don't have any custom code. He said it is fine for now to not have this functionality. If we fork, modify the code and ship libuv as a separate NuGet package this is will be even worse from maintenance point of view.

davidfowl:
@moozzyk Part of the deal was about not making it any harder to build than it is today.

DamianEdwards:
I think we said we'd do a build-time package for libuv, but only publish it to our feeds. Kestrel build would extract the dll and package it up in itself.

moozzyk:
How a separate package that needs to be maintained separately (repo, CI, publishing) and then extracted in a different repo and copied is simpler than just a project that lives in the repo with the project that needs it?

davidfowl:
@moozzyk harder for us. Easier for any contributor that wants to just clone and run kestrel.

moozzyk:
OK. Someone create a repo and I will move the code. Do we want to copy the code or use a submodule at specific (release) tag. I personally would prefer to use a submodule.

moozzyk:
I moved these changes to the new repo: https://github.com/aspnet/libuv-build/tree/dev

PR #219, opened 2015-09-22
Add third party notices for libuv

### LINKED ISSUE #218: Include 3rd party notice file in NuGet package

Eilon:
Fixes #218 

@DamianEdwards @moozzyk

### DIFF
--- a/content/thirdpartynotices.txt
+++ b/content/thirdpartynotices.txt
@@ -1,0 +1,32 @@
+This file is based on or incorporates material from the projects listed below
+(Third Party IP). The original copyright notice and the license under which
+Microsoft received such Third Party IP, are set forth below. Such licenses and
+notices are provided for informational purposes only. Microsoft licenses the
+Third Party IP to you under the licensing terms for the Microsoft product.
+Microsoft reserves all other rights not expressly granted under this agreement,
+whether by implication, estoppel or otherwise. 
+
+libuv v. 1.7.3
+
+Copyright Joyent, Inc. and other Node contributors. All rights reserved.
+ Provided for Informational Purposes Only
+
+MIT License 
+
+Permission is hereby granted, free of charge, to any person obtaining a copy of
+this software and associated documentation files (the Software), to deal in the
+Software without restriction, including without limitation the rights to use,
+copy, modify, merge, publish, distribute, sublicense, and/or sell copies of the
+Software, and to permit persons to whom the Software is furnished to do so,
+subject to the following conditions: 
+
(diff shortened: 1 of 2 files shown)


DamianEdwards:
:shipit:

We'll need to add this to the new libuv repo too I'm guessing.

Eilon:
@DamianEdwards this is the only place where libuv will be redistributed, so it's the only place it should be needed. If you feel otherwise let's chat offline.

content/thirdpartynotices.txt
```diff
@@ -0,0 +1,32 @@
+This file is based on or incorporates material from the projects listed below
+(Third Party IP). The original copyright notice and the license under which
+Microsoft received such Third Party IP, are set forth below. Such licenses and
+notices are provided for informational purposes only. Microsoft licenses the
+Third Party IP to you under the licensing terms for the Microsoft product.
+Microsoft reserves all other rights not expressly granted under this agreement,
+whether by implication, estoppel or otherwise. 
+
+libuv v. 1.7.3
```
moozzyk: I took v. 1.7.4 - we will need to rev it once @lodejard change is merged to their repo

Eilon: Can we ship beta8 with 1.7.4 as-is? Or do we need 1.7.4+ @lodejard 's change?

muratg: Obviously we won't be able to send the PR and get the changes in in time for beta 8, so if we need the change, we'll need to hack it in.  Unless mainline scenarios are broken without the fix, we should probably go with the 1.7.4 as-is.   cc @DamianEdwards

moozzyk: @halter73 mentioned that should be fine to ship beta8 without the fix.

Eilon: Can we use 1.7.3 then? Or do we need 1.7.4?

moozzyk: @Eilon I don't think it makes a difference for us at this point. The @lodejard's fix is in neither version so either you change the license or I change the submodule to point to 1.7.3 instead of 1.7.4

Eilon: It's easier if we just use 1.7.3 for right now. We're talking to the libuv folks about the change right now.

moozzyk: OK

Eilon:
https://github.com/aspnet/KestrelHttpServer/commit/c4bf10d6e9d77f75fd2a413c87a6186ac52d3c69

Issue #163, opened 2015-08-14
Date header is slow


DamianEdwards:
We should generate the Date response header value on a fixed cadence and just use that. Example in https://github.com/aspnet/KestrelHttpServer/tree/damianedwards/date-header-perf

I'll change this to create a new type that manages the generation of the value including the creation and disposal of the timer. That'll be created by `KestrelEngine` and flowed through to the other bits, all the way to `FrameHeaders` so that `FrameHeaders.Generated.cs` can use it grab the date value (here https://github.com/aspnet/KestrelHttpServer/blob/dev/src/Microsoft.AspNet.Server.Kestrel/Http/FrameHeaders.Generated.cs#L12)

PR #220, opened 2015-09-24
Date header performance

DamianEdwards:
#163

Uses a Timer to update a shared date string that is sent for the "Date" response header.
The Timer doesn't start until a request comes in, then shuts off after an idle period of no requests, and finally starts again if requests come in again.

NOTE: This is now rebased on #207 so until that gets merged, this PR contains all those changes too.

### DIFF
--- a/src/Microsoft.AspNet.Server.Kestrel/Http/DateHeaderValueManager.cs
+++ b/src/Microsoft.AspNet.Server.Kestrel/Http/DateHeaderValueManager.cs
@@ -1,0 +1,145 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Threading;
+using Microsoft.AspNet.Server.Kestrel.Infrastructure;
+
+namespace Microsoft.AspNet.Server.Kestrel.Http
+{
+    /// <summary>
+    /// Manages the generation of the date header value.
+    /// </summary>
+    public class DateHeaderValueManager : IDisposable
+    {
+        private readonly ISystemClock _systemClock;
+        private readonly TimeSpan _timeWithoutRequestsUntilIdle;
+        private readonly TimeSpan _timerInterval;
+
+        private volatile string _dateValue;
+        private object _timerLocker = new object();
+        private bool _isDisposed = false;
+        private bool _hadRequestsSinceLastTimerTick = false;
+        private Timer _dateValueTimer;
+        private DateTimeOffset _lastRequestSeen = DateTimeOffset.MinValue;
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="DateHeaderValueManager"/> class.
+        /// </summary>
+        public DateHeaderValueManager()
+            : this(
+                  systemClock: new SystemClock(),
+                  timeWithoutRequestsUntilIdle: TimeSpan.FromSeconds(10),
+                  timerInterval: TimeSpan.FromSeconds(1))
+        {
+
+        }
+
+        // Internal for testing
+        internal DateHeaderValueManager(
+            ISystemClock systemClock,
+            TimeSpan timeWithoutRequestsUntilIdle,
+            TimeSpan timerInterval)
+        {
+            _systemClock = systemClock;
+            _timeWithoutRequestsUntilIdle = timeWithoutRequestsUntilIdle;
+            _timerInterval = timerInterval;
+        }
+
+        /// <summary>
+        /// Returns a value representing the current server date/time for use in the HTTP "Date" response header
+        /// in accordance with http://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.18
+        /// </summary>
+        /// <returns>The value.</returns>
+        public string GetDateHeaderValue()
+        {
+            PumpTimer();
+
+            // See https://msdn.microsoft.com/en-us/library/az4se3k1(v=vs.110).aspx#RFC1123 for info on the format
+            // string used here.
+            // The null-coalesce here is to protect against returning null after Dispose() is called, at which
+            // point _dateValue will be null forever after.
+            return _dateValue ?? _systemClock.UtcNow.ToString("r");
+        }
+
+        /// <summary>
+        /// Releases all resources used by the current instance of <see cref="DateHeaderValueManager"/>.
+        /// </summary>
+        public void Dispose()
+        {
+            lock (_timerLocker)
+            {
+                if (_dateValueTimer != null)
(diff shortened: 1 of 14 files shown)


src/Microsoft.AspNet.Server.Kestrel/Http/DateHeaderValueManager.cs
```diff
@@ -0,0 +1,149 @@
...
+{
+    /// <summary>
+    /// Manages the generation of the date header value.
+    /// </summary>
+    public class DateHeaderValueManager : IDisposable
+    {
+        private readonly ISystemClock _systemClock;
+        private readonly TimeSpan _timeWithoutRequestsUntilIdle;
+        private readonly TimeSpan _timerInterval;
+        private readonly uint _timerTicksWithoutRequestsUntilIdle;
```
muratg: `double`? and _timerTicksSinceLastRequest can be `double` as well to avoid casting issues & equality check between `int` and `uint`.  And equality check can become comparison to avoid double precision issues.

Tragetaschen: `TimeSpan`? And increment by `_timerInterval` in the timer. (plus `>=` instead of `==`)

muratg: Even better :)

DamianEdwards: OK, but I did it a different way to what you suggested :smile:

DamianEdwards:
Updated

Tratcher:
https://github.com/aspnet/KestrelHttpServer/issues/223

DamianEdwards:
Rebased on #207

AspNetSmurfLab:
With this change we see throughput go from ~244,000 RPS (with the changes in #207) to ~273,000 RPS. That's a nearly 12% increase in throughput.

DamianEdwards:
ping @halter73 @lodejard @davidfowl

src/Microsoft.AspNet.Server.Kestrel/Http/DateHeaderValueManager.cs
```diff
@@ -0,0 +1,145 @@
...
+            }
+        }
+
+        // Called by the Timer (background) thread
+        private void UpdateDateValue(object state)
+        {
+            var now = _systemClock.UtcNow;
+
+            // See http://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.18 for required format of Date header
+            _dateValue = now.ToString("r");
```
lodejard: Move "r" to Constants class?

src/Microsoft.AspNet.Server.Kestrel/Http/DateHeaderValueManager.cs
```diff
@@ -0,0 +1,145 @@
...
+        }
+
+        /// <summary>
+        /// Releases all resources used by the current instance of <see cref="DateHeaderValueManager"/>.
+        /// </summary>
+        public void Dispose()
+        {
+            lock (_timerLocker)
+            {
+                if (_dateValueTimer != null)
```
lodejard: Move if() into DisposeTimer where it's being altered

src/Microsoft.AspNet.Server.Kestrel/ServerFactory.cs
```diff
@@ -53,9 +54,16 @@ public IDisposable Start(IFeatureCollection serverFeatures, Func<IFeatureCollect
             try
             {
                 var information = (KestrelServerInformation)serverFeatures.Get<IKestrelServerInformation>();
-                var engine = new KestrelEngine(_libraryManager, new ServiceContext { AppShutdown = _appShutdownService, Log = new KestrelTrace(_logger) });
+                var dateHeaderValueManager = new DateHeaderValueManager();
```
lodejard: Provide default "new" in ServiceContect ctor instead of this code path?

src/Microsoft.AspNet.Server.Kestrel/ServerFactory.cs
```diff
@@ -53,9 +54,16 @@ public IDisposable Start(IFeatureCollection serverFeatures, Func<IFeatureCollect
...
+                var engine = new KestrelEngine(_libraryManager, new ServiceContext
+                {
+                    AppShutdown = _appShutdownService,
+                    Log = new KestrelTrace(_logger),
+                    DateHeaderValueManager = dateHeaderValueManager
+                });
 
                 disposables.Push(engine);
+                disposables.Push(dateHeaderValueManager);
```
lodejard: oh, my

src/Microsoft.AspNet.Server.Kestrel/ServerFactory.cs
DamianEdwards: This is a little messy, as whoever creates the `DateHeaderManager` should be disposing it, and `ServiceContext` isn't disposable, and everything else derives from it, and you don't want them to destroy the default instance of `DateHeaderManager` in their base. We might have to fix this more generically later for things like `MemoryPool` as well that hang off of `ServiceContext`.

src/Microsoft.AspNet.Server.Kestrel/ServerFactory.cs
DamianEdwards: Yeah it's messy, but this ensures that "he who creates it disposes it". We should look at this all up later.

src/Microsoft.AspNet.Server.Kestrel/Http/DateHeaderValueManager.cs
DamianEdwards: Do you really want me to do that? :smile:

rynowak: If you're not going to do that, at least move `http://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.18` to a resource string :trollface:

DamianEdwards: OK doing that, it's actually kinda nice.

src/Microsoft.AspNet.Server.Kestrel/Http/DateHeaderValueManager.cs
DamianEdwards: :+1:

DamianEdwards:
Updated again

halter73:
:shipit: assuming @lodejard is making the lifetime/dispose changes.

PR #224, opened 2015-09-25
Fast path header enumerator

benaadams:
Use strongly typed enumerable as interface's allocates.

Is updated #210 but forked from current.


Tratcher:
:+1:

davidfowl:
@halter73 @muratg beta8?

Issue #98, opened 2015-05-08
Microsoft.AspNet.Server.Kestrel.Http.Listener ignores host argument


srdjantot:
Listener always binds to IPAddress.Any host. This can lead to problems when security policy allow only specific IP to bind to.

Are there any reasons for listener to ignore supplied host argument?

```csharp
public Task StartAsync(
            string scheme,
            string host,
            int port,
            KestrelThread thread,
            Func<Frame, Task> application)
        {
            Thread = thread;
            Application = application;

            var tcs = new TaskCompletionSource<int>();
            Thread.Post(_ =>
            {
                try
                {
                    ListenSocket = new UvTcpHandle();
                    ListenSocket.Init(Thread.Loop, Thread.QueueCloseHandle);
                    ListenSocket.Bind(new IPEndPoint(IPAddress.Any, port));
                    ListenSocket.Listen(10, _connectionCallback, this);
                    tcs.SetResult(0);
                }
                catch (Exception ex)
                {
                    tcs.SetException(ex);
                }
            }, null);
            return tcs.Task;
        }
```

davidfowl:
Looks like a bug to me

Tratcher:
Allow binding to IPs or *.
If localhost is requested, bind to 127.0.0.1 and ::1.
If any other host is requested bind to *.
For *, bind to both IPv4 and IPv6.

halter73:
I think this is a dupe of https://github.com/aspnet/KestrelHttpServer/issues/57. Should we close one of these issues?

Tratcher:
Taken care of.

PR #225, opened 2015-09-25
Bind to specific IP addresses if provided with any

halter73:
This still only applies to IPv4.

#98

@Tratcher @DamianEdwards @davidfowl


Tratcher:
:shipit:

PR #236, opened 2015-10-02
Sanitize loops

benaadams:
Use jit recognised ```for (var i; i < a.length; i++)``` for bounds check elimination.
Make some loop constructs more clear with intent e.g. ```while(true)``` vs ```for( ; ; )```

### DIFF
--- a/tools/Microsoft.AspNet.Server.Kestrel.GeneratedCode/KnownHeaders.cs
+++ b/tools/Microsoft.AspNet.Server.Kestrel.GeneratedCode/KnownHeaders.cs
@@ -59,7 +59,7 @@ namespace Microsoft.AspNet.Server.Kestrel.GeneratedCode
             {
                 ulong mask = 0;
                 ulong comp = 0;
-                for (var scan = 0; scan != count; ++scan)
+                for (var scan = 0; scan < count; scan++)
                 {
                     var ch = (byte)name[offset + count - scan - 1];
                     var isAlpha = (ch >= 'A' && ch <= 'Z') || (ch >= 'a' && ch <= 'z');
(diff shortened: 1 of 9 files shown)


tools/Microsoft.AspNet.Server.Kestrel.GeneratedCode/KnownHeaders.cs
```diff
@@ -59,7 +59,7 @@ protected string Term(string name, int offset, int count, string array, string s
             {
                 ulong mask = 0;
                 ulong comp = 0;
-                for (var scan = 0; scan != count; ++scan)
+                for (var scan = 0; scan < count; scan++)
```
Tratcher: @halter73 where is this file generated from?

benaadams: I think that *is* the generator; so its use is compile time, not necessarily perf sensitive.

I'm coming from the view that its best not to have examples of c++ iterator for-loop syntax in the code base as its not very good for c# for-loops which can have the extra overhead of bounds checking on each iteration if you don't follow the jit's happy pattern. As generations of devs will look to this and the other code bases for inspiration; but also probably not pay close attention to the context.

halter73: This is the generator.

@lodejard uses some some unconventional looping styles sometimes. I hadn't considered this would break JITing heuristics. I think that's a good enough reason to accept this PR.

benaadams: > unconventional looping styles 

Its how you get maximum flexibility in C++ so you can swap between and array and and iterator without rewriting the loop checks; however C# has extra overheads when doing loops and uses iterators/enumerators via ```foreach```; and the jitter is happy to give up early trying to workout if a loop is bounds safe. 

There is a better breakdown somewhere from the team, which I can't find, but this may provide some insight: http://www.codeproject.com/Articles/844781/Digging-Into-NET-Loop-Performance-Bounds-checking

benaadams:
Travis seems grumpy

Tratcher:
:shipit: Please squash.

benaadams:
Done

PR #238, opened 2015-10-03
Parallel Bit Count Fast

benaadams:
Faster count using parallel sideways addition


benaadams:
Probably needs a bit of testing; though not certain how to confirm the 2^64 combinations

benaadams:
Motivation: https://github.com/dotnet/corefx/issues/2209#issuecomment-139636114

src/Microsoft.AspNet.Server.Kestrel/Http/FrameHeaders.Generated.cs
```diff
@@ -5623,159 +5422,13 @@ public StringValues HeaderWWWAuthenticate
...
-                }
-            
-            return count;
+            // Parallel bit count for a 64-bit integer
+            var v = (ulong)_bits;
+            v = v - ((v >> 1) & 0x5555555555555555);
+            v = (v & 0x3333333333333333) + ((v >> 2) & 0x3333333333333333);
+            v = (v + (v >> 4) & 0x0f0f0f0f0f0f0f0f);
+            var count = (int)((v * 0x0101010101010101) >> 56);
+            return count + (MaybeUnknown?.Count ?? 0);
```
rynowak: do you care about boxing here?

benaadams: Definitely! Boxing, my shadowy nemesis... I can't see it though, what's boxing?

rynowak: IIRC `MaybeUnknown?.Count` has to hoist `.Count` into a `Nullable<int>`. Profiler would know for sure

benaadams: Did a dive, even installed [Clr Heap Allocation Analyzer](https://visualstudiogallery.msdn.microsoft.com/f9b47b93-8675-4ae0-9c52-5da8027c4bb8/) to see what it made of it; but didn't get any highlights.

I think its ok as ```Nullable<int>``` is a ```struct Nullable<T> where T : struct``` so is a struct containing a struct; hopefully doesn't do anything bad internally.

benaadams: Will do a memory profile on it just to be sure...

lodejard:
Nice! That's insane. For unit testing maybe add a few test cases by copying that algorithm specifically into a test class rather than test it indirectly via dictionary Count property? That was you can assert a handful of boundary conditions... Like all 0's==0, all 1's==64, a single 0 for each bit==63, a single 1 for each bit==1, and maybe a few others like every-odd-bit==32, every-even-bit==32, two-on-two-off, etc...

All just values for inline theory data of course

benaadams:
Could do bitcount as an extension method on a long? Would be able to test same code as is used? Where should it go though?

benaadams:
Something more like that with tests and whatnot?

benaadams:
Had merge issues so moved to  #245

PR #239, opened 2015-10-03
List/Array IFeatureCollection extras and Type object caching

benaadams:


src/Microsoft.AspNet.Server.Kestrel/Http/Frame.Generated.cs
```diff
@@ -7,109 +7,183 @@ namespace Microsoft.AspNet.Server.Kestrel.Http
...
 
         private object FastFeatureGet(Type key)
         {
-            if (key == typeof(global::Microsoft.AspNet.Http.Features.IHttpRequestFeature))
+            if (key == IHttpRequestFeatureType)
             {
-                return _currentIHttpRequestFeature;
+                if ((_featureOverridenFlags & flagIHttpRequestFeature) == 0L)
+                {
+                    return this as global::Microsoft.AspNet.Http.Features.IHttpRequestFeature;
```
benaadams: Not sure this casting is necessary?

lodejard:
Sorry, but can you rebase onto the PR that was just made?

tools/Microsoft.AspNet.Server.Kestrel.GeneratedCode/FrameFeatureCollection.cs
```diff
@@ -0,0 +1,144 @@
...
+{{
+    public partial class Frame
+    {{
+        {Each(commonFeatures.Select((feature, index) => new { feature, index }), entry => $@"
+        private const long flag{entry.feature.Name} = {1 << entry.index};")}
+
+        {Each(commonFeatures, feature => $@"
+        private static readonly Type {feature.Name}Type = typeof(global::{feature.FullName});")}
+
+        private long _featureOverridenFlags = 0L;
```
lodejard: I don't think we need the flag, TBH, because the distinction in all cases is "null or not" on the field. Know what I mean? Because the IFeatureCollection doesn't have Count, Remove, or ContainsKey, TryGet, there's no distinction between not being set and being null.

Plus you'd still need to null out all the fields on reset anyway, because you don't want a previously used collection to hold a reference to objects that were assigned in the previous request.

tools/Microsoft.AspNet.Server.Kestrel.GeneratedCode/FrameFeatureCollection.cs
```diff
@@ -0,0 +1,144 @@
...
+
+namespace Microsoft.AspNet.Server.Kestrel.Http 
+{{
+    public partial class Frame
+    {{
+        {Each(commonFeatures.Select((feature, index) => new { feature, index }), entry => $@"
+        private const long flag{entry.feature.Name} = {1 << entry.index};")}
+
+        {Each(commonFeatures, feature => $@"
+        private static readonly Type {feature.Name}Type = typeof(global::{feature.FullName});")}
```
lodejard: caching typeof has a measurable impact?

tools/Microsoft.AspNet.Server.Kestrel.GeneratedCode/FrameFeatureCollection.cs
benaadams: Was scrapping all the fields as they were all ```this``` references when not overridden;  however in the latest PR they aren't so am changing things a little.

benaadams: Updated, am using ```_featureOverridenFlags``` for ```IHttpRequestFeature```, ```IHttpResponseFeature```, ```IHttpUpgradeFeature``` which are generally ```this``` so use one int rather than 3 pointers for it.

tools/Microsoft.AspNet.Server.Kestrel.GeneratedCode/FrameFeatureCollection.cs
benaadams: I think so; but will bench

benaadams: Interestingly its a really terrible thing to do for equality tests; as that has a runtime fast path.

However, its good for variable assignment; so changed to do both approaches.

benaadams: Actually I need to retest; the equality was terrible with what is happening and didn't test cached type against Type variable (tested two typeof() equalling each other which is very fast)

benaadams: Closish; though comparing to typeof is better

lodejard:
I really don't like the additional complexity of putting interfaces into different categories, plus this would mean a feature-wrapping pattern around the implemented interfaces will go down the slow path rather than be backed by a field.

I do like the optimizations around type comparisons, using List instead of Dictionary, and we should order the interfaces by the frequency/likelihood we expect them to be gotted and setted. But please remove the different behaviors for groups of interfaces and bitmask, and go back to field per "well-known interface".

The frame itself isn't re-allocated per request, it's one per active client, so the members of this class won't be adding to the per-request memory pressure in a way that would make the _currentXxx field removal significant.

benaadams:
:ok:

benaadams:
Done

tools/Microsoft.AspNet.Server.Kestrel.GeneratedCode/FrameFeatureCollection.cs
```diff
@@ -25,26 +25,40 @@ public virtual void BeforeCompile(BeforeCompileContext context)
...
+            {
+                typeof(IHttpAuthenticationFeature),
                 typeof(IQueryFeature),
-                typeof(IFormFeature),
+                typeof(IFormFeature)
+            };
+
+            var sometimesFeatures = new[]
+            {
+                typeof(IHttpUpgradeFeature),
```
Tratcher: I called out these less common and rare features because I don't think they're worth first class treatment in this collection. The collection must handle arbitrary 3rd party features so these should just be lumped in with those.

benaadams: Only does always+common+implemented. ```IHttpUpgradeFeature``` gets picked up because its in the implemented group as part of frame so needs resetting back to ```this```

Tratcher: Ok, then why even list them here?

benaadams: They aren't in the generated code (though just removed some cached types); its more that their existence has been acknowledged and their fate decided. Whereas if they weren't there someone may add them - thinking they has been missed.

Probably worth adding a comment.

benaadams: Added comment to make this clearer

lodejard: @tratcher the fact they are less common or rare has no bearing. The collection has interfaces that are known in advance which is one less reason to create the list in any series of requests. Just like well-known headers in iis/httpsys structs. Those don't decide that headers which virtually never occur don't get an intptr's worth of memory to represent them.

The only thing we need to know is what feature types you can reach from our systems, to avoid creating the extra collection, and which are provided by the server, to avoid the 'as' casting from a known type.

I'm responding from a phone so it's a bit difficult to tell. Is that the state of this PR's implementation?

benaadams: @lodejard The 6 features currently not included with objects and which will use the backing list are

``IResponseCookiesFeature``
``IItemsFeature``
``ITlsConnectionFeature``
``IHttpWebSocketFeature``
``ISessionFeature``
``IHttpSendFileFeature``

Do you believe this to be incorrect?

lodejard:
Yes, sorry for the churn if I wasn't clear before. Interfaces are either known, or known and implemented. Please remove the other categorization.

benaadams:
Added all features back to object backing; still kept categorisations for ordering.

lodejard:
Thanks!

Issue #231, opened 2015-09-28
Bind to IPv6 loopback address when the "localhost" is specified


halter73:
Currently Kestrel only binds to the IPv4 loopback address (127.0.0.1) and not the IPv6 loopback address (::1) when localhost is specified.

#98

Tratcher:
Also for host names and *, bind to IPv4 Any (0.0.0.0) and IPv6 Any (::)

Tragetaschen:
Binding to IPv6Any already binds to IPv4Any. Libuv is using dual stack sockets by default and needs an explicit bind flag to disable that.

halter73:
@Tragetaschen Thanks for the heads up. It appears this only applies to IPv6Any. My testing shows that binding to IPv6Loopback does not bind to IPv4Loopback. Binding to IPv4Any also doesn't seem bind to IPv6Any.

grahamehorner:
not sure if my experience is related to this issue; but I'm unable to get kestrel to listen on the hostname when the host has multiple network interfaces; even the localhost is unable to access the web application via the hostname; however localhost shows that kestrel is listening on the loopback interface.

esko22:
Not sure if what I am seeing is also related but since switching to beta8 I have had a new issue pop up. Granted my use case is not common but I am running a Windows 10 VM on my Mac to host / debug my api host. Until switching to Beta8 I was able to make calls directly against the api hosted in the VM via port forwarding with no issues. 

The only way I can get this to work now is by launching Kestrel with the vm host name instead of localhost as I previously had in beta7. The odd thing is that on the Mac side I can make REST calls using either the vm host name or localhost and it resolves just fine. As soon as I try to host it on Kestrel with anything other then the VM host name the request times out with no error response.

Tratcher:
@esko22 I think this is expected. Before beta8 kestrel would always listen publicly on Any/*, no matter what you specified. Now if you specify localhost it listens only on `::1`. As far as the VM is concerned the Mac is a separate machine so you need to listen on a public IP or *, not localhost.

esko22:
Thanks @Tratcher. Using * works great.

cesarbs:
Is this still an issue? Is there something to be fixed?

Tratcher:
Yes, https://github.com/aspnet/KestrelHttpServer/pull/242#issuecomment-145688074

DamianEdwards:
post RC2

cesarbs:
[...]

Issue #234, opened 2015-09-30
Unable to read big json file from Request.Body


Tratcher:
_From @anfomin on September 30, 2015 15:1_

I'm trying to read big JSON file (UTF8) from `Request.Body` stream:
```C#
string content;
using (var sr = new StreamReader(Request.Body, Encoding.Default))
	content = sr.ReadToEnd();
```

At the beginning it's ok, but `content` variable contains invalid character sequence after some offset.

Sample repo: https://github.com/anfomin/DNX-MVC-bigfile. There is file `sample.json`. Just run `dnx kestrel` and make:
```http
POST http://localhost:5000
```
with body = `sample.json` file.

You will find mismatches since line 84.

Source:
![2015-09-30 19 58 00](https://cloud.githubusercontent.com/assets/1905503/10196698/9d2e92be-67ad-11e5-8948-254595776df4.png)

Result:
![2015-09-30 19 58 18](https://cloud.githubusercontent.com/assets/1905503/10196702/9e6a8c64-67ad-11e5-8ebc-f5e4432d113f.png)

I've tried latest DNX 1.0.0-beta8-15736 from https://www.myget.org/F/aspnetvnext/api/v2 and 1.0.0-rc1-15775 from https://www.myget.org/F/aspnetcidev/api/v2.

_Copied from original issue: aspnet/Mvc#3235_

Tratcher:
_From @Eilon on September 30, 2015 18:49_

@kichalla can you take a look?

Tratcher:
_From @kichalla on September 30, 2015 19:51_

@anfomin I suppose you are running on Mono on Linux? was curious if Encoding.Default was doing something here but looks like some content is entirely missing based on your snapshot above...just to isolate the issue, can you try saving the read content to a file and see if the contents are any different? also does this issue repro consistently for you?

Tratcher:
_From @anfomin on September 30, 2015 20:38_

@kichalla I tried this sample both on Windows and Mac, result is the same.

Also I tried to save `Request.Body` to a text file via `File.CreateText("...")`. Result file contains the same corruptions. You can remove `Encoding.Default`, the result does not change.

The problem is `Request.Body` because when I change input stream to `File.Open(...)` then output is correct.

Tratcher:
_From @kichalla on September 30, 2015 21:2_

@anfomin you are right, I was able to repro it on a Windows machine now..investigating...

left -> actual file content
right -> content read from request body and saved to a file
[...]

Issue #232, opened 2015-09-28
Multi-loop Kestrel on Windows blocked by libuv change


DamianEdwards:
We have a [PR out to libuv](https://github.com/libuv/libuv/pull/540) to fix an issue that will re-enable the multi-loop support on Windows that we lost when we stopped shipping our own fork of libuv.

muratg:
Just talked to @lodejard.  Moving is issues to RC2.

AspNetSmurfLab:
Latest info: the PR in libuv was merged! However, trying multi-loop on Windows with the updated libuv bits yields the following error:

```
Microsoft.AspNet.Server.Kestrel.Networking.UvException: Error -4071 EINVAL invalid argument
   at Microsoft.AspNet.Server.Kestrel.Networking.Libuv.Check(Int32 statusCode) in C:\code\github\aspnet\KestrelHttpServe
r\src\Microsoft.AspNet.Server.Kestrel\Networking\Libuv.cs:line 114
   at Microsoft.AspNet.Server.Kestrel.Networking.UvWriteReq.Write2(UvStreamHandle handle, ArraySegment`1 bufs, UvStreamH
andle sendHandle, Action`4 callback, Object state) in C:\code\github\aspnet\KestrelHttpServer\src\Microsoft.AspNet.Serve
r.Kestrel\Networking\UvWriteReq.cs:line 133
   at Microsoft.AspNet.Server.Kestrel.Http.ListenerPrimary.DispatchConnection(UvStreamHandle socket) in C:\code\github\a
spnet\KestrelHttpServer\src\Microsoft.AspNet.Server.Kestrel\Http\ListenerPrimary.cs:line 104
   at Microsoft.AspNet.Server.Kestrel.Networking.UvStreamHandle.UvConnectionCb(IntPtr handle, Int32 status) in C:\code\g
ithub\aspnet\KestrelHttpServer\src\Microsoft.AspNet.Server.Kestrel\Networking\UvStreamHandle.cs:line 137
   at Microsoft.AspNet.Server.Kestrel.Networking.Libuv.NativeMethods.uv_run(UvLoopHandle handle, Int32 mode)
   at Microsoft.AspNet.Server.Kestrel.Networking.Libuv.run(UvLoopHandle handle, Int32 mode) in C:\code\github\aspnet\Kes
trelHttpServer\src\Microsoft.AspNet.Server.Kestrel\Networking\Libuv.cs:line 150
[...]

PR #242, opened 2015-10-05
Add functional tests, bind to IPv6Any

Tratcher:
Functional tests for address bindings and a partial fix for #231 by binding to IPv6 by default. It uses dual mode sockets so binding to IPv6Any will also accept IPv4 requests.

Also includes disabled functional tests that repro #234 and #232.

@halter73 @lodejard @muratg


halter73:
This should fix https://github.com/aspnet/KestrelHttpServer/issues/231

Tratcher:
Only half of #231. It doesn't do the localhost -> ::1 & 127.0.0.1 binding.

halter73:
Yep. Just looked at your comment for the CorrectIPEndpointsAreCreated test. :shipit:

PR #243, opened 2015-10-06
Frame implementing mutable IFeatureCollection

lodejard:
Reducing allocations and avoiding dictionary operations for well-known feature interfaces

### DIFF
--- a/samples/SampleApp/Startup.cs
+++ b/samples/SampleApp/Startup.cs
@@ -6,6 +6,7 @@ using Microsoft.AspNet.Builder;
 using Microsoft.AspNet.Http;
 using Microsoft.Extensions.Logging;
 using Microsoft.Dnx.Runtime;
+using Microsoft.AspNet.Server.Kestrel;
 
 #if DNX451
 using System.IO;
--- a/samples/SampleApp/Startup.cs
+++ b/samples/SampleApp/Startup.cs
@@ -19,6 +19,9 @@ namespace SampleApp
     {
         public void Configure(IApplicationBuilder app, ILoggerFactory loggerFactory, IApplicationEnvironment env)
         {
+            var ksi = app.ServerFeatures[typeof(IKestrelServerInformation)] as IKestrelServerInformation;
+            //ksi.ThreadCount = 4;
+
             loggerFactory.MinimumLevel = LogLevel.Debug;
 
             loggerFactory.AddConsole(LogLevel.Debug);
@@ -45,6 +48,8 @@ namespace SampleApp
                     context.Request.PathBase,
                     context.Request.Path,
                     context.Request.QueryString);
+
+                foreach (var q in context.Request.Query) { }
 
                 context.Response.ContentLength = 11;
                 context.Response.ContentType = "text/plain";
--- a/samples/SampleApp/Startup.cs
+++ b/samples/SampleApp/Startup.cs
@@ -4,9 +4,10 @@
 using System;
 using Microsoft.AspNet.Builder;
 using Microsoft.AspNet.Http;
-using Microsoft.Extensions.Logging;
-using Microsoft.Dnx.Runtime;
+using Microsoft.AspNet.Http.Features;
 using Microsoft.AspNet.Server.Kestrel;
+using Microsoft.Dnx.Runtime;
+using Microsoft.Extensions.Logging;
 
 #if DNX451
 using System.IO;
@@ -20,7 +21,7 @@ namespace SampleApp
     {
         public void Configure(IApplicationBuilder app, ILoggerFactory loggerFactory, IApplicationEnvironment env)
         {
-            var ksi = app.ServerFeatures[typeof(IKestrelServerInformation)] as IKestrelServerInformation;
+            var ksi = app.ServerFeatures.Get<IKestrelServerInformation>();
             //ksi.ThreadCount = 4;
 
             loggerFactory.MinimumLevel = LogLevel.Debug;
@@ -49,8 +50,6 @@ namespace SampleApp
                     context.Request.PathBase,
                     context.Request.Path,
                     context.Request.QueryString);
-
-                foreach (var q in context.Request.Query) { }
 
                 context.Response.ContentLength = 11;
                 context.Response.ContentType = "text/plain";
--- a/src/Microsoft.AspNet.Server.Kestrel/Http/Frame.Generated.cs
+++ b/src/Microsoft.AspNet.Server.Kestrel/Http/Frame.Generated.cs
@@ -296,67 +296,67 @@ namespace Microsoft.AspNet.Server.Kestrel.Http
         {
             if ((_featureOverridenFlags & flagIHttpRequestFeature) == 0L)
             {
-                yield return new KeyValuePair<Type, object>(typeof(global::Microsoft.AspNet.Http.Features.IHttpRequestFeature), this as global::Microsoft.AspNet.Http.Features.IHttpRequestFeature);
+                yield return new KeyValuePair<Type, object>(IHttpRequestFeatureType, this as global::Microsoft.AspNet.Http.Features.IHttpRequestFeature);
             }
             if ((_featureOverridenFlags & flagIHttpResponseFeature) == 0L)
             {
-                yield return new KeyValuePair<Type, object>(typeof(global::Microsoft.AspNet.Http.Features.IHttpResponseFeature), this as global::Microsoft.AspNet.Http.Features.IHttpResponseFeature);
+                yield return new KeyValuePair<Type, object>(IHttpResponseFeatureType, this as global::Microsoft.AspNet.Http.Features.IHttpResponseFeature);
             }
             if ((_featureOverridenFlags & flagIHttpRequestIdentifierFeature) == 0L)
             {
-                yield return new KeyValuePair<Type, object>(typeof(global::Microsoft.AspNet.Http.Features.IHttpRequestIdentifierFeature), this as global::Microsoft.AspNet.Http.Features.IHttpRequestIdentifierFeature);
+                yield return new KeyValuePair<Type, object>(IHttpRequestIdentifierFeatureType, this as global::Microsoft.AspNet.Http.Features.IHttpRequestIdentifierFeature);
             }
             if ((_featureOverridenFlags & flagIHttpSendFileFeature) == 0L)
             {
-                yield return new KeyValuePair<Type, object>(typeof(global::Microsoft.AspNet.Http.Features.IHttpSendFileFeature), this as global::Microsoft.AspNet.Http.Features.IHttpSendFileFeature);
+                yield return new KeyValuePair<Type, object>(IHttpSendFileFeatureType, this as global::Microsoft.AspNet.Http.Features.IHttpSendFileFeature);
             }
             if ((_featureOverridenFlags & flagIServiceProvidersFeature) == 0L)
             {
-                yield return new KeyValuePair<Type, object>(typeof(global::Microsoft.AspNet.Http.Features.Internal.IServiceProvidersFeature), this as global::Microsoft.AspNet.Http.Features.Internal.IServiceProvidersFeature);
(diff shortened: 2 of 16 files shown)


samples/SampleApp/Startup.cs
```diff
@@ -19,6 +20,9 @@ public class Startup
     {
         public void Configure(IApplicationBuilder app, ILoggerFactory loggerFactory, IApplicationEnvironment env)
         {
+            var ksi = app.ServerFeatures[typeof(IKestrelServerInformation)] as IKestrelServerInformation;
```
Tratcher: `app.ServerFeatures.Get<IKestrelServerInformation>()`

samples/SampleApp/Startup.cs
```diff
@@ -46,6 +50,8 @@ public void Configure(IApplicationBuilder app, ILoggerFactory loggerFactory, IAp
                     context.Request.Path,
                     context.Request.QueryString);
 
+                foreach (var q in context.Request.Query) { }
```
Tratcher: ?

src/Microsoft.AspNet.Server.Kestrel/Http/Frame.Generated.cs
```diff
@@ -0,0 +1,297 @@
+
+using System;
+using System.Collections.Generic;
+
+namespace Microsoft.AspNet.Server.Kestrel.Http 
+{
+    public partial class Frame
+    {
+        
+        private object _currentIHttpRequestFeature;
```
Tratcher: Why are these `object` rather than their respective interfaces?

src/Microsoft.AspNet.Server.Kestrel/Http/Frame.Generated.cs
```diff
@@ -0,0 +1,297 @@
...
+            _currentIHttpRequestLifetimeFeature = this as global::Microsoft.AspNet.Http.Features.IHttpRequestLifetimeFeature;
+            _currentIQueryFeature = this as global::Microsoft.AspNet.Http.Features.Internal.IQueryFeature;
+            _currentIFormFeature = this as global::Microsoft.AspNet.Http.Features.Internal.IFormFeature;
+            _currentIResponseCookiesFeature = this as global::Microsoft.AspNet.Http.Features.Internal.IResponseCookiesFeature;
+            _currentIItemsFeature = this as global::Microsoft.AspNet.Http.Features.Internal.IItemsFeature;
+            _currentIHttpConnectionFeature = this as global::Microsoft.AspNet.Http.Features.IHttpConnectionFeature;
+            _currentITlsConnectionFeature = this as global::Microsoft.AspNet.Http.Features.ITlsConnectionFeature;
+            _currentIHttpUpgradeFeature = this as global::Microsoft.AspNet.Http.Features.IHttpUpgradeFeature;
+            _currentIHttpWebSocketFeature = this as global::Microsoft.AspNet.Http.Features.IHttpWebSocketFeature;
+            _currentISessionFeature = this as global::Microsoft.AspNet.Http.Features.ISessionFeature;
```
Tratcher: Why bother doing this `as` cast for features Frame does not currently implement, not to mention features that it probably never will implement?

samples/SampleApp/Startup.cs
```diff
@@ -6,6 +6,7 @@
 using Microsoft.AspNet.Http;
 using Microsoft.Extensions.Logging;
 using Microsoft.Dnx.Runtime;
+using Microsoft.AspNet.Server.Kestrel;
```
rynowak: sort

samples/SampleApp/Startup.cs
rynowak: Is this used for anything?

src/Microsoft.AspNet.Server.Kestrel/Http/Frame.Generated.cs
```diff
@@ -0,0 +1,297 @@
...
+        private object _currentIHttpRequestLifetimeFeature;
+        private object _currentIQueryFeature;
+        private object _currentIFormFeature;
+        private object _currentIResponseCookiesFeature;
+        private object _currentIItemsFeature;
+        private object _currentIHttpConnectionFeature;
+        private object _currentITlsConnectionFeature;
+        private object _currentIHttpUpgradeFeature;
+        private object _currentIHttpWebSocketFeature;
+        private object _currentISessionFeature;
```
Tratcher: You're trading CPU for object size which will cause more GCs. Almost no request will populate all of these together. Have you considered something more dynamic? E.g. do this for the few features that will be used on every request. For other features start with a `List<KVP<Type, object>>` and if it grows beyond some threshold then convert it to a `Dictionary<Type, object>>`.
A similar approach might be useful for headers.

Tratcher: Always present:
- IHttpRequestFeature
- IHttpResponseFeature
- IHttpRequestIdentifierFeature
- IServiceProvidersFeature
- IHttpRequestLifetimeFeature (eventually Kestrel will implement this)
- IHttpConnectionFeature (eventually Kestrel will implement this)

Common:
- IHttpAuthenticationFeature
- IQueryFeature
- IFormFeature

Sometimes:
- IResponseCookiesFeature
- IItemsFeature
- ITlsConnectionFeature (Client cert only)
- IHttpUpgradeFeature
- IHttpWebSocketFeature
- ISessionFeature

Rare/Never:
- IHttpSendFileFeature

davidfowl: Need to measure the differences before we can say for sure

test/Microsoft.AspNet.Server.KestrelTests/ConnectionFilterTests.cs
```diff
@@ -6,23 +6,26 @@
 using Microsoft.AspNet.Server.Kestrel.Filter;
 using Microsoft.AspNet.Server.Kestrel.Http;
 using Xunit;
+using Microsoft.AspNet.Http.Features;
```
Tratcher: nit: sort

test/Microsoft.AspNet.Server.KestrelTests/ChunkedResponseTests.cs
```diff
@@ -1,6 +1,7 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using Microsoft.AspNet.Http.Features;
```
Tratcher: nit: sort

samples/SampleApp/Startup.cs
lodejard: :ok:

src/Microsoft.AspNet.Server.Kestrel/Http/Frame.Generated.cs
```diff
@@ -0,0 +1,297 @@
...
+        private void FastReset()
+        {
+            _currentIHttpRequestFeature = this as global::Microsoft.AspNet.Http.Features.IHttpRequestFeature;
+            _currentIHttpResponseFeature = this as global::Microsoft.AspNet.Http.Features.IHttpResponseFeature;
+            _currentIHttpRequestIdentifierFeature = this as global::Microsoft.AspNet.Http.Features.IHttpRequestIdentifierFeature;
+            _currentIHttpSendFileFeature = this as global::Microsoft.AspNet.Http.Features.IHttpSendFileFeature;
+            _currentIServiceProvidersFeature = this as global::Microsoft.AspNet.Http.Features.Internal.IServiceProvidersFeature;
+            _currentIHttpAuthenticationFeature = this as global::Microsoft.AspNet.Http.Features.Authentication.IHttpAuthenticationFeature;
+            _currentIHttpRequestLifetimeFeature = this as global::Microsoft.AspNet.Http.Features.IHttpRequestLifetimeFeature;
+            _currentIQueryFeature = this as global::Microsoft.AspNet.Http.Features.Internal.IQueryFeature;
```
Tratcher: If we think `Internal` features like IQueryFeature are worth embedding then we should move them to the main Features package.

samples/SampleApp/Startup.cs
lodejard: :ok:

it's used to change ThreadCount (once that's working again)

samples/SampleApp/Startup.cs
lodejard: Was to force that feature interface to be created, but doesn't make sense to keep as part of the sample. Removed.

src/Microsoft.AspNet.Server.Kestrel/Http/Frame.Generated.cs
lodejard: Because they're only `[]` and `[]=`ed as object. Downcasting at this level could make the behavior different compared to `Dictionary<Type,Object>` on one hand, and would add a tiny bit of cpu on assignment that will be repeated anyway when a caller that asks for the feature downcasts again anyway.

src/Microsoft.AspNet.Server.Kestrel/Http/Frame.Generated.cs
lodejard: Yeah, I was thinking that also. The `as` cast is just so the generated code remains correct when features are added or removed from the frame.

That said - it's probably better to hard-code those assignments at code-generate time and put a comment on the Frame partial that implements interfaces telling the developer to change the related file.

lodejard: :ok:

test/Microsoft.AspNet.Server.KestrelTests/ConnectionFilterTests.cs
lodejard: :ok:

test/Microsoft.AspNet.Server.KestrelTests/ChunkedResponseTests.cs
lodejard: :ok:

src/Microsoft.AspNet.Server.Kestrel/Http/Frame.Generated.cs
lodejard: Sure, want to file an issue in the abstractions repo?

src/Microsoft.AspNet.Server.Kestrel/Http/Frame.Generated.cs
lodejard: It might also make a difference that the Frame object is allocated per-connection rather than per-request? So these objects fields aren't directly contributing to per-request memory pressure quite as aggressively.

The `List` vs `Dictionary` is a really interesting one. I do recall hearing that for small populations list scanning outperforms a hash table. Since the features added per request will always be a very limited set it might be better to use List instead of Dictionary all the time. Definitely worth measuring.

That said - we just declared all the well-known features used by our benchmarks as member fields... So it wouldn't actually make a difference to what we're measuring at this point. :\

benaadams: Picked up in https://github.com/aspnet/KestrelHttpServer/pull/239 List should be faster for < 16 items on x86 and < 8 items on x64 as its a single memory access to get the array in L1 cache (64 bytes); then iterating the array is just processing L1 cache.

Store the 3 types  ```IHttpRequestFeature```, ```IHttpResponseFeature```, ```IHttpUpgradeFeature``` in their common ```this``` state as a single ```int``` and only use object cache for the always and common features; and list beyond that.

muratg:
@lodejard Could you sign the CLA?  :smile:

lodejard:
@muratg I have twice now... No idea why the system keeps calling me out like that! I'll sign it again.

src/Microsoft.AspNet.Server.Kestrel/Http/Frame.cs
```diff
@@ -725,5 +728,5 @@ public bool StatusCanHaveBody(int statusCode)
                    statusCode != 205 &&
                    statusCode != 304;
         }
-    }
+            }
```
cesarbs: nit: fix indendation

halter73:
:shipit:

PR #245, opened 2015-10-07
Parallel Bit Count Fast

benaadams:
From #238

But with out horrible merginess issues


benaadams:
Need to add some tests as suggested by @lodejard https://github.com/aspnet/KestrelHttpServer/pull/238#issuecomment-146007312

benaadams:
Might change for similar code in corefx https://github.com/dotnet/corefx/blob/master/src/System.Reflection.Metadata/src/System/Reflection/Internal/Utilities/BitArithmetic.cs

lodejard:
Ah, nice find! If it matches similar code in corefx the need for test cases around the math probably even goes way down.

halter73:
I know this code pretty well. When Microsoft.AspNet.Server.Kestrel.GeneratedCode was first merged in I made a local branch where I compared the performance of all the bit counting algorithms described in this wikipedia article: https://en.wikipedia.org/wiki/Hamming_weight

You seem to be using the popcount_3 algorithm. All the algorithms on the wiki were faster than what we have currently have, but none of the algorithms made any significant difference in the context of of the overall request.

The main reason I held of on a PR though, was because I was unsure of the patent/copyright status of the algorithms on the Wikipedia page. Only [this algorithm](https://en.wikipedia.org/wiki/Hamming_weight#cite_ref-7) seemed old/obvious enough to certainly be in the clear. Interestingly that was the fasted algorithm given only a couple of set bits.

@Eilon What do we do in this case for small (4 line), semi-well known algorithms?

Eilon:
@halter73 let's take this offline.

benaadams:
Changed to use corefx implementation and added comment to source reflecting this.

halter73:
Nice. @davidfowl mentioned to me over lunch that we already use this algo in coreclr somewhere.

benaadams:
Can always ask them to surface it rather than marking it ```internal``` :stuck_out_tongue_closed_eyes:

benaadams:
Or expose the single cpu instruction for it :P https://github.com/dotnet/corefx/issues/2209

src/Microsoft.AspNet.Server.Kestrel/Infrastructure/LongExtensions.cs
```diff
@@ -0,0 +1,24 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.AspNet.Server.Kestrel.Extensions
+{
+    public static class LongExtensions
+    {
+        public static int BitCount(this long value)
```
davidfowl: internal

src/Microsoft.AspNet.Server.Kestrel/Infrastructure/LongExtensions.cs
```diff
@@ -0,0 +1,24 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.AspNet.Server.Kestrel.Extensions
+{
+    public static class LongExtensions
```
davidfowl: internal

davidfowl: Shouldn't be an extension method

benaadams:
Done

PR #253, opened 2015-10-12
Better handle blocks not taken taken from slabs

halter73:
- If the block needs to be larger than 4032 bytes it won't be backed by a slab
- Previously, MemoryPoolBlock2.Unpin was never called, but this doesn't really matter unless the block isn't backed by a slab
- Remove unused create function
- Optimize SocketInputStream.Dispose()
- If we want to add a test to verify blocks are unpinned, we will need to make some changes to MemoryPool2 so it can be mocked.


halter73:
@lodejard

lodejard:
:shipit:

Issue #129, opened 2015-07-14
Graceful crashing, logging


Tratcher:
Log all caught exceptions, keep running where possible.

Tratcher:
RE:
https://github.com/aspnet/KestrelHttpServer/issues/43
https://github.com/aspnet/KestrelHttpServer/issues/25

https://github.com/aspnet/KestrelHttpServer/blob/dev/src/Microsoft.AspNet.Server.Kestrel/Http/Frame.cs#L238

muratg:
@lodejard @Tratcher  Are you working on this? If not, @troydai should be able to look into this.

Tratcher:
@halter73 is making some improvements here. #254

PR #254, opened 2015-10-12
Log unhandled exceptions thrown from the AppFunc

halter73:


halter73:
@Tratcher @moozzyk

Tratcher:
Do you want to check these while you're at it? https://github.com/aspnet/KestrelHttpServer/issues/129

halter73:
@Tratcher @davidfowl @lodejard With this PR, Kestrel should no log all application errors and keep running where possible addressing #129.

I also did a little cleanup of our connection shutdown logic in Frame.ProduceEnd, since we gracefully shutdown the connection in Frame.RequestProcessingAsync.

src/Microsoft.AspNet.Server.Kestrel/Http/Frame.cs
```diff
@@ -261,12 +262,19 @@ private void FireOnStarting()
             {
                 foreach (var entry in onStarting)
                 {
-                    entry.Key.Invoke(entry.Value).Wait();
+                    try
```
Tratcher: This try/catch should be outside the foreach, we don't want to keep executing the OnStarted callbacks if the request has failed. OnCompleted is different.

src/Microsoft.AspNet.Server.Kestrel/Http/Frame.cs
```diff
@@ -411,28 +413,43 @@ public void ProduceContinue()
...
-        public void ProduceStart(bool immediate = true, bool appCompleted = false)
+        public async Task ProduceStartAndFireOnStarting(bool immediate = true)
+        {
+            if (_responseStarted) return;
+
+            await FireOnStarting();
+           
+            if (_applicationFailed) 
+            {
+                throw new ObjectDisposedException(typeof(Frame).FullName);
```
Tratcher: We should at least include a message saying that the response was aborted due to an application exception. We could even include the exception as a nested exception.

Tratcher:
:watch:

halter73:
@Tratcher good?

Tratcher:
:shipit:

PR #255, opened 2015-10-13
Make LibuvCopier work on CoreCLR even without DNX env variables

halter73:
- This should fix our AppVeyor builds

### DIFF
--- a/tools/Microsoft.AspNet.Server.Kestrel.LibuvCopier/Program.cs
+++ b/tools/Microsoft.AspNet.Server.Kestrel.LibuvCopier/Program.cs
@@ -25,9 +25,18 @@ namespace Microsoft.AspNet.Server.Kestrel.LibuvCopier
                 {
                     var dnxFolder = Environment.GetEnvironmentVariable("DNX_HOME");
 
-                    if (string.IsNullOrEmpty(dnxFolder))
+                    var firstCandidate = dnxFolder?.Split(';')
+                                                  ?.Select(path => Environment.ExpandEnvironmentVariables(path))
+                                                  ?.Where(path => Directory.Exists(path))
+                                                  ?.FirstOrDefault();
+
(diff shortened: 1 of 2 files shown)


halter73:
@moozzyk

tools/Microsoft.AspNet.Server.Kestrel.LibuvCopier/Program.cs
```diff
@@ -59,5 +56,23 @@ public void Main(string[] args)
...
+            return Environment.GetFolderPath(Environment.SpecialFolder.UserProfile);
+#else
+            if (_runtimeEnv.OperatingSystem == "Windows")
+            {
+                return Environment.GetEnvironmentVariable("USERPROFILE") ??
+                    Environment.GetEnvironmentVariable("HOMEDRIVE") + Environment.GetEnvironmentVariable("HOMEPATH");
+            }
+            else
+            {
+                return Environment.GetEnvironmentVariable("HOME");
```
akoeplinger: FYI this might not always be set, see https://github.com/dotnet/corefx/pull/2889. I guess it's time to expose this from CoreFX :smile:

tools/Microsoft.AspNet.Server.Kestrel.LibuvCopier/Program.cs
```diff
@@ -17,23 +25,12 @@ public void Main(string[] args)
                 {
                     var dnxFolder = Environment.GetEnvironmentVariable("DNX_HOME");
```
dougbu: When set `$env:DNX_HOME` may be a semicolon-separated list of paths. When not set, at least `dnvm.ps1` and `dnvm.sh` will fall back to checking `$env:DNX_USER_HOME` and `$env:DNX_GLOBAL_HOME` before assuming a subdirectory of the user's home.

PR #256, opened 2015-10-13
Minor perf and logging support changes

lodejard:
Uses header properties directly in a few places

Provides a value for Response.Stream.Length in support of that logging the response content size

### DIFF
--- a/src/Microsoft.AspNet.Server.Kestrel/Http/Frame.FeatureCollection.cs
+++ b/src/Microsoft.AspNet.Server.Kestrel/Http/Frame.FeatureCollection.cs
@@ -266,15 +266,14 @@ namespace Microsoft.AspNet.Server.Kestrel.Http
         {
             StatusCode = 101;
             ReasonPhrase = "Switching Protocols";
-            ResponseHeaders["Connection"] = "Upgrade";
-            if (!ResponseHeaders.ContainsKey("Upgrade"))
+            _responseHeaders.HeaderConnection = "Upgrade";
+
+            if (StringValues.IsNullOrEmpty(_responseHeaders.HeaderUpgrade) &&
+                !StringValues.IsNullOrEmpty(_requestHeaders.HeaderUpgrade))
             {
-                StringValues values;
-                if (RequestHeaders.TryGetValue("Upgrade", out values))
-                {
-                    ResponseHeaders["Upgrade"] = values;
-                }
+                _responseHeaders.HeaderUpgrade = _requestHeaders.HeaderUpgrade;
             }
+
             ProduceStart();
             return Task.FromResult(DuplexStream);
         }
--- a/src/Microsoft.AspNet.Server.Kestrel/Http/Frame.cs
+++ b/src/Microsoft.AspNet.Server.Kestrel/Http/Frame.cs
@@ -402,12 +402,16 @@ namespace Microsoft.AspNet.Server.Kestrel.Http
         {
             if (_responseStarted) return;
 
-            StringValues expect;
-            if (HttpVersion.Equals("HTTP/1.1") &&
-                RequestHeaders.TryGetValue("Expect", out expect) &&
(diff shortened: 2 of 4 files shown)


src/Microsoft.AspNet.Server.Kestrel/Http/FrameResponseStream.cs
```diff
@@ -23,13 +23,7 @@ public FrameResponseStream(FrameContext context)
...
         public override bool CanWrite => true;
 
-        public override long Length
-        {
-            get
-            {
-                throw new NotImplementedException();
-            }
-        }
+        public override long Length => Position;
```
Tratcher: Length is only expected to be available for seekable streams. https://msdn.microsoft.com/en-us/library/system.io.stream.length(v=vs.110).aspx `NotSupportedException: A class derived from Stream does not support seeking.`  Position has the same expectation. https://msdn.microsoft.com/en-us/library/system.io.stream.position(v=vs.110).aspx

halter73: Maybe we just add a new property to FrameResponseStream. `BytesWritten` perhaps?

@lodejard where were you planning on logging the this?

lodejard: From hosting, corresponding to "bytes written" you see in server logs.

halter73: Is this logging meant to be server agnostic? Would using Stream.Length work for WebListener?

Tratcher: No it won't, and it won't work for Nowin either. https://github.com/aspnet/Hosting/pull/403#discussion_r41928042

lodejard: Both of those could easily be updated. But the idea you suggested of having a request statistics feature should also be considered.

src/Microsoft.AspNet.Server.Kestrel/Http/Frame.cs
```diff
@@ -402,12 +402,16 @@ public void ProduceContinue()
...
             if (_responseStarted) return;
 
-            StringValues expect;
-            if (HttpVersion.Equals("HTTP/1.1") &&
-                RequestHeaders.TryGetValue("Expect", out expect) &&
-                (expect.FirstOrDefault() ?? "").Equals("100-continue", StringComparison.OrdinalIgnoreCase))
+            if (HttpVersion.Equals("HTTP/1.1"))
             {
-                SocketOutput.Write(_continueBytes);
+                foreach(var expect in _requestHeaders.HeaderExpect)
```
troydai: nit: format here?

muratg: space after foreach

src/Microsoft.AspNet.Server.Kestrel/Http/Frame.FeatureCollection.cs
```diff
@@ -266,15 +266,14 @@ Task<Stream> IHttpUpgradeFeature.UpgradeAsync()
         {
             StatusCode = 101;
             ReasonPhrase = "Switching Protocols";
-            ResponseHeaders["Connection"] = "Upgrade";
-            if (!ResponseHeaders.ContainsKey("Upgrade"))
+            _responseHeaders.HeaderConnection = "Upgrade";
+
+            if (StringValues.IsNullOrEmpty(_responseHeaders.HeaderUpgrade) &&
+                !StringValues.IsNullOrEmpty(_requestHeaders.HeaderUpgrade))
```
cesarbs: What if one of those is whitespace only?

halter73:
@lodejard do you still want to merge this?

halter73:
I think this is too stale to remain open.

Issue #237, opened 2015-10-02
UvReadCb allocating an Exception per request


rynowak:
I'm not sure what the root cause of this is because the exception is *never actually thrown*. Everything with the site seems to be working fine.

My profiler found this:
![image](https://cloud.githubusercontent.com/assets/1430011/10259194/51eb4b12-691a-11e5-9cfa-f6cfe6091ffc.png)

It's also allocating an `object[]` and calling `string.Concat` as well as a `UvException` which is what causes it to be so visible in the profile.

rynowak:
Looks like the exception is passed here and then ignored - https://github.com/aspnet/KestrelHttpServer/blob/318f3b7145a3aa96a5d90aad4e192d121b1bfb48/src/Microsoft.AspNet.Server.Kestrel/Http/ListenerSecondary.cs#L62

muratg:
Ugh that looks ugly. @halter73 @lodejard

PR #261, opened 2015-10-15
Don't allocate Exceptions unnecessarily

halter73:
- Allocate Exceptions in the ReadStart callbacks if necessary instead
  of in UvStreamHandle.
- This also fixes a bug in ListenerSecondary where it should have
  previously been looking at the error code instead of the read count.

#237

### DIFF
--- a/src/Microsoft.AspNet.Server.Kestrel/Http/Connection.cs
+++ b/src/Microsoft.AspNet.Server.Kestrel/Http/Connection.cs
@@ -13,7 +13,7 @@ namespace Microsoft.AspNet.Server.Kestrel.Http
 {
     public class Connection : ConnectionContext, IConnectionControl
     {
-        private static readonly Action<UvStreamHandle, int, int, Exception, object> _readCallback = ReadCallback;
+        private static readonly Action<UvStreamHandle, int, object> _readCallback = ReadCallback;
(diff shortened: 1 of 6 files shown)


halter73:
@lodejard @rynowak

rynowak:
:dancers:

src/Microsoft.AspNet.Server.Kestrel/Http/Connection.cs
```diff
@@ -113,28 +113,35 @@ private Libuv.uv_buf_t OnAlloc(UvStreamHandle handle, int suggestedSize)
...
         }
 
-        private void OnRead(UvStreamHandle handle, int readCount, int errorCode, Exception error)
+        private void OnRead(UvStreamHandle handle, int status)
         {
-            var normalRead = readCount != 0 && errorCode == 0;
-            var normalDone = readCount == 0 && (errorCode == 0 || errorCode == Constants.ECONNRESET || errorCode == Constants.EOF);
+            var normalRead = status > 0;
+            var normalDone = status == 0 || status == Constants.ECONNRESET || status == Constants.EOF;
             var errorDone = !(normalDone || normalRead);
```
rynowak: Seems like we can come up with less confusing names for all of these.

PR #268, opened 2015-10-20
Clean up

troydai:
Just clean up using statements in a few files I came across.


troydai:
Ping @halter73

halter73:
:shipit:

I would have probably snuck this in another PR :smile:

Issue #124, opened 2015-07-08
Path should be unescaped


lodejard:
Not sure if kestrel is doing this yet, but the path portion (not the query portion) of the request line should be unescaped.

troydai:
I'm validating current Kestrel behavior

troydai:
Looks like an issue:

send: `http://localhost:5000/path(%201)/path2?key=value%202&key2=value2`
kestrel:
```
GET /path(%25201)/path2?key=value%202&key2=value2
```

Tratcher:
Compare to weblistener. We should also check what happens when behind IIS.

troydai:
To: @Tratcher @lodejard @halter73 @blowdart 

Note:
- I'll add the logic to `https://github.com/aspnet/HttpAbstractions/tree/dev/src/Microsoft.Extensions.WebEncoders.Core` so as to able to share the path decoding loigc
- Different from Http.Sys, %2F will not be unescaped so as not to alter the meaning of an Url path.

troydai:
Fixed in https://github.com/aspnet/KestrelHttpServer/commit/52f4fa91e3f68c5f8a2e726432d85806024fd834

PR #269, opened 2015-10-20
Prototype: unescapse path

troydai:
Issue #124 

To unescape the path while read the first line to the request. 

Note:
- Implementation of the decoding is in `HttpAbstraction`. It is currently being reviewed: https://github.com/aspnet/HttpAbstractions/pull/448.
- A new method is added to `MemoryPoolIterator2` to return the raw `char[]` instead of a `string` as a result the decoding is able to be operated in place to avoid additional allocation.
- The change will not increase additional memory allocation.

Follow up:
- [ ] Add test cases.
- [ ] Performance review afterwards.

/cc @halter73 @Tratcher @davidfowl @lodejard @muratg

### DIFF
--- a/src/Microsoft.AspNet.Server.Kestrel/Http/Frame.cs
+++ b/src/Microsoft.AspNet.Server.Kestrel/Http/Frame.cs
@@ -12,6 +12,7 @@ using Microsoft.AspNet.Http;
 using Microsoft.AspNet.Server.Kestrel.Infrastructure;
 using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Primitives;
+using Microsoft.Extensions.WebEncoders;
 
 // ReSharper disable AccessToModifiedClosure
 
@@ -420,8 +421,8 @@ namespace Microsoft.AspNet.Server.Kestrel.Http
             if (_responseStarted) return;
 
             await FireOnStarting();
-           
-            if (_applicationException != null) 
+
+            if (_applicationException != null)
             {
                 throw new ObjectDisposedException(
                     "The response has been aborted due to an unhandled application exception.",
@@ -596,7 +597,9 @@ namespace Microsoft.AspNet.Server.Kestrel.Http
                 {
                     return false;
                 }
-                var requestUri = begin.GetString(scan);
+
+                var beginPath = begin;
+                var endPath = scan;
 
                 var queryString = "";
                 if (chFound == '?')
@@ -622,6 +625,11 @@ namespace Microsoft.AspNet.Server.Kestrel.Http
                 {
                     return false;
                 }
+
+                char[] requestUriChars;
+                var rawPathLength = beginPath.GetCharArray(endPath, out requestUriChars);
+                var decodedLength = UrlPathDecoder.DecodeInPlace(requestUriChars, rawPathLength);
+                var requestUri = new string(requestUriChars, 0, decodedLength);
 
                 consumed = scan;
                 Method = method;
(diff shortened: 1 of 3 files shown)


src/Microsoft.AspNet.Server.Kestrel/Http/Frame.cs
```diff
@@ -623,6 +626,11 @@ private bool TakeStartLine(SocketInput input)
                     return false;
                 }
 
+                char[] requestUriChars;
+                var rawPathLength = beginPath.GetCharArray(endPath, out requestUriChars);
+                var decodedLength = UrlPathDecoder.DecodeInPlace(requestUriChars, rawPathLength);
```
lodejard: DecodeInPlace should take beginPath/endPath iterators, modify the bytes in-place to unescape % sequences, and return the new endPath position rather than copy to an intermediate char[]. This also means the DecodeInPlace method does not need to understand utf-8 sequences, because the subsequent conversion from byte[] to string via UTF8 encoder will take care of that concern.

src/Microsoft.AspNet.Server.Kestrel/Http/Frame.cs
```diff
@@ -596,7 +597,9 @@ private bool TakeStartLine(SocketInput input)
                 {
                     return false;
                 }
-                var requestUri = begin.GetString(scan);
+
+                var beginPath = begin;
+                var endPath = scan;
```
lodejard: the line above `scan.Seek(' ', '?')` should be changed to `scan.Seek(' ', '?', '%', '+')` - only if chFound returns '%' or '+' would you set a flag `needsDecoding` and call `scan.Seek(' ', '?')` to find the real end of the path portion. in this way you skip the entire decoding step altogether for the majority of paths that contain entirely face-value bytes

troydai: Agree on seek the `%` to help skip unnecessary decoding. Is `+` really necessary?

Tratcher: `+` is not a valid uri encoding for space, it's primarily used in html encoding and sometimes in queries.

src/Microsoft.AspNet.Server.Kestrel/Http/Frame.cs
troydai: To achieve the goal of in place updating in the memory pool block I need to significantly change the `UrlPathDecoder` so that it accepts two `iterator` like objects to operator. The existing logic heavily depends on random access for verify UTF8 coding.

I think it's do doable but we won't be able to share the `Decoder` since it will be highly specified for `Kestrel`. And I'm more worried about progress at this point to redo this decoder at this point. I'm more than willing to make it happen after rc1.

src/Microsoft.AspNet.Server.Kestrel/Http/Frame.cs
```diff
@@ -623,6 +626,11 @@ private bool TakeStartLine(SocketInput input)
                     return false;
                 }
 
+                char[] requestUriChars;
```
Tratcher: requestPathChars

src/Microsoft.AspNet.Server.Kestrel/Http/Frame.cs
troydai: All right. I'm working out a prototype on this idea now. I'll send a separate PR for that. But I'd like to keep this floating as a fall back.

troydai: This is the prototype: https://github.com/aspnet/KestrelHttpServer/pull/274

troydai:
Replaced by https://github.com/aspnet/KestrelHttpServer/pull/274

PR #272, opened 2015-10-21
React to hosting changes

pakrym:
@Tratcher @davidfowl

### DIFF
--- a/samples/SampleApp/Program.cs
+++ b/samples/SampleApp/Program.cs
@@ -7,18 +7,10 @@ namespace SampleApp
 {
     public class Program
     {
-        private IServiceProvider _services;
-
-        public Program(IServiceProvider services)
+        public static void Main(string[] args)
         {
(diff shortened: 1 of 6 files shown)


samples/SampleApp/Program.cs
```diff
@@ -7,16 +7,9 @@ namespace SampleApp
...
-            _services = services;
-        }
-
-        public void Main(string[] args)
+        public static void Main(string[] args)
         {
-            new Microsoft.AspNet.Hosting.Program(_services).Main(new[] {
+            Microsoft.AspNet.Hosting.Program.Main(new[] {
                 "--server", "Microsoft.AspNet.Server.Kestrel"
             });
```
Tratcher: It looks like there's an existing bug here where it discards the incoming args. You can concat those like it does in the other Program.cs.

Tratcher:
:shipit:

PR #274, opened 2015-10-21
Unescape string in memory entirely.

troydai:
Escape the path url.

Feature:
- Operate completely in the memory pool without allocate memory
- UTF-8 decode once only.

/cc @lodejard @Tratcher @halter73 @muratg 

- [x] Add a `put` function to `MemoryPoolIterator2` for saving byte correctly
- [x] Add UTF8 verification during the pass

### DIFF
--- a/src/Microsoft.AspNet.Server.Kestrel/Http/Frame.cs
+++ b/src/Microsoft.AspNet.Server.Kestrel/Http/Frame.cs
@@ -420,8 +420,8 @@ namespace Microsoft.AspNet.Server.Kestrel.Http
             if (_responseStarted) return;
 
             await FireOnStarting();
-           
-            if (_applicationException != null) 
+
+            if (_applicationException != null)
             {
                 throw new ObjectDisposedException(
                     "The response has been aborted due to an unhandled application exception.",
@@ -591,12 +591,17 @@ namespace Microsoft.AspNet.Server.Kestrel.Http
 
                 scan.Take();
                 begin = scan;
-                var chFound = scan.Seek(' ', '?');
-                if (chFound == -1)
+
+                var needDecode = false;
+                var chFound = scan.Seek(' ', '?', '%');
+                if (chFound == '%')
                 {
-                    return false;
+                    needDecode = true;
+                    chFound = scan.Seek(' ', '?');
                 }
-                var requestUri = begin.GetString(scan);
+
+                var pathBegin = begin;
+                var pathEnd = scan;
 
                 var queryString = "";
                 if (chFound == '?')
@@ -623,9 +628,16 @@ namespace Microsoft.AspNet.Server.Kestrel.Http
                     return false;
                 }
 
+                if (needDecode)
+                {
+                    pathEnd = UrlPathDecoder.Unescape(pathBegin, pathEnd);
+                }
+
+                var requestUrlPath = pathBegin.GetString(pathEnd);
+
                 consumed = scan;
                 Method = method;
-                RequestUri = requestUri;
+                RequestUri = requestUrlPath;
                 QueryString = queryString;
                 HttpVersion = httpVersion;
                 Path = RequestUri;
--- a/src/Microsoft.AspNet.Server.Kestrel/Http/UrlPathDecoder.cs
+++ b/src/Microsoft.AspNet.Server.Kestrel/Http/UrlPathDecoder.cs
@@ -1,0 +1,306 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.AspNet.Server.Kestrel.Infrastructure;
+
+namespace Microsoft.AspNet.Server.Kestrel.Http
+{
+    public class UrlPathDecoder
+    {
+        /// <summary>
+        /// Unescapes the string between given memory iterators in place.
+        /// </summary>
+        /// <param name="start">The iterator points to the beginning of the sequence.</param>
+        /// <param name="end">The iterator points to the byte behind the end of the sequence.</param>
+        /// <returns>The iterator points to the byte behind the end of the processed sequence.</returns>
+        public static MemoryPoolIterator2 Unescape(MemoryPoolIterator2 start, MemoryPoolIterator2 end)
+        {
+            // the slot to read the input
+            var reader = start;
+
+            // the slot to write the unescaped byte
+            var writer = reader;
+
+            while (true)
+            {
+                if (CompareIterators(ref reader, ref end))
+                {
+                    return writer;
+                }
+
+                if (reader.Peek() == '%')
+                {
+                    var decodeReader = reader;
+
+                    // If decoding process succeeds, the writer iterator will be moved
+                    // to the next write-ready location. On the other hand if the scanned
+                    // percent-encodings cannot be interpreted as sequence of UTF-8 octets,
+                    // these bytes should be copied to output as is. 
+                    // The decodeReader iterator is always moved to the first byte not yet 
+                    // be scanned after the process. A failed decoding means the chars
+                    // between the reader and decodeReader can be copied to output untouched. 
+                    if (!DecodeCore(ref decodeReader, ref writer, end))
+                    {
+                        Copy(reader, decodeReader, ref writer);
+                    }
+
+                    reader = decodeReader;
+                }
+                else
+                {
+                    writer.Put((byte)reader.Take());
+                }
+            }
+        }
+
+        /// <summary>
+        /// Unescape the percent-encodings
+        /// </summary>
+        /// <param name="reader">The iterator point to the first % char</param>
+        /// <param name="writer">The place to write to</param>
+        /// <param name="end">The end of the sequence</param>
+        private static bool DecodeCore(ref MemoryPoolIterator2 reader, ref MemoryPoolIterator2 writer, MemoryPoolIterator2 end)
+        {
+            // preserves the original head. if the percent-encodings cannot be interpreted as sequence of UTF-8 octets,
+            // bytes from this till the last scanned one will be copied to the memory pointed by writer.
+            var byte1 = UnescapePercentEncoding(ref reader, end);
+            if (byte1 == -1)
+            {
+                return false;
+            }
+
+            if (byte1 <= 0x7F)
+            {
+                // first byte < U+007f, it is a single byte ASCII
+                writer.Put((byte)byte1);
+                return true;
+            }
+
+            int byte2 = 0, byte3 = 0, byte4 = 0;
+
+            // anticipate more bytes
+            var currentDecodeBits = 0;
+            var byteCount = 1;
+            var expectValueMin = 0;
+            if ((byte1 & 0xE0) == 0xC0)
+            {
+                // 110x xxxx, expect one more byte
+                currentDecodeBits = byte1 & 0x1F;
+                byteCount = 2;
+                expectValueMin = 0x80;
+            }
+            else if ((byte1 & 0xF0) == 0xE0)
+            {
+                // 1110 xxxx, expect two more bytes
+                currentDecodeBits = byte1 & 0x0F;
+                byteCount = 3;
+                expectValueMin = 0x800;
+            }
+            else if ((byte1 & 0xF8) == 0xF0)
+            {
+                // 1111 0xxx, expect three more bytes
+                currentDecodeBits = byte1 & 0x07;
+                byteCount = 4;
+                expectValueMin = 0x10000;
+            }
+            else
+            {
+                // invalid first byte
+                return false;
+            }
+
+            var remainingBytes = byteCount - 1;
+            while (remainingBytes > 0)
+            {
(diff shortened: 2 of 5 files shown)


src/Microsoft.AspNet.Server.Kestrel/Http/Decoder.cs
```diff
@@ -0,0 +1,117 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.AspNet.Server.Kestrel.Infrastructure;
+
+namespace Microsoft.AspNet.Server.Kestrel.Http
+{
+    public class Decoder
```
Tratcher: UrlPathDecoder

src/Microsoft.AspNet.Server.Kestrel/Http/Frame.cs
```diff
@@ -623,6 +634,13 @@ private bool TakeStartLine(SocketInput input)
                     return false;
                 }
 
+                if (needDecode)
+                {
+                    pathEnd = Decoder.Unescape(pathBegin, pathEnd);
+                }
+
+                var requestUri = pathBegin.GetString(pathEnd);
```
Tratcher: What happens for invalid utf-8 sequences? Throw? Abort the request?

src/Microsoft.AspNet.Server.Kestrel/Http/UrlPathDecoder.cs
```diff
@@ -0,0 +1,304 @@
...
+                writer.Put((byte)byte2);
+            }
+            if (byteCount > 2)
+            {
+                writer.Put((byte)byte3);
+            }
+            if (byteCount > 3)
+            {
+                writer.Put((byte)byte4);
+            }
```
troydai: Using these if-else structure to avoid allocate a byte array.

src/Microsoft.AspNet.Server.Kestrel/Http/UrlPathDecoder.cs
```diff
@@ -0,0 +1,304 @@
...
+                    byte2 = nextByte;
+                }
+                else if (byteCount - remainingBytes == 3)
+                {
+                    byte3 = nextByte;
+                }
+                else if (byteCount - remainingBytes == 4)
+                {
+                    byte4 = nextByte;
+                }
```
troydai: Using these if-else structure to avoid allocate a byte array.

troydai:
Ping

src/Microsoft.AspNet.Server.Kestrel/Http/UrlPathDecoder.cs
```diff
@@ -0,0 +1,304 @@
...
+                if (CompareIterators(ref reader, ref end))
+                {
+                    return writer;
+                }
+
+                if (reader.Peek() == '%')
+                {
+                    var decodeReader = reader;
+
+                    // If decoding process successes, the writer iterator will be moved
```
cesarbs: super nit: succeeds.

src/Microsoft.AspNet.Server.Kestrel/Http/Frame.cs
```diff
@@ -12,6 +12,7 @@
 using Microsoft.AspNet.Server.Kestrel.Infrastructure;
 using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Primitives;
+using Microsoft.Extensions.WebEncoders;
```
Tratcher: You're not using WebEncoders anymore, correct?

troydai: Correct. Will remove.

src/Microsoft.AspNet.Server.Kestrel/Http/UrlPathDecoder.cs
troydai: Thanks. Will update.

src/Microsoft.AspNet.Server.Kestrel/Http/Frame.cs
```diff
@@ -596,7 +597,17 @@ private bool TakeStartLine(SocketInput input)
                 {
                     return false;
                 }
-                var requestUri = begin.GetString(scan);
+
+                var needDecode = true;
+                var percent = begin;
+                var percentFound = percent.Seek('%');
+                if (percentFound == -1)
```
Tratcher: nit: `var needDecode = percentFound >= 0;` or `var needDecode = percent.Seek('%') >= 0;`

troydai: Will update.

test/Microsoft.AspNet.Server.KestrelTests/UrlPathDecoder.cs
```diff
@@ -0,0 +1,171 @@
...
+        [InlineData("%C1%BF")]
+        [InlineData("%E0%80%AF")]
+        [InlineData("%E0%9F%BF")]
+        [InlineData("%F0%80%80%AF")]
+        [InlineData("%F0%8F%8F%BF")]
+        // Incomplete
+        [InlineData("%")]
+        [InlineData("%%")]
+        [InlineData("%A")]
+        [InlineData("%Y")]
```
Tratcher: You should still assert the expected output here.

test/Microsoft.AspNet.Server.KestrelTests/UrlPathDecoder.cs
```diff
@@ -0,0 +1,171 @@
...
+        {
+            PositiveAssert(raw, expect);
+        }
+
+        [Theory]
+        // Test the borderline cases of overlong UTF8.
+        [InlineData("%C2%80")]
+        [InlineData("%C2%80")]
+        [InlineData("%E0%A0%80")]
+        [InlineData("%F0%90%80%80")]
```
Tratcher: Assert the expected output

test/Microsoft.AspNet.Server.KestrelTests/UrlPathDecoder.cs
```diff
@@ -0,0 +1,171 @@
...
+        [InlineData("%E0%9F%BF")]
+        [InlineData("%F0%80%80%AF")]
+        [InlineData("%F0%8F%8F%BF")]
+        // Incomplete
+        [InlineData("%")]
+        [InlineData("%%")]
+        [InlineData("%A")]
+        [InlineData("%Y")]
+        // [InlineData("http://xn--9zt52a.example.org/%e2%80%ae")]
+        public void InvalidUTF8(string raw)
```
Tratcher: What about mixed valid and invalid? `%%32`, `%C0%A4%32`, `%32%C0%A4`

troydai: Good point. I'll add this test.

test/Microsoft.AspNet.Server.KestrelTests/UrlPathDecoder.cs
troydai: Will comparing the result to itself.

test/Microsoft.AspNet.Server.KestrelTests/UrlPathDecoder.cs
troydai: Will update.

troydai:
Ping, updated.

test/Microsoft.AspNet.Server.KestrelTests/UrlPathDecoder.cs
troydai: Look into the `NegativeAssert` you will find it now compare the result to the raw string. For these negative case input equals to output.

test/Microsoft.AspNet.Server.KestrelTests/UrlPathDecoder.cs
```diff
@@ -0,0 +1,171 @@
...
+        // Test the borderline cases of overlong UTF8.
+        [InlineData("%C2%80", "\u0080")]
+        [InlineData("%E0%A0%80", "\u0800")]
+        [InlineData("%F0%90%80%80", "\U00010000")]
+        [InlineData("%63", "c")]
+        [InlineData("%32","2")]
+        [InlineData("%20"," ")]
+        [InlineData("%%32", "%2")]
+        [InlineData("%%20", "% ")]
+        [InlineData("%C0%A4%32", "%C0%A42")]
```
Tratcher: These mixed invalid&valid should be a separate test case.

src/Microsoft.AspNet.Server.Kestrel/Http/Frame.cs
```diff
@@ -596,7 +596,12 @@ private bool TakeStartLine(SocketInput input)
                 {
                     return false;
                 }
-                var requestUri = begin.GetString(scan);
+
+                var percent = begin;
+                var needDecode = percent.Seek('%') >= 0;
```
Tratcher: This should be after the query handling, you don't care if there are % in the query.

halter73: I think you'll need a version of Seek that takes an end iterator.

src/Microsoft.AspNet.Server.Kestrel/Infrastructure/MemoryPoolIterator2.cs
```diff
@@ -295,6 +298,45 @@ public int Seek(int char0, int char1)
             }
         }
 
+        /// <summary>
+        /// Save the data at the current location then move to the next available space.
+        /// </summary>
+        /// <param name="data">The byte to be saved.</param>
+        /// <returns>0 if the operation successes. -1 if can't find available space.</returns>
+        public int Put(byte data)
```
halter73: Why not return a bool and name it TryPut or something?

src/Microsoft.AspNet.Server.Kestrel/Infrastructure/MemoryPoolIterator2.cs
```diff
@@ -295,6 +298,45 @@ public int Seek(int char0, int char1)
...
+
+            var block = _block;
+            var index = _index;
+            while (true)
+            {
+                if (index < block.End)
+                {
+                    _block = block;
+                    _index = _index + 1;
+                    block.Array[index] = data;
```
halter73: I think you should exit the method here. You only when block.Next == null in this loop.

Add a test for this.

src/Microsoft.AspNet.Server.Kestrel/Http/Frame.cs
troydai: Adding a `Iterator.Seek(char0. char1, char2)`

src/Microsoft.AspNet.Server.Kestrel/Infrastructure/MemoryPoolIterator2.cs
troydai: Missing the `return` here. Will ensure a test is added to cover this scenario.

troydai:
Added iteration to fix a few bugs found in review. I'm adding test cases to cover them.

troydai:
Ping @lodejard @Tratcher @halter73 @muratg @davidfowl

troydai:
Ping

src/Microsoft.AspNet.Server.Kestrel/Http/Frame.cs
```diff
@@ -591,12 +591,38 @@ private bool TakeStartLine(SocketInput input)
 
                 scan.Take();
                 begin = scan;
-                var chFound = scan.Seek(' ', '?');
-                if (chFound == -1)
+
+                int chFound = -1;
```
Tratcher: ch? Character? Separator?

src/Microsoft.AspNet.Server.Kestrel/Http/Frame.cs
```diff
@@ -591,12 +591,38 @@ private bool TakeStartLine(SocketInput input)
...
+                bool needDecode = false;
+                while (true)
```
Tratcher: this while true is wierd. At most you're only ever going to call seek twice, correct?
```
chFound = scan.Seek(' ', '?', '%');
if (chFound == %)
{
  needsDecode = true;
  chFound = scan.Seek(' ', '?');
}
```

src/Microsoft.AspNet.Server.Kestrel/Http/Frame.cs
```diff
@@ -623,6 +628,13 @@ private bool TakeStartLine(SocketInput input)
                     return false;
                 }
 
+                if (needDecode)
+                {
+                    pathEnd = UrlPathDecoder.Unescape(pathBegin, pathEnd);
+                }
+
+                var requestUri = pathBegin.GetString(pathEnd);
```
Tratcher: requestPath

Tratcher:
:shipit:

troydai:
all right. I'll wait a bit for Travis.

Issue #270, opened 2015-10-21
Using Kestrel stand-alone sometimes gives 404's


Tragetaschen:
I thought I give Kestrel a try in beta8 on Windows. When I'm running from within Visual Studio through IIS Express, everything is fine, as is with running an Owin-based server stack on the command line. However, when I use Kestrel from the command line, I see reproducible 404's.

My application is as follows:
- An angular based SPA with static files being served to the browser
- A SignalR (the current 3.0.0 state) connection
- API requests to control the server and get additional data

After a refresh in the browser, everything is set up and the SignalR connection is running. The next POST request works and is handled correctly. If I "Edit and Resend" the request  in Firefox, it will 404 as does any other request. This also happens in Edge.

When I'm in this state:
- If I navigate away in my SPA (which aborts the SignalR connection) and come back, the status is reset and I have one succeeding request
- If I refresh the page, I will also get a 404, but the next refresh succeeds normally
- If I wait long enough (keep-alive?) exactly one request again succeeds

I have included a middleware in the request pipeline both before and after the `app.UseSignalR()` call in `Startup` and this sees all the request and the valid and correct path (the first one that succeeds and the same second one that fails).

If it helps, here is the Request:
```
…POST http://localhost:5000/api/v1/U100/Spectrum/Stop
Host: localhost:5000
User-Agent: Mozilla/5.0 (Windows NT 10.0; WOW64; rv:41.0) Gecko/20100101 Firefox/41.0
Accept: application/json, text/plain, */*
Accept-Language: en-US,en;q=0.5
Accept-Encoding: gzip, deflate
DNT: 1
Content-Type: application/json;charset=utf-8
Referer: http://localhost:5000/U100/Spectrum
Content-Length: 1
Connection: keep-alive
Pragma: no-cache
Cache-Control: no-cache

1
```

kayle:
I'm seeing what is probably a related issue. dnx kestrel works fine from the command line, but when running through IIS inside Visual Studio I'm getting 404 responses. I debugged it to the point of finding that the HttpMethodConstraint was failing since the request Method was set to "{}GET". Not sure where the extra "{}" prefix came from. Guessing the problem is from Frame.TakeStartLine not being at the right position in stream.

Only started seeing this problem after upgrading from beta7 to beta8.

Tragetaschen:
Yeah, that's it for me, too.

[...]

PR #275, opened 2015-10-21
Consume the full request body when the app does not

halter73:


src/Microsoft.AspNet.Server.Kestrel/Http/Frame.cs
```diff
@@ -25,6 +25,7 @@ public partial class Frame : FrameContext, IFrameControl
         private static readonly ArraySegment<byte> _continueBytes = CreateAsciiByteArraySegment("HTTP/1.1 100 Continue\r\n\r\n");
         private static readonly ArraySegment<byte> _emptyData = new ArraySegment<byte>(new byte[0]);
         private static readonly byte[] _hex = Encoding.ASCII.GetBytes("0123456789abcdef");
+        private static readonly Stream _nullStream = new NullStream();
```
Tratcher: Is Stream.Null available?

multiarc: We could use Dispose pattern on MessageBody to consume data that application not read to move input stream instead of read to buffer and write null. I can provide pull request.

multiarc:
We could use Dispose pattern on MessageBody to consume data that application not read to move input stream instead of read to buffer and write null. I can provide this with pull request.

src/Microsoft.AspNet.Server.Kestrel/Http/Frame.cs
halter73: I *knew* there must be an existing null stream implementation available. Thanks for pointing it out.

halter73:
@multiarc What makes your PR better? I like the CopyToAsync approach, because there is less new code.

multiarc:
CopyTo approach does the same, because ReadAsyncImplementation looks for the _inputLength and count down it.

src/Microsoft.AspNet.Server.Kestrel/Http/NullStream.cs
```diff
@@ -0,0 +1,69 @@
...
+        public override bool CanRead => false;
+        public override bool CanSeek => false;
+
+        public override bool CanWrite => true;
+
+        public override long Length
+        {
+            get
+            {
+                throw new NotImplementedException();
```
cesarbs: Wouldn't it be better to have all operations as noops, in case the implementation of CopyToAsync changes and starts using one of them?

Tratcher:
#270

Tratcher:
@muratg This is an important fix for RC1.

muratg:
@Tratcher Does this fix #270?  Since we have a PR already, I put it in RC1.

Tratcher:
Yes, failing to consume the full body causes corruption of the next request.

src/Microsoft.AspNet.Server.Kestrel/Http/NullStream.cs
halter73: I hope CopyToAsync doesn't start using Length on the destination stream, but I agree that it's better safe than sorry. I checked Stream.Null (which I'm switching to), and Length does not throw.

halter73:
@muratg @Tratcher I just rebased and reran tests after switching to Stream.Null and everything looks good.

I agree that @multiarc's change is more efficient, but it should only make much of a difference if the app doesn't fully consume the request body. I also think that this PR is safer for RC1 since it's smaller.

Am I good to merge this into dev?

Tratcher:
Good enough for RC1, but it wouldn't hurt to sanity check with Lou on the general flow.

PR #276, opened 2015-10-21
Properly handle headers with empty values

halter73:


Tratcher:
:shipit: Have you also filed a bug against HttpClient for sending the empty header?

halter73:
I just filed an issue about the empty header: https://github.com/dotnet/corefx/issues/4065

I might have to add some more details to that issue and file another issue about ByteArrayContent, but I'll do that tomorrow when I can hopefully get my dev environment behaving again.

Tratcher:
I'd guess ByteArrayContent is a manifestation of the same issue because it doesn't include a content-type header by default. I'd be curious to see the wireshark trace.

troydai:
:shipit:

halter73:
@lodejard I know you aren't a huge fan of creating an extra copy of the scan iterator to implement this fix, but are you OK with getting this PR in as is for RC1?

lodejard:
:shipit:

src/Microsoft.AspNet.Server.Kestrel/Http/Frame.cs
```diff
@@ -739,7 +752,7 @@ private bool TakeMessageHeaders(SocketInput input)
                         }
 
                         consumed = scan;
-                        _requestHeaders.Append(name.Array, name.Offset, name.Count, value);
+                        requestHeaders.Append(name.Array, name.Offset, name.Count, value);
```
davidfowl: Why this change

halter73: For testing.

PR #277, opened 2015-10-22
#270 Fix application POST stream consumption

multiarc:
Alternative to https://github.com/aspnet/KestrelHttpServer/pull/275

### DIFF
--- a/src/Microsoft.AspNet.Server.Kestrel/Http/MessageBody.cs
+++ b/src/Microsoft.AspNet.Server.Kestrel/Http/MessageBody.cs
@@ -10,7 +10,7 @@ using System.IO;
 
 namespace Microsoft.AspNet.Server.Kestrel.Http
 {
-    public abstract class MessageBody
+    public abstract class MessageBody : IDisposable
     {
         private FrameContext _context;
         private int _send100Continue = 1;
@@ -98,6 +98,13 @@ namespace Microsoft.AspNet.Server.Kestrel.Http
             return true;
         }
 
+        protected abstract void Dispose(bool disposing);
+
+        public void Dispose()
+        {
+            Dispose(true);
+            GC.SuppressFinalize(this);
+        }
 
         class ForRemainingData : MessageBody
         {
@@ -109,6 +116,11 @@ namespace Microsoft.AspNet.Server.Kestrel.Http
             public override Task<int> ReadAsyncImplementation(ArraySegment<byte> buffer, CancellationToken cancellationToken)
             {
                 return _context.SocketInput.ReadAsync(buffer);
+            }
+            
+            protected override void Dispose(bool disposing)
+            {
+                
             }
         }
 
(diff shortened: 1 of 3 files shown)


src/Microsoft.AspNet.Server.Kestrel/Http/MessageBody.cs
```diff
@@ -326,6 +349,17 @@ private static bool TakeChunkedLine(SocketInput baton, ref int chunkSizeOut)
                 }
             }
 
+            protected override void Dispose(bool disposing)
+            {
+                if (_inputLength > 0)
```
halter73: I think this only consumes the next chunk, not the entire request body.

src/Microsoft.AspNet.Server.Kestrel/Infrastructure/MemoryPoolIterator2.cs
```diff
@@ -137,6 +137,28 @@ public int Peek()
             }
         }
 
+        public MemoryPoolIterator2 Skip(int length)
+        {
+            var block = _block;
+            var index = _index;
+            var remaining = length;
+            var following = block.End - index;
```
troydai: The `following` should be recalculate when moves to a new block.

src/Microsoft.AspNet.Server.Kestrel/Infrastructure/MemoryPoolIterator2.cs
```diff
@@ -137,6 +137,28 @@ public int Peek()
...
+                {
+                    return new MemoryPoolIterator2(block, index + remaining);
+                }
+                if (block.Next == null)
+                {
+                    return new MemoryPoolIterator2(block, index + following);
+                }
+                remaining -= following;
+                block = block.Next;
+                index = block.Start;
```
troydai: The three branches of condition will be more readable if organized in the form of `if - else if - else`

src/Microsoft.AspNet.Server.Kestrel/Http/MessageBody.cs
multiarc: What if we have keep alive connection? Would this run several times for every chunk?

src/Microsoft.AspNet.Server.Kestrel/Infrastructure/MemoryPoolIterator2.cs
multiarc: Agree, let me fix.

src/Microsoft.AspNet.Server.Kestrel/Infrastructure/MemoryPoolIterator2.cs
multiarc: Not sure how much this would be readable, but added else statements.

halter73:
Closing this seen we merged https://github.com/aspnet/KestrelHttpServer/pull/300 and this doesn't consume all chunks.

If there is a measurable performance improvement, we can reconsider this approach.

PR #285, opened 2015-10-23
React to breaking changes in StringValues

rynowak:


test/Microsoft.AspNet.Server.KestrelTests/FrameRequestHeadersTests.cs
```diff
@@ -121,8 +121,8 @@ public void ContainsAndContainsKeyWork()
             IDictionary<string, StringValues> headers = new FrameRequestHeaders();
             var kv1 = new KeyValuePair<string, StringValues>("host", new[] { "localhost" });
             var kv2 = new KeyValuePair<string, StringValues>("custom", new[] { "value" });
-            var kv1b = new KeyValuePair<string, StringValues>("host", new[] { "localhost" });
-            var kv2b = new KeyValuePair<string, StringValues>("custom", new[] { "value" });
+            var kv1b = new KeyValuePair<string, StringValues>("host", new[] { "not-localhost" });
+            var kv2b = new KeyValuePair<string, StringValues>("custom", new[] { "not-value" });
```
rynowak: This test was verifying the wrong thing.

src/Microsoft.AspNet.Server.Kestrel/Http/FrameHeaders.cs
```diff
@@ -152,7 +152,7 @@ IEnumerator IEnumerable.GetEnumerator()
             StringValues value;
             return
                 TryGetValueFast(item.Key, out value) &&
-                object.Equals(value, item.Value) &&
+                value.Equals(item.Value) &&
```
rynowak: The old check here could never return true. You're boxing two structs to compare them for reference equality. That's always false.

Tratcher:
:shipit:

troydai:
:shipit: 

I'm looking for any other similar wrong use of value type in Kestrel.

Issue #286, opened 2015-10-24
[Perf] [Check] High Allocator DoWriteIfNeeded


benaadams:
Items 2 & 3 on highest allocators in plaintext test

![most-memory](http://aoa.blob.core.windows.net/aspnet/most-memory.png)

`Microsoft.AspNet.Server.Kestrel.Http.SocketOutput.WriteContext`

Allocations

![alloc](http://aoa.blob.core.windows.net/aspnet/DoWriteIfNeeded-allocations.png)

![alloc](http://aoa.blob.core.windows.net/aspnet/list-add-allocs.png)

Writes alloc on List.Add not `GCHandle.Alloc` for clarity

Bytes

![bytes](http://aoa.blob.core.windows.net/aspnet/DoWriteIfNeeded-bytes.png)

![bytes](http://aoa.blob.core.windows.net/aspnet/list-add-bytes.png)

Writes alloc on List.Add not `GCHandle.Alloc` for clarity

Further up the stack...

`DoPostWork` is burning 346,416 bytes in calling the `Action<object, object> CallbackAdapter` method (the byte difference in the method) though I'm not entirely sure why?

![bytes](http://aoa.blob.core.windows.net/aspnet/dopostwork-bytes.png)

benaadams:
Added PR for the closures; doesn't resolves everything though

benaadams:
Redoing post closure cleanup

PR #287, opened 2015-10-24
Reduce closure allocation

benaadams:
Removes a couple closure allocations identified in https://github.com/aspnet/KestrelHttpServer/issues/286

### DIFF
--- a/src/Microsoft.AspNet.Server.Kestrel/Http/Connection.cs
+++ b/src/Microsoft.AspNet.Server.Kestrel/Http/Connection.cs
@@ -12,8 +12,8 @@ namespace Microsoft.AspNet.Server.Kestrel.Http
 {
     public class Connection : ConnectionContext, IConnectionControl
     {
-        private static readonly Action<UvStreamHandle, int, object> _readCallback = ReadCallback;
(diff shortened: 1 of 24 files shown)


benaadams:
Still 15 closure captures; but size and number is reduced

benaadams:
2002 requests before 

img

2004 requests after

img

benaadams:
Needs to go further; some of the closures remaining are too heavy...

benaadams:
Next iteration for 2005 requests

allocs

Need to clean up names before I add commit

test/Microsoft.AspNet.Server.KestrelTests/Program.cs
```diff
@@ -27,7 +28,7 @@ public class Program
 
         public int Main()
         {
-            return new Xunit.Runner.Dnx.Program(env, sp, _libraryManager).Main(new string[]
+            return new Xunit.Runner.Dnx.Program(env, sp, _libraryManager, _shutdown).Main(new string[]
```
davidfowl: Revert that

benaadams:
No sure about this.

PR #292, opened 2015-10-27
Update System.Net.Sockets dependency in test project to 4.1.0

halter73:
- This only affects the test project
- The Socket.*Async extension methods are available for net46 but not net451
  so we ifdef.

### DIFF
--- a/test/Microsoft.AspNet.Server.KestrelTests/project.json
+++ b/test/Microsoft.AspNet.Server.KestrelTests/project.json
@@ -16,7 +16,7 @@
             "dependencies": {
                 "System.Diagnostics.TraceSource": "4.0.0-beta-*",
                 "System.IO": "4.0.11-beta-*",
-                "System.Net.Sockets": "4.0.10-beta-*",
+                "System.Net.Sockets": "4.1.0-beta-*",
                 "System.Runtime.Handles": "4.0.1-beta-*"
             }
         }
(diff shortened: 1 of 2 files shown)


test/Microsoft.AspNet.Server.KestrelTests/project.json
```diff
@@ -1,12 +1,13 @@
...
-        "xunit.runner.aspnet": "2.0.0-aspnet-*",
+        "System.Net.Sockets": "4.1.0-beta-*",
         "Microsoft.AspNet.Server.Kestrel": "1.0.0-*",
-        "Microsoft.AspNet.Testing": "1.0.0-*"
+        "Microsoft.AspNet.Testing": "1.0.0-*",
+        "xunit.runner.aspnet": "2.0.0-aspnet-*"
     },
     "frameworks": {
-        "dnx451": {
+        "dnx46": {
```
davidfowl: Nah, this will start to fail on mono

halter73: :( I guess I have to ifdef then?

davidfowl: yah

halter73:
Now with more `#if`!

davidfowl:
Look at that the CI is green! :shipit:

troydai:
:shipit:

halter73:
We're still waiting on a CoherenceSigned build before pushing anything to dev, correct?

Issue #244, opened 2015-10-07
Use DllImport for Libuv native binaries


davidfowl:
/cc @moozzyk

moozzyk:
Depends on https://github.com/aspnet/dnx/issues/402

muratg:
@moozzyk where are we with this one?

moozzyk:
@muratg - this worked on Friday. Need to retest after latest changes to native lib loading and possibly change a layout a little bit - "osx" is not a valid fallback anymore. One breaking change is that it will require to link the libuv as `libuv.so` and not `libuv.so.1` on Linux (I think this is why Travis is failing on my branch). Will try to send a PR for this but ultimately in the worst case RC1 can ship with what we have right now.

moozzyk:
Fixed in d4853f9b7c31a28cecac8ac7e965f8f2f783c06b. Needs an announcement for libuv name change on Linux

muratg:
@moozzyk Please post the announcement issue when you're back. That would be still before RC1 release.

PR #293, opened 2015-10-27
Switching to using the built-in loading of native libs

moozzyk:
#244

### DIFF
--- a/src/Microsoft.AspNet.Server.Kestrel/KestrelEngine.cs
+++ b/src/Microsoft.AspNet.Server.Kestrel/KestrelEngine.cs
@@ -3,14 +3,12 @@
 
 using System;
 using System.Collections.Generic;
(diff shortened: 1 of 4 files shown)


davidfowl:
!!!!!!!!!!!! YASSS !!!!

src/Microsoft.AspNet.Server.Kestrel/KestrelEngine.cs
```diff
@@ -20,45 +18,6 @@ public KestrelEngine(ILibraryManager libraryManager, ServiceContext context)
             : this(context)
         {
             Libuv = new Libuv();
-
```
davidfowl: Remove the ILibraryManager ctor arguments.

davidfowl:
:shipit:

PR #294, opened 2015-10-27
Removing dependency on Microsoft.Extensions.PlatformAbstractions

moozzyk:
The test runner still needs PlatformAbstractions


davidfowl:
:shipit:

PR #295, opened 2015-10-27
Don't call uname repeatedly

moozzyk:
We seem to call uname each time we want to allocate a buffer to
check we are running on non-Windows


davidfowl:
We should avoid this altogether actually.

moozzyk:
Actually I broke it in my PR - we did use to cache it. Will merge back caching. Removal will be separate.

Issue #296, opened 2015-10-27
Consuming request body allocates a lot


rynowak:
Introduced by this change: https://github.com/aspnet/KestrelHttpServer/commit/fb01ea39185c857dcf8230922a648e9b242427f3

Allocations for 3000 requests to a simple API site:
![image](https://cloud.githubusercontent.com/assets/1430011/10774035/23285594-7cbc-11e5-866d-05ba3bb2ce85.png)

This is 80kb per request. The `Stream` needs to override `CopyToAsync`

Tratcher:
CopyToAsync isn't virtual. Really you want a Drain method that re-uses the buffer per connection.

rynowak:
https://msdn.microsoft.com/en-us/library/hh158751(v=vs.110).aspx

The longest overload is

benaadams:
Should be 81Kb? Just read to null buffer rather than Copy? (Read+Write)

https://github.com/aspnet/KestrelHttpServer/pull/300

rynowak:
Oh my! you're right, numbers are hard. That is 80kb.

I was wondering why my buffer pooling prototype improved allocations by -240mb :laughing:

benaadams:
Should also be resolved :+1:

PR #298, opened 2015-10-28
Don't call RequestBody.CopyToAsync(Stream.Null) unnecessarily

halter73:
- The default CopyToAsync implementation allocates an ~82KB buffer

https://github.com/aspnet/KestrelHttpServer/issues/296

### DIFF
--- a/src/Microsoft.AspNet.Server.Kestrel/Http/Frame.cs
+++ b/src/Microsoft.AspNet.Server.Kestrel/Http/Frame.cs
@@ -197,8 +197,11 @@ namespace Microsoft.AspNet.Server.Kestrel.Http
 
                             await ProduceEnd();
 
-                            // Finish reading the request body in case the app did not.
-                            await RequestBody.CopyToAsync(Stream.Null);
+                            if (!MessageBody.LocalIntakeFin)
+                            {
+                                // Finish reading the request body in case the app did not.
+                                await RequestBody.CopyToAsync(Stream.Null);
+                            }
                         }
 
                         terminated = !_keepAlive;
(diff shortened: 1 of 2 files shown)


src/Microsoft.AspNet.Server.Kestrel/Http/MessageBody.cs
```diff
@@ -98,7 +99,6 @@ public static bool TryGet(IDictionary<string, StringValues> headers, string name
             return true;
         }
 
-
         class ForRemainingData : MessageBody
```
Tratcher: Which MessageBody implementation is used for a normal GET request?

halter73: `ForContentLength` with length 0. You can see how it's selected in [MassageBody.For](https://github.com/aspnet/KestrelHttpServer/blob/shalter/copy-to-less/src/Microsoft.AspNet.Server.Kestrel/Http/MessageBody.cs#L73).

halter73:
It's probably worth pointing out that this PR doesn't change anything if:

- The app doesn't fully consume the request body
OR
- The request is not a keep alive request AND the request body is empty

I still think this covers most cases since HTTP/1.1 defaults to keep alive requests.

src/Microsoft.AspNet.Server.Kestrel/Http/Frame.cs
```diff
@@ -197,8 +197,11 @@ public Task Stop()
 
                             await ProduceEnd();
 
-                            // Finish reading the request body in case the app did not.
-                            await RequestBody.CopyToAsync(Stream.Null);
+                            if (!MessageBody.LocalIntakeFin)
```
halter73: This could be `!MessageBody.LocalIntakeFin && _keepAlive`. If it's not a keep-alive request there is no risk of corrupting the next request.

Is this better?

Tratcher: Even in a non-keep-alive request you should gracefully drain the request body. This helps avoid backlog conditions on the client.

halter73:
This was fixed by https://github.com/aspnet/KestrelHttpServer/pull/300 instead.

PR #300, opened 2015-10-28
Read to null buffer

benaadams:
Read to null buffer rather than Stream.Null as the CopyToAsync will
create a `new byte[]` on each call
https://github.com/dotnet/coreclr/blob/bc146608854d1db9cdbcc0b08029a87754e12b49/src/mscorlib/src/System/IO/Stream.cs#L218
of size 81920 bytes!
https://github.com/dotnet/coreclr/blob/bc146608854d1db9cdbcc0b08029a87754e12b49/src/mscorlib/src/System/IO/Stream.cs#L48

### DIFF
--- a/src/Microsoft.AspNet.Server.Kestrel/Http/Frame.cs
+++ b/src/Microsoft.AspNet.Server.Kestrel/Http/Frame.cs
@@ -29,6 +29,7 @@ namespace Microsoft.AspNet.Server.Kestrel.Http
         private readonly object _onStartingSync = new Object();
         private readonly object _onCompletedSync = new Object();
         private readonly FrameRequestHeaders _requestHeaders = new FrameRequestHeaders();
+        private readonly byte[] _nullBuffer = new byte[4096];
         private readonly FrameResponseHeaders _responseHeaders = new FrameResponseHeaders();
 
         private List<KeyValuePair<Func<object, Task>, object>> _onStarting;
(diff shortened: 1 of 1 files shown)


src/Microsoft.AspNet.Server.Kestrel/Http/Frame.cs
```diff
@@ -198,7 +199,10 @@ public Task Stop()
                             await ProduceEnd();
 
                             // Finish reading the request body in case the app did not.
-                            await RequestBody.CopyToAsync(Stream.Null);
+                            while (await RequestBody.ReadAsync(_nullBuffer, 0, _nullBuffer.Length) != 0)
+                            {
+                                ;
```
troydai: I think it's fine to leave the while body empty.

halter73:
Nice. I think I prefer this to https://github.com/aspnet/KestrelHttpServer/pull/298

Tratcher:
@rynowak @muratg

muratg:
:+1:

src/Microsoft.AspNet.Server.Kestrel/Http/Frame.cs
benaadams: Might add comment instead? I find empty loop bodies jarring - like the loop is there by mistake

rynowak: That's what I'd do :+1: Always follow your :heart:

Issue #302, opened 2015-10-28
Don't mask libuv load failures with a NullReferenceException


halter73:
When Kestrel fails to load libuv, instead of giving a clear error such as `System.DllNotFoundException: Unable to load DLL 'libuv': The specified module could not be found.`, the following error is given:

    System.NullReferenceException: Object reference not set to an instance of an object.
       at Microsoft.AspNet.Server.Kestrel.Networking.UvAsyncHandle.Send()
       at Microsoft.AspNet.Server.Kestrel.KestrelThread.Post(Action`1 callback, Object state)
       at Microsoft.AspNet.Server.Kestrel.KestrelThread.Stop(TimeSpan timeout)
       at Microsoft.AspNet.Server.Kestrel.KestrelEngine.Dispose()
       at Microsoft.AspNet.Server.Kestrel.ServerFactory.<>c__DisplayClass4_0.<Start>b__0()
       at Microsoft.AspNet.Server.Kestrel.Disposable.Dispose(Boolean disposing)
       at Microsoft.AspNet.Server.Kestrel.Disposable.Dispose()
       at Microsoft.AspNet.Server.Kestrel.ServerFactory.Start(IFeatureCollection serverFeatures, Func`2 application)
       at Microsoft.AspNet.Hosting.Internal.HostingEngine.Start()
       at Microsoft.AspNet.Hosting.WebApplication.Run(Type startupType, String[] args)
       at Microsoft.AspNet.Hosting.WebApplication.Run(String[] args)
       at Microsoft.AspNet.Hosting.Program.Main(String[] args)
       at Microsoft.AspNet.Server.Kestrel.Program.Main(String[] args)
       at System.RuntimeMethodHandle.InvokeMethod(Object target, Object[] arguments, Signature sig, Boolean constructor)
       at System.Reflection.RuntimeMethodInfo.UnsafeInvokeInternal(Object obj, Object[] parameters, Object[] arguments)
       at System.Reflection.RuntimeMethodInfo.Invoke(Object obj, BindingFlags invokeAttr, Binder binder, Object[] parameters, CultureInfo culture)
       at Microsoft.Dnx.Runtime.Common.EntryPointExecutor.Execute(Assembly assembly, String[] args, IServiceProvider serviceProvider)
    --- End of stack trace from previous location where exception was thrown ---
       at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
[...]

PR #303, opened 2015-10-28
Do nothing in KestrelThread.Stop if libuv fails to load

halter73:
This prevents DllNotFoundExceptions from being masked by NullReferenceExceptions

 #302

### DIFF
--- a/src/Microsoft.AspNet.Server.Kestrel/Infrastructure/KestrelThread.cs
+++ b/src/Microsoft.AspNet.Server.Kestrel/Infrastructure/KestrelThread.cs
@@ -30,6 +30,7 @@ namespace Microsoft.AspNet.Server.Kestrel
         private Queue<CloseHandle> _closeHandleRunning = new Queue<CloseHandle>();
         private object _workSync = new Object();
         private bool _stopImmediate = false;
+        private bool _initCompleted = false;
         private ExceptionDispatchInfo _closeError;
         private IKestrelTrace _log;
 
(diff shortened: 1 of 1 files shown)


halter73:
The new output is:

    System.DllNotFoundException: Unable to load DLL 'libuv': The specified module could not be found.
     (Exception from HRESULT: 0x8007007E)
       at Microsoft.AspNet.Server.Kestrel.Networking.Libuv.NativeMethods.uv_loop_size()
       at Microsoft.AspNet.Server.Kestrel.Networking.Libuv.loop_size()
       at Microsoft.AspNet.Server.Kestrel.Networking.UvLoopHandle.Init(Libuv uv)
       at Microsoft.AspNet.Server.Kestrel.KestrelThread.ThreadStart(Object parameter)
 
instead of the NullRef output shown in the issue.

davidfowl:
This needs to go into release

src/Microsoft.AspNet.Server.Kestrel/Infrastructure/KestrelThread.cs
```diff
@@ -30,6 +30,7 @@ public class KestrelThread
         private Queue<CloseHandle> _closeHandleRunning = new Queue<CloseHandle>();
         private object _workSync = new Object();
         private bool _stopImmediate = false;
+        private bool _initCompleted = false;
```
Eilon: For future reference: don't need to initialize fields to default values.

halter73: I prefer [the terser route](https://github.com/aspnet/KestrelHttpServer/commit/3fb33119eec2550a9434e8761ad4c744bd528370), but when in Rome...

Eilon: I believe Rome eventually burned down. We don't want that :smile:

Eilon: (But I don't care for this checkin.)

Issue #258, opened 2015-10-13
Call new IHttpContextFactory.Dispose(...) from request end


Eilon:
In here: https://github.com/aspnet/KestrelHttpServer/blob/dev/src/Microsoft.AspNet.Server.Kestrel/Http/Frame.cs#L175-L196

Before the `try` do the create, and in the `finally` do the dispose.

See new API in https://github.com/aspnet/Hosting/issues/404.

muratg:
Same as WebListener, your hosting changes have potential to break these.

JunTaoLuo:
merged to dev

PR #307, opened 2015-10-29
Reacting to Hosting changes

JunTaoLuo:
#258 and also responding to https://github.com/aspnet/Hosting/pull/445

### DIFF
--- a/src/Microsoft.AspNet.Server.Kestrel/Http/Frame.cs
+++ b/src/Microsoft.AspNet.Server.Kestrel/Http/Frame.cs
@@ -206,9 +206,10 @@ namespace Microsoft.AspNet.Server.Kestrel.Http
                         ResponseBody = new FrameResponseStream(this);
                         DuplexStream = new FrameDuplexStream(RequestBody, ResponseBody);
 
+                        var httpContext = HttpContextFactory.Create(this);
                         try
                         {
-                            await Application.Invoke(this).ConfigureAwait(false);
+                            await Application.Invoke(httpContext).ConfigureAwait(false);
                         }
                         catch (Exception ex)
                         {
@@ -226,6 +227,8 @@ namespace Microsoft.AspNet.Server.Kestrel.Http
                             }
 
                             await FireOnCompleted();
+
+                            HttpContextFactory.Dispose(httpContext);
 
                             await ProduceEnd();
 
--- a/src/Microsoft.AspNet.Server.Kestrel/Http/Listener.cs
+++ b/src/Microsoft.AspNet.Server.Kestrel/Http/Listener.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Threading.Tasks;
+using Microsoft.AspNet.Http;
 using Microsoft.AspNet.Server.Kestrel.Networking;
 using Microsoft.Extensions.Logging;
 
@@ -23,7 +24,7 @@ namespace Microsoft.AspNet.Server.Kestrel.Http
         public Task StartAsync(
             ServerAddress address,
             KestrelThread thread,
-            Func<Frame, Task> application)
+            RequestDelegate application)
         {
             ServerAddress = address;
             Thread = thread;
--- a/src/Microsoft.AspNet.Server.Kestrel/KestrelServer.cs
+++ b/src/Microsoft.AspNet.Server.Kestrel/KestrelServer.cs
@@ -1,0 +1,128 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using Microsoft.AspNet.Hosting;
+using Microsoft.AspNet.Hosting.Server;
+using Microsoft.AspNet.Http;
+using Microsoft.AspNet.Http.Features;
+using Microsoft.AspNet.Server.Kestrel.Http;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.AspNet.Server.Kestrel
+{
+    public class KestrelServer : IServer
+    {
+        private Stack<IDisposable> _disposables;
+        private readonly IApplicationLifetime _applicationLifetime;
+        private readonly ILogger _logger;
+        private readonly IHttpContextFactory _httpContextFactory;
+
+        public KestrelServer(IFeatureCollection features, IApplicationLifetime applicationLifetime, ILogger logger, IHttpContextFactory httpContextFactory)
+        {
+            if (features == null)
+            {
+                throw new ArgumentNullException(nameof(features));
+            }
+
+            if (applicationLifetime == null)
+            {
+                throw new ArgumentNullException(nameof(applicationLifetime));
+            }
+
+            if (logger == null)
+            {
+                throw new ArgumentNullException(nameof(logger));
+            }
+
+            if (httpContextFactory == null)
+            {
+                throw new ArgumentNullException(nameof(httpContextFactory));
+            }
+
+            _applicationLifetime = applicationLifetime;
+            _logger = logger;
+            Features = features;
+            _httpContextFactory = httpContextFactory;
+        }
+
+        public IFeatureCollection Features { get; }
+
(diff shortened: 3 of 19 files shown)


src/Microsoft.AspNet.Server.Kestrel/Server.cs
```diff
@@ -0,0 +1,120 @@
...
+namespace Microsoft.AspNet.Server.Kestrel
+{
+    public class Server : IServer
+    {
+        private IApplicationLifetime _applicationLifetime;
+        private ILogger _logger;
+        private Stack<IDisposable> _disposables;
+        private IHttpContextFactory _httpContextFactory;
+
+        public IFeatureCollection Features { get; }
```
troydai: Put property after constructor.

src/Microsoft.AspNet.Server.Kestrel/Server.cs
```diff
@@ -0,0 +1,120 @@
...
+
+        public Server(IFeatureCollection features, IApplicationLifetime applicationLifetime, ILogger logger, IHttpContextFactory httpContextFactory)
+        {
+            if (features == null)
+            {
+                throw new ArgumentNullException(nameof(features));
+            }
+            if (applicationLifetime == null)
```
troydai: Empty line before if statement

src/Microsoft.AspNet.Server.Kestrel/Server.cs
```diff
@@ -0,0 +1,120 @@
...
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.AspNet.Server.Kestrel
+{
+    public class Server : IServer
+    {
+        private IApplicationLifetime _applicationLifetime;
+        private ILogger _logger;
+        private Stack<IDisposable> _disposables;
+        private IHttpContextFactory _httpContextFactory;
```
troydai: Make them readonly

src/Microsoft.AspNet.Server.Kestrel/Server.cs
```diff
@@ -0,0 +1,120 @@
...
+using Microsoft.AspNet.Hosting;
+using Microsoft.AspNet.Hosting.Server;
+using Microsoft.AspNet.Http;
+using Microsoft.AspNet.Http.Features;
+using Microsoft.AspNet.Server.Kestrel.Http;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.AspNet.Server.Kestrel
+{
+    public class Server : IServer
```
troydai: Rename to `KestrelServer`?

src/Microsoft.AspNet.Server.Kestrel/Server.cs
```diff
@@ -0,0 +1,120 @@
...
+
+            _applicationLifetime = applicationLifetime;
+            _logger = logger;
+            Features = features;
+            _httpContextFactory = httpContextFactory;
+        }
+
+        public void Start(RequestDelegate requestDelegate)
+        {
+            _disposables = new Stack<IDisposable>();
```
troydai: A trap here. If the start is invoked twice, the previous list of disposable will be forgotten and cause leak if any.

troydai: Though it rarely happens. How about throw an exception here for that case.

src/Microsoft.AspNet.Server.Kestrel/Server.cs
```diff
@@ -0,0 +1,120 @@
...
+
+                if (information.ThreadCount < 0)
+                {
+                    throw new ArgumentOutOfRangeException(nameof(information.ThreadCount),
+                        information.ThreadCount,
+                        "ThreadCount cannot be negative");
+                }
+
+                engine.Start(information.ThreadCount == 0 ? 1 : information.ThreadCount);
+                bool atLeastOneListener = false;
```
troydai: var

src/Microsoft.AspNet.Server.Kestrel/Server.cs
```diff
@@ -0,0 +1,120 @@
...
+        }
+
+        public void Dispose()
+        {
+            if (_disposables != null)
+            {
+                foreach (var disposable in _disposables)
+                {
+                    disposable.Dispose();
+                }
```
troydai: Clean the _disposables afterwards to prevent objects been disposed twice. It happens when `Start` is finished and `Dispose` is invoked again.

src/Microsoft.AspNet.Server.Kestrel/Server.cs
troydai: Also make it internal.

src/Microsoft.AspNet.Server.Kestrel/Server.cs
halter73: Should we throw if Start is called twice?

src/Microsoft.AspNet.Server.Kestrel/ServiceContext.cs
```diff
@@ -31,6 +33,8 @@ public ServiceContext(ServiceContext context)
 
         public IKestrelTrace Log { get; set; }
 
+        public IHttpContextFactory HttpContextFactory { get; set; }
```
troydai: Does it need to be settable given that it is initialized in constructor?

troydai:
:watch: very straightforward change.

src/Microsoft.AspNet.Server.Kestrel/Http/Frame.cs
```diff
@@ -184,6 +185,8 @@ public Task Stop()
                         }
                         finally
                         {
+                            HttpContextFactory.Dispose(httpContext);
```
halter73: This needs to happen after `FireOnStarting` and `FireOnCompleted`. Both these methods execute user code.

src/Microsoft.AspNet.Server.Kestrel/ServiceContext.cs
JunTaoLuo: Yes we use object initializers. Like in https://github.com/aspnet/KestrelHttpServer/blob/dev/src/Microsoft.AspNet.Server.Kestrel/ServerFactory.cs#L58-L62

troydai: :+1:

JunTaoLuo:
cc @lodejard

src/Microsoft.AspNet.Server.Kestrel/Http/Listener.cs
```diff
@@ -3,6 +3,7 @@
 
 using System;
 using System.Threading.Tasks;
+using Microsoft.AspNet.Builder;
```
davidfowl: https://github.com/aspnet/HttpAbstractions/pull/455

src/Microsoft.AspNet.Server.Kestrel/KestrelServer.cs
```diff
@@ -0,0 +1,129 @@
...
+using Microsoft.AspNet.Hosting;
+using Microsoft.AspNet.Hosting.Server;
+using Microsoft.AspNet.Http;
+using Microsoft.AspNet.Http.Features;
+using Microsoft.AspNet.Server.Kestrel.Http;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.AspNet.Server.Kestrel
+{
+    internal class KestrelServer : IServer
```
davidfowl: Public

src/Microsoft.AspNet.Server.Kestrel/KestrelServer.cs
```diff
@@ -0,0 +1,129 @@
...
+            }
+
+            if (httpContextFactory == null)
+            {
+                throw new ArgumentNullException(nameof(httpContextFactory));
+            }
+
+            _applicationLifetime = applicationLifetime;
+            _logger = logger;
+            Features = features;
```
davidfowl: I think we should create the feature collection in the server itself.

JunTaoLuo:
ping @troydai @halter73 @davidfowl

troydai:
:shipit:

PR #309, opened 2015-10-31
[For Testing] Go faster stripes MkII - Collated changes

benaadams:
Updated  #305

Have now split into 10 PRs:

#310 MemoryPool2 Stack->Queue, Allocate returns newest
#311 Task.Run -> ThreadPool.QueueUserWorkItem
#312 Reduce GetString allocs and conversions
#315 Lockless socket output, lower allocs (4 inter dependent) commits
#316 Remove KestrelThread locks
#317 Connection - remove closure allocation 
#318 Listener - remove closure allocation
#319 ListenerPrimary - remove closure allocation
#320 ListenerSecondary - reduce closure allocation
#321 Reduce delegate allocation

Can't currently performance test as the feeds aren't matching source

PR #310, opened 2015-11-01
MemoryPool2 Stack->Queue, Allocate returns newest

benaadams:
Rather than having a hot end for add **and** removal with a ConcurrentStack;
add to end and remove from top with ConcurrentQueue to reduce contention.

Rather than a while loop on Allocate; return last memory block created
rather than returning it to the pool and checking if one can be removed.

Behaviour remains unchanged (choice of collection is independent of api behaviour)

### DIFF
--- a/src/Microsoft.AspNet.Server.Kestrel/Infrastructure/MemoryPool2.cs
+++ b/src/Microsoft.AspNet.Server.Kestrel/Infrastructure/MemoryPool2.cs
@@ -40,13 +40,13 @@ namespace Microsoft.AspNet.Server.Kestrel.Infrastructure
         /// Thread-safe collection of blocks which are currently in the pool. A slab will pre-allocate all of the block tracking objects
         /// and add them to this collection. When memory is requested it is taken from here first, and when it is returned it is re-added.
         /// </summary>
-        private readonly ConcurrentStack<MemoryPoolBlock2> _blocks = new ConcurrentStack<MemoryPoolBlock2>();
+        private readonly ConcurrentQueue<MemoryPoolBlock2> _blocks = new ConcurrentQueue<MemoryPoolBlock2>();
 
         /// <summary>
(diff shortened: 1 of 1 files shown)


halter73:
I haven't tested this. But intuitively, wouldn't a stack be more likely to reduce cache misses?

I think we should test this change in isolation before merging it.

benaadams:
Am evaluating currently; need to go faster though. Will close the PRs for now until I have some numbers?

rynowak:
Agreed that we should test this change in isolation. When @DamianEdwards and I looked at contention data we didn't see significant amounts of contention on the stack.

benaadams:
The issue is contention hits spinlocks in user mode before moving to kernel locks; under high contention they burn a lot of cpu; don't think spin locks are measured; only when they transition to kernel are they picked up. Have hit 1M rps sustaned https://github.com/aspnet/KestrelHttpServer/pull/309 but latency has increased which means it gets worse at lower connection levels; so closed the PRs till can get to the bottom of it.

rynowak:
Ok cool, if we've actually seen data that shows this is a hotspot then that's different.

rynowak:
> The issue is contention hits spinlocks in user mode before moving to kernel locks; under high contention they burn a lot of cpu; don't think spin locks are measured; only when they transition to kernel are they picked up

For stuff in `System.Collections.Concurrent` - they do ETW tracing for these events which is what we were looking at. Also all of these are lock-free, I don't believe you ever get kernal mode sync.

benaadams:
Just meant all the changes holistically together; this being one of - but latency goes from 6ms to 16ms; so the changes are untenable; which is why I've closed the PRs. Can always add more bandwidth, more cpu, more servers; but can't change the speed of light, so increasing latency is an awful outcome; and latency is a tcp killer.

My hope is that I can up the throughput *and* bring the latency back down; which should speed up the server so that these changes will be meaningful enough to be significant (good or bad) - at the moment they are just fiddling at the edges :-/

benaadams:
Raised PR #330 just for the allocate

PR #311, opened 2015-11-01
Task.Run -> ThreadPool.QueueUserWorkItem

benaadams:
Task.Run eventually ends up being QueueUserWorkItem.
The returned task is ignored, so no added goodness.
Short running item.

Cut out the middleman

### DIFF
--- a/src/Microsoft.AspNet.Server.Kestrel/Http/SocketInput.cs
+++ b/src/Microsoft.AspNet.Server.Kestrel/Http/SocketInput.cs
@@ -128,7 +128,7 @@ namespace Microsoft.AspNet.Server.Kestrel.Http
             if (awaitableState != _awaitableIsCompleted &&
                 awaitableState != _awaitableIsNotCompleted)
             {
-                Task.Run(awaitableState);
+                ThreadPool.QueueUserWorkItem((o) => ((Action)o)(), awaitableState);
             }
         }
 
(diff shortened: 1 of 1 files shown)


src/Microsoft.AspNet.Server.Kestrel/Http/SocketInput.cs
```diff
@@ -194,7 +190,7 @@ public void OnCompleted(Action continuation)
             }
             else if (awaitableState == _awaitableIsCompleted)
             {
-                Task.Run(continuation);
+                ThreadPool.QueueUserWorkItem((o) => continuation());
```
rynowak: Should this become an `Action<object>`? Isn't this going to allocate a lambda?

benaadams: Gah, yeah. Changed

benaadams:

benaadams:
There is a flag switch that differs between the two paths (`compressedStack`); will need to evaluate more.

src/Microsoft.AspNet.Server.Kestrel/Http/SocketInput.cs
```diff
@@ -97,11 +97,7 @@ public void IncomingComplete(int count, Exception error)
                     {
                         _head = _tail = _pinned;
                     }
-                    else if (_tail == _pinned)
```
davidfowl: avoid the unrelated changes in PR, they're pretty distracting.

PR #312, opened 2015-11-01
Reduce GetString allocs and conversions

### LINKED ISSUE #291: [Perf] High Allocator MemoryPoolIterator2.GetString
Drill down on Issue #288
Measurements post PR #287

Item 1 on highest allocators in plaintext test

most-memory

Allocations (2003 requests)

alloc

Bytes (2003 requests)

bytes

Also the HTTP portion of the request is ASCII bytes rather than UTF8 or other encoding https://tools.ietf.org/html/rfc7230#section-3.2.4

Note: Path can be decoded to utf8 before being converted to string so may be utf8

benaadams:
Resolves  #291

### DIFF
--- a/src/Microsoft.AspNet.Server.Kestrel/Http/Frame.cs
+++ b/src/Microsoft.AspNet.Server.Kestrel/Http/Frame.cs
@@ -629,7 +629,7 @@ namespace Microsoft.AspNet.Server.Kestrel.Http
                 {
                     return false;
                 }
-                var method = begin.GetString(scan);
+                var method = begin.GetAsciiString(scan);
 
                 scan.Take();
                 begin = scan;
@@ -653,7 +653,7 @@ namespace Microsoft.AspNet.Server.Kestrel.Http
                     {
                         return false;
                     }
-                    queryString = begin.GetString(scan);
+                    queryString = begin.GetAsciiString(scan);
                 }
 
                 scan.Take();
@@ -662,7 +662,7 @@ namespace Microsoft.AspNet.Server.Kestrel.Http
                 {
                     return false;
                 }
-                var httpVersion = begin.GetString(scan);
+                var httpVersion = begin.GetAsciiString(scan);
 
                 scan.Take();
                 if (scan.Take() != '\n')
@@ -670,12 +670,21 @@ namespace Microsoft.AspNet.Server.Kestrel.Http
                     return false;
                 }
 
+                // URIs are always encoded/escaped to ASCII https://tools.ietf.org/html/rfc3986#page-11 
+                // Multibyte Internationalized Resource Identifiers (IRIs) are first converted to utf8; 
+                // then encoded/escaped to ASCII  https://www.ietf.org/rfc/rfc3987.txt "Mapping of IRIs to URIs"
+                string requestUrlPath;
                 if (needDecode)
                 {
+                    // URI was encoded, unescape and then parse as utf8
                     pathEnd = UrlPathDecoder.Unescape(pathBegin, pathEnd);
+                    requestUrlPath = pathBegin.GetUtf8String(pathEnd);
                 }
-
-                var requestUrlPath = pathBegin.GetString(pathEnd);
+                else
+                {
+                    // URI wasn't encoded, parse as ASCII
+                    requestUrlPath = pathBegin.GetAsciiString(pathEnd);
+                }
 
                 consumed = scan;
                 Method = method;
@@ -689,11 +698,6 @@ namespace Microsoft.AspNet.Server.Kestrel.Http
             {
                 input.ConsumingComplete(consumed, scan);
             }
-        }
-
-        static string GetString(ArraySegment<byte> range, int startIndex, int endIndex)
-        {
-            return Encoding.UTF8.GetString(range.Array, range.Offset + startIndex, endIndex - startIndex);
         }
 
         public static bool TakeMessageHeaders(SocketInput input, FrameRequestHeaders requestHeaders)
@@ -787,7 +791,7 @@ namespace Microsoft.AspNet.Server.Kestrel.Http
                         }
 
                         var name = beginName.GetArraySegment(endName);
-                        var value = beginValue.GetString(endValue);
+                        var value = beginValue.GetAsciiString(endValue);
                         if (wrapping)
                         {
                             value = value.Replace("\r\n", " ");
--- a/src/Microsoft.AspNet.Server.Kestrel/Infrastructure/MemoryPoolIterator2.cs
+++ b/src/Microsoft.AspNet.Server.Kestrel/Infrastructure/MemoryPoolIterator2.cs
@@ -4,7 +4,6 @@
 using System;
 using System.Linq;
 using System.Numerics;
-using System.Text;
 
 namespace Microsoft.AspNet.Server.Kestrel.Infrastructure
 {
@@ -21,8 +20,6 @@ namespace Microsoft.AspNet.Server.Kestrel.Infrastructure
         /// Used as an argument in the vector dot product that determines matching character index.
         /// </summary>
         private static Vector<byte> _dotIndex = new Vector<byte>(Enumerable.Range(0, Vector<byte>.Count).Select(x => (byte)-x).ToArray());
-
-        private static Encoding _utf8 = Encoding.UTF8;
 
         private MemoryPoolBlock2 _block;
         private int _index;
@@ -486,101 +483,6 @@ namespace Microsoft.AspNet.Server.Kestrel.Infrastructure
                     index = block.Start;
                 }
             }
-        }
-
-        public string GetString(MemoryPoolIterator2 end)
-        {
-            if (IsDefault || end.IsDefault)
-            {
-                return default(string);
-            }
-            if (end._block == _block)
-            {
-                return _utf8.GetString(_block.Array, _index, end._index - _index);
-            }
-
-            var decoder = _utf8.GetDecoder();
-
-            var length = GetLength(end);
-            var charLength = length * 2;
-            var chars = new char[charLength];
-            var charIndex = 0;
-
-            var block = _block;
-            var index = _index;
-            var remaining = length;
-            while (true)
-            {
-                int bytesUsed;
-                int charsUsed;
-                bool completed;
-                var following = block.End - index;
-                if (remaining <= following)
-                {
-                    decoder.Convert(
-                        block.Array,
-                        index,
-                        remaining,
-                        chars,
-                        charIndex,
-                        charLength - charIndex,
-                        true,
-                        out bytesUsed,
-                        out charsUsed,
-                        out completed);
-                    return new string(chars, 0, charIndex + charsUsed);
-                }
-                else if (block.Next == null)
-                {
-                    decoder.Convert(
-                        block.Array,
-                        index,
-                        following,
-                        chars,
-                        charIndex,
-                        charLength - charIndex,
-                        true,
-                        out bytesUsed,
-                        out charsUsed,
-                        out completed);
-                    return new string(chars, 0, charIndex + charsUsed);
-                }
-                else
-                {
-                    decoder.Convert(
-                        block.Array,
-                        index,
-                        following,
-                        chars,
-                        charIndex,
-                        charLength - charIndex,
-                        false,
-                        out bytesUsed,
-                        out charsUsed,
-                        out completed);
-                    charIndex += charsUsed;
-                    remaining -= following;
-                    block = block.Next;
-                    index = block.Start;
-                }
-            }
-        }
-
-        public ArraySegment<byte> GetArraySegment(MemoryPoolIterator2 end)
-        {
-            if (IsDefault || end.IsDefault)
-            {
-                return default(ArraySegment<byte>);
-            }
-            if (end._block == _block)
-            {
-                return new ArraySegment<byte>(_block.Array, _index, end._index - _index);
-            }
-
-            var length = GetLength(end);
-            var array = new byte[length];
-            CopyTo(array, 0, length, out length);
-            return new ArraySegment<byte>(array, 0, length);
         }
 
         public MemoryPoolIterator2 CopyTo(byte[] array, int offset, int count, out int actual)
--- a/src/Microsoft.AspNet.Server.Kestrel/Infrastructure/MemoryPoolIterator2Extenstions.cs
+++ b/src/Microsoft.AspNet.Server.Kestrel/Infrastructure/MemoryPoolIterator2Extenstions.cs
@@ -1,0 +1,207 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Text;
+
+namespace Microsoft.AspNet.Server.Kestrel.Infrastructure
+{
+    public static class MemoryPoolIterator2Extenstions
+    {
+        private const int _maxStackAllocBytes = 16384;
+
+        private static Encoding _utf8 = Encoding.UTF8;
+
+        private static unsafe string GetAsciiStringStack(byte[] input, int inputOffset, int length)
+        {
+            // avoid declaring other local vars, or doing work with stackalloc
+            // to prevent the .locals init cil flag , see: https://github.com/dotnet/coreclr/issues/1279
+            char* output = stackalloc char[length];
+
+            return GetAsciiStringImplementation(output, input, inputOffset, length);
+        }
+        private static unsafe string GetAsciiStringImplementation(char* output, byte[] input, int inputOffset, int length)
+        {
+            for (var i = 0; i < length; i++)
+            {
+                output[i] = (char)input[inputOffset + i];
+            }
+
+            return new string(output, 0, length);
+        }
+
+        private static unsafe string GetAsciiStringStack(MemoryPoolBlock2 start, MemoryPoolIterator2 end, int inputOffset, int length)
+        {
+            // avoid declaring other local vars, or doing work with stackalloc
+            // to prevent the .locals init cil flag , see: https://github.com/dotnet/coreclr/issues/1279
+            char* output = stackalloc char[length];
+
+            return GetAsciiStringImplementation(output, start, end, inputOffset, length);
+        }
+
+        private unsafe static string GetAsciiStringHeap(MemoryPoolBlock2 start, MemoryPoolIterator2 end, int inputOffset, int length)
+        {
+            var buffer = new char[length];
+
+            fixed (char* output = buffer)
+            {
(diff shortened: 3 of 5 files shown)


src/Microsoft.AspNet.Server.Kestrel/Infrastructure/MemoryPoolIterator2.cs
```diff
@@ -488,7 +488,76 @@ public int GetLength(MemoryPoolIterator2 end)
             }
         }
 
-        public string GetString(MemoryPoolIterator2 end)
+        private static unsafe string SingleBlockAsciiString(byte[] input, int offset, int length)
+        {
+            // avoid declaring other local vars, or doing work with stackalloc
+            // to prevent the .locals init cil flag , see: https://github.com/dotnet/coreclr/issues/1279
+            char* output = stackalloc char[length];
```
davidfowl: How big can length be?

benaadams: less than 1MB; call it 512k to be safe. Headers shouldn't be larger than 8k; probably should add a check though?

benaadams: Largest header will be cookie and issues with it being > 4093 bytes http://stackoverflow.com/questions/640938/what-is-the-maximum-size-of-a-web-browsers-cookies-key

benaadams: Similar issues with url length: http://stackoverflow.com/questions/417142/what-is-the-maximum-length-of-a-url-in-different-browsers

benaadams: Added 16k max

benaadams:
Bad PR

benaadams:

src/Microsoft.AspNet.Server.Kestrel/Infrastructure/MemoryPoolIterator2.cs
```diff
@@ -10,6 +10,7 @@ namespace Microsoft.AspNet.Server.Kestrel.Infrastructure
 {
     public struct MemoryPoolIterator2
     {
+        private const int _maxHeaderStackLength = 16384;
```
halter73: Is there a specific reason we don't want to stackalloc a char array larger than this?

src/Microsoft.AspNet.Server.Kestrel/Infrastructure/MemoryPoolIterator2.cs
```diff
@@ -488,20 +489,108 @@ public int GetLength(MemoryPoolIterator2 end)
...
+            // to prevent the .locals init cil flag , see: https://github.com/dotnet/coreclr/issues/1279
+            char* output = stackalloc char[length];
+
+            return MultiBlockAsciiIter(output, startBlock, ref end, offset, length);
+        }
+        
+        private static unsafe string MultiBlockAsciiIter(char* output, MemoryPoolBlock2 startBlock, ref MemoryPoolIterator2 end, int inputOffset, int length)
+        {
+            // avoid declaring other local vars, or doing work with stackalloc
+            // to prevent the .locals init cil flag , see: https://github.com/dotnet/coreclr/issues/1279
```
halter73: Remove this comment from methods that don't stackalloc.

src/Microsoft.AspNet.Server.Kestrel/Infrastructure/MemoryPoolIterator2.cs
```diff
@@ -488,20 +489,108 @@ public int GetLength(MemoryPoolIterator2 end)
...
             if (IsDefault || end.IsDefault)
             {
                 return default(string);
             }
+
+            var length = GetLength(ref end);
+
+            if (length > _maxHeaderStackLength)
+            {
+                return GetUtf8String(ref end);
```
halter73: Seems weird to call `GetUtf8String` from `GetAsciiString`. Maybe we should call `GetUtf8String` `GetAsciiStringSlow` or something and use `Encoding.ASCII`.

src/Microsoft.AspNet.Server.Kestrel/Infrastructure/MemoryPoolIterator2.cs
benaadams: Total headers are generally constrained < 8k/16k So individual headers shouldn't be larger than 16k; if you are going larger than that you probably have other issues to resolve than performance - Added some related comments to: https://github.com/aspnet/KestrelHttpServer/pull/313/files#diff-6b7953b837b6ad7c921259c907520639R14

Can go larger; but that was my reasoning.

For upper levels; total stack size aspnet4 x86 was 256k, x64 was 512k; .net 1MB - not sure what the current defaults are.

src/Microsoft.AspNet.Server.Kestrel/Infrastructure/MemoryPoolIterator2.cs
benaadams: Added `GetAsciiStringHeap` method?

src/Microsoft.AspNet.Server.Kestrel/Infrastructure/MemoryPoolIterator2.cs
benaadams: For ref: 
nginx [default 8k]( http://nginx.org/en/docs/http/ngx_http_core_module.html#large_client_header_buffers)
Apache [default 8190 bytes](http://httpd.apache.org/docs/2.4/mod/core.html#limitrequestfieldsize)
IIS default 16384 bytes

benaadams:
Mono keeps jamming ;-/

benaadams:
Was rebased on #337 to reduce probability of CI failures; removing last commit as don't think it necesserily helped anything

src/Microsoft.AspNet.Server.Kestrel/Infrastructure/MemoryPoolIterator2.cs
NickCraver: To add even more reasoning here: HTTP/2 compression will further reduce the average header size. I agree 16k is a good limit given what we see out in the world.

halter73:
This looks good to me, but since this is a pretty big change I would like someone else to also sign off on it.

@lodejard @davidfowl  @troydai thoughts?

davidfowl:
Is there anyway we can take advantage of CopyTo here? We already have the memory pool interator

benaadams:
@davidfowl you mean move into  MemoryPoolIterator2 rather than have it in Frame? GetNextString kinda thing?

troydai:
It looks good to me. The change is straightforward. One thing we __may or may not__ to do is to extract the logics of _taking two iterators to generate a string_ to a separate class. The logic here expand the iterator class which in my opinion should be kept concise. The logic of creating string with two iterators is very much self-contained.

src/Microsoft.AspNet.Server.Kestrel/Infrastructure/MemoryPoolIterator2.cs
davidfowl: This value is in the wrong layer MemoryPoolIterator2 doesn't understand what it's looking at. That's the caller's job.

benaadams: Its to prevent the stack blowing up; so if length is larger than this it heap allocates instead. Maybe a rename like `_maxStackAllocBytes`?

src/Microsoft.AspNet.Server.Kestrel/Infrastructure/MemoryPoolIterator2.cs
benaadams: Is just a limit on the max it will attempt with `stackalloc` before using `new` and heap allocating instead; isn't actually a parsing limit.

benaadams:
Renamed constant

benaadams:
I think this can be cleaned up

benaadams:
Simpler?

benaadams:
> Is there anyway we can take advantage of CopyTo here?

Its the byte -> char conversion that needs to be explicit to get it in right format for string; CopyTo is byte-> byte

> One thing we may or may not to do is to extract the logics of taking two iterators to generate a string to a separate class.

@troydai  Changed the string from MemoryPoolIterator to Extenstion methods; better?

benaadams:
For the Utf8->char*(string) it can be done faster with intrinsics but I don't think Vector exposes all the needed methods (see http://woboq.com/blog/utf-8-processing-using-simd.html)

src/Microsoft.AspNet.Server.Kestrel/Infrastructure/MemoryPoolIterator2Extenstions.cs
```diff
@@ -0,0 +1,207 @@
...
+                {
+                    return new string(output, 0, length);
+                }
+
+                block = block.Next;
+                inputOffset = block.Start;
+            }
+        }
+
+        private static string GetAsciiStringHeap(MemoryPoolBlock2 start, MemoryPoolIterator2 end, int inputOffset, int length)
```
davidfowl: In the name of code reuse:

```C#
private static string GetAsciiStringHeap(MemoryPoolBlock2 start, MemoryPoolIterator2 end, int inputOffset, int length)
{
    var buffer = new char[length];
    fixed (char* output = buffer)
    {
        return MultiBlockAsciiIter(output, start, end, inputOffset, length);
    }
}

private static unsafe string MultiBlockAsciiIter(char* output, MemoryPoolBlock2 start, MemoryPoolIterator2 end, int inputOffset, int length)
{
  ...
}
```

benaadams: Much better, done

troydai:
:shipit:

src/Microsoft.AspNet.Server.Kestrel/Http/Frame.cs
```diff
@@ -670,12 +670,16 @@ private bool TakeStartLine(SocketInput input)
...
 
+                string requestUrlPath;
                 if (needDecode)
                 {
                     pathEnd = UrlPathDecoder.Unescape(pathBegin, pathEnd);
+                    requestUrlPath = pathBegin.GetUtf8String(pathEnd);
+                }
+                else
+                {
+                    requestUrlPath = pathBegin.GetAsciiString(pathEnd);
```
davidfowl: Always use Utf8 for the path.

src/Microsoft.AspNet.Server.Kestrel/Http/Frame.cs
```diff
@@ -675,7 +675,7 @@ private bool TakeStartLine(SocketInput input)
                     pathEnd = UrlPathDecoder.Unescape(pathBegin, pathEnd);
                 }
 
-                var requestUrlPath = pathBegin.GetString(pathEnd);
+                var requestUrlPath = pathBegin.GetUtf8String(pathEnd);
```
Tratcher: GetAsciiString

Tratcher: Wait, let me read this closer.

benaadams: @davidfowl just told me to always Utf8 the path

davidfowl: Lol @benaadams wires crossed

src/Microsoft.AspNet.Server.Kestrel/Infrastructure/MemoryPoolIterator2Extenstions.cs
```diff
@@ -0,0 +1,186 @@
...
+            while (true)
+            {
+                int following = (block != endBlock ? block.End : endIndex) - inputOffset;
+
+                if (following > 0)
+                {
+                    var input = block.Array;
+                    for (var i = 0; i < following; i++)
+                    {
+                        output[i + outputOffset] = (char)input[i + inputOffset];
```
Tratcher: The real question is how do you detect and report when this is assumption wrong? E.g. when the byte > 7F (utf-8). Otherwise the app just sees corrupt bytes in their fields.

benaadams: That's because it is corrupt data

https://tools.ietf.org/html/rfc7230#section-3.2.4

> Historically, HTTP has allowed field content with text in the
ISO-8859-1 charset [ISO-8859-1], supporting other charsets only
through use of [RFC2047] encoding. In practice, most HTTP header
field values use only a subset of the US-ASCII charset [USASCII].
Newly defined header fields SHOULD limit their field values to
US-ASCII octets. A recipient SHOULD treat other octets in field
content (obs-text) as opaque data.

benaadams: Or rather "opaque octets" to go with the lingo...

Tratcher: Sure, but how do you report that it is corrupt data?

src/Microsoft.AspNet.Server.Kestrel/Http/Frame.cs
Tratcher: It is utf8 because line 675 is already doing the first level hex un-escape.

benaadams: Wasn't in previous commit: https://github.com/benaadams/KestrelHttpServer/commit/6769e1e042691475e1e69abc9dc4077c10c6d9d2

Tratcher: Oooh, https://github.com/benaadams/KestrelHttpServer/commit/6769e1e042691475e1e69abc9dc4077c10c6d9d2 was better. Sorry about that.

src/Microsoft.AspNet.Server.Kestrel/Infrastructure/MemoryPoolIterator2Extenstions.cs
benaadams: > A recipient SHOULD treat other octets in field content (obs-text) as opaque data.

Says ignore it to me? Let it be read as the underlying byte value?

src/Microsoft.AspNet.Server.Kestrel/Http/Frame.cs
benaadams: @davidfowl revert?

src/Microsoft.AspNet.Server.Kestrel/Infrastructure/MemoryPoolIterator2Extenstions.cs
Tratcher: We don't have a good mechanism for treating it as `opaque data` because we only expose strings and you have to convert to a string somehow. If there was a cheap way to detect bad data I'd prefer to do a verbose log before we mangled it.

benaadams: But dotnet strings aren't C strings; they can handle it :)

```
string s = new string(new char[] { (char)0, (char)2, (char)3 });
if ((byte)s[0] == 0)
{
    // yes its true
}
```
So if I use a bunch of binary headers its cool; as long as I stay away from `\r\n' and whatever the wrap thing is

benaadams: Whereas converting to utf8 you kill everything >= 128 so you have mangled it

src/Microsoft.AspNet.Server.Kestrel/Http/Frame.cs
benaadams: Question is are paths always url encoded?

benaadams: https://tools.ietf.org/html/rfc3986#page-11 seems to suggest so; but looking for a more recent one...

benaadams: Internationalized Resource Identifiers (IRIs) https://www.ietf.org/rfc/rfc3987.txt Mapping of IRIs to URIs; says yes? Convert to utf8 then url-encode to ascii - so reverting would be a safe transformation?

src/Microsoft.AspNet.Server.Kestrel/Infrastructure/MemoryPoolIterator2Extenstions.cs
benaadams: Also don't pass into any interop C string processors or they may read it truncated at they will use the zero as a terminator; but that format of strings is bad for many many reasons... and interpreting it truncated is no so bad as reading past the end.

davidfowl: We should file an issue for this AFAIK ASCIIEncoding turns things into ? /cc @halter73

src/Microsoft.AspNet.Server.Kestrel/Http/Frame.cs
benaadams: Reverted

src/Microsoft.AspNet.Server.Kestrel/Infrastructure/MemoryPoolIterator2Extenstions.cs
benaadams: Gone back to previous implementation of not using Encoding.ASCII but just casting the bytes to (char) which seems safer for the `opaque data`?

benaadams: Rechecking my assumptions about 255...

benaadams:
Adding some tests for this

src/Microsoft.AspNet.Server.Kestrel/Infrastructure/MemoryPoolIterator2Extenstions.cs
benaadams: bytes 0 and 255 are ok in string; added tests for the 3 paths

src/Microsoft.AspNet.Server.Kestrel/Http/Frame.cs
benaadams: Changed and reverted as discussed in outdated diff

Issue #181, opened 2015-09-01
Connection not terminated for invalid/ill-formed requests


moozzyk:
If a client sends an incorrect request the connection should be terminated. What happens now is that the connection is not terminated and new data are appended to the buffer but the data from the buffer are never removed so the buffer grows. Kestrel does throw if the data is incorrect (https://github.com/aspnet/KestrelHttpServer/blob/dev/src/Microsoft.AspNet.Server.Kestrel/Http/Frame.cs#L560)  but the exception is then caught logged and swallowed (https://github.com/aspnet/KestrelHttpServer/blob/dev/src/Microsoft.AspNet.Server.Kestrel/Http/Connection.cs#L87)

Repro steps:
- With telnet send \r\n to the server

Expected behavior:
- the connection is closed

Actual behavior:
- you can continue typing. with each new character an exception will be thrown on the server side. buffer grows. the connection is never closed

Should we do some fuzzing tests in general?

Tragetaschen:
Generally, any malformed request currently throws off Kestrel in various ways. I had created a [couple of failing tests](https://github.com/Tragetaschen/KestrelHttpServer/blob/pinvoke-cleanup-beta6/test/Microsoft.AspNet.Server.KestrelTests/Issues.cs) where generally a connection would leak, but I guess no one looked at it.

BTW, is there any way to test for leaking connections? @DamianEdwards mentioned these in various community standups and I guess they have been fixed, but there doesn't seem to be infrastructure to actually verify it.

halter73:
[This commit](https://github.com/aspnet/KestrelHttpServer/commit/094b8efbf8b760fb3929e821e8eadc5f28d06727) which just made it into RC1 allows Kestrel to handle requests with empty headers.

We still should add fuzz testing and limit start line length, header length, etc...

DamianEdwards:
This is a potential stress issue so I'd like us to take a look at this for RC2.

benaadams:
The last commit of https://github.com/aspnet/KestrelHttpServer/pull/619 should resolve this

benaadams:
Though not all @Tragetaschen's scenarios

cesarbs:
Fixed in https://github.com/aspnet/KestrelHttpServer/issues/683.

cesarbs:
Actually this is a dup of https://github.com/aspnet/KestrelHttpServer/issues/683.

PR #313, opened 2015-11-01
[Wip] Don't process infinite request lengths

benaadams:
Would cause unbounded memory growth.

Can I haz prize?

### DIFF
--- a/src/Microsoft.AspNet.Server.Kestrel/Infrastructure/MemoryPoolIterator2.cs
+++ b/src/Microsoft.AspNet.Server.Kestrel/Infrastructure/MemoryPoolIterator2.cs
@@ -11,15 +11,6 @@ namespace Microsoft.AspNet.Server.Kestrel.Infrastructure
 {
     public struct MemoryPoolIterator2
     {
-        // TODO: should be configurable? Default .NET 4.5 lengths are, configurable upwards:
-        // maxUrlLength = 260 (windows path length?)
-        // maxQueryStringLength = 2048
-        // Needs cooperation of reverse proxy if used
-        // In wild max cookie size ~4093 bytes http://stackoverflow.com/questions/640938/what-is-the-maximum-size-of-a-web-browsers-cookies-key
-        // In wild max url length ~2000 bytes http://stackoverflow.com/questions/417142/what-is-the-maximum-length-of-a-url-in-different-browsers
-        // Security issues around unbounded limits
-        private const int _maxHeaderLength = 16384;
-
         /// <summary>
         /// Array of "minus one" bytes of the length of SIMD operations on the current hardware. Used as an argument in the
         /// vector dot product that counts matching character occurrence.
--- a/src/Microsoft.AspNet.Server.Kestrel/Infrastructure/MemoryPoolIterator2.cs
+++ b/src/Microsoft.AspNet.Server.Kestrel/Infrastructure/MemoryPoolIterator2.cs
@@ -2,6 +2,7 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.IO;
 using System.Linq;
 using System.Numerics;
 using System.Text;
@@ -10,6 +11,15 @@ namespace Microsoft.AspNet.Server.Kestrel.Infrastructure
 {
     public struct MemoryPoolIterator2
     {
+        // TODO: should be configurable? Default .NET 4.5 lengths are, configurable upwards:
+        // maxUrlLength = 260 (windows path length?)
+        // maxQueryStringLength = 2048
+        // Needs cooperation of reverse proxy if used
+        // In wild max cookie size ~4093 bytes http://stackoverflow.com/questions/640938/what-is-the-maximum-size-of-a-web-browsers-cookies-key
+        // In wild max url length ~2000 bytes http://stackoverflow.com/questions/417142/what-is-the-maximum-length-of-a-url-in-different-browsers
+        // Security issues around unbounded limits
+        private const int _maxHeaderLength = 16384;
+
         /// <summary>
         /// Array of "minus one" bytes of the length of SIMD operations on the current hardware. Used as an argument in the
         /// vector dot product that counts matching character occurrence.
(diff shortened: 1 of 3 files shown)


Tratcher:
RE: https://github.com/aspnet/KestrelHttpServer/issues/181#issuecomment-151241986

benaadams:
@Tratcher could you provide more context? Related, covered by, clashing etc 

Or just my prize comment :) Which I think is also invalidated by creating a public PR

Tratcher:
Just referencing our existing notes on the topic.

Tratcher:
@blowdart

blowdart:
@benaadams In case your question was serious if you'd have reported it though the bug bounty rather than a github issue and we hadn't already known then it would have qualified.

benaadams:
@blowdart yeah, wasn't really serious and knew would invalidate but creating a public patch. 

Serious follow up though; security issues post live (rc2?) would you like them addressed privately i.e. don't post a patch for it to public repo?

blowdart:
@benaadams Yea, once we get past RC1 escrow I'd much rather they were reported directly. And that way they'd be bounty eligible if you report through MSRC ;)

src/Microsoft.AspNet.Server.Kestrel/Infrastructure/MemoryPoolIterator2.cs
```diff
@@ -10,6 +11,15 @@ namespace Microsoft.AspNet.Server.Kestrel.Infrastructure
...
     public struct MemoryPoolIterator2
     {
+        // TODO: should be configurable? Default .NET 4.5 lengths are, configurable upwards:
+        // maxUrlLength = 260 (windows path length?)
+        // maxQueryStringLength = 2048
+        // Needs cooperation of reverse proxy if used
+        // In wild max cookie size ~4093 bytes http://stackoverflow.com/questions/640938/what-is-the-maximum-size-of-a-web-browsers-cookies-key
+        // In wild max url length ~2000 bytes http://stackoverflow.com/questions/417142/what-is-the-maximum-length-of-a-url-in-different-browsers
+        // Security issues around unbounded limits
+        private const int _maxHeaderLength = 16384;
```
davidfowl: Any limits like this need to come from configuration with reasonable defaults

NickCraver: +1 - Though we would use this externally with no problems, there are some internal...optimizations...that involve large headers for us. 16k would break on them.

benaadams: Need to work out how to do that exactly... probably shouldn't be too long.

benaadams:
Still need to plumb in the values; but making sure I'm reading from config in the right way/place etc

benaadams:
Currently this isn't a show stopper as in the recommended hosting models (fronted by IIS or ngnix) they would terminate these requests at their configured limits.

src/Microsoft.AspNet.Server.Kestrel/KestrelServerInformation.cs
```diff
@@ -17,6 +17,10 @@ public class KestrelServerInformation : IKestrelServerInformation, IServerAddres
 
         public bool NoDelay { get; set;  }
 
+        public int MaxHeaderBytes { get; set; } = 16384; // 16kB
```
Tratcher: Something is out of sync with this PR. This property is read from config but the server never checks it.

src/Microsoft.AspNet.Server.Kestrel/Infrastructure/MemoryPoolIterator2.cs
```diff
@@ -170,6 +174,11 @@ public int Seek(int char0)
                 }
                 while (block.End != index)
                 {
+                    if (_processed > _maxHeaderLength)
```
Tratcher: The definition of this variable is missing?

src/Microsoft.AspNet.Server.Kestrel/KestrelServerInformation.cs
benaadams: PR is incomplete; just making sure I'm reading correctly?

Tratcher: Your config approach looks reasonable.

src/Microsoft.AspNet.Server.Kestrel/KestrelServerInformation.cs
```diff
@@ -26,6 +30,36 @@ public void Initialize(IConfiguration configuration)
             {
                 Addresses.Add(url);
             }
+
+            var maxHeaderBytes = configuration["request.maxHeaderBytes"];
```
Tratcher: `request.` => `server.`

src/Microsoft.AspNet.Server.Kestrel/Infrastructure/MemoryPoolIterator2.cs
benaadams: trying to work out how to get it there :-/

benaadams: Any suggestions? Always pass it to constructor is what I'm looking at currently... Hoping there is a better way.

Tratcher: @halter73 ?

halter73: We could put these limits directly on ServiceContext like we do for ConnectionFilter. It might be best to put the entire IKestrelServerInformation object in ServiceContext.

benaadams:
Major issue is its not exact but in Vector.Size increments is that a problem?

src/Microsoft.AspNet.Server.Kestrel/Infrastructure/MemoryPoolIterator2.cs
halter73: I also think that the limits need to be passed as arguments to functions like Seek. Just in case it is ever used from something other than headers.

halter73:
On my machine, `Vector<byte>.Count` is 16. I know it could be larger on other machines, but I don't think we change our Seek algorithm to scan in smaller strides just to be able to enforce more precise limits.

src/Microsoft.AspNet.Server.Kestrel/Infrastructure/MemoryPoolIterator2.cs
benaadams: return a `ResultType { int count, bool limitReached }` object and add a `limit` into each function and the caller can decide what to do? (e.g. throw BadRequest)

halter73: Sounds good. Another option might be to throw directly and let the caller wrap the exception.

benaadams:
Been struggling what to do with `MaxUploadBytes` but that can be a filter; sorting this out now.

benaadams:
I'm closing this, and will raise issue instead as I ended up accidentally deleting my repo with the changes in :(

PR #314, opened 2015-11-01
Stop streams on finish

### LINKED ISSUE #263: Poison streams after the app func completes
Any attempt to write to the response body or read from the request body after the app func completes should fail with an `ObjectDisposedException`.

benaadams:
Can't use dispose (or close) as can be disposed too early by user code

Resolves #263

### DIFF
--- a/src/Microsoft.AspNet.Server.Kestrel/Http/FrameRequestStream.cs
+++ b/src/Microsoft.AspNet.Server.Kestrel/Http/FrameRequestStream.cs
@@ -53,7 +53,7 @@ namespace Microsoft.AspNet.Server.Kestrel.Http
         {
             if (_stopped)
             {
-                throw new ObjectDisposedException("RequestStream has been disposed");
+                throw new ObjectDisposedException(nameof(FrameRequestStream));
             }
 
             return ReadAsync(buffer, offset, count).Result;
@@ -64,7 +64,7 @@ namespace Microsoft.AspNet.Server.Kestrel.Http
         {
             if (_stopped)
             {
-                throw new ObjectDisposedException("RequestStream has been disposed");
+                throw new ObjectDisposedException(nameof(FrameRequestStream));
             }
 
             var task = ReadAsync(buffer, offset, count, CancellationToken.None, state);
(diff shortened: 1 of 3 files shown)


src/Microsoft.AspNet.Server.Kestrel/Http/FrameRequestStream.cs
```diff
@@ -102,5 +123,10 @@ public override void Write(byte[] buffer, int offset, int count)
         {
             throw new NotImplementedException();
         }
+
+        public void StopAcceptingReads()
```
Tratcher: It's worth including your comment here about why you can't use Dispose.

benaadams: Added and referenced example of this

src/Microsoft.AspNet.Server.Kestrel/Http/FrameRequestStream.cs
```diff
@@ -50,12 +51,22 @@ public override void SetLength(long value)
 
         public override int Read(byte[] buffer, int offset, int count)
         {
+            if (_stopped)
+            {
+                throw new ObjectDisposedException("RequestStream has been disposed");
```
halter73: This ctor expects the object name instead of an exception message.

benaadams: Corrected

benaadams:
Bumped for CI retest rather than doing a whitespace change and force push as branch https://github.com/aspnet/KestrelHttpServer/tree/shalter/request-lifetime has built on it.

NickCraver:
+1

halter73:
I went against the normal pattern of merging this and rebased instead. It was more convenient since I was working against these changes in my request-lifetime PR.

Here is the last of the 3 relevant commits: https://github.com/aspnet/KestrelHttpServer/commit/1cbb273b0d6d752ed557a78de38a5b83a8223d0e

benaadams:
Yar; you stole my rebase; force push ability ;-)

PR #315, opened 2015-11-01
Lockless socket output, lower allocs

benaadams:
Inter dependent items from PR #309 as individual commits


benaadams:
Closing for now until have some numbers; need to improve latency...

PR #316, opened 2015-11-01
Remove KestrelThread locks

benaadams:
Remove KestrelThread locks, high contention locks are expensive, low contention cheap. ConcurrentQueue in producer/consumer contention as add and remove are seperated


benaadams:

PR #317, opened 2015-11-01
Connection - remove closure allocation

benaadams:
No closure or delegate allocation now

### DIFF
--- a/src/Microsoft.AspNet.Server.Kestrel/Http/Connection.cs
+++ b/src/Microsoft.AspNet.Server.Kestrel/Http/Connection.cs
@@ -13,8 +13,10 @@ namespace Microsoft.AspNet.Server.Kestrel.Http
 {
     public class Connection : ConnectionContext, IConnectionControl
     {
-        private static readonly Action<UvStreamHandle, int, object> _readCallback = ReadCallback;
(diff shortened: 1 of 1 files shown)


src/Microsoft.AspNet.Server.Kestrel/Http/Connection.cs
```diff
@@ -13,8 +13,10 @@ namespace Microsoft.AspNet.Server.Kestrel.Http
 {
     public class Connection : ConnectionContext, IConnectionControl
     {
-        private static readonly Action<UvStreamHandle, int, object> _readCallback = ReadCallback;
-        private static readonly Func<UvStreamHandle, int, object, Libuv.uv_buf_t> _allocCallback = AllocCallback;
+        private static readonly Action<UvStreamHandle, int, object> _readCallback = 
+            (handle, status, state) => ReadCallback(handle, status, state);
```
halter73: Do you have to use the `=>` to remove the closure allocation? I think I remember that being the case at one point, but I figured that was a compiler bug that should be fixed by now.

benaadams: Yep; I'd agree is bug; but not fixed yet: https://github.com/dotnet/roslyn/issues/5835

PR #320, opened 2015-11-01
ListenerSecondary - reduce closure allocation

benaadams:
Simplify lamba structure

~~Still captures `buf` and `ptr`~~

### DIFF
--- a/src/Microsoft.AspNet.Server.Kestrel/Http/ListenerSecondary.cs
+++ b/src/Microsoft.AspNet.Server.Kestrel/Http/ListenerSecondary.cs
@@ -92,11 +92,8 @@ namespace Microsoft.AspNet.Server.Kestrel.Http
             {
                 DispatchPipe.ReadStart(
                     (handle, status2, state) => ((ListenerSecondary)state)._buf,
-                    (handle, status2, state) => 
-                        {
-                            var listener = ((ListenerSecondary)state);
-                            listener.ReadStartCallback(handle, status2, listener._ptr);
-                        }, this);
+                    (handle, status2, state) => ((ListenerSecondary)state).ReadStartCallback(handle, status2), 
+                    this);
 
                 tcs.SetResult(0);
             }
@@ -107,7 +104,7 @@ namespace Microsoft.AspNet.Server.Kestrel.Http
             }
         }
 
-        private void ReadStartCallback(UvStreamHandle handle, int status, IntPtr ptr)
+        private void ReadStartCallback(UvStreamHandle handle, int status)
         {
             if (status < 0)
             {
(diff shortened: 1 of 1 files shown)


src/Microsoft.AspNet.Server.Kestrel/Http/ListenerSecondary.cs
```diff
@@ -29,95 +31,112 @@ protected ListenerSecondary(ServiceContext serviceContext) : base(serviceContext
...
+                DispatchPipe.Dispose();
+                tcs.SetException(ex);
+            }
+        }
+
+        private static void ConnectCallback(UvConnectRequest connect, int status, Exception error, TaskCompletionSource<int> tcs)
+        {
+            var listener = (ListenerSecondary)tcs.Task.AsyncState;
+            listener.ConnectedCallback(connect, status, error, tcs);
+        }
```
halter73: Nit: missing newline.

src/Microsoft.AspNet.Server.Kestrel/Http/ListenerSecondary.cs
```diff
@@ -29,95 +31,112 @@ protected ListenerSecondary(ServiceContext serviceContext) : base(serviceContext
...
+                tcs.SetException(error);
+                return;
+            }
+
+            try
+            {
+                var ptr = Marshal.AllocHGlobal(4);
+                var buf = Thread.Loop.Libuv.buf_init(ptr, 4);
+
+                DispatchPipe.ReadStart(
```
halter73: Both of these callbacks result in closure allocation because of buf and ptr. This is probably one of the more important places to avoid it, since it is per connection. Since we don't actually use the buffer, I think we could make the pointer and buffer a field, and only free it when ListenerSecondary is disposed.

What do you think?

benaadams: Sounds good; didn't know enough to make the call. Will make the change.

src/Microsoft.AspNet.Server.Kestrel/Http/ListenerSecondary.cs
benaadams: Added

src/Microsoft.AspNet.Server.Kestrel/Http/ListenerSecondary.cs
```diff
@@ -29,95 +33,117 @@ protected ListenerSecondary(ServiceContext serviceContext) : base(serviceContext
...
+            if (error != null)
+            {
+                tcs.SetException(error);
+                return;
+            }
+
+            try
+            {
+                _ptr = Marshal.AllocHGlobal(4);
+                _buf = Thread.Loop.Libuv.buf_init(_ptr, 4);
```
halter73: I would just initialize these fields in the ctor. Then we can have only one call to `Marshal.FreeHGlobal` in Dispose.

benaadams: Done

halter73:
I merged this into dev. I didn't catch your last commit, but I made similar changes after testing the change on linux where I could run the ZeroToTenThreads functional test and verify that ListenerSecondary still works.

PR #321, opened 2015-11-01
Reduce delegate allocation

benaadams:


khellang:
@benaadams You're on a roll! Seen this? https://github.com/dotnet/roslyn/issues/5835

benaadams:
@khellang Soon; hopefully, but not yet... :(

src/Microsoft.AspNet.Server.Kestrel/Http/MessageBody.cs
```diff
@@ -94,7 +94,7 @@ public static bool TryGet(IDictionary<string, StringValues> headers, string name
                 value = values[0];
                 return true;
             }
-            value = String.Join(",", values);
+            value = string.Join(",", values.ToArray());
```
halter73: Before this change wasn't [this overload of string.Join](https://github.com/dotnet/coreclr/blob/33323a204ab92b5f8762461b8915b7ffa6d89e69/src/mscorlib/src/System/String.cs#L156) being used? Is there a reason to avoid that overload?

benaadams: Yep, boxes enumerator to `IEnumerator`; as items > 1 we know `.ToArray` will return an underlying array without creation.

PR #324, opened 2015-11-03
Strong name everything.

cesarbs:
https://github.com/aspnet/dnx/issues/2615

cc @davidfowl @Eilon @muratg @moozzyk @pranavkm @Tratcher @halter73 @troydai


tools/Microsoft.StandardsPolice/project.json
```diff
@@ -1,6 +1,10 @@
 ﻿{
   "version": "1.0.0-*",
 
+  "compilationOptions": {
```
Eilon: Hmm this one shouldn't be needed I don't think. This is just a build tool, nothing compiles against it as far as I know.

cesarbs: There's a reference to it in src/Microsoft.AspNet.Server.Kestrel/project.json.

Eilon: It's a build-time dependency, not a real dependency, so it shouldn't matter. Did you have any issues without it?

cesarbs: It wouldn't build.

Eilon: That doesn't sound right to me - nothing should be using it as a reference. Can you show what the error was?

cesarbs: Building Microsoft.AspNet.Server.Kestrel for DNX,Version=v4.5.1
Error: Could not load file or assembly 'Microsoft.StandardsPolice, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null' or one of its dependencies. A strongly-named assembly is required. (Exception from HRESULT: 0x80131044)
fail: Exit code 1 from cmd

Eilon: I think it's certainly harmless for RC1 so feel free to ignore this for now, but @lodejard @davidfowl doesn't this look wrong to you? The strong-named-ness of a build-time dependency shouldn't affect the main project should it?

davidfowl: Hmmmm, @pakrym are we still strong naming meta assemblies? @Eilon we need to figure out if this will affect everyone doing ICompileModule(s)

pakrym: @davidfowl @Eilon https://github.com/aspnet/dnx/issues/3123

pakrym: @Eilon @davidfowl we compile build time dependent projects using their own compilation options to separate assembly, thats why if it's not signed we get load exception.

Eilon: @pakrym I see. @davidfowl is it worth logging a DNX bug to track this? Would this still be an issue in the future?

PR #325, opened 2015-11-03
Support for IHttpRequestLifetimeFeature

halter73:


src/Microsoft.AspNet.Server.Kestrel/Http/Frame.FeatureCollection.cs
```diff
@@ -297,5 +302,26 @@ Task<Stream> IHttpUpgradeFeature.UpgradeAsync()
...
         IEnumerator IEnumerable.GetEnumerator() => FastEnumerable().GetEnumerator();
+
+        void IHttpRequestLifetimeFeature.Abort()
+        {
+            _requestProcessingStopping = true;
+            _requestAbort = true;
+
+            try
+            {
+                _requestAbortCts?.Cancel();
```
Tratcher: Do this last, it may block in user code.

src/Microsoft.AspNet.Server.Kestrel/Http/Frame.cs
```diff
@@ -41,6 +41,9 @@ public partial class Frame : FrameContext, IFrameControl
         private bool _requestProcessingStarted;
         private Task _requestProcessingTask;
         private volatile bool _requestProcessingStopping; // volatile, see: https://msdn.microsoft.com/en-us/library/x13ttww7.aspx
+        private volatile bool _requestAbort;
```
Tratcher: _requestAborted

src/Microsoft.AspNet.Server.Kestrel/Http/FrameRequestStream.cs
```diff
@@ -102,5 +109,11 @@ public override void Write(byte[] buffer, int offset, int count)
         {
             throw new NotImplementedException();
         }
+
+        protected override void Dispose(bool disposing)
```
Tratcher: Use an internal API instead, you don't want random application code to dispose the streams.

Tratcher:
We also need to detect client disconnects and fire the CT.

davidfowl:
Why is travis failing

src/Microsoft.AspNet.Server.Kestrel/Http/Connection.cs
```diff
@@ -98,6 +98,20 @@ public void Start()
             }
         }
 
+        public void Abort()
+        {
+            if (_frame != null)
+            {
+                // Frame.Abort calls user code while tis method is always
```
davidfowl: typo tis -> this

src/Microsoft.AspNet.Server.Kestrel/Http/Frame.cs
```diff
@@ -200,12 +230,17 @@ public Task Stop()
...
-                        ResponseBody = new FrameResponseStream(this);
+                        var messageBody = MessageBody.For(HttpVersion, _requestHeaders, this);
+                        _keepAlive = messageBody.RequestKeepAlive;
+                        _requestBody = new FrameRequestStream(messageBody);
+                        RequestBody = _requestBody;
+                        _responseBody = new FrameResponseStream(this);
+                        ResponseBody = _responseBody;
                         DuplexStream = new FrameDuplexStream(RequestBody, ResponseBody);
 
+                        _requestAbortCts = CancellationTokenSource.CreateLinkedTokenSource(_disconnectCts.Token);
```
davidfowl: This has never been efficient... We should measure

benaadams:
> Why is travis failing

Hang, will work eventually after you resubmit it a couple times  #338

benaadams:
Some of the constructors for `SocketOutput` in `SocketOutputTests.cs` still need to be updated

src/Microsoft.AspNet.Server.Kestrel/Http/Frame.cs
```diff
@@ -168,6 +174,30 @@ public Task Stop()
...
+        /// </summary>
+        public void Abort()
+        {
+            _requestProcessingStopping = true;
+            _requestAborted = true;
+
+            _requestBody?.Abort();
+            _responseBody?.Abort();
+
+            ConnectionControl.End(ProduceEndType.SocketDisconnect);
```
Tratcher: Can this throw?

src/Microsoft.AspNet.Server.Kestrel/Http/Frame.cs
halter73: I've found that there isn't a measurable perf hit if no one registers to the RequestAborted token.

When running the wrk against https://github.com/aspnet/benchmarks' plaintext endpoint and registering with the RequestAborted token on each request there is about a 10% decrease in throughput.

PR #326, opened 2015-11-03
Move logging to new style (remove allocs)

benaadams:
and additionally boxing param issue shown below:

img

Takes allocations for logging calls to zero

### DIFF
--- a/src/Microsoft.AspNet.Server.Kestrel/Infrastructure/KestrelTrace.cs
+++ b/src/Microsoft.AspNet.Server.Kestrel/Infrastructure/KestrelTrace.cs
@@ -12,7 +12,34 @@ namespace Microsoft.AspNet.Server.Kestrel
     /// </summary>
     public class KestrelTrace : IKestrelTrace
     {
+        private static Action<ILogger, long, Exception> _connectionStart;
+        private static Action<ILogger, long, Exception> _connectionStop;
+        private static Action<ILogger, long, Exception> _connectionPause;
+        private static Action<ILogger, long, Exception> _connectionResume;
+        private static Action<ILogger, long, Exception> _connectionReadFin;
+        private static Action<ILogger, long, Exception> _connectionWriteFin;
+        private static Action<ILogger, long, int, Exception> _connectionWroteFin;
+        private static Action<ILogger, long, Exception> _connectionKeepAlive;
+        private static Action<ILogger, long, Exception> _connectionDisconnect;
+
         protected readonly ILogger _logger;
+
+        static KestrelTrace()
+        {
+            _connectionStart = LoggerMessage.Define<long>(LogLevel.Debug, 1, @"Connection id ""{ConnectionId}"" started.");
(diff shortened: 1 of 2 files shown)


rynowak:
We already have a pattern for this https://github.com/aspnet/Mvc/blob/dev/src/Microsoft.AspNet.Mvc.Core/Logging/InnerAttributeRouteLoggerExtensions.cs#L6

benaadams:
That's too weird for me :)

Raised issue #327 for it instead

benaadams:
Actually, quite like it, updated.

Resolves  #327

rynowak:
The pattern for this resolves a few perf issues by doing things like avoiding `params object[]` on both paths, and pre-parsing and caching the format string for you.

If you need to do significant work before calling the delegate, do a manual `IsEnabled` check.

We've measured this and verified that it's allocation free when logging is off, and allocates one object when logging is on. We might take it further in the next milestone to get rid of that allocation when logging is on as well.

src/Microsoft.AspNet.Server.Kestrel/Infrastructure/KestrelTrace.cs
```diff
@@ -12,6 +12,32 @@ namespace Microsoft.AspNet.Server.Kestrel
...
+            _connectionStop = LoggerMessage.Define<long>(LogLevel.Debug, 2, @"Connection id ""{ConnectionId}"" stopped.");
+            // ConnectionRead 3
+            _connectionPause = LoggerMessage.Define<long>(LogLevel.Debug, 4, @"Connection id ""{ConnectionId}"" paused.");
+            _connectionResume = LoggerMessage.Define<long>(LogLevel.Debug, 5, @"Connection id ""{ConnectionId}"" resumed.");
+            _connectionReadFin = LoggerMessage.Define<long>(LogLevel.Debug, 6, @"Connection id ""{ConnectionId}"" received FIN.");
+            _connectionWriteFin = LoggerMessage.Define<long>(LogLevel.Debug, 7, @"Connection id ""{ConnectionId}"" sending FIN.");
+            _connectionWroteFin = LoggerMessage.Define<long, int>(LogLevel.Debug, 8, @"Connection id ""{ConnectionId}"" sent FIN with status ""{Status}"".");
+            _connectionKeepAlive = LoggerMessage.Define<long>(LogLevel.Debug, 9, @"Connection id ""{ConnectionId}"" completed keep alive response.");
+            _connectionDisconnect = LoggerMessage.Define<long>(LogLevel.Error, 10, @"Connection id ""{ConnectionId}"" disconnected.");
```
halter73: Should ApplicationError also use this patern?

benaadams: I'd imagine so; but would need the following overload (as `Exception` is is a built in param)?

```csharp
static Action<ILogger, Exception> Define(LogLevel logLevel, int eventId, string formatString)
```
in `Microsoft.Extensions.Logging.Abstractions/LoggerMessage.cs` which doesn't currently exist.

benaadams:
Rebased on #337

benaadams:
removed rebase

PR #329, opened 2015-11-04
Don't alloc array per header

benaadams:
i.e `new StringValues(string)` rather than `new StringValues(new [] {
string } )`

Removes an alloc per header per request


davidfowl:
This is a no brainer

PR #330, opened 2015-11-04
MemoryPool2 Allocate returns newest

benaadams:
Rather than a while loop on Allocate; return last memory block created
rather than returning it to the pool and checking if one can be removed.

### DIFF
--- a/src/Microsoft.AspNet.Server.Kestrel/Infrastructure/MemoryPool2.cs
+++ b/src/Microsoft.AspNet.Server.Kestrel/Infrastructure/MemoryPool2.cs
@@ -96,11 +96,11 @@ namespace Microsoft.AspNet.Server.Kestrel.Infrastructure
             var basePtr = slab.ArrayPtr;
             var firstOffset = (int)((_blockStride - 1) - ((ulong)(basePtr + _blockStride - 1) % _blockStride));
 
-            var poolAllocationLength = _slabLength - (_blockLength + _blockStride);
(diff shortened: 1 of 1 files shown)


src/Microsoft.AspNet.Server.Kestrel/Infrastructure/MemoryPool2.cs
```diff
@@ -99,8 +96,11 @@ private void AllocateSlab()
             var basePtr = slab.ArrayPtr;
             var firstOffset = (int)((_blockStride - 1) - ((ulong)(basePtr + _blockStride - 1) % _blockStride));
 
-            for (var offset = firstOffset;
-                offset + _blockLength <= _slabLength;
+            var poolAllocationLength = _slabLength - (_blockLength + _blockStride);
```
halter73: I think semantically it makes more sense for `poolAllocationLength` to be `_slabLength - _blockStride`. The for condition then could go back to `offset + _blockLength < poolAllocationLength`.

halter73:
:shipit: @lodejard what do you think?

src/Microsoft.AspNet.Server.Kestrel/Infrastructure/MemoryPool2.cs
benaadams: Much more readable; changed

lodejard:
Makes sense

PR #331, opened 2015-11-05
GC.Collect before pinning starts

benaadams:
After hitting with very heavy load Kestrel seems to idle lower

gc-off

to

gc-on

may just be coincidence/timing though...


davidfowl:
Travis failing

benaadams:
Mono hates me

benaadams:
Was rebased on #337 to reduce probability of CI failures

benaadams:
Wrong place #362 does it betterer

PR #335, opened 2015-11-06
Consume body without allocating or reading

### LINKED ISSUE #334: RequestBody draining still allocates per-request buffers
This change: https://github.com/aspnet/KestrelHttpServer/commit/4bba074d778933b312011df0ade8ff875250cded

Fixed a serious allocation issue, but also added a 4kb buffer to each frame, which is more per-request allocation. Simply just making this `static` as well may not do what we want, as this might lead to a lot of contention for this cache line.

I think we need to revisit the original proposal which was to implement `CopyToStreamAsync` in a smarter way.

Allocation data for 3000 requests:
image

In my sample app, this is responsible for 12/82mb allocated

benaadams:
Resolves #334

### DIFF
--- a/src/Microsoft.AspNet.Server.Kestrel/Filter/LibuvStream.cs
+++ b/src/Microsoft.AspNet.Server.Kestrel/Filter/LibuvStream.cs
@@ -85,7 +85,7 @@ namespace Microsoft.AspNet.Server.Kestrel.Filter
 
         private Task<int> ReadAsync(ArraySegment<byte> buffer)
         {
-            return _input.ReadAsync(buffer);
+            return _input.ReadAsync(buffer.Array, buffer.Offset, buffer.Count);
         }
     }
 }
--- a/src/Microsoft.AspNet.Server.Kestrel/Http/Frame.cs
+++ b/src/Microsoft.AspNet.Server.Kestrel/Http/Frame.cs
@@ -232,10 +232,8 @@ namespace Microsoft.AspNet.Server.Kestrel.Http
 
                             await ProduceEnd();
 
-                            while (await MessageBody.SkipAsync() != 0)
-                            {
-                                // Finish reading the request body in case the app did not.
-                            }
+                            // Finish reading the request body in case the app did not.
+                            await MessageBody.Consume();
                         }
 
                         terminated = !_keepAlive;
--- a/src/Microsoft.AspNet.Server.Kestrel/Http/Frame.cs
+++ b/src/Microsoft.AspNet.Server.Kestrel/Http/Frame.cs
@@ -232,7 +232,7 @@ namespace Microsoft.AspNet.Server.Kestrel.Http
 
                             await ProduceEnd();
 
-                            while (await RequestBody.ReadAsync(_nullBuffer, 0, _nullBuffer.Length) != 0)
+                            while (await MessageBody.SkipAsync() != 0)
                             {
                                 // Finish reading the request body in case the app did not.
                             }
--- a/src/Microsoft.AspNet.Server.Kestrel/Http/MessageBody.cs
+++ b/src/Microsoft.AspNet.Server.Kestrel/Http/MessageBody.cs
@@ -38,7 +38,25 @@ namespace Microsoft.AspNet.Server.Kestrel.Http
             return result;
         }
 
+        public Task<int> SkipAsync(CancellationToken cancellationToken = default(CancellationToken))
+        {
+            Task<int> result = null;
+            var send100Continue = 0;
+            result = SkipImplementation(cancellationToken);
+            if (!result.IsCompleted)
+            {
+                send100Continue = Interlocked.Exchange(ref _send100Continue, 0);
+            }
+            if (send100Continue == 1)
+            {
+                _context.FrameControl.ProduceContinue();
+            }
+            return result;
+        }
+
         public abstract Task<int> ReadAsyncImplementation(ArraySegment<byte> buffer, CancellationToken cancellationToken);
+
+        public abstract Task<int> SkipImplementation(CancellationToken cancellationToken);
 
         public static MessageBody For(
             string httpVersion,
@@ -110,6 +128,10 @@ namespace Microsoft.AspNet.Server.Kestrel.Http
             {
                 return _context.SocketInput.ReadAsync(buffer);
             }
+            public override Task<int> SkipImplementation(CancellationToken cancellationToken)
+            {
+                return _context.SocketInput.SkipAsync(4096);
+            }
         }
 
         class ForContentLength : MessageBody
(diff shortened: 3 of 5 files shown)


benaadams:
Rebased on #337

davidfowl:
This needs to be implemented in a non allocating way.

benaadams:
Not sure I understand, it only allocates if the header name > 4032 bytes so it shouldn't ever allocate?

benaadams:
Stack to Queue test by @NickCraver  https://twitter.com/Nick_Craver/status/665503509958258688

benaadams:
Skip rather than Read

src/Microsoft.AspNet.Server.Kestrel/Infrastructure/MemoryPoolIterator2.cs
```diff
@@ -566,7 +566,7 @@ public string GetString(MemoryPoolIterator2 end)
             }
         }
 
-        public ArraySegment<byte> GetArraySegment(MemoryPoolIterator2 end)
+        public ArraySegment<byte> GetArraySegment(MemoryPoolIterator2 end, ArraySegment<byte> scratchBuffer)
```
davidfowl: This seems unrelated

src/Microsoft.AspNet.Server.Kestrel/Infrastructure/MemoryPool2.cs
```diff
@@ -40,7 +46,7 @@ public class MemoryPool2 : IDisposable
         /// Thread-safe collection of blocks which are currently in the pool. A slab will pre-allocate all of the block tracking objects
         /// and add them to this collection. When memory is requested it is taken from here first, and when it is returned it is re-added.
         /// </summary>
-        private readonly ConcurrentStack<MemoryPoolBlock2> _blocks = new ConcurrentStack<MemoryPoolBlock2>();
+        private readonly ConcurrentQueue<MemoryPoolBlock2> _blocks = new ConcurrentQueue<MemoryPoolBlock2>();
```
davidfowl: This is pretty unrelated as well

src/Microsoft.AspNet.Server.Kestrel/Http/Frame.cs
```diff
@@ -231,8 +230,8 @@ public Task Stop()
                             HttpContextFactory.Dispose(httpContext);
 
                             await ProduceEnd();
-
-                            while (await RequestBody.ReadAsync(_nullBuffer, 0, _nullBuffer.Length) != 0)
+                            
+                            while (await MessageBody.SkipAsync() != 0)
```
davidfowl: I think this should be a single call instead of a loop.

`RequestBody.Consume()`

src/Microsoft.AspNet.Server.Kestrel/Http/Frame.cs
```diff
@@ -689,12 +688,7 @@ private bool TakeStartLine(SocketInput input)
             }
         }
 
-        static string GetString(ArraySegment<byte> range, int startIndex, int endIndex)
-        {
-            return Encoding.UTF8.GetString(range.Array, range.Offset + startIndex, endIndex - startIndex);
-        }
-
-        public static bool TakeMessageHeaders(SocketInput input, FrameRequestHeaders requestHeaders)
+        public static bool TakeMessageHeaders(SocketInput input, FrameRequestHeaders requestHeaders, MemoryPool2 memorypool)
```
davidfowl: This change seems unrelated.

src/Microsoft.AspNet.Server.Kestrel/Http/MessageBody.cs
```diff
@@ -236,6 +279,78 @@ public ForChunkedEncoding(bool keepAlive, FrameContext context)
                 return 0;
             }
 
+            public override async Task<int> SkipAsyncImplementation(CancellationToken cancellationToken)
```
davidfowl: Feels like all of the duplication can be removed by just passing a delegate (though the out parameter won't work :smile: )

benaadams:
Moved unrelated to #390 and #391

benaadams:
@davidfowl does the last commit sort out your code reuse concerns?

davidfowl:
@benaadams Yes, exactly what I had in mind (except the 100 continue changes)

src/Microsoft.AspNet.Server.Kestrel/Http/MessageBody.cs
```diff
@@ -38,6 +38,34 @@ public Task<int> ReadAsync(ArraySegment<byte> buffer, CancellationToken cancella
             return result;
         }
 
+        public async Task Consume(CancellationToken cancellationToken = default(CancellationToken))
+        {
+            Task<int> result;
+            var firstLoop = true;
+            do
+            {
+                result = ReadAsyncImplementation(default(ArraySegment<byte>), cancellationToken);
```
davidfowl: Why not await here?

benaadams: Was copying the function above. Following the logic of that function; the task hasn't completed because no data; check whether their is no data because the client is waiting for a "100 continue" else the await will block infinitely as it waiting for data that's dependent on a send; if so do the send and return the incomplete task?

Since we are doing the full loop here; if we pass over that condition; we know the task has already completed. If the result is `0` then return as everything is done; otherwise move to next loop; so we only await if the task is not complete.

benaadams: Also `Interlocked` is much more expensive than a local var and only one  `100 continue` is sent, so only check it once; though `firstLoop` is probably the wrong name.

src/Microsoft.AspNet.Server.Kestrel/Filter/LibuvStream.cs
```diff
@@ -85,7 +85,7 @@ public override void Flush()
 
         private Task<int> ReadAsync(ArraySegment<byte> buffer)
         {
-            return _input.ReadAsync(buffer);
+            return _input.ReadAsync(buffer.Array, buffer.Offset, buffer.Count);
```
rynowak: Worth just overloading this on `SocketInput`? Notice a bunch of other methods here create an array segment to call this method.

At least least, you could clean all of those up to avoid creating the `ArraySegment`

src/Microsoft.AspNet.Server.Kestrel/Http/Frame.cs
```diff
@@ -232,10 +232,8 @@ public Task Stop()
 
                             await ProduceEnd();
 
-                            while (await RequestBody.ReadAsync(_nullBuffer, 0, _nullBuffer.Length) != 0)
-                            {
-                                // Finish reading the request body in case the app did not.
-                            }
+                            // Finish reading the request body in case the app did not.
+                            await MessageBody.Consume();
```
rynowak: :dancer:

src/Microsoft.AspNet.Server.Kestrel/Filter/LibuvStream.cs
benaadams: Can't create an `ArraySegement` with a null buffer + specified length [or it will throw an exception](https://github.com/dotnet/coreclr/blob/bc146608854d1db9cdbcc0b08029a87754e12b49/src/mscorlib/src/System/ArraySegment.cs#L45).

However a consumption limit needs to be passed through to `CopyTo` else it will consume the whole input rather than just the input related to the current request. Is a bit of abstraction leakiness to make the code reusable; rather than copy paste repeated...

rynowak:
:fireworks:

Issue #259, opened 2015-10-14
Ran into memory corruption running tests


davidfowl:
![image](https://cloud.githubusercontent.com/assets/95136/10477909/c2c685a2-720e-11e5-926b-a4d15320eb9c.png)

Didn't happen twice in a row though.

halter73:
It seems like the same issue came up in an AppVeyor build:

victorhurdugaci:
Happened on the CI too:

```
 [xunit-test] Unhandled Exception: System.AccessViolationException: Attempted to read or write protected memory. This is often an indication that other memory is corrupt.
[13:39:57]E:				 [xunit-test]    at Microsoft.AspNet.Server.Kestrel.Networking.Libuv.async_send(UvAsyncHandle handle)
[13:39:57]E:				 [xunit-test]    at Microsoft.AspNet.Server.Kestrel.KestrelThread.Post[T](Action`1 callback, T state)
[13:39:57]E:				 [xunit-test]    at Microsoft.AspNet.Server.Kestrel.Http.SocketOutput.ScheduleWrite()
[13:39:57]E:				 [xunit-test]    at Microsoft.AspNet.Server.Kestrel.Http.SocketOutput.Write(ArraySegment`1 buffer, Action`3 callback, Object state, Boolean immediate, Boolean socketShutdownSend, Boolean socketDisconnect)
[13:39:57]E:				 [xunit-test]    at Microsoft.AspNet.Server.Kestrel.Http.SocketOutput.End(ProduceEndType endType)
[13:39:57]E:				 [xunit-test]    at Microsoft.AspNet.Server.Kestrel.Http.Connection.Microsoft.AspNet.Server.Kestrel.Http.IConnectionControl.End(ProduceEndType endType)
[13:39:57]E:				 [xunit-test]    at Microsoft.AspNet.Server.Kestrel.Http.Frame.ProduceEnd(Exception ex)
[13:39:57]E:				 [xunit-test]    at Microsoft.AspNet.Server.Kestrel.Http.Frame.<RequestProcessingAsync>d__77.MoveNext()
[13:39:57]E:				 [xunit-test]    at System.Runtime.CompilerServices.AsyncMethodBuilderCore.MoveNextRunner.InvokeMoveNext(Object stateMachine)
[...]

Issue #299, opened 2015-10-28
Don't throw UvException directly to users


Tratcher:
Kestrel request and response Streams should throw IOExceptions, not UvExceptions.
```
An exception of type 'Microsoft.AspNet.Server.Kestrel.Networking.UvException' occurred in mscorlib.dll but was not handled in user code

Additional information: Error -4081 ECANCELED operation canceled

   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at Microsoft.AspNet.Server.Kestrel.Http.Frame.<WriteChunkedAsync>d__88.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at Microsoft.AspNet.Server.Kestrel.Http.Frame.<WriteAsync>d__86.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.GetResult()
   at Glimpse.Server.Resources.ServerSentEventResponse.<SendData>d__5.MoveNext() in C:\Users\antho\Projects\Glimpse.Prototype\src\Glimpse.Server\Resources\HttpContextExtensions.cs:line 100
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.GetResult()
   at Glimpse.Server.Internal.Resources.MessageStreamResource.<>c__DisplayClass4_0.<<Configure>b__1>d.MoveNext() in C:\Users\antho\Projects\Glimpse.Prototype\src\Glimpse.Server\Internal\Resources\MessageStreamResource.cs:line 63
[...]

PR #337, opened 2015-11-06
Don't post to closed socket (CI Fix)

### LINKED ISSUE #336: Handle Invalid
A handle is getting lost somewhere and either finalised or early disposed; it most often shows up on mono and randomly causes tests to fail; mostly tarvis in but occasionally also happens on appveyor.

Haven't be able to track down the cause

e.g. https://github.com/aspnet/KestrelHttpServer/pull/335 with the following resilts: https://travis-ci.org/aspnet/KestrelHttpServer/builds/89566455

> info:   workingdir: test/Microsoft.AspNet.Server.Kestrel.FunctionalTests
  Discovering: Microsoft.AspNet.Server.Kestrel.FunctionalTests
  Discovered:  Microsoft.AspNet.Server.Kestrel.FunctionalTests
  Starting:    Microsoft.AspNet.Server.Kestrel.FunctionalTests
Stacktrace:
Native stacktrace:
	mono() [0x4b23dc]
	mono() [0x508a0e]
	mono() [0x428fad]
	/lib/x86_64-linux-gnu/libpthread.so.0(+0xfcb0) [0x7fd272101cb0]
	/home/travis/libuvinstall/lib/libuv.so(uv_async_send+0x4) [0x7fd25eae0844]
	[0x40e97908]
Debug info from gdb:
>  =================================================================
Got a SIGSEGV while executing native code. This usually indicates
a fatal error in the mono runtime or one of the native libraries 
used by your application.
>  =================================================================

If I submit again it will likely pass the test.

benaadams:
Resolves #336

### DIFF
--- a/src/Microsoft.AspNet.Server.Kestrel/Http/SocketOutput.cs
+++ b/src/Microsoft.AspNet.Server.Kestrel/Http/SocketOutput.cs
@@ -140,6 +140,10 @@ namespace Microsoft.AspNet.Server.Kestrel.Http
             {
                 _thread.Post(_this => _this.WriteAllPending(), this);
             }
+            else
+            {
+                CompleteAllCallbacksWithError(_lastWriteError ?? new InvalidOperationException("Socket is closed"));
+            }
         }
 
         // This is called on the libuv event loop
@@ -175,6 +179,20 @@ namespace Microsoft.AspNet.Server.Kestrel.Http
                 }
 
                 throw;
+            }
+        }
+        
+        private void CompleteAllCallbacksWithError(Exception error)
+        {
+            lock (_lockObj)
+            {
+                while (_callbacksPending.Count > 0)
+                {
+                    var callbackContext = _callbacksPending.Dequeue();
+
+                    // callback(error, state, calledInline)
+                    callbackContext.Callback(error, callbackContext.State, false);
+                }
             }
         }
 
--- a/src/Microsoft.AspNet.Server.Kestrel/Http/SocketOutput.cs
+++ b/src/Microsoft.AspNet.Server.Kestrel/Http/SocketOutput.cs
@@ -135,7 +135,11 @@ namespace Microsoft.AspNet.Server.Kestrel.Http
 
         private void ScheduleWrite()
         {
-            _thread.Post(_this => _this.WriteAllPending(), this);
+            // Don't post write to closed socket
+            if (!_socket.IsClosed)
+            {
+                _thread.Post(_this => _this.WriteAllPending(), this);
+            }
         }
 
         // This is called on the libuv event loop
--- a/src/Microsoft.AspNet.Server.Kestrel/Networking/Libuv.cs
+++ b/src/Microsoft.AspNet.Server.Kestrel/Networking/Libuv.cs
@@ -2,6 +2,7 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.Diagnostics;
 using System.Runtime.InteropServices;
 
 namespace Microsoft.AspNet.Server.Kestrel.Networking
@@ -199,6 +200,9 @@ namespace Microsoft.AspNet.Server.Kestrel.Networking
         protected Func<UvAsyncHandle, int> _uv_async_send;
         public void async_send(UvAsyncHandle handle)
         {
+            // Can't Assert with .Validate as that checks threadId
+            // and this function is to post to correct thread.
+            Debug.Assert(!handle.IsInvalid, "Handle is invalid");
             Check(_uv_async_send(handle));
         }
 


benaadams:
Also resolves #259

halter73:
[I agree](https://github.com/aspnet/KestrelHttpServer/issues/259#issuecomment-148865655) that #336 and #259 are likely the same issue.

I had considered trying a fix like this, but I don't think checking `SafeHandle.IsClosed` is sufficient to avoid the AVs. SafeHandles should throw an ODE if used after being closed like the following:

    System.ObjectDisposedException: Safe handle has been closed
       at System.Runtime.InteropServices.SafeHandle.DangerousAddRef(Boolean& success)
       at System.StubHelpers.StubHelpers.SafeHandleAddRef(SafeHandle pHandle, Boolean& success)
       at Microsoft.AspNet.Server.Kestrel.Networking.Libuv.NativeMethods.uv_async_send(UvAsyncHandle handle)
       at Microsoft.AspNet.Server.Kestrel.Networking.Libuv.async_send(UvAsyncHandle handle) in C:\Users\shalter\dev\Universe\KestrelHttpServer\src\Microsoft.AspNet.Server.Kestrel\Networking\Libuv.cs:line 202
       at Microsoft.AspNet.Server.Kestrel.Networking.UvAsyncHandle.Send() in C:\Users\shalter\dev\Universe\KestrelHttpServer\src\Microsoft.AspNet.Server.Kestrel\Networking\UvAsyncHandle.cs:line 37

 This isn't the worst thing as long as long as we wrap the ODE in an `IOException` (https://github.com/aspnet/KestrelHttpServer/issues/299).

I actually modified the AsyncCanBeSent test to try to force the AV and ran it for a couple hours today, but I had no luck:

    [Fact]
    public void AsyncCanBeSent()
    {
        UvAsyncHandle triggerRef = null;
        var tasks = new Task[2];

        tasks[0] = Task.Run(() =>
        {
            while (true)
            {
                var loop = new UvLoopHandle(_logger);
                loop.Init(_uv);
                var trigger = new UvAsyncHandle(_logger);
                trigger.Init(loop, () =>
                {
                    called = true;
                    trigger.Dispose();
                });

                triggerRef = trigger;

                trigger.Send();
                loop.Run();
                loop.Dispose();
            }
        });

        tasks[1] = Task.Run(() =>
        {
            while (true)
            {
                if (triggerRef != null)
                {
                    try
                    {
                        triggerRef.Send();
                    }
                    catch (ObjectDisposedException ex)
                    {
                    }
                }
            }
        });

        Task.WhenAny(tasks).Result.Wait();
    }

I think there's more to this than meets the eye.

src/Microsoft.AspNet.Server.Kestrel/Networking/Libuv.cs
```diff
@@ -199,6 +200,9 @@ public void async_init(UvLoopHandle loop, UvAsyncHandle handle, uv_async_cb cb)
         protected Func<UvAsyncHandle, int> _uv_async_send;
         public void async_send(UvAsyncHandle handle)
         {
+            // Can't Assert with .Validate as that checks threadId
+            // and this function is to post to correct thread.
+            Debug.Assert(!handle.IsInvalid, "Handle is invalid");
```
halter73: If this was intended to prevent AVs, it could only have any affect on debug builds. I don't think even Trace.Assert would save us from the AVs we've been seeing.

This might be nice to have, but I doubt this fixes any bugs.

src/Microsoft.AspNet.Server.Kestrel/Http/SocketOutput.cs
```diff
@@ -135,7 +135,15 @@ public void End(ProduceEndType endType)
 
         private void ScheduleWrite()
         {
-            _thread.Post(_this => _this.WriteAllPending(), this);
+            // Don't post write to closed socket
+            if (!_socket.IsClosed)
```
halter73: Is this meant to be a perf optimization? `_thread.Post` shouldn't fail because the socket is closed.

We only actually use `_socket` in the `WriteContext` methods where `_socket.IsClosed` is always checked first thing.

src/Microsoft.AspNet.Server.Kestrel/Networking/Libuv.cs
benaadams: Helped me track down the bug; only needed for debug.

src/Microsoft.AspNet.Server.Kestrel/Http/SocketOutput.cs
benaadams: The issue is the loop is the gone hence all the calls erroring are actually `uv_async_send` so its post that fails. However the socket has been successfully closed on the server; so a check to see if the socket is closed fixes it; rather than checking if the loop is valid.

It occurs when `socketShutdownSend` has happened; the client closes the socket, the server picks this up and then the close is kicked off; so when there is a slight delay between `socketShutdownSend` and `socketDisconnect` while the loop has terminated because its been asked to and all the sockets are closed.

Its potentially only in a full shutdown scenarios (e.g. the tests or controlled app recycle)

benaadams:
Its when loop has been shutdown so trying to queue `socketDisconnect` fails; because the client shut it down as part of `socketShutdownSend` and the loop picked it up, then decided it had no outstanding work so was safe to shutdown.

benaadams:
The tests will cause this to occur more than I imagine it would happen in the wild as I think it only occurs on shutdown.

benaadams:
Basically this fix is don't post to terminated loop; but the socket is also closed and is more local.

benaadams:
This is too annoying, rebasing other back PRs onto it.

src/Microsoft.AspNet.Server.Kestrel/Networking/Libuv.cs
benaadams: If you put just this change in and run ./build in Windows; debug when the Assert pops up you can clearly see what's going on.

src/Microsoft.AspNet.Server.Kestrel/Http/SocketOutput.cs
davidfowl: It might be better to expose something from KestrelThread so that callers know?

benaadams: I think its only in a shutdown race situation; happens in both windows and mono. When it happens windows seems ok with it, but mono very unhappy.

benaadams: for example `OnWriteCompleted` calls it

benaadams: Will mitigate this in a later PR; but this will stop the AV for now.

Tragetaschen:
@davidfowl That what "callers know" is normally the `Connection`, but the current design of Kestrel [fires and forgets](https://github.com/aspnet/KestrelHttpServer/blob/dev/src/Microsoft.AspNet.Server.Kestrel/Http/Listener.cs#L79) any client request and knowledge about active connections. I'd make an educated guess that this decision is at the bottom of all this. That's for example the reason, the rude shutdown method of the loop exists in the first place, because there is nothing to properly shut down active connections for example on the `Frame` level. Instead it's done at the socket level.

benaadams:
@Tragetaschen not really, the problem is caused by Kestrel being well behaved; and the client being local and 0 ms away.

So Kestrel waits for the "socket shutdown" message to be "on the wire" before tearing down the socket, otherwise the client will never receive it and will remain in a connected state; however the client receives that message instantly so initiates its own shutdown, then libuv cleans up all before the confirm of the message being on the wire comes back; at which point Kestrel then tries to dispose the socket and fails because everything is already gone.

benaadams:
@Tragetaschen also looks like your other issue is addressed by this: `IHttpRequestLifetimeFeature` https://github.com/aspnet/KestrelHttpServer/pull/325

halter73:
If you take my `AsyncCanBeSent` test case and replace `trigger.Dispose()` with `trigger.DangerousClose()` like we call in `KestrelThread.ThreadStart` while tearing down, the AV repros immediately.

I think the call to `DangerousClose` was meant to prevent ODE's from being thrown from `KestrelThread.Stop`, but the way it is currently written there will always be a race condition in that code.

I think for now it's best to catch the potential ODEs in `KestrelThread.Stop` and simply call `UvAsyncHandle.Dispose` inside of `KestrelThread.ThreadStart`.

We can then get rid of `UvAsyncHandle.DangerousClose`. It's way too dangerous for me :fearful: 

@lodejard @davidfowl @benaadams What do you all think?

Here's my first attempt at a PR: https://github.com/aspnet/KestrelHttpServer/pull/347

benaadams:
Better fix by @halter73 in #347, closing this

PR #343, opened 2015-11-08
Dispose of MemoryPool2; suppress its finalizers

benaadams:
This is to unpin the MemoryPoolSlab2 buffers and SuppressFinalize on all the MemoryPoolBlock2 objects

### DIFF
--- a/test/Microsoft.AspNet.Server.KestrelTests/TestInput.cs
+++ b/test/Microsoft.AspNet.Server.KestrelTests/TestInput.cs
@@ -9,19 +9,26 @@ using Microsoft.AspNet.Server.Kestrel.Infrastructure;
 
 namespace Microsoft.AspNet.Server.KestrelTests
 {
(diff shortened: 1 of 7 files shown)


benaadams:
Was rebased on #337 to reduce probability of CI failures

test/Microsoft.AspNet.Server.KestrelTests/TestInput.cs
```diff
@@ -9,21 +9,28 @@
 
 namespace Microsoft.AspNet.Server.KestrelTests
 {
-    class TestInput : IConnectionControl, IFrameControl
```
halter73: Doesn't this require change MessageBodyTests?

benaadams: Added

benaadams:
This is probably better #402

PR #350, opened 2015-11-10
Remove sync block->task

### LINKED ISSUE #349: Blocking call to ProduceStart in UpgradeAsync?
https://github.com/aspnet/KestrelHttpServer/blob/dev/src/Microsoft.AspNet.Server.Kestrel/Http/Frame.FeatureCollection.cs#L293

/cc @lodejard @halter73

benaadams:
await rather than synchronously blocking and then returning a Task.FromResult

Resolves  #349

### DIFF
--- a/src/Microsoft.AspNet.Server.Kestrel/Http/Frame.FeatureCollection.cs
+++ b/src/Microsoft.AspNet.Server.Kestrel/Http/Frame.FeatureCollection.cs
@@ -277,7 +277,7 @@ namespace Microsoft.AspNet.Server.Kestrel.Http
             OnCompleted(callback, state);
         }
 
-        Task<Stream> IHttpUpgradeFeature.UpgradeAsync()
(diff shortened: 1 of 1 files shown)


benaadams:
Feeds are updating; will resubmit after

NickCraver:
+1 - tiny note: failures are now AggregateExceptions here due to the change. Honestly though, due to what's underneath that ship long since sailed (and isn't a bad thing - just noting). 

Only tangentially related: I noticed some related issues when looking at this, e.g.:

```C#
public async Task FlushAsync(CancellationToken cancellationToken)
{
    await ProduceStartAndFireOnStarting(immediate: false);
    await SocketOutput.WriteAsync(_emptyData, immediate: true);
}
```
I'm wondering why the token isn't passed into `SocketOutput.WriteAsync()` (which has a cancellation token overload). I haven't looked at this before tonight - so don't want to make assumptions on reasoning. Given that WriteAsync does just below...I'm assuming this is an oversight.

PR #352, opened 2015-11-10
Fix get IPAddress for Darwin

troydai:
The `sockaddr` on Darwin is different from the one on Windows and Linux, which uses a 8-bit integer as the first field representing length.
``` C++
struct sockaddr {
	__uint8_t	sa_len;		/* total length */
	sa_family_t	sa_family;	/* [XSI] address family */
	char		sa_data[14];	/* [XSI] addr value (actually larger) */
};
```
Source: [socket.h](http://www.opensource.apple.com/source/xnu/xnu-1456.1.26/bsd/sys/socket.h)

### DIFF
--- a/src/Microsoft.AspNet.Server.Kestrel/Networking/SockAddr.cs
+++ b/src/Microsoft.AspNet.Server.Kestrel/Networking/SockAddr.cs
@@ -53,6 +53,11 @@ namespace Microsoft.AspNet.Server.Kestrel.Networking
             // 0000 0000 0000 0000     the second unint8 field for sa family type
             // 0000 0000 0000 0000     http://www.opensource.apple.com/source/xnu/xnu-1456.1.26/bsd/sys/socket.h
             // 0000 0000 0000 0000
+            //
+            // Reference:
+            //  - Windows: https://msdn.microsoft.com/en-us/library/windows/desktop/ms740506(v=vs.85).aspx
+            //  - Linux: https://github.com/torvalds/linux/blob/master/include/linux/socket.h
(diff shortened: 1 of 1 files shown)


Tratcher:
:cry: :shipit:

src/Microsoft.AspNet.Server.Kestrel/Networking/SockAddr.cs
```diff
@@ -41,18 +41,37 @@ public IPEndPoint GetIPEndPoint()
...
+            // Example 4: 127.0.0.1:52798, on a Mac OS
+            //
+            // 0100 007F 3ECE 0210  => sa_family == AF_INET (02) Note that struct sockaddr on mac use
+            // 0000 0000 0000 0000     the second unint8 field for sa family type
+            // 0000 0000 0000 0000     http://www.opensource.apple.com/source/xnu/xnu-1456.1.26/bsd/sys/socket.h
+            // 0000 0000 0000 0000
+            //
+            // Reference:
+            //  - Windows: https://msdn.microsoft.com/en-us/library/windows/desktop/ms740506(v=vs.85).aspx
+            //  - Linux: https://github.com/torvalds/linux/blob/master/include/linux/socket.h
```
akoeplinger: linking to master is not future proof, better to use a permalink like https://github.com/torvalds/linux/blob/6a13feb9c82803e2b815eca72fa7a9f5561d7861/include/linux/socket.h

troydai: Thanks for the advice. I'll update the comments.

PR #353, opened 2015-11-10
SocketOutput to Tasks

benaadams:
Use Task, CompletedTask and TaskCompletionSource instead of CallbackContext

Removes all of the CallbackContext creations; cuts down on callbacks. Also creates less TaskCompletionSources

### DIFF
--- a/src/Microsoft.AspNet.Server.Kestrel/Http/SocketOutput.cs
+++ b/src/Microsoft.AspNet.Server.Kestrel/Http/SocketOutput.cs
@@ -31,7 +31,7 @@ namespace Microsoft.AspNet.Server.Kestrel.Http
         private int _numBytesPreCompleted = 0;
         private Exception _lastWriteError;
         private WriteContext _nextWriteContext;
-        private readonly Queue<CallbackContext> _callbacksPending;
+        private readonly Queue<TaskCompletionSource<object>> _tasksPending;
 
         public SocketOutput(KestrelThread thread, UvStreamHandle socket, long connectionId, IKestrelTrace log)
         {
@@ -39,13 +39,11 @@ namespace Microsoft.AspNet.Server.Kestrel.Http
             _socket = socket;
             _connectionId = connectionId;
             _log = log;
-            _callbacksPending = new Queue<CallbackContext>();
+            _tasksPending = new Queue<TaskCompletionSource<object>>();
         }
 
-        public void Write(
+        public Task WriteAsync(
             ArraySegment<byte> buffer,
-            Action<Exception, object, bool> callback,
-            object state,
             bool immediate = true,
             bool socketShutdownSend = false,
             bool socketDisconnect = false)
@@ -58,8 +56,8 @@ namespace Microsoft.AspNet.Server.Kestrel.Http
                 buffer = new ArraySegment<byte>(copy);
                 _log.ConnectionWrite(_connectionId, buffer.Count);
             }
-
-            bool triggerCallbackNow = false;
+            
+            TaskCompletionSource<object> tcs = null;
 
             lock (_lockObj)
             {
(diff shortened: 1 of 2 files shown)


src/Microsoft.AspNet.Server.Kestrel/Http/SocketOutput.cs
```diff
@@ -80,23 +78,25 @@ public SocketOutput(KestrelThread thread, UvStreamHandle socket, long connection
...
                 else
                 {
-                    _callbacksPending.Enqueue(new CallbackContext
-                    {
-                        Callback = callback,
-                        State = state,
-                        BytesToWrite = buffer.Count
-                    });
+                    // immediate==false calls always return complete tasks, because there is guaranteed
+                    // to be a subsequent immediate==true call which will go down one of the previous code-paths
```
halter73: I think this is a logic bug. Since this write isn't being tracked in the `_tasksPending` queue, you need to consider this write as pre-completed (`_numBytesPreCompleted += buffer.Count`).

This bug could cause `_numBytesPreCompleted` to become negative leading to assertion failures in debug builds. In release builds, this could delay write throttling.

I think our existing SocketOutputTests don't catch this bug because they were written before the `immediate` parameter was introduced. As a result, the tests always use the default value for `immediate` which is true.

Once you fix this, could you add a regression test to SocketOutputTests?

halter73:
Gotta love PR's with more deletions than additions :smile: I'm particularly happy to get rid of that `calledInline` parameter that the write callback had.

:shipit: Once the pre-completed bytes thing is fixed.

src/Microsoft.AspNet.Server.Kestrel/Http/SocketOutput.cs
benaadams: But the travis re-run pain! Sure...

benaadams: > In release builds, this could delay write throttling.

Might need to fiddle with the buffer sizes after this then

halter73: I hear you. I wish @davidfowl would ease up a little on requiring travis to pass given the current state of Kestrel on Mono. For now, it's just motivation for me to fix the timeouts :persevere:

halter73: > Might need to fiddle with the buffer sizes after this then

Why is that? What do you intend to change?

benaadams: If its currently overwriting to the socket; the profile may change if it throttles earlier - may be good, may be bad, may not effect anything :)

halter73: Once you fix this, it will start throttling at the same point it did before the PR. Did you already profile with these changes?

benaadams: I've discovered my profiling (in t'Cloud) is slightly "time of day" effected; so not great at testing smaller changes :(

And larger changes aren't great for PRs; so trying to break the larger one into chunks. Will paste some memory profiles while waiting for Travis to work...

benaadams:
It passed! So numbers...

4000 requests:

Before bytes 2,000,384

before-bytes

After bytes 1,488,000 (75% of before)

after-bytes

Before allocations 32,006

before-allocs

After allocations 20,000 (62% of before)

after-allocs

test/Microsoft.AspNet.Server.KestrelTests/SocketOutputTests.cs
```diff
@@ -115,6 +115,82 @@ public void WritesDontCompleteImmediatelyWhenTooManyBytesAreAlreadyPreCompleted(
...
+                // Assert
+                // The third write should pre-complete since it is not immediate, even though too many.
+                Assert.True(completedWh.Wait(1000));
+                // Arrange
+                completedWh.Reset();
+                // Act
+                socketOutput.WriteAsync(fullBuffer).ContinueWith(onCompleted);
+                // Assert 
+                // Too many bytes are already pre-completed for the fourth write to pre-complete.
+                Assert.False(completedWh.Wait(1000));
```
halter73: I think this test would pass even if you aren't counting non-immediate writes as pre-completed.

Since you've already written halfBuffer normally without completing it yet, I would expect writing the fullBuffer would induce throttling in either case.

src/Microsoft.AspNet.Server.Kestrel/Http/SocketOutput.cs
```diff
@@ -198,20 +195,30 @@ private void OnWriteCompleted(Queue<ArraySegment<byte>> writtenBuffers, int stat
...
+                    else
+                    {
+                        // error is closure captured 
+                        ThreadPool.QueueUserWorkItem(
+                            (o) => ((TaskCompletionSource<object>)o).SetException(error), 
+                            tcs);
+                    }
                 }
 
                 // Now that the while loop has completed the following invariants should hold true:
```
halter73: It's hidden in this PR, but I don't think `Debug.Assert(_numBytesPreCompleted <= _maxBytesPreCompleted);` is still valid since we can now exceed that given non-immediate writes.

I would just remove that assertion.

benaadams: Removed

benaadams:
Phew! :dancer:

halter73:
:sunglasses:

Issue #356, opened 2015-11-10
SocketOutput sometimes pre-completes too many writes which breaks throttling


halter73:
Given a bunch of pending write operations, once the first pending write completes, all the subsequent pending writes that are small enough will pre-complete. Given the right circumstances, it is possible to pre-complete far more than the expected 64KB worth of writes.

The reason this likely wasn't noticed until now, is because it would require multiple unawaited writes to the response stream to reproduce. It's still a bug though.

PR #357, opened 2015-11-11
Don't pre-complete too many writes

halter73:
#356

### DIFF
--- a/test/Microsoft.AspNet.Server.KestrelTests/SocketOutputTests.cs
+++ b/test/Microsoft.AspNet.Server.KestrelTests/SocketOutputTests.cs
@@ -192,6 +192,84 @@ namespace Microsoft.AspNet.Server.KestrelTests
             }
         }
 
+        [Fact]
+        public void WritesDontGetCompletedTooQuickly()
+        {
+            // This should match _maxBytesPreCompleted in SocketOutput
+            var maxBytesPreCompleted = 65536;
+            var completeQueue = new Queue<Action<int>>();
+            var onWriteWh = new ManualResetEventSlim();
+
+            // Arrange
+            var mockLibuv = new MockLibuv
+            {
+                OnWrite = (socket, buffers, triggerCompleted) =>
+                {
+                    completeQueue.Enqueue(triggerCompleted);
+                    onWriteWh.Set();
+
+                    return 0;
+                }
+            };
+
+            using (var kestrelEngine = new KestrelEngine(mockLibuv, new TestServiceContext()))
+            {
+                kestrelEngine.Start(count: 1);
+
+                var kestrelThread = kestrelEngine.Threads[0];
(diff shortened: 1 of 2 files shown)


test/Microsoft.AspNet.Server.KestrelTests/SocketOutputTests.cs
```diff
@@ -192,6 +192,83 @@ public void WritesDontCompleteImmediatelyWhenTooManyBytesIncludingNonImmediateAr
             }
         }
 
+        [Fact]
+        public void WritesDontGetCompletedTooQuickly()
+        {
+            // This should match _maxBytesPreCompleted in SocketOutput
+            var maxBytesPreCompleted = 65536;
```
cesarbs: nit: 64 * 1024 makes it more clear it's 64KB.

test/Microsoft.AspNet.Server.KestrelTests/SocketOutputTests.cs
```diff
@@ -192,6 +192,83 @@ public void WritesDontCompleteImmediatelyWhenTooManyBytesIncludingNonImmediateAr
             }
         }
 
+        [Fact]
+        public void WritesDontGetCompletedTooQuickly()
```
cesarbs: You'll probably have to skip this test on Mono (https://github.com/aspnet/KestrelHttpServer/pull/373).

test/Microsoft.AspNet.Server.KestrelTests/SocketOutputTests.cs
```diff
@@ -192,6 +192,83 @@ public void WritesDontCompleteImmediatelyWhenTooManyBytesIncludingNonImmediateAr
...
+                socketOutput.WriteAsync(buffer).ContinueWith(onCompleted);
+                // Assert
+                // The first write should pre-complete since it is <= _maxBytesPreCompleted.
+                Assert.True(completedWh.Wait(1000));
+                Assert.True(onWriteWh.Wait(1000));
+                // Arrange
+                completedWh.Reset();
+                onWriteWh.Reset();
+
+                // Act
```
cesarbs: Break this into multiple tests so you act, assert and arrange only once in each one.

test/Microsoft.AspNet.Server.KestrelTests/SocketOutputTests.cs
halter73: This uses `MockLibuv`, so it should be fine.

test/Microsoft.AspNet.Server.KestrelTests/SocketOutputTests.cs
halter73: I'm not going to split this into multiple tests, but these comments aren't as clear as the could be.

The first act/assert/arrange is really still setup with some sanity checking.

Issue #338, opened 2015-11-06
Travis jamms/hangs


benaadams:
Frequently Travis will jam when running tests; this is between tests before moving onto the next exec step e.g.

> === TEST EXECUTION SUMMARY ===
   Microsoft.AspNet.Server.KestrelTests  Total: 218, Errors: 0, Failed: 0, Skipped: 4, Time: 19.726s
No output has been received in the last 10 minutes, this potentially indicates a stalled build or something wrong with the build itself.
The build has been terminated

benaadams:
Multiple PRs is unrelated :(

benaadams:
For reference next lines should be
> info: Exec
info:   program: cmd
info:   commandline: /C dnx test

benaadams:
Looks like this might be fixed in mono 4.2.0

akoeplinger:
You can try out 4.2 by adding `mono: alpha` to .travis.yml :)

dougbu:
Not recommended since Mono 4.2 has other issues at least when used with Ubuntu 12.04. I will soon migrate our Travis configuration to use the new Ubuntu 14.04 (Trusty) systems. We should look at what works and what doesn't (with stable Mono releases) after that migration.

akoeplinger:
@dougbu I'd be interested to hear about the 12.04 vs 14.04 issues? Mono 4.2 will be released very soon.

dougbu:
For example https://bugzilla.xamarin.com/show_bug.cgi?id=30018 was only fixed in Mono 4.3 and it breaks `dnu restore` fairly reliably on Ubuntu 14.04.

Further we see significant MVC test regressions when running with the Mono Alpha feed.

Another push toward Ubuntu 14.04 is the fact .NET Core has targeted that specific release.

akoeplinger:
> For example https://bugzilla.xamarin.com/show_bug.cgi?id=30018 was only fixed in Mono 4.3 and it breaks dnu restore fairly reliably on Ubuntu 14.04.

I'll make sure this gets into the first service release for Mono 4.2

> Further we see significant MVC test regressions when running with the Mono Alpha feed.

That's unfortunate... MVC would need to stay on 4.0 for the time being then.

dougbu:
> MVC would need to stay on 4.0 for the time being then.

To be clear: That was an example of problems we've seen w/ more recent Mono builds than 4.0.4.

[...]

PR #358, opened 2015-11-11
Pass cancellation tokens down consistently

NickCraver:
I'm assuming these are oversights since the usage is inconsistent, but it's possible there's a reason here. If there is, this PR serves only as a mild waste of time - in which case: sorry!

### DIFF
--- a/src/Microsoft.AspNet.Server.Kestrel/Http/FrameRequestStream.cs
+++ b/src/Microsoft.AspNet.Server.Kestrel/Http/FrameRequestStream.cs
@@ -68,14 +68,8 @@ namespace Microsoft.AspNet.Server.Kestrel.Http
         {
             return ((Task<int>)asyncResult).Result;
         }
-#endif
 
-        public override Task<int> ReadAsync(byte[] buffer, int offset, int count, CancellationToken cancellationToken)
-        {
-            return _body.ReadAsync(new ArraySegment<byte>(buffer, offset, count), cancellationToken);
-        }
-
-        public Task<int> ReadAsync(byte[] buffer, int offset, int count, CancellationToken cancellationToken, object state)
+        private Task<int> ReadAsync(byte[] buffer, int offset, int count, CancellationToken cancellationToken, object state)
         {
             var tcs = new TaskCompletionSource<int>(state);
             var task = _body.ReadAsync(new ArraySegment<byte>(buffer, offset, count), cancellationToken);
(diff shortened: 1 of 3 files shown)


src/Microsoft.AspNet.Server.Kestrel/Http/FrameRequestStream.cs
```diff
@@ -94,7 +94,7 @@ public Task<int> ReadAsync(byte[] buffer, int offset, int count, CancellationTok
                 {
                     tcs2.SetResult(task2.Result);
                 }
-            }, tcs);
+            }, tcs, cancellationToken);
             return tcs.Task;
```
benaadams: I think this entire method can be deleted; it isn't used and it isn't an override of Stream.

Tratcher: Isn't it used to implement BeginRead?

benaadams: Ah yeah, make it private then; also it passes in `CancellationToken.None`, no harm in passing it along though.

NickCraver: I'm not against `private` here - but in that case, it should just be moved into the `#if NET451` altogether IMO - is anyone external possibly depending on this?

benaadams: Is kinda internal; hence why the suggestion; before anyone takes a dependency on it...

NickCraver: I'm all for it - @Tratcher, if you're +1 I'll commit in.

halter73: :+1: for making this private.

Tratcher: private and  #if NET451

NickCraver: done

NickCraver:
Possibly blind here - not finding the fail in the Travis CI log, what am I missing?

benaadams:
mono < 4.2 (in alpha) is grumpy particularly with this repository https://github.com/aspnet/KestrelHttpServer/issues/338

Normally have to do several "add whitespace"->"fixup"->"force push"->"fail"->"remove whitespace"->"fixup"->"force push"->"fail"->repeat cycles till it passes.

NickCraver:
Well that's kinda crap - any chance we can clean that up immensely and allow pinging to trigger via @dotnet-bot?

PR #360, opened 2015-11-11
.Result -> .GetAwaiter().GetResult()

benaadams:


davidfowl:
I forgot to reply, you missed EndRead

benaadams:
Added  #377 for EndRead

PR #361, opened 2015-11-11
Trim exposed KestrelThread api

benaadams:
object based PostAsync can be removed
object based Post can be made private and strongly typed


src/Microsoft.AspNet.Server.Kestrel/Infrastructure/KestrelThread.cs
```diff
@@ -64,15 +64,15 @@ public void Stop(TimeSpan timeout)
                 return;
             }
 
-            Post(OnStop, null);
```
halter73: I would prefer making this PR a little small by changing this to:

    Post(t => t.OnStop());

Then the OnStop* methods don't need to be changed.

benaadams:
Done

PR #362, opened 2015-11-11
Partitioned memory pools

benaadams:
The increase in use of memory pools is leading to and will inevitably lead to greater and greater contention.

Contention

This loosens the contention on the memorypool by partitioning `ProcessorCount` times and choosing the first pool to try as `ManagedThreadId % ProcessorCount`.

Also steals from other pools when preferred pool is empty.

### DIFF
--- a/src/Microsoft.AspNet.Server.Kestrel/Http/ListenerContext.cs
+++ b/src/Microsoft.AspNet.Server.Kestrel/Http/ListenerContext.cs
@@ -10,13 +10,11 @@ namespace Microsoft.AspNet.Server.Kestrel.Http
     {
         public ListenerContext()
         {
-            Memory2 = new MemoryPool2();
         }
 
         public ListenerContext(ServiceContext serviceContext) 
             : base(serviceContext)
         {
-            Memory2 = new MemoryPool2();
         }
 
         public ListenerContext(ListenerContext listenerContext)
@@ -25,7 +23,6 @@ namespace Microsoft.AspNet.Server.Kestrel.Http
             ServerAddress = listenerContext.ServerAddress;
             Thread = listenerContext.Thread;
             Application = listenerContext.Application;
-            Memory2 = listenerContext.Memory2;
             Log = listenerContext.Log;
         }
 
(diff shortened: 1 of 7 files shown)


benaadams:
I was aiming for a reduction in operating memory which didn't happen; however it did bump rps instead

from

>pts/pipeline.lua -- 16
Running 30s test @ http://...
    Latency     7.72ms   18.68ms 317.79ms   94.45%
    Req/Sec    22.76k     3.24k   56.97k    80.32%
  21528335 requests in 30.10s, 2.65GB read
Requests/sec: 715261.68
Transfer/sec:     90.04MB

to

> pts/pipeline.lua -- 16
Running 30s test @ http://...
    Latency     8.03ms   20.19ms 295.46ms   93.90%
    Req/Sec    24.52k     3.35k   46.80k    79.64%
  23504876 requests in 30.10s, 2.89GB read
Requests/sec: 780945.45
Transfer/sec:     98.31MB

YMMV

benaadams:
Increasing the allocation block size from 32 to 128 causes the performance to drop back to previous levels.

davidfowl:
Not a fan of this being in the ctor. It could be in a factory method though

benaadams:
Going to change it up then...

benaadams:
@halter73 give this a go? 

Taken @rynowak's round robining idea from https://github.com/aspnet/KestrelHttpServer/pull/390#issuecomment-157203190; but using `ManagedThreadId % ProcessorCount` rather than `Interlocked.Increment` which would put contention on a single numeric; though lighter than CAS. Hopefully it will still provide enough distribution.

DamianEdwards:
Rebased on 7e8a405 and measured. This one seems totally broken. It gets slower with each run and leaves the dnx process eating CPU even when not serving requests.

benaadams:
Yeah found it to be broken last night ;-(

benaadams:
@DamianEdwards is fixed

AspNetSmurfLab:
Measured again. It's very slightly up from the `dev` branch baseline (~499k):
```
    Latency     6.30ms    7.52ms 100.86ms   92.93%
    Req/Sec    15.84k     1.39k   32.81k    88.84%
  5071775 requests in 10.10s, 638.46MB read
Requests/sec: 502177.87
Transfer/sec:     63.22MB
```

benaadams:
This is more a pre-solution for the contention that will happen with more memory pool use with SocketOutput and the pre-computed Headers change

benaadams:
See @halter73 is working on something far better; closing this...

PR #363, opened 2015-11-11
Resuse writes, initalize queues

### LINKED ISSUE #165: Pool UvWriteReq objects
It might help.

### LINKED ISSUE #289: [Perf] High Allocator DoWriteIfNeeded
Drill down on Issue #288
Measurements post PR #287

Item 5 on highest allocators in plaintext test

most-memory

Allocations (2003 requests)

alloc

alloc

Writes alloc on List.Add not `GCHandle.Alloc` for clarity

Bytes (2003 requests)

bytes

bytes

Writes alloc on List.Add not `GCHandle.Alloc` for clarity

### LINKED ISSUE #290: [Perf] High Allocator SocketOutput.Write
Drill down on Issue #288
Measurements post PR #287

Item 2 on highest allocators in plaintext test

most-memory

`Microsoft.AspNet.Server.Kestrel.Http.SocketOutput.Write`

Allocations (2003 requests)

allocations

Bytes (2003 requests)

bytes

benaadams:
Reuse UvWriteReq on the socket rather than recreating for each write - pooled at per connection level.

Pre initialize the circular buffers; rather than resizing as work is added; with a copy on resize.

Saves 1.12 MB per 4k requests; see https://github.com/aspnet/KestrelHttpServer/pull/363#issuecomment-159145892 for details.

Resolves #165 Pool UvWriteReq objects
Resolves #289 High Allocator DoWriteIfNeeded
Resolves #290 High Allocator SocketOutput.Write

### DIFF
--- a/src/Microsoft.AspNet.Server.Kestrel/Http/SocketOutput.cs
+++ b/src/Microsoft.AspNet.Server.Kestrel/Http/SocketOutput.cs
@@ -17,6 +17,7 @@ namespace Microsoft.AspNet.Server.Kestrel.Http
         private const int _maxPendingWrites = 3;
         private const int _maxBytesPreCompleted = 65536;
         private const int _initialTaskQueues = 64;
+        private const int _maxPooledWriteContexts = 32;
 
         private static WaitCallback _returnBlocks = (state) => ReturnBlocks((MemoryPoolBlock2)state);
         private static MemoryPoolIterator2 _defaultIterator;
@@ -39,6 +40,7 @@ namespace Microsoft.AspNet.Server.Kestrel.Http
 
         // This locks access to to all of the below fields
         private readonly object _contextLock = new object();
+        private bool _isDisposed = false;
 
         // The number of write operations that have been scheduled so far
         // but have not completed.
@@ -49,6 +51,7 @@ namespace Microsoft.AspNet.Server.Kestrel.Http
         private WriteContext _nextWriteContext;
         private readonly Queue<TaskCompletionSource<object>> _tasksPending;
         private readonly Queue<TaskCompletionSource<object>> _tasksCompleted;
+        private readonly Queue<WriteContext> _writeContextPool;
 
         public SocketOutput(
             KestrelThread thread,
@@ -67,6 +70,7 @@ namespace Microsoft.AspNet.Server.Kestrel.Http
             _threadPool = threadPool;
             _tasksPending = new Queue<TaskCompletionSource<object>>(_initialTaskQueues);
             _tasksCompleted = new Queue<TaskCompletionSource<object>>(_initialTaskQueues);
+            _writeContextPool = new Queue<WriteContext>(_maxPooledWriteContexts);
 
             _head = memory.Lease();
             _tail = _head;
@@ -93,7 +97,14 @@ namespace Microsoft.AspNet.Server.Kestrel.Http
             {
                 if (_nextWriteContext == null)
                 {
-                    _nextWriteContext = new WriteContext(this);
+                    if (_writeContextPool.Count > 0)
+                    {
+                        _nextWriteContext = _writeContextPool.Dequeue();
+                    }
+                    else
+                    {
+                        _nextWriteContext = new WriteContext(this);
+                    }
                 }
 
                 if (socketShutdownSend)
@@ -275,9 +286,12 @@ namespace Microsoft.AspNet.Server.Kestrel.Http
         }
 
         // This is called on the libuv event loop
-        private void OnWriteCompleted(int bytesWritten, int status, Exception error)
+        private void OnWriteCompleted(WriteContext writeContext)
         {
-            _log.ConnectionWriteCallback(_connectionId, status);
+            var bytesWritten = writeContext.ByteCount;
+            var status = writeContext.WriteStatus;
+            var error = writeContext.WriteError;
+
 
             if (error != null)
             {
@@ -291,6 +305,7 @@ namespace Microsoft.AspNet.Server.Kestrel.Http
 
             lock (_contextLock)
             {
+                PoolWriteContext(writeContext);
                 if (_nextWriteContext != null)
                 {
                     scheduleWrite = true;
@@ -333,11 +348,11 @@ namespace Microsoft.AspNet.Server.Kestrel.Http
                 }
             }
 
+            _log.ConnectionWriteCallback(_connectionId, status);
+
             if (scheduleWrite)
             {
-                // ScheduleWrite();
-                // on right thread, fairness issues?
-                WriteAllPending();
+                ScheduleWrite();
             }
 
             _tasksCompleted.Clear();
@@ -365,6 +380,32 @@ namespace Microsoft.AspNet.Server.Kestrel.Http
 
                 _head = null;
                 _tail = null;
+            }
+        }
+
+        private void PoolWriteContext(WriteContext writeContext)
+        {
+            // called inside _contextLock
+            if (!_isDisposed && _writeContextPool.Count < _maxPooledWriteContexts)
+            {
+                writeContext.Reset();
+                _writeContextPool.Enqueue(writeContext);
+            }
+            else
+            {
+                writeContext.Dispose();
+            }
+        }
+
+        private void Dispose()
+        {
+            lock (_contextLock)
+            {
+                _isDisposed = true;
+                while (_writeContextPool.Count > 0)
+                {
+                    _writeContextPool.Dequeue().Dispose();
+                }
             }
         }
 
@@ -409,14 +450,14 @@ namespace Microsoft.AspNet.Server.Kestrel.Http
             buffers++;
         }
 
-        private class WriteContext
+        private class WriteContext : IDisposable
         {
             private static WaitCallback _returnWrittenBlocks = (state) => ReturnWrittenBlocks((MemoryPoolBlock2)state);
 
             private MemoryPoolIterator2 _lockedStart;
             private MemoryPoolIterator2 _lockedEnd;
             private int _bufferCount;
-            private int _byteCount;
+            public int ByteCount;
 
             public SocketOutput Self;
 
@@ -426,11 +467,15 @@ namespace Microsoft.AspNet.Server.Kestrel.Http
             public int WriteStatus;
             public Exception WriteError;
 
+            private UvWriteReq _writeReq;
+
             public int ShutdownSendStatus;
 
             public WriteContext(SocketOutput self)
             {
                 Self = self;
+                _writeReq = new UvWriteReq(Self._log);
+                _writeReq.Init(Self._thread.Loop);
             }
 
             /// <summary>
@@ -440,18 +485,19 @@ namespace Microsoft.AspNet.Server.Kestrel.Http
             {
                 LockWrite();
 
-                if (_byteCount == 0 || Self._socket.IsClosed)
+                if (ByteCount == 0 || Self._socket.IsClosed)
                 {
                     DoShutdownIfNeeded();
                     return;
                 }
 
-                var writeReq = new UvWriteReq(Self._log);
-                writeReq.Init(Self._thread.Loop);
+                // Sample values locally in case write completes inline
+                // to allow block to be Reset and still complete this function
+                var lockedEndBlock = _lockedEnd.Block;
+                var lockedEndIndex = _lockedEnd.Index;
 
-                writeReq.Write(Self._socket, _lockedStart, _lockedEnd, _bufferCount, (_writeReq, status, error, state) =>
+                _writeReq.Write(Self._socket, _lockedStart, _lockedEnd, _bufferCount, (_writeReq, status, error, state) =>
                 {
-                    _writeReq.Dispose();
                     var _this = (WriteContext)state;
                     _this.ScheduleReturnFullyWrittenBlocks();
                     _this.WriteStatus = status;
@@ -459,8 +505,8 @@ namespace Microsoft.AspNet.Server.Kestrel.Http
                     _this.DoShutdownIfNeeded();
                 }, this);
 
-                Self._head = _lockedEnd.Block;
-                Self._head.Start = _lockedEnd.Index;
+                Self._head = lockedEndBlock;
+                Self._head.Start = lockedEndIndex;
             }
 
             /// <summary>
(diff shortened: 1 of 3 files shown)


benaadams:
meh

halter73:
@benaadams When profiling, we see a fair amount of time spent resizing queues (though this is mostly in SocketOutput). Why did you close this PR?

benaadams:
I didn't feel it was worthwhile to go round the Travis repeat loop for to try to get to pass; will rebase.

benaadams:
Rebased and added SocketOutput

benaadams:
Allocation reductions for 8k requests

Before 32,023 objects; 2,032,968 bytes

before

After 12,059 objects; 947,168 bytes

after

37% of the objects than before; 46% of the bytes of before

halter73:
I think @cesarbs is pretty close to fixing our travis woes. I wouldn't bother going through the travis repeat loop until that gets in. Sorry that it's been such a pain for you.

I should be able kick off reruns for all failing PRs myself once that's in.

benaadams:
Removed the allocs from `DoWriteIfNeeded` so down by another 212800

benaadams:
And from UvWriteReq.Write

benaadams:
Boom! Use `MemoryPool2` rather than `new byte[buffer.Count]` in write and `SocketOutput` alloc free

DamianEdwards:
Rebased on 7e8a405 and measured. Basically the same. I really need to profile it to see if the size we're setting them to is high enough.

```
dev (7e8a405) - 499k
363 (2a32d33) - 501k (+0.4%)
```

AspNetSmurfLab:
Still spending some time in resizing it seems:
image

benaadams:
@DamianEdwards that's quite a nice tool; what are you using? I'm have a hard time measuring GC with core; but 44.5% still ouch!

mattwarren:
@benaadams I'm pretty sure that screenshot is from [JetBrains DotTrace](https://www.google.co.uk/search?q=jetbrains+dottrace+screenshots+3.5&espv=2&biw=1257&bih=929&source=lnms&tbm=isch&sa=X&ved=0CAcQ_AUoAmoVChMIscS6qfWZyQIVxAcsCh0GdAVY&dpr=1#tbm=isch&q=jetbrains+dottrace+screenshot&imgrc=6596Vzm1wXxoAM%3A)

khellang:
Yep, looks like and older version of dotTrace :smile:

benaadams:
With 45% GC load though all bets are off on algorithmic improvements; need to get that overhead down first.

mattwarren:
Out of interest, which version of the GC were these tests run under? Server, Client, Background Server, etc?

benaadams:
@DamianEdwards  could you retest? Obv resizing the queues will add to GC pressure.

Have increased the queues from and array of 16 pointers to an array of 64 pointers - shouldn't cause much difference in allocation expense i.e. objects allocated is the same and they are reused.

AspNetSmurfLab:
Good results, nearly 10% improvement:
```
    Latency     4.17ms    2.70ms 181.45ms   74.48%
    Req/Sec    17.22k     1.20k   29.55k    85.90%
  5518027 requests in 10.10s, 694.64MB read
Requests/sec: 546339.52
Transfer/sec:     68.78MB
```

DamianEdwards:
@halter73 @davidfowl @loudej pls review

davidfowl:
This change isn't just queue sizes it's pooling he write context and it will conflict with the socket output changes

benaadams:
Yes this needs revisiting post socket output

benaadams:
Update!

Rebased; renamed, changed to work with the SocketOutput changes. Results as follows....

Before WriteAsync bytes used 352,000 bytes

write-async-before

After WriteAsync bytes used 888 bytes; 888 bytes? Crazy town!

write-async-after

Before DoWriteIfNeeded 768,000 bytes

DoWrite

After DoWriteIfNeeded doesn't use any bytes at all, so I have nothing to screenshot O_o

So instead, the full SocketOutput stack:

Before

socket-output-before

After

socket-output-after

The mind boggles... :dancer:

src/Microsoft.AspNet.Server.Kestrel/Http/SocketOutput.cs
```diff
@@ -440,7 +481,11 @@ public void DoWriteIfNeeded()
                 }, this);
 
                 Self._head = _lockedEnd.Block;
-                Self._head.Start = _lockedEnd.Index;
```
benaadams: Discovered race here; don't know if was with these changes or is in current build.

halter73: Nice catch.

If you just want to avoid the NullRefEx, you should do the following:

    var newHead = _lockedEnd.Block;
    newHead.Start = _lockedEnd.Index;

    if (Self._head != null)
    {
        Self._head = newHead;
    }

With the current change, you could still get a null ref if `_head` is set to null between the if comparison and the actual assignment to `Self._head.Start`. This makes the race a lot tighter than it was before, but it's still there.

To be safe, all modifications to _head should be made with the `_returnLock`. This fixes the race because `_head` is only set to null when the `_returnLock` is taken.

src/Microsoft.AspNet.Server.Kestrel/Http/SocketOutput.cs
```diff
@@ -370,6 +385,32 @@ private void ReturnAllBlocks()
...
+                writeContext.Reset();
+                _writeContextPool.Enqueue(writeContext);
+            }
+            else
+            {
+                writeContext.Dispose();
+            }
+        }
+
+        public void Dispose()
```
halter73: Who calls this? Is SocketOutput disposable?

benaadams: No, probably should be private? (Or renamed) - `DoDisconnectIfNeeded` calls it.

src/Microsoft.AspNet.Server.Kestrel/Http/SocketOutput.cs
halter73: **Wait a minute!** I think this race must have been caused by a change in this PR

I was wondering how I didn't see this was a race when I originally wrote this code, and now I'm pretty sure it's because `_head` is only ever supposed to modified in the SocketOutput ctor and on the libuv thread. If that's the case, this race should be impossible.

@benaadams Do you have any idea how `_head` would be modified off of the libuv thread with this change? I'm not seeing it yet.

benaadams: All I can think is the `Write` callback actually completes before the next lines are run? Was going to move the set above the `Write` but didn't feel comfortable enough with the code.

halter73: @benaadams You don't want to move the set above the write in case write throws. If that's the case, we still want to ensure that the blocks get returned to the pool eventually.

Even if the write callback gets executed inline `Self._head` should never be null after `Self._head = _lockedEnd.Block`. We already know `_lockedEnd.Block` can't be null because `LockWrite` would exit early leaving `ByteCount` at zero.

src/Microsoft.AspNet.Server.Kestrel/Http/SocketOutput.cs
halter73: I'd make it private since it doesn't implement the interface.

src/Microsoft.AspNet.Server.Kestrel/Http/SocketOutput.cs
benaadams: Wow, ok it is because the `Write` callback is completing before the rest of `DoWriteIfNeeded` so `OnWriteCompleted` is called before `_head` is set to `_lockedEnd.Block`

callstack

benaadams: Ahhhhh.... its the behaviour of `MockLibuv` that's causing this.

benaadams: Fixed it but there would be a recursion error as `WriteAllPending` is called at the end of `OnWriteComplete` so when the stack unwinds _head could be set to an old value. Mitigating...

benaadams: Added throttle fix to https://github.com/aspnet/KestrelHttpServer/pull/427

benaadams:
Bug fix in this PR moving back to `ScheduleWrite` from `WriteAllPending`; however as that may cause a regression have added https://github.com/aspnet/KestrelHttpServer/pull/427 to resolve the regression.

benaadams:
Rebased

benaadams:
Still one of the big players after recent changes; nice to see `byte[]` has dropped of the list though

write-context

AspNetSmurfLab:
Here are the numbers from wrk when running the plaintext benchmark with and without this change:

__Baseline:__

	wrk -c 256 -t 32 -d 30 -s ./scripts/pipeline.lua http://10.0.0.100:5001/plaintext -- 16

	Running 30s test @ http://10.0.0.100:5001/plaintext
	    Latency    14.00ms   37.09ms 443.35ms   92.45%
	    Req/Sec    21.68k     2.62k  139.04k    89.94%
	  20722668 requests in 30.10s, 2.55GB read
	Requests/sec: 688462.29
	Transfer/sec:     86.67MB

	Running 30s test @ http://10.0.0.100:5001/plaintext
	    Latency    15.18ms   36.59ms 640.33ms   90.76%
	    Req/Sec    21.75k     2.69k  146.35k    90.92%
	  20786049 requests in 30.10s, 2.56GB read
	Requests/sec: 690577.45
	Transfer/sec:     86.93MB

	Running 30s test @ http://10.0.0.100:5001/plaintext
	    Latency    16.44ms   45.55ms   1.09s    92.51%
	    Req/Sec    21.65k     2.28k  185.05k    88.12%
	  20695354 requests in 30.10s, 2.54GB read
	Requests/sec: 687599.80
	Transfer/sec:     86.56MB


__After merge:__

	Running 30s test @ http://10.0.0.100:5001/plaintext
	    Latency     4.26ms    5.24ms 137.84ms   95.14%
	    Req/Sec    21.69k     1.11k   34.26k    73.50%
	  20739007 requests in 30.04s, 2.55GB read
	Requests/sec: 690491.69
	Transfer/sec:     86.92MB

	Running 30s test @ http://10.0.0.100:5001/plaintext
	    Latency     4.24ms    5.42ms 218.06ms   95.36%
	    Req/Sec    21.74k     1.07k   48.29k    72.91%
	  20789200 requests in 30.05s, 2.56GB read
	Requests/sec: 691930.78
	Transfer/sec:     87.10MB

	Running 30s test @ http://10.0.0.100:5001/plaintext
	    Latency     4.21ms    5.03ms 150.64ms   94.96%
	    Req/Sec    21.58k     1.03k   30.87k    69.44%
	  20637557 requests in 30.03s, 2.54GB read
	Requests/sec: 687156.69
	Transfer/sec:     86.50MB

Nothing too significant RPS-wise.

benaadams:
@halter73 Yeah, but look at latency...!

benaadams:
Rebased

benaadams:
I'll take the x4 latency improvement with no detriment to RPS :+1:

muratg:
@benaadams LogLevel.Verbose was recently renamed to LogLevel.Trace (https://github.com/aspnet/Announcements/issues/124).  Do you have the up-to-date logging packages?

khellang:
@muratg It fails both locally and on the CI server; https://ci.appveyor.com/project/aspnet-ci/kestrelhttpserver/build/1.0.867#L509.

Also, the Travis build seems to be stuck; https://travis-ci.org/aspnet/KestrelHttpServer/pull_requests

muratg:
I guess the updated packages were not pushed to the feeds yet for some reason.  cc: @pranavkm

benaadams:
@muratg timing? Feeds updated 20 hours ago logging change went in 19 hours ago.

halter73:
We're going to need to get new numbers: https://github.com/aspnet/KestrelHttpServer/pull/458#issuecomment-162992708

benaadams:
Merged  #427 with this as it contains a regress fix.

halter73:
@benaadams What do you think of [this idea](https://github.com/aspnet/KestrelHttpServer/commit/9c2706650912bac974cab373f79d150826a13c3f) of pooling UvWriteReqs instead of WriteContext. I started this change with the intention of continuing to pool WriteContexts, but have a shared pool per thread. I quickly realized this wouldn't be possible (without locking or a thread-safe data structure) since SocketOutput.WriteAsync is called from an arbitrary ThreadPool thread.

So at the cost of one extra allocation (that of the WriteContext), we can now have a bigger pool shared by all connections managed by another thread. I am thinking about opening a PR with this change, but since this is an alternative to pooling WriteContexts, I figured we could discuss it here first.

benaadams:
@halter73 Well... minimising `UvWriteReq` in all ways can only be a good thing being both a finalized object and having allocated native memory. 

However... `WriteContext` is a pretty big object; so what about pooling the `UvWriteReq` on the thread as you've done; but also pooling the `WriteContext` on the socket as it was before (so two queue references)?

halter73:
I'm not sure `WriteContext` is big enough to be worth pooling. Isn't the `UvWriteReq` the main component of `WriteContext` that currently makes it big?

The `WriteContext` pool only living for the lifetime of a connection instead of the lifetime of the server also makes me disinclined to add it.

benaadams:
On x64 the size of `WriteContext` without the `UvWriteReq` allocation is

Component| bytes|total
----|-----|-----
MemoryPoolIterator2 _lockedStart;|-|0
    ptr _block;|8|8
    int _index;|4|12
MemoryPoolIterator2 _lockedEnd;|-|12
    ptr _block;|8|20
    int _index;|4|24
int _bufferCount;|4|28
int ByteCount;|4|32
ptr Self;|8|40
bool SocketShutdownSend;|4|44
bool SocketDisconnect;|4|48
int WriteStatus;|4|52
ptr WriteError;|8|60
ptr _writeReq;|8|68
int ShutdownSendStatus;|4|72
object _syncblk |4|76
object TypeHandle |8|84
object alignment|4|88

88 bytes; for 4k requests (write+shutdown+disconnect) where they only have a single write 1 MB

The number of writes would be higher than the number of requests as there is generally more than 1 write per request.

However, I agree with moving `UvWriteReq` to a more global pool rather than pooling with `WriteContext` as it has higher costs (finalizer+native memory)

benaadams:
@halter73 which would be [this on top](https://github.com/benaadams/KestrelHttpServer/commit/9ed2b85e7c36eb98c1bc81bc5f4810c9aa24ee74) of yours; look good?

PR #367, opened 2015-11-11
Precomputed header bytes

benaadams:
Encode all the standard preamble once for all requests


benaadams:
705,348.96 rps -> 1,030,679.42 rps

/cc @davidfowl @halter73

rynowak:
That's some SRS RPS :+1:

benaadams:
1,069,312.35 on the 3rd wrk run (with Kestrel warmed up)

benaadams:
If `SocketOutput` used `MemoryPool2` internally and there was a `Write(+Async)` overload that took `(MemoryPoolBlock2 block, int length, ...)` and which took ownership of the pooled block (and Released it at write callback); then the copy step at the start of `public Task WriteAsync` could be skipped for the header preamble.

AspNetSmurfLab:
New numbers from SmurfLab with this PR, up from ~497k previously:

```
    Latency    10.36ms   23.70ms 316.86ms   93.90%
    Req/Sec    18.86k     3.02k   35.23k    77.74%
  5912136 requests in 10.10s, 744.25MB read
  Socket errors: connect 0, read 0, write 337, timeout 0
Requests/sec: 585370.49
Transfer/sec:     73.69MB
```

benaadams:
Try non pipelined? I'm also seeing an improvement there.

benaadams:
Added precomputing Http Version bytes also

AspNetSmurfLab:
Wow. With the addition of 371b2ad we break 600K:

```
    Latency    11.26ms   25.12ms 265.05ms   92.61%
    Req/Sec    19.57k     3.08k   43.89k    82.55%
  6115424 requests in 10.10s, 769.84MB read
  Socket errors: connect 0, read 0, write 300, timeout 0
Requests/sec: 605502.09
Transfer/sec:     76.22MB
```

AspNetSmurfLab:
And non-pipelined results:

```
    Latency     1.52ms    1.35ms  56.63ms   96.06%
    Req/Sec     5.68k   667.38     8.41k    83.08%
  1778313 requests in 10.10s, 223.86MB read
  Socket errors: connect 0, read 0, write 552, timeout 0
Requests/sec: 176086.22
Transfer/sec:     22.17MB
```

benaadams:
Have one last additional item for Server and Date; just need to sort out tests

AspNetSmurfLab:
Latest results, getting higher:

```
    Latency    10.50ms   23.15ms 230.54ms   92.07%
    Req/Sec    19.83k     2.93k   35.72k    85.37%
  6204867 requests in 10.10s, 781.10MB read
  Socket errors: connect 0, read 0, write 587, timeout 0
Requests/sec: 614363.71
Transfer/sec:     77.34MB
```

benaadams:
Added Server and Date precomputation; while still allowing swap out/clearing

benaadams:
Wow, Travis is really unhappy, going to give up on it for a bit; hopefully something is updating and not just broken.

AspNetSmurfLab:
Keep it coming :smile: 

```
    Latency     8.42ms   19.76ms 293.82ms   93.96%
    Req/Sec    20.22k     3.08k   40.68k    86.43%
  6302045 requests in 10.10s, 793.33MB read
  Socket errors: connect 0, read 0, write 549, timeout 0
Requests/sec: 623966.09
Transfer/sec:     78.55MB
```

src/Microsoft.AspNet.Server.Kestrel/Http/DateHeaderValueManager.cs
```diff
@@ -17,6 +18,9 @@ public class DateHeaderValueManager : IDisposable
         private readonly TimeSpan _timerInterval;
 
         private volatile string _dateValue;
+        private volatile int _activeDateBytes;
```
halter73: I'm not a huge fan of the bit twiddling. Could we make this a bool and use `_activeDateBytes = !_activeDateBytes` instead of `_activeDateBytes ^= 1`?

src/Microsoft.AspNet.Server.Kestrel/Http/DateHeaderValueManager.cs
```diff
@@ -17,6 +18,9 @@ public class DateHeaderValueManager : IDisposable
...
+        private byte[] _dateBytes0 = Encoding.ASCII.GetBytes("DDD, dd mmm yyyy hh:mm:ss GMT");
```
halter73: Any reason for setting initial values for the `_dateBytes`?

NickCraver: Proper array length :)

halter73: Oops. I thought this was reallocating an array each time.

Might as well mark this a readonly.

benaadams: Done

src/Microsoft.AspNet.Server.Kestrel/Http/DateHeaderValueManager.cs
benaadams: Done

benaadams:
Also rebased

benaadams:
Also passed Travis first time, it must like it...

lodejard:
I really like this, but to go the whole way with it SocketOutput really needs to be memory-pool-block based instead of byte[] segment based... Do we have that change queued up anywhere?

benaadams:
Have a commit that passes the memoryblock through and the SocketOutput held onto it and freed it as well as using memoryblocks internally - however the results were indeterminate. I believe from putting a much higher pressure on memorypool.

However, the results of this are conculsive; and can work out the improved pass through later - @davidfowl gets upset when I put too many changes in a PR ;-)

lodejard:
@davidfowl gets upset by a lot of things :)

I'm putting together a "for instance" commit on top of your branch, but yeah `SocketOutput` needs to become block-aware to make this change happen all the way

benaadams:
@lodejard various test repositories https://github.com/benaadams/KestrelHttpServer/tree/lockless-socket-output and https://github.com/benaadams/KestrelHttpServer/tree/socket-memorypool2

lodejard:
Hello, here's a commit I was going to send as a PR to your fork's branch:

https://github.com/lodejard/KestrelHttpServer/commit/7a265d94a8539c70c1e023ed318666465001d2df

but you've already rebased and made changes on top of it :) so I'm just gonna send this as reference material instead of trying to replant it on top of your new changes

it removes a lot of the special-case code in the generated response headers, but also accomplishes what you are working towards for blitting well-known byte[] material wherever possible

lodejard:
Added another commit - fixing a silly mistake in SetRawXxx which wasn't setting the _rawXxx field as well

See: https://github.com/aspnet/KestrelHttpServer/compare/dev...lodejard:faster-headers

benaadams:
There are a couple bonuses there; on CreateResponseHeader you've gone async?

@davidfowl criticised me for invoking the async state machine in a different PR; which I felt was unfair as he was using my own arguments against me; to address this I had to go a third way https://github.com/aspnet/KestrelHttpServer/pull/391/files maybe something similar can be used?

lodejard:
To be honest I suspect the "async" is will be temporary - @halter73 is also looking into making a change to `SocketOutput` to make it more like `SocketInput` in that the buffers will be head/tail block references. The Frame will be able to call methods like ProduceStart/ProduceComplete in order to access the tail's end iterator and blit directly onto the unused space of the output buffers already present. At that point it wouldn't need to await any Tasks anymore, because it's the SocketOutput's job to return blocks after they're written.

So just like SocketInput, it can be a rolling series of packed/pinned buffers. Should help the pipeline throughput test scenario in particular because a number of 'hello world' responses can probably fit into each 4k page.

DamianEdwards:
I tried to get your modifications running @lodejard but the performance is atrocious and wrk reports thousands of errors. I suspect I've done something wrong trying to fetch your changes into a local branch though, so I'll try again with a fresh clone of your fork.

DamianEdwards:
OK tried a fresh clone of your fork @lodejard and switched to the `faster-headers` branch and I still see the same result, so less likely it's my fault :smile:

lodejard:
No guarantees, :), more of a for-instance commit

benaadams:
@lodejard @halter73 sounds really interesting! Looking forward to it

benaadams:
@lodejard should I copy the changes in?

benaadams:
@lodejard am merging

benaadams:
@DamianEdwards should be good to test

benaadams:
Needs tweaks

benaadams:
May behave better on  #362

AspNetSmurfLab:
Latest results. Note that these are down a littler from the earlier results, likely due to the change to be multi-loop by default now:
```
    Latency    11.73ms   24.73ms 267.14ms   91.27%
    Req/Sec    19.01k     1.73k   31.57k    79.63%
  6094081 requests in 10.10s, 767.15MB read
Requests/sec: 603365.93
Transfer/sec:     75.95MB
```

src/Microsoft.AspNet.Server.Kestrel/Infrastructure/MemoryPoolIterator2.cs
```diff
@@ -530,5 +531,109 @@ public MemoryPoolIterator2 CopyTo(byte[] array, int offset, int count, out int a
...
+                        }
+                        block.End += copied;
+                        _block = block;
+                        _index = block.End;
+                    }
+                }
+            }
+        }
+
+        public void CopyFromAscii(byte[] data)
```
davidfowl: Why is this called CopyFromAscii?

benaadams: Thought that's what you said? `CopyAsciiFrom`?

benaadams: Ahh; just the string method?

src/Microsoft.AspNet.Server.Kestrel/Http/ReasonPhrases.cs
```diff
@@ -2,11 +2,65 @@
...
+        private static readonly byte[] _bytesStatus501 = Encoding.ASCII.GetBytes("501 Not Implemented");
+        private static readonly byte[] _bytesStatus502 = Encoding.ASCII.GetBytes("502 Bad Gateway");
+        private static readonly byte[] _bytesStatus503 = Encoding.ASCII.GetBytes("503 Service Unavailable");
+        private static readonly byte[] _bytesStatus504 = Encoding.ASCII.GetBytes("504 Gateway Timeout");
+        private static readonly byte[] _bytesStatus505 = Encoding.ASCII.GetBytes("505 HTTP Version Not Supported");
+        private static readonly byte[] _bytesStatus506 = Encoding.ASCII.GetBytes("506 Variant Also Negotiates");
+        private static readonly byte[] _bytesStatus507 = Encoding.ASCII.GetBytes("507 Insufficient Storage");
+        private static readonly byte[] _bytesStatus510 = Encoding.ASCII.GetBytes("510 Not Extended");
+
         public static string ToStatus(int statusCode, string reasonPhrase = null)
```
davidfowl: Is this still used?

src/Microsoft.AspNet.Server.Kestrel/Http/Frame.cs
```diff
@@ -655,21 +665,58 @@ public void ProduceContinue()
...
+                _responseHeaders.SetRawConnection("keep-alive", _bytesConnectionKeepAlive);
+            }
+
+            end.CopyFromAscii(_httpVersion == HttpVersionType.Http1_1 ? _bytesHttpVersion1_1 : _bytesHttpVersion1_0);
+            end.CopyFromAscii(statusBytes);
+            _responseHeaders.CopyTo(ref end);
+            end.CopyFromAscii(_bytesEndHeaders, 0, _bytesEndHeaders.Length);
+
+            // TODO: change this to SocketOutput.ProduceStart/ProduceComplete once that change is made 
+            var scan = begin.Block;
```
davidfowl: I think we should wait on @halter73 's change to do this part

src/Microsoft.AspNet.Server.Kestrel/Http/ReasonPhrases.cs
benaadams: Yep by https://github.com/aspnet/KestrelHttpServer/pull/367/files#diff-3ad18666c41931c02b7bd1f3226cb709R582 which calls `ToStatusBytes`

src/Microsoft.AspNet.Server.Kestrel/Http/Frame.cs
benaadams: Do you mean the whole PR? This bit passes the individual ArraySegements into WriteAsync as it can't currently take a set of memoryblocks. Will change after @halter73's change to just pass the memory chain in (I believe?)

davidfowl: I'm actually interested to see what it'll look like.

benaadams:
Rebased on #410 doing integration now

/cc @halter73

lodejard:
I just filed https://github.com/aspnet/Common/issues/64 we might want to consider.

It's related to this PR in that it would open a channel from app to server for pre-computed ascii byte[] response headers to be written.

davidfowl:
There's some duplication here with the CopyFrom logic. This looks much cleaner with @halter73 's changes

src/Microsoft.AspNet.Server.Kestrel/Http/Frame.cs
```diff
@@ -29,6 +29,17 @@ public partial class Frame : FrameContext, IFrameControl
         private static readonly ArraySegment<byte> _emptyData = new ArraySegment<byte>(new byte[0]);
         private static readonly byte[] _hex = Encoding.ASCII.GetBytes("0123456789abcdef");
 
+        private static readonly byte[] _bytesConnectionClose = Encoding.ASCII.GetBytes("\r\nConnection: close");
+        private static readonly byte[] _bytesConnectionKeepAlive = Encoding.ASCII.GetBytes("\r\nConnection: keep-alive");
```
davidfowl: Should move these out of frame into a static class? Similar to reason phrases?

src/Microsoft.AspNet.Server.Kestrel/Http/Frame.cs
```diff
@@ -654,21 +664,30 @@ public void ProduceContinue()
...
+            count += end.CopyFrom(statusBytes);
+            count += _responseHeaders.CopyTo(ref end);
+            count += end.CopyFrom(_bytesEndHeaders, 0, _bytesEndHeaders.Length);
+
+            SocketOutput.ProducingComplete(end, count);
+
+            if (immediate)
             {
-                writer.Write("Connection: keep-alive\r\n\r\n");
+                return SocketOutput.WriteAsync(default(ArraySegment<byte>), immediate: true);
```
davidfowl: Can we call flush?

benaadams: Flush doesn't exist on `ISocketOutput` and `ProducingComplete` stipulates "A write operation isn't guaranteed to be scheduled unless Write(Async) is called afterwards"

benaadams:
Seem to be spitting out the right length data but in lots of zero bytes; not sure I've integrated right.

src/Microsoft.AspNet.Server.Kestrel/Http/Frame.cs
```diff
@@ -654,21 +664,30 @@ public void ProduceContinue()
...
-            else if (_keepAlive && hasConnection == false && HttpVersion == "HTTP/1.0")
+            else if (_keepAlive && _responseHeaders.HasConnection == false && _httpVersion == HttpVersionType.Http1_0)
+            {
+                _responseHeaders.SetRawConnection("keep-alive", _bytesConnectionKeepAlive);
+            }
+
+            count += end.CopyFrom(_httpVersion == HttpVersionType.Http1_1 ? _bytesHttpVersion1_1 : _bytesHttpVersion1_0);
+            count += end.CopyFrom(statusBytes);
+            count += _responseHeaders.CopyTo(ref end);
+            count += end.CopyFrom(_bytesEndHeaders, 0, _bytesEndHeaders.Length);
```
halter73: This looks good. I was considering making CopyFrom mutable. I wonder if CopyTo should also mutate for consistency.

benaadams: Needed to get the count for the count; though I have 27 failing tests I need to work out.

lodejard: huh? absolutely CopyFrom should mutate the iterator, doesn't it? CopyTo can be changed to match.

ProducingComplete should only take (end), or (begin,end), but not the (count). The SocketOutput knows what to write based on the original tail's iterator returned, and the complete iterator passed in

src/Microsoft.AspNet.Server.Kestrel/Infrastructure/MemoryPoolIterator2.cs
```diff
@@ -530,5 +531,114 @@ public MemoryPoolIterator2 CopyTo(byte[] array, int offset, int count, out int a
...
+                {
+                    if (blockRemaining == 0)
+                    {
+                        block.Next = block.Pool.Lease();
+                        block = block.Next;
+                        blockRemaining = block.Data.Count;
+                        blockRemainingMinusSpan = blockRemaining - 3;
+                    }
+
+                    fixed (byte* pOutput = block.Data.Array)
```
halter73: How much more faster is this over normal iteration over the byte array?

src/Microsoft.AspNet.Server.Kestrel/Http/Frame.cs
benaadams: the count is for bytesPrecompted rather than traversing the iterator to work it out.

src/Microsoft.AspNet.Server.Kestrel/Infrastructure/MemoryPoolIterator2.cs
benaadams: Have a PR with results for that https://github.com/aspnet/KestrelHttpServer/pull/399#issuecomment-157728877 main issue is casting char to byte; or in the referenced PR byte to char; else could use block copy.

src/Microsoft.AspNet.Server.Kestrel/Http/Frame.cs
lodejard: The majority of the cases are where the entire response fits in one block, so having SocketOutput call begin.GetLength(end) will just subtract the index integers.

halter73: Yep. The count is only to simplify write-behind buffering calculation. We do traverse the blocks and count how many bytes we're actually writing before calling `uv_write`.

src/Microsoft.AspNet.Server.Kestrel/Http/Frame.cs
halter73: You could call Frame.Flush, but that does exactly this after wasting time calling `ProduceStartAndFireOnStarting` which would hopefully no-op anyway.

benaadams:
5 tests to fix

benaadams:
Yay!

src/Microsoft.AspNet.Server.Kestrel/Http/Frame.cs
benaadams: Less clutter?

benaadams:
Rebased on SocketOuput changes; fingers crossed for performance

benaadams:
Looks like it might be a good one, al lot of the allocations drop out

Before

pre

After

post

benaadams:
Bug in the copy (hence the System.Exception.ToString having high allocs)

benaadams:
Adding #411, #408 the allocations change to the following

post-extras-allocs

benaadams:
RPS isn't great; but CPU is low, so I think there is high lock contention - making some minor adjusts.

benaadams:
Reduced the in lock work; but also turned out I had receive side scaling turned off. When I turn that on... Oh my...

```
    Latency    11.67ms   34.29ms 657.95ms   90.26%
    Req/Sec    40.62k     4.81k   76.06k    77.62%
  77224953 requests in 1.00m, 9.49GB read
Requests/sec: 1284937.56
Transfer/sec:    161.75MB

```

halter73:
Nice. I'm merging this in now. I will plan to make a few more changes, like changing the signature of ProducingComplete to not take a count, but that can be done in another PR.

benaadams:
:checkered_flag: time to rebase all the things...

PR #373, opened 2015-11-12
Skip tests on Mono to prevent CI hangs.

cesarbs:
We're having issues with hangs on Travis and the CI after running functional tests on Mono. I've identified the reason for this, which is that KestrelThread.ThreadStart() hangs on the second call it makes to _loop.Run(). However, I haven't found a solution to this yet, so I'd like to disable these tests on Mono for now, so we can get past the hangs that we've been seeing.

cc @halter73 @lodejard @Tratcher

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,10 @@
 language: csharp
-sudo: false
+sudo: required
+dist: trusty
+addons:
+  apt:
+    packages:
+      - libunwind8
 install:
   - curl -sSL https://github.com/libuv/libuv/archive/v1.4.2.tar.gz | tar zxfv - -C /tmp && cd /tmp/libuv-1.4.2/
   - sh autogen.sh
@@ -9,4 +14,5 @@ install:
   - export LD_LIBRARY_PATH="$HOME/libuvinstall/lib"
   - cd $OLDPWD
 script:
+  - export KOREBUILD_TEST_DNXCORE=1
   - ./build.sh --quiet verify
--- a/test/Microsoft.AspNet.Server.Kestrel.FunctionalTests/IPv6SupportedConditionAttribute.cs
+++ b/test/Microsoft.AspNet.Server.Kestrel.FunctionalTests/IPv6SupportedConditionAttribute.cs
@@ -1,0 +1,48 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Net;
+using System.Net.Sockets;
+using Microsoft.AspNet.Testing.xunit;
+
+namespace Microsoft.AspNet.Server.Kestrel.FunctionalTests
+{
+    [AttributeUsage(AttributeTargets.Method, AllowMultiple = false)]
+    public class IPv6SupportedConditionAttribute : Attribute, ITestCondition
+    {
+        private static Lazy<bool> _ipv6Supported = new Lazy<bool>(CanBindToIPv6Address);
+
+        public bool IsMet
+        {
+            get
+            {
+                return _ipv6Supported.Value;
+            }
+        }
+
+        public string SkipReason
+        {
+            get
+            {
+                return "IPv6 not supported on the host.";
+            }
+        }
+
+        private static bool CanBindToIPv6Address()
+        {
+            try
+            {
+                using (var socket = new Socket(AddressFamily.InterNetworkV6, SocketType.Stream, ProtocolType.Tcp))
+                {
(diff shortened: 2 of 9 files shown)


Tratcher:
:shipit:

halter73:
Now the timeout occurs after KestrelTests :cry:

cesarbs:
@troydai

cesarbs:
Waiting on https://github.com/aspnet/Testing/pull/187 so we can skip tests on Travis when they bind to ::1.

cesarbs:
Per @victorhurdugaci's suggestion, instead of skipping IPv6 tests based on the environment variable on Travis, I added a runtime check for IPv6 support on the machine that's running the tests. That way we don't have to worry about environment settings and whatnot.

test/Microsoft.AspNet.Server.Kestrel.FunctionalTests/IPv6SupportedConditionAttribute.cs
```diff
@@ -0,0 +1,49 @@
...
+using System.Net;
+using System.Net.Sockets;
+using Microsoft.AspNet.Testing.xunit;
+
+namespace Microsoft.AspNet.Server.Kestrel.FunctionalTests
+{
+    [AttributeUsage(AttributeTargets.Method, AllowMultiple = false)]
+    public class IPv6SupportedConditionAttribute : Attribute, ITestCondition
+    {
+		private readonly Lazy<bool> _ipv6Supported = new Lazy<bool>(CanBindToIPv6Address);
```
victorhurdugaci: indentation

test/Microsoft.AspNet.Server.Kestrel.FunctionalTests/IPv6SupportedConditionAttribute.cs
```diff
@@ -0,0 +1,49 @@
...
+
+        public string SkipReason
+        {
+            get
+            {
+                return "IPv6 not supported on the host.";
+            }
+        }
+
+		private static bool CanBindToIPv6Address()
```
victorhurdugaci: indentation

test/Microsoft.AspNet.Server.Kestrel.FunctionalTests/IPv6SupportedConditionAttribute.cs
victorhurdugaci: Maybe make it static. This will probably not change between test runs

benaadams:
Addressed the blocking second call in #379 though Travis still hangs

benaadams:
nvm solution didn't work :(

troydai:
:shipit:

.travis.yml
```diff
@@ -1,5 +1,10 @@
 language: csharp
-sudo: false
+sudo: required
```
halter73: I would drop `sudo: required` if we can.

test/Microsoft.AspNet.Server.Kestrel.FunctionalTests/IPv6SupportedConditionAttribute.cs
```diff
@@ -0,0 +1,49 @@
...
+                    IPAddress hostIP = Dns.GetHostEntryAsync("::1").Result.AddressList[0];
+                    socket.Bind(new IPEndPoint(hostIP, 8787));
+                    return true;
+                }
+            }
+            catch (SocketException)
+            {
+                return false;
+            }
+		}
```
halter73: ident

.travis.yml
cesarbs: Can't: http://docs.travis-ci.com/user/trusty-ci-environment/

.travis.yml
```diff
@@ -1,5 +1,10 @@
...
+dist: trusty
```
dougbu: Doesn't this hit the issue in Mono that @BrennanConroy fixed? If yes, you probably need to wait for a KoreBuild workaround I'll be adding over the weekend, enabling `dnu restore` using DNX Core.

cesarbs: I'll check this in anyways. Once your workaround gets in, that problem will be resolved.

benaadams:
w00p! :dancer:

.travis.yml
dougbu: But between now and then Travis builds will fail even more frequently won't they?

benaadams: Haha, if you look at this list I'm not sure they can fail more frequently: https://travis-ci.org/aspnet/KestrelHttpServer/pull_requests

benaadams: It was about 19 in 20 that would fail

PR #378, opened 2015-11-13
Network thread count defaults

### LINKED ISSUE #375: Set IO thread count to half of logical CPU count by default
Using an IO thread (libuv loop) for every 2 logical CPUs in the system appears to be a good default value based on benchmarking.

``` C#
IKestrelServerInformation.ThreadCount = Environment.ProcessorCount >> 1;
```

benaadams:
Resolves #375


benaadams:
Need to keep these tests switched off for mono?

PR #379, opened 2015-11-13
Don't block on shutdown

benaadams:
Addresses issue mentioned in  #373

> We're having issues with hangs on Travis and the CI after running functional tests on Mono. I've identified the reason for this, which is that KestrelThread.ThreadStart() hangs on the second call it makes to _loop.Run()


benaadams:
Doesn't like dispose if stuff still outstanding.

PR #380, opened 2015-11-13
OnCompleted doesn't need threadpool

benaadams:
Is called by the awaiting thread, rather than the awaiter's thread

### DIFF
--- a/src/Microsoft.AspNet.Server.Kestrel/Http/SocketInput.cs
+++ b/src/Microsoft.AspNet.Server.Kestrel/Http/SocketInput.cs
@@ -200,6 +200,7 @@ namespace Microsoft.AspNet.Server.Kestrel.Http
             else
             {
                 // THIS IS AN ERROR STATE - ONLY ONE WAITER CAN WAIT
+                throw new InvalidOperationException("Already an awaited read");
             }
         }
 
--- a/src/Microsoft.AspNet.Server.Kestrel/Http/SocketInput.cs
+++ b/src/Microsoft.AspNet.Server.Kestrel/Http/SocketInput.cs
@@ -5,7 +5,6 @@ using System;
 using System.IO;
 using System.Runtime.CompilerServices;
 using System.Threading;
-using System.Threading.Tasks;
 using Microsoft.AspNet.Server.Kestrel.Infrastructure;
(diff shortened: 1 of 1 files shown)


src/Microsoft.AspNet.Server.Kestrel/Http/SocketInput.cs
```diff
@@ -194,11 +194,13 @@ public void OnCompleted(Action continuation)
             }
             else if (awaitableState == _awaitableIsCompleted)
             {
-                ThreadPool.QueueUserWorkItem((o) => ((Action)o)(), continuation);
+                // Read already completed, continue on calling thread
+                continuation();
```
davidfowl: Hmmmm, stack diving concerns? Is this always safe to call inline?

/cc @stephentoub

stephentoub: From a quick glance, this is suspectible.  An awaiter's OnCompleted should always queue the invocation of the delegate.

benaadams:
Closing

src/Microsoft.AspNet.Server.Kestrel/Http/SocketInput.cs
benaadams: @stephentoub is there a way to detect if you are already on threadpool/top of stack/async complete; rather than doing a double queue? E.g. the stack trace on https://github.com/aspnet/KestrelHttpServer/issues/451 is clearly top of stack.

stephentoub: Why are we trying hard to avoid queuing here? This should be rare: assuming this call is part of an await, we just checked to see if it already completed, so it would have needed to complete between that check and this one immediately after. Are you seeing this case a high percentage of the time?

benaadams: No; just came back to this code investigating why I hit the 3rd state neither IsNotCompleted or IsCompleted and the stack trace showed top of stack. However I only hit this 3rd state 3 times in 30 mins under load; and as you point out it being completed in the interval is probably a rare state.

PR #381, opened 2015-11-13
Set NoDelay as default

### LINKED ISSUE #374: Enable TCP no-delay (disable nagling) by default
This has been the default in IIS since 6.0. We should follow suit.

benaadams:
Resolves  #374


halter73:
You're taking all my bugs! Not that I'm complaining :smile:

benaadams:
Was looking for something to test the Travis with that wouldn't unnessarily clog AppVeyor :grin:

Issue #371, opened 2015-11-12
Lots of byte[] arrays allocated by kestrel


rynowak:
Allocations for 3000 requests

![image](https://cloud.githubusercontent.com/assets/1430011/11108566/ac3eb094-889c-11e5-8cd9-ec421a95a5bd.png)

rynowak:
Looks like we regressed here: https://github.com/aspnet/KestrelHttpServer/commit/d08b83a1c7241bca66c6934facb2410f118489c3

It should be safe to statically cache the platform right?

DamianEdwards:
Probably should change to use `System.Runtime.InteropServices.RuntimeInformation` instead (package & a class).

davidfowl:
Yep!

troydai:
Fixed in https://github.com/aspnet/KestrelHttpServer/commit/2e2dfd87c1dcd5c286679276670becc9b34fc8c4

PR #387, opened 2015-11-15
Fix byte[] array allocation in IsDarwin

troydai:
/cc @halter73 @Tratcher


davidfowl:
:shipit:

NickCraver:
Resolves #371

benaadams:
Also currently (without this change) throws an exception on the start of every connection on windows

halter73:
:ship:

PR #390, opened 2015-11-15
ConcurrentStack -> ConcurrentQueue

benaadams:
Stack to Queue test by @NickCraver https://twitter.com/Nick_Craver/status/665503509958258688

### DIFF
--- a/src/Microsoft.AspNet.Server.Kestrel/Infrastructure/MemoryPool2.cs
+++ b/src/Microsoft.AspNet.Server.Kestrel/Infrastructure/MemoryPool2.cs
@@ -40,7 +40,7 @@ namespace Microsoft.AspNet.Server.Kestrel.Infrastructure
         /// Thread-safe collection of blocks which are currently in the pool. A slab will pre-allocate all of the block tracking objects
         /// and add them to this collection. When memory is requested it is taken from here first, and when it is returned it is re-added.
         /// </summary>
-        private readonly ConcurrentStack<MemoryPoolBlock2> _blocks = new ConcurrentStack<MemoryPoolBlock2>();
+        private readonly ConcurrentQueue<MemoryPoolBlock2> _blocks = new ConcurrentQueue<MemoryPoolBlock2>();
 
         /// <summary>
         /// Thread-safe collection of slabs which have been allocated by this pool. As long as a slab is in this collection and slab.IsActive, 
@@ -75,7 +75,7 @@ namespace Microsoft.AspNet.Server.Kestrel.Infrastructure
             }
 
             MemoryPoolBlock2 block;
-            if (_blocks.TryPop(out block))
+            if (_blocks.TryDequeue(out block))
             {
                 // block successfully taken from the stack - return it
                 return block;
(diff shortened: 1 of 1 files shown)


NickCraver:
+1 - `ConcurrentQueue` should be (and is measured to be) faster at higher mixed workloads and we have a 1:1 here. 

Reference for the curious: http://blogs.msdn.com/b/pfxteam/archive/2010/04/26/9997562.aspx

benaadams:
ConcurrentBag might work here; its a memory pig though :(

lodejard:
Isn't it better to provide the most-recently-returned block, rather than provide the least-recently-used block, when someone asks for memory? That's why it was a stack instead of a queue in the first place

benaadams:
The issue is caused by contention by both [readers](https://github.com/dotnet/corefx/blob/master/src/System.Collections.Concurrent/src/System/Collections/Concurrent/ConcurrentStack.cs#L292) and [writers](https://github.com/dotnet/corefx/blob/master/src/System.Collections.Concurrent/src/System/Collections/Concurrent/ConcurrentStack.cs#L488) on the `_head` with a `ConcurrentStack` so when it hits high throughput the Interlocked fails and it moves to burning cpu with [`PushCore`](https://github.com/dotnet/corefx/blob/master/src/System.Collections.Concurrent/src/System/Collections/Concurrent/ConcurrentStack.cs#L380) and [`TryPopCore`](https://github.com/dotnet/corefx/blob/master/src/System.Collections.Concurrent/src/System/Collections/Concurrent/ConcurrentStack.cs#L603) in cpu spins - though it will always make forward progress.

With `ConcurrentQueue` it [enqueues to tail and dequeues from tail](https://github.com/dotnet/corefx/blob/master/src/System.Collections.Concurrent/src/System/Collections/Concurrent/ConcurrentQueue.cs#L534) which are different locations so writes only contend with writes and reads only contend with reads.

Also it allocates individual Nodes for every push vs `ConcurrentQueue` using array chunks and reusing them; so allocating less.

benaadams:
Though something to measure (perhaps after the other changes to up throughput)

@NickCraver saw an increase from 931,289.67 RPS to 1,191,674.89 RPS when this was added on top of the other changes.

rynowak:
If contention on the memory pool is having this much of an impact on perf, it might be worth investigating if round-distributing to separate pools is an improvement.

benaadams:
@rynowak I push them out so each `KestrelThread` has there own pool in this PR https://github.com/aspnet/KestrelHttpServer/pull/343/files#diff-2ee816fea70408fbc7f21874eccb9211R35 also so they can be disposed.

lodejard:
Okay, reducing push/pull contention sounds like a reasonable cause for the change. 

Doesn't each thread already have it's own pool2?

lodejard:
Well - each uv loop more than each thread. The fact that some of the work is done on worker threads means the pool isn't truly isolated to the loop

benaadams:
> Doesn't each thread already have it's own pool2?

PRs not been merged yet ;-)

PR #391, opened 2015-11-15
Don't alloc header names

benaadams:


benaadams:
Split from #335

src/Microsoft.AspNet.Server.Kestrel/Http/Frame.cs
```diff
@@ -318,6 +308,36 @@ public void Abort()
             }
         }
 
+        private async Task<bool> ReadStartLineAsync()
```
davidfowl: Moar async state machines.

src/Microsoft.AspNet.Server.Kestrel/Http/Frame.cs
```diff
@@ -838,7 +860,7 @@ public static bool TakeMessageHeaders(SocketInput input, FrameRequestHeaders req
                             continue;
                         }
 
-                        var name = beginName.GetArraySegment(endName);
+                        var name = beginName.GetArraySegment(endName, memoryBlock.Data);
```
davidfowl: I'm not sure how often header names are split across blocks... Seems edgy

src/Microsoft.AspNet.Server.Kestrel/Http/Frame.cs
benaadams: Dodged!

DamianEdwards:
Rebased on 7e8a405 and measured. Made it worse.

```
dev (7e8a405) - 499k
391 (e0ebb12) - 465k (-7%)
```

benaadams:
@DamianEdwards hopefully it will behave better on  #362

benaadams:
Closing this taken some legibility improvements to  #424

PR #393, opened 2015-11-16
Disconnect is Debug not Error

### LINKED ISSUE #392: Connection id "30" disconnected.
Hello, 

Last build of Kestrel logs connection disconnections "Connection id "30" disconnected." as errors.
Since they are not real errors, is there a way to prevent Kestrel to log them in default logger ?

benaadams:
Resolves #392

Correction to issue caused by #326


src/Microsoft.AspNet.Server.Kestrel/Infrastructure/KestrelTrace.cs
```diff
@@ -35,7 +35,7 @@ static KestrelTrace()
             _connectionWriteFin = LoggerMessage.Define<long>(LogLevel.Debug, 7, @"Connection id ""{ConnectionId}"" sending FIN.");
             _connectionWroteFin = LoggerMessage.Define<long, int>(LogLevel.Debug, 8, @"Connection id ""{ConnectionId}"" sent FIN with status ""{Status}"".");
             _connectionKeepAlive = LoggerMessage.Define<long>(LogLevel.Debug, 9, @"Connection id ""{ConnectionId}"" completed keep alive response.");
-            _connectionDisconnect = LoggerMessage.Define<long>(LogLevel.Error, 10, @"Connection id ""{ConnectionId}"" disconnected.");
+            _connectionDisconnect = LoggerMessage.Define<long>(LogLevel.Debug, 10, @"Connection id ""{ConnectionId}"" disconnected.");
```
Tratcher: Verbose?

benaadams: This just matches what it was previously.

PR #396, opened 2015-11-17
Initialize addresses property with urls array

khellang:
Initialize the `Addresses` property using the `urls` array, instead of allocating a backing array with the wrong size.

### DIFF
--- a/src/Microsoft.AspNet.Server.Kestrel/KestrelServerInformation.cs
+++ b/src/Microsoft.AspNet.Server.Kestrel/KestrelServerInformation.cs
@@ -11,7 +11,7 @@ namespace Microsoft.AspNet.Server.Kestrel
 {
     public class KestrelServerInformation : IKestrelServerInformation, IServerAddressesFeature
     {
-        public ICollection<string> Addresses { get; } = new List<string>();
+        public ICollection<string> Addresses { get; private set; }
 
         public int ThreadCount { get; set; }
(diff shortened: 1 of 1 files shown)


src/Microsoft.AspNet.Server.Kestrel/KestrelServerInformation.cs
```diff
@@ -22,10 +22,7 @@ public class KestrelServerInformation : IKestrelServerInformation, IServerAddres
         public void Initialize(IConfiguration configuration)
         {
             var urls = configuration["server.urls"] ?? string.Empty;
-            foreach (var url in urls.Split(new[] { ';' }, StringSplitOptions.RemoveEmptyEntries))
-            {
-                Addresses.Add(url);
-            }
+            Addresses = urls.Split(new[] { ';' }, StringSplitOptions.RemoveEmptyEntries);
```
davidfowl: I think this breaks here:

https://github.com/aspnet/Hosting/blob/dev/src/Microsoft.AspNet.Hosting/Internal/HostingEngine.cs#L262

We don't want the collection to be read only.

khellang: Ugh. Yeah, fixed.

davidfowl:
Spoke to @khellang offline and closing this as a result

PR #397, opened 2015-11-17
Made IKestrelServerInformation.ThreadCount nullable

khellang:
Noticed this when looking at #378. I think this looks better than `if (information.ThreadCount > 0)` :smile:


khellang:
When I think about it, why even bother doing all of the above if `information.ThreadCount.HasValue`? It's just being overridden anyway? // @benaadams

khellang:
Replaced by #417.

PR #401, opened 2015-11-18
Johluo/generic iserver

JunTaoLuo:
Reacting to https://github.com/aspnet/Hosting/pull/463

### DIFF
--- a/src/Microsoft.AspNet.Server.Kestrel/KestrelServer.cs
+++ b/src/Microsoft.AspNet.Server.Kestrel/KestrelServer.cs
@@ -5,7 +5,6 @@ using System;
 using System.Collections.Generic;
 using Microsoft.AspNet.Hosting;
 using Microsoft.AspNet.Hosting.Server;
-using Microsoft.AspNet.Http;
 using Microsoft.AspNet.Http.Features;
 using Microsoft.AspNet.Server.Kestrel.Http;
 using Microsoft.Extensions.Logging;
@@ -17,9 +16,8 @@ namespace Microsoft.AspNet.Server.Kestrel
         private Stack<IDisposable> _disposables;
         private readonly IApplicationLifetime _applicationLifetime;
         private readonly ILogger _logger;
-        private readonly IHttpContextFactory _httpContextFactory;
 
-        public KestrelServer(IFeatureCollection features, IApplicationLifetime applicationLifetime, ILogger logger, IHttpContextFactory httpContextFactory)
+        public KestrelServer(IFeatureCollection features, IApplicationLifetime applicationLifetime, ILogger logger)
         {
             if (features == null)
             {
@@ -36,20 +34,14 @@ namespace Microsoft.AspNet.Server.Kestrel
                 throw new ArgumentNullException(nameof(logger));
             }
 
-            if (httpContextFactory == null)
-            {
-                throw new ArgumentNullException(nameof(httpContextFactory));
-            }
-
             _applicationLifetime = applicationLifetime;
             _logger = logger;
             Features = features;
-            _httpContextFactory = httpContextFactory;
         }
 
         public IFeatureCollection Features { get; }
 
-        public void Start(RequestDelegate requestDelegate)
+        public void Start<TContext>(IHttpApplication<TContext> application)
         {
             if (_disposables != null)
             {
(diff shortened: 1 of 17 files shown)


src/Microsoft.AspNet.Server.Kestrel/KestrelServer.cs
```diff
@@ -64,9 +64,9 @@ public void Start(RequestDelegate requestDelegate)
                 var dateHeaderValueManager = new DateHeaderValueManager();
                 var engine = new KestrelEngine(new ServiceContext
                 {
+                    FrameFactory = (context, remoteEP, localEP) => { return new Frame<THttpContext>(application, context, remoteEP, localEP); },
```
halter73: Nit: Either remove the curly braces or make this multiple lines.

test/Microsoft.AspNet.Server.KestrelTests/EngineTests.cs
```diff
@@ -91,10 +92,14 @@ public void EngineCanStartAndStop(ServiceContext testContext)
         [FrameworkSkipCondition(RuntimeFrameworks.Mono, SkipReason = "Test hangs after execution on Mono.")]
         public void ListenerCanCreateAndDispose(ServiceContext testContext)
         {
+            testContext.FrameFactory = (connectionContext, remoteEP, localEP) =>
+            {
+                return new Frame<HttpContext>(new DummyApplication(App), connectionContext, remoteEP, localEP);
+            };
```
halter73: Maybe create `TestServiceContext.App` so this can turn into `testContext.App = App;` The setter can do exactly what this above does.

src/Microsoft.AspNet.Server.Kestrel/Http/Frame.Templated.cs
```diff
@@ -0,0 +1,128 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
```
halter73: Do we have a standard way of of naming generic classes that have a non-generic counterpart?

src/Microsoft.AspNet.Server.Kestrel/Http/Frame.Templated.cs
```diff
@@ -0,0 +1,128 @@
...
+
+                        var context = _application.CreateContext(this);
+                        try
+                        {
+                            await _application.ProcessRequestAsync(context).ConfigureAwait(false);
+                        }
+                        catch (Exception ex)
+                        {
+                            ReportApplicationError(ex);
+                            _application.DisposeContext(context, ex);
```
halter73: I think this is disposing the context too soon. `FireOnCompleted` is yet to run.

src/Microsoft.AspNet.Server.Kestrel/Http/Frame.Templated.cs
```diff
@@ -0,0 +1,128 @@
...
+                            if (!_responseStarted && _applicationException == null)
+                            {
+                                await FireOnStarting();
+                            }
+
+                            await FireOnCompleted();
+
+                            if (_applicationException == null)
+                            {
+                                _application.DisposeContext(context, null);
```
halter73: The context might never be disposed with this logic. Example case: _application.ProcessRequestAsync completes successfully, but FireOnStarting/Completed sets _applicationException.

JunTaoLuo:
This needs a rebase.

src/Microsoft.AspNet.Server.Kestrel/Http/Frame.Templated.cs
JunTaoLuo: I've seen *OfT

JunTaoLuo:
:up::date:

JunTaoLuo:
@Tratcher @davidfowl @lodejard @halter73

halter73:
Too bad we can't see the tests pass until the other changes are in.

src/Microsoft.AspNet.Server.Kestrel/Http/Frame.Templated.cs
halter73: Let's do that then.

JunTaoLuo:
@halter73, need volkswagen.

src/Microsoft.AspNet.Server.Kestrel/Http/Frame.Templated.cs
JunTaoLuo: Though files like https://github.com/aspnet/Logging/blob/dev/src/Microsoft.Extensions.Logging/LoggerOfT.cs do exist, I would actually suggest `Frame.OfT.cs` since that places it under `Frame.cs` under VS. Thoughts?

halter73: I expect my heavy use of `ctrlr+,` is going to cause me to dislike `Frame.OfT.cs`. I'm already thoroughly annoyed that `Frame.FeatureCollection.cs` is the first result when I type "frame".

Do what you think is right <sup>but I might really dislike you if you make the wrong decision</sup>

halter73:
:car:

halter73:
:shipit:

JunTaoLuo:
Tested locally, merged to dev

PR #402, opened 2015-11-18
Only use pooled buffers

benaadams:
Only use pooled blocks.
Don't finalize blocks outside debug.

For the cost of just including a finalizer on allocation see:  
http://www.philosophicalgeek.com/2014/09/29/digging-into-net-object-allocation-fundamentals/

### DIFF
--- a/src/Microsoft.AspNet.Server.Kestrel/Infrastructure/MemoryPoolBlock2.cs
+++ b/src/Microsoft.AspNet.Server.Kestrel/Infrastructure/MemoryPoolBlock2.cs
@@ -80,7 +80,7 @@ namespace Microsoft.AspNet.Server.Kestrel.Infrastructure
         ~MemoryPoolBlock2()
         {
             Debug.Assert(!_pinHandle.IsAllocated, "Ad-hoc memory block wasn't unpinned");
-            // Debug.Assert(Slab == null || !Slab.IsActive, "Block being garbage collected instead of returned to pool");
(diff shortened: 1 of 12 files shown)


src/Microsoft.AspNet.Server.Kestrel/Infrastructure/MemoryPoolBlock2.cs
```diff
@@ -74,6 +74,9 @@ protected MemoryPoolBlock2()
         /// </summary>
         public MemoryPoolBlock2 Next { get; set; }
 
+#if DEBUG
```
halter73: Maybe we should have a Debug.Assert for when `Slab != null` to ensure that we are always returning the block (in our tests at least).

benaadams: Also probably shouldn't add blocks that aren't pooled to the pool in `Return`

benaadams:
Have something better

src/Microsoft.AspNet.Server.Kestrel/Infrastructure/MemoryPoolBlock2.cs
benaadams: Removed non-pooled blocks instead (from `Lease`)

benaadams:
Revisit this later

benaadams:
Needs some revitalization...

Issue #370, opened 2015-11-12
If a ConnectionFilter throws synchronously the server crashes


halter73:
The exception will be logged inside of UvStreamHandle.UvConnectionCb and rethrown.

    Hosting environment: Production
    Now listening on: http://localhost:5000
    Now listening on: https://localhost:5001
    Application started. Press Ctrl+C to shut down.
    dbug: Microsoft.AspNet.Server.Kestrel[1]
          Connection id "1" started.
    fail: Microsoft.AspNet.Server.Kestrel[0]
          UvConnectionCb
          System.Exception: Exception of type 'System.Exception' was thrown.
             at Microsoft.AspNet.Server.Kestrel.Filter.NoOpConnectionFilter.OnConnection(ConnectionFilterContext context) in C:\Users\shalter\dev\Universe\KestrelHttpServer\src\Microsoft.AspNet.Server.Kestrel\Filter\NoOpConnectionFilter.cs:line 13
             at Microsoft.AspNet.Server.Kestrel.Http.Connection.Start() in C:\Users\shalter\dev\Universe\KestrelHttpServer\src\Microsoft.AspNet.Server.Kestrel\Http\Connection.cs:line 82
             at Microsoft.AspNet.Server.Kestrel.Http.Listener.DispatchConnection(UvStreamHandle socket) in C:\Users\shalter\dev\Universe\KestrelHttpServer\src\Microsoft.AspNet.Server.Kestrel\Http\Listener.cs:line 80
             at Microsoft.AspNet.Server.Kestrel.Http.TcpListener.OnConnection(UvStreamHandle listenSocket, Int32 status) in C:\Users\shalter\dev\Universe\KestrelHttpServer\src\Microsoft.AspNet.Server.Kestrel\Http\TcpListener.cs:line 53
             at Microsoft.AspNet.Server.Kestrel.Http.Listener.ConnectionCallback(UvStreamHandle stream, Int32 status, Exception error, Object state) in C:\Users\shalter\dev\Universe\KestrelHttpServer\src\Microsoft.AspNet.Server.Kestrel\Http\Listener.cs:line 66
             at Microsoft.AspNet.Server.Kestrel.Http.TcpListener.<>c.<CreateListenSocket>b__1_0(UvStreamHandle stream, Int32 status, Exception error, Object state) in C:\Users\shalter\dev\Universe\KestrelHttpServer\src\Microsoft.AspNet.Server.Kestrel\Http\TcpListener.cs:line 28

[...]

PR #403, opened 2015-11-19
Don't crash the server if a connection filter throws synchronously.

cesarbs:
https://github.com/aspnet/KestrelHttpServer/issues/370

cc @halter73 @Tratcher

### DIFF
--- a/test/Microsoft.AspNet.Server.KestrelTests/ConnectionFilterTests.cs
+++ b/test/Microsoft.AspNet.Server.KestrelTests/ConnectionFilterTests.cs
@@ -1,6 +1,7 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System;
 using System.IO;
 using System.Threading.Tasks;
 using Microsoft.AspNet.Http;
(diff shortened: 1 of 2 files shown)


test/Microsoft.AspNet.Server.KestrelTests/ConnectionFilterTests.cs
```diff
@@ -80,7 +81,36 @@ public class ConnectionFilterTests
                         "Hello World!");
                 }
             }
-       }
+        }
+
+        [ConditionalFact]
+        [FrameworkSkipCondition(RuntimeFrameworks.Mono, SkipReason = "Test hangs after execution on Mono.")]
```
halter73: We no longer need this skip condition, right? Though I guess we should remove them all in a separate commit.

test/Microsoft.AspNet.Server.KestrelTests/ConnectionFilterTests.cs
```diff
@@ -80,7 +81,36 @@ public class ConnectionFilterTests
...
+                    {
+                        await connection.SendEnd(
+                            "POST / HTTP/1.0",
+                            "",
+                            "Hello World?");
+                    }
+                    catch (IOException)
+                    {
+                        // Will throw because the exception in the connection filter will close the connection.
+                        Assert.True(true);
```
halter73: Did this test fail before your change? If I remember correctly, the entire server would crash in the scenario previously, so I would expect an exception from SendEnd before this change.

Maybe before TestServer.Dispose was throwing?

cesarbs: Before the change the xunit runner crashes along with the server :)

test/Microsoft.AspNet.Server.KestrelTests/ConnectionFilterTests.cs
cesarbs: Yeah, it looks like the problem is gone. I'd like to do a little more testing though.

test/Microsoft.AspNet.Server.KestrelTests/ConnectionFilterTests.cs
halter73: :laughing:

halter73:
Oops. I thought I gave this a :ship: last night. :shipit:

PR #408, opened 2015-11-19
Remove CreateLinkedTokenSource

### LINKED ISSUE #407: CreateLinkedTokenSource is high alloc
Using `CreateLinkedTokenSource` allocates more memory than the HttpContext! O_o

Token source

`_requestAbortCts = CancellationTokenSource.CreateLinkedTokenSource(_disconnectCts.Token);`

https://github.com/aspnet/KestrelHttpServer/blob/dev/src/Microsoft.AspNet.Server.Kestrel/Http/Frame.cs#L251

/cc @davidfowl @halter73

benaadams:
Resolves  #407

### DIFF
--- a/src/Microsoft.AspNet.Server.Kestrel/Http/Frame.cs
+++ b/src/Microsoft.AspNet.Server.Kestrel/Http/Frame.cs
@@ -199,7 +199,15 @@ namespace Microsoft.AspNet.Server.Kestrel.Http
                 ConnectionControl.End(ProduceEndType.SocketDisconnect);
                 SocketInput.AbortAwaiting();
 
-                _abortedCts?.Cancel();
+                try
+                {
+                    _abortedCts?.Cancel();
+                }
+                catch (ObjectDisposedException)
+                {
+                    // Don't log ODEs thrown from _abortedCts.Cancel()
+                    // If _abortedCts is disposed, the app has already completed.
+                }
             }
             catch (Exception ex)
             {
--- a/src/Microsoft.AspNet.Server.Kestrel/Http/Frame.cs
+++ b/src/Microsoft.AspNet.Server.Kestrel/Http/Frame.cs
@@ -43,8 +43,7 @@ namespace Microsoft.AspNet.Server.Kestrel.Http
         private Task _requestProcessingTask;
         private volatile bool _requestProcessingStopping; // volatile, see: https://msdn.microsoft.com/en-us/library/x13ttww7.aspx
         private volatile bool _requestAborted;
-        private CancellationTokenSource _disconnectCts = new CancellationTokenSource();
-        private CancellationTokenSource _requestAbortCts;
+        private CancellationTokenSource _disconnectOrAbortedCts = new CancellationTokenSource();
 
         private FrameRequestStream _requestBody;
         private FrameResponseStream _responseBody;
@@ -145,8 +144,6 @@ namespace Microsoft.AspNet.Server.Kestrel.Http
             }
 
             _prepareRequest?.Invoke(this);
-
-            _requestAbortCts?.Dispose();
         }
 
         public void ResetResponseHeaders()
@@ -199,7 +196,7 @@ namespace Microsoft.AspNet.Server.Kestrel.Http
                 ConnectionControl.End(ProduceEndType.SocketDisconnect);
                 SocketInput.AbortAwaiting();
 
-                _disconnectCts.Cancel();
+                _disconnectOrAbortedCts.Cancel();
             }
             catch (Exception ex)
             {
@@ -248,8 +245,7 @@ namespace Microsoft.AspNet.Server.Kestrel.Http
                         ResponseBody = _responseBody;
                         DuplexStream = new FrameDuplexStream(RequestBody, ResponseBody);
 
-                        _requestAbortCts = CancellationTokenSource.CreateLinkedTokenSource(_disconnectCts.Token);
-                        RequestAborted = _requestAbortCts.Token;
+                        RequestAborted = _disconnectOrAbortedCts.Token;
 
                         var httpContext = HttpContextFactory.Create(this);
                         try
@@ -302,7 +298,7 @@ namespace Microsoft.AspNet.Server.Kestrel.Http
             {
                 try
                 {
-                    _disconnectCts.Dispose();
+                    _disconnectOrAbortedCts.Dispose();
 
                     // If _requestAborted is set, the connection has already been closed.
                     if (!_requestAborted)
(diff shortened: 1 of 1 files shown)


src/Microsoft.AspNet.Server.Kestrel/Http/Frame.cs
```diff
@@ -248,8 +245,7 @@ public void Abort()
                         ResponseBody = _responseBody;
                         DuplexStream = new FrameDuplexStream(RequestBody, ResponseBody);
 
-                        _requestAbortCts = CancellationTokenSource.CreateLinkedTokenSource(_disconnectCts.Token);
-                        RequestAborted = _requestAbortCts.Token;
+                        RequestAborted = _disconnectOrAbortedCts.Token;
```
Tratcher: I think this is incorrect. RequestAborted needs to be unique per request, not per connection. We don't want registrations from one request to fire for the next request.

benaadams: Currently the behaviour is after the current request finishes it triggers the abort token; and if you actually call `Abort()` it aborts and disconnects?

Will change to maintain that behaviour

Tratcher: After a successful request the token should be disposed without triggering to discard all the registrations.

benaadams: Actually does it do anything particularly? Its only activated on `Abort` but that also disconnects

Tratcher: It should activate for disconnects and Abort, and Abort also disconnects. It tells the application to abandon any processing and unwind the middleware chain.

benaadams: K, might be looking at this wrong. Does the `_disconnectCts` do anything particularly beyond the `_requestAbortCts`?

halter73: @benaadams It avoids extra synchronization logic that might be required to avoid canceling an already disposed. `_requestAbortCts` to cancel when called from places like `Connection.Abort()`.

Turns out that there is still an issue with canceling a disposed `_disconnectCts`, but at least that isn't being disposed after every request.

benaadams: Switched it round?

Tratcher:
That might work. @halter73?

benaadams:
Need a `CancellationTokenSourceSlim` that can have its registrations reset; unfortunately  `CancellationTokenSource` is pretty heavyweight and `CancellationToken`'s constructor is internal and only takes `CancellationTokenSource` :(

benaadams:
Or a dump registrations on `CancellationToken`

benaadams:
Added https://github.com/dotnet/coreclr/issues/2113

src/Microsoft.AspNet.Server.Kestrel/Http/Frame.cs
```diff
@@ -199,12 +199,17 @@ public void Abort()
                 ConnectionControl.End(ProduceEndType.SocketDisconnect);
                 SocketInput.AbortAwaiting();
 
-                _disconnectCts.Cancel();
+                _abortedCts?.Cancel();
```
halter73: Maybe wrap this call specifically since we're not locking to avoid ODEs (however unlikely). This prevents us from logging errors that we know can happen, but the app developer can do nothing about.

        try
        {
            _abortedCts.Cancel();
        }
        catch (ObjectDisposedException)
        {
            // Don't log ODEs thrown from _abortedCts.Cancel()
            // If _disconnectCts is disposed, the app has already completed.
        }

halter73:
I think this changes behavior slightly. If Connection.Abort was called between requests and an app registered a callback with the RequestAborted token on the next request, the callback would be triggered immediately.

In the end, I don't think this matters since both the request and response are poisoned. Just thought it would be worth mentioning that the behavior changes slightly.

halter73:
The *next* request and response streams aren't poisoned I guess, but we still close the socket and poison SocketInput so I still think we're OK.

benaadams:
> If Connection.Abort was called between requests and an app registered a callback with the RequestAborted token on the next request, the callback would be triggered immediately.

Isn't that correct because that request if it managed to start would be correctly aborted? Or do you mean it should be receiving a different error?

halter73:
@benaadams That quote was me trying to describe the old behavior. I don't think the abort callback for the next request will be called after this change.

benaadams:
Ah kk

benaadams:
@stephentoub's added changes completely clear up the allocations

src/Microsoft.AspNet.Server.Kestrel/Http/Frame.cs
```diff
@@ -146,7 +185,8 @@ public void Reset()
 
             _prepareRequest?.Invoke(this);
 
-            _requestAbortCts?.Dispose();
+            _manuallySetRequestAbortToken = null;
+            _abortedCts = null;
```
halter73: Don't we need to to dispose the last _abortedCts (if it exists) to clean up registrations?

benaadams: /cc @stephentoub

stephentoub: No.  CancellationTokenSource.Dispose does a few things:
- Disposes of the ManualResetEvent that's lazily-allocated when the WaitHandle is accessed
- Disposes of the Timer that's created if the CT's time-based ctor or its CancelAfter method is used
- Disposes of the linked registrations created from a CTS created with CreateLinkedTokenSource

The third one isn't relevant as this CTS isn't being created with CreateLinkedTokenSource.
The second one isn't relevant as we're not using timers.
And while the third one is possibly relevant, it's a) rare that code these days uses CancellationToken.WaitHandle, and b) if it does get lazily-allocated, it'll still get cleaned up via the finalizer.

Given all that, it didn't seem worthwhile to Dispose of it and have to have other code that worried about delayed with the resulting race conditions on its use.

halter73: @stephentoub Thanks for the detailed explanation. :shipit:

src/Microsoft.AspNet.Server.Kestrel/Http/Frame.cs
```diff
@@ -93,8 +93,47 @@ public Frame(ConnectionContext context)
...
         public Stream DuplexStream { get; set; }
 
-        public CancellationToken RequestAborted { get; set; }
+        public CancellationToken RequestAborted
+        {
+            get
+            {
+                // If a request abort token was previously explicitly set, return it.
+                if (_manuallySetRequestAbortToken.HasValue)
+                    return _manuallySetRequestAbortToken.Value;
```
davidfowl: nit: style, always use braces

PR #410, opened 2015-11-20
Use MemoryPool2 for SocketOutput Buffers

halter73:
Add ProducingStart and ProducingComplete methods to ISocketOutput.
These new methods can help prevent double buffering when encoding.

### DIFF
--- a/src/Microsoft.AspNet.Server.Kestrel/Filter/StreamSocketOutput.cs
+++ b/src/Microsoft.AspNet.Server.Kestrel/Filter/StreamSocketOutput.cs
@@ -13,10 +13,13 @@ namespace Microsoft.AspNet.Server.Kestrel.Filter
     public class StreamSocketOutput : ISocketOutput
     {
         private readonly Stream _outputStream;
+        private readonly MemoryPool2 _memory;
+        private MemoryPoolBlock2 _producingBlock;
 
-        public StreamSocketOutput(Stream outputStream)
+        public StreamSocketOutput(Stream outputStream, MemoryPool2 memory)
         {
             _outputStream = outputStream;
+            _memory = memory;
         }
 
         void ISocketOutput.Write(ArraySegment<byte> buffer, bool immediate)
@@ -29,6 +32,28 @@ namespace Microsoft.AspNet.Server.Kestrel.Filter
             // TODO: Use _outputStream.WriteAsync
             _outputStream.Write(buffer.Array, buffer.Offset, buffer.Count);
             return TaskUtilities.CompletedTask;
+        }
+
+        public MemoryPoolIterator2 ProducingStart()
+        {
+            _producingBlock = _memory.Lease();
+            return new MemoryPoolIterator2(_producingBlock);
+        }
+
+        public void ProducingComplete(MemoryPoolIterator2 end, int count)
+        {
+            var block = _producingBlock;
+            while (block != end.Block)
+            {
+                _outputStream.Write(block.Data.Array, block.Data.Offset, block.Data.Count);
+
+                var returnBlock = block;
+                block = block.Next;
+                returnBlock.Pool?.Return(returnBlock);
+            }
+
+            _outputStream.Write(end.Block.Array, end.Block.Data.Offset, end.Index);
+            end.Block.Pool?.Return(end.Block);
         }
     }
 }
--- a/src/Microsoft.AspNet.Server.Kestrel/Http/SocketInput.cs
+++ b/src/Microsoft.AspNet.Server.Kestrel/Http/SocketInput.cs
@@ -64,7 +64,7 @@ namespace Microsoft.AspNet.Server.Kestrel.Http
                 {
                     _pinned = _tail;
                     var data = new ArraySegment<byte>(_pinned.Data.Array, _pinned.End, _pinned.Data.Offset + _pinned.Data.Count - _pinned.End);
-                    var dataPtr = _pinned.Pin();
+                    var dataPtr = _pinned.Pin() + _pinned.End;
                     return new IncomingBuffer
                     {
                         Data = data,
@@ -77,7 +77,7 @@ namespace Microsoft.AspNet.Server.Kestrel.Http
             return new IncomingBuffer
             {
                 Data = _pinned.Data,
-                DataPtr = _pinned.Pin()
+                DataPtr = _pinned.Pin() + _pinned.End
             };
         }
 
--- a/src/Microsoft.AspNet.Server.Kestrel/Http/SocketOutput.cs
+++ b/src/Microsoft.AspNet.Server.Kestrel/Http/SocketOutput.cs
@@ -23,8 +23,18 @@ namespace Microsoft.AspNet.Server.Kestrel.Http
         private readonly long _connectionId;
         private readonly IKestrelTrace _log;
 
+        // This locks all access to _tail, _isProducing and _returnFromOnProducingComplete.
+        // _head does not require a lock, since it is only used in the ctor and uv thread.
+        private readonly object _returnLock = new object();
+
+        private MemoryPoolBlock2 _head;
+        private MemoryPoolBlock2 _tail;
+
+        private bool _isProducing;
+        private MemoryPoolBlock2 _returnFromOnProducingComplete;
+
         // This locks access to to all of the below fields
-        private readonly object _lockObj = new object();
+        private readonly object _contextLock = new object();
 
         // The number of write operations that have been scheduled so far
         // but have not completed.
@@ -38,6 +48,7 @@ namespace Microsoft.AspNet.Server.Kestrel.Http
         public SocketOutput(
             KestrelThread thread,
             UvStreamHandle socket,
+            MemoryPool2 memory,
             Connection connection,
             long connectionId,
             IKestrelTrace log)
@@ -48,6 +59,9 @@ namespace Microsoft.AspNet.Server.Kestrel.Http
             _connectionId = connectionId;
             _log = log;
             _tasksPending = new Queue<TaskCompletionSource<object>>();
+
+            _head = memory.Lease();
+            _tail = _head;
         }
 
         public Task WriteAsync(
@@ -56,28 +70,20 @@ namespace Microsoft.AspNet.Server.Kestrel.Http
             bool socketShutdownSend = false,
             bool socketDisconnect = false)
         {
-            //TODO: need buffering that works
-            if (buffer.Array != null)
-            {
-                var copy = new byte[buffer.Count];
-                Array.Copy(buffer.Array, buffer.Offset, copy, 0, buffer.Count);
-                buffer = new ArraySegment<byte>(copy);
-                _log.ConnectionWrite(_connectionId, buffer.Count);
-            }
+            var tail = ProducingStart();
+            tail = tail.CopyFrom(buffer);
+            // We do our own accounting below
+            ProducingComplete(tail, count: 0);
 
             TaskCompletionSource<object> tcs = null;
 
-            lock (_lockObj)
+            lock (_contextLock)
             {
                 if (_nextWriteContext == null)
                 {
                     _nextWriteContext = new WriteContext(this);
                 }
 
-                if (buffer.Array != null)
-                {
-                    _nextWriteContext.Buffers.Enqueue(buffer);
-                }
                 if (socketShutdownSend)
                 {
                     _nextWriteContext.SocketShutdownSend = true;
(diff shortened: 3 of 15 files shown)


src/Microsoft.AspNet.Server.Kestrel/Filter/StreamSocketOutput.cs
```diff
@@ -30,5 +33,27 @@ Task ISocketOutput.WriteAsync(ArraySegment<byte> buffer, bool immediate, Cancell
...
+            _producingBlock = _memory.Lease();
+            return new MemoryPoolIterator2(_producingBlock);
+        }
+
+        public void ProducingComplete(MemoryPoolIterator2 end, int count)
+        {
+            var block = _producingBlock;
+            while (block != end.Block)
+            {
+                _outputStream.Write(block.Data.Array, block.Data.Offset, block.Data.Count);
```
davidfowl: Sync writes?

src/Microsoft.AspNet.Server.Kestrel/Filter/StreamSocketOutput.cs
```diff
@@ -30,5 +33,27 @@ Task ISocketOutput.WriteAsync(ArraySegment<byte> buffer, bool immediate, Cancell
...
             return TaskUtilities.CompletedTask;
         }
+
+        public MemoryPoolIterator2 ProducingStart()
+        {
+            _producingBlock = _memory.Lease();
+            return new MemoryPoolIterator2(_producingBlock);
+        }
+
+        public void ProducingComplete(MemoryPoolIterator2 end, int count)
```
davidfowl: Should this have an async version?

benaadams:
Not completely sure I follow, is the process:

-> ProducingStart
Add stuff to returned MemoryPoolIterator2 (including new blocks)
-> ProducingComplete

Not sure about ProducingComplete only taking end block though? Shouldn't it be taking the start block?

benaadams:
nvm Was reading the same mixed classes; is clearer now

src/Microsoft.AspNet.Server.Kestrel/Http/SocketInput.cs
```diff
@@ -64,7 +64,7 @@ public IncomingBuffer IncomingStart(int minimumSize)
                 {
                     _pinned = _tail;
                     var data = new ArraySegment<byte>(_pinned.Data.Array, _pinned.End, _pinned.Data.Offset + _pinned.Data.Count - _pinned.End);
-                    var dataPtr = _pinned.Pin();
+                    var dataPtr = _pinned.Pin() + _pinned.End;
```
davidfowl: So this was always wrong... Did you add a test for it?

benaadams: Changed the pin function in this PR to return a different value

davidfowl: Ah  missed that.

src/Microsoft.AspNet.Server.Kestrel/Http/SocketOutput.cs
```diff
@@ -138,6 +144,58 @@ public void End(ProduceEndType endType)
...
+            {
+                Debug.Assert(!_isProducing);
+                _isProducing = true;
+
+                if (_tail == null)
+                {
+                    throw new IOException("The socket has been closed.");
+                }
+
+                return new MemoryPoolIterator2(_tail, _tail.End);
```
benaadams: Shouldn't this behave like the `StreamSocketOutput.ProducingStart()`; use no lock and return a `MemoryPoolIterator2` with a fresh LeasedBlock so its disconnected? Rather than essentially returning the active queue.

src/Microsoft.AspNet.Server.Kestrel/Http/SocketOutput.cs
```diff
@@ -138,6 +144,58 @@ public void End(ProduceEndType endType)
...
+            }
+        }
+
+        public void ProducingComplete(MemoryPoolIterator2 end, int count)
```
benaadams: Following on from above; take `start` and `end` then something like
```csharp
_tail.Next = start
_tail = end;
```

benaadams: So the creation of the next segment in the chain is independent of what's already in the chain and the socket is working on

src/Microsoft.AspNet.Server.Kestrel/Http/SocketOutput.cs
```diff
@@ -348,13 +432,62 @@ public void DoDisconnectIfNeeded()
...
+                {
+                    var returnBlock = block;
+                    block = block.Next;
+
+                    returnBlock.Unpin();
+                    returnBlock.Pool?.Return(returnBlock);
+                }
+            }
+
+            private void LockWrite()
```
benaadams: Not sure how this inert the blocks currently passed out on ProducingStart as part of the MemoryPoolIterator2

src/Microsoft.AspNet.Server.Kestrel/Http/SocketOutput.cs
benaadams: e.g. lock isn't maintained after return

src/Microsoft.AspNet.Server.Kestrel/Http/SocketOutput.cs
benaadams: Then its just appended to the end of the current chain, if it hasn't been sent yet.

src/Microsoft.AspNet.Server.Kestrel/Http/SocketOutput.cs
benaadams: This should set `_head` and `_tail` to null?

src/Microsoft.AspNet.Server.Kestrel/Http/SocketOutput.cs
```diff
@@ -48,6 +59,9 @@ public class SocketOutput : ISocketOutput
             _connectionId = connectionId;
             _log = log;
             _tasksPending = new Queue<TaskCompletionSource<object>>();
+
+            _head = memory.Lease();
```
benaadams: might not need to lease a block here, if you go with `null` being nothing to write.

src/Microsoft.AspNet.Server.Kestrel/Filter/StreamSocketOutput.cs
halter73: `StreamSocketOutput.WriteAsync` also does sync writes. There is a TODO comment to use WriteAsync. We need to figure out why that breaks HTTPS.

src/Microsoft.AspNet.Server.Kestrel/Http/SocketOutput.cs
halter73: The lock here is not to protect against overlapping writes from user code.

This is to prevent NullRefExceptions when the socket is closed on the uv thread (which calls ReturnAllBlocks and nulls out _tail) during the call to ProducingStart.

src/Microsoft.AspNet.Server.Kestrel/Filter/StreamSocketOutput.cs
halter73: I think in practice this will only be used for what were `immediate: false` writes. We can certainly make an async version of this if we find the need.

src/Microsoft.AspNet.Server.Kestrel/Http/SocketOutput.cs
halter73: I almost wrote this to do lazy leasing, but decided against it after I figured that we should almost never be creating a new SocketOutput without writing to it. What I hadn't really taken into account is the potential negative perf impact of leasing blocks longer than necessary.

Currently `null` means the socket has been closed, but I guess I can add a bool to track that instead.

src/Microsoft.AspNet.Server.Kestrel/Http/SocketOutput.cs
halter73: The design is to do your own appending that potentially links multiple blocks (like CopyFrom does). Once your done appending blocks, you simply pass back an iterator pointing to the end of the response data that was just appended.

Note that this design *requires* that you fill the entire previous block before linking to the `Next` block. I originally had a bunch of `Debug.Asserts` in `LockWrite` to verify this is the case, but decided this was a little noisy.

src/Microsoft.AspNet.Server.Kestrel/Http/SocketOutput.cs
halter73: This should not set `_head` and `_tail` to null. Those should only be set to null after the socket is closed.

I don't think I understand your first question.

benaadams: There are two threads operating on the block chain? libuv and the writer? How do they coordinate?

src/Microsoft.AspNet.Server.Kestrel/Infrastructure/MemoryPoolBlock2.cs
```diff
@@ -98,24 +98,27 @@ protected MemoryPoolBlock2()
...
         {
-            Debug.Assert(!_pinHandle.IsAllocated);
-
             if (_dataArrayPtr != IntPtr.Zero)
             {
                 // this is a slab managed block - use the native address of the slab which is always locked
-                return _dataArrayPtr + End;
+                return _dataArrayPtr;
+            }
+            else if (_pinHandle.IsAllocated)
```
lodejard: Is this to enable Pin to be called more than once on the same block? Doesn't that imply Pin/Unpin calls might be nested? If that's the case there should probably be a pin-count tracked so Unpin only frees the GCHandle when the pin-count is reduced to zero.

src/Microsoft.AspNet.Server.Kestrel/Http/SocketOutput.cs
halter73: Only the writer can advance the _tail. Only the libuv thread can advance _head.

The libuv thread writes all the data between _head and a snapshot it takes of _tail. Once the write call is made _head is advanced to the _tail snapshot (_lockedEnd). Once the write is complete, all the blocks between the original _head (_lockedStart) and the _tailSnapshot (excluding the _tail block which might still be written to) are returned to the pool.

The socket is closed from the libuv thread. At this time it will take the _returnLock and set _head and _tail to _null after returning all blocks between _head and _tail inclusive.

If the writer is currently between a call to ProducingStart and ProducingComplete when the socket is closed, the _tail will not be immediately returned to the pool. Instead, what was the _tail will be stored in _returnFromOnProducingComplete. When ProducingComplete is then called, all the blocks starting _returnFromOnProducingComplete will be immediately returned to the pool.

benaadams: K :+1: 

Almost finished merging

src/Microsoft.AspNet.Server.Kestrel/Infrastructure/MemoryPoolBlock2.cs
halter73: Unpin is still only called once for each block (notice that `Debug.Assert(_pinHandle.IsAllocated);` is still in Unpin).

Pin can called multiple times because it is simpler that finding storing the address externally of blocks which are written multiple times (Only non-overlapping segments of the same block are written of course). The block is only returned once _head moves past the block.

halter73: OK I'm changing this to just Unpin all written blocks, not just the blocks that are being returned, when writes complete.

This simplifies the code to boot.

src/Microsoft.AspNet.Server.Kestrel/Http/SocketOutput.cs
lodejard: Yeah, would be nice if the SocketOutput could return itself to a `_head == _tail == null` state when there's known to be neither writes nor ProducingStart in effect at the time. Would help reduce memory pressure from long-running responses (web sockets or server-sent events) and reduce memory pressure for open connections waiting for the next request to arrive.

src/Microsoft.AspNet.Server.Kestrel/Http/SocketOutput.cs
lodejard: Actually it might not literally require all of the blocks to be filled... As long as the block.End is updated in the iter.CopyFrom method, the SocketOutput can use the block .Begin and .End properties to know what regions are occupied by the data to be written

benaadams:
@halter73 is this used correctly in https://github.com/aspnet/KestrelHttpServer/pull/367 ?

src/Microsoft.AspNet.Server.Kestrel/Filter/StreamSocketOutput.cs
```diff
@@ -30,5 +33,27 @@ Task ISocketOutput.WriteAsync(ArraySegment<byte> buffer, bool immediate, Cancell
...
+            while (block != end.Block)
+            {
+                _outputStream.Write(block.Data.Array, block.Data.Offset, block.Data.Count);
+
+                var returnBlock = block;
+                block = block.Next;
+                returnBlock.Pool?.Return(returnBlock);
+            }
+
+            _outputStream.Write(end.Block.Array, end.Block.Data.Offset, end.Index);
```
benaadams: end.Index isn't right here is it? Should be a count? end.Index - end.Block.Data.Offset?

halter73: Correct. Guess I need to add some StreamSocketOutputTests.

benaadams: Fixed in  #367 btw

Issue #214, opened 2015-09-19
ServerAddresses should support base paths


davidfowl:
http://localhost:5000/somebasePath

Should make PathBase = someBasePath

muratg:
@davidfowl In RC1? Priority?

Tratcher:
This is required to support IIS virtual directories https://github.com/aspnet/IISIntegration/issues/14

developer1998:
For production environments this is a must to have the sites functioning as Applications under a domain. Thank you for your efforts.

PR #412, opened 2015-11-20
Support PathBase (#214).

cesarbs:
Using `ServerAddress.Path` as PathBase. Had to change ServerAddress to not allow `Path` to be `/` or end with a `/`, since we don't want that for PathBase.

I'm in doubt about the difference between a request to `/foo` and `/foo/` when PathBase is `/foo`. Should Path be empty in the first case (which OWIN allows) and `/` in the second case, or should it be `/` in both cases? I tried forcing it to be `/` in both cases but that caused some functional tests to fail.

I'd appreciate some input on this.

#214

cc @Tratcher @halter73 @davidfowl

### DIFF
--- a/samples/SampleApp/Startup.cs
+++ b/samples/SampleApp/Startup.cs
@@ -48,6 +48,10 @@ namespace SampleApp
                     context.Request.PathBase,
                     context.Request.Path,
                     context.Request.QueryString);
+                Console.WriteLine($"Method: {context.Request.Method}");
+                Console.WriteLine($"PathBase: {context.Request.PathBase}");
+                Console.WriteLine($"Path: {context.Request.Path}");
+                Console.WriteLine($"QueryString: {context.Request.QueryString}");
 
                 var connectionFeature = context.Connection;
                 Console.WriteLine($"Peer: {connectionFeature.RemoteIpAddress?.ToString()} {connectionFeature.RemotePort}");
--- a/src/Microsoft.AspNet.Server.Kestrel/Http/Frame.cs
+++ b/src/Microsoft.AspNet.Server.Kestrel/Http/Frame.cs
@@ -11,7 +11,6 @@ using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.AspNet.Http;
 using Microsoft.AspNet.Http.Features;
-using Microsoft.AspNet.Server.Kestrel.Filter;
 using Microsoft.AspNet.Server.Kestrel.Infrastructure;
 using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Primitives;
@@ -70,6 +69,8 @@ namespace Microsoft.AspNet.Server.Kestrel.Http
         private readonly IPEndPoint _remoteEndPoint;
         private readonly Action<IFeatureCollection> _prepareRequest;
 
+        private readonly string _pathBase;
+
         public Frame(ConnectionContext context)
             : this(context, remoteEndPoint: null, localEndPoint: null, prepareRequest: null)
         {
@@ -84,6 +85,7 @@ namespace Microsoft.AspNet.Server.Kestrel.Http
             _remoteEndPoint = remoteEndPoint;
             _localEndPoint = localEndPoint;
             _prepareRequest = prepareRequest;
+            _pathBase = context.ServerAddress.PathBase;
 
             FrameControl = this;
             Reset();
@@ -92,6 +94,7 @@ namespace Microsoft.AspNet.Server.Kestrel.Http
         public string Scheme { get; set; }
         public string Method { get; set; }
         public string RequestUri { get; set; }
+        public string PathBase { get; set; }
         public string Path { get; set; }
         public string QueryString { get; set; }
         public string HttpVersion
@@ -198,6 +201,7 @@ namespace Microsoft.AspNet.Server.Kestrel.Http
             Scheme = null;
             Method = null;
             RequestUri = null;
+            PathBase = null;
             Path = null;
             QueryString = null;
             _httpVersion = HttpVersionType.Unknown;
@@ -809,13 +813,49 @@ namespace Microsoft.AspNet.Server.Kestrel.Http
                 RequestUri = requestUrlPath;
                 QueryString = queryString;
                 HttpVersion = httpVersion;
-                Path = RequestUri;
+
+                bool caseMatches;
+
+                if (!string.IsNullOrEmpty(_pathBase) &&
+                    (requestUrlPath.Length == _pathBase.Length || (requestUrlPath.Length > _pathBase.Length && requestUrlPath[_pathBase.Length] == '/')) &&
+                    RequestUrlStartsWithPathBase(requestUrlPath, out caseMatches))
+                {
+                    PathBase = caseMatches ? _pathBase : requestUrlPath.Substring(0, _pathBase.Length);
+                    Path = requestUrlPath.Substring(_pathBase.Length);
+                }
+                else
+                {
+                    Path = requestUrlPath;
+                }
+
                 return true;
             }
             finally
             {
                 input.ConsumingComplete(consumed, scan);
             }
+        }
+
+        private bool RequestUrlStartsWithPathBase(string requestUrl, out bool caseMatches)
+        {
+            caseMatches = true;
+
+            for (var i = 0; i < _pathBase.Length; i++)
+            {
+                if (requestUrl[i] != _pathBase[i])
+                {
+                    if (char.ToLowerInvariant(requestUrl[i]) == char.ToLowerInvariant(_pathBase[i]))
+                    {
+                        caseMatches = false;
+                    }
+                    else
+                    {
+                        return false;
+                    }
+                }
+            }
+
+            return true;
         }
 
         public static bool TakeMessageHeaders(SocketInput input, FrameRequestHeaders requestHeaders)
--- a/src/Microsoft.AspNet.Server.Kestrel/Infrastructure/MemoryPoolIterator2Extensions.cs
+++ b/src/Microsoft.AspNet.Server.Kestrel/Infrastructure/MemoryPoolIterator2Extensions.cs
@@ -6,7 +6,7 @@ using System.Text;
 
 namespace Microsoft.AspNet.Server.Kestrel.Infrastructure
 {
-    public static class MemoryPoolIterator2Extenstions
+    public static class MemoryPoolIterator2Extensions
     {
         private const int _maxStackAllocBytes = 16384;
 
--- a/test/Microsoft.AspNet.Server.Kestrel.FunctionalTests/PathBaseTests.cs
+++ b/test/Microsoft.AspNet.Server.Kestrel.FunctionalTests/PathBaseTests.cs
@@ -1,0 +1,113 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Collections.Generic;
+using System.Net.Http;
+using System.Threading.Tasks;
+using Microsoft.AspNet.Builder;
+using Microsoft.AspNet.Hosting;
+using Microsoft.AspNet.Http;
+using Microsoft.AspNet.Testing.xunit;
+using Microsoft.Extensions.Configuration;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
+using Xunit;
+
+namespace Microsoft.AspNet.Server.Kestrel.FunctionalTests
+{
+    public class PathBaseTests
+    {
+        [ConditionalTheory]
+        [InlineData("http://localhost:8791/base", "http://localhost:8791/base", "/base", "")]
+        [InlineData("http://localhost:8791/base", "http://localhost:8791/base/", "/base", "/")]
+        [InlineData("http://localhost:8791/base", "http://localhost:8791/base/something", "/base", "/something")]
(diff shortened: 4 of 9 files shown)


samples/SampleApp/Startup.cs
```diff
@@ -39,7 +39,7 @@ public void Configure(IApplicationBuilder app, ILoggerFactory loggerFactory, IAp
                 Console.WriteLine("Could not find certificate at '{0}'. HTTPS is not enabled.", testCertPath);
             }
 
-            app.UseKestrelConnectionLogging();
+            //app.UseKestrelConnectionLogging();
```
cesarbs: Revert.

samples/SampleApp/project.json
```diff
@@ -15,8 +15,8 @@
         }
     },
     "commands": {
-        "web": "Microsoft.AspNet.Hosting --server Microsoft.AspNet.Server.Kestrel --server.urls http://localhost:5000;https://localhost:5001",
-        "kestrel": "Microsoft.AspNet.Server.Kestrel --server.urls http://localhost:5000;https://localhost:5001",
+        "web": "Microsoft.AspNet.Hosting --server Microsoft.AspNet.Server.Kestrel --server.urls http://localhost:5000/base;https://localhost:5001",
```
cesarbs: Revert.

cesarbs:
Follow up on my original comment: WebListener sets `Path` to `/` when the request URL ends with a trailing slash, but leaves it empty when it doesn't. That's the current behavior I have for Kestrel in this PR.

src/Microsoft.AspNet.Server.Kestrel/Http/Frame.cs
```diff
@@ -744,7 +749,17 @@ private bool TakeStartLine(SocketInput input)
                 RequestUri = requestUrlPath;
                 QueryString = queryString;
                 HttpVersion = httpVersion;
-                Path = RequestUri;
+
+                if (!string.IsNullOrEmpty(_pathBase) && requestUrlPath.StartsWith(_pathBase))
```
Tratcher: case insensitive compare, then make sure you set PathBase to the requestUrlPath substring to preserve the user casing.

Also, StartsWith isn't good enough. "/foobar" starts with "/foo".

test/Microsoft.AspNet.Server.Kestrel.FunctionalTests/PathBaseTests.cs
```diff
@@ -0,0 +1,97 @@
...
+                            Path = context.Request.Path.Value
+                        }));
+                    });
+                });
+
+            using (var app = builder.Build().Start())
+            {
+                using (var client = new HttpClient())
+                {
+                    var response = await client.GetAsync(requestAddress);
```
Tratcher: HttpClient will always turn http://localhost to http://localhost/, so you can't test that senario here.

src/Microsoft.AspNet.Server.Kestrel/Http/Frame.cs
```diff
@@ -744,7 +751,19 @@ private bool TakeStartLine(SocketInput input)
...
                 QueryString = queryString;
                 HttpVersion = httpVersion;
-                Path = RequestUri;
+
+                if (!string.IsNullOrEmpty(_pathBase) &&
+                    (requestUrlPath.Equals(_pathBase, StringComparison.OrdinalIgnoreCase) ||
+                     requestUrlPath.StartsWith(_pathBaseWithSlash, StringComparison.OrdinalIgnoreCase)))
+                {
+                    PathBase = requestUrlPath.Substring(0, _pathBase.Length);
+                    Path = requestUrlPath.Substring(_pathBase.Length);
```
cesarbs: I'm going to rewrite this to use iterators and also avoid scanning the string multiple times.

src/Microsoft.AspNet.Server.Kestrel/Http/Frame.cs
```diff
@@ -744,7 +751,50 @@ private bool TakeStartLine(SocketInput input)
                 RequestUri = requestUrlPath;
                 QueryString = queryString;
                 HttpVersion = httpVersion;
-                Path = RequestUri;
+
+                var pathBaseEnd = pathBegin;
```
cesarbs: This is not the final revision of it - I just pushed so I can continue working on it later from somewhere else.

cesarbs:
Ping.

src/Microsoft.AspNet.Server.Kestrel/Http/Frame.cs
```diff
@@ -71,6 +72,7 @@ public Frame(ConnectionContext context)
             _remoteEndPoint = remoteEndPoint;
             _localEndPoint = localEndPoint;
             _prepareRequest = prepareRequest;
+            _pathBase = context?.ServerAddress?.PathBase;
```
davidfowl: Can those really all be null?

src/Microsoft.AspNet.Server.Kestrel/Infrastructure/MemoryPoolIterator2Extensions.cs
```diff
@@ -203,5 +203,31 @@ public static ArraySegment<byte> GetArraySegment(this MemoryPoolIterator2 start,
             start.CopyTo(array, 0, length, out length);
             return new ArraySegment<byte>(array, 0, length);
         }
+
+        public static bool StartsWithIgnoreCase(this MemoryPoolIterator2 start, string prefix, out MemoryPoolIterator2 end, out bool caseMatches)
```
davidfowl: You may want to profile this, we went through extreme lengths to avoid stuff like this on the per request path (in kestrel at least) :smile:

Tratcher:
:+1: on the tests. Good luck with the perf. /cc: @benaadams

src/Microsoft.AspNet.Server.Kestrel/Http/Frame.cs
halter73: I'm fairly certain (maybe with the exception of some tests) that neither context nor ServerAddress can be null.

src/Microsoft.AspNet.Server.Kestrel/Infrastructure/MemoryPoolIterator2Extensions.cs
```diff
@@ -203,5 +203,31 @@ public static ArraySegment<byte> GetArraySegment(this MemoryPoolIterator2 start,
...
+            caseMatches = true;
+
+            var i = 0;
+            for (i = 0; i < prefix.Length; i++)
+            {
+                var c = end.Take();
+
+                if (c != prefix[i])
+                {
+                    if (char.ToLowerInvariant((char)c) == char.ToLowerInvariant(prefix[i]))
```
halter73: I wonder if you could improve perf for the ASCII case by writing your own `ToLowerInvariantAscii` or something like that.

benaadams: `Vector.BitwiseOr(new Vector<byte>(0x20), v)` ...

src/Microsoft.AspNet.Server.Kestrel/Infrastructure/MemoryPoolIterator2Extensions.cs
halter73: :+1: for profiling. @davidfowl Do you have any specific ideas for improving the perf of these PR?

halter73:
:+1: on the tests

src/Microsoft.AspNet.Server.Kestrel/Infrastructure/MemoryPoolIterator2Extensions.cs
cesarbs: @halter73 char.ToLowerInvariant already optimizes the call when it's ASCII. It ends up here:

http://referencesource.microsoft.com/#mscorlib/system/globalization/textinfo.cs,145c4c32b3b4bace

src/Microsoft.AspNet.Server.Kestrel/Http/Frame.cs
```diff
@@ -744,7 +748,31 @@ private bool TakeStartLine(SocketInput input)
...
                 QueryString = queryString;
                 HttpVersion = httpVersion;
-                Path = RequestUri;
+
+                MemoryPoolIterator2 pathBaseEnd;
+                bool caseMatches;
+
+                if (!string.IsNullOrEmpty(_pathBase) &&
+                    requestUrlPath.Length >= _pathBase.Length &&
+                    pathBegin.StartsWithIgnoreCase(_pathBase, out pathBaseEnd, out caseMatches) &&
```
benaadams: if _pathBase is a fixed string/char array; isn't it excessive to create a MemoryPoolIterator2 out of it for the purpose of comparison? Could also create a _pathBaseLower at Frame initialisation time; rather than doing it on each request per char again and again?

src/Microsoft.AspNet.Server.Kestrel/Infrastructure/MemoryPoolIterator2Extensions.cs
```diff
@@ -203,5 +203,31 @@ public static ArraySegment<byte> GetArraySegment(this MemoryPoolIterator2 start,
...
             return new ArraySegment<byte>(array, 0, length);
         }
+
+        public static bool StartsWithIgnoreCase(this MemoryPoolIterator2 start, string prefix, out MemoryPoolIterator2 end, out bool caseMatches)
+        {
+            end = start;
+            caseMatches = true;
+
+            var i = 0;
+            for (i = 0; i < prefix.Length; i++)
```
davidfowl: Does the iterator have a way of scanning length chars?

src/Microsoft.AspNet.Server.Kestrel/Infrastructure/MemoryPoolIterator2Extensions.cs
halter73: @cesarbs that's a little more efficient than I thought, but that checks if each character is in the ascii range, which might be unnecessary. Could we try @benaadams suggestion and vectorize the operation to see if it performs better?

src/Microsoft.AspNet.Server.Kestrel/Http/Frame.cs
halter73: We would like to know if the case matches too. If so, we don't need to allocate a new string and can just do `PathBase = _pathBase`. Otherwise, we have to create a new correctly cased string.

src/Microsoft.AspNet.Server.Kestrel/Infrastructure/MemoryPoolIterator2Extensions.cs
halter73: I don't think that exists yet. How do you expect that to work? You pass a length and get an `IEnumerable<char>`? Something callback based?

src/Microsoft.AspNet.Server.Kestrel/Infrastructure/MemoryPoolIterator2Extensions.cs
Tratcher: Keep in mind that the path has already been un-escaped, so `c` may be utf-8 bytes. @cesarbs add a test for base paths that contain unicode characters: `/b𤭢se`

halter73: Correct, but only if `needDecode` is true. We can have a special ASCII path if we want to.

src/Microsoft.AspNet.Server.Kestrel/Infrastructure/MemoryPoolIterator2Extensions.cs
davidfowl: `ArraySegment<byte>`

halter73: Would a new array be allocated if the chars traverse blocks?

davidfowl: I guess that would kinda suck huh :smile:

cesarbs:
I've stepped away from using iterators to do the path base comparison. The code started looking quite messy when I included UTF-8 characters in the game.

Since I already have a decoded string for the whole request URL, I'm simply using that to compare against _pathBase. I did a quick check to make sure that the new code doesn't significantly impact our requests per seconds and it's not the case.

I see little advantage in using iterators anyways, since what we're trying to optimize for is reduced string allocations. With iterators or not we still have to allocate a new string if the path base in the request doesn't match case with the registered path base. And we will always have to allocate a new string for the path.

halter73:
Squash this.

cesarbs:
@halter73 Done :+1:

halter73:
@davidfowl @cesarbs explained to me that doing an iterator based comparison against pathbase made no measurable improvement in RPS (using wrk) over the PR as it currently stands.

Given your aversion to complicating the codebase for no rps/memory gains, I assume your fine with taking this PR as is? Technically, we might be able to reduce allocations if the full path is exactly equal to the PathBase, but I'm not sure if that is worth special casing to remove one string allocation per impacted request.

We could always look at improving the performance of this down the road.

halter73:
:shipit:

davidfowl:
:shipit: 

/cc @DamianEdwards measure when this is merged

PR #415, opened 2015-11-22
loopback fast-path and zero-copy send (Windows)

benaadams:
Have to do it this way due to 2 open issues in libuv

Loopback fast path: https://github.com/libuv/libuv/issues/489
Winsock send buffer 0: https://github.com/libuv/libuv/issues/74

Currently copy the data 3 times: app -> bytes, bytes -> socketoutput, socketoutput -> winsock
This drops the last copy, can reduce the second copy after #367 is merged.


benaadams:
Related: https://github.com/aspnet/IISIntegration/issues/29

PR #416, opened 2015-11-22
Enable TCP Loopback Fast Path (Windows)

### LINKED ISSUE #420: Use TCP Loopback Fast Path on Windows
As the recommended hosting model is to front Kestrel with a reverse proxy server (e.g. HttpPlatformHandler+IIS) which works over localhost, use the TCP Loopback Fast Path on Windows; where it skips the IP layer. 

TCP Loopback Fast Path behaviour reference: http://blogs.technet.com/b/wincat/archive/2012/12/05/fast-tcp-loopback-performance-and-low-latency-with-windows-server-2012-tcp-loopback-fast-path.aspx

The only wrinkle is both end-points need the fast path enabled to receive any advantage - so this may need changes to HttpPlatformHandler if it doesn't already have it enabled. 

There appears to be no disadvantage to having it enabled but not used (e.g. other socket not enabled). 

When it is used the disadvantage is it loops back further up the stack so fewer packsniffers can listen in; depending on where they insert themselves in the stack. Though localhost is problematic for most of them anyway as they generally listen at the inet level.

Related: *"Confirm HttpPlatformHandler uses Fast TCP Loopback"* aspnet/IISIntegration#29

benaadams:
Enable TCP Loopback Fast Path; where the IP layer is skipped, for lower latency for localhost comms. For example when fronted by the HttpPlatformHandler+IIS reverse proxy

TCP Loopback Fast Path reference: http://blogs.technet.com/b/wincat/archive/2012/12/05/fast-tcp-loopback-performance-and-low-latency-with-windows-server-2012-tcp-loopback-fast-path.aspx

Recommended work around for open issue in libuv *"Enable Fast TCP Loopback on Win"* libuv/libuv#489

Related: *"Confirm HttpPlatformHandler uses Fast TCP Loopback"* aspnet/IISIntegration#29

Resolves #420

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Kestrel/Networking/Libuv.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/Networking/Libuv.cs
@@ -27,6 +27,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Networking
                 _uv_stop = NativeDarwinMonoMethods.uv_stop;
                 _uv_ref = NativeDarwinMonoMethods.uv_ref;
                 _uv_unref = NativeDarwinMonoMethods.uv_unref;
+                _uv_fileno = NativeDarwinMonoMethods.uv_fileno;
                 _uv_close = NativeDarwinMonoMethods.uv_close;
                 _uv_async_init = NativeDarwinMonoMethods.uv_async_init;
                 _uv_async_send = NativeDarwinMonoMethods.uv_async_send;
@@ -68,6 +69,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Networking
                 _uv_stop = NativeMethods.uv_stop;
                 _uv_ref = NativeMethods.uv_ref;
                 _uv_unref = NativeMethods.uv_unref;
+                _uv_fileno = NativeMethods.uv_fileno;
                 _uv_close = NativeMethods.uv_close;
                 _uv_async_init = NativeMethods.uv_async_init;
                 _uv_async_send = NativeMethods.uv_async_send;
@@ -178,6 +180,15 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Networking
         }
 
         [UnmanagedFunctionPointer(CallingConvention.Cdecl)]
+        protected delegate int uv_fileno_func(UvHandle handle, ref IntPtr socket);
+        protected uv_fileno_func _uv_fileno;
+        public int uv_fileno(UvHandle handle, ref IntPtr socket)
+        {
+            handle.Validate();
+            return Check(_uv_fileno(handle, ref socket));
+        }
+
+        [UnmanagedFunctionPointer(CallingConvention.Cdecl)]
         public delegate void uv_close_cb(IntPtr handle);
         protected Action<IntPtr, uv_close_cb> _uv_close;
         public void close(UvHandle handle, uv_close_cb close_cb)
@@ -221,6 +232,40 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Networking
         {
             handle.Validate();
             Check(_uv_tcp_bind(handle, ref addr, flags));
+            if (PlatformApis.IsWindows)
+            {
+                tcp_bind_windows_extras(handle);
+            }
+        }
+
+        private unsafe void tcp_bind_windows_extras(UvTcpHandle handle)
+        {
+            const int SIO_LOOPBACK_FAST_PATH = -1744830448; // IOC_IN | IOC_WS2 | 16;
+            const int WSAEOPNOTSUPP = 10000 + 45; // (WSABASEERR+45)
+            const int SOCKET_ERROR = -1;
+
+            var socket = IntPtr.Zero;
+            Check(_uv_fileno(handle, ref socket));
+
+            // Enable loopback fast-path for lower latency for localhost comms, like HttpPlatformHandler fronting
+            // http://blogs.technet.com/b/wincat/archive/2012/12/05/fast-tcp-loopback-performance-and-low-latency-with-windows-server-2012-tcp-loopback-fast-path.aspx
+            // https://github.com/libuv/libuv/issues/489
+            var optionValue = 1;
+            uint dwBytes = 0u;
+
+            var result = NativeMethods.WSAIoctl(socket, SIO_LOOPBACK_FAST_PATH, &optionValue, sizeof(int), null, 0, out dwBytes, IntPtr.Zero, IntPtr.Zero);
+            if (result == SOCKET_ERROR)
+            {
+                var errorId = NativeMethods.WSAGetLastError();
+                if (errorId == WSAEOPNOTSUPP)
+                {
+                    // This system is not >= Windows Server 2012, and the call is not supported.
+                }
+                else
+                {
+                    Check(errorId);
+                }
+            }
         }
 
         protected Func<UvTcpHandle, IntPtr, int> _uv_tcp_open;
@@ -502,6 +547,9 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Networking
             public static extern void uv_unref(UvHandle handle);
 
             [DllImport("libuv", CallingConvention = CallingConvention.Cdecl)]
+            public static extern int uv_fileno(UvHandle handle, ref IntPtr socket);
+
+            [DllImport("libuv", CallingConvention = CallingConvention.Cdecl)]
             public static extern void uv_close(IntPtr handle, uv_close_cb close_cb);
 
             [DllImport("libuv", CallingConvention = CallingConvention.Cdecl)]
@@ -587,6 +635,22 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Networking
 
             [DllImport("libuv", CallingConvention = CallingConvention.Cdecl)]
             unsafe public static extern int uv_walk(UvLoopHandle loop, uv_walk_cb walk_cb, IntPtr arg);
+
+            [DllImport("WS2_32.dll", CallingConvention = CallingConvention.Winapi)]
+            unsafe public static extern int WSAIoctl(
+                IntPtr socket,
+                int dwIoControlCode,
+                int* lpvInBuffer,
+                uint cbInBuffer,
+                int* lpvOutBuffer,
+                int cbOutBuffer,
+                out uint lpcbBytesReturned,
+                IntPtr lpOverlapped,
+                IntPtr lpCompletionRoutine
+            );
+            
+            [DllImport("WS2_32.dll", CallingConvention = CallingConvention.Winapi)]
+            public static extern int WSAGetLastError();
         }
 
         private static class NativeDarwinMonoMethods
@@ -608,6 +672,9 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Networking
 
             [DllImport("__Internal", CallingConvention = CallingConvention.Cdecl)]
             public static extern void uv_unref(UvHandle handle);
+
+            [DllImport("__Internal", CallingConvention = CallingConvention.Cdecl)]
+            public static extern int uv_fileno(UvHandle handle, ref IntPtr socket);
 
             [DllImport("__Internal", CallingConvention = CallingConvention.Cdecl)]
             public static extern void uv_close(IntPtr handle, uv_close_cb close_cb);
--- a/test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/TestConnection.cs
+++ b/test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/TestConnection.cs
@@ -1,0 +1,35 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Net;
+using System.Net.Sockets;
+using Microsoft.AspNetCore.Server.Kestrel.Networking;
+
+namespace Microsoft.AspNetCore.Server.Kestrel.FunctionalTests
+{
+    public class TestConnection
+    {
+        public static Socket CreateConnectedLoopbackSocket(int port)
+        {
+            var socket = new Socket(AddressFamily.InterNetwork, SocketType.Stream, ProtocolType.Tcp);
+            if (PlatformApis.IsWindows)
+            {
+                const int SIO_LOOPBACK_FAST_PATH = -1744830448;
+                var optionInValue = BitConverter.GetBytes(1);
+                try
+                {
+                    socket.IOControl(SIO_LOOPBACK_FAST_PATH, optionInValue, null);
+                }
+                catch
+                {
+                    // If the operating system version on this machine did
+                    // not support SIO_LOOPBACK_FAST_PATH (i.e. version
+                    // prior to Windows 8 / Windows Server 2012), handle the exception
+                }
+            }
+            socket.Connect(new IPEndPoint(IPAddress.Loopback, port));
+            return socket;
+        }
+    }
+}
--- a/test/Microsoft.AspNetCore.Server.KestrelTests/EngineTests.cs
+++ b/test/Microsoft.AspNetCore.Server.KestrelTests/EngineTests.cs
@@ -118,24 +118,7 @@ namespace Microsoft.AspNetCore.Server.KestrelTests
             var started = engine.CreateServer(address);
 
             Console.WriteLine("Started");
-            var socket = new Socket(AddressFamily.InterNetwork, SocketType.Stream, ProtocolType.Tcp);
-            if (PlatformApis.IsWindows)
-            {
-                const int SIO_LOOPBACK_FAST_PATH = (-1744830448);
-                var optionInValue = BitConverter.GetBytes(1);
-                try
-                {
-                    socket.IOControl(SIO_LOOPBACK_FAST_PATH, optionInValue, null);
-                }
-                catch
-                {
-                    // If the operating system version on this machine did
-                    // not support SIO_LOOPBACK_FAST_PATH (i.e. version
-                    // prior to Windows 8 / Windows Server 2012), handle the exception
-                }
-            }
-            socket.NoDelay = true;
-            socket.Connect(new IPEndPoint(IPAddress.Loopback, port));
+            var socket = TestConnection.CreateConnectedLoopbackSocket(port);
             socket.Send(Encoding.ASCII.GetBytes("POST / HTTP/1.0\r\n\r\nHello World"));
             socket.Shutdown(SocketShutdown.Send);
             var buffer = new byte[8192];
@@ -491,24 +474,7 @@ namespace Microsoft.AspNetCore.Server.KestrelTests
         {
             using (var server = new TestServer(App, testContext))
             {
-                var socket = new Socket(SocketType.Stream, ProtocolType.Tcp);
-                if (PlatformApis.IsWindows)
-                {
-                    const int SIO_LOOPBACK_FAST_PATH = (-1744830448);
-                    var optionInValue = BitConverter.GetBytes(1);
-                    try
-                    {
-                        socket.IOControl(SIO_LOOPBACK_FAST_PATH, optionInValue, null);
-                    }
-                    catch
-                    {
-                        // If the operating system version on this machine did
-                        // not support SIO_LOOPBACK_FAST_PATH (i.e. version
-                        // prior to Windows 8 / Windows Server 2012), handle the exception
-                    }
-                }
-                socket.NoDelay = true;
-                socket.Connect(IPAddress.Loopback, server.Port);
+                var socket = TestConnection.CreateConnectedLoopbackSocket(server.Port);
                 await Task.Delay(200);
                 socket.Dispose();
 
--- a/test/Microsoft.AspNetCore.Server.KestrelTests/EngineTests.cs
+++ b/test/Microsoft.AspNetCore.Server.KestrelTests/EngineTests.cs
@@ -14,6 +14,7 @@ using Microsoft.AspNetCore.Http.Features;
 using Microsoft.AspNetCore.Server.Kestrel;
 using Microsoft.AspNetCore.Server.Kestrel.Filter;
 using Microsoft.AspNetCore.Server.Kestrel.Infrastructure;
+using Microsoft.AspNetCore.Server.Kestrel.Networking;
 using Microsoft.AspNetCore.Testing.xunit;
 using Microsoft.Extensions.Logging;
 using Xunit;
@@ -118,6 +119,22 @@ namespace Microsoft.AspNetCore.Server.KestrelTests
 
             Console.WriteLine("Started");
             var socket = new Socket(AddressFamily.InterNetwork, SocketType.Stream, ProtocolType.Tcp);
+            if (PlatformApis.IsWindows)
+            {
+                const int SIO_LOOPBACK_FAST_PATH = (-1744830448);
+                var optionInValue = BitConverter.GetBytes(1);
+                try
+                {
+                    socket.IOControl(SIO_LOOPBACK_FAST_PATH, optionInValue, null);
+                }
+                catch
+                {
+                    // If the operating system version on this machine did
+                    // not support SIO_LOOPBACK_FAST_PATH (i.e. version
+                    // prior to Windows 8 / Windows Server 2012), handle the exception
+                }
+            }
+            socket.NoDelay = true;
             socket.Connect(new IPEndPoint(IPAddress.Loopback, port));
             socket.Send(Encoding.ASCII.GetBytes("POST / HTTP/1.0\r\n\r\nHello World"));
             socket.Shutdown(SocketShutdown.Send);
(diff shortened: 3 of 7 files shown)


benaadams:
AppVeyor back down < 10 mins ;  but is it sensible?

/cc @halter73

src/Microsoft.AspNet.Server.Kestrel/Infrastructure/KestrelThread.cs
```diff
@@ -120,6 +124,18 @@ private void OnStopRude(object obj)
                 IntPtr.Zero);
         }
 
+        private void OnStopRuder(object obj)
```
davidfowl: Why?

src/Microsoft.AspNet.Server.Kestrel/KestrelEngine.cs
```diff
@@ -43,7 +43,7 @@ public void Dispose()
         {
             foreach (var thread in Threads)
             {
-                thread.Stop(TimeSpan.FromSeconds(2.5));
+                thread.Stop(TimeSpan.FromSeconds(1.0));
```
davidfowl: Why?

src/Microsoft.AspNet.Server.Kestrel/Infrastructure/KestrelThread.cs
benaadams: Libuv reference counts and uses that to determine if it can shutdown loop. `OnStop` unreferences thread' `OnStopRude` calls `Dispose` on all the items still referenced and they call close which hopefully unreferences them. So `OnStopRuder` explicitly unreferences them.

src/Microsoft.AspNet.Server.Kestrel/KestrelEngine.cs
benaadams: In use the number gets multiplied by 4 as it moves through the various `_thread.Join` so `2.5` works out at `10` secs. Will divide internally to maintain meaning?

benaadams:
Resolves #420

src/Microsoft.AspNet.Server.Kestrel/Infrastructure/KestrelThread.cs
benaadams: Actually will split this to different PR

src/Microsoft.AspNet.Server.Kestrel/Networking/Libuv.cs
```diff
@@ -216,6 +227,40 @@ public void tcp_bind(UvTcpHandle handle, ref SockAddr addr, int flags)
...
+            const int WSAEOPNOTSUPP = 10000 + 45; // (WSABASEERR+45)
+            const int SOCKET_ERROR = -1;
+
+            IntPtr socket = IntPtr.Zero;
+            Check(_uv_fileno(handle, ref socket));
+
+            // Enable loopback fast-path for lower latency for localhost comms, like HttpPlatformHandler fronting
+            // http://blogs.technet.com/b/wincat/archive/2012/12/05/fast-tcp-loopback-performance-and-low-latency-with-windows-server-2012-tcp-loopback-fast-path.aspx
+            // https://github.com/libuv/libuv/issues/489
+            int OptionValue = 1;
```
davidfowl: nit: casing

src/Microsoft.AspNet.Server.Kestrel/Networking/Libuv.cs
```diff
@@ -216,6 +227,40 @@ public void tcp_bind(UvTcpHandle handle, ref SockAddr addr, int flags)
         {
             handle.Validate();
             Check(_uv_tcp_bind(handle, ref addr, flags));
+            if (PlatformApis.IsWindows)
+            {
+                tcp_bind_windows_extras(handle);
```
halter73: Should we only call this when binding to a loopback address.

benaadams: No is only off by default for "just in-case" compat; and doesn't seem to have side effect; both sockets need it activated for it to trigger (http://blogs.technet.com/b/wincat/archive/2012/12/05/fast-tcp-loopback-performance-and-low-latency-with-windows-server-2012-tcp-loopback-fast-path.aspx)

> To enable the fast path: 

> 1. Both ends of the TCP session must enable the capability by setting the SIO_LOOPBACK_FAST_PATH IOCTL.
2. The process initiating the TCP connection must set the SIO_LOOPBACK_FAST_PATH IOCTL **prior** to establishing the TCP session.
3. The target of the connection request must set the SIO_LOOPBACK_FAST_PATH IOCTL on the *listen socket*, that is, **prior** to accepting the connection.

And

> The default behavior of the TCP loopback interface is unchanged in Windows Server 2012 thereby preserving compatibility.   Adding support for the loopback fast path enhancement to an existing or new application requires adding the code to set the SIO_LOOPBACK_FAST_PATH  IOCTL in the manner described above.

Unless didn't do it on an specific ip bind; rather than all ip addresses, and localhost bind.

muratg: I think so too, I don't think fast TCP lookback is available otherwise.

benaadams:
Rebased

src/Microsoft.AspNet.Server.Kestrel/Networking/Libuv.cs
```diff
@@ -216,6 +227,40 @@ public void tcp_bind(UvTcpHandle handle, ref SockAddr addr, int flags)
...
+            }
+        }
+
+        private unsafe void tcp_bind_windows_extras(UvTcpHandle handle)
+        {
+            const int SIO_LOOPBACK_FAST_PATH = -1744830448; // IOC_IN | IOC_WS2 | 16;
+            const int WSAEOPNOTSUPP = 10000 + 45; // (WSABASEERR+45)
+            const int SOCKET_ERROR = -1;
+
+            IntPtr socket = IntPtr.Zero;
```
muratg: nit: var

src/Microsoft.AspNet.Server.Kestrel/Networking/Libuv.cs
```diff
@@ -216,6 +227,40 @@ public void tcp_bind(UvTcpHandle handle, ref SockAddr addr, int flags)
...
+            Check(_uv_fileno(handle, ref socket));
+
+            // Enable loopback fast-path for lower latency for localhost comms, like HttpPlatformHandler fronting
+            // http://blogs.technet.com/b/wincat/archive/2012/12/05/fast-tcp-loopback-performance-and-low-latency-with-windows-server-2012-tcp-loopback-fast-path.aspx
+            // https://github.com/libuv/libuv/issues/489
+            int optionValue = 1;
```
muratg: Ditto

test/Microsoft.AspNet.Server.KestrelTests/EngineTests.cs
```diff
@@ -387,6 +403,22 @@ public void ConnectionCanReadAndWrite(TestServiceContext testContext)
...
+                    byte[] optionInValue = BitConverter.GetBytes(1);
+                    try
+                    {
+                        socket.IOControl(SIO_LOOPBACK_FAST_PATH, optionInValue, null);
+                    }
+                    catch (Exception e)
+                    {
+                        // If the operating system version on this machine did
+                        // not support SIO_LOOPBACK_FAST_PATH (i.e. version
+                        // prior to Windows 8 / Windows Server 2012), handle the exception
```
muratg: Do we know what the exception is and catch that particular exception?

test/Microsoft.AspNet.Server.KestrelTests/EngineTests.cs
```diff
@@ -116,6 +116,22 @@ public void ConnectionCanReadAndWrite(TestServiceContext testContext)
...
             var socket = new Socket(AddressFamily.InterNetwork, SocketType.Stream, ProtocolType.Tcp);
+            if (PlatformApis.IsWindows)
+            {
+                const int SIO_LOOPBACK_FAST_PATH = (-1744830448);
+                byte[] optionInValue = BitConverter.GetBytes(1);
+                try
+                {
+                    socket.IOControl(SIO_LOOPBACK_FAST_PATH, optionInValue, null);
+                }
+                catch (Exception e)
```
muratg: Do we know what the exception is and catch that particular exception?

(And below)

benaadams: In the code provided on TechNet in the Server & Tools blog which seems to be the main detail around the announcement of this feature they are catching `SystemException`; however this exception type does seem to exist in coreclr. Hazarding a guess it would be a `SocketException` with `ErrorCode` 10045

benaadams: However; `SocketException` does not support the `ErrorCode` property in coreclr. (Also is just in tests?)

muratg: OK, if CoreCLR doesn't support it properly yet it should be fine to catch `Exception` since these are just tests.  I was just being pedantic :smile:

src/Microsoft.AspNet.Server.Kestrel/Networking/Libuv.cs
halter73: I guess my question is what will happen if someone tries to bind to multiple network interfaces (using different SockAddrs) on the same port. Will we get an error from WSAIoctl on the second call to tcp_bind?

muratg:
Looks good to me, but I'll wait for @davidfowl or @halter73 to give the squirrel.

src/Microsoft.AspNet.Server.Kestrel/Networking/Libuv.cs
davidfowl: Test on down level windows platforms?

benaadams:
More deets https://msdn.microsoft.com/en-us/library/windows/desktop/jj841212(v=vs.85).aspx

> The socket that plans to initiate the connection request must apply this IOCTL before making the connection request. So the socket used by the connect, ConnectEx, WSAConnect, WSAConnectByList, or WSAConnectByName function to initiate the connection must apply this IOCTL to use the fast path for loopback operations.

> The socket that is listening for the connection request must apply this IOCTL before accepting the connection. So the socket used by the listen function must apply this IOCTL so any sockets that are accepted will use the fast path for loopback. Any sockets returned by the listen function and passed to the accept, AcceptEx, or WSAAccept function will be marked to use the special fast path for loopback operations.

> Once an application establishes the connection on a loopback interface using the fast path, all packets for the lifetime of the connection must use the fast path.

> Applying SIO_LOOPBACK_FAST_PATH to a socket which will be connected to a non-loopback path will have no effect.

^--- suggests no side effects; when not loopback

> This TCP loopback optimization results in packets that flow through the Transport Layer (TL) instead of the traditional loopback through the Network Layer. This optimization improves the latency for loopback packets. Once an application opts in for a connection level setting to use the loopback fast path, all packets will follow the loopback path. For loopback communications, congestion and packet drop are not expected. The notion of congestion control and reliable delivery in TCP will be unnecessary. This, however, is not true for flow control. Without flow control, the sender can overwhelm the receive buffer, leading to erroneous TCP loopback behavior. The flow control in the TCP optimized loopback path is maintained by placing send requests in a queue. When receive buffer is full, the TCP/IP stack guarantees that the sends won't complete until the queue is serviced maintaining flow control.

src/Microsoft.AspNet.Server.Kestrel/Networking/Libuv.cs
benaadams: > What will happen if someone tries to bind to multiple network interfaces (using different SockAddrs) on the same port. Will they get an error result from WSAIoctl on the second call to tcp_bind?

No issue its per socket; also one of them would likely be not loopback, when it becomes a noop

benaadams: @davidfowl I spent a while trying to get stuff to run on 2008 R2 SP1 but dnvm; wasn't completely happy and it had troubles with nuget; and I couldn't run build - so not sure as to the effects to the patch.

src/Microsoft.AspNet.Server.Kestrel/Networking/Libuv.cs
```diff
@@ -582,6 +630,22 @@ private static class NativeMethods
 
             [DllImport("libuv", CallingConvention = CallingConvention.Cdecl)]
             unsafe public static extern int uv_walk(UvLoopHandle loop, uv_walk_cb walk_cb, IntPtr arg);
+
+            [DllImport("WS2_32.dll", CallingConvention = CallingConvention.Winapi)]
```
moozzyk: Does this work on Nano?

cesarbs:
Tested this on Windows 7, 8 and 2008 R2, both local connection and connection from another machine. Everything looks fine except I can't do a remote connection to 2008 R2. Not caused by the change though, I can't do it even without it. Might be a bug we have to look into, but I'm actually more suspicious it's something that needs to be setup correctly.

cesarbs:
Next I'll make sure this does't cause issues on Linux and Mac.

benaadams:
Opened windows advanced firewall ports on 2008?

cesarbs:
@benaadams Thanks, that worked. For 7 and 8 I only had to create an endpoint in Azure portal, but 2008 required that extra step.

cesarbs:
@benaadams Can you rebase?

benaadams:
Rebased

cesarbs:
Awesome, thanks.

src/Microsoft.AspNetCore.Server.Kestrel/Networking/Libuv.cs
```diff
@@ -610,6 +674,9 @@ private static class NativeDarwinMonoMethods
             public static extern void uv_unref(UvHandle handle);
 
             [DllImport("__Internal", CallingConvention = CallingConvention.Cdecl)]
+            public static extern int uv_fileno(UvHandle handle, ref IntPtr socket);
```
moozzyk: This is only used on Darwin Mono so may not be needed if the feature is windows specific...

test/Microsoft.AspNetCore.Server.KestrelTests/EngineTests.cs
```diff
@@ -118,6 +119,22 @@ public void ConnectionCanReadAndWrite(TestServiceContext testContext)
 
             Console.WriteLine("Started");
             var socket = new Socket(AddressFamily.InterNetwork, SocketType.Stream, ProtocolType.Tcp);
+            if (PlatformApis.IsWindows)
```
moozzyk: create a helper function?

src/Microsoft.AspNetCore.Server.Kestrel/Networking/Libuv.cs
benaadams: Would end up with mismatched function implements? So if someone did want to use the `fileno` in future calling the function on darwin mono would return incorrect info (if replaced with a dummy function)

moozzyk: If they need it they would have to add them. I don't really care too much - the `__Internal` is going away soon. It won't/can't work with dotnet and even [now](https://github.com/aspnet/KestrelHttpServer/commit/fe75aa47dbe4209600511543fb5b2f9aca055c8a) when using dotnet the regular DllImports are being used on Mac on Mono.

benaadams: I mean, I don't know what the `uv_fileno` function should resolve to on Darwin Mono if taken out. 

With this in it resolves to the libuv function - if I take it out should it resolve to a function shim with body: `throw new NotImplementedExecption("Libuv function not mapped on on Darwin Mono")`?

moozzyk: I think you are right. It's better and easier to have `uv_fileno` in __Internal just as any other function than do a special dance just to avoid it.

cesarbs:
Doesn't cause any issues on Linux and Mac either. :shipit:

cesarbs:
Nope, no :shipit: :disappointed: 

@benaadams Did you notice any instability in test runs after your change? I've fetched your PR locally and I'm seeing wild test failures (each time a different one). Not seeing the same on dev.

benaadams:
@cesarbs don't get any test errors (Win10) - but I do get hang after first set of tests with all versions e.g.
```
Published to test\Microsoft.AspNetCore.Server.KestrelTests\.testPublish\dnx451
Published 1/1 projects successfully
info: Copying HTML.xslt
info: Copying NUnitXml.xslt
info: Copying xunit.abstractions.dll
info: Copying xunit.console.exe
info: Copying xunit.console.exe.config
info: Copying xunit.console.x86.exe
info: Copying xunit.console.x86.exe.config
info: Copying xunit.runner.reporters.desktop.dll
info: Copying xunit.runner.utility.desktop.dll
info: Copying xUnit1.xslt
info: Copying xunit.execution.desktop.dll
info: Exec
info:   program: I:\Work\GitHub\KestrelHttpServer\test\Microsoft.AspNetCore.Server.KestrelTests\.testPublish\dnx451\xunit.console.exe
info:   commandline: Microsoft.AspNetCore.Server.KestrelTests.dll
info:   workingdir: test\Microsoft.AspNetCore.Server.KestrelTests\.testPublish\dnx451
xUnit.net Console Runner (64-bit .NET 4.0.30319.42000)
  Discovering: Microsoft.AspNetCore.Server.KestrelTests
  Discovered:  Microsoft.AspNetCore.Server.KestrelTests
  Starting:    Microsoft.AspNetCore.Server.KestrelTests
Vector.IsHardwareAccelerated == False
Vector<byte>.Count == 16
Task.Run
Connected
Task.Run
Connected
count 10 ABCDEABCDE
count 0
  Finished:    Microsoft.AspNetCore.Server.KestrelTests
=== TEST EXECUTION SUMMARY ===
   Microsoft.AspNetCore.Server.KestrelTests  Total: 268, Errors: 0, Failed: 0, Skipped: 0, Time: 4.512s
info: Exec
info:   program: dotnet
info:   commandline: test
info:   workingdir: test\Microsoft.AspNetCore.Server.Kestrel.FunctionalTests
xUnit.net DNX Runner (64-bit win7-x64)
  Finished:    Microsoft.AspNetCore.Server.Kestrel.FunctionalTests
=== TEST EXECUTION SUMMARY ===
   Microsoft.AspNetCore.Server.Kestrel.FunctionalTests  Total: 48, Errors: 0, Failed: 0, Skipped: 0, Time: 42.782s
```
Hang...

cesarbs:
@benaadams Yep, hanging here too. For me it's hanging before the test summary is displayed.

cesarbs:
@benaadams Have you see this in any of your runs?

    Microsoft.AspNetCore.Server.KestrelTests.NetworkingTests.SocketCanRead [FAIL]
      System.Net.Sockets.SocketException : A connect request was made on an already connected socket
      Stack Trace:
           at System.Net.Sockets.Socket.BeginConnectEx(EndPoint remoteEP, Boolean flowContext, AsyncCallback callback, Object state)
           at System.Net.Sockets.Socket.BeginConnect(EndPoint remoteEP, AsyncCallback callback, Object state)
           at System.Threading.Tasks.TaskFactory`1.FromAsyncImpl[TArg1](Func`4 beginMethod, Func`2 endFunction, Action`1 endAction, TArg1 arg1, Object state, TaskCreationOptions creationOptions)
           at System.Net.Sockets.SocketTaskExtensions.ConnectAsync(Socket socket, EndPoint remoteEndPoint)
        D:\src\aspnet\KestrelHttpServer\test\Microsoft.AspNetCore.Server.KestrelTests\NetworkingTests.cs(143,0): at Microsoft.AspNetCore.Server.KestrelTests.NetworkingTests.<>c__DisplayClass7_0.<<SocketCanRead>b__3>d.MoveNext()
        D:\src\aspnet\KestrelHttpServer\test\Microsoft.AspNetCore.Server.KestrelTests\NetworkingTests.cs(151,0): at Microsoft.AspNetCore.Server.KestrelTests.NetworkingTests.<SocketCanRead>d__7.MoveNext()

benaadams:
Wow, never - coreclr bug? Are you on later feed? What runtime, OS?

cesarbs:
Win10, volatile feed. This is on CoreCLR.

benaadams:
Was using ci; trying volatile

cesarbs:
How are you using ci without hitting this:

`Unable to resolve Microsoft.DotNet.CoreHost (>= 0.0.1-beta-00001) for DNXCore,Version=v5.0.`

?

benaadams:
Used the dotnet feed for something else; think it put it in my local cache

benaadams:
Freezing at first test summary on volatile rather than second :(

cesarbs:
Removing calls to `ConnectAsync` from lines 143 and 221 of `NetworkingTests.cs` got me past errors. Still hanging after test summary though.

test/Microsoft.AspNetCore.Server.KestrelTests/TestConnection.cs
```diff
@@ -125,5 +125,28 @@ public void Dispose()
             var text = new string(ch, 0, count);
             Assert.Equal("", text);
         }
+
+        public static Socket CreateConnectedLoopbackSocket(int port, AddressFamily addressFamily = AddressFamily.InterNetwork)
+        {
+            var socket = new Socket(addressFamily, SocketType.Stream, ProtocolType.Tcp);
+            if (PlatformApis.IsWindows)
+            {
+                const int SIO_LOOPBACK_FAST_PATH = (-1744830448);
```
cesarbs: Why the parens?

test/Microsoft.AspNetCore.Server.KestrelTests/TestConnection.cs
```diff
@@ -125,5 +125,28 @@ public void Dispose()
...
+                    socket.IOControl(SIO_LOOPBACK_FAST_PATH, optionInValue, null);
+                }
+                catch
+                {
+                    // If the operating system version on this machine did
+                    // not support SIO_LOOPBACK_FAST_PATH (i.e. version
+                    // prior to Windows 8 / Windows Server 2012), handle the exception
+                }
+            }
+            socket.NoDelay = true;
```
cesarbs: Why `NoDelay`?

test/Microsoft.AspNetCore.Server.KestrelTests/TestConnection.cs
```diff
@@ -125,5 +125,28 @@ public void Dispose()
             var text = new string(ch, 0, count);
             Assert.Equal("", text);
         }
+
+        public static Socket CreateConnectedLoopbackSocket(int port, AddressFamily addressFamily = AddressFamily.InterNetwork)
```
cesarbs: I don't see any calls overriding the optional `addressFamily` parameter. Move to method body?

test/Microsoft.AspNetCore.Server.KestrelTests/TestConnection.cs
cesarbs: After removed it I got my first non-hanging run. Will run a few more times.

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/TestConnection.cs
```diff
@@ -0,0 +1,36 @@
...
+                    socket.IOControl(SIO_LOOPBACK_FAST_PATH, optionInValue, null);
+                }
+                catch
+                {
+                    // If the operating system version on this machine did
+                    // not support SIO_LOOPBACK_FAST_PATH (i.e. version
+                    // prior to Windows 8 / Windows Server 2012), handle the exception
+                }
+            }
+            socket.NoDelay = true;
```
cesarbs: Removed this too and got rid of the last hang. You might want to follow up with the .NET team on this.

cesarbs:
:ship: after removing offending lines.

benaadams:
> Removing calls to ConnectAsync from lines 143 and 221 of NetworkingTests.cs got me past errors. Still hanging after test summary though.

D'oh... double connect

test/Microsoft.AspNetCore.Server.KestrelTests/TestConnection.cs
benaadams: Copy paste code from the blog post on example.

benaadams:
Needs a squish

benaadams:
Removed lines, made the changes.

cesarbs:
@benaadams Still seeing hangs?

benaadams:
> Still seeing hangs?

Yep; but see hangs with everything. Tried https://github.com/aspnet/KestrelHttpServer/pull/633 to resolve them, didn't help.

Not sure if its kestrel shutdown, tests, xunit or dotnet stuff - but started happening for me on move to dotnet from dnx.

Know which is the xunit repo that the tests reference?

cesarbs:
This is the runner that we're using: https://github.com/xunit/xunit/tree/master/src/xunit.console

benaadams:
Always seem get stuck here

benaadams:
Other thread is stuck at which looks to be `_process.WaitForExit();`

benaadams:
K and the process is stuff because the libuv is still in `Run`

kestrel

benaadams:
Usual place https://github.com/aspnet/KestrelHttpServer/blob/dev/src/Microsoft.AspNetCore.Server.Kestrel/Infrastructure/KestrelThread.cs#L280

benaadams:
@cesarbs  Added `"Talk to the hand" stop` to https://github.com/aspnet/KestrelHttpServer/pull/633 which sorts out the hangs.

cesarbs:
Once #623 is merged the hands should be gone. As far as this PR goes, everything looks good now.

PR #417, opened 2015-11-22
Moved ThreadCount initialization up the pipeline

### LINKED ISSUE #404: The default thread count should be reflected on KestrelServerInformation.ThreadCount during application start
The default thread count is currently determined by `KestrelEngine` **after** the application's `Conifgure` method has been called. This means the application can't inspect the default value for `IKestrelServerInformation.ThreadCount`, like it can for `IKestrelServerInformation.NoDelay`. Essentially, the following code should echo the correct defaults:

``` C#
public void Configure(IApplicationBuilder app)
{
    var kestrel = app.ServerFeatures[typeof(IKestrelServerInformation)] as IKestrelServerInformation;

    if (kestrel != null)
    {
        Console.WriteLine($"Kestrel thread count: {kestrel.ThreadCount}");
        Console.WriteLine($"Kestrel TCP no-delay: {kestrel.NoDelay}");
    }
```

khellang:
Closes #404. Replaces #397.

### DIFF
--- a/src/Microsoft.AspNet.Server.Kestrel/KestrelServer.cs
+++ b/src/Microsoft.AspNet.Server.Kestrel/KestrelServer.cs
@@ -73,34 +73,13 @@ namespace Microsoft.AspNet.Server.Kestrel
                 _disposables.Push(engine);
                 _disposables.Push(dateHeaderValueManager);
 
-                // Actual core count would be a better number 
-                // rather than logical cores which includes hyper-threaded cores.
-                // Divide by 2 for hyper-threading, and good defaults (still need threads to do webserving).
-                // Can be user overriden using IKestrelServerInformation.ThreadCount
-                var threadCount = Environment.ProcessorCount >> 1;
+                var threadCount = information.ThreadCount;
 
-                if (threadCount < 1)
+                if (threadCount <= 0)
                 {
-                    // Ensure shifted value is at least one
-                    threadCount = 1;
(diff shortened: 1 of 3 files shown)


muratg:
@halter73 Could you have a look?

src/Microsoft.AspNet.Server.Kestrel/KestrelServer.cs
```diff
@@ -75,35 +75,14 @@ public void Start(RequestDelegate requestDelegate)
...
 
-                // Actual core count would be a better number 
-                // rather than logical cores which includes hyper-threaded cores.
-                // Divide by 2 for hyper-threading, and good defaults (still need threads to do webserving).
-                // Can be user overriden using IKestrelServerInformation.ThreadCount
-                var threadCount = Environment.ProcessorCount >> 1;
+                var threadCount = information.ThreadCount;
 
-                if (threadCount < 1)
+                if (threadCount < 0)
```
halter73: This PR looks good, sorry for the delayed response. The only thing I would change is making this `threadCount <= 0` and the exception message ""ThreadCount must be positive."

khellang: Fixed! :+1:

khellang:
Squashed and ready to go :wink:

halter73:
Thanks!

PR #421, opened 2015-11-23
Moar stopping power, Timeout is total timout

benaadams:
AppVeyor has got up to 23 mins to run; this brings it back down to 9 mins

### DIFF
--- a/src/Microsoft.AspNet.Server.Kestrel/Infrastructure/KestrelThread.cs
+++ b/src/Microsoft.AspNet.Server.Kestrel/Infrastructure/KestrelThread.cs
@@ -64,16 +64,18 @@ namespace Microsoft.AspNet.Server.Kestrel
                 return;
             }
 
+            var stepTimeout = (int)(timeout.TotalMilliseconds / 3); 
+
             Post(t => t.OnStop());
-            if (!_thread.Join((int)timeout.TotalMilliseconds))
+            if (!_thread.Join(stepTimeout))
             {
                 try
                 {
                     Post(t => t.OnStopRude());
-                    if (!_thread.Join((int)timeout.TotalMilliseconds))
+                    if (!_thread.Join(stepTimeout))
                     {
                         Post(t => t.OnStopImmediate());
-                        if (!_thread.Join((int)timeout.TotalMilliseconds))
+                        if (!_thread.Join(stepTimeout))
                         {
 #if NET451
                             _thread.Abort();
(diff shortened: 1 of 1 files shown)


src/Microsoft.AspNet.Server.Kestrel/Infrastructure/KestrelThread.cs
```diff
@@ -120,6 +126,18 @@ private void OnStopRude(object obj)
...
         }
 
+        private void OnStopForceUnreference(object obj)
+        {
+            _engine.Libuv.walk(
+                _loop,
+                (ptr, arg) =>
+                {
+                    var handle = UvMemory.FromIntPtr<UvHandle>(ptr);
+                    _loop.Libuv.unref(handle);
```
halter73: Do you know what situations would prevent uv_close from unrefing a handle?

benaadams: Not sure; though I found it was doing it when debugging. Best guess is its multi-dispose protection from by the SafeHandle, and they get added back somehow, but will only dispose once?

benaadams: For example `_post.Unreference()` followed by `if (handle != _post)` in the harsher stop, suggests a similar encounter of this previously.

benaadams:
Close to what was happening in `DangerousClose`; but it just decrements libuv's reference count rather than calling `Dispose+ReleaseHandle`; also `Dispose` should have been called in the previous stop.

halter73:
Are you sure Stop is being sped up as a result of the `OnStopForceUnreference` change? That is the most questionable part of the PR to me.

It seems odd that there would even be anything left to unref if we already called `uv_close` on all the referenced handles. I wonder if we are only seeing quicker AppVeyor runs because this effectively decreased the Stop timeout by a factor of 3.

I'm going to try to get @lodejard to look at this.

halter73:
OK. I took a look at the libuv source code myself to see what's going on. At least on Windows, I can now see why `uv_close` will not unref the handle.

If you look at [uv_tcp_endgame](https://github.com/libuv/libuv/blob/v1.7.5/src/win/tcp.c#L211) which is triggered by `uv_close` on TCP handles, you'll see that the `uv__handle_close` macro (the thing that actually does the unreffing) is only called if `reqs_pending == 0`.

Basically, `uv_close` isn't as forceful as I thought.

benaadams:
I think `uv_tcp_endgame` might be over promising in name :)

halter73:
@lodejard suggested that simply calling `uv_stop` is the more straightforward way of gracelessly killing the loop over unreffing all the handles. I'm inclined to agree.

halter73:
Since we're already calling `uv_stop`` inside of `OnStopImmediate` after `OnStopForceUnreference`, I think we can remove `OnStopForceUnreference`. I don't mind shortening the timeout though.

benaadams:
Changed; start your watch :)

halter73:
"10 min 39 sec" on AppVeyor. Good enough for me!

PR #422, opened 2015-11-23
Don't throw exception out of threadpool

benaadams:
Would like to log with logger in the catches but would need to be with a static var to not allocate. Suggestions?

/cc @davidfowl

### DIFF
--- a/src/Microsoft.AspNet.Server.Kestrel/Infrastructure/TaskUtilities.cs
+++ b/src/Microsoft.AspNet.Server.Kestrel/Infrastructure/TaskUtilities.cs
@@ -1,7 +1,10 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System;
+using System.Threading;
 using System.Threading.Tasks;
+using Microsoft.AspNet.Server.Kestrel.Http;
 
 namespace Microsoft.AspNet.Server.Kestrel.Infrastructure
 {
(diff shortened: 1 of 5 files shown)


src/Microsoft.AspNet.Server.Kestrel/Infrastructure/TaskUtilities.cs
```diff
@@ -12,5 +15,72 @@ public static class TaskUtilities
...
+            {
+                // log with a static logger in some way?
+            }
+        };
+
+        private static WaitCallback _completeTcs = (o) =>
+        {
+            try
+            {
+                ((TaskCompletionSource<object>)o).SetResult(null);
```
halter73: Was there a situation where this was throwing before? If there was, we could just change this to call TrySetResult which never throws. We can also call TrySetException in ErrorOnThreadPool.

benaadams: No, and I doubt it ever would as this is the only place that ever sets it and it is called only once - however not catching an exception at the top level of a threadpool thread; which is at the top of the stack, will crash the process (by design) as its kinda like `async void`

benaadams: Actually might be better to add a callback to ThreadPool for unhandled execptions

benaadams:
Added https://github.com/dotnet/coreclr/issues/2144 which would probably be easier and better?

src/Microsoft.AspNet.Server.Kestrel/Infrastructure/TaskUtilities.cs
halter73: Given that we aren't likely to see the callback added very soon, I think we might want something like this.

The main thing preventing me from merging this is the fact that it doesn't log the exceptions.

halter73:
I think the correct thing to do is make sure that any action we call QueueUserWorkItem with *cannot* throw.

If thre are any places we call QueueUserWorkItem with an action that could possibly throw, we should open another PR that at the very least captures the exception and logs it.

src/Microsoft.AspNet.Server.Kestrel/Infrastructure/TaskUtilities.cs
benaadams: Agreed; though wasn't sure how to get am instance of logger in here? Would setting a static from startup work?

benaadams:
Not sure they are easy places; SocketInput and SocketOutput completions - will do some testing see if its a problem.

halter73:
We can reopen if you strongly disagree. I do see the benefit of having all the try/catch blocks in one place in order to avoid missing stuff, but not at the cost of swallowing exceptions.

Another alternative might be to make the logger a parameter to the `TaskUtilities` methods.

benaadams:
@halter73 updated in  #443 currying to embed the log in closures so its scoped rather than using static

PR #424, opened 2015-11-24
Improved RequestProcessingAsync legibility

benaadams:
With precomputed tasks

Hopefully #391, the good bits...


benaadams:
myget down :(

benaadams:
@mattwarren indeed; and the magic continues :+1:

benaadams:
Not good will revisit.

PR #427, opened 2015-11-25
Process cascaded work immediately

benaadams:
Without waiting for next libuv pass; due to it being added to the `_workAdding` buffer and having to wait to the next pass to move to the `_workRunning` buffer to be submitted; which will currently throttle writes to network and handle close speed.

This is to get the events into the libuv loop as early as possible as it determines its sleep time between loops based on pending work. Otherwise end up with a throttled message pump.

Fix for potential regression in #363 due to bug fix.

### DIFF
--- a/src/Microsoft.AspNet.Server.Kestrel/Infrastructure/KestrelThread.cs
+++ b/src/Microsoft.AspNet.Server.Kestrel/Infrastructure/KestrelThread.cs
@@ -18,6 +18,11 @@ namespace Microsoft.AspNet.Server.Kestrel
     /// </summary>
     public class KestrelThread
     {
+        // maximum times the work queues swapped and are processed in a single pass
+        // as completing a task may immediately have write data to put on the network
+        // otherwise it needs to wait till the next pass of the libuv loop
+        private const int _maxLoops = 8;
+
         private static Action<object, object> _threadCallbackAdapter = (callback, state) => ((Action<KestrelThread>)callback).Invoke((KestrelThread)state);
         private KestrelEngine _engine;
         private readonly IApplicationLifetime _appLifetime;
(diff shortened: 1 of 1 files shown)


src/Microsoft.AspNet.Server.Kestrel/Infrastructure/KestrelThread.cs
```diff
@@ -262,11 +267,22 @@ private void ThreadStart(object parameter)
...
         {
-            DoPostWork();
-            DoPostCloseHandle();
+            var loopsRemaining = _maxLoops;
+            bool wasWork;
+            do
+            {
+                // Queues swapped
+                wasWork = DoPostWork();
+                wasWork = DoPostCloseHandle() || wasWork;
```
halter73: Why is it important to call the DoPost methods twice in one iteration? I think it would be cleaner to just double `_maxLoops`.

halter73:
I think we need to measure this once #363 is in.

src/Microsoft.AspNet.Server.Kestrel/Infrastructure/KestrelThread.cs
benaadams: I started with that; then changed to this, didn't add reason :(

I think it was to make sure a minimum of a double loop had always been done and both queues processed in case a task had snuck in; to get the tasks into the Libuv queue as early as possible for lower latency. 

libuv-loop

But if it was `DoPostWork` that added an item to the queue then `wasWork` should return `true` so that would be ok as it would do another loop. If `wasWork` returned false, then `DoPostWork` and `DoPostCloseHandle` would both have returned pretty fast so probably wouldn't be anything extra there?

My worry was the libuv loop sleeping (15ms?) because it had nothing to do; however an item sitting in the thread queue ready to go; but missed due to `uv_async_send` [coalescing](http://docs.libuv.org/en/latest/async.html#c.uv_async_send) and blocked on the add behind the lock; but I was probably being over cautious?

halter73: I don't think coalescing is an issue since we always call `uv_async_send` after enqueing to `_workAdding`. We've verified that calling `uv_async_send` will not coalesce if it is the first call after the post callback *starts*.

benaadams: Yeah; will think a while and see if the reason comes to me; else revert.

benaadams: Actually have taken a better approach - have reverted; then will submit another PR if I can come up with a good reason backed by metrics :)

benaadams:
Merged with #363 as is response to that.

PR #428, opened 2015-11-25
Loose local vars, for lighter state machine

benaadams:
Trims 8kb over 4k requests (20 bytes per request); not huge in the scheme of things; but every little helps...


benaadams:
Gah

Issue #430, opened 2015-11-27
Proposal: Improve the precision and speed of DateHeaderValueManager


tenor:
The `System.Timer` that fetches the time from the RTC in DateHeaderValueManager has an interval of one second. That means the precision can not be higher than one second.
Also, calls to `GetDateHeaderValue` call `PumpTimer()` which involves a `lock`, effectively serializing all calls to this method.

The following proposal improves the precision and speed of DateHeaderValueManager:

Instead of `PumpTimer()`, have `StartTimer()` and `StopTimer()` methods.
`StartTimer` will:

1. Calculate a `_lastReadTime` long value. (more on this later)
2. Calculate a `_lastReadStringTime` string or byte array value.
3. If the object isn't disposed,  set a volatile\interlocked boolean value called `_timerIsRunning` to true and start the timer loop.

The timer loop, will read the UTC time from the RTC every 10 ms, convert it to ticks and store it in a long called `_currentTime`.

`GetDateValue()` will look like this:

```csharp
 byte[]/string GetDateValue()
{
  if(!_timerIsRunning){StartTimer();}
     return ReadDateValue();
}

private ReadDateValue()
{
long currentTime;
if(_disposed)
{
 currentTime = GetFromRTC();
}
else
{
  currentTime = ReadCurrentTime(); //This performs a dirty read on _currentTime, or Interlocked.Read on 32-bit systems.
}

if (_lastReadTime - currentTime > threshold \* 100 ms*\)
{
  if (interlocked.CompareExchange(ref _lastReadTime, ref currentTime)
  {
    var _newTimeString = CalculateTimeString(_lastReadTime);
    Interlocked.CompareExchange(ref _lastReadTimeString, ref _newTimeString);
  }
}

 return _lastReadTimeString;

}
```

This proposal eliminates the `lock` in `PumpTimer()` and increases the time precision to 100ms.
I'm not sure this will move the overall performance needle in Kestrel but I think it's worth checking out.

I can provide a PR if the team finds this approach interesting.

davidfowl:
Make sure to show some performance results with the pull request. That'll be more convincing than anything else.

khellang:
> That means the precision can not be higher than one second.

Just curious, why would you need a precision higher than one second when format used is RFC1123, e.g. `Sat, 28 Nov 2015 20:45:30 GMT`?

tenor:
@khellang  :smile:  I didn't even look at that time format. I'll have to rethink this.
[...]

PR #431, opened 2015-11-29
Speed up DateHeaderValueManager

tenor:
This is the PR for #430 

I removed PumpTimer() and replaced it with StartTimer() and introduced a new StopTimer() method which pauses the timer instead of disposing it.

It's about 2 to 3 times faster than the existing manager and it's more precise because it polls the RTC every 200ms.

Here are some notes I made while working on the change:

1. The older `ISystemClock` used `DateTimeOffset.UtcNow`. That property is about 10 times slower than `DateTime.UtcNow`. I've changed `ISystemClock` to use `DateTime.UtcNow` instead.
Is there a reason why `DateTimeOffset` is preferred? I can change it back without having any effect on the performance. It's only called in the timer loop.
*Update: I've changed it back to DateTimeOffset due to failing unit tests.*

2. The `_activeDateBytes` flag toggles between reading `_dateBytes0` and `_dateBytes1`. I understand this is done to prevent allocations but it looks what's needed here is a buffer pool, so you can say `pool.GetBuffer(size)` and work with the provided buffer.

Here are the results for the speed test. You can download the speed test project from [here](https://www.dropbox.com/s/dxnacgbzwxfgbfd/ImprovedDateHeaderTime.zip?dl=0)

On dnx x64:
```
Testing Old Manager....
7 threads @ 50,000,000 calls per thread.
Total Time: 00:00:32.0746731


Testing New Manager....
7 threads @ 50,000,000 calls per thread.
Total Time: 00:00:11.7850176
Test Complete!
```

On dnx x86:
```
Testing Old Manager....
7 threads @ 50,000,000 calls per thread.
Total Time: 00:00:21.1245817


Testing New Manager....
7 threads @ 50,000,000 calls per thread.
Total Time: 00:00:11.2784120
Test Complete!


```

### DIFF
--- a/src/Microsoft.AspNet.Server.Kestrel/Http/DateHeaderValueManager.cs
+++ b/src/Microsoft.AspNet.Server.Kestrel/Http/DateHeaderValueManager.cs
@@ -22,10 +22,14 @@ namespace Microsoft.AspNet.Server.Kestrel.Http
         private readonly byte[] _dateBytes0 = Encoding.ASCII.GetBytes("\r\nDate: DDD, dd mmm yyyy hh:mm:ss GMT");
         private readonly byte[] _dateBytes1 = Encoding.ASCII.GetBytes("\r\nDate: DDD, dd mmm yyyy hh:mm:ss GMT");
         private object _timerLocker = new object();
-        private bool _isDisposed = false;
-        private bool _hadRequestsSinceLastTimerTick = false;
+        private volatile bool _isDisposed = false;
+        private volatile bool _hadRequestsSinceLastTimerTick = false;
         private Timer _dateValueTimer;
-        private DateTimeOffset _lastRequestSeen = DateTimeOffset.MinValue;
+        private long _lastRequestSeenTicks;
+        private long _lastReadDateTimeTicks;
+        private readonly bool _is64BitSystem;
+        private readonly long _ticksInOneSecond;
+        private volatile bool _timerIsRunning;
 
         /// <summary>
         /// Initializes a new instance of the <see cref="DateHeaderValueManager"/> class.
@@ -34,9 +38,8 @@ namespace Microsoft.AspNet.Server.Kestrel.Http
             : this(
                   systemClock: new SystemClock(),
                   timeWithoutRequestsUntilIdle: TimeSpan.FromSeconds(10),
-                  timerInterval: TimeSpan.FromSeconds(1))
+                  timerInterval: TimeSpan.FromMilliseconds(200))
         {
-
         }
 
         // Internal for testing
@@ -48,6 +51,8 @@ namespace Microsoft.AspNet.Server.Kestrel.Http
             _systemClock = systemClock;
             _timeWithoutRequestsUntilIdle = timeWithoutRequestsUntilIdle;
             _timerInterval = timerInterval;
+            _dateValueTimer = new Timer(TimerLoop, state: null, dueTime: Timeout.Infinite, period: Timeout.Infinite);
+            _is64BitSystem = IntPtr.Size >= 8;
         }
 
         /// <summary>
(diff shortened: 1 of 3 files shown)


src/Microsoft.AspNet.Server.Kestrel/Http/DateHeaderValueManager.cs
```diff
@@ -34,9 +37,8 @@ public DateHeaderValueManager()
             : this(
                   systemClock: new SystemClock(),
                   timeWithoutRequestsUntilIdle: TimeSpan.FromSeconds(10),
-                  timerInterval: TimeSpan.FromSeconds(1))
+                  timerInterval: TimeSpan.FromMilliseconds(200))
```
benaadams: Not sure running 5 times as often is necessary? As it doesn't need to be highly accurate (reasons given in issue)

davidfowl:
@tenor How does this affect the RPS?

src/Microsoft.AspNet.Server.Kestrel/Http/DateHeaderValueManager.cs
tenor: The timerInterval only affects the precision and has no effect on the performance gains.
GetDatexxx method calls will still be 2-3 times faster regardless of what is set as the timer interval.

tenor:
@davidfowl I expect to see some slight RPS gains but nothing significant.
I only removed one impediment in the request path but there are probably other bottlenecks in the request path.

davidfowl:
If there's no real gain in RPS or overall memory reduction then there's no point.

tenor:
Say you have a pipe with multiple clogs. Someone clears one clog but that fix is rejected because the overall flow is not impacted, then that pipe will never get de-clogged.

The cool thing about open source is that over time, a lot of those bottlenecks tend to be found and fixed.

src/Microsoft.AspNet.Server.Kestrel/Http/DateHeaderValueManager.cs
halter73: I would remove this change from this PR then. If you think it's worth it to increase the precision of the date header, just submit another PR with this one line change.

src/Microsoft.AspNet.Server.Kestrel/Http/DateHeaderValueManager.cs
```diff
@@ -57,103 +61,159 @@ public DateHeaderValueManager()
...
+            _activeDateBytes = !_activeDateBytes;
+        }
+
+        private void WriteLongThreadSafe(ref long location, long value)
         {
-            if (_dateValueTimer != null)
+            Interlocked.Exchange(ref location, value);
+        }
+
+        private long ReadLongThreadSafe(ref long location)
```
halter73: Do we know if this is more efficient on 64-bit machines than simply calling Interlocked.Read?

src/Microsoft.AspNet.Server.Kestrel/Http/DateHeaderValueManager.cs
```diff
@@ -57,103 +61,159 @@ public DateHeaderValueManager()
...
+        /// <param name="value">A DateTimeOffset value</param>
+        private void SetDateValues(DateTimeOffset value)
+        {
+            // See http://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.18 for required format of Date header
+            _dateValue = value.ToString(Constants.RFC1123DateFormat);
+            Encoding.ASCII.GetBytes(_dateValue, 0, _dateValue.Length, !_activeDateBytes ? _dateBytes0 : _dateBytes1, "\r\nDate: ".Length);
+            _activeDateBytes = !_activeDateBytes;
+        }
+
+        private void WriteLongThreadSafe(ref long location, long value)
```
halter73: I think Interlocked.Exchange can be called directly.

src/Microsoft.AspNet.Server.Kestrel/Http/DateHeaderValueManager.cs
```diff
@@ -57,103 +61,159 @@ public DateHeaderValueManager()
...
-            PumpTimer();
-
-            // See https://msdn.microsoft.com/en-us/library/az4se3k1(v=vs.110).aspx#RFC1123 for info on the format
-            // string used here.
-            // The null-coalesce here is to protect against returning null after Dispose() is called, at which
-            // point _dateValue will be null forever after.
-            return _dateValue ?? _systemClock.UtcNow.ToString(Constants.RFC1123DateFormat);
+            _hadRequestsSinceLastTimerTick = true;
+            PrepareDateValues();
+            return _dateValue;
```
halter73: So after this is disposed, GetDateHeaderValue() will forever return the value as it was when it was disposed.

This seems fine to me, but given the old comment, I want to verify this is OK. @DamianEdwards?

halter73:
OK. I just saw the RPS comments. @davidfowl Are you planning on closing this?

src/Microsoft.AspNet.Server.Kestrel/Http/DateHeaderValueManager.cs
tenor: @halter73 Sure, I'll change this to one second and submit another PR that sets it to 400ms.

src/Microsoft.AspNet.Server.Kestrel/Http/DateHeaderValueManager.cs
tenor: Yes, it will, however,  `PrepareDateValues()` will return the date from the real time clock if `_isDisposed` is true. The older code behaved this way too.

src/Microsoft.AspNet.Server.Kestrel/Http/DateHeaderValueManager.cs
tenor: Yes, Interlocked.Read is mostly useful for 32-bit systems. It eliminates the possibility of having a torn read, which can't happen when reading a 64 bit value on a 64 bit machine (since .NET aligns the word properly).

src/Microsoft.AspNet.Server.Kestrel/Http/DateHeaderValueManager.cs
tenor: The ReadLong/WriteLongThreadSafe methods are really there to remind anyone modifying this class that they need to use Interlocked methods when getting/setting those fields.
Currently, there'll only be one thread reading/writing those fields at any time (in the TimerLoop), there's no need to use InterlockedXXX to access them since it's just one thread.

I added those methods because you can't be too paranoid with multi-threaded code. In fact I should add `[MethodImpl(MethodImplOptions.NoInlining)]` attribute to both methods, even though I don't think they are candidates for inlining.

halter73:
I talked to @davidfowl in person and he doesn't really mind if I merge this. I want include this, since this a pretty limited change, and you produced some nice micro benchmark numbers.

I do have some style changes that I would like to make to this. Normally I would just comment on the relevant lines in the PR, but I made the changes myself so I could ensure that they wouldn't hurt performance (namely removing WriteLongThreadSafe and ReadLongThreadSafe).

Here are the proposed changes: https://github.com/aspnet/KestrelHttpServer/commit/5580181dfe09d71747ea559f04402074f5ef0806

I ran your speed test project and saw no perf hit on both 32-bit and 64-bit dnx.

@tenor Do my proposed changes look good to you? If so, I can just merge this with my changes.

P.S. In the future it would be better to create a github repo instead of linking to a zip archive, but it's no big deal.

tenor:
@halter73 Your changes look good to me. Please merge. :smile:

halter73:
Done! Thanks again.

tenor:
You're welcome! Thanks for looking into the PR.

davidfowl:
@tenor so you're going to send more now right?

tenor:
@davidfowl Sure, if I find more stuff. :smirk:

PR #432, opened 2015-11-29
Value tasks + Less Async

benaadams:
About as far as I can go without breaking Stream inheritance

ValueTask<T> from https://github.com/dotnet/corefx/issues/4708


benaadams:
Well, more could be done for `ForChunkedEncoding : MessageBody` `ReadAsyncAwaited` but it has 3 awaits in 4 while loops so it would have dubious value and the code complexity required be high.

benaadams:
Though cutting stream; would 1/3 the allocations here

streams

Might give it a go..?

benaadams:
Revisit when ValueTask lands; added bits to "Lighter async state + Task fast paths" #446

PR #434, opened 2015-11-29
OnConnection->OnConnectionAsync

### LINKED ISSUE #433: IConnectionFilter.OnConnection should end Async
Async methods in public Api should be suffixed as such

```csharp
public interface IConnectionFilter
{
    Task OnConnection(ConnectionFilterContext context);
}
```
Should be
```csharp
public interface IConnectionFilter
{
    Task OnConnectionAsync(ConnectionFilterContext context);
}
```

benaadams:
Resolves  #433

### DIFF
--- a/src/Microsoft.AspNet.Server.Kestrel.Https/HttpsConnectionFilter.cs
+++ b/src/Microsoft.AspNet.Server.Kestrel.Https/HttpsConnectionFilter.cs
@@ -36,9 +36,9 @@ namespace Microsoft.AspNet.Server.Kestrel.Https
             _previous = previous;
         }
 
-        public async Task OnConnection(ConnectionFilterContext context)
+        public async Task OnConnectionAsync(ConnectionFilterContext context)
         {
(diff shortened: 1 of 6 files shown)


halter73:
On one hand, I remember removing all the Async suffixes from Hub and PersistentConnection methods before releasing SignalR 1.0. On the other hand, [MVC uses the Async suffix](https://github.com/aspnet/Mvc/blob/6875ee55d3c8632a92432b4b80b139f293f14804/src/Microsoft.AspNet.Mvc.Core/Formatters/OutputFormatter.cs#L181) even when there isn't a synchronous alternative to the async method. I'm sure there are plenty more example of this.

@davidfowl @lodejard I assume we want to take this for consistency? <sup>even though I'm not the biggest fan</sup>

benaadams:
I think the -Async suffix is more about highlighting the return value is/needs to be a `Task` which potentially hasn't completed at the end of the call; whereas the non-Async versions you would assume its finished doing what you asked it to do when the call returns.

PR #437, opened 2015-11-30
[Not for merging] IStreams

benaadams:
Requires too much upstream working HttpAbstractions; and then everywhere :-/

streams

Though [frame changes](https://github.com/benaadams/KestrelHttpServer/blob/alt-stream/src/Microsoft.AspNet.Server.Kestrel/Http/Frame.cs#L335-L338) using this and supporting [FrameDuplexStream changes](https://github.com/benaadams/KestrelHttpServer/blob/alt-stream/src/Microsoft.AspNet.Server.Kestrel/Http/FrameDuplexStream.cs) would cut the above allocs from 640,000 bytes to 160,000 bytes for 4k requests; a 75% reduction in allocation.


halter73:
Did you open this PR so you could reference it in your "Stream - Interface segregation" issue and "Value Tasks PR"? Should we close this for now?

PR #441, opened 2015-12-02
Reduce string allocations per request

cesarbs:
There is a fixed set of methods and a fixed set of HTTP versions that we'll see in most requests. When we see those we can reuse the same string instead of allocating a new one. If not, we fall back to allocating a new string.

Since all methods we're checking for plus the HTTP versions fit in 8 bytes, I'm pre-computing longs containing those bytes to make the comparisons faster.

cc @halter73 @davidfowl @benaadams @DamianEdwards

### DIFF
--- a/src/Microsoft.AspNet.Server.Kestrel/Infrastructure/MemoryPoolIterator2Extensions.cs
+++ b/src/Microsoft.AspNet.Server.Kestrel/Infrastructure/MemoryPoolIterator2Extensions.cs
@@ -2,6 +2,7 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.Diagnostics;
 using System.Text;
 
 namespace Microsoft.AspNet.Server.Kestrel.Infrastructure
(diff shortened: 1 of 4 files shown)


benaadams:
Related: https://github.com/aspnet/KestrelHttpServer/pull/411

benaadams:
Related: https://github.com/aspnet/KestrelHttpServer/pull/411 which removes all string allocations on repeated requests on a keep alive - however this resolves common allocators across all requests

src/Microsoft.AspNet.Server.Kestrel/Infrastructure/MemoryPoolIterator2Extensions.cs
```diff
@@ -203,5 +243,92 @@ public static ArraySegment<byte> GetArraySegment(this MemoryPoolIterator2 start,
...
+                httpVersion = Http10Version;
+            }
+            else if (scanLong == _http11VersionLong)
+            {
+                httpVersion = Http11Version;
+            }
+
+            if (httpVersion != null)
+            {
+                for (int i = 0; i < 8; i++) scan.Take();
```
halter73: Can re avoid this by having PeekLong advance the iterator for us.

halter73:
@benaadams After discussing #411 with @davidfowl and @cesarbs we were thinking about closing it due to the relative complexity of managing a per-connection invalidating cache to reduce string allocations.

I didn't what to start that discussion before offering an alternative, and this is it. I prefer this approach, because it has a better (or at least easier to analyze) worst case performance both memory and cpu-wise. This might also have a better memory footprint per-connection since we don't have to reallocate a new cache each time.

@benaadams What do you think?

src/Microsoft.AspNet.Server.Kestrel/Infrastructure/MemoryPoolIterator2Extensions.cs
```diff
@@ -203,5 +243,92 @@ public static ArraySegment<byte> GetArraySegment(this MemoryPoolIterator2 start,
...
+
+            if (scanLong == _httpConnectMethodLong)
+            {
+                httpMethod = HttpConnectMethod;
+            }
+            else if (((scanLong ^ _httpDeleteMethodLong) << 8) == 0)
+            {
+                httpMethod = HttpDeleteMethod;
+            }
+            else if (((scanLong ^ _httpGetMethodLong) << 32) == 0)
```
halter73: Super nitpicky, but we might as well check for GETs and POSTs first since I assume they are the most common methods.

src/Microsoft.AspNet.Server.Kestrel/Infrastructure/MemoryPoolIterator2Extensions.cs
```diff
@@ -203,5 +243,92 @@ public static ArraySegment<byte> GetArraySegment(this MemoryPoolIterator2 start,
...
+        {
+            httpVersion = null;
+            var scanLong = scan.PeekLong();
+
+            if (scanLong == -1)
+            {
+                return false;
+            }
+
+            if (scanLong == _http10VersionLong)
```
halter73: ditto for checking for HTTP/1.1 first.

benaadams:
@halter73 :+1: to this as it resolves expected shared strings in known locations; also the two changes aren't in conflict.

Want a discussion over on other issue? Or was this the discussion? :wink:

halter73:
You are right that the two changes aren't mutually exclusive, but I'm still not sold on the StringPool yet. I was hoping this change would help convince you that the StringPool isn't necessary :wink:

We can can continue discussing that over on the issue though.

cesarbs:
I was able to optimize this even further. Given we have a set of 11 known strings that will never change, I went looking for a divisor that would yield a unique modulo value for each string's long representation. Turns out there is such a value (37). So now we have a perfect hash of the known strings and matching the input to those is just a matter of clearing uninteresting bits in the input and looking up the resulting value, then making sure the longs are actually the same.

src/Microsoft.AspNet.Server.Kestrel/Properties/launchSettings.json
```diff
@@ -0,0 +1,3 @@
+{
```
cesarbs: revert

test/Microsoft.AspNet.Server.Kestrel.FunctionalTests/Properties/launchSettings.json
```diff
@@ -0,0 +1,3 @@
+{
```
cesarbs: revert

test/Microsoft.AspNet.Server.KestrelTests/Properties/launchSettings.json
```diff
@@ -0,0 +1,3 @@
+{
```
cesarbs: revert

src/Microsoft.AspNet.Server.Kestrel.Https/Properties/launchSettings.json
```diff
@@ -0,0 +1,3 @@
+{
```
cesarbs: revert

benaadams:
> So now we have a perfect hash of the known strings and matching the input to those is just a matter of ...

Now we're cooking :+1:

pakrym:
I think we need to make sure that somebody won't accidentally get hash collision when adding new string to known strings. And add more comments to intialization code, because without this issue context it looks spooky

cesarbs:
@pakrym Goot point :+1: I'll add comments explaining what is going on.

cesarbs:
This is really unfortunate, but the 30% figure from my initial tests where for very simple requests (no headers, I should've thought better about that). With more realistic requests the reduction is a lot smaller (around 1%) :disappointed:

I'm looking for more places where I can apply a similar optimization.

benaadams:
LGTM

Only comment would be `GetKnownString` could be split into two `GetHttpVersionString` and `GetMethodString` but might be something for future when the hash is broken and their are more types.

halter73:
I'm waiting on verification that this will work on big-endian architectures before merging.

cesarbs:
@halter73 It's going to be hard to check that. Raspberry Pi distros set the processor do little endian. Actually little endian seems to be the default on ARM environments. I'm not sure it's worth verifying this right now. I could enter a bug to track that and move on with this change.

cesarbs:
I've emailed @stephentoub asking if they test Core on big endian. If they do I might try the same as them to test this.

benaadams:
Would there be problems with the Frame header collection also?

cesarbs:
@benaadams Can you elaborate? I don't see what you mean.

benaadams:
@cesarbs like https://github.com/aspnet/KestrelHttpServer/blob/dev/src/Microsoft.AspNet.Server.Kestrel/Http/FrameHeaders.Generated.cs#L8127

cesarbs:
@benaadams Oh, with regard to endianess, you mean? I hadn't looked at that code, but it's likely it might be affected by it.

cesarbs:
Ok, I did much better controlled tests and these are the results I got:

Before:

image


After:

image

Before:

image

After:

image

(Look at the allocation percentages in the last two)

That looks like some improvement to me :grinning: 

I tested it with wrk, from a remote machine:

    wrk -c 256 -t 32 -d 10 http://<local address>:5000

cesarbs:
@stephentoub replied that they don't test Core on big endian. Again I'd say we can postpone verification on big endian.

halter73:
@cesarbs Aside from potentially addressing further feedback, do you think this PR is complete?

stephentoub:
> @stephentoub replied that they don't test Core on big endian. Again I'd say we can postpone verification on big endian.

I'd suggest at least adding a ```Debug.Assert(BitConverter.IsLittleEndian)``` to the relevant code.

cesarbs:
@halter73 Yes.

cesarbs:
@stephentoub I'd rather not. This code likely does not work on big endian, **but** it doesn't break things. It'll just go through a less optimized path.

stephentoub:
> I'd rather not. This code likely does not work on big endian, but it doesn't break things. It'll just go through a less optimized path

I've not looked at the code in depth. Just looked like wrong values would be computed in big endian. If that's not true, then an assert isn't valuable. But if running on a big endian system would start resulting in erroneous results, an assert would help to point out the problem immediately, and there's little downside. Like I said, though, I've not looked much at the code, so you know better than I whether there's a problem.

cesarbs:
@stephentoub On a second thought, I noticed someone could craft some weird requests (not necessarily malicious) and this would not behave as expected. I'll add the assert.

cesarbs:
Went with a slightly different approach. Instead of an assert I'm just checking `BitConverter.IsLittleEndian` and skipping the optimized path if it's false.

cesarbs:
Squashed.

AspNetSmurfLab:
Branchmarks:

**dev**

	Running 15s test @ http://10.0.0.100:5001/plaintext
		Latency     5.10ms   12.44ms 393.12ms   96.10%
		Req/Sec    24.62k     2.11k   67.62k    93.12%
	11787922 requests in 15.10s, 1.45GB read
	Socket errors: connect 0, read 0, write 177, timeout 0
	Requests/sec: 780789.10
	Transfer/sec:     98.29MB

**This PR**

	Running 15s test @ http://10.0.0.100:5001/plaintext
		Latency     4.94ms   12.55ms 211.73ms   94.72%
		Req/Sec    25.26k     1.87k   54.54k    87.10%
	12106391 requests in 15.10s, 1.49GB read
	Requests/sec: 801743.82
	Transfer/sec:    100.93MB

@cesarbs ran those several times and the RPS was consistently around those marks for each branch.

benaadams:
LGTM! :+1:

cesarbs:
Ping.

halter73:
:shipit:

cesarbs:
Yay :grinning:

PR #443, opened 2015-12-02
Log ThreadPool exceptions

benaadams:
Updated  #422 - Rebased and logging the exceptions now

Curry `IKestrelTrace` in `ThreadPoolActions` ctor so its only closure captured once; rather than per call.

/cc @halter73 improved?

### DIFF
--- a/src/Microsoft.AspNet.Server.Kestrel/ServiceContext.cs
+++ b/src/Microsoft.AspNet.Server.Kestrel/ServiceContext.cs
@@ -21,7 +21,7 @@ namespace Microsoft.AspNet.Server.Kestrel
         {
             AppLifetime = context.AppLifetime;
             Log = context.Log;
-            ThreadPoolActions = context.ThreadPoolActions;
+            ThreadPool = context.ThreadPool;
             FrameFactory = context.FrameFactory;
             DateHeaderValueManager = context.DateHeaderValueManager;
             ConnectionFilter = context.ConnectionFilter;
@@ -32,7 +32,7 @@ namespace Microsoft.AspNet.Server.Kestrel
 
         public IKestrelTrace Log { get; set; }
 
-        public ThreadPoolActions ThreadPoolActions { get; set; }
+        public IThreadPool ThreadPool { get; set; }
 
         public Func<ConnectionContext, IPEndPoint, IPEndPoint, Action<IFeatureCollection>, Frame> FrameFactory { get; set; }
 
--- a/src/Microsoft.AspNet.Server.Kestrel/Filter/FilteredStreamAdapter.cs
+++ b/src/Microsoft.AspNet.Server.Kestrel/Filter/FilteredStreamAdapter.cs
@@ -20,9 +20,9 @@ namespace Microsoft.AspNet.Server.Kestrel.Filter
             Stream filteredStream,
             MemoryPool2 memory,
             IKestrelTrace logger,
-            ThreadPoolActions threadPoolActions)
+            IThreadPool threadPool)
         {
-            SocketInput = new SocketInput(memory, threadPoolActions);
+            SocketInput = new SocketInput(memory, threadPool);
             SocketOutput = new StreamSocketOutput(filteredStream, memory);
 
             _log = logger;
--- a/src/Microsoft.AspNet.Server.Kestrel/Http/Connection.cs
+++ b/src/Microsoft.AspNet.Server.Kestrel/Http/Connection.cs
@@ -41,8 +41,8 @@ namespace Microsoft.AspNet.Server.Kestrel.Http
 
             _connectionId = Interlocked.Increment(ref _lastConnectionId);
 
-            _rawSocketInput = new SocketInput(Memory2, ThreadPoolActions);
-            _rawSocketOutput = new SocketOutput(Thread, _socket, Memory2, this, _connectionId, Log, ThreadPoolActions);
+            _rawSocketInput = new SocketInput(Memory2, ThreadPool);
+            _rawSocketOutput = new SocketOutput(Thread, _socket, Memory2, this, _connectionId, Log, ThreadPool);
         }
 
         public void Start()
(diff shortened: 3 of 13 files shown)


src/Microsoft.AspNet.Server.Kestrel/Infrastructure/ThreadPoolActions.cs
```diff
@@ -0,0 +1,153 @@
...
+            while (block != null)
+            {
+                var returningBlock = block;
+                block = returningBlock.Next;
+
+                returningBlock.Pool?.Return(returningBlock);
+            }
+        }
+
+        private static void ReturnWrittenBlocksInternal(MemoryPoolBlock2 block)
```
halter73: I don't like moving the WaitCallback implementations outside of the class queuing the work. I noticed we were doing this with Frame.Abort in the previous PR and probably should have brought it up then.

I know this is being done to avoid per-call closure allocations, but I think think can be done by making the state parameter include the logger we could achieve the same thing.

benaadams: State is an object would box as a struct; and would need an allocation to include the logger or a change in data structures. However... can do it another way. 5 mins...

benaadams: Might have to admit defeat; the issue is the `try {} catch {}` wrapper more than the logger as the `WaitCallback` is the wrapper and the state needs `Action+state` or `Action+state+logger`.

What if the callback called back into a static method in the correct class? Still a bit tightly coupled though...

halter73: I don't think we have to admit defeat quite yet. In this case for example, couldn't the WriteContext itself be the state object? We could add a logger field.

halter73: I don't think we have to admit defeat quite yet. In this example, couldn't the state object be the WriteContext itself? You could access the logger via `WriteContext.Self._log`. Hopefully we could do something similar for the other ones.

benaadams: Would clash with #363 - however... ReturnBlocks shouldn't currently ever throw, also Frame.Abort; revert both to ThreadPool? (probably spending too much time on it) Then just do it for user code triggering stuff?

halter73: I'm OK with that.

src/Microsoft.AspNet.Server.Kestrel/Infrastructure/ThreadPoolActions.cs
```diff
@@ -0,0 +1,153 @@
...
+                {
+                    _log.ApplicationError(e);
+                }
+            };
+
+            _completeTcs = (o) =>
+            {
+                try
+                {
+                    ((TaskCompletionSource<object>)o).SetResult(null);
```
halter73: If we use the `TrySet` methods we could avoid the try/catch.

benaadams: It stops the setting of the value throwing rather than the synchronous running of the continuations from throwing. From digging into Task & tcs its probably worthwhile anyway as `SetResult` calls `TrySetResult`; then decides to throw based on the return.

However this would be the top of the stack the the continuations (which is why I think its put on the threadpool); so I think the try catch is still needed.

benaadams: (For reference it ends up in this code, eventually, on SetResult https://github.com/dotnet/coreclr/blob/29526273e376bdf647102388260907b70778f9df/src/mscorlib/src/System/Threading/Tasks/Task.cs#L3581)

halter73: Good call.

-----Original Message-----

src/Microsoft.AspNet.Server.Kestrel/ServiceContext.cs
```diff
@@ -29,7 +32,20 @@ public ServiceContext(ServiceContext context)
...
+        public IKestrelTrace Log
+        {
+            get
+            {
+                return _log;
+            }
+            set
+            {
+                _log = value;
+                ThreadPoolActions = new ThreadPoolActions(_log);
```
halter73: Nit: Can we just set this in KestrelServer rather than having a custom getter and setter for Log?

benaadams: Done

src/Microsoft.AspNet.Server.Kestrel/ServiceContext.cs
```diff
@@ -20,7 +22,8 @@ public ServiceContext()
         public ServiceContext(ServiceContext context)
         {
             AppLifetime = context.AppLifetime;
-            Log = context.Log;
+            _log = context.Log;
```
halter73: You gotta change this back too.

benaadams: Gah, done.

src/Microsoft.AspNet.Server.Kestrel/Infrastructure/ThreadPoolActions.cs
```diff
@@ -0,0 +1,74 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.AspNet.Server.Kestrel.Http;
+
+namespace Microsoft.AspNet.Server.Kestrel.Infrastructure
+{
+    public class ThreadPoolActions
```
davidfowl: IThreadPool?

benaadams: Betterer?

PR #446, opened 2015-12-03
Lighter async state + Task fast paths

benaadams:
Selected excerpts from "Value tasks + Less Async" #432


src/Microsoft.AspNet.Server.Kestrel/Http/Frame.cs
```diff
@@ -501,10 +523,27 @@ public void ProduceContinue()
...
         {
-            if (_responseStarted) return;
+            if (_responseStarted) return TaskUtilities.CompletedTask;
+
+            if (_onStarting != null)
+            {
+                return FireOnStartingProduceStart(immediate: immediate);
+            }
 
+            if (_applicationException != null)
```
davidfowl: This isn't new code right?

benaadams: No; its just surfaced here for the non-async path; previous gets triggered if it goes down the async path of `FireOnStartingProduceStart` 3 lines above.

PR #447, opened 2015-12-03
Use pooled memory for Stream CopyToAsync

benaadams:
Default buffer size Is crazy large (82kB); and as is used for lifetime of connection; per connection - likely ends up in Gen2

(Ref https://github.com/dotnet/coreclr/issues/2223)

### DIFF
--- a/src/Microsoft.AspNet.Server.Kestrel/Filter/FilteredStreamAdapter.cs
+++ b/src/Microsoft.AspNet.Server.Kestrel/Filter/FilteredStreamAdapter.cs
@@ -28,9 +28,13 @@ namespace Microsoft.AspNet.Server.Kestrel.Filter
             _filteredStream = filteredStream;
             _socketInputStream = new SocketInputStream(SocketInput);
 
-            // Don't use 81920 byte buffer
-            _filteredStream.CopyToAsync(_socketInputStream, 4096).ContinueWith((task, state) =>
+            var block = memory.Lease();
+            // Use pooled block for copy
+            _filteredStream.CopyToAsync(_socketInputStream, block).ContinueWith((task, state) =>
             {
+                var returnedBlock = task.Result;
+                returnedBlock.Pool?.Return(returnedBlock);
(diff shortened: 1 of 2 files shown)


src/Microsoft.AspNet.Server.Kestrel/Filter/FilteredStreamAdapter.cs
```diff
@@ -28,8 +28,13 @@ public class FilteredStreamAdapter
             _filteredStream = filteredStream;
             _socketInputStream = new SocketInputStream(SocketInput);
 
-            _filteredStream.CopyToAsync(_socketInputStream).ContinueWith((task, state) =>
+            var block = memory.Lease();
+            // Use pooled block for copy
+            _filteredStream.CopyToAsync(_socketInputStream, block).ContinueWith((task, state) =>
             {
+                var returnedBlock = task.Result;
+                returnedBlock.Pool?.Return(returnedBlock);
```
davidfowl: Why can the pool be null?

benaadams: If `SocketInput` asks for blocks larger than a pooled block; and then returns them to pool - had a related PR https://github.com/aspnet/KestrelHttpServer/pull/402 to stop it; wasn't sure - too much churn at the time.

halter73: In this case, since we aren't providing a minimum size to `memory.Lease()`, value it pool should never be null.

benaadams: Weird; thought non-pooled blocks could get returned; but looks like that changed way back...

benaadams: Added PR #448 to address this.

PR #453, opened 2015-12-05
Concurrent read error propagation fix

### LINKED ISSUE #451: Concurrent reads are not supported consequences
```
Unhandled Exception: System.InvalidOperationException: Concurrent reads are not supported.
   at Microsoft.AspNet.Server.Kestrel.Http.SocketInput.OnCompleted(Action continuation)
   at System.Runtime.CompilerServices.AsyncTaskMethodBuilder`1.AwaitUnsafeOnCompleted[TAwaiter,TStateMachine](TAwaiter& awaiter, TStateMachine& stateMachine)
   at System.Runtime.CompilerServices.AsyncMethodBuilderCore.<>c.<ThrowAsync>b__6_1(Object state)
   at System.Threading.ExecutionContext.Run(ExecutionContext executionContext, ContextCallback callback, Object state)
   at System.Threading.QueueUserWorkItemCallback.System.Threading.IThreadPoolWorkItem.ExecuteWorkItem()
   at System.Threading.ThreadPoolWorkQueue.Dispatch()
```

benaadams:
Resolves #451 server crash

```
Unhandled Exception: System.InvalidOperationException: Concurrent reads are not supported.
   at Microsoft.AspNet.Server.Kestrel.Http.SocketInput.OnCompleted(Action continuation)
   at System.Runtime.CompilerServices.AsyncTaskMethodBuilder`1.AwaitUnsafeOnCompleted[TAwaiter,TStateMachine](TAwaiter& awaiter, TStateMachine& stateMachine)
   at System.Runtime.CompilerServices.AsyncMethodBuilderCore.<>c.<ThrowAsync>b__6_1(Object state)
   at System.Threading.ExecutionContext.Run(ExecutionContext executionContext, ContextCallback callback, Object state)
   at System.Threading.QueueUserWorkItemCallback.System.Threading.IThreadPoolWorkItem.ExecuteWorkItem()
   at System.Threading.ThreadPoolWorkQueue.Dispatch()
```

/cc @halter73 @lodejard

### DIFF
--- a/src/Microsoft.AspNet.Server.Kestrel/Http/SocketInput.cs
+++ b/src/Microsoft.AspNet.Server.Kestrel/Http/SocketInput.cs
@@ -216,8 +216,16 @@ namespace Microsoft.AspNet.Server.Kestrel.Http
             }
             else
             {
-                // THIS IS AN ERROR STATE - ONLY ONE WAITER CAN WAIT
-                throw new InvalidOperationException("Concurrent reads are not supported.");
+                _awaitableError = new InvalidOperationException("Concurrent reads are not supported.");
+
+                awaitableState = Interlocked.Exchange(
+                    ref _awaitableState,
+                    _awaitableIsCompleted);
+
+                _manualResetEvent.Set();
+
+                _threadPool.Run(continuation);
+                _threadPool.Run(awaitableState);
             }
         }
 
--- a/test/Microsoft.AspNet.Server.KestrelTests/SocketInputTests.cs
+++ b/test/Microsoft.AspNet.Server.KestrelTests/SocketInputTests.cs
@@ -1,0 +1,94 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Threading.Tasks;
+using Microsoft.AspNet.Server.Kestrel;
+using Microsoft.AspNet.Server.Kestrel.Http;
+using Microsoft.AspNet.Server.Kestrel.Infrastructure;
+using Xunit;
+
+namespace Microsoft.AspNet.Server.KestrelTests
+{
+    public class SocketInputTests
+    {
+        [Fact]
+        public async Task ConcurrentReadsFailGracefully()
+        {
+            // Arrange
+            var trace = new KestrelTrace(new TestKestrelTrace());
+            var ltp = new LoggingThreadPool(trace);
+            using (var memory2 = new MemoryPool2())
+            {
+                var socketInput = new SocketInput(memory2, ltp);
+
+                var task0Threw = false;
+                var task1Threw = false;
+                var task2Threw = false;
+
+
+                var task0 = AwaitAsTaskAsync(socketInput);
+
+                Assert.False(task0.IsFaulted);
+
+                var task = task0.ContinueWith(
+                    (t) =>
+                    {
+                        TestConcurrentFaultedTask(t);
+                        task0Threw = true;
+                    },
+                    TaskContinuationOptions.OnlyOnFaulted);
+
+                Assert.False(task0.IsFaulted);
+
+                // Awaiting/continuing two tasks faults both
+
+                var task1 = AwaitAsTaskAsync(socketInput);
+
+                await task1.ContinueWith(
+                    (t) =>
+                    {
+                        TestConcurrentFaultedTask(t);
+                        task1Threw = true;
+                    },
+                    TaskContinuationOptions.OnlyOnFaulted);
+
+                await task;
+
+                Assert.True(task0.IsFaulted);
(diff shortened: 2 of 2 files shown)


src/Microsoft.AspNet.Server.Kestrel/Http/SocketInput.cs
```diff
@@ -217,7 +217,10 @@ public void OnCompleted(Action continuation)
             else
             {
                 // THIS IS AN ERROR STATE - ONLY ONE WAITER CAN WAIT
-                throw new InvalidOperationException("Concurrent reads are not supported.");
```
halter73: I wonder if we should also set `_awaitableError` so the exception gets surfaced to the reader.

davidfowl: And also trigger the callback right?

halter73: Right. @lodejard suggested putting all `await SocketInput`s behind a [TaskQueue](https://github.com/SignalR/SignalR/blob/3a219c101587333d562f04f30903d50ad9773e1c/src/Microsoft.AspNet.SignalR.Core/Infrastructure/TaskQueue.cs) to prevent the possibility of concurrent reads. Should I create an issue for that? It would be interesting to see how that affects performance.

davidfowl: Yes, make an issue.

halter73: Here you go: https://github.com/aspnet/KestrelHttpServer/issues/456

davidfowl: Also think we should make the error bubble to app code instead of throwing on some background thread. Set the error and trigger the callback.

halter73: Sounds good to me :+1:

davidfowl: Reuse the logic in AbortAwaiting.

benaadams: Done

lodejard:
Should we put in the concurrent queue instead of a short term fix?

benaadams:
@lodejard Then I'd never find where it happened, am not intentionally double reading :(

Occurred 3 times in 30 mins across about 100M+ WebSocket messages with coreclr x64; over https using Kestrel.Https filter- I'd have perferred if it errorred and told me where it was happening - so I could fix either kestrel, filter, websocket middleware; ssl stream or my app :)

Still can't track it down or reproduce in load testing :(

benaadams:
Though obviously the server crashing is a worse situation... :P

benaadams:
Added change to error both reads

halter73:
@benaadams Has this fix helped you pinpoint the source of the concurrent reads?

benaadams:
@halter73 not yet, as I can't get a concurrent read to trigger - working on it. It happened during weekend playtest where webserver was open to the public to try; so I've got to get a bit messier.

I'm going to trigger a concurrent read purposefully and make sure the outcome is well behaved (disconnect would be well behaved - as far as I'm concerned); and handled (reconnect, re-establish control to spaceship) - and I might have to wait till the next public pre-alpha test to catch it in the act.

Though will be running some extensive testing meanwhile as the websockets weren't 100% well behaved so want to track those issues down too.

halter73:
Cool. It's awesome to have real-world testing going on with so many messages being sent.

I think we should get a unit test in for this before merging this PR.

benaadams:
Ok, will make sure I can trigger it :D

benaadams:
> Cool. It's awesome to have real-world testing going on with so many messages being sent.

On the positive flip-side... 

When we go live we are going to be running on Service Fabric, so it will be migrating the servers/microservices around and the effect will essentially be crashing the connections as they hop from host to host - so I did get to test our client reconnection logic. 

By the 3rd crash I was on the ball; restarted pretty quick and no-one really noticed it. I haven't been brave enough to restart the server mid-test in any test so far to check if it really works! So from that side of things am pretty pleased :+1:

benaadams:
Found the error while setting up tests if you do something like this it counts as a double await:
```csharp
await AwaitAsTaskAsync(SocketInput socketInput).ContinueWith(
    (t) =>
    {
        taskThrew = true;
    },
    TaskContinuationOptions.OnlyOnFaulted);

private async Task AwaitAsTaskAsync(SocketInput socketInput)
{
    await socketInput;
}
```

benaadams:
Added test

halter73:
Nice! I was about to write the test myself (which is probably fair since I added the throw). I just ran your test with the existing code and it crashed the test runner. Good sign :smile:

PR #455, opened 2015-12-07
Trim req/resp streams by a field each

benaadams:
Can only be in one of the states; not both

### DIFF
--- a/src/Microsoft.AspNet.Server.Kestrel/Http/FrameRequestStream.cs
+++ b/src/Microsoft.AspNet.Server.Kestrel/Http/FrameRequestStream.cs
@@ -11,8 +11,7 @@ namespace Microsoft.AspNet.Server.Kestrel.Http
     public class FrameRequestStream : Stream
     {
         private readonly MessageBody _body;
-        private bool _stopped;
-        private bool _aborted;
+        private StreamState _state;
 
         public FrameRequestStream(MessageBody body)
         {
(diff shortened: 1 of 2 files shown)


src/Microsoft.AspNet.Server.Kestrel/Http/FrameRequestStream.cs
```diff
@@ -145,14 +116,38 @@ public void StopAcceptingReads()
...
         public void Abort()
         {
             // We don't want to throw an ODE until the app func actually completes.
             // If the request is aborted, we throw an IOException instead.
-            _aborted = true;
+            if (_state != StreamState.Disposed)
+            {
+                _state = StreamState.Aborted;
+            }
+        }
```
halter73: Nit: needs newline

src/Microsoft.AspNet.Server.Kestrel/Http/FrameResponseStream.cs
```diff
@@ -110,14 +81,39 @@ public void StopAcceptingWrites()
...
+            switch (_state)
+            {
+                case StreamState.Open:
+                    return;
+                case StreamState.Disposed:
+                    throw new ObjectDisposedException(nameof(FrameResponseStream));
+                case StreamState.Aborted:
+                    throw new IOException("The request has been aborted.");
+                default:
+                    throw new IOException("Invalid Stream State.");
```
halter73: Does the compilere require this considering StreamState has only 3 defined values that are already checked?

src/Microsoft.AspNet.Server.Kestrel/Http/FrameRequestStream.cs
benaadams: Done

src/Microsoft.AspNet.Server.Kestrel/Http/FrameResponseStream.cs
benaadams: Remove

PR #457, opened 2015-12-07
Configurable thread count

khellang:


src/Microsoft.AspNet.Server.Kestrel/KestrelServerInformation.cs
```diff
@@ -39,5 +39,34 @@ private static ICollection<string> GetAddresses(IConfiguration configuration)
 
             return addresses;
         }
+
+        private static int GetThreadCount(IConfiguration configuration)
+        {
+            var threadCountString = configuration["server.threadCount"];
```
khellang: Can this return `null`?

halter73: @Tratcher What are your thoughts on this? Should all the properties on IKestrelServerInformation be configurable through IConfiguration? Or should Kestrel-specific things like ThreadCount and NoDelay should those only be configurable by directly modifying the IKestrelServerInformation object?

src/Microsoft.AspNet.Server.Kestrel/ServerFactory.cs
```diff
@@ -27,36 +27,11 @@ public ServerFactory(IApplicationLifetime appLifetime, ILoggerFactory loggerFact
...
-            var threadCount = GetThreadCount();
-            var information = new KestrelServerInformation(configuration, threadCount);
+            var information = new KestrelServerInformation(configuration);
             var serverFeatures = new FeatureCollection();
             serverFeatures.Set<IKestrelServerInformation>(information);
             serverFeatures.Set<IServerAddressesFeature>(information);
             return new KestrelServer(serverFeatures, _appLifetime, _loggerFactory.CreateLogger("Microsoft.AspNet.Server.Kestrel"));
         }
-
-        private static int GetThreadCount()
```
halter73: Did you move this to make it easier to test?

test/Microsoft.AspNet.Server.KestrelTests/KestrelServerInformationTests.cs
```diff
@@ -0,0 +1,70 @@
...
+
+using System;
+using System.Collections.Generic;
+using Microsoft.AspNet.Server.Kestrel;
+using Microsoft.Extensions.Configuration;
+using Xunit;
+
+namespace Microsoft.AspNet.Server.KestrelTests
+{
+    public class KestrelServerInformationTests
```
halter73: While we're adding tests, can we add a test to verify that when you try to call `KestrelServer.Start` with `ThreadCount==0`, it throws?

src/Microsoft.AspNet.Server.Kestrel/KestrelServerInformation.cs
Tratcher: Adding more things to the config is just fine, but give kestrel specific entries like this a kestrel specific name. E.g. `kestrel.threadCount`.

src/Microsoft.AspNet.Server.Kestrel/KestrelServerInformation.cs
```diff
@@ -39,5 +39,34 @@ private static ICollection<string> GetAddresses(IConfiguration configuration)
...
+
+            int threadCount;
+            if (string.IsNullOrEmpty(threadCountString) || !int.TryParse(threadCountString, out threadCount))
```
Tratcher: Invariant Culture.

Do you want to range check the result here?

src/Microsoft.AspNet.Server.Kestrel/ServerFactory.cs
khellang: Not really, but yes, I guess it kinda makes it easier to test as well :smile:

test/Microsoft.AspNet.Server.KestrelTests/KestrelServerInformationTests.cs
khellang: I can send another PR with this :+1:

src/Microsoft.AspNet.Server.Kestrel/KestrelServerInformation.cs
khellang: I guess it makes sense to range check things as early as possible, but it's already checked before it's used (since it can be overriden in `Configure`. What do you think? Fail fast?

src/Microsoft.AspNet.Server.Kestrel/KestrelServerInformation.cs
khellang: @Tratcher So you want me to change this to `kestrel.threadCount`?

test/Microsoft.AspNet.Server.KestrelTests/KestrelServerInformationTests.cs
halter73: Or you can add it to this PR. Either way is fine.

src/Microsoft.AspNet.Server.Kestrel/KestrelServerInformation.cs
halter73: I'm OK with just checking in KestrelServer.

src/Microsoft.AspNet.Server.Kestrel/KestrelServerInformation.cs
halter73: Yes. Can you also add `kestrel.NoDelay` for consistency?

khellang: Also, are you saying that `kestre.noDelay` is up for grabs as well? :smile:

src/Microsoft.AspNet.Server.Kestrel/KestrelServerInformation.cs
khellang: What if I move the check into the ThreadCount setter? That way it would ensure that the value is always valid?

test/Microsoft.AspNet.Server.KestrelTests/KestrelServerInformationTests.cs
khellang: I'll keep it for another PR. I'm behind on http://24pullrequests.com/ :wink:

src/Microsoft.AspNet.Server.Kestrel/KestrelServerInformation.cs
khellang: Oh, didn't see the comment until I refreshed. I'll add it ASAP.

Issue #265, opened 2015-10-19
Microsoft.AspNet.Server.Kestrel.FunctionalTests failures on CoreCLR on Linux


cesarbs:
info: Exec
    info:   program: /usr/bin/env
    info:   commandline: bash -c ". \"/home/cesars/src/aspnet/universe/KestrelHttpServer/packages/KoreBuild/build/dnvm.sh\"; dnvm run default -runtime coreclr test"
    info:   workingdir: test/Microsoft.AspNet.Server.Kestrel.FunctionalTests
    xUnit.net DNX Runner (64-bit DNXCore 5.0)
      Discovering: Microsoft.AspNet.Server.Kestrel.FunctionalTests
      Discovered:  Microsoft.AspNet.Server.Kestrel.FunctionalTests
      Starting:    Microsoft.AspNet.Server.Kestrel.FunctionalTests
          System.Threading.Tasks.TaskCanceledException : A task was canceled.
          Stack Trace:
               at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
               at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
               at System.Runtime.CompilerServices.TaskAwaiter`1.GetResult()
               at Microsoft.AspNet.Server.Kestrel.FunctionalTests.ResponseTests.<LargeDownload>d__0.MoveNext()
            --- End of stack trace from previous location where exception was thrown ---
               at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
               at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
            --- End of stack trace from previous location where exception was thrown ---
               at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
               at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
            --- End of stack trace from previous location where exception was thrown ---
               at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
               at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
        Microsoft.AspNet.Server.Kestrel.FunctionalTests.ThreadCountTests.ZeroToTenThreads [SKIP]
[...]

PR #465, opened 2015-12-09
Skip more tests on Mono to prevent hangs

halter73:
- It seems like I missed some FrameworkSkipConditions when I merged these tests


halter73:
This should help address https://github.com/aspnet/KestrelHttpServer/issues/265

Tratcher:
:shipit:

JunTaoLuo:
:shipit: ignore the appveyor/brach failure, I accidentally cancelled that one when cancelling my other tests.

PR #467, opened 2015-12-11
Use ValueTask

benaadams:
For lighter async and sync reads when already completed synchronously 

+2.6% RPS (Variable)

Before
    Latency     6.99ms   16.43ms 298.51ms   97.60%
    Req/Sec    65.03k     7.15k  214.14k    85.26%
  60288749 requests in 30.10s, 7.41GB read
Requests/sec: 2002987.40
Transfer/sec:    252.15MB

After 
    Latency     7.42ms   19.18ms 640.49ms   96.76%
    Req/Sec    64.64k    11.13k  162.59k    89.40%
  61826961 requests in 30.06s, 7.60GB read
  Socket errors: connect 35, read 0, write 0, timeout 27
Requests/sec: 2056488.77
Transfer/sec:    258.88MB

### DIFF
--- a/src/Microsoft.AspNet.Server.Kestrel/Filter/LibuvStream.cs
+++ b/src/Microsoft.AspNet.Server.Kestrel/Filter/LibuvStream.cs
@@ -11,8 +11,12 @@ namespace Microsoft.AspNet.Server.Kestrel.Filter
 {
     public class LibuvStream : Stream
     {
+        private readonly static Task<int> _initialCachedTask = Task.FromResult(0);
+
         private readonly SocketInput _input;
         private readonly ISocketOutput _output;
+
+        private Task<int> _cachedTask = _initialCachedTask;
 
         public LibuvStream(SocketInput input, ISocketOutput output)
         {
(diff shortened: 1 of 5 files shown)


src/Microsoft.AspNet.Server.Kestrel/Filter/LibuvStream.cs
```diff
@@ -11,9 +11,13 @@ namespace Microsoft.AspNet.Server.Kestrel.Filter
 {
     public class LibuvStream : Stream
     {
+        private readonly static Task<int> _initalCachedTask = Task.FromResult(0);
```
halter73: Nit: spelling

src/Microsoft.AspNet.Server.Kestrel/Http/MessageBody.cs
```diff
@@ -156,19 +156,39 @@ public ForContentLength(bool keepAlive, int contentLength, FrameContext context)
                 _inputLength = _contentLength;
             }
 
-            public override async Task<int> ReadAsyncImplementation(ArraySegment<byte> buffer, CancellationToken cancellationToken)
+            public override ValueTask<int> ReadAsyncImplementation(ArraySegment<byte> buffer, CancellationToken cancellationToken)
             {
                 var input = _context.SocketInput;
 
-                var limit = buffer.Array == null ? _inputLength : Math.Min(buffer.Count, _inputLength);
```
halter73: Is Math.Min slower? If not, I think this version is a little easier to read.

benaadams: Probably not in any significant way

halter73:
:shipit:

benaadams:
Rebased and addressed feedback

halter73:
Thanks!

PR #466, opened 2015-12-10
SocketOutput Allocation Reductions

benaadams:
Continuation of "Resuse writes, initalize queues" #363

Resolves #165 Pool UvWriteReq objects
Resolves #289 High Allocator DoWriteIfNeeded
Resolves #290 High Allocator SocketOutput.Write

PR #469, opened 2015-12-12
[Proposal] Pool WriteContext per thread

### LINKED ISSUE #165: Pool UvWriteReq objects

### LINKED ISSUE #289: [Perf] High Allocator DoWriteIfNeeded

### LINKED ISSUE #290: [Perf] High Allocator SocketOutput.Write

benaadams:
Continuation of SocketOutput Allocation Reductions #466

With extra of "Pool WriteContext per thread rather than connection"

Resolves #165 Pool UvWriteReq objects
Resolves #289 High Allocator DoWriteIfNeeded
Resolves #290 High Allocator SocketOutput.Write

### DIFF
--- a/src/Microsoft.AspNet.Server.Kestrel/Http/SocketOutput.cs
+++ b/src/Microsoft.AspNet.Server.Kestrel/Http/SocketOutput.cs
@@ -17,6 +17,7 @@ namespace Microsoft.AspNet.Server.Kestrel.Http
         private const int _maxPendingWrites = 3;
         private const int _maxBytesPreCompleted = 65536;
         private const int _initialTaskQueues = 64;
+        private const int _maxPooledWriteContexts = 32;
 
         private static WaitCallback _returnBlocks = (state) => ReturnBlocks((MemoryPoolBlock2)state);
 
@@ -38,6 +39,7 @@ namespace Microsoft.AspNet.Server.Kestrel.Http
 
         // This locks access to to all of the below fields
         private readonly object _contextLock = new object();
+        private bool _isDisposed = false;
 
         // The number of write operations that have been scheduled so far
         // but have not completed.
@@ -48,6 +50,7 @@ namespace Microsoft.AspNet.Server.Kestrel.Http
         private WriteContext _nextWriteContext;
         private readonly Queue<TaskCompletionSource<object>> _tasksPending;
         private readonly Queue<TaskCompletionSource<object>> _tasksCompleted;
+        private readonly Queue<WriteContext> _writeContextPool;
 
         public SocketOutput(
             KestrelThread thread,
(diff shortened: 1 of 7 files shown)


benaadams:
Need to look into this a bit more, weird things happening.

benaadams:
Wider pooling obv needed Self resetting...

src/Microsoft.AspNet.Server.Kestrel/Http/SocketOutput.cs
```diff
@@ -14,12 +14,18 @@ namespace Microsoft.AspNet.Server.Kestrel.Http
...
+        public const int MaxPooledWriteReqs = 1024;
+
         private const int _maxPendingWrites = 3;
         private const int _maxBytesPreCompleted = 65536;
         private const int _initialTaskQueues = 64;
+        private const int _maxWriteContextsPooledPerThread = 128;
 
         private static WaitCallback _returnBlocks = (state) => ReturnBlocks((MemoryPoolBlock2)state);
 
+        [ThreadStatic]
```
halter73: I'm afraid of adding a ThreadStatic. We don't use ThreadStatics anywhere else in Kestrel that I'm aware of.

I just checked the default maximum number of worker threads on my machine, and it was 32,767. I don't like having potentially that many static queues. Considering that each queue could pool up to 128 WriteContexts (and you just calculated the size WriteContext being 88 bytes), that's up to ~370MB of WriteContexts that will never be freed. That's never even considering the size of the queues themselves.

halter73:
There's nothing in here that I need if I just want to merge #466, right?

benaadams:
> There's nothing in here that I need if I just want to merge #466, right?

@halter73 Nope. #466 is good to merge.

src/Microsoft.AspNet.Server.Kestrel/Http/SocketOutput.cs
benaadams: Don't hold off merging #466 based on this (no issues in that one; and this is straight on top)

Know what you mean about `ThreadStatic`s though been doing some investigation. 

So the `ThreadStatic` is GC'd after thread exit; assuming it has no other outstanding strong references - which should be ok for this. Objects with a finalizer aren't so good as they'll hang around till the finalizer runs - but the WriteContext that's been reset is just a POCO managed type; so the queue will be happily GC'd when the threadpool scales back down.

If you are up to 32k of live threadpool threads; then the 32GB of thread stack space they need might be more of an issue than the 370MB of pooled `WriteContext`s.

However... I'm not sure this PR is a great win; it should reduce the memory for a very large number of connections - but that scenerio will probably have all sorts of other stuff to look at.

Happy to close for now and come back to; if it is an issue?

benaadams:
Closing this; can revisit later

PR #472, opened 2015-12-13
[Proposal] Reduce contention

benaadams:
Seeing "only" a 5% gain on top for this (though that is +70k rps)

Before
```
Running 30s test @ http://.../plaintext
    Latency     8.68ms   14.73ms 333.56ms   95.18%
    Req/Sec    47.81k     4.58k  122.10k    90.28%
  44317883 requests in 30.10s, 5.45GB read
Requests/sec: 1472388.96
```
After
```
Running 30s test @ http://.../plaintext
    Latency     9.18ms   18.09ms 305.19ms   95.63%
    Req/Sec    48.55k     8.78k  290.43k    91.02%
  46410810 requests in 30.10s, 5.71GB read
Requests/sec: 1541996.54
```
Gets a little exotic with `WeakReference<Thread>` so it can free pinned memory from the `ThreadStatic` MemoryPools as the threadpool scales back down without using finalizers.

### DIFF
--- a/src/Microsoft.AspNet.Server.Kestrel/Http/SocketInput.cs
+++ b/src/Microsoft.AspNet.Server.Kestrel/Http/SocketInput.cs
@@ -5,7 +5,6 @@ using System;
 using System.IO;
 using System.Runtime.CompilerServices;
 using System.Threading;
-using System.Threading.Tasks;
 using Microsoft.AspNet.Server.Kestrel.Infrastructure;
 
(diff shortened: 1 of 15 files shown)


benaadams:
On further testing, that increase could be within measurable error margins... O_o

benaadams:
Need to re-benchmark

benaadams:
meh

PR #473, opened 2015-12-13
Output thread except to console in testing

benaadams:
Without it errors in thrown in libuv loop (in test) either throw `NotImplementedException` from `LifetimeNotImplemented` or worse only show up as `System.Threading.Tasks.TaskCanceledException : A task was canceled.` on the "client" side.

Which doesn't help narrow down causes.


benaadams:
Meh.

Issue #484, opened 2015-12-14
Add ExecutionTimeout parameter


benaadams:
Timeout between two successive body reads which if exceeded the client is determined to be disconnected and the request aborted.

Failure state: `408 Request Timeout`

**ngnix** [default](http://nginx.org/en/docs/http/ngx_http_core_module.html#client_body_timeout) `client_body_timeout` 60s

Implementation-wise the timer can use the same timer as [Keep-alive](https://github.com/aspnet/KestrelHttpServer/issues/464); as a multi stage/state timer.

When deployed in recommended configuration with a reverse proxy (e.g. IIS/nginx) this would be handled by the proxy - however, this wouldn't apply in a pure Kestrel deployment.

Tratcher:
Make sure this doesn't break websockets.

benaadams:
@Tratcher I don't think so; but https isn't an upgrade request is it?

Tratcher:
>https isn't an upgrade request is it?

Huh?

benaadams:
> Huh?

My plan is to switch off the timers on opaque stream upgrade and assume they were handling that interaction (would also need to be same for MaxUploadBytes) - just making sure there is no corner cases I hadn't thought of (https in this case)

Tratcher:
Ah, sounds good.

https shouldn't be any different from http in this scenario.

benaadams:
Would be easier to have this as a maxRequestExecutionTimeout, that includes body reads, that work?

/cc @blowdart

benaadams:
Hmmm... may cause issue with large file uploads :-/

blowdart:
Hurrah, timeouts :D

DamianEdwards:
post RC-2

Issue #483, opened 2015-12-14
Add HeaderCompleteTimeout parameter


benaadams:
Time in which from the header preamble starts it must be completed. If exceeded the client is determined to be disconnected and the request aborted.

Failure state: `408 Request Timeout`

*"The client did not produce a request within the time that the server was prepared to wait. The client MAY repeat the request without modifications at any later time."*

**ngnix** [default](http://nginx.org/en/docs/http/ngx_http_core_module.html#client_header_timeout) `client_header_timeout` 60s

Implementation-wise the timer can use the same timer as [Keep-alive](https://github.com/aspnet/KestrelHttpServer/issues/464); as a multi stage/state timer.

When deployed in recommended configuration with a reverse proxy (e.g. IIS/nginx) this would be handled by the proxy - however, this wouldn't apply in a pure Kestrel deployment.

benaadams:
@halter73 @DamianEdwards this still RC2? Will finish in next couple days if so.

DamianEdwards:
We'll be in beginning of our ASK mode by then, which is very high bar. Late features like this don't meet the bar at this stage I'm afraid.

benaadams:
Is ok, is just marked RC2 still rather than backlog

benaadams:
Is ok - is handled by reverse poxies in suggested deployment, is just marked RC2 still rather than backlog

benaadams:
Also its not a very simple change, so I wouldn't put it in RC2 at this time :)

muratg:
This will be covered

Issue #464, opened 2015-12-09
Add keep-alive timeout parameter


benaadams:

muratg:
Yeah, I think we should do this.  

cc @lodejard

muratg:
@benaadams It looks like you're working on this.  Looking forward to see the PR!

cesarbs:
@benaadams Are you working on this? I guess not since it's been 21 days since @muratg's comment :smile: I'll take this if no one's working on it.

benaadams:
@cesarbs I was quite close with #485 though it needs rebasing; I can finish it of with improves if you can workout  #569? Haven't managed to work out what's going on there.

cesarbs:
@benaadams I can't look into #569 immediately, so go ahead with #485 as it is right now.

DamianEdwards:
post RC2

darrelmiller:
It would be nice if this could be called "ConnectionTimeout" instead of Keep-Alive timeout.  IIS calls it [ConnectionTimeout](http://www.iis.net/configreference/system.applicationhost/sites/sitedefaults/limits).  By calling it Keep-Alive timeout there is an inference that is somehow related to the HTTP/1.0 extension `keep-alive` header and the [failed attempt to resurrect it for a different purpose.](https://tools.ietf.org/id/draft-thomson-hybi-http-timeout-01.html#rfc.section.2.1). 

Admittedly though, Apache does call it [KeepAliveTimeout](http://httpd.apache.org/docs/2.2/mod/core.html)

cesarbs:
@darrelmiller It might happen https://github.com/aspnet/KestrelHttpServer/pull/1102 :smile:

PR #485, opened 2015-12-14
[Needs Tests] Add timeouts

benaadams:
For #464, #483, #484

Still needs
- [X] KeepAliveTimeout
- [X] HeadersCompleteTimeout
- [X] ExecutionTimeout
- [X] Read from config
- [X] Handle bad config
- [X] Switching off on Upgrades (e.g. WebSockets)
- [X] Cancelled/Aborted Exception (Easier to distinguish connection close events over more troubling errors)
- [ ] Tests

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Kestrel/IKestrelServerInformation.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/IKestrelServerInformation.cs
@@ -1,12 +1,19 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System;
 using Microsoft.AspNetCore.Server.Kestrel.Filter;
 
 namespace Microsoft.AspNetCore.Server.Kestrel
 {
     public interface IKestrelServerInformation
     {
+        TimeSpan ExecutionTimeout { get; set; }
+
+        TimeSpan HeadersCompleteTimeout { get; set; }
+
+        TimeSpan KeepAliveTimeout { get; set; }
+
         int ThreadCount { get; set; }
 
         bool NoDelay { get; set; }
--- a/src/Microsoft.AspNetCore.Server.Kestrel/KestrelServerInformation.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/KestrelServerInformation.cs
@@ -23,9 +23,18 @@ namespace Microsoft.AspNetCore.Server.Kestrel
             ThreadCount = GetThreadCount(configuration);
             NoDelay = GetNoDelay(configuration);
             ReuseStreams = GetReuseStreams(configuration);
+            HeadersCompleteTimeout = GetTimeout(configuration, "kestrel.headersCompleteTimeout", defaultSeconds: 30);
+            ExecutionTimeout = GetTimeout(configuration, "server.executionTimeout", defaultSeconds: 110);
+            KeepAliveTimeout = GetTimeout(configuration, "server.keepAliveTimeout", defaultSeconds: 160);
         }
 
         public ICollection<string> Addresses { get; }
+
+        public TimeSpan ExecutionTimeout { get; set; }
+
+        public TimeSpan HeadersCompleteTimeout { get; set; }
+
+        public TimeSpan KeepAliveTimeout { get; set; }
 
         public int ThreadCount { get; set; }
 
@@ -122,6 +131,24 @@ namespace Microsoft.AspNetCore.Server.Kestrel
             }
 
             return false;
+        }
+
+        private static TimeSpan GetTimeout(IConfiguration configuration, string configurationKey, int defaultSeconds)
+        {
+            var timeoutString = configuration[configurationKey];
+
+            if (string.IsNullOrEmpty(timeoutString))
+            {
+                return TimeSpan.FromSeconds(defaultSeconds);
+            }
+
+            TimeSpan timeout;
+            if (TimeSpan.TryParse(timeoutString, out timeout))
+            {
+                return timeout;
+            }
+
+            return TimeSpan.FromSeconds(defaultSeconds);
         }
     }
 }
--- a/src/Microsoft.AspNetCore.Server.Kestrel/Http/Frame.FeatureCollection.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/Http/Frame.FeatureCollection.cs
@@ -11,7 +11,6 @@ using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.Http.Features;
-using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Primitives;
 
 namespace Microsoft.AspNetCore.Server.Kestrel.Http
@@ -314,7 +313,11 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Http
 
             await FlushAsync(default(CancellationToken));
 
-            return DuplexStream;
+            if (_frameState.TransitionToState(RequestState.UpgradedRequest) == RequestState.UpgradedRequest)
+            {
+                return DuplexStream;
+            }
+            throw new IOException("Failed to upgrade request");
         }
 
         IEnumerator<KeyValuePair<Type, object>> IEnumerable<KeyValuePair<Type, object>>.GetEnumerator() => FastEnumerable().GetEnumerator();
--- a/src/Microsoft.AspNetCore.Server.Kestrel/Http/Frame.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/Http/Frame.cs
@@ -53,10 +53,8 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Http
 
         protected List<KeyValuePair<Func<object, Task>, object>> _onCompleted;
 
-        private bool _requestProcessingStarted;
+        protected FrameState _frameState;
         private Task _requestProcessingTask;
-        protected volatile bool _requestProcessingStopping; // volatile, see: https://msdn.microsoft.com/en-us/library/x13ttww7.aspx
-        protected int _requestAborted;
         protected CancellationTokenSource _abortedCts;
         protected CancellationToken? _manuallySetRequestAbortToken;
 
@@ -92,12 +90,14 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Http
             _localEndPoint = localEndPoint;
             _prepareRequest = prepareRequest;
             _pathBase = context.ServerAddress.PathBase;
-            if (ReuseStreams)
+            if (Settings.ReuseStreams)
             {
                 _requestBody = new FrameRequestStream();
                 _responseBody = new FrameResponseStream(this);
                 _duplexStream = new FrameDuplexStream(_requestBody, _responseBody);
             }
+
+            _frameState = new FrameState(this, context.Settings);
 
             FrameControl = this;
             Reset();
@@ -167,7 +167,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Http
                 var cts = _abortedCts;
                 return
                     cts != null ? cts.Token :
-                    (Volatile.Read(ref _requestAborted) == 1) ? new CancellationToken(true) :
+                    _frameState.CurrentState == RequestState.Aborted ? new CancellationToken(true) :
                     RequestAbortedSource.Token;
             }
             set
@@ -185,20 +185,26 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Http
                 // Get the abort token, lazily-initializing it if necessary.
                 // Make sure it's canceled if an abort request already came in.
                 var cts = LazyInitializer.EnsureInitialized(ref _abortedCts, () => new CancellationTokenSource());
-                if (Volatile.Read(ref _requestAborted) == 1)
+                if (_frameState.CurrentState == RequestState.Aborted)
                 {
                     cts.Cancel();
                 }
                 return cts;
             }
         }
+
         public bool HasResponseStarted
         {
             get { return _responseStarted; }
         }
 
-        public void Reset()
+        public bool Reset()
         {
+            if (_frameState.TransitionToState(RequestState.Waiting) != RequestState.Waiting)
+            {
+                return false;
+            }
+
             _onStarting = null;
             _onCompleted = null;
 
@@ -237,6 +243,8 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Http
 
             _manuallySetRequestAbortToken = null;
             _abortedCts = null;
+
+            return true;
         }
 
         public void ResetResponseHeaders()
@@ -255,9 +263,8 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Http
         /// </summary>
         public void Start()
         {
-            if (!_requestProcessingStarted)
+            if (_frameState.TransitionToState(RequestState.Waiting) == RequestState.Waiting)
             {
-                _requestProcessingStarted = true;
                 _requestProcessingTask =
                     Task.Factory.StartNew(
                         (o) => ((Frame)o).RequestProcessingAsync(),
@@ -276,10 +283,8 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Http
         /// </summary>
         public Task Stop()
         {
-            if (!_requestProcessingStopping)
-            {
-                _requestProcessingStopping = true;
-            }
+            _frameState.TransitionToState(RequestState.Stopped);
+
             return _requestProcessingTask ?? TaskUtilities.CompletedTask;
         }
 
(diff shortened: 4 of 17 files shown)


src/Microsoft.AspNet.Server.Kestrel/Http/Frame.cs
```diff
@@ -185,8 +196,91 @@ public bool HasResponseStarted
...
+        private void TimeoutRequest()
+        {
+            if (TransitionToState(FrameState.Timeout) == FrameState.Timeout)
+            {
+                SocketInput?.AbortAwaiting();
+            }
+        }
+        protected int TransitionToState(int state)
+        {
+#pragma warning disable CS0420 // A reference to a volatile field will not be treated as volatile
```
muratg: Isn't disabling this potentially dangerous?

benaadams: Is a spurious compile warning though perhaps I can just go 
`int prevState = _frameState;` 
rather than 
`int prevState = Volatile.Read(ref _frameState);`
and not get the error

muratg: Yeah, my worry is that passing volatile fields as ref (or out) may have different behavior on different CPU architectures.

benaadams: Now you raise is... changed to normal `int` and changed all reads to `Volatile.Read` as while `volatile` gives compiler guarantees which is enough on Intel; it doesn't give CPU guarantees which may nor be enough on arm. (Don't know how the traditional C# memory memory model works on other chipssets). Also I'm using the value with `Interlocked` which wants `ref`

Also I've rolled two flags into one state and added some extras; so safety first.

src/Microsoft.AspNet.Server.Kestrel/Http/Frame.cs
```diff
@@ -39,6 +39,13 @@ public abstract partial class Frame : FrameContext, IFrameControl
         private static readonly byte[] _bytesDate = Encoding.ASCII.GetBytes("Date: ");
         private static readonly byte[] _bytesEndHeaders = Encoding.ASCII.GetBytes("\r\n\r\n");
 
+        private static readonly TimerCallback _timeoutRequest = (o) => ((Frame)o).TimeoutRequest();
+        protected readonly Timer _timeout;
+        /* TODO: Need to move to config */
```
Tratcher: And make them TimeSpans

benaadams:
Added keep-alive as: **`server`**`.keepAliveTimeout`
Other two as **`kestrel`**`.headersCompleteTimeout` and **`kestrel`**`.bodyKeepAliveTimeout`
That work?

src/Microsoft.AspNet.Server.Kestrel/Http/Frame.cs
```diff
@@ -185,8 +190,89 @@ public bool HasResponseStarted
             get { return _responseStarted; }
         }
 
+        private void TimeoutRequest()
+        {
+            if (TransitionToState(FrameState.Timeout) == FrameState.Timeout)
+            {
+                SocketInput?.AbortAwaiting();
+            }
+        }
```
cesarbs: nit: add newline

src/Microsoft.AspNet.Server.Kestrel/KestrelServerInformation.cs
```diff
@@ -107,5 +117,59 @@ private static bool GetNoDelay(IConfiguration configuration)
...
+            }
+
+            TimeSpan keepAliveTimeout;
+            if (TimeSpan.TryParse(bodyKeepAliveTimeoutString, out keepAliveTimeout))
+            {
+                return keepAliveTimeout;
+            }
+
+            return TimeSpan.FromSeconds(60);
+        }
```
cesarbs: The three methods above are extremely similar to each other. Refactor into a single method with parameters for each timeout?

benaadams:
Just need to add tests (I think)

cesarbs:
Yes, please :smile:

benaadams:
Still needs more work in light of https://github.com/aspnet/KestrelHttpServer/issues/566#issuecomment-171009932

benaadams:
Should have it finished tomorrow.

src/Microsoft.AspNet.Server.Kestrel/Http/SocketOutput.cs
```diff
@@ -388,6 +387,22 @@ void ISocketOutput.Write(ArraySegment<byte> buffer, bool immediate, bool chunk)
 
         Task ISocketOutput.WriteAsync(ArraySegment<byte> buffer, bool immediate, bool chunk, CancellationToken cancellationToken)
         {
+            if (cancellationToken.IsCancellationRequested)
```
halter73: We want to register with the cancellation token for writes that actually result in an incomplete task being returned.

src/Microsoft.AspNet.Server.Kestrel/Infrastructure/KestrelTrace.cs
```diff
@@ -39,6 +40,7 @@ static KestrelTrace()
             // ConnectionWrite: Reserved: 11
             // ConnectionWriteCallback: Reserved: 12
             // ApplicationError: Reserved: 13 - LoggerMessage.Define overload not present 
+            _connectionDisconnectedWrite = LoggerMessage.Define<long, int>(LogLevel.Debug, 14, @"Connection id ""{ConnectionId}"" write of ""{count}"" bytes to disconnected client.");
         }
```
halter73: We also want to log errors surfaced by libuv to the write callback since we will no longer throw `IOExceptions` from `SocketOutput.WriteAsync` (though we really should have done this all along since there was never a guarantee that the call to `WriteAsync` would actually see the error before). We decided that these errors should use `LogLevel.Information`.

src/Microsoft.AspNet.Server.Kestrel/Http/FrameRequestStream.cs
```diff
@@ -165,7 +165,26 @@ private void ValidateState()
...
+            }
+        }
+
+        private void ValidateStateAsync(CancellationToken cancellationToken)
+        {
+            if (_state == FrameStreamState.Closed)
+            {
+                throw new ObjectDisposedException(nameof(FrameRequestStream));
+            }
+            if (cancellationToken.IsCancellationRequested)
```
halter73: Is there a reason this pattern is better than `cancellationToken.ThrowIfCancellationRequested()`?

src/Microsoft.AspNet.Server.Kestrel/Http/FrameRequestStream.cs
```diff
@@ -165,7 +165,26 @@ private void ValidateState()
                 case FrameStreamState.Closed:
                     throw new ObjectDisposedException(nameof(FrameRequestStream));
                 case FrameStreamState.Aborted:
-                    throw new IOException("The request has been aborted.");
+                    throw new OperationCanceledException("The request has been aborted.");
```
halter73: Is it wrong to throw a `TaskCanceledException` from a synchronous call to Read or Write? It is an `OperationCanceledException` so the handling code should be the same. I think it's best to have one ValidateState method if possible.

src/Microsoft.AspNet.Server.Kestrel/Http/Frame.StateMachine.cs
```diff
@@ -0,0 +1,143 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Diagnostics;
+using System.Threading;
+
+namespace Microsoft.AspNet.Server.Kestrel.Http
+{
+    public partial class Frame
```
halter73: How hard would it be to move this logic into its own class? I really like having the state machine logic in its own file, but its own class would be even better.

src/Microsoft.AspNet.Server.Kestrel/Http/Frame.StateMachine.cs
```diff
@@ -0,0 +1,143 @@
...
+                    } while (prevState < FrameState.Timeout);
+                    return prevState < FrameState.Timeout ? FrameState.Timeout : prevState;
+                case FrameState.Stopped:
+                    if (prevState >= FrameState.Stopped) return prevState;
+                    // can transition to Stopped from states below it
+                    do
+                    {
+                        prevState = Interlocked.CompareExchange(ref _frameState, FrameState.Stopped, prevState);
+                    } while (prevState < FrameState.Stopped);
+                    return prevState < FrameState.Stopped ? FrameState.Stopped : prevState;
```
halter73: Isn't it impossible for `prevState < FrameState.Stopped` here?

src/Microsoft.AspNet.Server.Kestrel/Http/Frame.cs
```diff
@@ -297,24 +302,28 @@ public Task Stop()
...
 
             _requestBody?.Abort();
             _responseBody?.Abort();
 
             try
             {
-                ConnectionControl.End(ProduceEndType.SocketDisconnect);
-                SocketInput.AbortAwaiting();
-                RequestAbortedSource.Cancel();
+                ConnectionControl?.End(ProduceEndType.SocketDisconnect);
```
halter73: Is `ConnectionControl` ever set to null? Same question for the other additions of `?.`

benaadams: Tests were throwing when I had the timeouts low; probably should fix the tests instead

src/Microsoft.AspNet.Server.Kestrel/Http/FrameRequestStream.cs
benaadams: No, but will throw `OperationCanceledException` cancelling a task or tcs with throw `TaskCanceledException`; I assume to differentiate whether it's throwing on sync part or async part. Standardize on `TaskCanceledException`?

src/Microsoft.AspNet.Server.Kestrel/Http/FrameRequestStream.cs
benaadams: After having crazy exception handling/checking for Websockets in 4.5+ (I think SignalR had similar) - I'm all up for a standardised exception :)

src/Microsoft.AspNet.Server.Kestrel/Infrastructure/KestrelTrace.cs
benaadams: Need some more clarity around the last parameter `Exception` parameter to logging as it does seem to come out. Adding one that takes an exception parameter.

src/Microsoft.AspNet.Server.Kestrel/Http/SocketOutput.cs
benaadams: Done; not sure perf effects; but on the plus side I've removed some boxing and unboxing :stuck_out_tongue_winking_eye:

benaadams:
Checkpoint

src/Microsoft.AspNet.Server.Kestrel/Http/SocketInput.cs
```diff
@@ -238,6 +239,14 @@ public void GetResult()
             var error = _awaitableError;
             if (error != null)
             {
+                if (error is OperationCanceledException)
```
benaadams: Wrong error handling

benaadams:
Question on how to handle error in `ProducingStart`

halter73:
@benaadams I think we return the default iterator from `ProducingStart` if there has been an error, and special case the `CopyFrom` methods to no-op.

src/Microsoft.AspNet.Server.Kestrel/Http/Frame.StateMachine.cs
benaadams: I think this could work

src/Microsoft.AspNet.Server.Kestrel/Http/SocketInput.cs
```diff
@@ -238,6 +239,14 @@ public void GetResult()
             var error = _awaitableError;
             if (error != null)
             {
+                if (error is TaskCanceledException)
```
halter73: Maybe this could be:

    if (error is InvalidOperationException || error is InvalidOperationException)
    {
       throw error;
    }

benaadams:
After discussion with @halter73 I've reopened #567 and moved the Exception/Cancellation code back there - will finish that first and then continue with the timeouts.

benaadams:
This will have to wait till after #567 as otherwise it will be a horror or a mergeclash

benaadams:
Rebased

benaadams:
build stalls when merged with the dotnet update - investigating

cesarbs:
:disappointed: Try setting `NO_PARALLEL_TEST_PROJECTS=Microsoft.AspNetCore.Server.Kestrel.FunctionalTests` (env. variable).

benaadams:
Having a little look as it works without the merge; trying just upping threads e.g.
`set ComPlus_ThreadPool_ForceMinWorkerThreads=2000`

src/Microsoft.AspNetCore.Server.Kestrel/KestrelServerInformation.cs
```diff
@@ -123,5 +132,23 @@ private static bool GetReuseStreams(IConfiguration configuration)
 
             return false;
         }
+
+        private static TimeSpan GetTimeout(IConfiguration configuration, string configurationKey, int defaultSeconds)
+        {
+            var timeoutString = configuration[configurationKey];
```
muratg: Why not pass timeoutString directly to this method?

benaadams:
Thread count didn't help; got a failed restore second time; third time (without `MinWorkerThreads` or `NO_PARALLEL_TEST_PROJECTS`) it was happy...

Might be a dotnet cli issue instead

cesarbs:
@benaadams Yeah, we think it might be dotnet, or the xunit console runner is doing something funny.

cesarbs:
@benaadams Did you see it hang consistently? I've only seen it hang on AppVeyor, can't repro on my box.

benaadams:
No; but also get on/off restore issues with `Build` it seems a little shaky atm compared to the dnx version

benaadams:
No; but also get on/off restore issues with `Build` when starting with a fresh repo; it seems a little shaky atm compared to the dnx version

(e.g. delete directory; then do fresh get from github)

halter73:
We should be able to add this for rc2. Do you still think you can write some test cases for this? Thanks!

benaadams:
I'd like advice on test cases; waiting for timeouts and CI never seem to mix well...

benaadams:
To clarify further; the timeouts are in seconds so can't go finer than that; and CI is flakey with timeout numbers ;-/

Write write some see how it goes

halter73:
I would make timeouts very short (maybe a tenth of a second) and verify that you get 500 responses (header timeout) or the that connection is gracefully (keepalive timeout)/ungracefully(execution timeout). Also verify that ongoing reads get aborted in the event you are in the executing phase.

You can be very generous with the amount of time you wait for the timeouts to trigger. I wouldn't mind if we waited 10 seconds (100 times what they are set to) for the timeouts to trigger. Hopefully that will be long enough for travis/appveyor. If not, we can adjust.

halter73:
Oh. And this isn't urgent. Feel free to wait until we get AppVeyor and Travis passing again (maybe someday :disappointed:) so you can verify the timeouts work.

benaadams:
Obviously I need to get it in before #623 - think of the merge! ;-)

src/Microsoft.AspNetCore.Server.Kestrel/IKestrelServerInformation.cs
```diff
@@ -1,12 +1,19 @@
...
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System;
 using Microsoft.AspNetCore.Server.Kestrel.Filter;
 
 namespace Microsoft.AspNetCore.Server.Kestrel
 {
     public interface IKestrelServerInformation
     {
+        TimeSpan ExecutionTimeout { get; set; }
```
Tratcher: KeepAlive and HeadersComplete are reasonable, but I don't think we need or want ExecutionTimeout at the server level.
A) The same functionality can be provided in app/middleware code.
B) The timeout is un-enforceable. You can't force an app to abandon a request, you can only close the socket and trip the token. The RequestAborted CancellationToken is at best a hint. There are no thread aborts like v4.
C) There are known scenarios this breaks with long running requests. It's one more setting on the server to hunt down and fiddle with when that happens.

benaadams: :ok:

benaadams:
@muratg @halter73 @cesarbs going to close this as realistically I won't have enough time to devote to it; the underlying code has changed a lot and there are multiple potential race conditions which have to be guarded against

MelbourneDeveloper:
I'm pretty confused. Is there currently an ExecutionTimeout in Kestrel? If so, how to we stop it, or change it?

I'm getting a WCF timeout that I can't explain unless it's something going on with a Kestrel request timeout.

davidfowl:
There's no execution timeout in kestrel (which isn't possible to do with async requests because if there's no thread running then there's nothing to time out). How would kestrel affect your WCF service? The ASP.NET core module has a 2 minute timeout for requests, maybe you're running into that?

PR #487, opened 2015-12-15
Amortize stream costs

### LINKED ISSUE #471: Can Middleware modify Response In OnCompleted callback?
Looking at [FrameOfT#98](https://github.com/aspnet/KestrelHttpServer/blob/dev/src/Microsoft.AspNet.Server.Kestrel/Http/FrameOfT.cs#L98).

It doesn't seem like there's anything to prevent Middleware from modifying the response in the `OnCompleted` callback.

Does `OnCompleted` mean the application response has been received by the server, and so Middleware can add extra stuff to the response stream (diagnostics for instance) 
 or does it mean the response has been fully received and has been sent (or is being sent) to the client, so Middleware can clean up any resources attached to this request?

benaadams:
Reuse streams over connection lifetime.

Reduction in 12,000 objects per 4k requests; and 640000 bytes

Also resolves #471 "Can Middleware modify Response In OnCompleted callback" /cc @cesarbs

### DIFF
--- a/src/Microsoft.AspNet.Server.Kestrel/Http/Frame.cs
+++ b/src/Microsoft.AspNet.Server.Kestrel/Http/Frame.cs
@@ -57,6 +57,7 @@ namespace Microsoft.AspNet.Server.Kestrel.Http
 
         internal FrameRequestStream _requestBody;
         internal FrameResponseStream _responseBody;
+        internal FrameDuplexStream _duplexStream;
 
         protected bool _responseStarted;
         protected bool _keepAlive;
@@ -86,6 +87,9 @@ namespace Microsoft.AspNet.Server.Kestrel.Http
             _localEndPoint = localEndPoint;
             _prepareRequest = prepareRequest;
             _pathBase = context.ServerAddress.PathBase;
+            _requestBody = new FrameRequestStream();
+            _responseBody = new FrameResponseStream(this);
+            _duplexStream = new FrameDuplexStream(_requestBody, _responseBody);
 
             FrameControl = this;
             Reset();
--- a/src/Microsoft.AspNet.Server.Kestrel/Http/Frame.cs
+++ b/src/Microsoft.AspNet.Server.Kestrel/Http/Frame.cs
@@ -87,9 +87,12 @@ namespace Microsoft.AspNet.Server.Kestrel.Http
             _localEndPoint = localEndPoint;
             _prepareRequest = prepareRequest;
             _pathBase = context.ServerAddress.PathBase;
-            _requestBody = new FrameRequestStream();
-            _responseBody = new FrameResponseStream(this);
-            _duplexStream = new FrameDuplexStream(_requestBody, _responseBody);
+            if (ReuseStreams)
+            {
+                _requestBody = new FrameRequestStream();
+                _responseBody = new FrameResponseStream(this);
+                _duplexStream = new FrameDuplexStream(_requestBody, _responseBody);
+            }
 
             FrameControl = this;
             Reset();
--- a/src/Microsoft.AspNet.Server.Kestrel/Http/FrameOfT.cs
+++ b/src/Microsoft.AspNet.Server.Kestrel/Http/FrameOfT.cs
@@ -91,6 +91,9 @@ namespace Microsoft.AspNet.Server.Kestrel.Http
                                 await FireOnStarting();
                             }
 
+                            _requestBody.PauseAcceptingReads();
+                            _responseBody.PauseAcceptingWrites();
+
                             if (_onCompleted != null)
                             {
                                 await FireOnCompleted();
@@ -101,10 +104,12 @@ namespace Microsoft.AspNet.Server.Kestrel.Http
                             // If _requestAbort is set, the connection has already been closed.
                             if (!_requestAborted)
                             {
+                                _responseBody.ResumeAcceptingWrites();
                                 await ProduceEnd();
 
                                 if (_keepAlive)
                                 {
+                                    _requestBody.ResumeAcceptingReads();
                                     // Finish reading the request body in case the app did not.
                                     await messageBody.Consume();
                                 }
--- a/src/Microsoft.AspNet.Server.Kestrel/Http/FrameOfT.cs
+++ b/src/Microsoft.AspNet.Server.Kestrel/Http/FrameOfT.cs
@@ -64,6 +64,15 @@ namespace Microsoft.AspNet.Server.Kestrel.Http
                     {
                         var messageBody = MessageBody.For(HttpVersion, _requestHeaders, this);
                         _keepAlive = messageBody.RequestKeepAlive;
+
+                        // _duplexStream may be null if flag switched while running
+                        if (!ReuseStreams || _duplexStream == null)
+                        {
+                            _requestBody = new FrameRequestStream();
+                            _responseBody = new FrameResponseStream(this);
+                            _duplexStream = new FrameDuplexStream(_requestBody, _responseBody);
+                        }
+
                         RequestBody = _requestBody.StartAcceptingReads(messageBody);
                         ResponseBody = _responseBody.StartAcceptingWrites();
                         DuplexStream = _duplexStream;
--- a/src/Microsoft.AspNet.Server.Kestrel/Http/FrameOfT.cs
+++ b/src/Microsoft.AspNet.Server.Kestrel/Http/FrameOfT.cs
@@ -64,11 +64,9 @@ namespace Microsoft.AspNet.Server.Kestrel.Http
                     {
                         var messageBody = MessageBody.For(HttpVersion, _requestHeaders, this);
                         _keepAlive = messageBody.RequestKeepAlive;
-                        _requestBody = new FrameRequestStream(messageBody);
-                        RequestBody = _requestBody;
-                        _responseBody = new FrameResponseStream(this);
-                        ResponseBody = _responseBody;
-                        DuplexStream = new FrameDuplexStream(RequestBody, ResponseBody);
+                        RequestBody = _requestBody.StartAcceptingReads(messageBody);
+                        ResponseBody = _responseBody.StartAcceptingWrites();
+                        DuplexStream = _duplexStream;
 
                         _abortedCts = null;
                         _manuallySetRequestAbortToken = null;
--- a/src/Microsoft.AspNet.Server.Kestrel/Http/FrameRequestStream.cs
+++ b/src/Microsoft.AspNet.Server.Kestrel/Http/FrameRequestStream.cs
@@ -123,6 +123,19 @@ namespace Microsoft.AspNet.Server.Kestrel.Http
             return this;
         }
 
+        public void PauseAcceptingReads()
+        {
+            _state = StreamState.Closed;
+        }
+
+        public void ResumeAcceptingReads()
+        {
+            if (_state == StreamState.Closed)
+            {
+                _state = StreamState.Open;
+            }
+        }
+
         public void StopAcceptingReads()
         {
             // Can't use dispose (or close) as can be disposed too early by user code
--- a/src/Microsoft.AspNet.Server.Kestrel/Http/FrameRequestStream.cs
+++ b/src/Microsoft.AspNet.Server.Kestrel/Http/FrameRequestStream.cs
@@ -11,11 +11,11 @@ namespace Microsoft.AspNet.Server.Kestrel.Http
     public class FrameRequestStream : Stream
     {
         private MessageBody _body;
-        private StreamState _state;
+        private FrameStreamState _state;
 
         public FrameRequestStream()
         {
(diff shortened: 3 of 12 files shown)


benaadams:
Byte allocs

before

after

Allocations on `(!_requestProcessingStopping)` line are for:
`var messageBody = MessageBody.For(HttpVersion, _requestHeaders, this);`

benaadams:
Will add a flag for this...

src/Microsoft.AspNet.Server.Kestrel/Http/FrameOfT.cs
```diff
@@ -64,11 +64,15 @@ public class Frame<TContext> : Frame
                     {
                         var messageBody = MessageBody.For(HttpVersion, _requestHeaders, this);
                         _keepAlive = messageBody.RequestKeepAlive;
-                        _requestBody = new FrameRequestStream(messageBody);
-                        RequestBody = _requestBody;
-                        _responseBody = new FrameResponseStream(this);
-                        ResponseBody = _responseBody;
-                        DuplexStream = new FrameDuplexStream(RequestBody, ResponseBody);
+                        if (!ReuseStreams)
```
cesarbs: nit: space before if block

src/Microsoft.AspNet.Server.Kestrel/Http/FrameOfT.cs
```diff
@@ -64,11 +64,15 @@ public class Frame<TContext> : Frame
...
+                        {
+                            _requestBody = new FrameRequestStream();
+                            _responseBody = new FrameResponseStream(this);
+                            _duplexStream = new FrameDuplexStream(_requestBody, _responseBody);
+                        }
```
cesarbs: nit: new line after if block

src/Microsoft.AspNet.Server.Kestrel/Http/FrameOfT.cs
```diff
@@ -93,6 +97,9 @@ public class Frame<TContext> : Frame
                                 await FireOnStarting();
                             }
 
+                            _requestBody.PauseAcceptingReads();
+                            _responseBody.PauseAcceptingWrites();
+
                             if (_onCompleted != null)
                             {
                                 await FireOnCompleted();
```
cesarbs: Could this throw?

src/Microsoft.AspNet.Server.Kestrel/Http/FrameRequestStream.cs
```diff
@@ -135,7 +160,7 @@ private void ValidateState()
             {
                 case StreamState.Open:
                     return;
-                case StreamState.Disposed:
+                case StreamState.Closed:
                     throw new ObjectDisposedException(nameof(FrameRequestStream));
```
cesarbs: A few thoughts here:

1) `ObjectDisposedException` doesn't seem like the right exception to throw here since the stream can switch between the `Open` and `Closed` states multiple times.
2) Should we really throw? How about just no-oping when someone attempts to operate on a closed stream?

src/Microsoft.AspNet.Server.Kestrel/Http/FrameResponseStream.cs
```diff
@@ -100,7 +125,7 @@ private void ValidateState()
             {
                 case StreamState.Open:
                     return;
-                case StreamState.Disposed:
+                case StreamState.Closed:
                     throw new ObjectDisposedException(nameof(FrameResponseStream));
```
cesarbs: Same comments from above apply here.

src/Microsoft.AspNet.Server.Kestrel/Http/FrameResponseStream.cs
```diff
@@ -110,7 +135,7 @@ private void ValidateState()
         private enum StreamState
```
cesarbs: This enum is duplicate.

src/Microsoft.AspNet.Server.Kestrel/KestrelServerInformation.cs
```diff
@@ -107,5 +110,23 @@ private static bool GetNoDelay(IConfiguration configuration)
...
+        {
+            var reuseStreamsString = configuration["kestrel.reuseStreams"];
+
+            if (string.IsNullOrEmpty(reuseStreamsString))
+            {
+                return false;
+            }
+
+            bool reuseStreams;
+            if (bool.TryParse(reuseStreamsString, out reuseStreams))
```
cesarbs: Does `TryParse` throw on `null`? If not you could eliminate the previous `if` block.

test/Microsoft.AspNet.Server.Kestrel.FunctionalTests/ReuseStreamsTests.cs
```diff
@@ -0,0 +1,108 @@
...
+                {
+                    return context.Request.Body.CopyToAsync(context.Response.Body);
+                });
+            });            
+
+            using (var app = hostBuilder.Build().Start())
+            {
+                using (var client = new HttpClient())
+                {
+
```
cesarbs: nit: remove empty line

test/Microsoft.AspNet.Server.Kestrel.FunctionalTests/ReuseStreamsTests.cs
```diff
@@ -0,0 +1,108 @@
...
+using Microsoft.AspNet.Http.Features;
+using Microsoft.Extensions.Configuration;
+using Xunit;
+
+namespace Microsoft.AspNet.Server.Kestrel.FunctionalTests
+{
+    public class ReuseStreamsTests
+    {
+        [Fact]
+        public async Task ReuseStreamsOn()
```
cesarbs: I'm sorry if I'm missing something obvious here, but I don't see how this test is effectively validating that the stream are being reused.

test/Microsoft.AspNet.Server.KestrelTests/KestrelServerInformationTests.cs
```diff
@@ -85,6 +85,38 @@ public void SetNoDelayUsingConfiguration()
             Assert.False(information.NoDelay);
         }
 
+
+        [Fact]
+        public void SetReuseStreamsUsingConfiguration()
```
cesarbs: Please split into separate tests for `"true"` and `"false"`. Also please add tests for `null`, `string.Empty` and an invalid string (i.e. not parsable as `bool`).

src/Microsoft.AspNet.Server.Kestrel/Http/FrameRequestStream.cs
benaadams: 1) To the user of the stream it should seem like its disposed; the implementation of it being reused shouldn't be exposed. Also in the non-reuse case it stays in the DIsposed state
2) Should throw as you are in the dangerzone writing to it in the reuse case.

test/Microsoft.AspNet.Server.Kestrel.FunctionalTests/ReuseStreamsTests.cs
benaadams: Its validating that it doesn't crash when the stream is reused/not reused (on the keep-alive connection); as the initialization and re-initialization paths are different.

src/Microsoft.AspNet.Server.Kestrel/Http/FrameOfT.cs
benaadams: Definitely. Concerned about the `DisposeContext` not getting called?

test/Microsoft.AspNet.Server.Kestrel.FunctionalTests/ReuseStreamsTests.cs
benaadams: Will add something to ensure resuse is and is not taking place

src/Microsoft.AspNet.Server.Kestrel/Http/FrameOfT.cs
benaadams: Oh, no it can't

src/Microsoft.AspNet.Server.Kestrel/Http/FrameRequestStream.cs
cesarbs: :+1:

src/Microsoft.AspNet.Server.Kestrel/Http/FrameOfT.cs
cesarbs: I'm concerned about leaving streams in the `Closed` state.

benaadams: If it throws, you move into the finally block and shutdown the connection and throw away the streams with it - they are only reused over the connection.

src/Microsoft.AspNet.Server.Kestrel/Http/FrameOfT.cs
benaadams: Done

test/Microsoft.AspNet.Server.Kestrel.FunctionalTests/ReuseStreamsTests.cs
benaadams: Is now

src/Microsoft.AspNet.Server.Kestrel/Http/FrameOfT.cs
cesarbs: Ok, nice.

test/Microsoft.AspNet.Server.Kestrel.FunctionalTests/ReuseStreamsTests.cs
```diff
@@ -0,0 +1,127 @@
...
+                        };
+                        request.Headers.Add("Connection", new string[] { "Keep-Alive" });
+                        var responseMessage = await client.SendAsync(request);
+                        var result = await responseMessage.Content.ReadAsStringAsync();
+                        Assert.Equal(content, result);
+                    }
+                }
+            }
+
+            Assert.True(streamCount == loopCount);
```
cesarbs: Assert.Equal

cesarbs:
LGTM, just address that `Assert.Equal` comment I've just left here.

I'll run some benchmarks on this.

Did you perform any real world tests with this change? No chance of stuff getting mixed up in the streams (i.e. something from one request spilling into the next)?

benaadams:
> Did you perform any real world tests with this change? 

Yes, though I'm around 1.5Mrps and hit a plateau (though can get 2M+ at pipeline 512) - my guess is lock contention/spinlock on libuv thread; so working that through.

> No chance of stuff getting mixed up in the streams (i.e. something from one request spilling into the next)?

No - however, if someone held onto a stream passed `oncomplete` and after the context had been "disposed" then tried to do something with it after the next request had started - they could write to the wrong stream (the correct stream being an thrown Exception) - however that would be a bad coding error.

But... people do dumb things; which is why I put it behind a flag that's off by default. So you need to actively switch it on and make that choice.

benaadams:
Though in effect its holding onto the socket rather than any state data; would have a similar issue if you held onto the `IFeatureCollection` and used that beyond the end of the request.

src/Microsoft.AspNet.Server.Kestrel/IKestrelServerInformation.cs
```diff
@@ -11,6 +11,8 @@ public interface IKestrelServerInformation
 
         bool NoDelay { get; set; }
 
+        bool ReuseStreams { get; set; }
```
DamianEdwards: Add an XML doc comment to describe what this option does please.

src/Microsoft.AspNet.Server.Kestrel/Http/Frame.cs
```diff
@@ -86,7 +87,12 @@ public Frame(ConnectionContext context)
             _localEndPoint = localEndPoint;
             _prepareRequest = prepareRequest;
             _pathBase = context.ServerAddress.PathBase;
-
+            if (ReuseStreams)
+            {
+                _requestBody = new FrameRequestStream();
+                _responseBody = new FrameResponseStream(this);
+                _duplexStream = new FrameDuplexStream(_requestBody, _responseBody);
+            }
```
halter73: nit:

src/Microsoft.AspNet.Server.Kestrel/IKestrelServerInformation.cs
benaadams: Added

src/Microsoft.AspNet.Server.Kestrel/Http/Frame.cs
benaadams: Added return

halter73:
:ship:

Issue #328, opened 2015-11-03
UvTcpHandle life not managed explicitly


migueldeicaza:
While investigating the peformance problems of Kestrel on Mono we identified a strange pause and a strange set of socket errors that would happen every once in a while.

The problem is caused because Kestrel depends on the GC finalization to release its hold on the UvTcpHandle instead of explicitly disposing the objects.

schani:
Specifically, the server will eventually become completely unresponsive after a while.  This can be exacerbated either by using a very large nursery (try `MONO_GC_PARAMS=nursery-size=1g`) or disabling finalization (`MONO_GC_DEBUG=do-not-finalize`).

halter73:
I just added the following to UvTcpHandle:

        ~UvTcpHandle()
        {
            System.Diagnostics.Debugger.Launch();
        }

And ran build.cmd which runs all our unit and functional tests on core and desktop clr. As far as I can tell, the finalizer never got hit.

This makes sense as a quick scan of the code seems to show a corresponding call to Dispose for every `new UvTcpHandle`.

Does anyone have any details that could help me figure out what conditions are necessary to prevent UvTcpHandles from being disposed?

schani:
Running the Kestrel server for this benchmark, on Mono: https://github.com/schani/benchmarks/tree/master/src/Benchmarks

I need to specify a small enough nursery, otherwise the GC never gets around to finalizing the objects.  I'm using

    MONO_GC_PARAMS=nursery-size=128k MONO_GC_DEBUG=log-finalizers

Then all I have to do is

    wrk -c 1024 -t 32 http://localhost:5001/plaintext

and I get lots of these:

    mono-gc-finalizers: <UvTcpHandle at 0x7fbd03fe0eb0> Starting finalizer checks.
    mono-gc-finalizers: <UvTcpHandle at 0x7fbd03fe0eb0> Registered finalizer as processed.
    mono-gc-finalizers: <UvTcpHandle at 0x7fbd03fe0eb0> Compiling finalizer.
    mono-gc-finalizers: <UvTcpHandle at 0x7fbd03fe0eb0> Calling finalizer.
    mono-gc-finalizers: <UvTcpHandle at 0x7fbd03fe0eb0> Returned from finalizer.

Running on 1.0.0-beta8, as per `project.json`.

benaadams:
[...]

PR #489, opened 2015-12-15
Dispose socket on Accept() error (#328).

cesarbs:
We're leaking sockets when there's an error on Accept() on an incoming connection. The socket is only disposed when finalizers are run, which is non-deterministic.

#328 

cc @halter73 @benaadams @migueldeicaza


halter73:
We should probably do the same for `UvPipeHandle`s in `PipeListener`.

cesarbs:
Now we have it in all relevant places. `ListenerSecondary` was already doing the right thing.

halter73:
Probably should have mentioned this before, but I think we should also move the calls to `DispatchConnection` into the try blocks.

cesarbs:
In that case we wouldn't need the explicit calls to `Dispose` since `DispatchConnection` should (indirectly) take care of it, right?

halter73:
@cesarbs It won't if it throws. Technically, double dispose shouldn't be unsafe anyway.

cesarbs:
@halter73 Done.

halter73:
:shipit:

PR #491, opened 2015-12-16
Use pre-encoded headers if possible

JunTaoLuo:
Kestrel update after pre-encoded bytes are added to `StringValues` in https://github.com/aspnet/Common/pull/68


benaadams:
May need to delete some lines in the generated ClearFast methods to get the full benefit of this change: https://github.com/aspnet/KestrelHttpServer/issues/493

benaadams:
Added PR for that issue "Faster Headers.ClearFast" #497

halter73:
Would there be a perf-hit to use pre-encoded StringValues instead of using the special-cased `FrameResponseHeaders.SetRaw(Connection|Date|TransferEncoding|ContentLength|Server)` methods we use today?

JunTaoLuo:
Revisiting this right now. Instead of rebasing on top of a month of work, I'll open a new PR.

PR #494, opened 2015-12-16
Don't use Vector<byte>.Dot

### LINKED ISSUE #493: Don't use Vector<byte>.Dot
`Vector<byte>.Dot` is not an intrinsic so quite slow; see https://github.com/dotnet/coreclr/issues/2359 so don't use it. 

> `Vector.Dot<byte>` is [not an intrinsic](https://github.com/dotnet/coreclr/blob/master/src/jit/simd.cpp#L2392) single CPU instruction and is instead composed of a [new object create] (https://github.com/dotnet/corefx/blob/master/src/System.Numerics.Vectors/src/System/Numerics/Vector.cs#L4142) followed by [32 byte multiply instructions and 32 byte addition instructions](https://github.com/dotnet/corefx/blob/master/src/System.Numerics.Vectors/src/System/Numerics/Vector.cs#L4145) (depending on arch). 
> dotproduct-cpu
> For the aspnet5 server Kestrel this adds up to being 8.23% of the *total* server CPU load as it makes heavy use of it when parsing the ASCII headers.

Need alternative implementation for `MemoryPoolIterator2.Seek`

benaadams:
23% improvement in RPS

Faster implementation than using `Vector<byte>.Dot`

Resolves #493

Before
    Latency     8.68ms   14.73ms 333.56ms   95.18%
    Req/Sec    47.81k     4.58k  122.10k    90.28%
  44317883 requests in 30.10s, 5.45GB read
Requests/sec: 1472388.96

After
    Latency    11.18ms   28.91ms 558.18ms   94.56%
    Req/Sec    59.52k     8.66k  282.83k    93.10%
  54796839 requests in 30.10s, 6.74GB read
Requests/sec: 1820474.06

### DIFF
--- a/src/Microsoft.AspNet.Server.Kestrel/Infrastructure/MemoryPoolIterator2.cs
+++ b/src/Microsoft.AspNet.Server.Kestrel/Infrastructure/MemoryPoolIterator2.cs
@@ -3,25 +3,12 @@
 
 using System;
 using System.Diagnostics;
-using System.Linq;
 using System.Numerics;
 
 namespace Microsoft.AspNet.Server.Kestrel.Infrastructure
 {
     public struct MemoryPoolIterator2
     {
-        /// <summary>
-        /// Array of "minus one" bytes of the length of SIMD operations on the current hardware. Used as an argument in the
-        /// vector dot product that counts matching character occurrence.
-        /// </summary>
-        private static Vector<byte> _dotCount = new Vector<byte>(Byte.MaxValue);
-
-        /// <summary>
-        /// Array of negative numbers starting at 0 and continuing for the length of SIMD operations on the current hardware.
-        /// Used as an argument in the vector dot product that determines matching character index.
-        /// </summary>
-        private static Vector<byte> _dotIndex = new Vector<byte>(Enumerable.Range(0, Vector<byte>.Count).Select(x => (byte)-x).ToArray());
-
         private MemoryPoolBlock2 _block;
         private int _index;
 
@@ -146,7 +133,6 @@ namespace Microsoft.AspNet.Server.Kestrel.Infrastructure
             }
 
             var byte0 = (byte)char0;
-            var vectorStride = Vector<byte>.Count;
             var ch0Vector = new Vector<byte>(byte0);
 
             var block = _block;
@@ -169,27 +155,20 @@ namespace Microsoft.AspNet.Server.Kestrel.Infrastructure
                 while (block.End != index)
                 {
                     var following = block.End - index;
-                    if (following >= vectorStride)
+                    if (following >= Vector<byte>.Count)
                     {
                         var data = new Vector<byte>(array, index);
                         var ch0Equals = Vector.Equals(data, ch0Vector);
-                        var ch0Count = Vector.Dot(ch0Equals, _dotCount);
 
-                        if (ch0Count == 0)
+                        if (ch0Equals.Equals(Vector<byte>.Zero))
                         {
-                            index += vectorStride;
+                            index += Vector<byte>.Count;
                             continue;
                         }
-                        else if (ch0Count == 1)
-                        {
-                            _block = block;
-                            _index = index + Vector.Dot(ch0Equals, _dotIndex);
-                            return char0;
-                        }
-                        else
-                        {
-                            following = vectorStride;
-                        }
+
+                        _block = block;
+                        _index = index + FindFirstEqualByte(ch0Equals);
+                        return char0;
                     }
                     while (following > 0)
                     {
@@ -215,7 +194,6 @@ namespace Microsoft.AspNet.Server.Kestrel.Infrastructure
 
             var byte0 = (byte)char0;
             var byte1 = (byte)char1;
-            var vectorStride = Vector<byte>.Count;
             var ch0Vector = new Vector<byte>(byte0);
             var ch1Vector = new Vector<byte>(byte1);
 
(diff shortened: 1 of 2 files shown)


benaadams:
Has a bug - adding more tests

halter73:
23%! That's great. We're going to run this in our perflab along with the StringValues change to a class (https://github.com/aspnet/Common/pull/68) and report back with the combined results.

src/Microsoft.AspNet.Server.Kestrel/Infrastructure/MemoryPoolIterator2.cs
```diff
@@ -146,7 +133,6 @@ public int Seek(int char0)
             }
 
             var byte0 = (byte)char0;
-            var vectorStride = Vector<byte>.Count;
```
davidfowl: why this change?

benaadams: Read that it's Jitted to a const; will try to find reference

src/Microsoft.AspNet.Server.Kestrel/Infrastructure/MemoryPoolIterator2.cs
```diff
@@ -417,6 +397,34 @@ public int Seek(int char0, int char1, int char2)
...
+            // Quasi-tree search
+            var vector64 = Vector.AsVectorInt64(chEquals);
+            for (var i = 0; i < Vector<long>.Count; i++)
+            {
+                var longValue = vector64[i];
+                if (longValue == 0) continue;
+
+                var shift = i << 1;
+                var offset = shift << 2;
+                var vector32 = Vector.AsVectorInt32(chEquals);
```
halter73: Out of curiosity, have you done any micro benchmarks to confirm that creating `vector32` to determine whether or not to check the first or second half of the long is faster than naively checking up to all 8 bytes for random/evenly distributed `chEquals` inputs?

benaadams: Is probably worth a test; though the `AsVectorType` conversions shouldn't* do anything post Jit in release as its all the same register and should just be changing the cpu instructions called on them. Also one of the reasons I did it with vector indexers `[]` rather than shift+cast; is endianness confuses me.

*doesn't mean they don't

halter73: Cool. I'm going to try to get this in quickly. Hard to argue with a 23% RPS gain!

benaadams: I can do a follow up PR if need be, after testing. Might be some stuff can get moved further down into coreclr/corefx types.

halter73: I did the microbenchmarking myself ([Program.cs](https://gist.github.com/halter73/236f9f300b7f5cfe1d40)), and it appears that it's better to do the "Quasi-tree search". At least it doesn't seem any worse. Here are the results from my laptop:

    C:\Users\shalter\dev\Universe\KestrelHttpServer\samples\ConsoleApp1 [dev +1 ~2 -0 !]> # FindFirstEqualByte
    C:\Users\shalter\dev\Universe\KestrelHttpServer\samples\ConsoleApp1 [dev +1 ~2 -0 !]> dnx --configuration Release ConsoleApp1
    1635
    C:\Users\shalter\dev\Universe\KestrelHttpServer\samples\ConsoleApp1 [dev +1 ~2 -0 !]> dnx --configuration Release ConsoleApp1
    1640
    C:\Users\shalter\dev\Universe\KestrelHttpServer\samples\ConsoleApp1 [dev +1 ~2 -0 !]> dnx --configuration Release ConsoleApp1
    1453

    C:\Users\shalter\dev\Universe\KestrelHttpServer\samples\ConsoleApp1 [dev +1 ~2 -0 !]> # FindFirstEqualByte2
    C:\Users\shalter\dev\Universe\KestrelHttpServer\samples\ConsoleApp1 [dev +1 ~2 -0 !]> dnx --configuration Release ConsoleApp1
    1715
    C:\Users\shalter\dev\Universe\KestrelHttpServer\samples\ConsoleApp1 [dev +1 ~2 -0 !]> dnx --configuration Release ConsoleApp1
    1697
    C:\Users\shalter\dev\Universe\KestrelHttpServer\samples\ConsoleApp1 [dev +1 ~2 -0 !]> dnx --configuration Release ConsoleApp1
    1688

Merging now.

darcythomas:
What is the environment that you ran these tests on? (So I can make some apple to apple comparisons)

halter73:
1.0.0-rc2-16312 coreclr x64 win dnx on my laptop (intel core i7-3367U)

halter73:
It's hard to make apple-to-apple comparisons unless you run the baseline on the exact same machine.

darcythomas:
Apple to pear comparisons perhaps :P 

I was wondering if it was something "Standard-ish" like a A3 VM instance on Azure, or something...

benaadams:
@darcythomas you need to run in release (SIMD optimizations are weakened in debug; so its possible to debug them). `Vector<float>.Dot` is hardware accelerated, but there isn't an equivalent cpu instruction for `Vector<byte>.Dot`.

On a A3 VM you might max out its network first; I'm testing with G4 VMs as already above 2.3GBps sends.

PR #497, opened 2015-12-16
Faster Headers.ClearFast

### LINKED ISSUE #492: Reduce CPU for Headers.ClearFast
`FrameResponseHeaders.ClearFast` and `FrameRequestHeaders.ClearFast` consume 4.75% cpu between them

cpu-trace

As they set (copy) 71 fields of the struct `StringValues.Empty`; all these can be dropped as they have already been "cleared" by setting `_bits = 0;`

Currently would be 71 small copies of 8 bytes (568 bytes) and 16 bytes with the https://github.com/aspnet/KestrelHttpServer/pull/491 changes (1136 bytes); which cloud be saved.

benaadams:
10% improvement in RPS

Issue (2 ClearFast lines)
cpu-trace

Before
> Running 30s test @ http://..../plaintext
    Latency     9.46ms   18.88ms 428.30ms   96.09%
    Req/Sec    46.14k     9.03k  194.88k    92.81%
  44125265 requests in 30.10s, 5.42GB read
Requests/sec: 1465986.87
Transfer/sec:    184.55MB

After
    Latency     9.76ms   20.82ms 366.01ms   95.03%
    Req/Sec    52.65k     5.22k  116.78k    90.51%
  48788229 requests in 30.10s, 6.00GB read
Requests/sec: 1620849.91
Transfer/sec:    204.04MB

Resolves #492

### DIFF
--- a/src/Microsoft.AspNet.Server.Kestrel/Http/FrameHeaders.Generated.cs
+++ b/src/Microsoft.AspNet.Server.Kestrel/Http/FrameHeaders.Generated.cs
@@ -60,7 +60,11 @@ namespace Microsoft.AspNet.Server.Kestrel.Http
         {
             get
             {
-                return _CacheControl;
+                if (((_bits & 1L) != 0))
+                {
+                    return _CacheControl;
+                }
+                return StringValues.Empty;
             }
             set
             {
@@ -72,7 +76,11 @@ namespace Microsoft.AspNet.Server.Kestrel.Http
         {
             get
             {
-                return _Connection;
+                if (((_bits & 2L) != 0))
+                {
+                    return _Connection;
+                }
+                return StringValues.Empty;
             }
             set
             {
@@ -84,7 +92,11 @@ namespace Microsoft.AspNet.Server.Kestrel.Http
         {
             get
             {
-                return _Date;
+                if (((_bits & 4L) != 0))
+                {
+                    return _Date;
+                }
+                return StringValues.Empty;
             }
             set
             {
@@ -96,7 +108,11 @@ namespace Microsoft.AspNet.Server.Kestrel.Http
         {
             get
             {
-                return _KeepAlive;
+                if (((_bits & 8L) != 0))
+                {
+                    return _KeepAlive;
+                }
+                return StringValues.Empty;
             }
             set
             {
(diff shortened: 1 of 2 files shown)


src/Microsoft.AspNet.Server.Kestrel/Http/FrameHeaders.Generated.cs
```diff
@@ -7441,43 +7682,6 @@ protected override bool RemoveFast(string key)
...
-            _Server = StringValues.Empty;
-            _SetCookie = StringValues.Empty;
-            _Vary = StringValues.Empty;
-            _WWWAuthenticate = StringValues.Empty;
-            
-            _rawConnection = null;
-            _rawDate = null;
-            _rawTransferEncoding = null;
-            _rawContentLength = null;
-            _rawServer = null;
```
halter73: The _raw fields are only read by `CopyToFast` which already has `if ({header.TestBit()})`, so no other changes needed here.

halter73:
Sweet!

lodejard:
Hmm... If the headers aren't zeroed out after a request, and the application is responding with computed strings, couldn't this result in more computed string entering the gen1/gen2 heaps?

I mean, yes better for raw RPS, but worse for normal application loads?

benaadams:
>  result in more computed string entering the gen1/gen2 heaps?

Assuming the setters are clearing the references (either by setting a full StringValue set, or by setting the non-set components to null/empty) then it would only be rarely set headers that would be promoted, as the other ones would be cleared in the next request set - its just avoiding both a double clear (clear+set); and a clear of items that were never set in the first place.

Not sure about the implications pre-encoded change though; haven't looked too closely.

lodejard:
I'm not so sure that's a rare case - typically browsers will keep a couple connections open at a time, and you requests to arrive in small bursts with longish delays between. If reset doesn't zero the memory then all of the strings from the previous response will still be referenced by the header object during those long delays. So if you have a thousand browsers in the world with your web page open, and the GC kicks off, all of the strings from the most recent response for the thousand connections will move into the next generation.

Pre-encoded and string literals are no problem though - those should be treated like static readonly constants once they are created, and should move into the gen2 heap and stay there forever.

benaadams:
Solution for that might be to pool the header collections between connections so they aren't active in the keep-alive wait state?

PR #500, opened 2015-12-17
Horizontal initialize repeat vectors once

benaadams:
+9.7% RPS

Before
    Latency     7.81ms   18.85ms 316.60ms   96.24%
    Req/Sec    64.00k    12.48k  271.04k    93.95%
  61204200 requests in 30.10s, 7.52GB read
Requests/sec: 2033518.93
Transfer/sec:    255.99MB

After 
    Latency     6.43ms   14.78ms 309.77ms   96.32%
    Req/Sec    72.50k     7.76k  303.39k    91.70%
  67177075 requests in 30.10s, 8.26GB read
Requests/sec: 2231776.12
Transfer/sec:    280.95MB

### DIFF
--- a/src/Microsoft.AspNet.Server.Kestrel/Http/Frame.cs
+++ b/src/Microsoft.AspNet.Server.Kestrel/Http/Frame.cs
@@ -6,6 +6,7 @@ using System.Collections.Generic;
 using System.IO;
 using System.Linq;
 using System.Net;
+using System.Numerics;
 using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
@@ -38,6 +39,12 @@ namespace Microsoft.AspNet.Server.Kestrel.Http
         private static readonly byte[] _bytesServer = Encoding.ASCII.GetBytes("\r\nServer: Kestrel");
         private static readonly byte[] _bytesDate = Encoding.ASCII.GetBytes("Date: ");
         private static readonly byte[] _bytesEndHeaders = Encoding.ASCII.GetBytes("\r\n\r\n");
+
+        private static readonly Vector<byte> _vectorCRs = new Vector<byte>((byte)'\r');
+        private static readonly Vector<byte> _vectorColons = new Vector<byte>((byte)':');
+        private static readonly Vector<byte> _vectorSpaces = new Vector<byte>((byte)' ');
+        private static readonly Vector<byte> _vectorQuestionMarks = new Vector<byte>((byte)'?');
+        private static readonly Vector<byte> _vectorPercentages = new Vector<byte>((byte)'%');
 
         private readonly object _onStartingSync = new Object();
         private readonly object _onCompletedSync = new Object();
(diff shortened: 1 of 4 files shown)


src/Microsoft.AspNet.Server.Kestrel/Infrastructure/MemoryPoolIterator2.cs
```diff
@@ -228,18 +228,13 @@ public int Seek(int char0)
             }
         }
 
-        public int Seek(int char0, int char1)
+        public int Seek(Vector<byte> byte0Vector, Vector<byte> byte1Vector)
```
halter73: Why not continue using `ch0Vector`, `ch0Equals`, `ch0Index` names?

Makes this PR a little easier to review.

cesarbs:
:shipit:

src/Microsoft.AspNet.Server.Kestrel/Http/Frame.cs
```diff
@@ -39,6 +40,12 @@ public abstract partial class Frame : FrameContext, IFrameControl
         private static readonly byte[] _bytesDate = Encoding.ASCII.GetBytes("Date: ");
         private static readonly byte[] _bytesEndHeaders = Encoding.ASCII.GetBytes("\r\n\r\n");
 
+        private static readonly Vector<byte> _vectorCRs = new Vector<byte>((byte)'\r');
```
halter73: It's impressive what a difference avoiding struct initialization is making at this point.

benaadams: I'm guessing its because it has to replicate the byte 32 times for each of the Vector's entries. Would probably get a similar gain by initializing it with a repeated array instead; except the framework would probably add some bounds checks...

src/Microsoft.AspNet.Server.Kestrel/Infrastructure/MemoryPoolIterator2.cs
benaadams: I thought someone wouldn't complaint they weren't representative; will change back :)

src/Microsoft.AspNet.Server.Kestrel/Http/Frame.cs
benaadams: Though there is a lot of weird code for the constructor in the jitter so I'm not actaully sure what it does https://github.com/dotnet/coreclr/blob/master/src/jit/simd.cpp#L1874 but whatever it is - its avoiding that :D

src/Microsoft.AspNet.Server.Kestrel/Infrastructure/MemoryPoolIterator2.cs
halter73: No need to change it. @cesarbs has already reviewed it to make sure you didn't screw anything up with the renames :stuck_out_tongue:

halter73:
:shipit:

PR #501, opened 2015-12-17
Mono not happy with client keep-alive header

benaadams:
*edit*
Moved test to correct place; to correctly test without relying on "keep-alive" behaviour of the client.


cesarbs:
Can you elaborate a little on this? Did you find any indications of why this is happening?

halter73:
Does this mean HttpClient is sometimes making all the requests on seperate connections on Mono? It might be best to have this as an `EngineTest` so we can be sure all requests are made on the same socket.

benaadams:
Not sure, isn't all the time; but enough to be annoying; and not always being Keep-alive isn't necessarily wrong; but bad for the test as it stands.

@halter73 yea; will move to `EngineTest` sounds like a good idea

benaadams:
@cesarbs this should fix the mono ci issues - if it doesn't cause hangs :P

benaadams:
Ignoring nuget being unhappy

halter73:
:shipit:

cesarbs:
:+1:

PR #510, opened 2015-12-20
Jit readonly static optimization

benaadams:
Allow Jit to optimize primitive statics to consts by marking them readonly https://github.com/dotnet/coreclr/issues/1079


cesarbs:
:shipit:

PR #511, opened 2015-12-20
Faster byte[] copy

benaadams:
This makes me sad that this is even possible to do (+4% RPS)...

Issue: "Array.Copy & Buffer.BlockCopy x2 to x3 slower < 1kB" https://github.com/dotnet/coreclr/issues/2430

[Performance Graph](http://aoa.blob.core.windows.net/aspnet/VectorizedCopy.png)

Until that is fixed, there is this, +4% RPS

Before

    Latency     7.31ms   17.04ms 430.36ms   96.48%
    Req/Sec    63.61k    11.37k  121.07k    89.55%
  60603911 requests in 30.04s, 7.45GB read
Requests/sec: 2017281.16
Transfer/sec:    253.95MB

    Latency     7.86ms   18.76ms 319.53ms   96.10%
    Req/Sec    65.93k     7.22k  233.05k    91.79%
  61092441 requests in 30.10s, 7.51GB read
Requests/sec: 2029658.31
Transfer/sec:    255.50MB

    Latency     7.76ms   18.81ms 338.18ms   96.30%
    Req/Sec    65.90k     6.02k  141.64k    88.76%
  61119465 requests in 30.10s, 7.51GB read
Requests/sec: 2030569.81
Transfer/sec:    255.62MB

After

    Latency     6.54ms   16.04ms 421.70ms   97.95%
    Req/Sec    66.30k    13.82k  277.14k    91.29%
  63126220 requests in 30.10s, 7.76GB read
Requests/sec: 2097240.79
Transfer/sec:    264.01MB

    Latency     6.39ms   15.65ms 315.77ms   98.11%
    Req/Sec    68.62k     7.74k  234.24k    87.15%
  63593990 requests in 30.10s, 7.82GB read
  Socket errors: connect 35, read 0, write 0, timeout 8
Requests/sec: 2112790.91
Transfer/sec:    265.97MB

    Latency     6.25ms   15.05ms 332.33ms   98.25%
    Req/Sec    66.43k    13.67k  283.71k    92.85%
  63515232 requests in 30.10s, 7.81GB read
Requests/sec: 2110181.96
Transfer/sec:    265.64MB

### DIFF
--- a/src/Microsoft.AspNet.Server.Kestrel/Filter/SocketInputStream.cs
+++ b/src/Microsoft.AspNet.Server.Kestrel/Filter/SocketInputStream.cs
@@ -5,6 +5,7 @@ using System;
 using System.IO;
 using System.Threading;
 using System.Threading.Tasks;
+using IllyriadGames.ByteArrayExtensions;
 using Microsoft.AspNet.Server.Kestrel.Http;
 using Microsoft.AspNet.Server.Kestrel.Infrastructure;
 
@@ -76,7 +77,7 @@ namespace Microsoft.AspNet.Server.Kestrel.Filter
         {
             var inputBuffer = _socketInput.IncomingStart(count);
 
-            Buffer.BlockCopy(buffer, offset, inputBuffer.Data.Array, inputBuffer.Data.Offset, count);
+            buffer.VectorizedCopy(offset, inputBuffer.Data.Array, inputBuffer.Data.Offset, count);
 
             _socketInput.IncomingComplete(count, error: null);
         }
--- a/src/Microsoft.AspNet.Server.Kestrel/Infrastructure/MemoryPoolIterator2.cs
+++ b/src/Microsoft.AspNet.Server.Kestrel/Infrastructure/MemoryPoolIterator2.cs
@@ -4,6 +4,7 @@
 using System;
 using System.Diagnostics;
 using System.Numerics;
+using IllyriadGames.ByteArrayExtensions;
 
 namespace Microsoft.AspNet.Server.Kestrel.Infrastructure
 {
@@ -552,7 +553,7 @@ namespace Microsoft.AspNet.Server.Kestrel.Infrastructure
                     actual = count;
                     if (array != null)
                     {
-                        Buffer.BlockCopy(block.Array, index, array, offset, remaining);
+                        block.Array.VectorizedCopy(index, array, offset, remaining);
                     }
                     return new MemoryPoolIterator2(block, index + remaining);
                 }
@@ -561,7 +562,7 @@ namespace Microsoft.AspNet.Server.Kestrel.Infrastructure
                     actual = count - remaining + following;
                     if (array != null)
                     {
-                        Buffer.BlockCopy(block.Array, index, array, offset, following);
+                        block.Array.VectorizedCopy(index, array, offset, following);
                     }
                     return new MemoryPoolIterator2(block, index + following);
                 }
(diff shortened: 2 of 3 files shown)


src/Microsoft.AspNet.Server.Kestrel/Filter/SocketInputStream.cs
```diff
@@ -5,6 +5,7 @@
 using System.IO;
 using System.Threading;
 using System.Threading.Tasks;
+using IllyriadGames.ByteArrayExtensions;
```
davidfowl: ??? 😀

benaadams: `VectorizedCopy` is in a standalone lib as it has a wide applicable use in many different situations; adding a reference seemed easier than including the source here, or Common and trying to workout what do with the file headers (here and in the standalone lib). Happy to take advice 😀

Hopefully the clr will change so it isn't needed; and everything can get the benefit, but equally that might either be a change to memmove in the C Standard Lib; or will be coreclr only for stability, so I'm not holding my breath :frowning:

benaadams: Added reference as specific version number; so the code referenced is under nuget's control not mine :wink:

davidfowl: But you know we're not going to have a dependency on  `"IllyriadGames.ByteArrayExtensions": "1.0.0-beta-10"` in our web server right? :smile:

benaadams: Doubt it :smile: 
What about a source file with a different copyright notice; but using same licence (Apache License, Version 2.0)? Not sure you'd be comfortable with the precedent either...

benaadams: Want to drop net451 and move to only supporting net46? I've just been made aware of `Buffer.MemoryCopy` which should offer better performance... (Doesn't appear till 4.6 and coreclr)

Tragetaschen: `net46` will likely break Mono. Not necessarily the most recent released version or master, but quite likely the one release that comes most closely to working: 4.0.x.

davidfowl: We're going to be moving to .NET 4.6 for RC2. We'll need to see how much of mono breaks when we do that. Hopefully 4.2.x will have enough 4.6 support to run.

benaadams:
Closed as still investigating

src/Microsoft.AspNet.Server.Kestrel/Filter/SocketInputStream.cs
Tragetaschen: (sorry for continuing here)
Switching to .NET 4.6 will break current Mono users. The switch to the .NET 4.6 reference source only happened [back in October](https://github.com/mono/referencesource/commits/mono/mscorlib/system/array.cs), so all versions of Mono (including the most recent 4.2 and the mostly-working 4.0) don't have this member. So things will go back the way they were in beta2: aspnet/dnx#1105

davidfowl: Yep, this is known. I'm working with the mono guys to figure out which version of mono will support .NET 4.6 reference assemblies. Hopefully by the time RC2 is ready we'll have a version of mono that supports that.

Anderman:
@benaadams Any updates on this. I think it is still possible to make it faster than `Buffer.MemoryCopy`. I can make a PR for this but that would be without `using IllyriadGames.ByteArrayExtensions` (-;

benaadams:
Holding off till post v1.0.0 RTM

Issue #512, opened 2015-12-20
Seek with Vector is 5 times slower then 'normal' Seek


Anderman:
I was looking for performance gains. Because for small strings there is some overhead in setting up the vector and finding the first byte.

But, I found out that the difference is also true for a long string. 

Probably I do something wrong. But all test still working after changing the code

normal search
```
public unsafe int Seek(byte search)
        {
            if (IsDefault)
            {
                return -1;
            }
            var block = _block;
            var index = _index;
            while (true)
            {
                while (block.End == index)
                {
                    if (block.Next == null)
                    {
                        _block = block;
                        _index = index;
                        return -1;
                    }
                    block = block.Next;
                    index = block.Start;
                }
                fixed (byte* ptr = block.Array)
                {
                    for(;index< block.End ;index++)
                    {
                        if (*(ptr + index) == serach)
                        {
                            _block = block;
                            _index = index;
                            return search;
                        }
                    }
                }
            }
```

benaadams:
Been doing some benchmarks around into this; a change needed for the current version is the vectors need to be passed by `ref` as they are a large memory copy otherwise.

benaadams:
Also you need to be running x64 (+RyuJIT) for Vectors to mean hardware acceleration (and currently Windows)

benaadams:
You are right about the non-optimal shorter than vector path though.

Anderman:
I did some test for both seek functions with the program below.
Both has an processor load of 60%
```
Start normal seek...
normal seek 451 call/ms
normal seek 450 call/ms
normal seek 450 call/ms
normal seek 448 call/ms
 Start vector seek...
Vector seek 44 call/ms
Vector seek 44 call/ms
Vector seek 44 call/ms
Vector seek 44 call/ms

```


test program
```
    public class Program
    {
        private static readonly Vector<byte> _vectorSpaces = new Vector<byte>((byte)' ');
        private static readonly Vector<byte> _vectorQuestionMark = new Vector<byte>((byte)'?');
        public static void Main()
        {
            var myStr = @"
[...]

PR #513, opened 2015-12-20
SeekFast

Anderman:
Improved seek performance by factor 5 to 10. See [issue] (https://github.com/aspnet/KestrelHttpServer/issues/512)


benaadams:
May be `Vector.IsHardwareAccelerated == false issue`

Anderman:
That was the problem

PR #514, opened 2015-12-21
MemoryPoolIterator2.Seek performance

### LINKED ISSUE #512: Seek with Vector is 5 times slower then 'normal' Seek

### LINKED ISSUE #515: MemoryPoolIterator2.Seek slow when Vector.IsHardwareAccelerated == false
e.g. linux and x86

Linux hopefully resolved in RC2? dotnet/coreclr#983

benaadams:
Static vectors by ref rather than by copy
Faster path when length < `Vector<byte>.Count`
Skip Vector path when `Vector.IsHardwareAccelerated != true` unless in debug/test (e.g. linux and x86)

Linux hopefully resolved in RC2? https://github.com/dotnet/coreclr/issues/983

Resolves #512
Resolves #515

### DIFF
--- a/src/Microsoft.AspNet.Server.Kestrel/Infrastructure/MemoryPoolIterator2.cs
+++ b/src/Microsoft.AspNet.Server.Kestrel/Infrastructure/MemoryPoolIterator2.cs
@@ -9,6 +9,8 @@ namespace Microsoft.AspNet.Server.Kestrel.Infrastructure
 {
     public struct MemoryPoolIterator2
     {
+        private readonly static int _vectorSpan = Vector<byte>.Count;
+
         private MemoryPoolBlock2 _block;
         private int _index;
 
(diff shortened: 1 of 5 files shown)


src/Microsoft.AspNet.Server.Kestrel/Infrastructure/MemoryPoolIterator2.cs
```diff
@@ -168,113 +170,141 @@ public unsafe long PeekLong()
...
-                    var following = block.End - index;
-                    if (following >= Vector<byte>.Count)
+#if !DEBUG // Need unit tests to test Vector path
+                    // Check will be Jitted away https://github.com/dotnet/coreclr/issues/1079
+                    if (Vector.IsHardwareAccelerated)
                     {
-                        var data = new Vector<byte>(array, index);
-                        var byte0Equals = Vector.Equals(data, byte0Vector);
+#endif
+                    if (following >= _vectorSpan)
```
cesarbs: Fix indentation.

cesarbs:
Did you do any benchmarks for this change?

benaadams:
Will revisit with #519 as source (or rather with hindsight of knowing where am going)

cesarbs:
I really like the passing as `ref` aspect of this PR. Are the other changes required because of it? If you could send a PR with only the change to using `ref`, that would be great to benchmark in our setup.

benaadams:
Removing the `readonly` is needed to pass by ref :(

benaadams:
@cesarbs updated in #524

PR #516, opened 2015-12-22
I/O Streamlining and Cleanup

benaadams:
I imagine will have to break this one up..?


cesarbs:
Breaking it up would be nice :smile:

cesarbs:
Also, what exactly is the purpose of these changes?

benaadams:
Will revisit with #519 as source (or rather with hindsight of knowing where am going)

PR #518, opened 2015-12-22
Remove dnx451 and dnxcore50 TFM's

henkmollema:
As per https://github.com/aspnet/Hosting/issues/531.

### DIFF
--- a/src/Microsoft.AspNet.Server.Kestrel/Networking/Libuv.cs
+++ b/src/Microsoft.AspNet.Server.Kestrel/Networking/Libuv.cs
@@ -12,7 +12,7 @@ namespace Microsoft.AspNet.Server.Kestrel.Networking
         {
             IsWindows = PlatformApis.IsWindows;
 
-            var isDarwinMono = !IsWindows && PlatformApis.IsDarwin;
+            var isDarwinMono = PlatformApis.IsDarwin && PlatformApis.IsMono;
 
             if (isDarwinMono)
             {
--- a/src/Microsoft.AspNet.Server.Kestrel/Networking/Libuv.cs
+++ b/src/Microsoft.AspNet.Server.Kestrel/Networking/Libuv.cs
@@ -12,12 +12,7 @@ namespace Microsoft.AspNet.Server.Kestrel.Networking
         {
             IsWindows = PlatformApis.IsWindows;
 
-            var isDarwinMono =
-#if DNX451
-                IsWindows ? false : PlatformApis.IsDarwin;
-#else
-                false;
-#endif
+            var isDarwinMono = !IsWindows && PlatformApis.IsDarwin;
 
             if (isDarwinMono)
             {
--- a/src/Microsoft.AspNet.Server.Kestrel/Networking/PlatformApis.cs
+++ b/src/Microsoft.AspNet.Server.Kestrel/Networking/PlatformApis.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Runtime.InteropServices;
+using Microsoft.Extensions.PlatformAbstractions;
 
 namespace Microsoft.AspNet.Server.Kestrel.Networking
 {
@@ -24,11 +25,15 @@ namespace Microsoft.AspNet.Server.Kestrel.Networking
                 IsDarwin = string.Equals(GetUname(), "Darwin", StringComparison.Ordinal);
             }
 #endif
+
+            IsMono = PlatformServices.Default.Runtime.RuntimeType == "Mono";
         }
 
         public static bool IsWindows { get; }
 
         public static bool IsDarwin { get; }
+
+        public static bool IsMono { get; }
 
         [DllImport("libc")]
         static extern int uname(IntPtr buf);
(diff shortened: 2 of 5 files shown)


henkmollema:
Meh, apparently it's not that simple..

davidfowl:
Looks like it might be the `ICompileModule`. We need to remove them.

cesarbs:
Did you check the code for any `#if DNX451` or `DNXCORE50`?

henkmollema:
@cesarbs yeah there are some. I was trying to get rid of them but I got some vague errors in `Microsoft.DNX.targets`. I'll try again in a minute.

henkmollema:
@davidfowl / @cesarbs what to do with this guy:
```csharp
var isDarwinMono =
#if DNX451
    IsWindows ? false : PlatformApis.IsDarwin;
#else
    false;
#endif
```

It's in `Libuv.cs`.

halter73:
Maybe this is the way to go?

    return Type.GetType("Mono.Runtime") != null;

http://stackoverflow.com/a/721194/719967

Are you able to test if everything works on Mac OS X w/ Mono after this change?

davidfowl:
Use `Microsoft.Extensions.PlatformAbstractions`

henkmollema:
It builds after removing the `StandardsPoliceCompileModule` usage, so the problem lies in there. Perhaps because `Microsoft.StandardsPolice` still targets `dnx451` and `dnxcore50`?

I'm not really sure how to test this correctly. Perhaps it's better if one of you guys take over?

davidfowl:
That sounds about right.

src/Microsoft.AspNet.Server.Kestrel/Networking/Libuv.cs
```diff
@@ -12,12 +12,7 @@ public Libuv()
         {
             IsWindows = PlatformApis.IsWindows;
 
-            var isDarwinMono =
-#if DNX451
-                IsWindows ? false : PlatformApis.IsDarwin;
-#else
-                false;
-#endif
+            var isDarwinMono = !IsWindows && PlatformApis.IsDarwin;
```
davidfowl: This check isn't right though.

henkmollema: I figured.. Not sure how to replace the `#if DNX451`..

davidfowl: Microsoft.Extensions.PlatformAbstractions:

https://github.com/aspnet/PlatformAbstractions/blob/dev/src/Microsoft.Extensions.PlatformAbstractions/IRuntimeEnvironment.cs#L29

RuntimeType

halter73: @henkmollema Do you plan on making the change to use RuntimeType?

henkmollema: Whoops, forgot about this one. Yes, I'll look into it right not.

/cc @davidfowl @halter73

henkmollema:
@halter73 @davidfowl updated it with `PlatformServices`. Is this correct now?

henkmollema:
And how about the `StandardsPoliceCompileModule`?

halter73:
Thanks. It looks good, but I'm not sure about removing `StandardsPoliceCompileModule`. Do you still get warnings for violations?

I'm going to look into it and get back to you.

davidfowl:
It's going to have to be removed as part of the move to the CLI.

src/Microsoft.AspNet.Server.Kestrel/Networking/PlatformApis.cs
```diff
@@ -24,12 +25,16 @@ static PlatformApis()
                 IsDarwin = string.Equals(GetUname(), "Darwin", StringComparison.Ordinal);
             }
 #endif
+
+            IsMono = PlatformServices.Default.Runtime.RuntimeType == "Mono";
```
davidfowl: :+1:

moozzyk: I think we will need/be able to get rid of this check and the entire `[DllImport("__Internal"` hack once we move away from dnx. Users will have to install libuv on Mac when using Mono.

henkmollema:
Do you guys want me to do anything for this PR?

/cc @davidfowl / @halter73

src/Microsoft.AspNet.Server.Kestrel/Networking/PlatformApis.cs
davidfowl: Yap

src/Microsoft.AspNet.Server.Kestrel/compiler/preprocess/SPCM.cs
```diff
@@ -1,6 +0,0 @@
-﻿namespace Microsoft.AspNet.Server.Kestrel
-{
-    public class StandardsPoliceCompileModule : Microsoft.StandardsPolice.StandardsPoliceCompileModule
```
halter73: So we are no longer getting any warning from standards police after this is removed, but I see the issue you are running into. Having any .cs file in the compiler/preprocess folder is causing dnu build to fail with a null ref violation.

@davidfowl Any idea why this is happening now?

davidfowl: Because of the target framework changes. It can't find the right dependencies in the lock file. Null ref is bad but DNX is gone anyways. 

We can turn this into an analyzer once we get support in the cli for that

halter73:
Thanks!

henkmollema:
:smile:

PR #519, opened 2015-12-22
Xmas combo

benaadams:
Best yet... (though will probably have to break it down into pieces?)

Azure G4 Peaking 2.7 GBps send 1.1GBps receive

344,334 rps Non pipelined w/ 3.44ms latency (32 threads and 1024 connections)
2,398,936 rps 16 pipelined w/ 5.66ms latency (32 threads and 1024 connections)

(3.9 - 4.2 Gbps send / 1.4 - 1.5 Gbps receive when pipelined 512)

PR #521, opened 2015-12-23
Fast path Consume when already complete

benaadams:
Fast path `Consume` when content complete by not using async state machine (is normally complete)

From #519

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Kestrel/Http/MessageBody.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/Http/MessageBody.cs
@@ -5,6 +5,7 @@ using System;
 using System.IO;
 using System.Threading;
 using System.Threading.Tasks;
+using Microsoft.AspNetCore.Server.Kestrel.Infrastructure;
 
 namespace Microsoft.AspNetCore.Server.Kestrel.Http
 {
@@ -35,7 +36,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Http
             return result;
         }
 
-        public async Task Consume(CancellationToken cancellationToken = default(CancellationToken))
+        public Task Consume(CancellationToken cancellationToken = default(CancellationToken))
         {
             ValueTask<int> result;
             var send100checked = false;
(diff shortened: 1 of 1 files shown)


src/Microsoft.AspNet.Server.Kestrel/Http/MessageBody.cs
```diff
@@ -38,35 +39,40 @@ public Task<int> ReadAsync(ArraySegment<byte> buffer, CancellationToken cancella
...
+            }
+            else if (result.GetAwaiter().GetResult() == 0)
+            {
+                // Completed Task, end of stream
+                return TaskUtilities.CompletedTask;
+            }
+            else
+            {
+                // Completed Task, but non-zero get next Task and await
+                return ConsumeAwaited(ReadAsyncImplementation(default(ArraySegment<byte>), cancellationToken), cancellationToken);
```
halter73: I don't understand this. Why not put his call to `ReadAsyncImplementation` inside of `ConsumeAwaited`? You already do this inside the while condition.

halter73:
Will have to be updated post ValueTask merge, but aside from my one comment this is good to :ship:

src/Microsoft.AspNet.Server.Kestrel/Http/MessageBody.cs
benaadams: Yeah is a bit weird; enough to make me look at it for a good few minutes before working out why.

Is because the non-completed one passes in the current task so is to match that. Will have a think on how it can be made better.

benaadams: Swapped the parameters; and made the task have a null default. Better?

benaadams:
This is a a bit fiddly with the ValueTask; merge to rebase, will take a little bit of time.

benaadams:
Closing this; I think I'd need to start again.

benaadams:
Rebased and reworked to work with the ValueTask changes

halter73:
:shipit:

Issue #304, opened 2015-10-29
Limit size of memory buffer when reading request


Tratcher:
As @halter73 explained it to me, kestrel will constantly read the incoming request body off the socket and buffer it in memory until the application retrieves it. This could consume significant amounts of memory if the app isn't draining it fast enough.

Should Kestrel implement IHttpBufferingFeature.DisableRequestBuffering to stop preemptively buffering the request body and only read it on-demand? This would also let you put back-pressure on the client if they are sending data too quickly. This is also something that may make sense to enable/disable at startup.

benaadams:
Or a `max buffered` property?

muratg:
@cesarbs Feel free to ping @troydai as he has some context on this one.

lodejard:
Having a max read-ahead would be a good idea. Disabling request buffering would be a different thing- that would be to stop the server or middleware from remembering the body, to disable "seek" rewinding

DamianEdwards:
post RC2

ChristineBoersen:
This is definitely an issue when sending an AJAX upload back to a controller with a **LARGE** file (hundreds of MB), as even if you use CopyToAsync to a filestream object, since the HttpContext.Request.Body stream was buffered, you run into an out of memory error. 

Is there a Kestrel Max Upload Size / Max Request Size equivalent to what used to be set in the web.config pre-kestrel that could be used as a workaround until this is resolved?

muratg:
Thanks @mikeharder!

PR #525, opened 2015-12-23
Simplify SocketInput, remove lock, only use pooled blocks

benaadams:
From #519

Partial #516

### DIFF
--- a/src/Microsoft.AspNet.Server.Kestrel/Http/SocketInput.cs
+++ b/src/Microsoft.AspNet.Server.Kestrel/Http/SocketInput.cs
@@ -116,13 +116,6 @@ namespace Microsoft.AspNet.Server.Kestrel.Http
             Complete();
         }
 
-        public void AbortAwaiting()
-        {
-            _awaitableError = new ObjectDisposedException(nameof(SocketInput), "The request was aborted");
-
-            Complete();
-        }
-
         private void Complete()
         {
             var awaitableState = Interlocked.Exchange(
@@ -175,6 +168,13 @@ namespace Microsoft.AspNet.Server.Kestrel.Http
                 returnStart = returnStart.Next;
                 returnBlock.Pool.Return(returnBlock);
             }
+        }
+
+        public void AbortAwaiting()
+        {
+            _awaitableError = new ObjectDisposedException(nameof(SocketInput), "The request was aborted");
+
+            Complete();
         }
 
         public SocketInput GetAwaiter()
--- a/src/Microsoft.AspNet.Server.Kestrel/Http/SocketInput.cs
+++ b/src/Microsoft.AspNet.Server.Kestrel/Http/SocketInput.cs
@@ -5,7 +5,6 @@ using System;
 using System.IO;
 using System.Runtime.CompilerServices;
 using System.Threading;
-using System.Threading.Tasks;
 using Microsoft.AspNet.Server.Kestrel.Infrastructure;
 
 namespace Microsoft.AspNet.Server.Kestrel.Http
@@ -25,7 +24,6 @@ namespace Microsoft.AspNet.Server.Kestrel.Http
         private MemoryPoolBlock2 _head;
         private MemoryPoolBlock2 _tail;
         private MemoryPoolBlock2 _pinned;
-        private readonly object _sync = new Object();
 
         public SocketInput(MemoryPool2 memory, IThreadPool threadPool)
         {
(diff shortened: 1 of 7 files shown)


src/Microsoft.AspNet.Server.Kestrel/Http/SocketInput.cs
```diff
@@ -34,98 +32,104 @@ public SocketInput(MemoryPool2 memory, IThreadPool threadPool)
...
             }
+            if (error != null)
+            {
+                _awaitableError = error;
+            }
+
+            Complete();
+        }
+
+        public void AbortAwaiting()
```
cesarbs: nit: move this back after `ConsumingComplete` so there's less impact on the diff.

src/Microsoft.AspNet.Server.Kestrel/Http/SocketInput.cs
```diff
@@ -25,7 +24,6 @@ public class SocketInput : ICriticalNotifyCompletion
         private MemoryPoolBlock2 _head;
         private MemoryPoolBlock2 _tail;
         private MemoryPoolBlock2 _pinned;
-        private readonly object _sync = new Object();
```
cesarbs: Love it when locks can be removed :+1: Are you 100% positive this won't introduce any race conditions?

benaadams:
@cesarbs yes the idea came from something @halter73 pointed out in the SocketOutput

`IncomingStart`+`IncomingComplete` is only called by the libuv thread; `IncomingData` is only called when its a wrapped stream (so libuv won't be calling the other two methods); so the Incoming methods are synchronized with each other already as there is only one thing using them.

Likewise the `Consuming` methods have only one thing using them (single producer, single consumer).

In the cross over between the two:

`Consuming` only modifies `_head`
`Incoming` only modifies `_tail`+`_pinned` *except* on the very first read when `_head == null`, when it sets it to point at the start of the data.

If `ConsumingStart` read null for `_head` then it doesn't change `_head` in `ConsumingComplete` 

```csharp
if (!consumed.IsDefault)
{
    returnStart = _head;
    returnEnd = consumed.Block;
    _head = consumed.Block;
    _head.Start = consumed.Index;
}
```
So there isn't really any cross over; with the data so the lock isn't needed.

The timing between the `Incoming` and `Consuming` is controlled by the `Interlocked.CompareExchange(ref _awaitableState...` and `ManualResetEventSlim _manualResetEvent` as before.

benaadams:
Rebased so can make another PR based on it

halter73:
:+1: It does seem very similar to SocketOutput (which shouldn't be to surprising), and I think your right that the lock was unnecessary.

I wouldn't mind a second opinion from @lodejard though.

cesarbs:
I love this PR because I think it'll make my life easier when I get to work on https://github.com/aspnet/KestrelHttpServer/issues/304 :grin:

benaadams:
Still can't workout why the `GetDateHeaderValue_ReturnsUpdatedValueAfterIdle` test periodically fails

halter73:
@lodejard Was a little worried about removing the locks, but I think it's safe (assuming there aren't concurrent reads) and there is quite a nice perf gain. Here are my numbers:

dev:

	    Latency     8.18ms   35.18ms 494.60ms   93.79%
	    Req/Sec    36.07k     2.78k   69.91k    87.91%
	  11552580 requests in 10.10s, 1.42GB read
	Requests/sec: 1143851.17
	Transfer/sec:    143.99MB

	    Latency     4.70ms   18.91ms 342.23ms   94.38%
	    Req/Sec    35.99k     2.53k   63.87k    90.36%
	  11525358 requests in 10.10s, 1.42GB read
	  Socket errors: connect 0, read 0, write 529, timeout 0
	Requests/sec: 1141167.60
	Transfer/sec:    143.66MB

	    Latency     2.81ms    5.60ms 177.49ms   95.18%
	    Req/Sec    35.88k     1.86k   53.53k    86.22%
	  11512022 requests in 10.10s, 1.42GB read
	Requests/sec: 1139994.68
	Transfer/sec:    143.51MB

benaadams/socket-input-merged:

	    Latency     2.70ms    6.02ms 386.69ms   93.37%
	    Req/Sec    36.70k     1.48k   49.61k    78.44%
	  11790897 requests in 10.10s, 1.45GB read
	Requests/sec: 1167464.67
	Transfer/sec:    146.97MB

	    Latency     2.46ms    3.77ms 265.56ms   94.96%
	    Req/Sec    36.45k     2.12k   57.27k    83.90%
	  11695314 requests in 10.10s, 1.44GB read
	Requests/sec: 1158028.82
	Transfer/sec:    145.78MB

	    Latency     3.44ms    7.98ms 294.07ms   94.29%
	    Req/Sec    36.71k     2.20k   63.16k    84.17%
	  11771692 requests in 10.10s, 1.45GB read
	Requests/sec: 1165591.15
	Transfer/sec:    146.73MB

benaadams/socket-input-merged-lock:


	    Latency     4.55ms   14.87ms 277.05ms   93.21%
	    Req/Sec    35.74k     2.29k   57.51k    85.31%
	  11454930 requests in 10.10s, 1.41GB read
	Requests/sec: 1134174.28
	Transfer/sec:    142.78MB


	    Latency     4.13ms   12.74ms 273.82ms   95.53%
	    Req/Sec    35.79k     2.02k   55.83k    82.63%
	  11484589 requests in 10.10s, 1.41GB read
	Requests/sec: 1137144.99
	Transfer/sec:    143.15MB


	    Latency     3.92ms   12.20ms 266.46ms   96.73%
	    Req/Sec    35.75k     2.26k   63.15k    87.27%
	  11458948 requests in 10.10s, 1.41GB read
	Requests/sec: 1134531.78
	Transfer/sec:    142.82MB

[shalter/socket-input-safe-consume](https://github.com/aspnet/KestrelHttpServer/commit/7e9dd9e2bcc1a2e374b7627d655865678c40c343):

	    Latency     3.03ms    6.70ms 262.35ms   96.15%
	    Req/Sec    36.27k     2.13k   52.20k    87.72%
	  11642067 requests in 10.10s, 1.43GB read
	Requests/sec: 1152997.37
	Transfer/sec:    145.15MB


	    Latency     4.58ms   19.06ms 253.21ms   97.02%
	    Req/Sec    36.47k     2.23k   60.40k    90.38%
	  11691776 requests in 10.10s, 1.44GB read
	Requests/sec: 1157669.21
	Transfer/sec:    145.73MB

	    Latency     2.89ms    7.11ms 158.40ms   97.67%
	    Req/Sec    36.61k     2.17k   64.40k    91.24%
	  11730125 requests in 10.10s, 1.44GB read
	Requests/sec: 1161390.24
	Transfer/sec:    146.20MB

I added a [small change on top of yours to throw for concurrent non-async reads](https://github.com/aspnet/KestrelHttpServer/commit/7e9dd9e2bcc1a2e374b7627d655865678c40c343). This should actually catch potential bugs that weren't caught before.

And, as shown above, it doesn't seem to detrimental to perf. At least it's not blocking the libuv thread.

@benaadams What do you think?

benaadams:
Your change is good and is better than the lock, as it works across the ConsumingStart/Complete and will surface bugs in user code - which wasn't covered by the lock.

The Interlocked also acts a memory barrier at cpu level (as does lock).  It could be argued that's its overly defensive as nowhere is it ever suggested an instance socket read is a threadsafe operation; but its potentially a public port; RTM is approaching - safety first, and more importantly I clearly managed to do it at some point in our application with the way we were using websockets ;)

The strong synchronisation is still covered by it having interlocked awaitable states and perhaps overly covered with the addition of the ManualResetEventSlim (async all the things!); however I never saw it show up as an item of interest in perf testing.

I'd go with your addition; revisit post RTM if everything gets so fast it starts showing up as an hot spot?

benaadams:
LGTM (with https://github.com/aspnet/KestrelHttpServer/commit/7e9dd9e2bcc1a2e374b7627d655865678c40c343) looking to see if it will trigger exceptions on the issues I experienced earlier

benaadams:
Though maybe some tests for the double read ;-)

benaadams:
In testing I don't trigger overlapped consumes; but have managed to get a confirmation of previous related fix
> Microsoft.AspNet.Server.Kestrel[0]
Connection shutdown abnormally
System.IO.IOException: Concurrent reads are not supported. ---> System.InvalidOperationException: Concurrent reads are not supported.
   --- End of inner exception stack trace ---
   at Microsoft.AspNet.Server.Kestrel.Http.SocketInput.GetResult()
   at Microsoft.AspNet.Server.Kestrel.Http.Frame`1.<RequestProcessingAsync>d__3.MoveNext()

benaadams:
Also found where it was previously going wrong; in somewhere in my or middleware handling of websockets close - which is a weird area anyway.

PR #526, opened 2015-12-23
[Wip] Differentiate write and write2 + simplify

benaadams:
UvWriteReq only uses pooled blocks
UvWrite2Req only has one buffer
`List<GCHandle>` isn't needed

From #519

### DIFF
--- a/src/Microsoft.AspNet.Server.Kestrel/Networking/UvWriteReq2.cs
+++ b/src/Microsoft.AspNet.Server.Kestrel/Networking/UvWriteReq2.cs
@@ -1,0 +1,114 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Runtime.InteropServices;
+using Microsoft.AspNet.Server.Kestrel.Infrastructure;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.AspNet.Server.Kestrel.Networking
+{
+    /// <summary>
+    /// Summary description for UvWriteRequest2
+    /// </summary>
+    public class UvWrite2Req : UvRequest
+    {
+        private readonly static Libuv.uv_write_cb _uv_write2_cb = (IntPtr ptr, int status) => UvWrite2Cb(ptr, status);
+
+        private IntPtr _bufs;
+
+        private Action<UvWrite2Req, int, Exception, object> _callback;
+        private object _state;
+        private const int BUFFER_COUNT = 1;
+
+        private GCHandle _pinUvWrite2Req;
+        private GCHandle _pinBuffer;
+        private bool _bufferIsPinned;
+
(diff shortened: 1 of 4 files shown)


src/Microsoft.AspNet.Server.Kestrel/Networking/UvWriteReq2.cs
```diff
@@ -0,0 +1,114 @@
...
+using System.Runtime.InteropServices;
+using Microsoft.AspNet.Server.Kestrel.Infrastructure;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.AspNet.Server.Kestrel.Networking
+{
+    /// <summary>
+    /// Summary description for UvWriteRequest2
+    /// </summary>
+    public class UvWrite2Req : UvRequest
```
cesarbs: The class name and file name are inconsistent. I personally prefer `UvWriteReq2`.

benaadams: Renamed

src/Microsoft.AspNet.Server.Kestrel/Networking/UvWriteReq2.cs
```diff
@@ -0,0 +1,114 @@
...
+            CreateMemory(
+                loop.Libuv,
+                loop.ThreadId,
+                requestSize + bufferSize);
+            _bufs = handle + requestSize;
+        }
+
+        public unsafe void Write2(
+            UvStreamHandle handle,
+            ArraySegment<byte> buf,
```
halter73: I wonder if it would be better to put the dummy buffer in this class and use a static `pBuffers` object. Better encapsulation probably.

halter73: Better yet. Create the dummy buffer in this class using `CreateMemory`. This way the only `GCHandle` we have to bother with is `_pinUvWrite2Req`.

benaadams: Made change for this but can't test due to https://github.com/aspnet/Home/issues/1207

halter73: @benaadams Do you still have this change?

benaadams: Yes, though will have to add it to PR after Saturday (Playtest this weekend so focused on that)

halter73: Sounds good to me. Good luck with the playtest!

benaadams:
Still needs a change as above

muratg:
@benaadams Still working on this?

benaadams:
Not at the moment; would be a slight breaking change though would be weird if anyone but Kestrel took a dependency on them

benaadams:
Going to close this and raise as an issue

PR #524, opened 2015-12-23
Much faster Seek perf + non vector path

benaadams:
From  #519
Updated #514

Static vectors by ref rather than by copy
Faster path when length < `Vector<byte>.Count`
Skip Vector path when `Vector.IsHardwareAccelerated != true` unless in debug/test (e.g. linux and x86)

Linux hopefully resolved in RC2? https://github.com/dotnet/coreclr/issues/983

Common headers have common offsets

Also has Faster output header handling #528 in it :-/ due to the generated code file.

Resolves #512
Resolves #515

PR #528, opened 2015-12-23
Faster output header handling

benaadams:
From  #519

### DIFF
--- a/src/Microsoft.AspNet.Server.Kestrel/Http/Frame.cs
+++ b/src/Microsoft.AspNet.Server.Kestrel/Http/Frame.cs
@@ -640,14 +640,13 @@ namespace Microsoft.AspNet.Server.Kestrel.Http
         {
             var begin = SocketOutput.ProducingStart();
             var end = begin;
-            if (_keepAlive)
+            if (_keepAlive && _responseHeaders.HasConnection)
             {
-                foreach (var connectionValue in _responseHeaders.HeaderConnection)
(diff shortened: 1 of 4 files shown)


davidfowl:
How much does this actually save?

benaadams:
It could save more, but are some oddities with nulls (being valid stored values but not output; is even a test for it and related things)

Otherwise a fair bit and moves `CreateResponseHeader` out of the heavy functions; as without it, it repeatedly does:

`new Enumerator -> Movenext -> Current -> Movenext` for every header; as well as keep-alive check

davidfowl:
@benaadams I'd like to see numbers. This looks like an uber micro-optimization.

benaadams:
Perf is in https://github.com/aspnet/KestrelHttpServer/pull/519#issuecomment-166891420

Numbers are hard to get matching; but look at the ratios for FrameResponseHeaders.CopyToFast - CopyFrom is the same in both

Pre change 
pre-change
Without
post-change

benaadams:
Can run longer perf tests if you want more equivalent numbers?

benaadams:
@davidfowl some might be related to: https://github.com/aspnet/Common/pull/74

benaadams:
`StringValues` enumerator is currently fairly inefficient; this might become a de-opt after PR https://github.com/aspnet/Common/pull/74

benaadams:
Merged with  #524 due to Generated file clash

Issue #187, opened 2015-09-03
Style: Kestrel Cleanup part 2


lodejard:
- [x] private on embedded types
 - [x] readonly on fields assigned only by constructors

glennc:
Is this for right now same as the other cleanup tasks?

muratg:
Just talked to @lodejard.  Moving is issues to RC2.

PR #529, opened 2015-12-24
Cleanup: private embedded types and readonly where appropriate (#187).

cesarbs:
https://github.com/aspnet/KestrelHttpServer/issues/187

cc @halter73 @lodejard

### DIFF
--- a/src/Microsoft.AspNet.Server.Kestrel/Http/FrameRequestHeaders.cs
+++ b/src/Microsoft.AspNet.Server.Kestrel/Http/FrameRequestHeaders.cs
@@ -21,11 +21,11 @@ namespace Microsoft.AspNet.Server.Kestrel.Http
 
         public partial struct Enumerator : IEnumerator<KeyValuePair<string, StringValues>>
         {
-            private FrameRequestHeaders _collection;
-            private long _bits;
+            private readonly FrameRequestHeaders _collection;
+            private readonly long _bits;
             private int _state;
             private KeyValuePair<string, StringValues> _current;
-            private bool _hasUnknown;
+            private readonly bool _hasUnknown;
             private Dictionary<string, StringValues>.Enumerator _unknownEnumerator;
 
             internal Enumerator(FrameRequestHeaders collection)
--- a/src/Microsoft.AspNet.Server.Kestrel/Http/FrameResponseHeaders.cs
+++ b/src/Microsoft.AspNet.Server.Kestrel/Http/FrameResponseHeaders.cs
@@ -10,8 +10,8 @@ namespace Microsoft.AspNet.Server.Kestrel.Http
 {
     public partial class FrameResponseHeaders : FrameHeaders
     {
-        private static byte[] _CrLf = new[] { (byte)'\r', (byte)'\n' };
-        private static byte[] _colonSpace = new[] { (byte)':', (byte)' ' };
+        private static readonly byte[] _CrLf = new[] { (byte)'\r', (byte)'\n' };
+        private static readonly byte[] _colonSpace = new[] { (byte)':', (byte)' ' };
 
         public bool HasConnection => HeaderConnection.Count != 0;
 
@@ -53,11 +53,11 @@ namespace Microsoft.AspNet.Server.Kestrel.Http
 
         public partial struct Enumerator : IEnumerator<KeyValuePair<string, StringValues>>
         {
-            private FrameResponseHeaders _collection;
-            private long _bits;
+            private readonly FrameResponseHeaders _collection;
+            private readonly long _bits;
             private int _state;
             private KeyValuePair<string, StringValues> _current;
-            private bool _hasUnknown;
+            private readonly bool _hasUnknown;
             private Dictionary<string, StringValues>.Enumerator _unknownEnumerator;
 
             internal Enumerator(FrameResponseHeaders collection)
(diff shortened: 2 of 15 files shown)


src/Microsoft.AspNet.Server.Kestrel/Http/FrameRequestHeaders.cs
```diff
@@ -21,11 +21,11 @@ public Enumerator GetEnumerator()
...
         {
-            private FrameRequestHeaders _collection;
-            private long _bits;
+            private readonly FrameRequestHeaders _collection;
+            private readonly long _bits;
             private int _state;
             private KeyValuePair<string, StringValues> _current;
-            private bool _hasUnknown;
+            private readonly bool _hasUnknown;
             private Dictionary<string, StringValues>.Enumerator _unknownEnumerator;
```
cesarbs: For whatever reason some tests (`Microsoft.AspNet.Server.KestrelTests.FrameRequestHeadersTests.EntriesCanBeEnumerated` and `Microsoft.AspNet.Server.KestrelTests.FrameRequestHeadersTests.KeysAndValuesCanBeEnumerated`) fail when I set `_unknownEnumerator` to `readonly`.

benaadams: Have to be careful with `readonly` of in a struct as the compiler causes a defensive copy to be taken to prevent mutation so you don't modify the original which will be why making the `_unknownEnumerator` readonly will cause the tests to fail (as the enumerator needs to mutate) is ok in classes. Ref: http://codeblog.jonskeet.uk/2014/07/16/micro-optimization-the-surprising-inefficiency-of-readonly-fields/

benaadams: Trying to find specific for enum; but here is example of don't make mutable struct readonly: https://github.com/dotnet/coreclr/blob/bc146608854d1db9cdbcc0b08029a87754e12b49/src/mscorlib/src/System/Runtime/CompilerServices/AsyncMethodBuilder.cs#L40

benaadams: Safest is not to make struct fields `readonly` as there be dragons.

src/Microsoft.AspNet.Server.Kestrel/Http/FrameResponseHeaders.cs
```diff
@@ -53,12 +53,12 @@ public void CopyTo(ref MemoryPoolIterator2 output)
...
-            private FrameResponseHeaders _collection;
-            private long _bits;
+            private readonly FrameResponseHeaders _collection;
+            private readonly long _bits;
             private int _state;
             private KeyValuePair<string, StringValues> _current;
-            private bool _hasUnknown;
-            private Dictionary<string, StringValues>.Enumerator _unknownEnumerator;
+            private readonly bool _hasUnknown;
+            private readonly Dictionary<string, StringValues>.Enumerator _unknownEnumerator;
```
benaadams: Don't make the struct fields `readonly`; dragons.

cesarbs:
Removed `readonly` from `struct` fields.

benaadams:
On further analysis, is only the structs in a struct, where the inner struct has methods or properties called on it that have issues. (So bool is ok to be readonly, but the inner enumerator struct will go weird)

cesarbs:
@benaadams `Microsoft.AspNet.Server.Kestrel.FunctionalTests.ReuseStreamsTests.ReuseStreamsOn` is failing consistenly in this PR. Any idea why that could be? I can't repro on my local Ubuntu VM.

benaadams:
@cesarbs yes, need to move how the test is done https://github.com/aspnet/KestrelHttpServer/pull/501 

Keep-alive is not enforced with the header on mono; so sometimes works, sometimes doesn't. Need to move it to an enginetest rather than how its currently done.

benaadams:
That PR isn't currently correct though.

cesarbs:
@benaadams The real issue is having mutable `readonly` `struct` fields. They are not mutated as expected because when the field is `readonly` a copy is made before invoking one of its members.

benaadams:
Yar:
> is only the structs in a struct, where the inner struct has methods or properties called on it that have issues

Though that sentence doesn't make that much sense reading it back

cesarbs:
:smile:

halter73:
From reading the Jon Skeet's blog, it appears the issue isn't at all related to whether or not the field is *in* a struct. It also doesn't matter whether or not the struct is modified. The compile adds the copies (i.e. `stloc ldloca.s`) because the value in the readonly field *might* be modified and the compiler isn't smart enough to know that it isn't. Note that the `Int256` struct used in Jon Skeet's example is never actually modified.

It seems that if perf matters, it's safest to just never put a struct in a read only field whether or not the field is inside of a struct or class.

benaadams:
https://github.com/dotnet/corefx/pull/4898#discussion_r47368263
> FWIW, that's an oversimplification. The compiler only adds the defensive copy of a readonly struct field in the case of an invocation off of that field, in order to prevent mutation to the thing that was supposed to be readonly.

Though I don't `readonly` structs anymore; is simpler; for everything else it seems to be a good idea.

benaadams:
As the behaviour also easily breaks enumerators and other things, without a compile error; as it allows you to modify them; but then doesn't change them.

halter73:
:shipit:

PR #532, opened 2015-12-26
Faster Take

benaadams:
From #519


src/Microsoft.AspNet.Server.Kestrel/Infrastructure/MemoryPoolIterator2.cs
```diff
@@ -87,7 +84,15 @@ public int Take()
...
                 }
-            }
+
+                if (index < block.End)
+                {
+                    _block = block;
+                    _index = index + 1;
+                    return block.Array[index];
+                }
+
```
cesarbs: nit: remove blank line

benaadams: Removed

cesarbs:
:ship:

PR #533, opened 2015-12-26
Faster Peek

benaadams:
From #519


cesarbs:
:shipit:

cesarbs:
But before I merge, did you run any benchmarks on this change?

benaadams:
Its one of the hot spot changes from the xmas combo #519 (i.e. worked each method until it was no longer a top hotspot then moved onto the next); was planning to benchmark each one in isolation though; just haven't looped back round to it.

PR #537, opened 2015-12-26
Only queue write when not queued

benaadams:
Reduces contention on Thread's work queue

### DIFF
--- a/src/Microsoft.AspNet.Server.Kestrel/Http/SocketOutput.cs
+++ b/src/Microsoft.AspNet.Server.Kestrel/Http/SocketOutput.cs
@@ -289,16 +289,9 @@ namespace Microsoft.AspNet.Server.Kestrel.Http
                 _connection.Abort();
             }
 
-            bool scheduleWrite = false;
-
             lock (_contextLock)
             {
                 PoolWriteContext(writeContext);
-                if (_nextWriteContext != null)
-                {
-                    scheduleWrite = true;
-                    _writePending = true;
-                }
 
                 // _numBytesPreCompleted can temporarily go negative in the event there are
                 // completed writes that we haven't triggered callbacks for yet.
@@ -334,12 +327,6 @@ namespace Microsoft.AspNet.Server.Kestrel.Http
             }
 
             _log.ConnectionWriteCallback(_connectionId, status);
-
-            if (scheduleWrite)
-            {
-                ScheduleWrite();
-            }
-
             _tasksCompleted.Clear();
         }
 
--- a/src/Microsoft.AspNet.Server.Kestrel/Http/SocketOutput.cs
+++ b/src/Microsoft.AspNet.Server.Kestrel/Http/SocketOutput.cs
@@ -16,7 +16,6 @@ namespace Microsoft.AspNet.Server.Kestrel.Http
     {
         public const int MaxPooledWriteReqs = 1024;
 
-        private const int _maxPendingWrites = 3;
         private const int _maxBytesPreCompleted = 65536;
         private const int _initialTaskQueues = 64;
         private const int _maxPooledWriteContexts = 32;
@@ -44,7 +43,7 @@ namespace Microsoft.AspNet.Server.Kestrel.Http
 
         // The number of write operations that have been scheduled so far
         // but have not completed.
-        private int _writesPending = 0;
+        private bool _writePending = false;
         private int _numBytesPreCompleted = 0;
         private Exception _lastWriteError;
         private WriteContext _nextWriteContext;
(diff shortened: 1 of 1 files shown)


src/Microsoft.AspNet.Server.Kestrel/Http/SocketOutput.cs
```diff
@@ -16,11 +16,13 @@ public class SocketOutput : ISocketOutput
     {
         public const int MaxPooledWriteReqs = 1024;
 
-        private const int _maxPendingWrites = 3;
```
halter73: It was @lodejard who initially suggested this pattern. I think that it had something to do about prioritizing future writes even if there was a pending write operation ready to handle the newly requested write.

I'm not sure I fully understood the explanation for having up to 3 pending write requests, because only having one queued write at a time still makes sense to me. I'll talk to him about it.

src/Microsoft.AspNet.Server.Kestrel/Http/SocketOutput.cs
```diff
@@ -312,10 +302,6 @@ private void OnWriteCompleted(WriteContext writeContext)
                 {
                     scheduleWrite = true;
```
halter73: Isn't this now redundant? Are we already guaranteed to have a queued write operation that will handle the `_nextWriteContext` based on the logic in WriteAsync?

src/Microsoft.AspNet.Server.Kestrel/Http/SocketOutput.cs
benaadams: It made more sense before as the `WriteContext` was more stateful with buffers; whereas now it picks up whatever is available at the time of the write.

src/Microsoft.AspNet.Server.Kestrel/Http/SocketOutput.cs
benaadams: Hmm... maybe. 

I also think the `WriteContext` and `SocketOutput` can be merged in someway, as essentially its only unique state is `SocketShutdownSend`+`SocketDisconnect`; but if those are use/triggered when all the blocks are processed rather than depending on the active `WriteContext` they can be moved to properties of `SocketOutput`?

src/Microsoft.AspNet.Server.Kestrel/Http/SocketOutput.cs
```diff
@@ -248,7 +248,10 @@ private static void ReturnBlocks(MemoryPoolBlock2 block)
 
         private void ScheduleWrite()
         {
-            _thread.Post(_this => _this.WriteAllPending(), this);
```
halter73: I don't see the need for `_writeQueueState` and all this Interlocked stuff.

We could simply create a `_writeQueued` bool instead and set it and `scheduleWrite` to true inside of `WriteAsync` with the `_contextLock` taken if `!_writeQueued && immediate`. Then you simply set `_writeQueued` to false inside of `WriteAllPending` where you once again have the `_contextLock`.

src/Microsoft.AspNet.Server.Kestrel/Http/SocketOutput.cs
halter73: There might be a way to do something like that. The key is to always do shutdown and disconnect in the right order and to never do either twice.

benaadams:
@halter73 I think that change would look like this? Is a bit simpler as you said.

Still need `bool scheduleWrite` so the `ScheduleWrite` call is done outside the lock

halter73:
This is great! Not only is it a lot simpler than before, it measurably improves our plaintext benchmark results by a fair amount.

Dev:

	    Latency     2.90ms    5.30ms 262.86ms   96.52%
	    Req/Sec    33.91k     2.58k   52.13k    76.57%
	  10887022 requests in 10.10s, 1.34GB read
	Requests/sec: 1077972.07
	Transfer/sec:    135.70MB

	    Latency     3.16ms    8.40ms 217.11ms   95.44%
	    Req/Sec    33.80k     2.53k   47.66k    70.31%
	  10854987 requests in 10.10s, 1.33GB read
	Requests/sec: 1074838.88
	Transfer/sec:    135.31MB

	    Latency     2.70ms    3.84ms 115.81ms   94.84%
	    Req/Sec    33.65k     2.25k   48.79k    70.22%
	  10806755 requests in 10.10s, 1.33GB read
	Requests/sec: 1069972.56
	Transfer/sec:    134.69MB

benaadams/pendingwrites-merged:

	    Latency     3.66ms   10.46ms 314.37ms   95.67%
	    Req/Sec    35.74k     2.21k   62.72k    88.98%
	  11457533 requests in 10.10s, 1.41GB read
	Requests/sec: 1134506.71
	Transfer/sec:    142.82MB

	    Latency     3.61ms   11.54ms 266.23ms   94.00%
	    Req/Sec    35.84k     2.06k   53.82k    86.93%
	  11492810 requests in 10.10s, 1.41GB read
	Requests/sec: 1137905.25
	Transfer/sec:    143.25MB

	    Latency     3.88ms   12.61ms 276.34ms   94.20%
	    Req/Sec    35.71k     1.69k   51.27k    81.51%
	  11458706 requests in 10.10s, 1.41GB read
	Requests/sec: 1134574.77
	Transfer/sec:    142.83MB

Issue #389, opened 2015-11-15
Add CORS headers to known reponse headers to avoid initialization of unknown headers code path


benaadams:
Particularly `Access-Control-Allow-Origin`

Related:

`Access-Control-Allow-Methods`
`Access-Control-Allow-Headers`
`Access-Control-Expose-Headers`
`Access-Control-Max-Age`

NickCraver:
Are you talking about adding them to the known headers, or actually adding on every response? I'm assuming the former here.

I'm for adding these and additionally pre-computing the bytes for all known header names and only eating the encoding cost of the value, but it's a non-trivial change (it'd be on top of #367, allowing `BlockCopy` usage for all headers). Given that almost every usage of headers in an interned string (very seldom is a *name* generated), we could shave a lot of repeated work off there. Thoughts?

benaadams:
Adding to known headers else added them causes initialization of the `Unknown` dictionary and their key gets converted to string: https://github.com/aspnet/KestrelHttpServer/blob/dev/src/Microsoft.AspNet.Server.Kestrel/Http/FrameHeaders.Generated.cs#L4508

I use them all the time; other than on the entry domain; all other work is done on subdomains and with ajax; images (cors needed for webgl/canvas) or websockets where these headers are added to every request.

benaadams:
@NickCraver just made me think; all unknown response headers keys could be converted to a byte array once and stored in dict for cached conversion of key

NickCraver:
I thought about that too - it could lead to unintended unbounded growth of that structure though. I can't think of a reasonable way around that for the case that *does* generate headers with some unbounded superset. I think we have to add all of the common ones (including these) and optimize that set as best we can and optimize unknown as much as possible *without* caching the result.

benaadams:
Should be ok as its server controlled response headers? If it was request headers I could see an issue as the memory growth would be controlled by an external third party; but server response headers should only be what is coded; and what's necessary for the app?

Unless you have a unique header key per request or similar?

NickCraver:
Yep - the unique-ish per response is the issue. I don't personally have a use case for this, but I've seen it in the wild....regardless of if I strongly disagree with it :) Some people use it for things like `Auth-<userid>: <token>` on proxy requests and such. It's not a good idea, but due to max per-header lengths in various proxies across the internet, I see why they had to get "creative"...

benaadams:
[...]

PR #543, opened 2015-12-30
Add CORS headers to known headers (#389).

cesarbs:
#389 

cc @halter73 @benaadams


halter73:
This seems to address #389 well. Might be worth adding a comment referencing http://www.w3.org/TR/cors/#syntax to show where the list headers come from.

I wonder if there is much of a downside to increasing the number of fields in this class. We could measure, but I doubt it would currently show up in our benchmarks since we reuse the header dictionaries for all requests made over the same connection.

However, if we go back to [resetting each individual field to avoid gen2 promotions](https://github.com/aspnet/KestrelHttpServer/pull/497#issuecomment-165606529) this could have a bigger impact.

cesarbs:
@halter73 I've added comments pointing to the link you suggested.

halter73:
:shipit:

PR #547, opened 2015-12-31
Find first byte fast

Anderman:
Improved FindFirstByte by 300%
tested wirh Coreclr 64 release

### DIFF
--- a/src/Microsoft.AspNet.Server.Kestrel/Infrastructure/MemoryPoolIterator2.cs
+++ b/src/Microsoft.AspNet.Server.Kestrel/Infrastructure/MemoryPoolIterator2.cs
@@ -9,7 +9,7 @@ namespace Microsoft.AspNet.Server.Kestrel.Infrastructure
 {
     public struct MemoryPoolIterator2
     {
-        private readonly static int _vectorSpan = Vector<byte>.Count; 
+        private static readonly int _vectorSpan = Vector<byte>.Count;
 
         private MemoryPoolBlock2 _block;
         private int _index;
@@ -203,7 +203,7 @@ namespace Microsoft.AspNet.Server.Kestrel.Infrastructure
                 long nextLong;
                 fixed (byte* ptr = &_block.Next.Array[_block.Next.Start])
                 {
-                    nextLong = *(long*)(ptr );
+                    nextLong = *(long*)(ptr);
                 }
 
                 return (blockLong >> (sizeof(long) - blockBytes) * 8) | (nextLong << (sizeof(long) - nextBytes) * 8);
@@ -537,7 +537,37 @@ namespace Microsoft.AspNet.Server.Kestrel.Infrastructure
             }
         }
 
-        private static int FindFirstEqualByte(ref Vector<byte> byteEquals)
+        /// <summary>
+        /// Find first byte
+        /// </summary>
+        /// <param  name="byteEquals"></param >
+        /// <returns>The first index of the result vector</returns>
+        /// <exception cref="InvalidOperationException">byteEquals = 0</exception>
+        internal static int FindFirstEqualByte(ref Vector<byte> byteEquals)
+        {
+            if (!BitConverter.IsLittleEndian) return FindFirstEqualByteSlow(ref byteEquals);
+
+            // Quasi-tree search
+            var vector64 = Vector.AsVectorInt64(byteEquals);
+            for (var i = 0; i < Vector<long>.Count; i++)
+            {
+                var longValue = vector64[i];
+                if (longValue == 0) continue;
+
+                return (i << 3) +
+                    ((longValue & 0x00000000ffffffff) > 0
+                        ? (longValue & 0x000000000000ffff) > 0
+                            ? (longValue & 0x00000000000000ff) > 0 ? 0 : 1
+                            : (longValue & 0x0000000000ff0000) > 0 ? 2 : 3
+                        : (longValue & 0x0000ffff00000000) > 0
+                            ? (longValue & 0x000000ff00000000) > 0 ? 4 : 5
+                            : (longValue & 0x00ff000000000000) > 0 ? 6 : 7);
+            }
+            throw new InvalidOperationException();
+        }
+
+        // Internal for testing
+        internal static int FindFirstEqualByteSlow(ref Vector<byte> byteEquals)
         {
             // Quasi-tree search
             var vector64 = Vector.AsVectorInt64(byteEquals);
--- a/test/Microsoft.AspNet.Server.KestrelTests/MemoryPoolIterator2Tests.cs
+++ b/test/Microsoft.AspNet.Server.KestrelTests/MemoryPoolIterator2Tests.cs
@@ -20,6 +20,34 @@ namespace Microsoft.AspNet.Server.KestrelTests
             _pool.Dispose();
         }
 
+        [Fact]
+        public void FindFirstByte()
+        {
+            var bytes = new byte[] {
+                0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff,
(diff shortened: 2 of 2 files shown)


src/Microsoft.AspNet.Server.Kestrel/Infrastructure/MemoryPoolIterator2.cs
```diff
@@ -439,7 +441,35 @@ public int Seek(Vector<byte> byte0Vector, Vector<byte> byte1Vector, Vector<byte>
...
+                    ((longValue & 0x00000000ffffffff) == 0
+                    ? (longValue & 0x0000ffff00000000) == 0
+                    ? (longValue & 0x00ff000000000000) == 0 ? 7 : 6
+                    : (longValue & 0x000000ff00000000) == 0 ? 5 : 4
+                    : (longValue & 0x000000000000ffff) == 0
+                    ? (longValue & 0x0000000000ff0000) == 0 ? 3 : 2
+                    : (longValue & 0x00000000000000ff) == 0 ? 1 : 0);
+            }
+            throw new InvalidOperationException();
+        }
```
halter73: Nit newline.

src/Microsoft.AspNet.Server.Kestrel/Infrastructure/MemoryPoolIterator2.cs
```diff
@@ -439,7 +441,35 @@ public int Seek(Vector<byte> byte0Vector, Vector<byte> byte1Vector, Vector<byte>
...
+        public static int _FindFirstEqualByte(Vector<byte> byteEquals)
```
halter73: Can we make this private an rename it to something like FindFirstEqualByteSlow?

src/Microsoft.AspNet.Server.Kestrel/Infrastructure/MemoryPoolIterator2.cs
```diff
@@ -439,7 +441,35 @@ public int Seek(Vector<byte> byte0Vector, Vector<byte> byte1Vector, Vector<byte>
             }
         }
 
-        private static int FindFirstEqualByte(Vector<byte> byteEquals)
+        /// <summary>
+        /// Find 
+        /// </summary>
+        /// <param name="byteEquals"></param>
+        /// <returns>The first index of the result vector if found otherwise return int.MaxValue</returns>
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
```
halter73: Think this is the first time we've used this attribute in Kestrel. It's probably better to profile and add these attributes all at once if/when we decide to do that.

src/Microsoft.AspNet.Server.Kestrel/Infrastructure/MemoryPoolIterator2.cs
```diff
@@ -347,6 +348,7 @@ public int Seek(Vector<byte> byte0Vector, Vector<byte> byte1Vector, Vector<byte>
                     var following = block.End - index;
                     if (following >= Vector<byte>.Count)
                     {
+
```
halter73: ?

src/Microsoft.AspNet.Server.Kestrel/Infrastructure/MemoryPoolIterator2.cs
```diff
@@ -439,7 +441,35 @@ public int Seek(Vector<byte> byte0Vector, Vector<byte> byte1Vector, Vector<byte>
...
+                if (longValue == 0) continue;
+
+                return (i << 3) +
+                    ((longValue & 0x00000000ffffffff) == 0
+                    ? (longValue & 0x0000ffff00000000) == 0
+                    ? (longValue & 0x00ff000000000000) == 0 ? 7 : 6
+                    : (longValue & 0x000000ff00000000) == 0 ? 5 : 4
+                    : (longValue & 0x000000000000ffff) == 0
+                    ? (longValue & 0x0000000000ff0000) == 0 ? 3 : 2
+                    : (longValue & 0x00000000000000ff) == 0 ? 1 : 0);
```
halter73: Can we make the branching of this expression more clear w/o hurting perf?

src/Microsoft.AspNet.Server.Kestrel/Infrastructure/MemoryPoolIterator2.cs
halter73: Since this is being called from a test, maybe make this internal instead and add a comment that it's accessible for testing.

halter73:
@Anderman ping

src/Microsoft.AspNet.Server.Kestrel/Infrastructure/MemoryPoolIterator2.cs
Anderman: No )-; it would not Inline. So I added the attr. 
A lot of very small functions will inline will inline  
like `public bool IsDefault => _block == null` 

But If  look to the code there are a lot of small functions that can make a big difference
Like peek, peeklong, isend, take. 

They can be split up in a peek and a peeknext block. So the inline code will be small

src/Microsoft.AspNet.Server.Kestrel/Infrastructure/MemoryPoolIterator2.cs
Anderman: Thx, I will update this PR.

test/Microsoft.AspNet.Server.KestrelTests/MemoryPoolIterator2Tests.cs
```diff
@@ -19,6 +19,33 @@ public void Dispose()
         {
             _pool.Dispose();
         }
```
halter73: Nit: add newline

src/Microsoft.AspNet.Server.Kestrel/Infrastructure/MemoryPoolIterator2.cs
halter73: Maybe indentation can help with this if we can't do the same thing with if blocks.

halter73:
Please rebase this so only your commits show up in the diff.

Can you also provide the code you used to test the perf improvements and the numbers you got, so I can compare?

src/Microsoft.AspNet.Server.Kestrel/Infrastructure/MemoryPoolIterator2.cs
```diff
@@ -537,7 +539,38 @@ public unsafe int Seek(ref Vector<byte> byte0Vector, ref Vector<byte> byte1Vecto
...
 
-        private static int FindFirstEqualByte(ref Vector<byte> byteEquals)
+        // Internal for testing
+        /// <summary>
+        /// Find first byte
+        /// </summary>
+        /// <param  name="byteEquals"></param >
+        /// <returns>The first index of the result vector</returns>
+        /// <exception cref="InvalidOperationException">byteEquals = 0</exception>
+        internal static int FindFirstEqualByte(ref Vector<byte> byteEquals)
```
cesarbs: I'm not very fond of the idea of putting "for testing" code outside of test code. Why does this have to here?

halter73: I have not problem with "for testing" comments (I was the one who suggested @Anderman should add it), but I meant it to be specifically for `FindFirstEqualByteSlow`. The point is that that method could/should be private, but we won't get any test coverage unless we call it explicitly unless we have any big endian test agents lying around.

Anderman: I removed the first comment

halter73:
Thanks!

PR #549, opened 2015-12-31
Move generic Get/Set into Interface

benaadams:
Paired change for https://github.com/aspnet/HttpAbstractions/pull/526


src/Microsoft.AspNet.Server.Kestrel/Http/Frame.FeatureCollection.cs
```diff
@@ -280,6 +280,16 @@ bool IHttpUpgradeFeature.IsUpgradableRequest
             set { FastFeatureSet(key, value); }
         }
 
+        public TFeature IFeatureCollection.Get<TFeature>()
```
davidfowl: This doesn't compile, remove the public

davidfowl:
Merged a fix

PR #553, opened 2016-01-05
Fix typo

sergey-tom:
Looks like value from wrong variable is passed to Marshal.FreeHGlobal method. So IntPtr.Zero is passed to the method and actual pointer is lost.


halter73:
Thanks!

PR #554, opened 2016-01-05
Faster common header handling

benaadams:


src/Microsoft.AspNet.Server.Kestrel/Infrastructure/Constants.cs
```diff
@@ -20,5 +24,164 @@ internal class Constants
...
+
+        internal static readonly byte[] HexBytes = Encoding.ASCII.GetBytes("0123456789abcdef");
+
+        internal static readonly byte[] HeaderEndChunkBytes = Encoding.ASCII.GetBytes("\r\n");
+        internal static readonly byte[] HeaderEndChunkedResponseBytes = Encoding.ASCII.GetBytes("0\r\n\r\n");
+        internal static readonly byte[] HeaderContinueBytes = Encoding.ASCII.GetBytes("HTTP/1.1 100 Continue\r\n\r\n");
+        internal static readonly byte[] HeaderEmptyDataBytes = new byte[0];
+
+        // method needed while this bug is active https://github.com/dotnet/coreclr/issues/2526
+        public static void InitalizeJitConstants()
```
halter73: Spelling

benaadams: This word seems to haunt me for some reason

src/Microsoft.AspNet.Server.Kestrel/Infrastructure/Constants.cs
```diff
@@ -20,5 +24,164 @@ internal class Constants
...
+        internal static readonly byte[] HexBytes = Encoding.ASCII.GetBytes("0123456789abcdef");
+
+        internal static readonly byte[] HeaderEndChunkBytes = Encoding.ASCII.GetBytes("\r\n");
+        internal static readonly byte[] HeaderEndChunkedResponseBytes = Encoding.ASCII.GetBytes("0\r\n\r\n");
+        internal static readonly byte[] HeaderContinueBytes = Encoding.ASCII.GetBytes("HTTP/1.1 100 Continue\r\n\r\n");
+        internal static readonly byte[] HeaderEmptyDataBytes = new byte[0];
+
+        // method needed while this bug is active https://github.com/dotnet/coreclr/issues/2526
+        public static void InitializeJitConstants()
+        {
```
halter73: I chatted w/ @jkotas, and he suggested this could be simplified to:

    private static bool initializer;

    public static void InitializeJitConstants()
    {
         initializer = BitConverter.IsLittleEndian;
    }

All static fields of a class get initialized at once.

He also suggested it's better to avoid static variables altogether. This could probably be achieved by making `Constants` non-static and referencing the singleton instance using a property in `ServiceContext`. I think we should probably do this instead, so we don't need to worry about static initialization order.

halter73:
Are the two commits in this PR related? Does one require the other?

benaadams:
Yea, was just looking at that - I think it was because it was related to the earlier commit that was in here; will split again.

benaadams:
Removed other PR

halter73:
Given the extra complexity this adds for the relatively small performance improvement, I'm not very inclined to merge this PR.

halter73:
This is really cool, but a little much for a 1% perf gain.

Issue #555, opened 2016-01-05
Do not use an Action for the awaitable status 


Anderman:
The SocketInput use an action as status. 
Actions are expensive to compare and in interlock statements

This statement is 10 times slower with action 
```
154:             if (awaitableState != _awaitableIsCompleted &&
155:                 awaitableState != _awaitableIsNotCompleted)
000007FCB1296789  mov         rdx,0E883067838h  
000007FCB1296793  mov         rdx,qword ptr [rdx]  
000007FCB1296796  mov         rcx,rsi  
000007FCB1296799  call        000007FD05E74680  
000007FCB129679E  test        eax,eax  
000007FCB12967A0  je          000007FCB12967C6  
000007FCB12967A2  mov         rdx,0E883067840h  
000007FCB12967AC  mov         rdx,qword ptr [rdx]  
000007FCB12967AF  mov         rcx,rsi  
000007FCB12967B2  call        000007FD05E74680  
000007FCB12967B7  test        eax,eax  
000007FCB12967B9  je          000007FCB12967C6  
```

Generated code with int 
``` 
   154:             if (awaitableState2 != _awaitableIsCompleted2 &&
   155:                 awaitableState2 != _awaitableIsNotCompleted2)
000007FCB1296573  cmp         eax,1  
000007FCB1296576  je          000007FCB1296583  
000007FCB1296578  cmp         eax,2  
000007FCB129657B  je          000007FCB1296583  
```

benaadams:
Resolved by https://github.com/aspnet/KestrelHttpServer/pull/558?

PR #556, opened 2016-01-05
Faster Awaitable

Anderman:
See [Compare of actions is slow](https://github.com/aspnet/KestrelHttpServer/issues/555)

### DIFF
--- a/src/Microsoft.AspNet.Server.Kestrel/Http/SocketInput.cs
+++ b/src/Microsoft.AspNet.Server.Kestrel/Http/SocketInput.cs
@@ -160,7 +160,7 @@ namespace Microsoft.AspNet.Server.Kestrel.Http
                 {
                     _manualResetEvent.Reset();
 
-                    var awaitableState = Interlocked.CompareExchange(
+                    Interlocked.CompareExchange(
                         ref _awaitableState,
                         _awaitableIsNotCompleted,
                         _awaitableIsCompleted);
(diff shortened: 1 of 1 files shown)


halter73:
Wouldn't the simpler change be to change `!=` to `!object.ReferenceEquals`?

I'm a little concerned that moving the assignment of the continuation outside of the call to `CompareExchange` might introduce a subtle bug.

Anderman:
Hmm, I don't think so but that looks even better without any risk.

halter73:
Any reason not to just do this instead? https://github.com/aspnet/KestrelHttpServer/pull/558

Anderman:
I still like the version with constants instead of ref to memory (slower)
and I was not completely convinced that this would lead to RC

But this not my daily business so I can be wrong.

The state is still checked / changed with the interlock statement.
The only thing that'is different is the `continuation action`. 

The producer reads only and the consumer does write and read.

So what could go wrong is that the producer reads an incorrect value if this value is set to late

```
var continuationPrev = _continuation;
            _continuation = continuation;

            var awaitableState = Interlocked.CompareExchange(
                ref _awaitableState,
                _awaitableIsContinue,
                _awaitableIsNotCompleted);
```
Maybe this can happen with out of order execution? Or if someone change the code.

halter73:
I haven't looked at it closely enough to determine for sure whether not an RC is there, but I know it's definitely possible to introduce on with this type of change, and I don't want to risk it.

> ref to memory (slower)

Are you sure about that? It's not like we're actually following the reference to check equality if we take my PR.

Anderman:
I understand. Should I close this one?

halter73:
@Anderman I can close this.

Thanks for the PR and filing the issue.

PR #557, opened 2016-01-06
Remove old code

Anderman:
Removed some unused code

### DIFF
--- a/src/Microsoft.AspNet.Server.Kestrel/Http/SocketInput.cs
+++ b/src/Microsoft.AspNet.Server.Kestrel/Http/SocketInput.cs
@@ -16,7 +16,6 @@ namespace Microsoft.AspNet.Server.Kestrel.Http
 
         private readonly MemoryPool2 _memory;
         private readonly IThreadPool _threadPool;
-        private readonly ManualResetEventSlim _manualResetEvent = new ManualResetEventSlim(false);
 
         private Action _awaitableState;
         private Exception _awaitableError;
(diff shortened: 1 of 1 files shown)


src/Microsoft.AspNet.Server.Kestrel/Http/SocketInput.cs
```diff
@@ -17,7 +16,6 @@ public class SocketInput : ICriticalNotifyCompletion
 
         private readonly MemoryPool2 _memory;
         private readonly IThreadPool _threadPool;
-        private readonly ManualResetEventSlim _manualResetEvent = new ManualResetEventSlim(false);
```
halter73: Removing this seems scary. I think this breaks reasonable assumptions about how GetResult() works.

Anderman:
Yeah, little bit scary. But I can't think of any situation when this is needed. And there is no test or comment why this is needed. 

The code is created by @lodejard

halter73:
I'm happy taking the first commit removing uncalled methods, but I don't see a compelling enough reason to remove `_manualResetEvent` from `SocketInput`.

cesarbs:
Same for me.

Anderman:
Ok, do you want me to remove the last commit?

cesarbs:
Yes, please.

halter73:
SocketInput's Skip and Take methods have been removed #525, and I don't want to remove the `_manualResetEvent`, so I'm closing this.

PR #558, opened 2016-01-06
Use object.ReferenceEquals when comparing Actions in SocketInput

halter73:


src/Microsoft.AspNet.Server.Kestrel/Http/SocketInput.cs
```diff
@@ -127,8 +127,8 @@ public void IncomingComplete(int count, Exception error)
                 _manualResetEvent.Set();
             }
 
-            if (awaitableState != _awaitableIsCompleted &&
-                awaitableState != _awaitableIsNotCompleted)
+            if (!object.ReferenceEquals(awaitableState, _awaitableIsCompleted) &&
```
moozzyk: Any reason to specify `object.`? you did not do this above...

halter73: No. I'll remove `object` from these changes.

halter73:
This is meant to address https://github.com/aspnet/KestrelHttpServer/issues/555

src/Microsoft.AspNet.Server.Kestrel/Http/SocketInput.cs
moozzyk: :+1:

cesarbs:
Did you see any perf improvement from this?

halter73:
Frankly, I didn't measure this one. Considering we are doing `Interlocked.CompareExchange` on the `_awaitableState` which is effectively doing a reference comparison too, I felt ReferenceEquals was more correct in addition to generating simpler IL.

cesarbs:
Sounds good. :ship:

cesarbs:
Merge?

Issue #470, opened 2015-12-12
Can OnCompleted callbacks be called even though OnStarting was never called?


tenor:
I'm looking at [FrameOfT#93](https://github.com/aspnet/KestrelHttpServer/blob/dev/src/Microsoft.AspNet.Server.Kestrel/Http/FrameOfT.cs#L93), and imagining the following scenario:

Application throws an exception, 
but hasn't called Response.Write() or Response.Flush(), which would have called FireOnStarting()

FireOnStarting() on line 93 is never called because _applicationException is set.
FireOnCompleted() (line 98) is called.

Is this an acceptable scenario?

tenor:
I can repro this scenario.
Create a new WebAPI application and change the api/values action to this:

```csharp
// GET: api/values
[HttpGet]
public IEnumerable<string> Get()
{
    bool onStartingCalled = false, onCompletedCalled = false;

    this.HttpContext.Response.OnStarting(() =>
    {
        return Task.Run(() =>
        {
            onStartingCalled = true;
        });
    });

    this.HttpContext.Response.OnCompleted(() =>
    {
        return Task.Run(() =>
        {
            System.Diagnostics.Debug.Assert(onStartingCalled);
            onCompletedCalled = true;
        });
    });

    //return new string[] { "value1", "value2" }; //Uncomment this to get the expected behavior

    throw new Exception();
}
```

benaadams:
Don't know, you might have stuff to clean up even if it threw exception and didn't start outputting?

Tratcher:
Yes, this is the behavior we want.

halter73:
@benaadams You are exactly right. We call the `OnCompleted` callback in spite of an `_applicationException`, because we want to allow users to treat it as a `finally` block of sorts. Firing `OnStarting` isn't necessary, because the application never attempted to even start writing a response.

I don't think we have any regression tests to verify we don't change this behavior, so I think we should keep this bug open until we add that.

muratg:
@JunTaoLuo

cesarbs:
Added regression test. Closing.

PR #561, opened 2016-01-08
Add regression test to verify OnComplete is called even when OnStarting is not (#470).

cesarbs:
#470

cc @halter73 @Tratcher

### DIFF
--- a/test/Microsoft.AspNet.Server.Kestrel.FunctionalTests/ResponseTests.cs
+++ b/test/Microsoft.AspNet.Server.Kestrel.FunctionalTests/ResponseTests.cs
@@ -1,8 +1,10 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Net;
 using System.Net.Http;
 using System.Threading.Tasks;
 using Microsoft.AspNet.Builder;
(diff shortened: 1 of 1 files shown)


test/Microsoft.AspNet.Server.Kestrel.FunctionalTests/ResponseTests.cs
```diff
@@ -125,6 +126,48 @@ public class ResponseTests
             }
         }
 
+        [Fact]
+        public async Task OnCompleteCalledEvenWhenOnStartingNotCalled()
+        {
+            var config = new ConfigurationBuilder()
+                .AddInMemoryCollection(new Dictionary<string, string>
+                {
+                    { "server.urls", "http://localhost:8793/" }
```
halter73: Not sure how much this really matters but the above tests were incrementing port numbers by one, so this one would use 8794 to follow the pattern. Might as well be consistent.

test/Microsoft.AspNet.Server.Kestrel.FunctionalTests/ResponseTests.cs
```diff
@@ -125,6 +126,48 @@ public class ResponseTests
...
+                        // Prevent OnStarting call (see Frame<T>.RequestProcessingAsync()).
+                        throw new Exception();
+                    });
+                });
+
+            using (var app = hostBuilder.Build())
+            {
+                app.Start();
+
+                using (var client = new HttpClient())
```
halter73: Might as well verify there is a 500 status code.

halter73:
:shipit:

PR #562, opened 2016-01-08
Make all write operations atomic

halter73:
This PR ensures that simultaneous calls to `SocketOutput.WriteAsync` won't result in a corrupted write.

This was mainly achieved by moving the `CopyFrom` call inside of `SocketOutput.WriteAsync` inside of the `_contenxtLock`. I also moved chunking down a layer inside of `SocketOutput`, so it could be protected by the same lock.

I tested this change using both the plaintext benchmark and a modified LargeResponseSample that used small chunking, and in both cases the performance was nearly identical <1% to dev.

When I first tried only moving `CopyFrom` inside of the lock, I saw performance degradation of a few percent. I was able to remedy this by using `Monitor.TryEnter` to avoid stalling libuv threads.

### DIFF
--- a/src/Microsoft.AspNet.Server.Kestrel/Filter/StreamSocketOutput.cs
+++ b/src/Microsoft.AspNet.Server.Kestrel/Filter/StreamSocketOutput.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.IO;
+using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.AspNet.Server.Kestrel.Http;
@@ -12,11 +13,14 @@ namespace Microsoft.AspNet.Server.Kestrel.Filter
 {
     public class StreamSocketOutput : ISocketOutput
     {
+        private static readonly byte[] _endChunkBytes = Encoding.ASCII.GetBytes("\r\n");
         private static readonly byte[] _nullBuffer = new byte[0];
 
         private readonly Stream _outputStream;
         private readonly MemoryPool2 _memory;
         private MemoryPoolBlock2 _producingBlock;
+
+        private object _writeLock = new object();
 
         public StreamSocketOutput(Stream outputStream, MemoryPool2 memory)
         {
@@ -24,15 +28,29 @@ namespace Microsoft.AspNet.Server.Kestrel.Filter
             _memory = memory;
         }
 
-        void ISocketOutput.Write(ArraySegment<byte> buffer, bool immediate)
+        public void Write(ArraySegment<byte> buffer, bool immediate, bool chunk)
         {
-            _outputStream.Write(buffer.Array ?? _nullBuffer, buffer.Offset, buffer.Count);
+            lock (_writeLock)
+            {
+                if (chunk && buffer.Array != null)
+                {
+                    var beginChunkBytes = ChunkWriter.BeginChunkBytes(buffer.Count);
+                    _outputStream.Write(beginChunkBytes.Array, beginChunkBytes.Offset, beginChunkBytes.Count);
+                }
+
+                _outputStream.Write(buffer.Array ?? _nullBuffer, buffer.Offset, buffer.Count);
+
+                if (chunk && buffer.Array != null)
+                {
+                    _outputStream.Write(_endChunkBytes, 0, _endChunkBytes.Length);
+                }
+            }
         }
 
-        Task ISocketOutput.WriteAsync(ArraySegment<byte> buffer, bool immediate, CancellationToken cancellationToken)
+        public Task WriteAsync(ArraySegment<byte> buffer, bool immediate, bool chunk, CancellationToken cancellationToken)
         {
             // TODO: Use _outputStream.WriteAsync
-            _outputStream.Write(buffer.Array ?? _nullBuffer, buffer.Offset, buffer.Count);
+            Write(buffer, immediate, chunk);
             return TaskUtilities.CompletedTask;
         }
 
--- a/src/Microsoft.AspNet.Server.Kestrel/Http/ChunkWriter.cs
+++ b/src/Microsoft.AspNet.Server.Kestrel/Http/ChunkWriter.cs
@@ -1,0 +1,62 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Text;
+using Microsoft.AspNet.Server.Kestrel.Infrastructure;
+
+namespace Microsoft.AspNet.Server.Kestrel.Http
+{
+    public static class ChunkWriter
+    {
+        private static readonly ArraySegment<byte> _endChunkBytes = CreateAsciiByteArraySegment("\r\n");
+        private static readonly byte[] _hex = Encoding.ASCII.GetBytes("0123456789abcdef");
+
+        private static ArraySegment<byte> CreateAsciiByteArraySegment(string text)
+        {
+            var bytes = Encoding.ASCII.GetBytes(text);
+            return new ArraySegment<byte>(bytes);
+        }
+
+        public static ArraySegment<byte> BeginChunkBytes(int dataCount)
+        {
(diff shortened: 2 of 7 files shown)


src/Microsoft.AspNet.Server.Kestrel/Filter/StreamSocketOutput.cs
```diff
@@ -12,27 +13,44 @@ namespace Microsoft.AspNet.Server.Kestrel.Filter
...
+                {
+                    _outputStream.Write(_endChunkBytes, 0, _endChunkBytes.Length);
+                }
+            }
         }
 
-        Task ISocketOutput.WriteAsync(ArraySegment<byte> buffer, bool immediate, CancellationToken cancellationToken)
+        public Task WriteAsync(ArraySegment<byte> buffer, bool immediate, bool chunk, CancellationToken cancellationToken)
         {
             // TODO: Use _outputStream.WriteAsync
```
cesarbs: Remove TODO?

halter73: This is still a todo though. Someday...

src/Microsoft.AspNet.Server.Kestrel/Http/ChunkWriter.cs
```diff
@@ -0,0 +1,62 @@
...
+            };
+
+            // Determine the most-significant non-zero nibble
+            int total, shift;
+            total = (dataCount > 0xffff) ? 0x10 : 0x00;
+            dataCount >>= total;
+            shift = (dataCount > 0x00ff) ? 0x08 : 0x00;
+            dataCount >>= shift;
+            total |= shift;
+            total |= (dataCount > 0x000f) ? 0x04 : 0x00;
```
cesarbs: I like this sorcery :+1:

src/Microsoft.AspNet.Server.Kestrel/Filter/StreamSocketOutput.cs
cesarbs: What's preventing us from addressing it?

src/Microsoft.AspNet.Server.Kestrel/Http/ChunkWriter.cs
cesarbs: Oh now I see it's just copied from another class :P

halter73: This was simply moved from Frame. Credit @lodejard.

src/Microsoft.AspNet.Server.Kestrel/Filter/StreamSocketOutput.cs
halter73: Debugging. All I know is calling `WriteAsync` on `SslStream` didn't work when last I tried it. I'm think the call never completed, but I'm not 100% on that. You are definitely right that I should have left a better comment when I first wrote this code.

benaadams:
I don't think this is right.

If you write to any stream in `System.IO` concurrently, you will mashup your data and there is no synchronization applied (e.g. MemoryStream https://github.com/dotnet/coreclr/blob/master/src/mscorlib/src/System/IO/MemoryStream.cs#L540-L613); its up the caller to synchronize their writes - which is normally automatic by waiting for one to complete before doing another.

Also this is making `SocketOutput` thread-safe for multiple writers but the `ISocketOutput` presentation is not thread-safe depending on what the filters are doing and not pushing it up to the caller to do the synchronization means the order of events that the filter creates and what the socket does may be mis-matched (e.g. if a filter logs, writes, logs that may cause the socket output to interleave differently to what is logged depending on timings)

benaadams:
Also a `Stream.CopyTo` does it in chunks so if that was used the atomicity of the `SocketOutput.Write` won't help.

src/Microsoft.AspNet.Server.Kestrel/Http/SocketOutput.cs
```diff
@@ -253,9 +265,9 @@ private void ScheduleWrite()
         // This is called on the libuv event loop
         private void WriteAllPending()
         {
-            WriteContext writingContext;
+            WriteContext writingContext = null;
 
-            lock (_contextLock)
+            if (Monitor.TryEnter(_contextLock))
```
cesarbs: Why?

benaadams:
However... SocketOutput does do leasing blocks and whatnot; so this change might be sensible to protect itself, rather than necessarily providing upstream benefit (or throwing a don't do concurrent writes execption?)

src/Microsoft.AspNet.Server.Kestrel/Http/SocketOutput.cs
halter73: We don't want to block libuv threads at all if we can avoid it.

benaadams:
The question is what's better?
```
var t1 = file1.CopyToAsync(Body.Response);
var t2 = file2.CopyToAsync(Body.Response);
await Task.WhenAll(t1, t2);
```
Current - unknown output, unknown state (I'd suggest this is bad)
This change - right byte count output; unknown interleaving, good state
Alternative - exception thrown to user code, good state

halter73:
@benaadams I was/am also a big proponent for throwing much like we do for ReadAsync.

The filter point you made is interesting, but currently mitigated by the lock I added to SocketOutputStream. The point you make about CopyTo is also a good one, you'll see similar issues if you use JsonSerializer to write to the response streams.

Still, my experience with SignalR has shown me that people *will* try to write concurrently, and won't see that it's a problem until their app is under stress even if we do throw. And in this case, I think it's important that our internal data structures are in a good sate (i.e. our linked blocks), especially if we can ensure this without much of a perf hit.

halter73:
> Alternative - exception thrown to user code, good state

I like this to. It could very well be achieved with a `Interlocked.CompareExchange`. I don't think it will have a real impact on perf, but it's nice to tell users that what they're doing is bad. I'll talk to @lodejard about this tomorrow.

benaadams:
Doing a guard may not cause a throw, to it could still do interleaving as per moving write into lock; but it might be good to have it as an api that *can* throw on concurrent (public api-wise)? A "be careful" modifier :)

src/Microsoft.AspNet.Server.Kestrel/Filter/StreamSocketOutput.cs
benaadams: Can you retest post https://github.com/aspnet/KestrelHttpServer/pull/567 ? Ran into this issue with Ssl WriteAsync as its sync https://github.com/aspnet/KestrelHttpServer/issues/565

On the upside; I think I fixed the issue :)

benaadams: K, so tests fail changing this to async. Will dig in and do a different PR if I get a resolution.

halter73: @benaadams Thanks for looking into this.

halter73:
@benaadams I don't think we should remove the other changes, just because we decide to throw instead of expand the lock. I added another commit to the end of this PR that throws instead. What do you think?

benaadams:
Yeh, think that's good - especially since throwing wasn't the cure all I thought it was going to be :(

halter73:
So, after further discussion with @lodejard and @cesarbs, I've flip-flopped back to not throwing for concurrent writes. The argument that won the day was that when the would-be-exception was thrown under load it would still only be one of two failure modes. There would still be a potential for out-of-order writes assuming order is actually important, but there is also a small chance you get an exception thrown as well. Also, unlike with concurrent reads, it's plausible that your app could work fine depending on the design with concurrent writes. The example given was logging, and how things like `Console.Out` don't throw for concurrent writes.

If we change our mind (and that's definitely possible), it should be relatively simple to change it to make it throw again.

benaadams:
:+1: makes sense

PR #567, opened 2016-01-10
SocketOutput exception completes all waiting tasks

### LINKED ISSUE #450: Kestrel aborts with dropped client connections
I'm trying to harden a service using kestrel and found a problem with dropped HTTP client connections.  If the HTTP client drops the connection while the application is writing to the stream, kestrel doesn't seem to detect this very well.  Even if the application code catches the write errors and the HTTP context is aborted, eventually the kestrel server will just exit with "Aborted."  This is relatively easy to exploit just by killing a lot of HTTP connections while a write is occurring.

I first tried this in coreclr-beta8, but it continues to be an issue in rc1 and rc1-update1.  It is relatively simple to reproduce the issue - I have a small application that runs using the bits on the microsoft/aspnet:1.0.0-rc1-update1-coreclr docker image.  The server code and bash script to reproduce the issue can be found in my [hardening-kestrel repo] (https://github.com/ninjarobot/hardening-kestrel).

### LINKED ISSUE #565: Doing synchronous writes while chunking can seems to block the thread indefinitely
Doing some testing on aspnet/Mvc#3516 I ran into this issue.

It's super easy to overwhelm the server when a large Razor page does synchronous writes. Doing 50 requests per second quickly starves the thread pool. 

I ended up with about 140 threads blocked on this callstack:
```
>	mscorlib.dll!System.Threading.Monitor.Wait(object obj, int millisecondsTimeout, bool exitContext) Line 203	C#
 	mscorlib.dll!System.Threading.Monitor.Wait(object obj, int millisecondsTimeout) Line 213	C#
 	mscorlib.dll!System.Threading.ManualResetEventSlim.Wait(int millisecondsTimeout, System.Threading.CancellationToken cancellationToken) Line 688	C#
 	mscorlib.dll!System.Threading.Tasks.Task.SpinThenBlockingWait(int millisecondsTimeout, System.Threading.CancellationToken cancellationToken) Line 3320	C#
 	mscorlib.dll!System.Threading.Tasks.Task.InternalWait(int millisecondsTimeout, System.Threading.CancellationToken cancellationToken) Line 3252	C#
 	mscorlib.dll!System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(System.Threading.Tasks.Task task) Line 148	C#
 	Microsoft.AspNet.Server.Kestrel.dll!Microsoft.AspNet.Server.Kestrel.Http.SocketOutput.Microsoft.AspNet.Server.Kestrel.Http.ISocketOutput.Write(System.ArraySegment<byte> buffer, bool immediate)	Unknown
 	Microsoft.AspNet.Server.Kestrel.dll!Microsoft.AspNet.Server.Kestrel.Http.Frame.WriteChunked(System.ArraySegment<byte> data)	Unknown
 	Microsoft.AspNet.Server.Kestrel.dll!Microsoft.AspNet.Server.Kestrel.Http.Frame.Write(System.ArraySegment<byte> data)	Unknown
 	Microsoft.AspNet.Server.Kestrel.dll!Microsoft.AspNet.Server.Kestrel.Http.FrameResponseStream.Write(byte[] buffer, int offset, int count)	Unknown
 	Microsoft.AspNet.Mvc.Core!Microsoft.AspNet.Mvc.HttpResponseStreamWriter.FlushInternal() Line 334	C#
 	Microsoft.AspNet.Mvc.Core!Microsoft.AspNet.Mvc.HttpResponseStreamWriter.Write(string value) Line 194	C#
 	Microsoft.AspNet.Mvc.ViewFeatures!Microsoft.AspNet.Mvc.ViewFeatures.Buffer.ViewBufferValue.WriteTo(System.IO.TextWriter writer, System.Text.Encodings.Web.HtmlEncoder encoder) Line 55	C#
 	Microsoft.AspNet.Mvc.ViewFeatures!Microsoft.AspNet.Mvc.ViewFeatures.Buffer.ViewBuffer.WriteTo(System.IO.TextWriter writer, System.Text.Encodings.Web.HtmlEncoder encoder) Line 145	C#
 	Microsoft.AspNet.Mvc.Razor!Microsoft.AspNet.Mvc.Razor.RazorView.RenderLayoutAsync(Microsoft.AspNet.Mvc.Rendering.ViewContext context, Microsoft.AspNet.Mvc.Razor.RazorTextWriter bodyWriter) Line 236	C#
 	Microsoft.AspNet.Mvc.Razor!Microsoft.AspNet.Mvc.Razor.RazorView.RenderAsync(Microsoft.AspNet.Mvc.Rendering.ViewContext context) Line 102	C#
 	Microsoft.AspNet.Mvc.ViewFeatures!Microsoft.AspNet.Mvc.ViewFeatures.ViewExecutor.ExecuteAsync(Microsoft.AspNet.Mvc.ActionContext actionContext, Microsoft.AspNet.Mvc.ViewEngines.IView view, Microsoft.AspNet.Mvc.ViewFeatures.ViewDataDictionary viewData, Microsoft.AspNet.Mvc.ViewFeatures.ITempDataDictionary tempData, Microsoft.Net.Http.Headers.MediaTypeHeaderValue contentType, int? statusCode) Line 180	C#
 	Microsoft.AspNet.Mvc.ViewFeatures!Microsoft.AspNet.Mvc.ViewFeatures.ViewResultExecutor.ExecuteAsync(Microsoft.AspNet.Mvc.ActionContext actionContext, Microsoft.AspNet.Mvc.ViewEngines.IView view, Microsoft.AspNet.Mvc.ViewResult viewResult) Line 165	C#
 	Microsoft.AspNet.Mvc.ViewFeatures!Microsoft.AspNet.Mvc.ViewResult.ExecuteResultAsync(Microsoft.AspNet.Mvc.ActionContext context) Line 75	C#
 	Microsoft.AspNet.Mvc.Core!Microsoft.AspNet.Mvc.Controllers.FilterActionInvoker.InvokeResultAsync(Microsoft.AspNet.Mvc.IActionResult result) Line 884	C#
 	Microsoft.AspNet.Mvc.Core!Microsoft.AspNet.Mvc.Controllers.FilterActionInvoker.InvokeResultFilterAsync() Line 853	C#
 	Microsoft.AspNet.Mvc.Core!Microsoft.AspNet.Mvc.Controllers.FilterActionInvoker.InvokeResultFilterAsync() Line 835	C#
 	Microsoft.AspNet.Mvc.Core!Microsoft.AspNet.Mvc.Controllers.FilterActionInvoker.InvokeAllResultFiltersAsync(Microsoft.AspNet.Mvc.IActionResult result) Line 738	C#
 	Microsoft.AspNet.Mvc.Core!Microsoft.AspNet.Mvc.Controllers.FilterActionInvoker.InvokeResourceFilterAsync() Line 466	C#
 	Microsoft.AspNet.Mvc.Core!Microsoft.AspNet.Mvc.Controllers.FilterActionInvoker.InvokeResourceFilterAsync() Line 399	C#
 	Microsoft.AspNet.Mvc.Core!Microsoft.AspNet.Mvc.Controllers.FilterActionInvoker.InvokeAllResourceFiltersAsync() Line 313	C#
 	Microsoft.AspNet.Mvc.Core!Microsoft.AspNet.Mvc.Controllers.FilterActionInvoker.InvokeAsync() Line 156	C#
 	Microsoft.AspNet.Mvc.Core!Microsoft.AspNet.Mvc.Infrastructure.MvcRouteHandler.InvokeActionAsync(Microsoft.AspNet.Http.HttpContext httpContext, Microsoft.AspNet.Mvc.Abstractions.ActionDescriptor actionDescriptor) Line 102	C#
 	Microsoft.AspNet.Mvc.Core!Microsoft.AspNet.Mvc.Infrastructure.MvcRouteHandler.RouteAsync.AnonymousMethod__0(Microsoft.AspNet.Http.HttpContext c) Line 71	C#
 	Microsoft.AspNet.Routing.dll!Microsoft.AspNet.Builder.RouterMiddleware.Invoke(Microsoft.AspNet.Http.HttpContext httpContext) Line 48	C#
 	Microsoft.AspNet.Diagnostics.dll!Microsoft.AspNet.Diagnostics.ExceptionHandlerMiddleware.Invoke(Microsoft.AspNet.Http.HttpContext context) Line 44	C#
 	Microsoft.AspNet.Hosting.dll!Microsoft.AspNet.Hosting.Internal.RequestServicesContainerMiddleware.Invoke(Microsoft.AspNet.Http.HttpContext httpContext) Line 55	C#
 	Microsoft.AspNet.Hosting.dll!Microsoft.AspNet.Hosting.Internal.HostingApplication.ProcessRequestAsync(Microsoft.AspNet.Hosting.Internal.HostingApplication.Context context) Line 81	C#
 	Microsoft.AspNet.Server.Kestrel.dll!Microsoft.AspNet.Server.Kestrel.Http.Frame<Microsoft.AspNet.Hosting.Internal.HostingApplication.Context>.RequestProcessingAsync()	Unknown
 	mscorlib.dll!System.Runtime.CompilerServices.AsyncTaskMethodBuilder.Start<Microsoft.AspNet.Server.Kestrel.Http.Frame<Microsoft.AspNet.Hosting.Internal.HostingApplication.Context>.<RequestProcessingAsync>d__3>(ref Microsoft.AspNet.Server.Kestrel.Http.Frame<Microsoft.AspNet.Hosting.Internal.HostingApplication.Context>.<RequestProcessingAsync>d__3 stateMachine) Line 316	C#
 	Microsoft.AspNet.Server.Kestrel.dll!Microsoft.AspNet.Server.Kestrel.Http.Frame<Microsoft.AspNet.Hosting.Internal.HostingApplication.Context>.RequestProcessingAsync()	Unknown
 	Microsoft.AspNet.Server.Kestrel.dll!Microsoft.AspNet.Server.Kestrel.Http.Frame.Start.AnonymousMethod__111_0(object o)	Unknown
 	mscorlib.dll!System.Threading.Tasks.Task<System.Threading.Tasks.Task>.InnerInvoke() Line 686	C#
 	mscorlib.dll!System.Threading.Tasks.Task.Execute() Line 2498	C#
 	mscorlib.dll!System.Threading.Tasks.Task.ExecutionContextCallback(object obj) Line 2861	C#
 	mscorlib.dll!System.Threading.ExecutionContext.RunInternal(System.Threading.ExecutionContext executionContext, System.Threading.ContextCallback callback, object state, bool preserveSyncCtx) Line 954	C#
 	mscorlib.dll!System.Threading.ExecutionContext.Run(System.Threading.ExecutionContext executionContext, System.Threading.ContextCallback callback, object state, bool preserveSyncCtx) Line 902	C#
 	mscorlib.dll!System.Threading.Tasks.Task.ExecuteWithThreadLocal(ref System.Threading.Tasks.Task currentTaskSlot) Line 2827	C#
 	mscorlib.dll!System.Threading.Tasks.Task.ExecuteEntry(bool bPreventDoubleExecution) Line 2767	C#
 	mscorlib.dll!System.Threading.Tasks.Task.System.Threading.IThreadPoolWorkItem.ExecuteWorkItem() Line 2704	C#
 	mscorlib.dll!System.Threading.ThreadPoolWorkQueue.Dispatch() Line 820	C#
 	mscorlib.dll!System.Threading._ThreadPoolWaitCallback.PerformWaitCallback() Line 1161	C#
```

It seems like these threads don't make any progress once they get into this state, and the server is unresponsive. More threads are slowly spawned to handle the queued requests until the server OOMs.

I suspect the tasks created by `Write` aren't completed once the client is gone.

### LINKED ISSUE #566: Consider behavior when a client disconnects
Currently today it's *possible* that you'll get I/O exceptions due to a client disconnect, but it's pretty hard to reason about exactly when that will happen and when it won't due to buffering. 

These exceptions can be troublesome in production because you didn't do anything wrong to cause them, and can't do anything to handle or prevent them. Now they are popping up in logs, and you have to try to reason about what's an actual bug vs an I/O exception due to disconnection.

------------

The proposal is that we provide a cancellation token for client disconnection that you can pass if you want to know about cancellation, and that the server by default would just ignore your output once a socket is closed.

Rough idea is that this token would be part of the connection feature, and not surfaced through `HttpContext` because it's so advanced.

benaadams:
Lightens #565
Resolves #450 
Resolves #566

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Kestrel/Http/Connection.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/Http/Connection.cs
@@ -109,7 +109,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Http
             }
         }
 
-        public void Abort()
+        public virtual void Abort()
         {
             if (_frame != null)
             {
--- a/src/Microsoft.AspNetCore.Server.Kestrel/Http/Frame.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/Http/Frame.cs
@@ -468,12 +468,12 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Http
 
         private void WriteChunked(ArraySegment<byte> data)
         {
-            SocketOutput.Write(data, immediate: false, chunk: true);
+            SocketOutput.Write(data, immediate: true, chunk: true);
         }
 
         private Task WriteChunkedAsync(ArraySegment<byte> data, CancellationToken cancellationToken)
         {
-            return SocketOutput.WriteAsync(data, immediate: false, chunk: true, cancellationToken: cancellationToken);
+            return SocketOutput.WriteAsync(data, immediate: true, chunk: true, cancellationToken: cancellationToken);
         }
 
         private Task WriteChunkedResponseSuffix()
--- a/src/Microsoft.AspNetCore.Server.Kestrel/Http/Frame.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/Http/Frame.cs
@@ -56,7 +56,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Http
         private bool _requestProcessingStarted;
         private Task _requestProcessingTask;
         protected volatile bool _requestProcessingStopping; // volatile, see: https://msdn.microsoft.com/en-us/library/x13ttww7.aspx
-        protected volatile bool _requestAborted;
+        protected int _requestAborted;
         protected CancellationTokenSource _abortedCts;
         protected CancellationToken? _manuallySetRequestAbortToken;
 
@@ -167,7 +167,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Http
                 var cts = _abortedCts;
                 return
                     cts != null ? cts.Token :
-                    _requestAborted ? new CancellationToken(true) :
+                    (Volatile.Read(ref _requestAborted) == 1) ? new CancellationToken(true) :
                     RequestAbortedSource.Token;
             }
             set
@@ -185,7 +185,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Http
                 // Get the abort token, lazily-initializing it if necessary.
                 // Make sure it's canceled if an abort request already came in.
                 var cts = LazyInitializer.EnsureInitialized(ref _abortedCts, () => new CancellationTokenSource());
-                if (_requestAborted)
+                if (Volatile.Read(ref _requestAborted) == 1)
                 {
                     cts.Cancel();
                 }
@@ -288,24 +288,31 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Http
         /// </summary>
         public void Abort()
         {
-            _requestProcessingStopping = true;
-            _requestAborted = true;
+            if (Interlocked.CompareExchange(ref _requestAborted, 1, 0) == 0)
+            {
+                _requestProcessingStopping = true;
 
-            _requestBody?.Abort();
-            _responseBody?.Abort();
+                _requestBody?.Abort();
+                _responseBody?.Abort();
 
-            try
-            {
-                ConnectionControl.End(ProduceEndType.SocketDisconnect);
-                SocketInput.AbortAwaiting();
-                RequestAbortedSource.Cancel();
-            }
-            catch (Exception ex)
-            {
-                Log.LogError("Abort", ex);
-            }
-            finally
-            {
+                try
+                {
+                    ConnectionControl.End(ProduceEndType.SocketDisconnect);
+                    SocketInput.AbortAwaiting();
+                }
+                catch (Exception ex)
+                {
+                    Log.LogError("Abort", ex);
+                }
+
+                try
+                {
+                    RequestAbortedSource.Cancel();
+                }
+                catch (Exception ex)
+                {
+                    Log.LogError("Abort", ex);
+                }
                 _abortedCts = null;
             }
         }
--- a/src/Microsoft.AspNetCore.Server.Kestrel/Http/FrameOfT.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/Http/FrameOfT.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Net;
+using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Hosting.Server;
 using Microsoft.AspNetCore.Http.Features;
@@ -111,7 +112,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Http
                             _application.DisposeContext(context, _applicationException);
 
                             // If _requestAbort is set, the connection has already been closed.
-                            if (!_requestAborted)
+                            if (Volatile.Read(ref _requestAborted) == 0)
                             {
                                 _responseBody.ResumeAcceptingWrites();
                                 await ProduceEnd();
@@ -148,7 +149,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Http
                     _abortedCts = null;
 
                     // If _requestAborted is set, the connection has already been closed.
-                    if (!_requestAborted)
+                    if (Volatile.Read(ref _requestAborted) == 0)
                     {
                         // Inform client no more data will ever arrive
                         ConnectionControl.End(ProduceEndType.SocketShutdownSend);
--- a/src/Microsoft.AspNetCore.Server.Kestrel/Http/FrameRequestStream.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/Http/FrameRequestStream.cs
@@ -5,6 +5,7 @@ using System;
 using System.IO;
 using System.Threading;
 using System.Threading.Tasks;
+using Microsoft.AspNet.Server.Kestrel.Infrastructure;
 
 namespace Microsoft.AspNetCore.Server.Kestrel.Http
 {
@@ -51,8 +52,6 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Http
 
         public override int Read(byte[] buffer, int offset, int count)
         {
-            ValidateState();
-
             // ValueTask uses .GetAwaiter().GetResult() if necessary
             return ReadAsync(buffer, offset, count).Result;
         }
@@ -60,7 +59,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Http
 #if NET451
         public override IAsyncResult BeginRead(byte[] buffer, int offset, int count, AsyncCallback callback, object state)
         {
-            ValidateState();
+            ValidateState(CancellationToken.None);
 
             var task = ReadAsync(buffer, offset, count, CancellationToken.None, state);
             if (callback != null)
@@ -77,7 +76,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Http
 
         private Task<int> ReadAsync(byte[] buffer, int offset, int count, CancellationToken cancellationToken, object state)
         {
-            ValidateState();
+            ValidateState(cancellationToken);
 
             var tcs = new TaskCompletionSource<int>(state);
             var task = _body.ReadAsync(new ArraySegment<byte>(buffer, offset, count), cancellationToken);
@@ -103,10 +102,13 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Http
 
         public override Task<int> ReadAsync(byte[] buffer, int offset, int count, CancellationToken cancellationToken)
         {
-            ValidateState();
-
-            // Needs .AsTask to match Stream's Async method return types
-            return _body.ReadAsync(new ArraySegment<byte>(buffer, offset, count), cancellationToken).AsTask();
+            var task = ValidateState(cancellationToken);
+            if (task == null)
+            {
+                // Needs .AsTask to match Stream's Async method return types
+                return _body.ReadAsync(new ArraySegment<byte>(buffer, offset, count), cancellationToken).AsTask();
+            }
+            return task;
         }
 
         public override void Write(byte[] buffer, int offset, int count)
@@ -149,24 +151,29 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Http
         public void Abort()
         {
             // We don't want to throw an ODE until the app func actually completes.
-            // If the request is aborted, we throw an IOException instead.
+            // If the request is aborted, we throw an TaskCanceledException instead.
             if (_state != FrameStreamState.Closed)
             {
                 _state = FrameStreamState.Aborted;
             }
         }
 
-        private void ValidateState()
+        private Task<int> ValidateState(CancellationToken cancellationToken)
         {
             switch (_state)
             {
                 case FrameStreamState.Open:
-                    return;
+                    if (cancellationToken.IsCancellationRequested)
+                    {
+                        return TaskUtilities.GetCancelledZeroTask();
+                    }
+                    break;
                 case FrameStreamState.Closed:
                     throw new ObjectDisposedException(nameof(FrameRequestStream));
                 case FrameStreamState.Aborted:
-                    throw new IOException("The request has been aborted.");
+                    return TaskUtilities.GetCancelledZeroTask();
             }
+            return null;
         }
     }
 }
--- a/src/Microsoft.AspNetCore.Server.Kestrel/Http/FrameResponseStream.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/Http/FrameResponseStream.cs
@@ -5,6 +5,7 @@ using System;
 using System.IO;
 using System.Threading;
 using System.Threading.Tasks;
+using Microsoft.AspNet.Server.Kestrel.Infrastructure;
 
 namespace Microsoft.AspNetCore.Server.Kestrel.Http
 {
@@ -37,16 +38,19 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Http
 
         public override void Flush()
         {
-            ValidateState();
+            ValidateState(CancellationToken.None);
 
             _context.FrameControl.Flush();
         }
 
         public override Task FlushAsync(CancellationToken cancellationToken)
         {
-            ValidateState();
-
-            return _context.FrameControl.FlushAsync(cancellationToken);
+            var task = ValidateState(cancellationToken);
+            if (task == null)
+            {
+                return _context.FrameControl.FlushAsync(cancellationToken);
+            }
+            return task;
         }
 
         public override long Seek(long offset, SeekOrigin origin)
@@ -66,16 +70,19 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Http
 
         public override void Write(byte[] buffer, int offset, int count)
         {
-            ValidateState();
+            ValidateState(CancellationToken.None);
 
             _context.FrameControl.Write(new ArraySegment<byte>(buffer, offset, count));
         }
 
         public override Task WriteAsync(byte[] buffer, int offset, int count, CancellationToken cancellationToken)
         {
-            ValidateState();
-
-            return _context.FrameControl.WriteAsync(new ArraySegment<byte>(buffer, offset, count), cancellationToken);
+            var task = ValidateState(cancellationToken);
+            if (task == null)
+            {
+                return _context.FrameControl.WriteAsync(new ArraySegment<byte>(buffer, offset, count), cancellationToken);
+            }
+            return task;
         }
 
         public Stream StartAcceptingWrites()
@@ -112,24 +119,36 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Http
         public void Abort()
         {
             // We don't want to throw an ODE until the app func actually completes.
-            // If the request is aborted, we throw an IOException instead.
             if (_state != FrameStreamState.Closed)
             {
                 _state = FrameStreamState.Aborted;
             }
         }
 
-        private void ValidateState()
+        private Task ValidateState(CancellationToken cancellationToken)
         {
             switch (_state)
             {
                 case FrameStreamState.Open:
-                    return;
+                    if (cancellationToken.IsCancellationRequested)
+                    {
+                        return TaskUtilities.GetCancelledTask(cancellationToken);
+                    }
+                    break;
                 case FrameStreamState.Closed:
                     throw new ObjectDisposedException(nameof(FrameResponseStream));
                 case FrameStreamState.Aborted:
-                    throw new IOException("The request has been aborted.");
+                    if (cancellationToken.CanBeCanceled)
+                    {
+                        // Aborted state only throws on write if cancellationToken requests it
+                        return TaskUtilities.GetCancelledTask(
+                            cancellationToken.IsCancellationRequested ? 
+                            cancellationToken : 
+                            new CancellationToken(true));
+                    }
+                    break;
             }
+            return null;
         }
     }
 }
--- a/src/Microsoft.AspNetCore.Server.Kestrel/Http/FrameResponseStream.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/Http/FrameResponseStream.cs
@@ -138,13 +138,10 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Http
                 case FrameStreamState.Closed:
                     throw new ObjectDisposedException(nameof(FrameResponseStream));
                 case FrameStreamState.Aborted:
-                    if (cancellationToken.CanBeCanceled)
+                    if (cancellationToken.IsCancellationRequested)
                     {
                         // Aborted state only throws on write if cancellationToken requests it
-                        return TaskUtilities.GetCancelledTask(
-                            cancellationToken.IsCancellationRequested ? 
-                            cancellationToken : 
-                            new CancellationToken(true));
+                        return TaskUtilities.GetCancelledTask(cancellationToken);
                     }
                     break;
             }
--- a/src/Microsoft.AspNetCore.Server.Kestrel/Http/MessageBody.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/Http/MessageBody.cs
@@ -2,11 +2,9 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
-using System.Collections.Generic;
 using System.IO;
 using System.Threading;
 using System.Threading.Tasks;
-using Microsoft.Extensions.Primitives;
 
 namespace Microsoft.AspNetCore.Server.Kestrel.Http
 {
--- a/src/Microsoft.AspNetCore.Server.Kestrel/Http/SocketInput.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/Http/SocketInput.cs
@@ -5,6 +5,7 @@ using System;
 using System.IO;
 using System.Runtime.CompilerServices;
 using System.Threading;
+using System.Threading.Tasks;
 using Microsoft.AspNetCore.Server.Kestrel.Infrastructure;
 
 namespace Microsoft.AspNetCore.Server.Kestrel.Http
@@ -184,7 +185,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Http
 
         public void AbortAwaiting()
         {
-            _awaitableError = new ObjectDisposedException(nameof(SocketInput), "The request was aborted");
+            _awaitableError = new TaskCanceledException("The request was aborted");
 
             Complete();
         }
@@ -238,6 +239,10 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Http
             var error = _awaitableError;
             if (error != null)
             {
+                if (error is TaskCanceledException || error is InvalidOperationException)
+                {
+                    throw error;
+                }
                 throw new IOException(error.Message, error);
             }
         }
--- a/src/Microsoft.AspNetCore.Server.Kestrel/Http/SocketOutput.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/Http/SocketOutput.cs
@@ -45,6 +45,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Http
         // The number of write operations that have been scheduled so far
         // but have not completed.
         private bool _writePending = false;
+        private bool _cancelled = false;
         private int _numBytesPreCompleted = 0;
         private Exception _lastWriteError;
         private WriteContext _nextWriteContext;
@@ -90,7 +91,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Http
 
             lock (_contextLock)
             {
-                if (_lastWriteError != null || _socket.IsClosed)
+                if (_socket.IsClosed)
                 {
                     _log.ConnectionDisconnectedWrite(_connectionId, buffer.Count, _lastWriteError);
 
@@ -164,7 +165,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Http
                         if (cancellationToken.IsCancellationRequested)
                         {
                             _connection.Abort();
-
+                            _cancelled = true;
                             return TaskUtilities.GetCancelledTask(cancellationToken);
                         }
                         else
@@ -295,7 +296,8 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Http
             {
                 // Abort the connection for any failed write
                 // Queued on threadpool so get it in as first op.
-                _connection?.Abort();
+                _connection.Abort();
+                _cancelled = true;
 
                 CompleteAllWrites();
             }
@@ -346,9 +348,9 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Http
         }
 
         // This may called on the libuv event loop
-        // This is always called with the _contextLock already acquired
         private void OnWriteCompleted(WriteContext writeContext)
         {
+            // Called inside _contextLock
             var bytesWritten = writeContext.ByteCount;
             var status = writeContext.WriteStatus;
             var error = writeContext.WriteError;
@@ -358,6 +360,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Http
                 // Abort the connection for any failed write
                 // Queued on threadpool so get it in as first op.
                 _connection.Abort();
+                _cancelled = true;
                 _lastWriteError = error;
             }
 
@@ -381,6 +384,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Http
 
         private void CompleteNextWrite(ref int bytesLeftToBuffer)
         {
+            // Called inside _contextLock
             var waitingTask = _tasksPending.Dequeue();
             var bytesToWrite = waitingTask.BytesToWrite;
 
@@ -416,6 +420,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Http
 
         private void CompleteFinishedWrites(int status)
         {
+            // Called inside _contextLock
             // bytesLeftToBuffer can be greater than _maxBytesPreCompleted
             // This allows large writes to complete once they've actually finished.
             var bytesLeftToBuffer = _maxBytesPreCompleted - _numBytesPreCompleted;
@@ -428,6 +433,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Http
 
         private void CompleteAllWrites()
         {
+            // Called inside _contextLock
             var writesToComplete = _tasksPending.Count > 0;
             var bytesLeftToBuffer = _maxBytesPreCompleted - _numBytesPreCompleted;
             while (_tasksPending.Count > 0)
@@ -468,7 +474,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Http
 
         private void PoolWriteContext(WriteContext writeContext)
         {
-            // called inside _contextLock
+            // Called inside _contextLock
             if (_writeContextPool.Count < _maxPooledWriteContexts)
             {
                 writeContext.Reset();
@@ -485,8 +491,13 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Http
         {
             if (cancellationToken.IsCancellationRequested)
             {
-                _connection?.Abort();
+                _connection.Abort();
+                _cancelled = true;
                 return TaskUtilities.GetCancelledTask(cancellationToken);
+            }
+            else if (_cancelled)
+            {
+                return TaskUtilities.CompletedTask;
             }
 
             return WriteAsync(buffer, cancellationToken, immediate, chunk);
--- a/src/Microsoft.AspNetCore.Server.Kestrel/Http/SocketOutput.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/Http/SocketOutput.cs
@@ -370,15 +370,15 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Http
             // completed writes that we haven't triggered callbacks for yet.
             _numBytesPreCompleted -= bytesWritten;
 
-            CompleteFinishedWrites(status);
-
-            if (error != null)
+            if (error == null)
             {
-                _log.ConnectionError(_connectionId, error);
+                CompleteFinishedWrites(status);
+                _log.ConnectionWriteCallback(_connectionId, status);
             }
             else
             {
-                _log.ConnectionWriteCallback(_connectionId, status);
+                CompleteAllWrites();
+                _log.ConnectionError(_connectionId, error);
             }
         }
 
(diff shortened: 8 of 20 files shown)


benaadams:
Confirmed this resolves the issue we were experiencing with Ssl (forced sync) and high connection churn.

src/Microsoft.AspNet.Server.Kestrel/Http/SocketOutput.cs
```diff
@@ -293,23 +297,35 @@ private void OnWriteCompleted(WriteContext writeContext)
...
+                        bytesLeftToBuffer -= bytesToWrite;
 
-                    _tasksCompleted.Enqueue(tcs);
+                        _tasksCompleted.Enqueue(tcs);
+                    }
+                }
+                else
+                {
+                    // Error all waiting tasks
+                    while (_tasksPending.Count > 0)
```
halter73: Thanks for doing this. I agree with this approach. This is type of change is what @rynowak and I were considering when he filed the issues at the end of last week.

I think we also want to fail all pending tasks if `OnWriteCompleted` is called after a `SocketDisconnect` just to be safe.

Another consideration is how this is going to interact with the per-request [IgnoreWriteExceptions](https://github.com/aspnet/KestrelHttpServer/issues/566#issuecomment-170194556) property we are likely to add. We still haven't decided whether or not it will be opt in or out, but I'm personally leaning toward opt-in.

This PR also conflicts a little with https://github.com/aspnet/KestrelHttpServer/pull/562 so one of us is going to have to rebase depending on the order these get in (and we know who controls that :stuck_out_tongue:). Seriously though, I'm going to try to merge #562 today and the rebase shouldn't be too difficult.

benaadams: > so one of us is going to have to rebase depending on the order these get in (and we know who controls that :stuck_out_tongue:

:laughing:

> Another consideration is how this is going to interact with the per-request IgnoreWriteExceptions property we are likely to add.

Isn't too hard; though would be an issue if someone was streaming something large but the client had disconnected as if they weren't listening for it they'd just keep writing to null?

halter73: We decided not to add an `IgnoreWriteExceptions` property. We will never throw from `SocketOutput.WriteAsync`. `FrameResponseStream` will be modified to honor the `CancellationToken` that is passed in for async writes.

If you want the `IgnoreWriteExceptions = false` behavior you will always pass in the `IHttpRequestLifetimeFeature.RequestAborted` token to your writes. If `FrameResponseStream` ever throws due to a canceled token it will abort the connection.

We will also plan to log all write failures using `LogLevel.Debug` correlated with the connection id. The connection id will be made accessible from user code via a new `IHttpConnectionFeature.ConnectionId` property.

benaadams: What about lines new lines 86-89 ---^

halter73: Are you talking about `if (_lastWriteError != null) throw ...`? If so, we'd remove that and just ignore the write while returning a completed task.

benaadams:
Going to roll this into #485; if that works for you?

halter73:
That will probably fine. We definitely appreciate the work. Try to separate the changes related to https://github.com/aspnet/KestrelHttpServer/issues/566#issuecomment-171009932 into their own commit if you can.

benaadams:
Will split into 2 commits when I know the shape of it

benaadams:
Need to sort out a new test

benaadams:
@halter73 Added test, probably could do with some comments - but otherwise seem ok?

src/Microsoft.AspNet.Server.Kestrel/Http/SocketOutput.cs
```diff
@@ -298,44 +320,72 @@ private void OnWriteCompleted(WriteContext writeContext)
...
                 else
                 {
-                    _threadPool.Error(tcs, _lastWriteError);
+                    _threadPool.Complete(waitingTask.CompletionSource);
                 }
             }
 
             _log.ConnectionWriteCallback(_connectionId, status);
         }
+        private void CancelOutstandingTasks(Exception error)
```
cesarbs: nit: add new line before method

src/Microsoft.AspNet.Server.Kestrel/Http/SocketOutput.cs
```diff
@@ -388,7 +438,23 @@ void ISocketOutput.Write(ArraySegment<byte> buffer, bool immediate, bool chunk)
 
         Task ISocketOutput.WriteAsync(ArraySegment<byte> buffer, bool immediate, bool chunk, CancellationToken cancellationToken)
         {
-            return WriteAsync(buffer, immediate, chunk);
+            if (cancellationToken.IsCancellationRequested)
+            {
+#if DOTNET5_4
+                return Task.FromCanceled(cancellationToken);
+#else
+                throw new TaskCanceledException();
```
cesarbs: This is a significant behavioral difference between the two frameworks. Isn't it possible to do something similar to what's being done for dotnet5.4?

benaadams: I'll have a dig; might be able to do something odd for 451 - like return a faulting lambda task

benaadams: Have something better

benaadams:
Made behaviour same between frameworks and added comments to new test

benaadams:
1 sec missed a spot

benaadams:
K; think that's the correct behaviour?

src/Microsoft.AspNet.Server.Kestrel/Http/FrameResponseStream.cs
```diff
@@ -66,14 +66,14 @@ public override int Read(byte[] buffer, int offset, int count)
...
-            ValidateState();
+            ValidateState(CancellationToken.None);
 
             _context.FrameControl.Write(new ArraySegment<byte>(buffer, offset, count));
         }
 
         public override Task WriteAsync(byte[] buffer, int offset, int count, CancellationToken cancellationToken)
         {
-            ValidateState();
+            ValidateState(cancellationToken);
```
benaadams: Should return cancelled task rather than throwing TaskCanceledException?

src/Microsoft.AspNet.Server.Kestrel/Http/Frame.cs
```diff
@@ -307,16 +307,19 @@ public void Abort()
...
             catch (Exception ex)
             {
                 Log.LogError("Abort", ex);
             }
-            finally
+
+            try
             {
-                _abortedCts = null;
+                RequestAbortedSource.Cancel();
```
halter73: I don't think we should swallow this error. I think there is a chance that exceptions from callbacks registered with the token will bubble up through this call. I would probably use `KestrelTrace.ApplicationError`.

src/Microsoft.AspNet.Server.Kestrel/Http/FrameRequestStream.cs
```diff
@@ -149,23 +149,27 @@ public void StopAcceptingReads()
...
-            // If the request is aborted, we throw an IOException instead.
+            // If the request is aborted, we throw an TaskCanceledException instead.
             if (_state != FrameStreamState.Closed)
             {
                 _state = FrameStreamState.Aborted;
             }
         }
 
-        private void ValidateState()
+        private void ValidateState(CancellationToken cancellationToken)
```
halter73: Should this return a Task like the `FrameResponseStream` version of `ValidateState` so we can return a faulted task instead of throw synchronously?

src/Microsoft.AspNet.Server.Kestrel/Http/SocketOutput.cs
```diff
@@ -298,45 +313,74 @@ private void OnWriteCompleted(WriteContext writeContext)
...
+        }
 
+        private void CompleteFinishedWrites(int status)
+        {
             // bytesLeftToBuffer can be greater than _maxBytesPreCompleted
             // This allows large writes to complete once they've actually finished.
             var bytesLeftToBuffer = _maxBytesPreCompleted - _numBytesPreCompleted;
             while (_tasksPending.Count > 0 &&
-                   (int)(_tasksPending.Peek().Task.AsyncState) <= bytesLeftToBuffer)
+                   (_tasksPending.Peek().BytesToWrite) <= bytesLeftToBuffer)
```
halter73: Back to the old way. I like the improved readability :smile:

src/Microsoft.AspNet.Server.Kestrel/Http/SocketOutput.cs
```diff
@@ -91,6 +91,11 @@ public class SocketOutput : ISocketOutput
                 if (buffer.Count > 0)
                 {
                     var tail = ProducingStart();
+                    if (tail.IsDefault)
```
halter73: I think we need to make `CopyFrom` itself no-op. I don't want to risk a `NullReferenceException` being thrown from `CreateResponseHeader` and the like.

src/Microsoft.AspNet.Server.Kestrel/Http/SocketOutput.cs
```diff
@@ -146,8 +151,16 @@ public class SocketOutput : ISocketOutput
                 else
                 {
                     // immediate write, which is not eligable for instant completion above
-                    tcs = new TaskCompletionSource<object>(buffer.Count);
-                    _tasksPending.Enqueue(tcs);
+                    tcs = new TaskCompletionSource<object>();
+                    if (!cancellationToken.IsCancellationRequested)
+                    {
+                        _tasksPending.Enqueue(new WaitingTask() { Token = cancellationToken, BytesToWrite = buffer.Count, CompletionSource = tcs });
```
halter73: I think we are going to need to register with the `cancellationToken` if it [CanBeCanceled](https://msdn.microsoft.com/en-us/library/system.threading.cancellationtoken.canbecanceled(v=vs.110).aspx). If the callback is triggered before the write completes we should abort the connection.

test/Microsoft.AspNet.Server.KestrelTests/SocketOutputTests.cs
```diff
@@ -199,6 +199,147 @@ public void WritesDontCompleteImmediatelyWhenTooManyBytesIncludingNonImmediateAr
...
+
+                while (completeQueue.Count > 0)
+                {
+                    completeQueue.Dequeue()(0);
+                }
+
+                await Task.Delay(2000).ConfigureAwait(false);
+
+                // Seventh task is completed and sucessful
+                Assert.True(task7Success.IsCompleted);
```
halter73: This assertion [failed on AppVeyor](https://github.com/aspnet/KestrelHttpServer/pull/567/files).

src/Microsoft.AspNet.Server.Kestrel/Http/SocketOutput.cs
```diff
@@ -298,45 +313,74 @@ private void OnWriteCompleted(WriteContext writeContext)
...
+                var bytesToWrite = waitingTask.BytesToWrite;
 
                 _numBytesPreCompleted += bytesToWrite;
                 bytesLeftToBuffer -= bytesToWrite;
 
-                if (_lastWriteError == null)
+                if (waitingTask.Token.IsCancellationRequested)
                 {
-                    _threadPool.Complete(tcs);
+                    _threadPool.Cancel(waitingTask.CompletionSource);
```
halter73: At this point since the write has actually already completed, there isn't really any need to cancel the task even if the token has somehow tripped. We might as well complete them all as before.

In practice, once we start registering with cancellable tokens and unregistering prior to completing/canceling the TCSs, it's pretty unlikely we would get to this point with a canceled token.

src/Microsoft.AspNet.Server.Kestrel/Http/SocketOutput.cs
benaadams: (box)(unbox)(unbox)?

benaadams: Oh; gotcha :grin:

src/Microsoft.AspNet.Server.Kestrel/Http/SocketOutput.cs
benaadams: Done

benaadams:
Still need to do task registration; and then rework the tests.

src/Microsoft.AspNet.Server.Kestrel/Http/SocketOutput.cs
benaadams: Done

benaadams:
Hopefully, this is now the correct behaviour?

benaadams:
Will investigate mono failure in morning.

src/Microsoft.AspNet.Server.Kestrel/Http/FrameResponseStream.cs
```diff
@@ -112,24 +119,36 @@ public void StopAcceptingWrites()
...
+                        return TaskUtilities.GetCancelledTask(cancellationToken);
+                    }
+                    break;
                 case FrameStreamState.Closed:
                     throw new ObjectDisposedException(nameof(FrameResponseStream));
                 case FrameStreamState.Aborted:
-                    throw new IOException("The request has been aborted.");
+                    if (cancellationToken.CanBeCanceled)
+                    {
+                        // Abroted state only throws on write if cancellationToken requests it
```
cesarbs: nit: typo (Abroted)

test/Microsoft.AspNet.Server.KestrelTests/EngineTests.cs
```diff
@@ -1127,7 +1127,7 @@ public void ConnectionCanReadAndWrite(TestServiceContext testContext)
                     writeTcs.SetException(ex);
 
                     // Give a chance for RequestAborted to trip before the app completes
-                    registrationWh.Wait(1000);
+                    registrationWh.Wait(2000);
```
cesarbs: Put this value in a `const`.

test/Microsoft.AspNet.Server.KestrelTests/EngineTests.cs
```diff
@@ -1119,7 +1119,7 @@ public void ConnectionCanReadAndWrite(TestServiceContext testContext)
                     // Ensure write is long enough to disable write-behind buffering
                     for (int i = 0; i < 10; i++)
                     {
-                        await response.WriteAsync(new string('a', 65537));
+                        await response.WriteAsync(new string('a', 65537)).ConfigureAwait(false);
```
cesarbs: Put hardcoded values in `const`s.

test/Microsoft.AspNet.Server.KestrelTests/EngineTests.cs
```diff
@@ -1141,16 +1141,16 @@ public void ConnectionCanReadAndWrite(TestServiceContext testContext)
...
                 }
 
                 connectionCloseWh.Set();
 
                 // Write failed
-                await Assert.ThrowsAsync<IOException>(async () => await writeTcs.Task);
+                await Assert.ThrowsAsync<TaskCanceledException>(async () => await writeTcs.Task);
                 // RequestAborted tripped
-                Assert.True(registrationWh.Wait(200));
+                Assert.True(registrationWh.Wait(2000));
```
cesarbs: `const`

benaadams:
Also see #589

src/Microsoft.AspNet.Server.Kestrel/Http/FrameOfT.cs
```diff
@@ -111,7 +112,7 @@ public class Frame<TContext> : Frame
                             _application.DisposeContext(context, _applicationException);
 
                             // If _requestAbort is set, the connection has already been closed.
-                            if (!_requestAborted)
+                            if (Volatile.Read(ref _requestAborted) == 0)
```
halter73: I don't thinks these volatile reads are really necessary. http://joeduffyblog.com/2010/12/04/sayonara-volatile/

src/Microsoft.AspNet.Server.Kestrel/Http/SocketOutput.cs
```diff
@@ -91,6 +92,21 @@ public class SocketOutput : ISocketOutput
...
                 {
                     var tail = ProducingStart();
+                    if (tail.IsDefault)
+                    {
+                        if (cancellationToken.CanBeCanceled)
+                        {
+                            return TaskUtilities.GetCancelledTask(
+                                cancellationToken.IsCancellationRequested ?
+                                cancellationToken :
+                                new CancellationToken(true));
```
halter73: We only want to throw if the token is actually canceled. This seems to be throwing in any case there is a cancelable token and the socket is closed.

src/Microsoft.AspNet.Server.Kestrel/Http/SocketOutput.cs
```diff
@@ -298,43 +353,72 @@ private void OnWriteCompleted(WriteContext writeContext)
...
+
+            while (_tasksPending.Count > 0)
+            {
+                var waitingTask = _tasksPending.Dequeue();
+
+                if (waitingTask.CancellationRegistration != null)
                 {
-                    _threadPool.Complete(tcs);
+                    // Only Cancellation tokens which accept cancellation
+                    // therefore have a registration should be cancelled
```
halter73: We should only cancel writes when we've verified the token *has* been cancelled not just that it can be cancelled.

src/Microsoft.AspNet.Server.Kestrel/Http/SocketOutput.cs
```diff
@@ -298,43 +353,72 @@ private void OnWriteCompleted(WriteContext writeContext)
...
-                _lastWriteError = new IOException(error.Message, error);
+                // _numBytesPreCompleted can temporarily go negative in the event there are
+                // completed writes that we haven't triggered callbacks for yet.
+                _numBytesPreCompleted -= bytesWritten;
 
-                // Abort the connection for any failed write.
-                _connection.Abort();
+                CompleteFinishedWrites(status);
             }
+            else if (error != null)
```
halter73: or just `else` :grinning:

src/Microsoft.AspNet.Server.Kestrel/Http/SocketOutput.cs
```diff
@@ -246,6 +289,18 @@ private void ProducingCompleteNoPreComplete(MemoryPoolIterator2 end)
             }
         }
 
+        private void CancellationTriggered()
+        {
+            lock (_contextLock)
+            {
+                // Abort the connection for any failed write
+                // Queued on threadpool so get it in as first op.
+                _connection?.Abort();
```
halter73: Is `?.` for tests? If not we should just call `_connection.Abort` since `_connection` should always be set in the ctor.

src/Microsoft.AspNet.Server.Kestrel/Http/SocketOutput.cs
benaadams: But the write has been cancelled? And if a token is passed in that `CanBeCanceled` it means something is listening for cancellation and will handle it.

Alternative is just to mark as complete, even though it has been cancelled?

src/Microsoft.AspNet.Server.Kestrel/Http/SocketOutput.cs
benaadams: (dupe response) But the write has been cancelled? And if a token is passed in that `CanBeCanceled` it means something is listening for cancellation and will handle it.

Alternative is just to mark as complete, even though it has been cancelled?

halter73: This is something I was really on the fence on too, but @Tratcher convinced me we should only cancel the task if the passed in token has actually been canceled.

The idea is that it's easier for the caller to reason about their code if the only way a write can fail is if the token that was passed in was actually canceled.

In theory, the behavior should be the same either way if the caller is always passing in the `RequestAborted` token. We just have to be extra careful to trip the token before determining whether to cancel or complete or cancel the task.

If you're not on board with this, we could probably wait until @Tratcher chimes in so we can have a group discussion. I'm probably not going to merge this until tomorrow anyway. Everyone here in Seattle has MLK day off :grin:

I think this PR is very close to being ready though.

src/Microsoft.AspNet.Server.Kestrel/Infrastructure/MemoryPoolIterator2.cs
```diff
@@ -728,6 +728,11 @@ public void CopyFrom(byte[] data, int offset, int count)
             Debug.Assert(_block.Next == null);
             Debug.Assert(_block.End == _index);
 
+            if (IsDefault)
```
halter73: The assertions should probably be moved to be after this check.

src/Microsoft.AspNet.Server.Kestrel/Infrastructure/TaskUtilities.cs
```diff
@@ -13,5 +14,23 @@ public static class TaskUtilities
...
+        public static Task GetCancelledTask(CancellationToken cancellationToken)
+        {
+#if DOTNET5_4
+            return Task.FromCanceled(cancellationToken);
+#else
+            var tcs = new TaskCompletionSource<object>();
+            tcs.TrySetCanceled();
+            return tcs.Task;
+#endif
+        }
```
halter73: Nit: newline

src/Microsoft.AspNet.Server.Kestrel/Http/FrameOfT.cs
halter73: On second thought, I've changed my mind. Keep the volatile reads, since the bool was volatile before. I think the only real downside is potential perf, and I'd rather not be responsible for another regression.

src/Microsoft.AspNet.Server.Kestrel/Http/SocketInput.cs
```diff
@@ -238,6 +239,10 @@ public void GetResult()
             var error = _awaitableError;
             if (error != null)
             {
+                if (error is TaskCanceledException || error is InvalidOperationException)
```
halter73: @Tratcher I would just like a second opinion on changing the `IOException` to `TaskCanceledExceptions` when the request is aborted. I think it's fine.

src/Microsoft.AspNet.Server.Kestrel/Http/FrameOfT.cs
benaadams: It wasn't a regression; they way it turned out it was a fact finding exercise :grin:

src/Microsoft.AspNet.Server.Kestrel/Http/SocketOutput.cs
benaadams: Yeah; should probably make a fake connection or something

src/Microsoft.AspNet.Server.Kestrel/Http/FrameOfT.cs
benaadams: You know someone, somewhere, is going to run a Kestrel websocket server on an arm Windows phone as part of their UWP app, and curse the memory model - just saying... :wink:

src/Microsoft.AspNet.Server.Kestrel/Http/SocketOutput.cs
benaadams: I think the `RequestAborted` token gets cancelled on the threadpool to deal with any registrations, so may need to use a flag/property or similar that's set sync on the connection rather than the token itself :-/

benaadams: Its sort of the approach from `Listening to Multiple Tokens Simultaneously` from https://msdn.microsoft.com/en-us/library/dd997364(v=vs.110).aspx with the internal and external token but without the expense of using `CreateLinkedTokenSource`; happy to change it pending discussion with @Tratcher

Tratcher: This whole flow is riddled with race conditions so you're never going to get a deterministic behavior. The one behavior we do want to guarantee is the following: When I call WriteAsync with a cancellable CT and the operation becomes async, then I should be able to cancel that CT and have the async operation end in a timely fashion. Even here there is a race here where the operation may complete successfully just before being cancelled, or it may throw a TaskCancelledException (or equivalent). We're not trying to grantee it throws, only that it ends. If there's a race then the cancelled CT will still fail next time it's checked and terminate operations.

benaadams: Might be easier to ask the questions with test behaviour; should this be a cancelled or completed Task? 

https://github.com/benaadams/KestrelHttpServer/blob/socket-error/test/Microsoft.AspNet.Server.KestrelTests/SocketOutputTests.cs#L298-L305

Tratcher: No, a Task's state should only be affected by directly associated CTs. E.g. a CT from a prior Write should not affect the Task from the next Write. This should make the implementation much simpler. Usually it won't matter as the same CT will be used with most/all operations.

Tratcher: I guess that wasn't a yes or no question... It should be completed, not cancelled.

benaadams:
K, Adapting to feedback, was waiting for feeds to update - which they now have

benaadams:
Need to run through this again in morning

benaadams:
Need to run through this again in morning; though I just think need a fake connection for tests to remove `?` in `_connection?.Abort();`

src/Microsoft.AspNet.Server.Kestrel/Infrastructure/TaskUtilities.cs
benaadams: Added

benaadams:
Annotate all function with expected locks (refactor guard)

benaadams:
@halter73 should be there? Issues with CI on Github though (libuv copy)

halter73:
@cesarbs should have a fix for the libuv copy issue. He discovered the issue is that the [Libuv-Copier](https://github.com/aspnet/KestrelHttpServer/blob/dev/tools/Microsoft.AspNet.Server.Kestrel.LibuvCopier/Program.cs) is looking for libuv in `~/.dnx/packages` instead of the [now used `~/.nuget/packages` directory](https://github.com/aspnet/dnx/commit/d233abfdec8f4eb3934fbe714c05c2b99440703e).

src/Microsoft.AspNet.Server.Kestrel/Http/SocketOutput.cs
```diff
@@ -317,43 +370,81 @@ private void OnWriteCompleted(WriteContext writeContext)
...
+            // This allows large writes to complete once they've actually finished.
+            var bytesLeftToBuffer = _maxBytesPreCompleted - _numBytesPreCompleted;
+            while (_tasksPending.Count > 0 &&
+                   (_tasksPending.Peek().BytesToWrite) <= bytesLeftToBuffer)
+            {
+                CompleteNextWrite(ref bytesLeftToBuffer);
+            }
+        }
+
+        private void CompleteAllWrites()
```
halter73: It seems to me that this is only getting called if the user passes in a canceled token, but that's only one scenario where this is getting called.

We need to complete all writes whenever there is an error, even if we don't wind up cancelling them.

benaadams: Hmm... let me run through the logic again

benaadams: Fixed, though I have it Completing written ones first on Error, then doing the others. Probably should just do all regardless of counts?

benaadams:
libuv copy grumpy

src/Microsoft.AspNet.Server.Kestrel/Http/SocketOutput.cs
halter73: >  I have it Completing written ones first on Error, then doing the others

This has no functional impact vs completing all the writes at once on error, right? If so, that's fine, as long as the code stays simple.

benaadams: No, but it looks a bit weird. Let me change it since the CI is failing anyway.

benaadams: Resolved, was just a case of moving one line of code - looks better now

test/Microsoft.AspNet.Server.KestrelTests/SocketOutputTests.cs
```diff
@@ -199,6 +199,116 @@ public void WritesDontCompleteImmediatelyWhenTooManyBytesIncludingNonImmediateAr
         }
 
         [Fact]
+        public async Task OnlyWritesRequestingCancellationAreErroredOnCancellation()
```
halter73: Note to self: We should add a test where a bunch of pending writes that get completed b/c of a failed write instead of a canceled token before merging.

I can do this myself since this will require some small modifications to MockLibuv (I know _uv_err_name and _uv_strerror will need to be implemented.)

There's also a pretty big chance that we might need to merge a more limited version of this PR that only ensures writes always get completed, and merge the CancellationToken related chances post-RC2.

benaadams: Hmm... you want to raise a CancellationToken issue to track? Will take about 5 mins to do a complete on error PR.

src/Microsoft.AspNet.Server.Kestrel/Http/SocketOutput.cs
```diff
@@ -317,43 +370,81 @@ private void OnWriteCompleted(WriteContext writeContext)
...
+            var writesToComplete = _tasksPending.Count > 0;
+            var bytesLeftToBuffer = _maxBytesPreCompleted - _numBytesPreCompleted;
+            while (_tasksPending.Count > 0)
+            {
+                CompleteNextWrite(ref bytesLeftToBuffer);
+            }
+
+            if (writesToComplete)
+            {
+                _log.ConnectionError(_connectionId, new TaskCanceledException("Connetcion"));
```
halter73: We should move this to `CancellationTriggered` and guard that method with an `if (!_canceled)`

halter73: Oh and a better exception message would be "Write operation canceled. Aborting connection."

src/Microsoft.AspNet.Server.Kestrel/Http/FrameOfT.cs
cesarbs: What is the difference between having a `volatile int` and doing `Volatile.Read`?

benaadams: `volatile` is compiler `Volatile` is cpu? Would be same on intel; ARM is dodgy

cesarbs: @benaadams I see, thanks for the clarification!

benaadams: Ah might be same; could be for VB - but is also more in your face :)

benaadams: Ah might be same; could be for VB - but is also more in your face :) Was compiler only at one point I believe - probably changed after RT and WinPhone.

benaadams: Jit replaced with fallback:
https://github.com/dotnet/coreclr/blob/master/src/mscorlib/src/System/Threading/Volatile.cs#L86-L94

benaadams: Jit replaced with fallback:
https://github.com/dotnet/coreclr/blob/master/src/mscorlib/src/System/Threading/Volatile.cs#L86-L94

Also it gets grumpy if you put `volatile` in `Interlocked`

src/Microsoft.AspNet.Server.Kestrel/Http/FrameResponseStream.cs
```diff
@@ -112,24 +119,36 @@ public void StopAcceptingWrites()
...
                     throw new ObjectDisposedException(nameof(FrameResponseStream));
                 case FrameStreamState.Aborted:
-                    throw new IOException("The request has been aborted.");
+                    if (cancellationToken.CanBeCanceled)
+                    {
+                        // Aborted state only throws on write if cancellationToken requests it
+                        return TaskUtilities.GetCancelledTask(
+                            cancellationToken.IsCancellationRequested ? 
+                            cancellationToken : 
+                            new CancellationToken(true));
```
benaadams: Pre-newer CancellationToken changes, needs to only throw if given `CancellationToken` is cancelled otherwise it should return `null`.

benaadams:
Revised test showed issue with v large files, latency of send start and cancellation:

Rebased-branch with added test and fix https://github.com/benaadams/KestrelHttpServer/tree/socket-error-merged-fix
Non-Rebased with added test and fix https://github.com/benaadams/KestrelHttpServer/tree/socket-error-non-rebased

Going to do some wrk testing with 1 GB files and check latencies

benaadams:
Bad day to be building :P

benaadams:
Linked #599

benaadams:
Rebased

PR #570, opened 2016-01-11
using reorder (nit)

benaadams:


halter73:
:+1:  I haven't got around to fixing it b/c it was an easy way to see if warnings are still showing up post #518 by just looking at the appveyor/travis logs.

Issue #273, opened 2015-10-21
Normalize URL


troydai:
It doesn't seem to me that Kestrel normalizes URL.

muratg:
Can you give some details?  A few examples?

cc @Tratcher

Tratcher:
https://www.ietf.org/rfc/rfc3987.txt
5.3.2.2.  Character Normalization
And maybe
5.3.2.4.  Path Segment Normalization

muratg:
@Tratcher @troydai Do we need to do this? What's the user impact if we don't do it?

Tratcher:
That's more a question for @blowdart. We're un-escaping arbitrary Unicode from remote users. Normalization helps avoid lots of known issues with strange unicode combinations. http://unicode.org/faq/normalization.html

muratg:
@blowdart Please advise :smile:

blowdart:
You must normalize. Just like IIS does.

cesarbs:
Do we want to preserve the original URL as a field in the context i.e. the original bytes and path before normalization and dot segment removal?

Tratcher:
Not right now.

DamianEdwards:
Need for RC2

PR #573, opened 2016-01-11
Normalize request path to NFC and remove/resolve dot segments (#273).

cesarbs:
#273 

cc @Tratcher @halter73 @blowdart

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Kestrel/Http/Frame.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/Http/Frame.cs
@@ -741,12 +741,15 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Http
                     // URI was encoded, unescape and then parse as utf8
                     pathEnd = UrlPathDecoder.Unescape(pathBegin, pathEnd);
                     requestUrlPath = pathBegin.GetUtf8String(pathEnd);
+                    requestUrlPath = PathNormalizer.NormalizeToNFC(requestUrlPath);
                 }
                 else
                 {
                     // URI wasn't encoded, parse as ASCII
                     requestUrlPath = pathBegin.GetAsciiString(pathEnd);
                 }
+
+                requestUrlPath = PathNormalizer.RemoveDotSegments(requestUrlPath);
 
                 consumed = scan;
                 Method = method;
--- a/src/Microsoft.AspNetCore.Server.Kestrel/Http/PathNormalizer.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/Http/PathNormalizer.cs
@@ -1,0 +1,132 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Buffers;
+using System.Text;
+
+namespace Microsoft.AspNetCore.Server.Kestrel.Http
+{
+    public static class PathNormalizer
+    {
+        public static string NormalizeToNFC(string path)
+        {
+            if (!path.IsNormalized(NormalizationForm.FormC))
+            {
+                path = path.Normalize(NormalizationForm.FormC);
+            }
+
+            return path;
+        }
+
+        public static string RemoveDotSegments(string path)
+        {
+            if (ContainsDotSegments(path))
+            {
+                var normalizedChars = ArrayPool<char>.Shared.Rent(path.Length);
+                var normalizedIndex = normalizedChars.Length;
+                var pathIndex = path.Length - 1;
+                var skipSegments = 0;
+
+                while (pathIndex >= 0)
+                {
+                    if (pathIndex >= 2 && path[pathIndex] == '.' && path[pathIndex - 1] == '.' && path[pathIndex - 2] == '/')
+                    {
+                        if (normalizedIndex == normalizedChars.Length || normalizedChars[normalizedIndex] != '/')
+                        {
+                            normalizedChars[--normalizedIndex] = '/';
+                        }
+
+                        skipSegments++;
+                        pathIndex -= 3;
+                    }
+                    else if (pathIndex >= 1 && path[pathIndex] == '.' && path[pathIndex - 1] == '/')
+                    {
+                        pathIndex -= 2;
+                    }
+                    else
+                    {
+                        while (pathIndex >= 0)
+                        {
+                            var lastChar = path[pathIndex];
+
+                            if (skipSegments == 0)
+                            {
+                                normalizedChars[--normalizedIndex] = lastChar;
+                            }
+
+                            pathIndex--;
+
+                            if (lastChar == '/')
+                            {
+                                break;
+                            }
+                        }
+
+                        if (skipSegments > 0)
+                        {
+                            skipSegments--;
+                        }
+                    }
+                }
+
+                path = new string(normalizedChars, normalizedIndex, normalizedChars.Length - normalizedIndex);
+                ArrayPool<char>.Shared.Return(normalizedChars);
+            }
+
+            return path;
+        }
+
+        private unsafe static bool ContainsDotSegments(string path)
+        {
+            fixed (char* ptr = path)
+            {
+                char* end = ptr + path.Length;
+
+                for (char* p = ptr; p < end; p++)
+                {
+                    if (*p == '/')
+                    {
+                        p++;
+                    }
+
+                    if (p == end)
+                    {
+                        return false;
+                    }
+
+                    if (*p == '.')
+                    {
+                        p++;
+
+                        if (p == end)
+                        {
+                            return true;
+                        }
+
+                        if (*p == '.')
+                        {
+                            p++;
+
+                            if (p == end)
+                            {
+                                return true;
+                            }
+
+                            if (*p == '/')
+                            {
+                                return true;
+                            }
+                        }
+                        else if (*p == '/')
+                        {
+                            return true;
+                        }
+                    }
+                }
+            }
+
+            return false;
+        }
+    }
+}
--- a/src/Microsoft.AspNetCore.Server.Kestrel/ServerAddress.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/ServerAddress.cs
@@ -4,6 +4,7 @@
 using System;
 using System.Diagnostics;
 using System.Globalization;
+using Microsoft.AspNetCore.Server.Kestrel.Http;
 using Microsoft.AspNetCore.Server.Kestrel.Infrastructure;
 
 namespace Microsoft.AspNetCore.Server.Kestrel
@@ -146,6 +147,8 @@ namespace Microsoft.AspNetCore.Server.Kestrel
             {
                 serverAddress.PathBase = url.Substring(pathDelimiterEnd);
             }
+
+            serverAddress.PathBase = PathNormalizer.NormalizeToNFC(serverAddress.PathBase);
 
             return serverAddress;
         }
--- a/src/Microsoft.AspNetCore.Server.Kestrel/project.json
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/project.json
@@ -6,6 +6,7 @@
     "url": "git://github.com/aspnet/kestrelhttpserver"
   },
   "dependencies": {
+    "System.Buffers": "4.0.0-*",
     "Microsoft.AspNetCore.Hosting": "1.0.0-*",
     "Microsoft.Extensions.Logging.Abstractions": "1.0.0-*",
     "Microsoft.Extensions.PlatformAbstractions": "1.0.0-*",
--- a/test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/RequestTests.cs
+++ b/test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/RequestTests.cs
@@ -3,7 +3,10 @@
 
 using System.Collections.Generic;
 using System.Globalization;
+using System.Net;
 using System.Net.Http;
+using System.Net.Sockets;
+using System.Text;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Builder;
 using Microsoft.AspNetCore.Hosting;
@@ -88,7 +91,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.FunctionalTests
         [IPv6SupportedCondition]
         public Task RemoteIPv6Address()
         {
-            return TestRemoteIPAddress("[::1]", "[::1]", "::1", "8792");
+            return TestRemoteIPAddress("[::1]", "[::1]", "::1", "8793");
         }
 
         [ConditionalFact]
@@ -97,7 +100,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.FunctionalTests
         {
             var config = new ConfigurationBuilder().AddInMemoryCollection(
                 new Dictionary<string, string> {
-                    { "server.urls", "http://localhost:8791" }
+                    { "server.urls", "http://localhost:8794" }
                 }).Build();
 
             var builder = new WebHostBuilder()
@@ -120,8 +123,59 @@ namespace Microsoft.AspNetCore.Server.Kestrel.FunctionalTests
                 client.DefaultRequestHeaders.Connection.Clear();
                 client.DefaultRequestHeaders.Connection.Add("close");
 
-                var response = await client.GetAsync("http://localhost:8791/");
+                var response = await client.GetAsync("http://localhost:8794/");
                 response.EnsureSuccessStatusCode();
+            }
+        }
+
+        [ConditionalFact]
+        [FrameworkSkipCondition(RuntimeFrameworks.Mono, SkipReason = "Test hangs after execution on Mono.")]
+        public async Task RequestPathIsNormalized()
+        {
+            var config = new ConfigurationBuilder().AddInMemoryCollection(
+                new Dictionary<string, string> {
+                    { "server.urls", "http://localhost:8795/\u0041\u030A" }
+                }).Build();
+
+            var builder = new WebHostBuilder()
+                .UseConfiguration(config)
+                .UseServer("Microsoft.AspNetCore.Server.Kestrel")
+                .Configure(app =>
+                {
+                    app.Run(async context =>
+                    {
+                        var connection = context.Connection;
+                        Assert.Equal("/\u00C5", context.Request.PathBase.Value);
+                        Assert.Equal("/B/\u00C5", context.Request.Path.Value);
+                        await context.Response.WriteAsync("hello, world");
+                    });
+                });
+
+            using (var host = builder.Build())
+            {
+                host.Start();
+
+                using (var socket = new Socket(AddressFamily.InterNetwork, SocketType.Stream, ProtocolType.Tcp))
+                {
+                    socket.Connect(new IPEndPoint(IPAddress.Loopback, 8795));
+                    socket.Send(Encoding.ASCII.GetBytes("GET /%41%CC%8A/A/../B/%41%CC%8A HTTP/1.1\r\n\r\n"));
+                    socket.Shutdown(SocketShutdown.Send);
+
+                    var response = new StringBuilder();
+                    var buffer = new byte[4096];
+                    while (true)
+                    {
+                        var length = socket.Receive(buffer);
+                        if (length == 0)
+                        {
+                            break;
+                        }
+                        
+                        response.Append(Encoding.ASCII.GetString(buffer, 0, length));
+                    }
+                    
+                    Assert.StartsWith("HTTP/1.1 200 OK", response.ToString());
+                }
             }
         }
 
--- a/test/Microsoft.AspNetCore.Server.KestrelTests/EngineTests.cs
+++ b/test/Microsoft.AspNetCore.Server.KestrelTests/EngineTests.cs
@@ -1120,7 +1120,7 @@ namespace Microsoft.AspNetCore.Server.KestrelTests
                 try
                 {
                     // Ensure write is long enough to disable write-behind buffering
-                    for (int i = 0; i < 10; i++)
+                    for (int i = 0; i < 100; i++)
                     {
                         await response.WriteAsync(largeString, lifetime.RequestAborted);
                     }
--- a/test/Microsoft.AspNetCore.Server.KestrelTests/PathNormalizerTests.cs
+++ b/test/Microsoft.AspNetCore.Server.KestrelTests/PathNormalizerTests.cs
@@ -1,0 +1,63 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Text;
+using Microsoft.AspNetCore.Server.Kestrel.Http;
+using Xunit;
+
+namespace Microsoft.AspNetCore.Server.KestrelTests
+{
+    public class PathNormalizerTests
+    {
+        [Theory]
+        [InlineData("/a", "/a")]
+        [InlineData("/a/", "/a/")]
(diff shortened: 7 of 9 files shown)


src/Microsoft.AspNet.Server.Kestrel/Http/Frame.cs
```diff
@@ -775,6 +775,11 @@ protected bool TakeStartLine(SocketInput input)
                     // URI was encoded, unescape and then parse as utf8
                     pathEnd = UrlPathDecoder.Unescape(pathBegin, pathEnd);
                     requestUrlPath = pathBegin.GetUtf8String(pathEnd);
+
+                    if (PathNormalizer.NeedsNormalization(requestUrlPath))
```
benaadams: Does this only effect URI encoded paths and not non-encoded paths?

benaadams: Ah yes

cesarbs:
Ping.

halter73:
I'm curious to see how this affects perf for requests that don't need normalization (e.g. the plaintext benchmark).

test/Microsoft.AspNet.Server.Kestrel.FunctionalTests/RequestTests.cs
```diff
@@ -120,7 +120,39 @@ public Task RemoteIPv6Address()
...
 
-                var response = await client.GetAsync("http://localhost:8791/");
+                var response = await client.GetAsync("http://localhost:8794/");
+                response.EnsureSuccessStatusCode();
+            }
+        }
+
+        [ConditionalFact]
+        [FrameworkSkipCondition(RuntimeFrameworks.Mono, SkipReason = "Test hangs after execution on Mono.")]
+        public async Task RequestPathIsNormalized()
```
cesarbs: Note to self: add test with PathBase.

benaadams:
It shouldn't effect it much as `needDecode` would be `false` it that case and it should just skip it?

halter73:
There shouldn't be much of an affect as long as `PathNormalizer.NeedsNormalization` isn't too expensive. I would like to know for sure that's the case.

src/Microsoft.AspNet.Server.Kestrel/Http/Frame.cs
Tratcher: needDecode and NeedsNormalization are now out of sync because dots don't trigger needDecode. Add a functional test because I don't think this code is actually executing right now.

test/Microsoft.AspNet.Server.Kestrel.FunctionalTests/RequestTests.cs
```diff
@@ -120,7 +120,39 @@ public Task RemoteIPv6Address()
...
+                        await context.Response.WriteAsync("hello, world");
+                    });
+                });
+
+            using (var app = builder.Build())
+            using (var client = new HttpClient())
+            {
+                app.Start();
+
+                var response = await client.GetAsync("http://localhost:8795/A/../%41%CC%8A");
```
Tratcher: You can't use HttpClient to test these as it will normalize them before sending them.

test/Microsoft.AspNet.Server.KestrelTests/PathNormalizerTests.cs
```diff
@@ -0,0 +1,145 @@
...
+namespace Microsoft.AspNet.Server.KestrelTests
+{
+    public class PathNormalizerTests
+    {
+        [Theory]
+        [InlineData("/a", "/a")]
+        [InlineData("/a/", "/a/")]
+        [InlineData("/a/b", "/a/b")]
+        [InlineData("/a/b/", "/a/b/")]
+        [InlineData("", ".")]
```
Tratcher: Path should always start with `/`, so `.` and `..` are undefined, you only need to check for `/.`, `/./`, `/..`, and `/../`

cesarbs: Do you mean we should not expect requests that don't start with a `/`? The RFC accounts for the cases where it starts with `./` and `../`.

Tratcher: But not in HTTP: https://tools.ietf.org/html/rfc7230#section-5.3

cesarbs:
@benaadams Actually I was wrong to apply the normalization only to the path with percent-encoded characters in the URL. Since normalization comprises of normalization to NFC + dot segment removal, I have to check for the need for normalization in the plain text path too (because there might be dot segments in the request path).

cesarbs:
Comments addressed, perf test still pending.

cesarbs:
There's a small perf hit in the plain text benchmark:

Before:

        Latency     4.67ms   12.42ms 187.86ms   94.60%
        Req/Sec    34.92k     3.16k   61.86k    77.44%
      11197736 requests in 10.10s, 1.38GB read
    Requests/sec: 1108681.79
    Transfer/sec:    139.57MB

        Latency     3.96ms    8.65ms 127.63ms   91.75%
        Req/Sec    34.84k     3.09k   44.80k    72.59%
      11194786 requests in 10.10s, 1.38GB read
    Requests/sec: 1108504.52
    Transfer/sec:    139.54MB

        Latency     9.66ms   22.11ms 378.68ms   89.15%
        Req/Sec    34.24k     3.55k   59.86k    73.48%
      10997567 requests in 10.10s, 1.35GB read
    Requests/sec: 1088929.91
    Transfer/sec:    137.08MB

After:

        Latency     7.54ms   22.43ms 334.46ms   93.61%
        Req/Sec    33.75k     3.49k   64.74k    78.30%
      10834047 requests in 10.10s, 1.33GB read
    Requests/sec: 1072774.90
    Transfer/sec:    135.05MB

        Latency     6.06ms   15.59ms 285.80ms   92.87%
        Req/Sec    33.99k     2.79k   51.70k    74.58%
      10911305 requests in 10.10s, 1.34GB read
    Requests/sec: 1080355.79
    Transfer/sec:    136.00MB

        Latency     4.93ms   11.70ms 254.75ms   92.00%
        Req/Sec    33.52k     3.10k   49.13k    74.06%
      10763402 requests in 10.10s, 1.32GB read
    Requests/sec: 1065676.09
    Transfer/sec:    134.15MB

Ratio between worse after and best before is 0.9786.

cesarbs:
The hit is significant when decoding + normalization are needed:

Before:

    Running 10s test @ http://10.0.0.100:5001/plaintext/%41%CC%8A
        Latency     3.79ms    7.68ms 200.81ms   96.34%
        Req/Sec    29.86k     3.22k   51.04k    73.78%
      9585223 requests in 10.10s, 1.18GB read
    Requests/sec: 949067.03
    Transfer/sec:    119.47MB

    Running 10s test @ http://10.0.0.100:5001/plaintext/%41%CC%8A
        Latency     4.21ms    9.23ms 129.26ms   94.94%
        Req/Sec    30.78k     2.88k   46.97k    78.71%
      9882682 requests in 10.10s, 1.21GB read
    Requests/sec: 978584.03
    Transfer/sec:    123.19MB

    Running 10s test @ http://10.0.0.100:5001/plaintext/%41%CC%8A
        Latency     4.91ms   12.42ms 235.40ms   94.12%
        Req/Sec    31.24k     2.58k   49.29k    77.63%
      10036237 requests in 10.10s, 1.23GB read
    Requests/sec: 993710.05
    Transfer/sec:    125.09MB

After:

    Running 10s test @ http://10.0.0.100:5001/plaintext/%41%CC%8A
        Latency     3.75ms    3.77ms  94.38ms   94.26%
        Req/Sec    22.60k     1.53k   32.20k    79.50%
      7252043 requests in 10.10s, 0.89GB read
    Requests/sec: 718004.23
    Transfer/sec:     90.39MB

    Running 10s test @ http://10.0.0.100:5001/plaintext/%41%CC%8A
        Latency     3.56ms    3.10ms  58.15ms   91.99%
        Req/Sec    22.61k     1.54k   31.35k    77.02%
      7255969 requests in 10.10s, 0.89GB read
      Socket errors: connect 0, read 0, write 81, timeout 0
    Requests/sec: 718417.96
    Transfer/sec:     90.44MB

    Running 10s test @ http://10.0.0.100:5001/plaintext/%41%CC%8A
        Latency     3.51ms    2.89ms  54.24ms   88.69%
        Req/Sec    22.50k     1.49k   32.51k    78.47%
      7218541 requests in 10.10s, 0.89GB read
    Requests/sec: 714747.15
    Transfer/sec:     89.98MB

benaadams:
When just normalization?

cesarbs:
@benaadams NFC normalization will only take place in paths with percent-encoded characters.

I'm making a small change that might make a difference - I'll only try to remove dot segments if I detect any. This will save cycles and allocations.

src/Microsoft.AspNet.Server.Kestrel/Http/PathNormalizer.cs
```diff
@@ -0,0 +1,137 @@
...
+        public static bool NeedsNormalization(string path)
+        {
+            return !path.IsNormalized(NormalizationForm.FormC) || ContainsDotSegments(path);
+        }
+
+        public static string Normalize(string path)
+        {
+            if (path.IndexOf('/') > -1)
+            {
+                var normalizedChars = new char[path.Length];
```
benaadams: System.Buffers?

halter73:
Do you have any data for normalizing paths?

cesarbs:
@halter73 Gathering some.

src/Microsoft.AspNet.Server.Kestrel/Http/PathNormalizer.cs
```diff
@@ -0,0 +1,137 @@
...
+                        {
+                            skipSegments--;
+                        }
+                    }
+                }
+
+                path = new string(normalizedChars, normalizedIndex, normalizedChars.Length - normalizedIndex);
+            }
+
+            if (!path.IsNormalized(NormalizationForm.FormC))
```
Tratcher: I don't see any reason dot compression and unicode normalization have to be run at the same time. If you separate them then you can limit the unicode normalization to only run in the needDecode scenario.

src/Microsoft.AspNet.Server.Kestrel/Http/PathNormalizer.cs
benaadams: Or jump in before the path string is created and work on the byte buffer?

cesarbs: I'm not familiar with it. Any pointers/examples?

benaadams:
```csharp
using System.Buffers;
...
var normalizedChars = ArrayPool<char>.Shared.Rent(path.Length);
// stuff
ArrayPool<char>.Shared.Return(normalizedChars);
```

project.json
```json
"System.Buffers": "4.0.0-*"
```

src/Microsoft.AspNet.Server.Kestrel/Http/PathNormalizer.cs
```diff
@@ -0,0 +1,132 @@
...
+
+namespace Microsoft.AspNet.Server.Kestrel.Http
+{
+    public static class PathNormalizer
+    {
+        public static string Normalize(string path)
+        {
+            if (ContainsDotSegments(path))
+            {
+                var normalizedChars = new char[path.Length];
```
benaadams:
```csharp
using System.Buffers;
...
var normalizedChars = ArrayPool<char>.Shared.Rent(path.Length);
// stuff
ArrayPool<char>.Shared.Return(normalizedChars);
```

project.json
```json
"System.Buffers": "4.0.0-*"
```

Doesn't need try catch for the return; will just be GC'd if you don't return it.

src/Microsoft.AspNet.Server.Kestrel/Http/PathNormalizer.cs
cesarbs: You're right, will change.

cesarbs:
Got some perf back by avoiding dot segment removal when not necessaryL

    Running 10s test @ http://10.0.0.100:5001/plaintext/%41%CC%8A
        Latency     3.91ms    5.80ms 173.92ms   95.48%
        Req/Sec    24.26k     1.88k   37.44k    75.29%
      7790526 requests in 10.10s, 0.96GB read
    Requests/sec: 771378.77
    Transfer/sec:     97.11MB

    Running 10s test @ http://10.0.0.100:5001/plaintext/%41%CC%8A
        Latency     3.52ms    4.20ms 102.13ms   94.37%
        Req/Sec    24.24k     1.69k   36.02k    77.64%
      7780465 requests in 10.10s, 0.96GB read
    Requests/sec: 770358.86
    Transfer/sec:     96.98MB

    Running 10s test @ http://10.0.0.100:5001/plaintext/%41%CC%8A
        Latency     3.63ms    4.17ms 113.08ms   93.35%
        Req/Sec    24.48k     1.76k   36.00k    74.83%
      7859579 requests in 10.10s, 0.97GB read
    Requests/sec: 778191.01
    Transfer/sec:     97.96MB

@Tratcher suggested a better change which I'll implement now.

cesarbs:
New plain text numbers. A veeery small improvement with suggested changes:

        Latency     5.83ms   16.09ms 271.85ms   93.81%
        Req/Sec    34.11k     2.59k   48.42k    75.19%
      10949164 requests in 10.10s, 1.35GB read
    Requests/sec: 1084253.42
    Transfer/sec:    136.49MB

        Latency     4.68ms   10.46ms 209.49ms   93.48%
        Req/Sec    34.01k     2.61k   47.20k    76.08%
      10923702 requests in 10.10s, 1.34GB read
    Requests/sec: 1081589.99
    Transfer/sec:    136.16MB

        Latency     6.48ms   17.26ms 297.70ms   93.62%
        Req/Sec    33.99k     3.81k   60.87k    85.81%
      10857619 requests in 10.10s, 1.33GB read
    Requests/sec: 1075094.28
    Transfer/sec:    135.34MB

src/Microsoft.AspNet.Server.Kestrel/Http/PathNormalizer.cs
```diff
@@ -0,0 +1,139 @@
...
+                }
+
+                path = new string(normalizedChars, normalizedIndex, normalizedChars.Length - normalizedIndex);
+                ArrayPool<char>.Shared.Return(normalizedChars);
+            }
+
+            return path;
+        }
+
+        private static bool ContainsDotSegments(string path)
```
halter73: Can we use `MemoryPool2Iterator2.Seek` to find any '/' or '.' characters? I think we need to ask @troydai if this could work and if any utf8 characters could be normalized to a '/' or a '.'.

src/Microsoft.AspNet.Server.Kestrel/Http/Frame.cs
```diff
@@ -731,6 +731,7 @@ protected bool TakeStartLine(SocketInput input)
                     // URI was encoded, unescape and then parse as utf8
                     pathEnd = UrlPathDecoder.Unescape(pathBegin, pathEnd);
                     requestUrlPath = pathBegin.GetUtf8String(pathEnd);
+                    requestUrlPath = PathNormalizer.NormalizeToNFC(requestUrlPath);
```
benaadams: Can this operate over `pathBegin`->`pathEnd` rather than the created string? Might be messy with the Utf8

src/Microsoft.AspNet.Server.Kestrel/Http/PathNormalizer.cs
benaadams: Check back on find to see if previous byte has high byte set? (e.g. >= 128)

src/Microsoft.AspNet.Server.Kestrel/Http/Frame.cs
cesarbs: It's a lot of work. Checking if the string needs normalization requires checking some large tables of Unicode characters. Also, I'd basically have to re-implement `string.Normalize()`, and that wouldn't even save us an allocation because the native code it calls requires an output buffer to put the results i.e. it's not done in place.

src/Microsoft.AspNet.Server.Kestrel/Http/PathNormalizer.cs
cesarbs: @benaadams I don't follow. How would that work?

benaadams: @cesarbs actually you don't need to back check  
@halter73 for Url Path Encoding `%2E` is a valid `.`, however, in utf8 bytes only `.` is `.` and only `/` is `/`

Good table at start of https://en.wikipedia.org/wiki/UTF-8#Description
> * Backward compatibility: One-byte codes are used only for the ASCII values 0 through 127. In this case the UTF-8 code has the same value as the ASCII code. The high-order bit of these codes is always 0. This means that ASCII text is valid UTF-8, and UTF-8 can be used for parsers expecting 8-bit extended ASCII even if they are not designed for UTF-8.

> * Clear distinction between multi-byte and single-byte characters: Code points larger than 127 are represented by multi-byte sequences, composed of a leading byte and one or more continuation bytes. The leading byte has two or more high-order 1s followed by a 0, while continuation bytes all have '10' in the high-order position.

src/Microsoft.AspNet.Server.Kestrel/Http/Frame.cs
benaadams: Just mean it currently does:
1, `Unescape` - loop each byte 
2. `GetUtf8String` (`new char[byteLength * 2]`+ `new string()`) + loop each byte -> char
3. `NormalizeToNFC` (`ArrayPool<char>`+ `new string()`) + loop each char

Could do
1. `Unescape` - loop each byte 
2. `NormalizeToNFC` work on the memory blocks (as `.` and `\` are the same bytes in utf8)
3. `GetUtf8String` (`ArrayPool<char>`+ `new string()`)

1 & 2 could be combined, but would be messy.

halter73:
We're hoping to move the path/dotsegment normilization logic to https://github.com/aspnet/FileSystem

src/Microsoft.AspNet.Server.Kestrel/Http/Frame.cs
cesarbs: `NormalizeToNFC` is not about `.` and `/`. It's about this: http://www.macchiato.com/unicode/nfc-faq

In theory we could implement the equivalents of `string.IsNormalized` and `string.Normalize` for Kestrel, using memory blocks. Unfortunately, we don't have time for it now :disappointed:

cesarbs:
Changing `ContainsDotSegments` to use a pointer instead of an index improved things:

        Latency     5.33ms   14.74ms 263.96ms   93.40%
        Req/Sec    35.23k     3.33k   61.66k    79.25%
      11212229 requests in 10.10s, 1.38GB read
      Socket errors: connect 0, read 0, write 542, timeout 0
    Requests/sec: 1110137.07
    Transfer/sec:    139.75MB

        Latency     5.64ms   16.03ms 243.53ms   93.68%
        Req/Sec    35.10k     3.50k   63.18k    84.33%
      11226167 requests in 10.10s, 1.38GB read
    Requests/sec: 1111526.44
    Transfer/sec:    139.92MB

        Latency     4.57ms   12.28ms 236.93ms   95.26%
        Req/Sec    34.36k     2.81k   52.99k    71.87%
      11025641 requests in 10.10s, 1.36GB read
    Requests/sec: 1091646.87
    Transfer/sec:    137.42MB

src/Microsoft.AspNet.Server.Kestrel/Http/Frame.cs
benaadams: Oh... joy :stuck_out_tongue_closed_eyes:

blowdart:
Path and dot segment isn't just about file systems though. So it needs to happen higher up.

Also we need to have an unnormalized property on the request for people who want to do really weird things.

src/Microsoft.AspNet.Server.Kestrel/Http/Frame.cs
benaadams: Ah sorry was concentrating on the `PathNormalizer` bit

Tratcher:
@blowdart can we design the GetRawUrl feature later? We'd need to decide how it flows through the entire stack.

blowdart:
Sure, it can wait till after RC2

Tratcher:
Ok, @cesarbs file a separate bug for that

src/Microsoft.AspNet.Server.Kestrel/Http/Frame.cs
benaadams: Correcting

It currently does:
Utf8
1, `Unescape` - loop each byte 
2. `GetUtf8String` (`new char[byteLength * 2]`+ `new string()`) + loop each byte -> char
3. `NormalizeToNFC` string stuffs
4. `ContainsDotSegments` over string
5. `RemoveDotSegments`  (`ArrayPool<char>`+ `new string()`) + loop each char

Ascii
1. .GetAsciiString(pathEnd); `new string()`
2. `ContainsDotSegments` over string
3. `PathNormalizer.RemoveDotSegments`  `new string()`

Could do for the Utf8 route
1. `Unescape` - loop each byte 
2. `ContainsDotSegments` seek()
3. `RemoveDotSegments` loop each byte as per 1, (`.` and `\` are the same bytes in utf8)
4. `GetUtf8String` (`ArrayPool<char>`+ `new string()`)
5. `NormalizeToNFC` string stuffs 

And Ascii
1. `ContainsDotSegments` seek()
2. `RemoveDotSegments` `memoryblock`s (as per `Unescape`)
3. .GetAsciiString(pathEnd); `new string()`

cesarbs:
Filed #594.

src/Microsoft.AspNet.Server.Kestrel/Http/Frame.FeatureCollection.cs
```diff
@@ -325,5 +325,8 @@ void IHttpRequestLifetimeFeature.Abort()
         {
             Abort();
         }
+
+        // TODO: remove before merging.
```
halter73: Don't forget!

src/Microsoft.AspNet.Server.Kestrel/Http/Frame.cs
halter73: I also like the idea of using seek(). I was talking to @cesarbs, but forgot that Unescape modifies the blocks, so I forgot it could work for both utf8 and ascii strings assuming `NormalizeToNFC` doesn't convert anything to a `.` and `\`.

Considering we're only taking a ~1% perf hit for the plaintext benchmark according to @cesarbs testing, I think that this can go in as is. But if @cesarbs has time to change it to use seek, I more than welcome it.

halter73:
:shipit:

src/Microsoft.AspNet.Server.Kestrel/Http/Frame.cs
cesarbs: @benaadams It's not possible to remove dot segments without incurring extra allocations OR making significant changes to memory blocks.

There are two ways to do it (that I thought of, if you have something better in mind please share :)):

1) Move forward in the string and keep a stack of segments to figure out which ones to skip. The problem here is that we'd have to allocate a stack and incur whatever cost `Push()` and `Pop()` would impose.

2) Go backward in the string and skip segments as `..`s are found. This is what I am doing now.

We can't do 2 with blocks/iterators at this time because that would require being able to navigate back in blocks. While adding a `back` pointer to blocks isn't a super hard change, I wouldn't like to introduce that change at this time.

benaadams: @cesarbs Can always move the `ContainsDotSegments` test out of the `RemoveDotSegments`; post `Unescape` for utf8 and pre-`GetAsciiString` for ascii, use `seek()` and set a flag?

Then do `RemoveDotSegments` based on flag?

PR #577, opened 2016-01-13
Bypass LibuvStream if no ConnectionFilter wraps it

halter73:
I haven't come up with a great way to test it, but it makes the stacks way nicer when debugging our SampleApp using port 5000 (which SslStream doesn't wrap).

I know that this might technically fall under "perf", but I think it's nice to avoid all the ConnectionFilter machinery if all the filters decided to not actually wrap the connection.


Tratcher:
Oh, I thought it already worked that way... :shipit:

halter73:
It really should have... That was always my intention, but it wasn't exactly broken before, so I didn't fix it until now :disappointed:

cesarbs:
Sweet :ship:

Issue #571, opened 2016-01-11
Change to build to use dotnet


victorhurdugaci:
Kestrel has a "special" build:

- It invokes DNX as a post build step
- It relies on compilation modules
- Might (not sure) spin up a new server that's not in proc as part of the tests

cc @cesarbs @halter73

victorhurdugaci:
At a minimum, you have to upgrade a few script and project.json files just like here: https://github.com/aspnet/Configuration/commit/9cfd22f1fc48d46d9aa35e645c4daf89b7987422

victorhurdugaci:
cc @muratg

halter73:
> * It invokes DNX as a post build step

I don't think this should be a huge issue. Basically all we need to do is run two command line apps (tools/Microsoft.AspNet.Server.Kestrel.GeneratedCode and tools/Microsoft.AspNet.Server.Kestrel.GeneratedCode) before building the main project.

> * It relies on compilation modules

I think we can remove this for now as long as we have a plan to move this to [an analyzer](https://github.com/aspnet/KestrelHttpServer/pull/518/files#r49260663).

> * Might (not sure) spin up a new server that's not in proc as part of the tests

Why do you think that? I'm pretty sure that's not the case.

cesarbs:
>I think we can remove this for now as long as we have a plan to move this to an analyzer.

Why do we care about this only in Kestrel? I don't see anything similar in other projects.

davidfowl:
We don't actually. We can remove it for now and fix it later. @lodejard wrote it to keep himself in check :smile:

PR #581, opened 2016-01-15
Build with dotnet (#571)

cesarbs:
- Move to new dotnet build.
- Fix tests so they can be run in parallel (no need to pass `-parallel none` to the xunit console runner, which right now can't be done as neatly as before).
- Because we don't have `postrestore` in the new world yet, move libuv copy to `precompile`.
- Change sample apps to run with `dotnet run`.
- Fix `FindFirstByte` tests so they don't throw on `Vector<byte>.Count` > 16.
- Remove `tools\Microsoft.StandardsPolice`.

#571 

cc @victorhurdugaci @halter73 @davidfowl


test/Microsoft.AspNet.Server.KestrelTests/TestServer.cs
```diff
@@ -48,5 +53,10 @@ public void Dispose()
             _server.Dispose();
             _engine.Dispose();
         }
+
+        public static int GetNextPort()
+        {
+            return _nextPort++;
```
halter73: If the tests are running in parallel, we might need to use `Interlocked.Increment` to avoid random failures.

src/Microsoft.AspNet.Server.Kestrel/project.json
```diff
@@ -52,13 +52,11 @@
...
-      "dnx -p ../../tools/Microsoft.AspNet.Server.Kestrel.GeneratedCode run Http/FrameHeaders.Generated.cs Http/Frame.Generated.cs"
-    ],
-    "postrestore": [
-      "dnu restore ../../tools/Microsoft.AspNet.Server.Kestrel.LibuvCopier",
-      "dnx -p ../../tools/Microsoft.AspNet.Server.Kestrel.LibuvCopier run"
+    "precompile": [
+      "dotnet restore ../../tools/Microsoft.AspNet.Server.Kestrel.GeneratedCode",
+      "dotnet run -p ../../tools/Microsoft.AspNet.Server.Kestrel.GeneratedCode Http/FrameHeaders.Generated.cs Http/Frame.Generated.cs",
+      "dotnet restore ../../tools/Microsoft.AspNet.Server.Kestrel.LibuvCopier",
+      "dotnet run -p ../../tools/Microsoft.AspNet.Server.Kestrel.LibuvCopier"
```
halter73: Do these work with both coreclr/mono? Does this require running build.sh/build.cmd?

halter73:
Might want to squash the "more" commit.

test/Microsoft.AspNet.Server.KestrelTests/TestServer.cs
cesarbs: Doh, I should've thought of that. Thanks :+1:

.travis.yml
```diff
@@ -11,7 +11,7 @@ addons:
     - libunwind8
     - zlib1g
 env:
-  - KOREBUILD_DNU_RESTORE_CORECLR=true KOREBUILD_TEST_DNXCORE=true
+  - KOREBUILD_TEST_DNXCORE=true
```
victorhurdugaci: Is this still used?

src/Microsoft.AspNet.Server.Kestrel/project.json
victorhurdugaci: dotnet is a native executable. doesn't require any extension and it's only coreclr

test/Microsoft.AspNet.Server.Kestrel.FunctionalTests/PortManager.cs
```diff
@@ -0,0 +1,15 @@
...
+
+namespace Microsoft.AspNet.Server.Kestrel.FunctionalTests
+{
+    public static class PortManager
+    {
+        private static int _nextPort = 8001;
+
+        public static int GetPort()
+        {
+            return _nextPort++;
```
victorhurdugaci: Interlocked ?

test/Microsoft.AspNet.Server.KestrelTests/HttpsConnectionFilterTests.cs
```diff
@@ -311,5 +310,17 @@ public class HttpsConnectionFilterTests
 #endif
             }
         }
+
+        private static string GetTestCertificatePath()
+        {
+            if (Directory.GetCurrentDirectory().Contains(".testPublish"))
```
victorhurdugaci: This is a dangerous assumptions. Maybe in the future we'll rename this folder. A suggestion for a better approach. Look for a project.json file in the current folder. If you don't find it. look in the parent folders until you find one. Then use that as the folder for the pfx file

test/Microsoft.AspNet.Server.KestrelTests/TestServer.cs
victorhurdugaci: Don't you have this code in PortManager ?

tools/Microsoft.AspNet.Server.Kestrel.GeneratedCode/project.json
```diff
@@ -1,28 +1,25 @@
...
-    "dependencies": {
-        "Microsoft.Dnx.Compilation.CSharp.Abstractions": "1.0.0-*",
-        "Microsoft.AspNet.Http.Features": "1.0.0-*",
-        "Microsoft.AspNet.Hosting": "1.0.0-*"
+  "version": "1.0.0-*",
+  "compilationOptions": {
+    "emitEntryPoint": true
+  },
+  "dependencies": {
+    "Microsoft.Dnx.Compilation.CSharp.Abstractions": "1.0.0-*",
```
victorhurdugaci: This is going away... Not sure what's the replacement in the dotnet world

samples/LargeResponseApp/project.json
```diff
@@ -8,9 +8,11 @@
   },
   "frameworks": {
     "dnx451": { },
-    "dnxcore50": { }
+    "dnxcore50": {
+      "dependencies": {
+        "NETStandard.Library": "1.0.0-rc2-23706"
```
moozzyk: Is it OK to pin the version?

tools/Microsoft.AspNet.Server.Kestrel.LibuvCopier/Program.cs
```diff
@@ -7,33 +7,24 @@ namespace Microsoft.AspNet.Server.Kestrel.LibuvCopier
...
-
-                    if (string.IsNullOrEmpty(firstCandidate))
-                    {
-                        dnxFolder = Path.Combine(GetHome(), ".dnx");
-                    }
-                    else
-                    {
-                        dnxFolder = firstCandidate;
-                    }
+                    packagesFolder = Environment.GetEnvironmentVariable("NUGET_PACKAGES");
```
moozzyk: all this dance was required to make sure we can find packages on dev boxes and different CIs (TC/travis etc.) - make sure that CIs pass before pushing this...

samples/LargeResponseApp/project.json
victorhurdugaci: Yes. That's the recommended approach

tools/Microsoft.AspNet.Server.Kestrel.LibuvCopier/project.json
```diff
@@ -26,6 +17,7 @@
         "System.Threading": "4.0.11-*",
         "System.IO.FileSystem": "4.0.1-*"
       }
-    }
+    },
+    "dnx451": { }
```
moozzyk: Is this needed? this tool is run only with dotnet and dotnet is always CoreClr...

tools/Microsoft.AspNet.Server.Kestrel.LibuvCopier/Program.cs
```diff
@@ -7,33 +7,24 @@ namespace Microsoft.AspNet.Server.Kestrel.LibuvCopier
...
+                }
+                else
+                {
+                    packagesFolder += ";" + (Environment.GetEnvironmentVariable("NUGET_PACKAGES") ?? string.Empty);
```
moozzyk: This line is weird and prone to getting out of sync with the other occurrence of `Environment.GetEnvironmentVariable("NUGET_PACKAGES")` - you could for instance add only `;` if packagesFolder is not empty and alwasy append `(Environment.GetEnvironmentVariable("NUGET_PACKAGES")`
I also don't think that `?? string.Empty` is required.

samples/LargeResponseApp/project.json
moozzyk: OK. Then I need to fix this in IISIntegration repo where I use *

tools/Microsoft.AspNet.Server.Kestrel.LibuvCopier/Program.cs
victorhurdugaci: Isn't `;` only a Windows separator?

moozzyk: good point - I guess before we always had only one folder...

victorhurdugaci: Use `Path.PathSeparator`

src/Microsoft.AspNet.Server.Kestrel/project.json
cesarbs: @halter73 No need to run build.cmd/build.sh.

.travis.yml
cesarbs: No, but I actually meant `KOREBUILD_TEST_SKIPMONO` :grin:

halter73:
Do you need to update .travis.yml?

src/Microsoft.AspNet.Server.Kestrel/project.json
cesarbs: I'll have to move these to build.cmd/build.sh. Weird stuff happens.

.travis.yml
```diff
@@ -22,5 +22,6 @@ install:
   - cd $OLDPWD
 mono:
   - 4.0.5
+osx_image: xcode7.1
 script:
-  - ./build.sh --quiet verify
+  - ./build.sh verify
```
halter73: Travis output:

    Usage: ./build.sh -m [makefile] -n [nuget] [ [targets] ]
		   -m [makefile]  The makefile.shade to execute
		   -n [nuget]     nuget.exe
		   [targets]      A space separated list of targets to run
	The command "./build.sh verify" exited with 1.
	Done. Your build exited with 1.

cesarbs:
@halter73 Yes. I've already done some changes and there will be another one in the commit addressing comments so far.

test/Microsoft.AspNet.Server.KestrelTests/HttpsConnectionFilterTests.cs
cesarbs: I'll use `"content": [ "TestResources/testCert.pfx" ]`, similar to what I did to include `hosting.json` in the sample apps. That way I don't have to go around looking for it at runtime.

tools/Microsoft.AspNet.Server.Kestrel.LibuvCopier/project.json
cesarbs: @halter73 wants to keep it in case someone wants to run it with Mono/full CLR.

src/Microsoft.AspNet.Server.Kestrel/project.json
```diff
@@ -52,13 +52,13 @@
     "keyFile": "../../tools/Key.snk"
   },
   "scripts": {
-    "prepare": [
-      "dnu restore ../../tools/Microsoft.AspNet.Server.Kestrel.GeneratedCode",
-      "dnx -p ../../tools/Microsoft.AspNet.Server.Kestrel.GeneratedCode run Http/FrameHeaders.Generated.cs Http/Frame.Generated.cs"
+    "prepublish": [
```
cesarbs: `prepublish` should be the same as `prepare`, right?

src/Microsoft.AspNet.Server.Kestrel/project.json
```diff
@@ -52,13 +52,13 @@
...
+      "dotnet restore ../../tools/Microsoft.AspNet.Server.Kestrel.GeneratedCode",
+      "dotnet run -p ../../tools/Microsoft.AspNet.Server.Kestrel.GeneratedCode Http/FrameHeaders.Generated.cs Http/Frame.Generated.cs"
     ],
-    "postrestore": [
-      "dnu restore ../../tools/Microsoft.AspNet.Server.Kestrel.LibuvCopier",
-      "dnx -p ../../tools/Microsoft.AspNet.Server.Kestrel.LibuvCopier run"
+    "precompile": [
```
cesarbs: I have to take this out of here because it gets runs over and over. I'm thinking of moving this step to `build.cmd`/`build.sh` until we have `postrestore` in nuget. Any suggestions for alternatives?

tools/Microsoft.AspNet.Server.Kestrel.LibuvCopier/Program.cs
cesarbs: Will do. @anurse can you provide some input here? For now if there's no `DOTNET_PACKAGES` or `NUGET_PACKAGES` set, I'm defaulting to `$HOME/.dnx/packages`, but I guess I'll have to change that to `$HOME/.nuget/packages` soon, right?

tools/Microsoft.AspNet.Server.Kestrel.LibuvCopier/project.json
moozzyk: OK

tools/Microsoft.AspNet.Server.Kestrel.LibuvCopier/Program.cs
anurse: Yes, that sounds correct.

src/Microsoft.AspNet.Server.Kestrel/project.json
anurse: :+1:

cesarbs:
Trying to figure out why it hangs after running tests on appveyor.

cesarbs:
All checks successful, yay!

tools/Microsoft.AspNet.Server.Kestrel.GeneratedCode/project.json
victorhurdugaci: What about this one?

cesarbs:
@davidfowl Where do I put it?

* prepublish: bad
* precompile: bad
* build script: bad

Putting it in the makefile is a temporary solution until we have prerestore and postrestore in nuget.

davidfowl:
FYI: The CLI supports analyzers (not in project references though).

PR #583, opened 2016-01-16
Removed IsLocal from ConnectionInfo

BrennanConroy:
https://github.com/aspnet/HttpAbstractions/issues/536
@halter73 @Tratcher


BrennanConroy:
Need to update this still, wait a couple mins

muratg:
What else needs changing here?

BrennanConroy:
Needed to remove another IsLocal from HttpConnectionFeature, now it should be good. Won't pass on Travis

Tratcher:
:shipit:

halter73:
:ship:

PR #586, opened 2016-01-16
Write chunks async; unblock sync-waits directly

### LINKED ISSUE #572: Low RPS with lots of response data + chunking
I'm only able to get about 500-550 rps using the following middleware over the **loopback** interface. I can get 1200+ with a content length set.

```
private readonly byte[] bytes = Encoding.UTF8.GetBytes(new string('a', 1024));

...

app.Run(async (context) =>
{
    for (var i = 0; i < 300; i++)
    {
        await context.Response.Body.WriteAsync(bytes, 0, bytes.Length);
    }
});
```

Note that this is intentionally exceeding the write-behind buffer and going through the chunked path. This only gets the cpu to ~15%. 

This is a synthetic version of a Razor benchmark I've been trying to improve.

### LINKED ISSUE #590: Chunk writing is sync when write is async
Means the async path hits a sync-block `.GetAwaiter().GetResult()` wait

benaadams:
Resolves waiting issues observed while investigating #572
Also resolves sync-waiting issues observed in #565 (but not the connection issues #567 deals with)

Resolves #590

### DIFF
--- a/src/Microsoft.AspNet.Server.Kestrel/Http/SocketOutput.cs
+++ b/src/Microsoft.AspNet.Server.Kestrel/Http/SocketOutput.cs
@@ -48,7 +48,7 @@ namespace Microsoft.AspNet.Server.Kestrel.Http
         private int _numBytesPreCompleted = 0;
         private Exception _lastWriteError;
         private WriteContext _nextWriteContext;
-        private readonly Queue<TaskCompletionSource<object>> _tasksPending;
+        private readonly Queue<WaitingTask> _tasksPending;
         private readonly Queue<WriteContext> _writeContextPool;
         private readonly Queue<UvWriteReq> _writeReqPool;
 
@@ -68,7 +68,7 @@ namespace Microsoft.AspNet.Server.Kestrel.Http
             _connectionId = connectionId;
             _log = log;
             _threadPool = threadPool;
-            _tasksPending = new Queue<TaskCompletionSource<object>>(_initialTaskQueues);
+            _tasksPending = new Queue<WaitingTask>(_initialTaskQueues);
             _writeContextPool = new Queue<WriteContext>(_maxPooledWriteContexts);
             _writeReqPool = writeReqPool;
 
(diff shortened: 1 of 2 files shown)


benaadams:
Also see #589

halter73:
:shipit: Nice catch.

src/Microsoft.AspNet.Server.Kestrel/Http/SocketOutput.cs
```diff
@@ -634,5 +644,13 @@ public void Reset()
                 ShutdownSendStatus = 0;
             }
         }
+
+        private struct WaitingTask
+        {
+            public bool isSync;
```
halter73: Nit: capitalization.

benaadams:
Re: the threadpool queuing for async vs not for sync blocked at some future point the tcs can use `TaskCreationOptions: RunContinuationsAsynchronously` https://github.com/dotnet/coreclr/pull/2718 and it will handle both automatically.

benaadams:
The IsSync and queuing SocketOutput tasks to threadpool, *should* be able to be skipped for 4.6+ and coreclr versions later than tomorrow's. A task post RTM when there is time and the effects can be investigated? (As the current performance characteristics are known)

src/Microsoft.AspNet.Server.Kestrel/Http/SocketOutput.cs
```diff
@@ -634,5 +644,13 @@ public void Reset()
                 ShutdownSendStatus = 0;
             }
         }
+
+        private struct WaitingTask
+        {
+            public bool IsSync;
+            public int BytesToWrite;
+            public IDisposable CancellationRegistration;
```
halter73: This property isn't used yet.

benaadams: Clairvoyant? Removed...

benaadams:
Failure as `IsLocal` removal needs to work through

benaadams:
reworded commit message to retest

halter73:
Yeah it might be a while until the packages on the aspnetcidev feed are updated. A lot of our builds are failing on our internal ci due to the transition from dnx to dotnet/cli.

Until then you can change your nuget.config to reference the aspnetvolatiledev feed. That doesn't help with travis/appveyor though :cry: Blame @BrennanConroy :stuck_out_tongue:

halter73:
Thanks!

PR #588, opened 2016-01-17
StreamSocketOutput WriteAsync as async (SSL)

### LINKED ISSUE #569: StreamSocketOutput.WriteAsync is Sync (SSL)
This is used for wrapped streams such as Ssl. There is currently an issue with it being async; so it is [currently implemented as sync](https://github.com/aspnet/KestrelHttpServer/blob/dev/src/Microsoft.AspNet.Server.Kestrel/Filter/StreamSocketOutput.cs#L50-L55).

```csharp
public Task WriteAsync(ArraySegment<byte> buffer, bool immediate, bool chunk, CancellationToken cancellationToken)
{
    // TODO: Use _outputStream.WriteAsync
    Write(buffer, immediate, chunk);
}
```

Tests do catch the issue current; e.g. switching to async causes the following failures:

`Microsoft.AspNet.Server.KestrelTests.HttpsConnectionFilterTests.CanReadAndWriteWithHttpsConnectionFilter [FAIL]`
`Microsoft.AspNet.Server.KestrelTests.HttpsConnectionFilterTests.HttpsSchemePassedToRequestFeature [FAIL]`
`Microsoft.AspNet.Server.KestrelTests.HttpsConnectionFilterTests.AllowCertificateContinuesWhenNoCertificate [FAIL]`

Blockers
- [x] https://github.com/dotnet/corefx/pull/5534
- [x] https://github.com/dotnet/coreclr/pull/2724
- [x] https://github.com/dotnet/corefx/pull/5541
- [x] dotnet/corefx#7800
- [x] dotnet/corefx#7819

benaadams:
This prevents https from being async

Blocked by https://github.com/dotnet/corefx/issues/5488 / https://github.com/dotnet/corefx/issues/5077

Blockers
- [x] https://github.com/dotnet/corefx/pull/5534
- [x] https://github.com/dotnet/coreclr/pull/2724
- [x] https://github.com/dotnet/corefx/pull/5541
- [x] dotnet/corefx#7800
- [x] dotnet/corefx#7819

Resolves #569

Works on coreclr, ifdef'd for full framework

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Kestrel/Filter/StreamSocketOutput.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/Filter/StreamSocketOutput.cs
@@ -91,7 +91,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Filter
             var beginChunkBytes = ChunkWriter.BeginChunkBytes(buffer.Count);
 
             await _outputStream.WriteAsync(beginChunkBytes.Array, beginChunkBytes.Offset, beginChunkBytes.Count, cancellationToken);
-            await _outputStream.WriteAsync(buffer.Array ?? _nullBuffer, buffer.Offset, buffer.Count, cancellationToken);
+            await _outputStream.WriteAsync(buffer.Array, buffer.Offset, buffer.Count, cancellationToken);
             await _outputStream.WriteAsync(_endChunkBytes, 0, _endChunkBytes.Length, cancellationToken);
         }
 
--- a/src/Microsoft.AspNetCore.Server.Kestrel/Filter/StreamSocketOutput.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/Filter/StreamSocketOutput.cs
@@ -73,9 +73,26 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Filter
 
         public Task WriteAsync(ArraySegment<byte> buffer, bool chunk, CancellationToken cancellationToken)
         {
-            // TODO: Use _outputStream.WriteAsync
+#if NET451
             Write(buffer, chunk);
             return TaskUtilities.CompletedTask;
+#else
+            if (chunk && buffer.Array != null)
+            {
+                return WriteAsyncChunked(buffer, cancellationToken);
+            }
+
+            return _outputStream.WriteAsync(buffer.Array ?? _nullBuffer, buffer.Offset, buffer.Count, cancellationToken);
+#endif
+        }
+
+        private async Task WriteAsyncChunked(ArraySegment<byte> buffer, CancellationToken cancellationToken)
+        {
+            var beginChunkBytes = ChunkWriter.BeginChunkBytes(buffer.Count);
+
+            await _outputStream.WriteAsync(beginChunkBytes.Array, beginChunkBytes.Offset, beginChunkBytes.Count, cancellationToken);
+            await _outputStream.WriteAsync(buffer.Array ?? _nullBuffer, buffer.Offset, buffer.Count, cancellationToken);
+            await _outputStream.WriteAsync(_endChunkBytes, 0, _endChunkBytes.Length, cancellationToken);
         }
 
         public MemoryPoolIterator ProducingStart()


benaadams:
One half https://github.com/dotnet/corefx/pull/5505 though I think it actually needs the ReadAsync also (as is on bidirectional stream and I saw blocking on Read)

benaadams:
Also this is relevant https://github.com/dotnet/coreclr/pull/2724

src/Microsoft.AspNet.Server.Kestrel/Filter/StreamSocketOutput.cs
```diff
@@ -49,8 +49,41 @@ public void Write(ArraySegment<byte> buffer, bool immediate, bool chunk)
...
+            {
+                return Task.WhenAll(preambleTask, dataTask, endTask);
+            }
+            if (!dataTask.IsCompleted)
+            {
+                return Task.WhenAll(dataTask, endTask);
+            }
+            if (!endTask.IsCompleted)
+            {
+                return endTask;
```
halter73: Is it wrong to simply return `endTask` if it's not completed? Will it ever complete prior to `preambleTask` or `dataTask`?

src/Microsoft.AspNet.Server.Kestrel/Filter/StreamSocketOutput.cs
```diff
@@ -49,8 +49,41 @@ public void Write(ArraySegment<byte> buffer, bool immediate, bool chunk)
...
+        private Task WriteAsyncChunked(ArraySegment<byte> buffer, CancellationToken cancellationToken)
+        {
+            var beginChunkBytes = ChunkWriter.BeginChunkBytes(buffer.Count);
+
+            // When calls return the tasks they have copied the data to the backing buffer
+            // so is safe to send one after another and await at end
+
+            var preambleTask = _outputStream.WriteAsync(beginChunkBytes.Array, beginChunkBytes.Offset, beginChunkBytes.Count, cancellationToken);
+            var dataTask = _outputStream.WriteAsync(buffer.Array ?? _nullBuffer, buffer.Offset, buffer.Count, cancellationToken);
+            var endTask = _outputStream.WriteAsync(_endChunkBytes, 0, _endChunkBytes.Length, cancellationToken);
```
halter73: Are we not using await for perf? Do we have measurements showing this is better?

src/Microsoft.AspNet.Server.Kestrel/Filter/StreamSocketOutput.cs
benaadams: That's a good point, no

src/Microsoft.AspNet.Server.Kestrel/Filter/StreamSocketOutput.cs
```diff
@@ -62,6 +95,7 @@ public MemoryPoolIterator2 ProducingStart()
 
         public void ProducingComplete(MemoryPoolIterator2 end)
         {
+            // sync writes with is immedate == true as over Stream
```
halter73: Is this comment explaining why we don't use `WriteAsync` here? I think we should use `WriteAsync` instead of write since the normal `SocketOutput.ProducingComplete`. We probably do need to trace async errors though.

benaadams: Then it would need to return a Task which would then change the `ISocketOutput` interface although... 1sec

benaadams: No :(

Also that its `immedate: true` rather than `immedate:false` as its over `Stream` rather than `ISocketOutput` which has those overloads.

benaadams: Then it would need to return a Task which would then change the `ISocketOutput` interface; its an issue that `immedate:false` can't be passed rather than the sync write per se, as its over `Stream` rather than `ISocketOutput` which has those overloads.

benaadams: Thinking something like?
```csharp
WriteAsync(...).ContinueWith((t) => log, TaskContinuationOptions.NotOnRanToCompletion);
```

halter73: I don't think it needs to return a Task. Nothing that calls `ProducingComplete` cares about waiting for the actual write to complete. There is also no error handling logic if `ProducingComplete` fails asynchronously.

You're right that the `immediate` parameter is ignored by `StreamSocketOutput`. All writes are treated as `immediate: true`. If we wanted to change that we would need to maintain a buffer in this class internally that would only get flushed for `immediate: true` writes. I haven't done this yet, since it has no functional impact, but it could be a nice way to get a perf gain in the future.

halter73: The `ContinueWith` thing is what I was thinking.

benaadams: Its a bit ugly, but added

src/Microsoft.AspNet.Server.Kestrel/Filter/StreamSocketOutput.cs
```diff
@@ -62,18 +84,49 @@ public MemoryPoolIterator2 ProducingStart()
...
+            MemoryPoolBlock2 block;
+            try
+            {
+                // When the Async calls return the tasks they have copied the data to the backing buffer
+                // also in writing in a lock so is safe to send one after another return the final task
+
+                block = _producingBlock;
+                while (block != end.Block)
+                {
+                    _outputStream.WriteAsync(block.Data.Array, block.Data.Offset, block.Data.Count, CancellationToken.None);
```
halter73: What if the wrapping stream produces an async error?

benaadams: Will have to await them one after another too as don't know the characteristics of the underlying stream :(

halter73: Yeah. It's kinda sad.

benaadams: ...unless it was only `immediate:true` when `Flush` is called - but that's probably just asking for trouble...

benaadams: Also means have to do `.GetAwaiter().GetResult()` else the function may return in an await point... but will only be one.

src/Microsoft.AspNet.Server.Kestrel/Filter/StreamSocketOutput.cs
```diff
@@ -62,17 +81,42 @@ public MemoryPoolIterator2 ProducingStart()
...
+            }
+        }
+
+        private async Task ProducingCompleteAsync(MemoryPoolIterator2 end)
+        {
+            MemoryPoolBlock2 block;
+            try
+            {
+                // When the Async calls return the tasks they have copied the data to the backing buffer
+                // also in writing in a lock so is safe to send one after another return the final task
```
benaadams: Remove comments, not doing this now

benaadams:
Last item https://github.com/dotnet/corefx/pull/5541 has just been merged

benaadams:
Still an issue here - investigating

muratg:
:eyes:

benaadams:
@muratg this is something I really want - however there have been so many strange issues with it all the way down to coreclr; I wouldn't feel confident with it going in rushed - maybe an item to pair with WebSockets timeline? (as the current workaround does work, with caveats)

benaadams:
e.g. backlog

cesarbs:
What are we doing with this? Is it still blocked on corefx?

benaadams:
It was still jamming when I last tried it; but I haven't had time to debug to find the problem. SslStream doesn't implement true `xxxAsync` methods but uses the Stream's BeginAsync/EndAsync. 

A change was made to corefx so these methods didn't become sync in coreclr with read and write blocking each other - however I don't think that entirely fixed the issue and I observed something similar happening in full framework.

However; that may have just been a point in time feed issue.

benaadams:
Looks like this might be the final blocker https://github.com/dotnet/corefx/pull/7800

benaadams:
Rebased still waiting for dotnet/corefx#7800 and/or dotnet/corefx#7819 to hit the feed to likely work

src/Microsoft.AspNetCore.Server.Kestrel/Filter/StreamSocketOutput.cs
```diff
@@ -62,17 +80,39 @@ public MemoryPoolIterator ProducingStart()
 
         public void ProducingComplete(MemoryPoolIterator end)
         {
-            var block = _producingBlock;
-            while (block != end.Block)
+            lock (_writeLock)
             {
-                _outputStream.Write(block.Data.Array, block.Data.Offset, block.Data.Count);
+                ProducingCompleteAsync(end).GetAwaiter().GetResult();
```
halter73: I wonder if it's better not to wait for the writes to complete in `ProducingComplete`. The real `SocketOutput.ProducingComplete` doesn't block.

src/Microsoft.AspNetCore.Server.Kestrel/Filter/StreamSocketOutput.cs
```diff
@@ -49,9 +52,24 @@ public void Write(ArraySegment<byte> buffer, bool chunk)
 
         public Task WriteAsync(ArraySegment<byte> buffer, bool chunk, CancellationToken cancellationToken)
         {
-            // TODO: Use _outputStream.WriteAsync
-            Write(buffer, chunk);
-            return TaskUtilities.CompletedTask;
+            lock (_writeLock)
```
halter73: The `_writeLock` was added to this class so the chunk prefix and suffix wouldn't get intermingled given concurrent/overlapping writes. The lock doesn't really work for async writes to a stream. I'm not sure how important of a concern this really is in the first place. It might be better just to remove the lock altogether.

src/Microsoft.AspNetCore.Server.Kestrel/Filter/StreamSocketOutput.cs
```diff
@@ -62,17 +80,39 @@ public MemoryPoolIterator ProducingStart()
...
 
+                await _outputStream.WriteAsync(end.Block.Array, end.Block.Data.Offset, end.Index - end.Block.Data.Offset, CancellationToken.None);
+            }
+            catch (Exception ex)
+            {
+                _logger.LogError(ex.Message, ex);
+            }
+
+            block = _producingBlock;
+            while (block != end.Block)
```
halter73: I like this better than the `_canWrite` flag in #750

src/Microsoft.AspNetCore.Server.Kestrel/Filter/StreamSocketOutput.cs
```diff
@@ -62,17 +80,39 @@ public MemoryPoolIterator ProducingStart()
...
+                {
+                    await _outputStream.WriteAsync(block.Data.Array, block.Data.Offset, block.Data.Count, CancellationToken.None);
+                    block = block.Next;
+                }
 
+                await _outputStream.WriteAsync(end.Block.Array, end.Block.Data.Offset, end.Index - end.Block.Data.Offset, CancellationToken.None);
+            }
+            catch (Exception ex)
+            {
+                _logger.LogError(ex.Message, ex);
```
halter73: I prefer logging this as a connection error with connection id like #750.

src/Microsoft.AspNetCore.Server.Kestrel/Filter/StreamSocketOutput.cs
benaadams: Its a sync memory write though? Issue is the unknown stream type

src/Microsoft.AspNetCore.Server.Kestrel/Filter/StreamSocketOutput.cs
cesarbs: The purpose of `_canWrite` is to prevent further attempts to call `Write()`.

benaadams:
Probably needs to rebase on #750 as this is waiting for a corefx fix to hit the feeds

src/Microsoft.AspNetCore.Server.Kestrel/Filter/StreamSocketOutput.cs
halter73: The main difference is that the real ProducingComplete would never wait for the write behind buffer to flush. This might by waiting for the Task returned by SocketOutput.WriteAsync. It's a difference in behavior that we probably don't account for in the calling code.

src/Microsoft.AspNetCore.Server.Kestrel/Filter/StreamSocketOutput.cs
halter73: Makes sense.

cesarbs:
@benaadams #750 now in.

benaadams:
Issue on full framework (block)

Checking coreclr

benaadams:
Works on coreclr; have ifdef'd for full framework

Raise issue for full framework?

src/Microsoft.AspNetCore.Server.Kestrel/Filter/StreamSocketOutput.cs
benaadams: Done

src/Microsoft.AspNetCore.Server.Kestrel/Filter/StreamSocketOutput.cs
benaadams: Done

benaadams:
Issue raised for full framework https://github.com/aspnet/KestrelHttpServer/issues/768

src/Microsoft.AspNetCore.Server.Kestrel/Filter/StreamSocketOutput.cs
```diff
@@ -73,9 +73,26 @@ public void Write(ArraySegment<byte> buffer, bool chunk)
...
+            return _outputStream.WriteAsync(buffer.Array ?? _nullBuffer, buffer.Offset, buffer.Count, cancellationToken);
+#endif
+        }
+
+        private async Task WriteAsyncChunked(ArraySegment<byte> buffer, CancellationToken cancellationToken)
+        {
+            var beginChunkBytes = ChunkWriter.BeginChunkBytes(buffer.Count);
+
+            await _outputStream.WriteAsync(beginChunkBytes.Array, beginChunkBytes.Offset, beginChunkBytes.Count, cancellationToken);
+            await _outputStream.WriteAsync(buffer.Array ?? _nullBuffer, buffer.Offset, buffer.Count, cancellationToken);
```
khellang: `buffer.Array ?? _nullBuffer`: the coalescing part of this expression will never be hit. buffer.Array is checked in `WriteAsync`

benaadams: True :)

cesarbs:
Bummer that it doesn't work on desktop clr, but that's better than nothing :smile: 

:shipit:

PR #592, opened 2016-01-20
Log request and response headers

JunTaoLuo:
Item 2 of #4.

cc @cesarbs @halter73

### DIFF
--- a/src/Microsoft.AspNet.Server.Kestrel/Http/Frame.cs
+++ b/src/Microsoft.AspNet.Server.Kestrel/Http/Frame.cs
@@ -318,6 +318,59 @@ namespace Microsoft.AspNet.Server.Kestrel.Http
         /// </summary>
         public abstract Task RequestProcessingAsync();
 
+        public string RequestToString()
+        {
+            var requestBuilder = new StringBuilder("Received request: ");
+
+            // GET /path?query HTTP/1.1
+            requestBuilder.Append(Method);
+            requestBuilder.Append(" ");
+            requestBuilder.Append(PathBase);
+            requestBuilder.Append(Path);
+            requestBuilder.Append(QueryString);
+            requestBuilder.Append(" ");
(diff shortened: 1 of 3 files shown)


src/Microsoft.AspNet.Server.Kestrel/Http/Frame.cs
```diff
@@ -655,6 +708,12 @@ private Task WriteSuffix()
 
             end.CopyFrom(_httpVersion == HttpVersionType.Http1_1 ? _bytesHttpVersion1_1 : _bytesHttpVersion1_0);
             end.CopyFrom(statusBytes);
+
+            if (Log.IsEnabled(LogLevel.Debug))
+            {
+                Log.LogDebug(ResponseToString());
```
JunTaoLuo: Possibly need eventids or something more than just a string

JunTaoLuo:
I thought it was best to log the headers and such closer to when they are actually read and written but I suppose when the HttpContext was created/disposed is good enough. I'll close this PR and open a new one in Hosting and remove the logging of headers in WebListener.

halter73:
@davidfowl I think it would be a good idea to expose a connection and request ID via the IHttpConnectionFeature and IHttpRequestFeature respectively. This should make it easier to correlate logs from different layers. I can open issues in Abstractions and Kestrel if you agree.

davidfowl:
`IHttpConnectionFeature.ConnectionId` I agree with but I don't think there should be a request id on the `IHttpRequestFeature`. I prefer what we have right now as it doesn't need to be implemented in the sever itself and it's not *required* for anything to function.

Tratcher:
I expect we'll need ConnectionId at some point. We already have IHttpRequestIdentifierFeature for request id. I agree that it would make correlation easier if the Kestrel implemented IHttpRequestIdentifierFeature. WebListener does.

Issue #598, opened 2016-01-24
Chunked responses are not flushed until the response completes


halter73:
This is obviously really bad for large responses, server-sent events, etc...

benaadams:
Closed?

PR #599, opened 2016-01-24
Flush chunked writes before the entire response completes

halter73:
Addresses #598

### DIFF
--- a/tools/Microsoft.AspNetCore.Server.Kestrel.LibuvCopier/Program.cs
+++ b/tools/Microsoft.AspNetCore.Server.Kestrel.LibuvCopier/Program.cs
@@ -11,30 +11,8 @@ namespace Microsoft.AspNetCore.Server.Kestrel.LibuvCopier
         {
             try
             {
-                var packagesFolder = Environment.GetEnvironmentVariable("DNX_PACKAGES");
-
-                if (string.IsNullOrEmpty(packagesFolder))
-                {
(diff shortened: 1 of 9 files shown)


tools/Microsoft.AspNetCore.Server.Kestrel.LibuvCopier/Program.cs
```diff
@@ -11,30 +11,8 @@ public void Main(string[] args)
...
-                        dnxFolder = Path.Combine(GetHome(), ".dnx");
-                    }
-                    else
-                    {
-                        dnxFolder = firstCandidate;
-                    }
-
-                    packagesFolder = Path.Combine(dnxFolder, "packages");
-                }
+                var packagesFolder = Environment.GetEnvironmentVariable("DNX_PACKAGES") ??
```
halter73: @davidfowl Should I even be looking at `$DNX_PACKAGES` anymore? Should I just do like you did in https://github.com/aspnet/dnx/commit/a440f9ae3461d6a9f078cbcf6b741d01e7cec95c and get rid of the GetHome method?

             var profileDirectory = Environment.GetEnvironmentVariable("USERPROFILE");
             if (string.IsNullOrEmpty(profileDirectory))
             {
                  profileDirectory = Environment.GetEnvironmentVariable("HOME");
             }
  		  
             return Path.Combine(profileDirectory, ".nuget", "packages");

PR #600, opened 2016-01-24
Remove spinning for blocked socket ops

benaadams:
As per https://github.com/dotnet/corefx/pull/5655

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Kestrel/Http/SocketInput.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/Http/SocketInput.cs
@@ -16,7 +16,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Http
 
         private readonly MemoryPool2 _memory;
         private readonly IThreadPool _threadPool;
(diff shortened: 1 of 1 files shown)


src/Microsoft.AspNetCore.Server.Kestrel/Http/SocketInput.cs
```diff
@@ -16,7 +16,7 @@ public class SocketInput : ICriticalNotifyCompletion
 
         private readonly MemoryPool2 _memory;
         private readonly IThreadPool _threadPool;
-        private readonly ManualResetEventSlim _manualResetEvent = new ManualResetEventSlim(false);
+        private readonly ManualResetEventSlim _manualResetEvent = new ManualResetEventSlim(false, 0);
```
muratg: In CoreFX case, they seem to have removed `false` also... should it be the same here?

benaadams: They added `false` :)

muratg: Yup, my bad :smile:

cesarbs:
LGTM

PR #603, opened 2016-01-26
Combined fixes #567 and #599

benaadams:
Combines fixes in #567 and #599
Merge and rename clash resolved

Should build!!


benaadams:
Don't know about Travis failure; doesn't seem to be a new test that would do that

test/Microsoft.AspNet.Server.KestrelTests/TestHelpers/MockConnection.cs
```diff
@@ -0,0 +1,25 @@
+using System.Threading;
+using Microsoft.AspNet.Server.Kestrel.Http;
+using Microsoft.AspNet.Server.Kestrel.Networking;
+
+namespace Microsoft.AspNet.Server.KestrelTests.TestHelpers
+{
+    public class MockConnection : Connection
```
halter73: This class should be removed.

benaadams:
Raised bug for Travis fail https://github.com/dotnet/corefx/issues/5694

PR #608, opened 2016-02-01
Reduce lifetime of header values

### LINKED ISSUE #607: Idle connections header values may be GC gen promoted
As raised by @lodejard in https://github.com/aspnet/KestrelHttpServer/pull/497#issuecomment-165606529

> typically browsers will keep a couple connections open at a time, and you requests to arrive in small bursts with longish delays between. If reset doesn't zero the memory then all of the strings from the previous response will still be referenced by the header object during those long delays.

This effects the common headers that are not dictionary backed (which are discarded)

### LINKED ISSUE #628: Throw InvalidOperationException if the user tries to modify response headers after the body starts
Sometimes users accidentally set response headers after the response starts and it can be very hard to debug why the client did not receive the header.
https://github.com/aspnet/Mvc/issues/4065#issuecomment-181638058

Throw an InvalidOperationException if the user tries to modify response headers after the first write to the body.

WebListener already does this. Helios used to as well.

benaadams:
For common header values (non-dictionary backed ones)

* `IHttpComponentFactory` controls header and stream creation, disposal
* Headers are attached at request start and detached at request end
* Streams are attached at message body start and detached at request end
* `HttpComponentFactory` implementation pools the headers, and streams *if* `ReuseStreams == true`
* Request header values are nulled before detach (User input could be of indeterminate length)
* If connection errors in-use headers & streams are not returned to pool

Faster cycling of header objects should cause the Response header values to remain in Gen0 as they are likely to be reset to a new value on each request. 

Request values are reset if they have been set

Streams are uninitialized and reinitialized with frame reference so as not to prolong its lifetime.

Also doesn't hold onto header collections or stream objects while the Frame is in a keep-alive wait.

Resolves #607
Resolves #628
Resolves https://github.com/aspnet/KestrelHttpServer/issues/629 when set > 0

Before:
headers

With this change, setting
```
{
  "kestrel.maxPooledStreams": 128,
  "kestrel.maxPooledHeaders": 128
}
```
After memory is down by 1.4GB, headers have moved out of the top spots
headers
And instead appear quite far down
headers

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Kestrel/Http/Frame.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/Http/Frame.cs
@@ -656,8 +656,10 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Http
             byte[] statusBytes,
             bool appCompleted)
         {
-            var end = SocketOutput.ProducingStart();
             var responseHeaders = _frameHeaders.ResponseHeaders;
+            responseHeaders.SetReadOnly();
+
+            var end = SocketOutput.ProducingStart();
             if (_keepAlive)
             {
                 foreach (var connectionValue in responseHeaders.HeaderConnection)
--- a/src/Microsoft.AspNetCore.Server.Kestrel/Http/Frame.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/Http/Frame.cs
@@ -34,8 +34,6 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Http
         private static readonly byte[] _bytesHttpVersion1_1 = Encoding.ASCII.GetBytes("HTTP/1.1 ");
         private static readonly byte[] _bytesContentLengthZero = Encoding.ASCII.GetBytes("\r\nContent-Length: 0");
         private static readonly byte[] _bytesSpace = Encoding.ASCII.GetBytes(" ");
-        private static readonly byte[] _bytesServer = Encoding.ASCII.GetBytes("\r\nServer: Kestrel");
-        private static readonly byte[] _bytesDate = Encoding.ASCII.GetBytes("Date: ");
         private static readonly byte[] _bytesEndHeaders = Encoding.ASCII.GetBytes("\r\n\r\n");
 
         private static Vector<byte> _vectorCRs = new Vector<byte>((byte)'\r');
@@ -46,8 +44,9 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Http
 
         private readonly object _onStartingSync = new Object();
         private readonly object _onCompletedSync = new Object();
-        protected readonly FrameRequestHeaders _requestHeaders = new FrameRequestHeaders();
-        private readonly FrameResponseHeaders _responseHeaders = new FrameResponseHeaders();
+
+        private Headers _frameHeaders;
+        private Streams _frameStreams;
 
         protected List<KeyValuePair<Func<object, Task>, object>> _onStarting;
 
@@ -59,10 +58,6 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Http
         protected int _requestAborted;
         protected CancellationTokenSource _abortedCts;
         protected CancellationToken? _manuallySetRequestAbortToken;
-
-        internal FrameRequestStream _requestBody;
-        internal FrameResponseStream _responseBody;
-        internal FrameDuplexStream _duplexStream;
 
         protected bool _responseStarted;
         protected bool _keepAlive;
@@ -92,12 +87,6 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Http
             _localEndPoint = localEndPoint;
             _prepareRequest = prepareRequest;
             _pathBase = context.ServerAddress.PathBase;
-            if (ReuseStreams)
-            {
-                _requestBody = new FrameRequestStream();
-                _responseBody = new FrameResponseStream(this);
-                _duplexStream = new FrameDuplexStream(_requestBody, _responseBody);
-            }
 
             FrameControl = this;
             Reset();
@@ -197,8 +186,48 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Http
             get { return _responseStarted; }
         }
 
+        protected FrameRequestHeaders FrameRequestHeaders => _frameHeaders.RequestHeaders;
+
+        public Frame InitializeHeaders()
+        {
+            _frameHeaders = HttpComponentFactory.CreateHeaders(DateHeaderValueManager);
+            RequestHeaders = _frameHeaders.RequestHeaders;
+            ResponseHeaders = _frameHeaders.ResponseHeaders;
+            return this;
+        }
+
+
+        public void InitializeStreams(MessageBody messageBody)
+        {
+            _frameStreams = HttpComponentFactory.CreateStreams(this);
+
+            RequestBody = _frameStreams.RequestBody.StartAcceptingReads(messageBody);
+            ResponseBody = _frameStreams.ResponseBody.StartAcceptingWrites();
+            DuplexStream = _frameStreams.DuplexStream;
+        }
+
+        public void PauseStreams()
+        {
+            _frameStreams.RequestBody.PauseAcceptingReads();
+            _frameStreams.ResponseBody.PauseAcceptingWrites();
+        }
+
+        public void ResumeStreams()
+        {
+            _frameStreams.RequestBody.ResumeAcceptingReads();
+            _frameStreams.ResponseBody.ResumeAcceptingWrites();
+        }
+
+        public void StopStreams()
+        {
+            _frameStreams.RequestBody.StopAcceptingReads();
+            _frameStreams.ResponseBody.StopAcceptingWrites();
+        }
+
         public void Reset()
         {
+            ResetComponents(poolingPermitted: true);
+
             _onStarting = null;
             _onCompleted = null;
 
@@ -207,8 +236,6 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Http
             _autoChunk = false;
             _applicationException = null;
 
-            _requestHeaders.Reset();
-            ResetResponseHeaders();
             ResetFeatureCollection();
 
             Scheme = null;
@@ -218,13 +245,8 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Http
             Path = null;
             QueryString = null;
             _httpVersion = HttpVersionType.Unknown;
-            RequestHeaders = _requestHeaders;
-            RequestBody = null;
             StatusCode = 200;
             ReasonPhrase = null;
-            ResponseHeaders = _responseHeaders;
-            ResponseBody = null;
-            DuplexStream = null;
 
             var httpConnectionFeature = this as IHttpConnectionFeature;
             httpConnectionFeature.RemoteIpAddress = _remoteEndPoint?.Address;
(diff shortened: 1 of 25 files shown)


src/Microsoft.AspNetCore.Server.Kestrel/Http/Frame.cs
```diff
@@ -197,8 +186,48 @@ public bool HasResponseStarted
...
+
+        public void PauseStreams()
+        {
+            _frameStreams.RequestBody.PauseAcceptingReads();
+            _frameStreams.ResponseBody.PauseAcceptingWrites();
+        }
+
+        public void ResumeStreams()
+        {
+            _frameStreams.ResponseBody.ResumeAcceptingWrites();
```
muratg: nit: this is ordered differently than Pause and Stop :)

benaadams:
Maybe thread level pooling

src/Microsoft.AspNetCore.Server.Kestrel/Http/FrameHeaders.Generated.cs
```diff
@@ -3737,6 +3737,51 @@ protected override bool RemoveFast(string key)
...
+            if (((_bits & 17179869184L) != 0)) _MaxForwards = default(StringValues);
+            if (((_bits & 34359738368L) != 0)) _ProxyAuthorization = default(StringValues);
+            if (((_bits & 68719476736L) != 0)) _Referer = default(StringValues);
+            if (((_bits & 137438953472L) != 0)) _Range = default(StringValues);
+            if (((_bits & 274877906944L) != 0)) _TE = default(StringValues);
+            if (((_bits & 549755813888L) != 0)) _Translate = default(StringValues);
+            if (((_bits & 1099511627776L) != 0)) _UserAgent = default(StringValues);
+            if (((_bits & 2199023255552L) != 0)) _Origin = default(StringValues);
+            if (((_bits & 4398046511104L) != 0)) _AccessControlRequestMethod = default(StringValues);
+            if (((_bits & 8796093022208L) != 0)) _AccessControlRequestHeaders = default(StringValues);
```
halter73: Is there any evidence this is faster than simply resetting all the fields without checking which bits are set like we did before?

tools/Microsoft.AspNetCore.Server.Kestrel.GeneratedCode/KnownHeaders.cs
```diff
@@ -371,7 +371,9 @@ protected override bool RemoveFast(string key)
             return MaybeUnknown?.Remove(key) ?? false;
         }}
         protected override void ClearFast()
-        {{
+        {{{(loop.ClassName != "FrameRequestHeaders" ? "" : Each(loop.Headers, header => $@"
+            if ({header.TestBit()}) _{header.Identifier} = default(StringValues);"))}
```
halter73: Why not do this for response headers as well? I see the comment in the PR about those being controlled by the app and not the client, but it still seems nice to keep response header values out of higher generations when browsers keep connections open.

src/Microsoft.AspNetCore.Server.Kestrel/Http/FrameOfT.cs
```diff
@@ -144,6 +135,8 @@ public class Frame<TContext> : Frame
             }
             finally
             {
+                // Error occurred, do not return components to pool
+                ResetComponents(poolingPermitted: false);
```
halter73: An error didn't necessarily occur here, right? The client could have just closed the connection. Why prevent pooling even if there was an error?

halter73: Actually, I think I understand now: you're afraid of buggy apps. I think header pooling should also be opt-in if we want to support it at all.

halter73: OK. I realize this isn't much worse than it was before since we were already reusing headers between requests on the same connection. I still think having shared configuration for reusing stream/header objects would be nice since in principle it's pretty similar.

 My point still stands about an error not necessarily having occurred here even if the headers are initialized. We have a couple places where we do:

    if (SocketInput.RemoteIntakeFin)
    {
        return;
    }

src/Microsoft.AspNetCore.Server.Kestrel/Http/Frame.cs
```diff
@@ -197,8 +186,48 @@ public bool HasResponseStarted
             get { return _responseStarted; }
         }
 
+        protected FrameRequestHeaders FrameRequestHeaders => _frameHeaders.RequestHeaders;
+
+        public Frame InitalizeHeaders()
```
halter73: Spelling

src/Microsoft.AspNetCore.Server.Kestrel/Http/Frame.cs
```diff
@@ -197,8 +186,48 @@ public bool HasResponseStarted
...
+        {
+            _frameHeaders = HttpComponentFactory.CreateHeaders(DateHeaderValueManager);
+            RequestHeaders = _frameHeaders.RequestHeaders;
+            ResponseHeaders = _frameHeaders.ResponseHeaders;
+            return this;
+        }
+
+
+        public void InitalizeStreams(MessageBody messageBody)
```
halter73: ditto

Tratcher:
Hey @benaadams, while you're at it... https://github.com/aspnet/KestrelHttpServer/issues/628

src/Microsoft.AspNetCore.Server.Kestrel/Http/Frame.cs
```diff
@@ -239,15 +261,26 @@ public void Reset()
...
-                DateHeaderValueManager.GetDateHeaderValue(),
-                DateHeaderValueManager.GetDateHeaderValueBytes());
-            _responseHeaders.SetRawServer(
-                "Kestrel",
-                _bytesServer);
+            if (_frameHeaders != null)
+            {
+                RequestHeaders = null;
+                ResponseHeaders = null;
+                var frameHeaders = _frameHeaders;
```
halter73: I'm not sure this pattern is really achieving anything.

tools/Microsoft.AspNetCore.Server.Kestrel.GeneratedCode/KnownHeaders.cs
benaadams: I was assuming that response header use are mostly the same per request and with them cycling much faster (per-request rather than per-"request on connection") they will get naturally cleared out in Gen0.

davidfowl: Measure measure measure

halter73: Sorry. What makes the response headers cycle much faster?

benaadams: The component factory is set at the `new KestrelEngine(new ServiceContext` level so shared across connections

benaadams: However with requests, the external-user could stick really long strings in rarely used headers like `Warning` (guessing) and it would hang around if not cleared.

src/Microsoft.AspNetCore.Server.Kestrel/Http/Frame.cs
benaadams: Disconnecting headers prior to pooling? In case of race?

src/Microsoft.AspNetCore.Server.Kestrel/Http/FrameHeaders.Generated.cs
benaadams: Could do it in one statement by putting them in a struct and doing `_headers = default(Headers)`

src/Microsoft.AspNetCore.Server.Kestrel/Http/MessageBody.cs
```diff
@@ -35,7 +36,7 @@ public ValueTask<int> ReadAsync(ArraySegment<byte> buffer, CancellationToken can
             return result;
         }
 
-        public async Task Consume(CancellationToken cancellationToken = default(CancellationToken))
+        public Task Consume(CancellationToken cancellationToken = default(CancellationToken))
```
benaadams: This shouldn't be in this set

src/Microsoft.AspNetCore.Server.Kestrel/Http/FrameOfT.cs
benaadams: Betterizing it

src/Microsoft.AspNetCore.Server.Kestrel/Http/Frame.cs
benaadams: Isn't race is to detect double call

src/Microsoft.AspNetCore.Server.Kestrel/Http/FrameOfT.cs
benaadams: Use numerics now for pool sizes, when zero it doesn't pool; by default is zero.

Only doesn't return to pool on error.

benaadams:
Need to look at clearing methods
> Is there any evidence this is faster than simply resetting all the fields without checking which bits are set like we did before?

benaadams:
Update description with measurements

benaadams:
Added fast header clear. Also you will need to switch on some pooling for high perf as headers now are  refreshed per request rather than per connection.

benaadams:
So with the inner struct clearing out all the header strings in ClearFast isn't much il

```csharp
  IL_0009:  ldflda     valuetype HeaderReferences FrameRequestHeaders::_headers
  IL_000e:  initobj HeaderReferences
```
Full function
```csharp
.method family hidebysig virtual instance void 
        ClearFast() cil managed
{
  // Code size       37 (0x25)
  .maxstack  8
  IL_0000:  ldarg.0
  IL_0001:  ldc.i4.0
  IL_0002:  conv.i8
  IL_0003:  stfld      int64 Microsoft.AspNetCore.Server.Kestrel.Http.FrameRequestHeaders::_bits
  IL_0008:  ldarg.0
  IL_0009:  ldflda     valuetype Microsoft.AspNetCore.Server.Kestrel.Http.FrameRequestHeaders/HeaderReferences Microsoft.AspNetCore.Server.Kestrel.Http.FrameRequestHeaders::_headers
  IL_000e:  initobj    Microsoft.AspNetCore.Server.Kestrel.Http.FrameRequestHeaders/HeaderReferences
  IL_0014:  ldarg.0
  IL_0015:  ldfld      class [System.Collections]System.Collections.Generic.Dictionary`2<string,valuetype [Microsoft.Extensions.Primitives]Microsoft.Extensions.Primitives.StringValues> Microsoft.AspNetCore.Server.Kestrel.Http.FrameHeaders::MaybeUnknown
  IL_001a:  dup
  IL_001b:  brtrue.s   IL_001f
  IL_001d:  pop
  IL_001e:  ret
  IL_001f:  call       instance void class [System.Collections]System.Collections.Generic.Dictionary`2<string,valuetype [Microsoft.Extensions.Primitives]Microsoft.Extensions.Primitives.StringValues>::Clear()
  IL_0024:  ret
} // end of method FrameRequestHeaders::ClearFast
```

halter73:
:ship:

src/Microsoft.AspNetCore.Server.Kestrel/Http/Frame.cs
halter73: Actually you still need to fix this spelling

src/Microsoft.AspNetCore.Server.Kestrel/Http/Frame.cs
halter73: ditto

src/Microsoft.AspNetCore.Server.Kestrel/Http/Frame.cs
```diff
@@ -197,8 +186,48 @@ public bool HasResponseStarted
...
         }
 
+        protected FrameRequestHeaders FrameRequestHeaders => _frameHeaders.RequestHeaders;
+
+        public Frame InitalizeHeaders()
+        {
+            _frameHeaders = HttpComponentFactory.CreateHeaders(DateHeaderValueManager);
+            RequestHeaders = _frameHeaders.RequestHeaders;
+            ResponseHeaders = _frameHeaders.ResponseHeaders;
+            return this;
```
halter73: Don't return `this`. I don't see you using this anywhere anyway.

halter73: On second thought, don't worry about this. I changed it while addressing my own PR feedback readying a merge.

Issue #520, opened 2015-12-22
System.IO.InvalidDataException in Kestrel.Http.MessageBody when posting large file


davenewza:
When trying to post a large file  to an API action using form-data, the following exception is thrown. I have found this to be around 1.5mb but appears to be inconsistent. 

This only seems to happen when hosting in an Azure Web App using the IISPlatformHandler. 

I am using ASP.NET 5 RC1 (1.0.0-rc1-final) packages.

Action:

        [HttpPost()]
        public IActionResult PostText()
        {
            if (Request.Body.CanSeek)
            {
                Request.Body.Position = 0;
            }

            var body = new StreamReader(Request.Body).ReadToEnd();

            ...
        }

[...]

PR #619, opened 2016-02-08
Consume chunked request fully

### LINKED ISSUE #617: Chunk encoding request will corrupt state for next reqeust
A sequence of two requests (sequential, not concurrency involved) will cause Kestrel to generate invalid HTTP method for the 3rd request.

* GET http://localhost:8080/replication/topology HTTP/1.1
* PUT http://localhost:8080/admin/databases/CanSaveAndLoad HTTP/1.1
 * Note that this request is chunked, and we are sending 147 bytes
 * The actual size doesn't seems to matter, only that it is chunked.
* GET http://localhost:8080/databases/CanSaveAndLoad/docs?&id=Raven%2FHilo%2Fusers&id=Raven%2FServerPrefixForHilo HTTP/1.1
 * *corruption* The GET method here is invalid. Instead of being 'GET', it is '\r\nGET'

From the behavior, I'm assuming that you are reading the remains of the previous request. 

Repro code is here: https://gist.github.com/ayende/989682910c8312629473
This is against "Microsoft.AspNet.Server.Kestrel": "1.0.0-rc1-final",
I tried to test against the current dev branch, but I couldn't build it properly (I'll create another issue about that).

benaadams:
Chunked requests have an extra `\r\n` at end

```
5\r\n
Hello\r\n
6\r\n
World\r\n
0\r\n
... Trailer headers ...
\r\n
```

Resolves #617

@Tratcher is this correct?

/cc @cesarbs @halter73 

Please note changes in tests

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Kestrel/Http/Frame.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/Http/Frame.cs
@@ -5,7 +5,6 @@ using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
-using System.Net;
 using System.Numerics;
 using System.Text;
 using System.Threading;
@@ -13,6 +12,7 @@ using System.Threading.Tasks;
 using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.Http.Features;
 using Microsoft.AspNetCore.Server.Kestrel.Infrastructure;
+using Microsoft.AspNetCore.Server.Kestrel.Exceptions;
 using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Primitives;
 
@@ -45,7 +45,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Http
         private readonly object _onStartingSync = new Object();
         private readonly object _onCompletedSync = new Object();
 
-        protected bool _poolingPermitted = true;
+        protected bool _corruptedRequest = false;
         private Headers _frameHeaders;
         private Streams _frameStreams;
 
@@ -211,7 +211,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Http
 
         public void Reset()
         {
-            ResetComponents(poolingPermitted: true);
+            ResetComponents();
 
             _onStarting = null;
             _onCompleted = null;
@@ -248,27 +248,23 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Http
             _abortedCts = null;
         }
 
-        protected void ResetComponents(bool poolingPermitted)
+        protected void ResetComponents()
         {
-            if (_frameHeaders != null)
+            var frameHeaders = Interlocked.Exchange(ref _frameHeaders, null);
+            if (frameHeaders != null)
             {
-                var frameHeaders = _frameHeaders;
-                _frameHeaders = null;
-
                 RequestHeaders = null;
                 ResponseHeaders = null;
-                HttpComponentFactory.DisposeHeaders(frameHeaders, poolingPermitted);
+                HttpComponentFactory.DisposeHeaders(frameHeaders);
             }
 
-            if (_frameStreams != null)
+            var frameStreams = Interlocked.Exchange(ref _frameStreams, null);
+            if (frameStreams != null)
             {
-                var frameStreams = _frameStreams;
-                _frameStreams = null;
-
                 RequestBody = null;
                 ResponseBody = null;
                 DuplexStream = null;
-                HttpComponentFactory.DisposeStreams(frameStreams, poolingPermitted);
+                HttpComponentFactory.DisposeStreams(frameStreams);
             }
         }
 
@@ -568,8 +564,19 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Http
 
         protected Task ProduceEnd()
         {
-            if (_applicationException != null)
+            if (_corruptedRequest || _applicationException != null)
             {
+                if (_corruptedRequest)
+                {
+                    // 400 Bad Request
+                    StatusCode = 400;
+                } 
+                else
+                {
+                    // 500 Internal Server Error
+                    StatusCode = 500;
+                }
+
                 if (_responseStarted)
                 {
                     // We can no longer respond with a 500, so we simply close the connection.
@@ -578,7 +585,6 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Http
                 }
                 else
                 {
-                    StatusCode = 500;
                     ReasonPhrase = null;
 
                     var responseHeaders = _frameHeaders.ResponseHeaders;
@@ -711,7 +717,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Http
             {
                 string method;
                 var begin = scan;
-                if (!begin.GetKnownMethod(ref scan,out method))
+                if (!begin.GetKnownMethod(ref scan, out method))
                 {
                     if (scan.Seek(ref _vectorSpaces) == -1)
                     {
@@ -834,7 +840,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Http
             return true;
         }
 
-        public static bool TakeMessageHeaders(SocketInput input, FrameRequestHeaders requestHeaders)
+        public bool TakeMessageHeaders(SocketInput input, FrameRequestHeaders requestHeaders)
         {
             var scan = input.ConsumingStart();
             var consumed = scan;
@@ -863,7 +869,10 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Http
                             consumed = scan;
                             return true;
                         }
-                        throw new InvalidDataException("Malformed request");
+
+                        ReportCorruptedHttpRequest(new BadHttpRequestException("Headers corrupted, invalid header sequence."));
+                        // Headers corrupted, parsing headers is complete
+                        return true;
                     }
 
                     while (
@@ -953,16 +962,27 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Http
                    statusCode != 304;
         }
 
+        public void ReportCorruptedHttpRequest(BadHttpRequestException ex)
+        {
+            _corruptedRequest = true;
+            Log.ConnectionBadRequest(ConnectionId, ex);
+        }
+
         protected void ReportApplicationError(Exception ex)
         {
             if (_applicationException == null)
             {
                 _applicationException = ex;
             }
+            else if (_applicationException is AggregateException)
+            {
+                _applicationException = new AggregateException(_applicationException, ex).Flatten();
+            }
             else
             {
                 _applicationException = new AggregateException(_applicationException, ex);
             }
+
             Log.ApplicationError(ex);
         }
 
--- a/src/Microsoft.AspNetCore.Server.Kestrel/Http/MessageBody.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/Http/MessageBody.cs
@@ -7,6 +7,7 @@ using System.Numerics;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Server.Kestrel.Infrastructure;
+using Microsoft.AspNetCore.Server.Kestrel.Exceptions;
 
 namespace Microsoft.AspNetCore.Server.Kestrel.Http
 {
@@ -118,10 +119,19 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Http
                 return new ForChunkedEncoding(keepAlive, headers, context);
             }
 
-            var contentLength = headers.HeaderContentLength.ToString();
-            if (contentLength.Length > 0)
+            var unparsedContentLength = headers.HeaderContentLength.ToString();
+            if (unparsedContentLength.Length > 0)
             {
-                return new ForContentLength(keepAlive, int.Parse(contentLength), context);
+                int contentLength;
+                if (!int.TryParse(unparsedContentLength, out contentLength) || contentLength < 0)
+                {
+                    context.ReportCorruptedHttpRequest(new BadHttpRequestException("Invalid content length."));
+                    return new ForContentLength(keepAlive, 0, context);
+                }
+                else
+                {
+                    return new ForContentLength(keepAlive, contentLength, context);
+                }
             }
 
             if (keepAlive)
@@ -130,6 +140,15 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Http
             }
 
             return new ForRemainingData(context);
+        }
+
+        private int ThrowBadRequestException(string message)
+        {
+            // returns int so can be used as item non-void function
+            var ex =  new BadHttpRequestException(message);
+            _context.ReportCorruptedHttpRequest(ex);
+
+            throw ex;
         }
 
         private class ForRemainingData : MessageBody
@@ -177,7 +196,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Http
                     _inputLength -= actual;
                     if (actual == 0)
                     {
-                        throw new InvalidDataException("Unexpected end of request content");
+                        ThrowBadRequestException("Unexpected end of request content");
                     }
                     return actual;
                 }
@@ -193,7 +212,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Http
                 _inputLength -= actual;
                 if (actual == 0)
                 {
-                    throw new InvalidDataException("Unexpected end of request content");
+                    ThrowBadRequestException("Unexpected end of request content");
                 }
 
                 return actual;
@@ -290,7 +309,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Http
 
                 if (_mode == Mode.TrailerHeaders)
                 {
-                    while (!Frame.TakeMessageHeaders(input, _requestHeaders))
+                    while (!_context.TakeMessageHeaders(input, _requestHeaders))
                     {
                         await GetDataAsync(input);
                     }
@@ -451,7 +470,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Http
                     }
                     else
                     {
-                        ThrowInvalidFormat();
+                        ThrowBadRequestException("Bad chunk suffix");
                     }
                 }
                 finally
@@ -489,7 +508,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Http
                 }
             }
 
-            private static int CalculateChunkSize(int extraHexDigit, int currentParsedSize)
+            private int CalculateChunkSize(int extraHexDigit, int currentParsedSize)
             {
                 checked
                 {
@@ -507,35 +526,24 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Http
                     }
                     else
                     {
-                        return ThrowInvalidFormat();
+                        return ThrowBadRequestException("Bad chunk size data");
                     }
                 }
             }
 
-            private static SocketInput GetDataAsync(SocketInput input)
+            private SocketInput GetDataAsync(SocketInput input)
             {
                 ThrowIfRequestIncomplete(input);
 
                 return input;
             }
 
-            private static void ThrowIfRequestIncomplete(SocketInput input)
+            private void ThrowIfRequestIncomplete(SocketInput input)
             {
                 if (input.RemoteIntakeFin)
                 {
-                    ThrowRequestIncomplete();
+                    ThrowBadRequestException("Chunked request incomplete");
                 }
-            }
-
-            private static int ThrowInvalidFormat()
-            {
-                // returns int so can be used as item non-void function
-                throw new InvalidOperationException("Bad request");
-            }
-
-            private static void ThrowRequestIncomplete()
-            {
-                throw new InvalidOperationException("Chunked request incomplete");
             }
 
             private enum Mode
--- a/src/Microsoft.AspNetCore.Server.Kestrel/Http/MessageBody.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/Http/MessageBody.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.IO;
+using System.Numerics;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Server.Kestrel.Infrastructure;
@@ -11,10 +12,10 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Http
 {
     public abstract class MessageBody
     {
-        private readonly FrameContext _context;
+        private readonly Frame _context;
         private int _send100Continue = 1;
 
-        protected MessageBody(FrameContext context)
+        protected MessageBody(Frame context)
         {
             _context = context;
         }
@@ -99,7 +100,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Http
         public static MessageBody For(
             string httpVersion,
             FrameRequestHeaders headers,
-            FrameContext context)
+            Frame context)
         {
             // see also http://tools.ietf.org/html/rfc2616#section-4.4
 
@@ -114,7 +115,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Http
             var transferEncoding = headers.HeaderTransferEncoding.ToString();
             if (transferEncoding.Length > 0)
             {
-                return new ForChunkedEncoding(keepAlive, context);
+                return new ForChunkedEncoding(keepAlive, headers, context);
             }
 
             var contentLength = headers.HeaderContentLength.ToString();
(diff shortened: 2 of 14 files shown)


src/Microsoft.AspNetCore.Server.Kestrel/Http/MessageBody.cs
```diff
@@ -259,6 +271,35 @@ public override ValueTask<int> ReadAsyncImplementation(ArraySegment<byte> buffer
...
+                    }
+                    else if (ch1 == '\r' && ch2 == '\n')
+                    {
+                        input.ConsumingComplete(scan, scan);
+                        _mode = Mode.Complete;
+                    }
+                    else
+                    {
+                        // Post request headers
+                        throw new NotImplementedException("INVALID REQUEST FORMAT");
```
cesarbs: I don't think `NotImplementedException` is the right exception to throw here. `InvalidOperationException` seems more appropriate.

src/Microsoft.AspNetCore.Server.Kestrel/Http/MessageBody.cs
```diff
@@ -298,9 +339,9 @@ private static bool TakeChunkedLine(SocketInput baton, ref int chunkSizeOut)
                             {
                                 throw new NotImplementedException("INVALID REQUEST FORMAT");
```
cesarbs: My comment regarding the exception type applies here too.

src/Microsoft.AspNetCore.Server.Kestrel/Http/MessageBody.cs
```diff
@@ -329,7 +370,7 @@ private static bool TakeChunkedLine(SocketInput baton, ref int chunkSizeOut)
...
```
cesarbs: Same (even though it was already there).

test/Microsoft.AspNetCore.Server.KestrelTests/EngineTests.cs
```diff
@@ -999,7 +1005,9 @@ public void ConnectionCanReadAndWrite(TestServiceContext testContext)
                         "HelloPOST / HTTP/1.1",
                         "Transfer-Encoding: chunked",
                         "",
-                        "C", "HelloChunked", "0",
+                        "C", "HelloChunked", 
+                        "0", "",
+                        "",
```
Tratcher: This appears to be testing for:
```
0\r\n
\r\n
\r\n
```
which is invalid. It should be:
```
0\r\n
\r\n
```
Or:
```
0\r\n
CustomHeader: custom value\r\n
\r\n
```

src/Microsoft.AspNetCore.Server.Kestrel/Http/MessageBody.cs
benaadams: This though could be valid http (trailing headers)

test/Microsoft.AspNetCore.Server.KestrelTests/EngineTests.cs
benaadams: Fixed

benaadams:
Resolved issues. @cesarbs question on whether to go `NotImplementedException` vs `InvalidOperationException` for unsupported but valid features (e.g. Chunked-extensions & Trailing headers)

src/Microsoft.AspNetCore.Server.Kestrel/Http/MessageBody.cs
```diff
@@ -255,8 +267,36 @@ public override ValueTask<int> ReadAsyncImplementation(ArraySegment<byte> buffer
...
+                        if (input.RemoteIntakeFin)
+                        {
+                            ThrowChunkedRequestIncomplete();
                         }
+
+                        await input;
+                    }
+                    else if (ch1 == '\r' && ch2 == '\n')
+                    {
+                        input.ConsumingComplete(scan, scan);
```
halter73: `ConsumingComplete` should be called in a finally block. We're not consistent about doing that in this class (I'll fix that soon), but that is the recommended calling pattern.

test/Microsoft.AspNetCore.Server.KestrelTests/EngineTests.cs
```diff
@@ -301,11 +301,14 @@ public void ConnectionCanReadAndWrite(TestServiceContext testContext)
             {
                 using (var connection = new TestConnection(server.Port))
                 {
-                    await connection.Send(
```
halter73: Maybe we should add a ChunkedRequestsTests class much like the ChunkedResponseTests one. EngineTests is getting kinda large. We could add a test method for trailing headers while we're at it.

src/Microsoft.AspNetCore.Server.Kestrel/Http/MessageBody.cs
```diff
@@ -178,6 +178,7 @@ public override ValueTask<int> ReadAsyncImplementation(ArraySegment<byte> buffer
         private class ForChunkedEncoding : MessageBody
         {
             private int _inputLength;
+            private int _chunkSize;
```
halter73: Out of curiosity, is there a reason for making this a field instead of a local variable? It seems that it's still only being used when `_mode == Mode.ChunkPrefix`

src/Microsoft.AspNetCore.Server.Kestrel/Http/MessageBody.cs
Tratcher: Oh, if you're not going to process incoming trailer headers you should drain them and file a bug to deal with them later.

src/Microsoft.AspNetCore.Server.Kestrel/Http/MessageBody.cs
benaadams: was important when `ChunkPrefix` -> `ChunkSuffix` to consume the trailer to detect the type after a return from `ChunkData`, isn't important now - will revert.

src/Microsoft.AspNetCore.Server.Kestrel/Http/MessageBody.cs
```diff
@@ -353,12 +394,32 @@ private static bool TakeChunkedLine(SocketInput baton, ref int chunkSizeOut)
                 }
             }
 
+            private static void ThrowInvalidFormat()
+            {
+                throw new InvalidOperationException("INVALID REQUEST FORMAT");
```
cesarbs: Can you switch to lowercase? Not sure why it was all uppercase to begin with, but since you're changing this it would be nice to be consistent (and better looking :smile:) with the rest.

benaadams: All caps is LogLevel.CRITICAL! ;-)

src/Microsoft.AspNetCore.Server.Kestrel/Http/MessageBody.cs
benaadams: I think the latest commit fixes this - but I can't compile it due to 

`Unable to resolve Microsoft.DotNet.CoreHost (>= 0.0.1-beta-00001) for DNXCore,Version=v5.0`

benaadams:
Capturing comment hidden in outdated diff
> Oh, if you're not going to process incoming trailer headers you should drain them and file a bug to deal with them later.

test/Microsoft.AspNetCore.Server.KestrelTests/EngineTests.cs
benaadams: Just checking if this can be reverted back to `Send`; would prefer to keep the test the same (other than correcting the input)

benaadams: Testing hangs without this change, but also putting it in matches the other tests - so am ok with it :)

benaadams:
Now parses the trailing headers /cc @davidfowl

benaadams:
A test

test/Microsoft.AspNetCore.Server.KestrelTests/ChunkedRequestTests.cs
```diff
@@ -0,0 +1,294 @@
...
+                Assert.Equal(string.Empty, request.Headers["X-Trailer-Header"]);
+
+                while(await request.Body.ReadAsync(buffer, 0, buffer.Length) != 0)
+                {
+                    // read to end
+                }
+
+                if (requestsReceived < requestCount)
+                {
+                    Assert.Equal(new string('a', requestsReceived), request.Headers["X-Trailer-Header"]);
```
halter73: Cool. Now I'm interested to see if this feature gets used even once in the wild. You should find a way to take advantage of this in one of your games :wink:

src/Microsoft.AspNetCore.Server.Kestrel/Http/MessageBody.cs
```diff
@@ -353,12 +462,34 @@ private static bool TakeChunkedLine(SocketInput baton, ref int chunkSizeOut)
...
+            }
+
+            private static void ThrowChunkedExtensionsNotSupported()
+            {
+                throw new NotImplementedException("Chunked-extensions not supported");
+            }
+
+            private static void ThrowTrailingHeadersNotSupported()
+            {
+                throw new NotImplementedException("Trailing headers not supported");
```
halter73: This error needs to be changed.

src/Microsoft.AspNetCore.Server.Kestrel/Http/MessageBody.cs
```diff
@@ -353,12 +462,34 @@ private static bool TakeChunkedLine(SocketInput baton, ref int chunkSizeOut)
...
             }
 
+            private static void ThrowInvalidFormat()
+            {
+                throw new InvalidOperationException("Bad Request");
+            }
+
+            private static void ThrowChunkedRequestIncomplete()
+            {
+                throw new InvalidOperationException("Chunked request incomplete");
```
halter73: Maybe this and the above would be better as FormatExceptions.

cesarbs: I'm not sure. `FormatException` seems more appropriate in places where you're using format strings, or parsing a string to convert it to some other type.

In our case I think we should use:

* `InvalidOperationException` where things are just wrong (like incomplete request).
* `NotSupportedException` when something is not supported (like chunked extensions).

src/Microsoft.AspNetCore.Server.Kestrel/Http/Frame.cs
```diff
@@ -46,7 +46,7 @@ public abstract partial class Frame : FrameContext, IFrameControl
 
         private readonly object _onStartingSync = new Object();
         private readonly object _onCompletedSync = new Object();
-        protected readonly FrameRequestHeaders _requestHeaders = new FrameRequestHeaders();
+        internal readonly FrameRequestHeaders _requestHeaders = new FrameRequestHeaders();
```
cesarbs: Why?

src/Microsoft.AspNetCore.Server.Kestrel/Http/MessageBody.cs
```diff
@@ -353,12 +462,34 @@ private static bool TakeChunkedLine(SocketInput baton, ref int chunkSizeOut)
                 }
             }
 
+            private static void ThrowInvalidFormat()
+            {
+                throw new InvalidOperationException("Bad Request");
```
cesarbs: nit for consistency: `Request` -> `request`

test/Microsoft.AspNetCore.Server.KestrelTests/ChunkedRequestTests.cs
```diff
@@ -0,0 +1,294 @@
...
+
+            public bool IsEnabled(LogLevel logLevel)
+            {
+                return true;
+            }
+
+            public void Log<TState>(LogLevel logLevel, EventId eventId, TState state, Exception exception, Func<TState, Exception, string> formatter)
+            {
+                // Application errors are logged using 13 as the eventId.
+                if (eventId.Id == 13)
```
cesarbs: Use a `const`.

test/Microsoft.AspNetCore.Server.KestrelTests/ChunkedRequestTests.cs
```diff
@@ -0,0 +1,294 @@
...
+        {
+            public Task OnConnectionAsync(ConnectionFilterContext context)
+            {
+                context.Connection = new LoggingStream(context.Connection, new TestApplicationErrorLogger());
+                return TaskUtilities.CompletedTask;
+            }
+        }
+    }
+}
+
```
cesarbs: Please add some tests for error conditions (invalid requests).

src/Microsoft.AspNetCore.Server.Kestrel/Http/MessageBody.cs
benaadams: Introduce some new errors? `HttpErrorException(numeric, message, deets)` wondering how easy it is to currently distinguish things and their various levels of meh -> serious

src/Microsoft.AspNetCore.Server.Kestrel/Http/Frame.cs
benaadams: So the chunked trailers can use the static method `while (!Frame.TakeMessageHeaders(input, frame._requestHeaders))`

src/Microsoft.AspNetCore.Server.Kestrel/Http/MessageBody.cs
benaadams: Ah `NotSupportedException` is much better than `NotImplementedException`

src/Microsoft.AspNetCore.Server.Kestrel/Http/Frame.cs
halter73: Might be better to make TakeMessageHeaders non-static. I think it only changes one test in FrameTests that now needs to create a frame, but that isn't to hard (see FrameFacts... I'm not sure why that is a separate class).

test/Microsoft.AspNetCore.Server.KestrelTests/ChunkedRequestTests.cs
benaadams: Might be a stretch even for me :wink: 

Linked to discussion from "Web Hypertext Application Technology Working Group" (whatwg) in https://github.com/aspnet/KestrelHttpServer/issues/622 - think it needs a custom client rather than browser.

src/Microsoft.AspNetCore.Server.Kestrel/Http/Frame.cs
benaadams: Actually headers are passed in; can swap this

test/Microsoft.AspNetCore.Server.KestrelTests/ChunkedRequestTests.cs
benaadams: Having issues with feeds and running tests (to test the tests) https://github.com/aspnet/Home/issues/1289

src/Microsoft.AspNetCore.Server.Kestrel/Http/MessageBody.cs
cesarbs: Let's go with that :+1:

benaadams:
I'm not sure this is right... is doing what it was doing before other than consuming chunked extensions; but the `TakeChunkedLine` function is weird and doesn't seem to interact properly with the flow.

benaadams:
I'm not sure this is right... is doing what it was doing before other than consuming chunked extensions; but the `TakeChunkedLine` function is weird and doesn't seem to interact properly with the flow.

May be what's causing https://github.com/aspnet/KestrelHttpServer/issues/520

test/Microsoft.AspNetCore.Server.KestrelTests/ChunkedRequestTests.cs
halter73: Have you gotten to this?

test/Microsoft.AspNetCore.Server.KestrelTests/ChunkedRequestTests.cs
```diff
@@ -0,0 +1,296 @@
...
+            var bytes = data.ToArray();
+
+            response.Headers.Clear();
+            response.Headers["Content-Length"] = bytes.Length.ToString();
+            await response.Body.WriteAsync(bytes, 0, bytes.Length);
+        }
+
+        [ConditionalTheory]
+        [MemberData(nameof(ConnectionFilterData))]
+        [FrameworkSkipCondition(RuntimeFrameworks.Mono, SkipReason = "Test hangs after execution on Mono.")]
```
halter73: You can remove this :smile:

benaadams:
`TakeChunkedLine` is correct, renaming stuff to make it more clear though

benaadams:
Tests added

test/Microsoft.AspNetCore.Server.KestrelTests/ChunkedRequestTests.cs
benaadams: Yep

benaadams:
Not quite sure how to deal with the connection being shutdown by invalid data; sometimes it throws, sometimes it doesn't

Have added a `catch (IOException)` but there is probably a better way? Can't see any invalid request tests to mimic

test/Microsoft.AspNetCore.Server.KestrelTests/ChunkedRequestTests.cs
cesarbs: Sweet, I like them :+1:

cesarbs:
I say this is ready to :ship:. @halter73 any concerns?

test/Microsoft.AspNetCore.Server.KestrelTests/ChunkedRequestTests.cs
```diff
@@ -0,0 +1,477 @@
...
+                    await connection.ReceiveEnd(
+                        "Content-Length: 0",
+                        "Server: Kestrel",
+                        "",
+                        "");
+                }
+            }
+        }
+
+        private class TestApplicationErrorLogger : ILogger
```
halter73: Can we move these private classes to TestHelpers? As simple as these are, I don't like having exact copies that could diverge.

benaadams: Done

halter73:
Do you see these errors? If not, I can investigate tomorrow.

    Microsoft.AspNetCore.Server.KestrelTests.ChunkedRequestTests.InvalidLengthResultsIn500(testContext: TestServiceContext { App = null, AppLifetime = LifetimeNotImplemented { ApplicationStarted = (throws NotImplementedException), ApplicationStopped = (throws NotImplementedException), ApplicationStopping = (throws NotImplementedException) }, DateHeaderValueManager = TestDateHeaderValueManager { }, FrameFactory = null, Log = TestKestrelTrace { }, ... }) [FAIL]
      System.IO.IOException : Unable to write data to the transport connection: An established connection was aborted by the software in your host machine.
      ---- System.Net.Sockets.SocketException : An established connection was aborted by the software in your host machine
      Stack Trace:
           at System.Net.Sockets.NetworkStream.BeginWrite(Byte[] buffer, Int32 offset, Int32 size, AsyncCallback callback, Object state)
           at System.IO.Stream.<>c.<BeginEndWriteAsync>b__53_0(Stream stream, ReadWriteParameters args, AsyncCallback callback, Object state)
           at System.Threading.Tasks.TaskFactory`1.FromAsyncTrim[TInstance,TArgs](TInstance thisRef, TArgs args, Func`5 beginMethod, Func`3 endMethod)
           at System.IO.Stream.BeginEndWriteAsync(Byte[] buffer, Int32 offset, Int32 count)
           at System.IO.Stream.WriteAsync(Byte[] buffer, Int32 offset, Int32 count, CancellationToken cancellationToken)
           at System.IO.Stream.WriteAsync(Byte[] buffer, Int32 offset, Int32 count)
           at System.IO.StreamWriter.<FlushAsyncInternal>d__68.MoveNext()
        C:\Users\shalter\dev\Universe\KestrelHttpServer\test\Microsoft.AspNetCore.Server.KestrelTests\TestConnection.cs(51,0): at Microsoft.AspNetCore.Server.KestrelTests.TestConnection.<Send>d__6.MoveNext()
        C:\Users\shalter\dev\Universe\KestrelHttpServer\test\Microsoft.AspNetCore.Server.KestrelTests\ChunkedRequestTests.cs(366,0): at Microsoft.AspNetCore.Server.KestrelTests.ChunkedRequestTests.<InvalidLengthResultsIn500>d__9.MoveNext()
        ----- Inner Stack Trace -----
           at System.Net.Sockets.Socket.BeginSend(Byte[] buffer, Int32 offset, Int32 size, SocketFlags socketFlags, AsyncCallback callback, Object state)
           at System.Net.Sockets.NetworkStream.BeginWrite(Byte[] buffer, Int32 offset, Int32 size, AsyncCallback callback, Object state)
    Microsoft.AspNetCore.Server.KestrelTests.ChunkedRequestTests.InvalidSizedDataResultsIn500(testContext: TestServiceContext { App = null, AppLifetime = LifetimeNotImplemented { ApplicationStarted = (throws NotImplementedException), ApplicationStopped = (throws NotImplementedException), ApplicationStopping = (throws NotImplementedException) }, DateHeaderValueManager = TestDateHeaderValueManager { }, FrameFactory = null, Log = TestKestrelTrace { }, ... }) [FAIL]
      System.IO.IOException : Unable to write data to the transport connection: An established connection was aborted by the software in your host machine.
      ---- System.Net.Sockets.SocketException : An established connection was aborted by the software in your host machine
      Stack Trace:
           at System.Net.Sockets.NetworkStream.BeginWrite(Byte[] buffer, Int32 offset, Int32 size, AsyncCallback callback, Object state)
           at System.IO.Stream.<>c.<BeginEndWriteAsync>b__53_0(Stream stream, ReadWriteParameters args, AsyncCallback callback, Object state)
           at System.Threading.Tasks.TaskFactory`1.FromAsyncTrim[TInstance,TArgs](TInstance thisRef, TArgs args, Func`5 beginMethod, Func`3 endMethod)
           at System.IO.Stream.BeginEndWriteAsync(Byte[] buffer, Int32 offset, Int32 count)
           at System.IO.Stream.WriteAsync(Byte[] buffer, Int32 offset, Int32 count, CancellationToken cancellationToken)
           at System.IO.Stream.WriteAsync(Byte[] buffer, Int32 offset, Int32 count)
           at System.IO.StreamWriter.<FlushAsyncInternal>d__68.MoveNext()
        C:\Users\shalter\dev\Universe\KestrelHttpServer\test\Microsoft.AspNetCore.Server.KestrelTests\TestConnection.cs(51,0): at Microsoft.AspNetCore.Server.KestrelTests.TestConnection.<Send>d__6.MoveNext()
        C:\Users\shalter\dev\Universe\KestrelHttpServer\test\Microsoft.AspNetCore.Server.KestrelTests\ChunkedRequestTests.cs(413,0): at Microsoft.AspNetCore.Server.KestrelTests.ChunkedRequestTests.<InvalidSizedDataResultsIn500>d__10.MoveNext()
        ----- Inner Stack Trace -----
           at System.Net.Sockets.Socket.BeginSend(Byte[] buffer, Int32 offset, Int32 size, SocketFlags socketFlags, AsyncCallback callback, Object state)
           at System.Net.Sockets.NetworkStream.BeginWrite(Byte[] buffer, Int32 offset, Int32 size, AsyncCallback callback, Object state)
    Finished:    Microsoft.AspNetCore.Server.KestrelTests
    === TEST EXECUTION SUMMARY ===
    Microsoft.AspNetCore.Server.KestrelTests  Total: 359, Errors: 0, Failed: 2, Skipped: 2, Time: 10.361s

benaadams:
On and off - get it with the other 500 throwing test also. Normally seems to happen if I run the tests repeatedly rather than in clean `cmd`

Can wrap in a `IOException` catch; but that obv misses the point a little. Think it must be either sending the error post start; or the `SendEnd` chops the connection prior to flushing sometimes?

benaadams:
However... Kestrel should probably also abort the connection for those tests; which cause the issue also.

benaadams:
K should now be returning 400 and chopping the connection; not getting the earlier exception either - might have been overflow junk into the next request causing the connection to chop?

benaadams:
Hmm, not perfect; if user code caught the exception and threw it away it would still go wrong...

benaadams:
Should still close connection even if user code handles it now.

benaadams:
@halter73 that error is caused by `Consume` throwing the exception on read; then chopping the connection.

benaadams:
@halter73 last commit resolved the issues?

benaadams:
Rebased, better bad request handling for headers and invalid content lengths

benaadams:
Re-rebased for newest; resolving errors for newer tests.

benaadams:
I'm getting an failure in the cctor for MessageBody.ForChunkedEncoding of Invalid Program exception on full framework.

If I move `private static Vector<byte> _vectorCRs = new Vector<byte>((byte)'\r');` into `MessageBody` then every frame test fails with same error. Bit odd...

benaadams:
AppVeyor: 
```
System.Diagnostics.Debug+DebugAssertException: Block being garbage collected instead of returned to pool 
   at System.Environment.GetStackTrace(Exception e, Boolean needFileInfo)
```

Issue #580, opened 2016-01-14
Properly return pooled blocks in SocketInput


halter73:
When I was investigating #579, I discovered that all of the blocks getting finalized in [the app](https://github.com/rynowak/mvc-sandbox/tree/f51625029ac437a7315961e5dc01e1b8b68a252a/workloads/BasicViews) I was using repro the issue were last leased by `SocketInput.IncomingStart`.

It appears that the last incoming block will never be returned even if it is fully consumed.

DamianEdwards:
This seems like a really simple fix, yes? If so, I'd like to see this done in RC2.

muratg:
If it's not a straightforward quick fix, let's move it to backlog

benaadams:
Measurement across 1M non keep-alive requests; majority of the highlighted `byte[]` is the blocks not being returned early enough from the `SocketInput`

![non-keepalive-connections](http://aoa.blob.core.windows.net/aspnet/non-keepalive-connections.png)

SajayAntony:
@sivagms - This might be the issue that Martin was trying to fix today due to the incorrect finalization. Do you want to take a look since you know the fix he was talking about?

benaadams:
@SajayAntony I assumed it was in this fix? (though haven't measured) https://github.com/aspnet/KestrelHttpServer/pull/623

SajayAntony:
btw.. @halter73 ryan has moved these benchmark and loadtest apps [here](https://github.com/aspnet/Performance/tree/dev/testapp)

@benaadams :tired_face:  I hadn't seen that PR and I've just joined this party :)

benaadams:
@halter73 added PR for how I generated this https://github.com/aspnet/Performance/pull/48

martinpf:
I noticed this as well last Friday.  Under load, with a significant number of new connections being made, the finalizer  queue falls behind and the memory blocks waiting in the queue  can quickly exceed 1 GB.   A simple fix for this would be to return the block held by SocketInput on connection close.

Issue #247, opened 2015-10-07
Kestrel takes 2-3 seconds to shutdown if any connections have been made


AspNetSmurfLab:
Kestrel is taking 2-3 seconds to shutdown (e.g. after Ctrl-C) if any connections have been made to it. This is causing the development experience in Visual Studio to be poor due to the long time it takes to get the server going again after making a code change.

## Repro steps
1. Start the server
1. Immediately Ctrl-C to end it without making a request
1. It exits immediately
1. Start the server again
1. Make a request from a browser then close the browser
1. Ctrl-C the kestrel server
1. It takes 2-3 seconds for the process to end

by @DamianEdwards

muratg:
OK, let's investigate what we can do for Beta 8.

halter73:
@muratg I tested the shalter/247 branch referenced above with the new WTE build. This seemed to reduce the time window where you could get a 502.3 response, but didn't eliminate it completely.

Bill thinks that the change to kill iisexpress on ctrl+f5 could fix the 502.3 issue as well, so I'm waiting on that change before doing any further work on this.

muratg:
@halter73 Sounds good.  If WTE change can eliminate the need to do this in Beta 8, please move to RC1.

muratg:
Assigning back to @halter73 as @cesarbs doesn't have cycles currently to work on this.   

@halter73 Please look into this when you're in-progress work is done as it's a bad developer experience problem.

salerth:
Not sure if this is the same bug/related, but kestrel wont respond (at all) to CTRL+C after it has served at least 1 page on OSX + CoreCLR.  If I CTRL+C before serving a page it exits properly.

Mono works fine and exits at any time.

`dnx-coreclr-darwin-x64.1.0.0-rc2-16249`

DamianEdwards:
Need for RC2

PR #623, opened 2016-02-10
Close connections as quickly and gracefully as possible on shutdown

halter73:
- Don't rely on the finalizer to return blocks referenced by SocketInput
- Make the time given for requests to complete gracefully configurable
- Close UvAsyncHandle properly to avoid zombie threads
- Verify that memory pool blocks aren't leaked in tests

#580 #247


src/Microsoft.AspNetCore.Server.Kestrel/Http/SocketInput.cs
```diff
@@ -246,5 +251,23 @@ public void GetResult()
                 throw new IOException(error.Message, error);
             }
         }
+
+        public void Dispose()
+        {
+            AbortAwaiting();
+
+            // Return all blocks
+            var block = _head;
```
davidfowl: Is this safe without locks?

halter73: This is only called after the socket is closed so there should not be any more incoming data. The AbortAwaiting call kills any consumers. (If all else fails, the finalizer still returns the block in production, but it would crash the process in debug builds)

src/Microsoft.AspNetCore.Server.Kestrel/Http/Connection.cs
```diff
@@ -174,8 +260,8 @@ private void OnRead(UvStreamHandle handle, int status)
             }
             else
             {
-                _socket.ReadStop();
                 Log.ConnectionReadFin(_connectionId);
+                _socket.ReadStop();
```
cesarbs: Why did you switch the order here?

src/Microsoft.AspNetCore.Server.Kestrel/Http/ConnectionManager.cs
```diff
@@ -0,0 +1,53 @@
...
+        {
+            _thread = thread;
+        }
+
+        // This must be called on the libuv event loop
+        public void WalkConnectionsAndClose()
+        {
+            if (_connectionStopTasks != null)
+            {
+                throw new InvalidOperationException(nameof(WalkConnectionsAndClose) + " cannot be called twice.");
```
cesarbs: Use $"<string>".

src/Microsoft.AspNetCore.Server.Kestrel/Http/ConnectionManager.cs
```diff
@@ -0,0 +1,53 @@
...
+                    _connectionStopTasks.Add(connection.StopAsync());
+                }
+            });
+        }
+
+        public Task WaitForConnectionCloseAsync()
+        {
+            if (_connectionStopTasks == null)
+            {
+                throw new InvalidOperationException(nameof(WalkConnectionsAndClose) + " must be called first.");
```
cesarbs: $"<string>" here too

src/Microsoft.AspNetCore.Server.Kestrel/IKestrelServerInformation.cs
```diff
@@ -9,6 +10,15 @@ public interface IKestrelServerInformation
     {
         int ThreadCount { get; set; }
 
+        /// <summary>
+        /// The amount of time after the server begins shutting down before connections will be forcefully closed.
+        /// By default, Kestrel will wait 5 seconds for any ongoing requests to complete before terminating
```
cesarbs: Why 5s?

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/PathBaseTests.cs
```diff
@@ -25,13 +24,12 @@ public class PathBaseTests
         [InlineData("/base/more", "/base/more", "/base/more", "")]
         [InlineData("/base/more", "/base/more/something", "/base/more", "/something")]
         [InlineData("/base/more", "/base/more/something/", "/base/more", "/something/")]
-        [FrameworkSkipCondition(RuntimeFrameworks.Mono, SkipReason = "Test hangs after execution on Mono.")]
```
cesarbs: :+1:

test/Microsoft.AspNetCore.Server.KestrelTests/EngineTests.cs
```diff
@@ -1125,6 +1096,7 @@ public void ConnectionCanReadAndWrite(TestServiceContext testContext)
                     for (int i = 0; i < 100; i++)
                     {
                         await response.WriteAsync(largeString, lifetime.RequestAborted);
+                        registrationWh.Wait(1000);
```
cesarbs: Should the loop count still be 100?

src/Microsoft.AspNetCore.Server.Kestrel/Http/Connection.cs
halter73: No reason. Will revert.

src/Microsoft.AspNetCore.Server.Kestrel/IKestrelServerInformation.cs
halter73: I picked a number that sounded good to me. I'm open to suggestions.

For most web applications that typically don't take 5 seconds or more to handle a single request, the timeout shouldn't be hit.

test/Microsoft.AspNetCore.Server.KestrelTests/EngineTests.cs
halter73: Tough one. It means that it would take at least 100 seconds to fail if we broke it, but as flaky as this test is particularly on the CI server, I don't particularly mind.

In practice, this test passes in less than a second on my machine.

benaadams:
Does this provide a deterministic disposal point for the connection? (rather than any finalization)

benaadams:
Wanted to do a pooled SocketOutput - but bit complicated atm

src/Microsoft.AspNetCore.Server.Kestrel/Infrastructure/KestrelThread.cs
```diff
@@ -71,6 +71,12 @@ public Task StartAsync()
...
         }
 
+        // This must be called from the libuv event loop.
+        public void AllowStop()
+        {
+            _post.Unreference();
+        }
+
         public void Stop(TimeSpan timeout)
         {
```
benaadams: Add
```csharp
#if !DEBUG
    if (_thread != null)
    {
        _thread.IsBackground = true;
    }
#end
```

halter73:
@benaadams the "deterministic disposal point for the connection" would be `Connection.OnSocketClosed`. Right now, the only time I'm aware of that it might not be called is during ungraceful shutdown if `KestrelThread.OnStopRude` disposes the socket.

I don't think there is a need for it currently, but if you want `OnSocketClosed` to *always* be called, you could simply cast the `UvHandle` like `ConnectionManager` does in `OnStopRude` and call `OnSocketClosed` there.

benaadams:
> the only time I'm aware of that it might not be called is during ungraceful shutdown if KestrelThread.OnStopRude disposes the socket.

Good enough for me :)

PR #624, opened 2016-02-11
Removing Mono specific DllImports

moozzyk:
With the conversion to dotnet nothing will preload the libuv.dylib into the process so the __Internal in DllImport won't work. Users will have to put libuv.lib in the default search location or set the DYLD_LIBRARY_PATH to the folder where libuv.dylib was restored or installed.

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Kestrel/Networking/Libuv.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/Networking/Libuv.cs
@@ -12,90 +12,44 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Networking
         {
             IsWindows = PlatformApis.IsWindows;
 
-            var isDarwinMono = PlatformApis.IsDarwin && PlatformApis.IsMono;
-
-            if (isDarwinMono)
+            _uv_loop_init = NativeMethods.uv_loop_init;
+            _uv_loop_close = NativeMethods.uv_loop_close;
+            _uv_run = NativeMethods.uv_run;
+            _uv_stop = NativeMethods.uv_stop;
+            _uv_ref = NativeMethods.uv_ref;
+            _uv_unref = NativeMethods.uv_unref;
+            _uv_close = NativeMethods.uv_close;
+            _uv_async_init = NativeMethods.uv_async_init;
+            _uv_async_send = NativeMethods.uv_async_send;
+            _uv_tcp_init = NativeMethods.uv_tcp_init;
+            _uv_tcp_bind = NativeMethods.uv_tcp_bind;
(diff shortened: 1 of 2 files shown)


davidfowl:
Won't this break DNX?

moozzyk:
do we care?

davidfowl:
@moozzyk Until we have working CLI tooling yes. As long as we have a good workaround for OSX that's fine (is OSX the only one?)

moozzyk:
@davidfowl - The workaround for dnx on Mono will be the same as for CLI on Mono - install libuv using homebrew or, possibly, set the DYLD_LIBRARY_PATH to point to the libuv.dylib. 
I don't think this change is very urgent but currently it is impossible to run Kestrel on Darwin Mono when using CLI

davidfowl:
> The workaround for dnx on Mono will be the same as for CLI on Mono - install libuv using homebrew or, possibly, set the DYLD_LIBRARY_PATH to point to the libuv.dylib. 
I don't think this change is very urgent but currently it is impossible to run Kestrel on Darwin Mono when using CLI

It'll work on mono because we copy it to the output folder with the CLI (even on build for net45x). I wonder if we should runtime check if we're on DNX and use it if we're there...

moozzyk:
It won't work on Mono and CLI because DllImport is trying to load `__Internal` and nothing loaded the dylib into the process (this was dnx's job before) so the exported functions couldn't be found. I built an app today to confirm this and, as expected, it failed due to not being able to invoke `uv.loop_size()` which is the first call to libuv Kestrel makes.

davidfowl:
@moozzyk best course of action is to check for DNX specifically at runtime. Do we have an API for that? Hmmm.

davidfowl:
Super hack:

https://github.com/aspnet/dnx/blob/dev/src/Microsoft.Dnx.Host/Bootstrapper.cs#L77

Check to see if DNX_SERVICEPROVIDER is set in the app domain data :smile:

moozzyk:
Can you check for a `Microsoft.Dnx*` assembly/type?

moozzyk:
DNX_SERVICEPROVIDER hack should work for now but I would really like to remove `__Internal` the code longer term.

davidfowl:
@moozzyk we should I just don't want to break anyone using things * in VS today.

moozzyk:
We haven't had coherence-signed in a long time and not sure when the lucky day will come... :smile:

benaadams:
Is why we all went aspnetvolatiledev :stuck_out_tongue_winking_eye:

moozzyk:
@benaadams - unfortunately yes.

moozzyk:
Closing this for now.

PR #633, opened 2016-02-17
Make the CI build work

benaadams:
Resolve compile warnings
Always shutdown

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -5,5 +5,3 @@ build_script:
 clone_depth: 1
 test: off
 deploy: off
-environment:
-  NO_PARALLEL_TEST_PROJECTS: Microsoft.AspNetCore.Server.Kestrel.FunctionalTests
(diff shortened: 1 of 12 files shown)


benaadams:
Server should always shut down after last commit in this 

/cc @cesarbs @halter73

benaadams:
Crazy town! Lights are green! --> https://github.com/aspnet/KestrelHttpServer/pull/635

cesarbs:
@halter73 has the proper fix here: https://github.com/aspnet/KestrelHttpServer/commit/7455e11b82e95175f157733219b82be84f34e0b8

halter73:
I'm fairly certain the CI hangs are fixed by [this commit](https://github.com/aspnet/KestrelHttpServer/commit/7455e11b82e95175f157733219b82be84f34e0b8). That's why I removed the same FrameworkSkipConditions in #623. The commit message should explain why the fix makes the second call to `uv_run` terminate reliably.

I think this is a more appropriate than setting the `IsBackground` property on the thread.

benaadams:
Is obviously better to have the cause dealt with rather than just the symptom :)

Might be worth still adding the the `IsBackground = false` for `#if !DEBUG`? So issues still get seen for testing but always stops in production (with the actual good clean up code in #623)

halter73:
I think the `#if !DEBUG` thing is a good idea.

benaadams:
Want to add it to #623? Will close this and the other one.

halter73:
Will do

PR #638, opened 2016-02-17
Add IHttpConnectionFeature.ConnectionId.

Tratcher:
Depends on https://github.com/aspnet/HttpAbstractions/pull/564
@halter73 

Most of the noise is that I changed the existing connectionId from long to string. I also cleaned up FrameFactory.

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Kestrel/Http/Connection.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/Http/Connection.cs
@@ -2,7 +2,6 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
-using System.Net;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Server.Kestrel.Filter;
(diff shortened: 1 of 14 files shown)


src/Microsoft.AspNetCore.Server.Kestrel/Http/Connection.cs
```diff
@@ -194,18 +192,18 @@ private void OnRead(UvStreamHandle handle, int status)
 
         private Frame CreateFrame()
         {
-            return FrameFactory(this, _remoteEndPoint, _localEndPoint, _filterContext?.PrepareRequest);
+            return FrameFactory(this);
```
halter73: :+1:

src/Microsoft.AspNetCore.Server.Kestrel/Http/Frame.cs
```diff
@@ -73,24 +73,18 @@ public abstract partial class Frame : FrameContext, IFrameControl
...
-                     Action<IFeatureCollection> prepareRequest)
             : base(context)
         {
-            _remoteEndPoint = remoteEndPoint;
-            _localEndPoint = localEndPoint;
-            _prepareRequest = prepareRequest;
+            _remoteEndPoint = context.RemoteEndPoint;
+            _localEndPoint = context.LocalEndPoint;
+            _prepareRequest = context.PrepareRequest;
+            _connectionId = context.ConnectionId;
```
halter73: This should happen in the ConnectionContext's copy ctor.

src/Microsoft.AspNetCore.Server.Kestrel/Http/ConnectionContext.cs
```diff
@@ -21,7 +25,17 @@ public ConnectionContext(ConnectionContext context) : base(context)
...
+
         public IConnectionControl ConnectionControl { get; set; }
+
+        public IPEndPoint RemoteEndPoint { get; set; }
+
+        public IPEndPoint LocalEndPoint { get; set; }
+
+        public string ConnectionId { get; set; }
+
+        public Action<IFeatureCollection> PrepareRequest { get; set; }
```
halter73: Update copy ctor.

Tratcher:
Updated

halter73:
:shipit:

src/Microsoft.AspNetCore.Server.Kestrel/Http/Connection.cs
```diff
@@ -32,23 +32,20 @@ public class Connection : ConnectionContext, IConnectionControl
...
-        private IPEndPoint _remoteEndPoint;
-        private IPEndPoint _localEndPoint;
-
         public Connection(ListenerContext context, UvStreamHandle socket) : base(context)
         {
             _socket = socket;
             ConnectionControl = this;
 
-            _connectionId = Interlocked.Increment(ref _lastConnectionId);
+            ConnectionId = Interlocked.Increment(ref _lastConnectionId).ToString(CultureInfo.InvariantCulture);
```
davidfowl: Not performant enough. Take a look at what we did with request id and copy it.

benaadams: As this is a property could it be lazy-init? Though might not work with the logging; as will currently always init.

Logging could be passed a `ConnectionContext` rather than `ConnectionId` and if enabled at that level then call `ConnectionId` which inits the value.

`Task.Id` in coreclr uses this approach https://github.com/dotnet/coreclr/pull/3139

Tratcher: Keep in mind that this is once per connection, not per request. I think Frame's copy constructor would always resolve it anyways.

Tratcher:
Updated

PR #642, opened 2016-02-18
Handle 0-byte reads correctly (fixes #520).

### LINKED ISSUE #520: System.IO.InvalidDataException in Kestrel.Http.MessageBody when posting large file

cesarbs:
We were incorrectly assuming a 0-byte read from libuv indicated a connection end. This is wrong, since according to the [docs](http://docs.libuv.org/en/v1.x/stream.html#c.uv_read_cb) when the read callback is called with a zero status that only means there is no data available to read at the moment. This can be safely ignored.

#520

cc @halter73

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Kestrel/Http/Connection.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/Http/Connection.cs
@@ -16,9 +16,9 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Http
         // Base32 encoding - in ascii sort order for easy text based sorting
         private static readonly string _encode32Chars = "0123456789ABCDEFGHIJKLMNOPQRSTUV";
 
-        private static readonly Action<UvStreamHandle, int, object> _readCallback = 
+        private static readonly Action<UvStreamHandle, int, object> _readCallback =
             (handle, status, state) => ReadCallback(handle, status, state);
-        private static readonly Func<UvStreamHandle, int, object, Libuv.uv_buf_t> _allocCallback = 
+        private static readonly Func<UvStreamHandle, int, object, Libuv.uv_buf_t> _allocCallback =
             (handle, suggestedsize, state) => AllocCallback(handle, suggestedsize, state);
 
         // Seed the _lastConnectionId for this application instance with
@@ -252,8 +252,18 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Http
 
         private void OnRead(UvStreamHandle handle, int status)
         {
+            if (status == 0)
+            {
+                // A zero status does not indicate an error or connection end. It indicates
+                // there is no data to be read right now.
+                // See the note at http://docs.libuv.org/en/v1.x/stream.html#c.uv_read_cb.
+                // We need to clean up whatever was allocated by OnAlloc.
+                _rawSocketInput.IncomingDeferred();
+                return;
+            }
+
             var normalRead = status > 0;
-            var normalDone = status == 0 || status == Constants.ECONNRESET || status == Constants.EOF;
+            var normalDone = status == Constants.ECONNRESET || status == Constants.EOF;
             var errorDone = !(normalDone || normalRead);
             var readCount = normalRead ? status : 0;
 
--- a/src/Microsoft.AspNetCore.Server.Kestrel/Http/SocketInput.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/Http/SocketInput.cs
@@ -2,6 +2,7 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.Diagnostics;
 using System.IO;
 using System.Runtime.CompilerServices;
 using System.Threading;
@@ -115,6 +116,21 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Http
             }
 
             Complete();
+        }
+
+        public void IncomingDeferred()
+        {
+            Debug.Assert(_pinned != null);
+
+            if (_pinned != null)
+            {
+                if (_pinned != _tail)
+                {
+                    _memory.Return(_pinned);
+                }
+
+                _pinned = null;
+            }
         }
 
         private void Complete()
--- a/test/Microsoft.AspNetCore.Server.KestrelTests/ConnectionTests.cs
+++ b/test/Microsoft.AspNetCore.Server.KestrelTests/ConnectionTests.cs
@@ -1,0 +1,46 @@
+﻿using System.Threading;
+using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Server.Kestrel;
+using Microsoft.AspNetCore.Server.Kestrel.Http;
+using Microsoft.AspNetCore.Server.Kestrel.Infrastructure;
+using Microsoft.AspNetCore.Server.Kestrel.Networking;
+using Microsoft.AspNetCore.Server.KestrelTests.TestHelpers;
+using Xunit;
+
+namespace Microsoft.AspNetCore.Server.KestrelTests
+{
+    public class ConnectionTests
+    {
+        [Fact]
+        public void DoesNotEndConnectionOnZeroRead()
+        {
+            var mockLibuv = new MockLibuv();
+            
+            using (var memory = new MemoryPool2())
+            using (var engine = new KestrelEngine(mockLibuv, new TestServiceContext()))
+            {
+                engine.Start(count: 1);
+
+                var trace = new TestKestrelTrace();
+                var context = new ListenerContext(new TestServiceContext())
+                {
+                    FrameFactory = connectionContext => new Frame<HttpContext>(
+                        new DummyApplication(httpContext => TaskUtilities.CompletedTask), connectionContext),
+                    Memory2 = memory,
+                    ServerAddress = ServerAddress.FromUrl($"http://localhost:{TestServer.GetNextPort()}"),
+                    Thread = engine.Threads[0]
+                };
+                var socket = new MockSocket(mockLibuv, Thread.CurrentThread.ManagedThreadId, trace);
+                var connection = new Connection(context, socket);
+                connection.Start();
+
+                Libuv.uv_buf_t ignored;
+                mockLibuv.AllocCallback(socket.InternalGetHandle(), 2048, out ignored);
+                mockLibuv.ReadCallback(socket.InternalGetHandle(), 0, ref ignored);
+                Assert.False(connection.SocketInput.RemoteIntakeFin);
+
+                connection.ConnectionControl.End(ProduceEndType.SocketDisconnect);
+            }
+        }
+    }
+}
--- a/test/Microsoft.AspNetCore.Server.KestrelTests/TestHelpers/MockLibuv.cs
+++ b/test/Microsoft.AspNetCore.Server.KestrelTests/TestHelpers/MockLibuv.cs
@@ -17,6 +17,12 @@ namespace Microsoft.AspNetCore.Server.KestrelTests.TestHelpers
         unsafe public MockLibuv()
             : base(onlyForTesting: true)
         {
+            _onWrite = (socket, buffers, triggerCompleted) =>
+            {
+                triggerCompleted(0);
+                return 0;
+            };
+
             _uv_write = UvWrite;
 
             _uv_async_send = postHandle =>
@@ -69,6 +75,8 @@ namespace Microsoft.AspNetCore.Server.KestrelTests.TestHelpers
             _uv_walk = (loop, callback, ignore) => 0;
             _uv_err_name = errno => IntPtr.Zero;
             _uv_strerror = errno => IntPtr.Zero;
+            _uv_read_start = UvReadStart;
+            _uv_read_stop = handle => 0;
         }
 
         public Func<UvStreamHandle, int, Action<int>, int> OnWrite
@@ -81,6 +89,17 @@ namespace Microsoft.AspNetCore.Server.KestrelTests.TestHelpers
             {
                 _onWrite = value;
             }
+        }
+
+        public uv_alloc_cb AllocCallback { get; set; }
+
+        public uv_read_cb ReadCallback { get; set; }
+
+        private int UvReadStart(UvStreamHandle handle, uv_alloc_cb allocCallback, uv_read_cb readCallback)
+        {
+            AllocCallback = allocCallback;
+            ReadCallback = readCallback;
+            return 0;
         }
 
         unsafe private int UvWrite(UvRequest req, UvStreamHandle handle, uv_buf_t* bufs, int nbufs, uv_write_cb cb)
--- a/test/Microsoft.AspNetCore.Server.KestrelTests/TestHelpers/MockSocket.cs
+++ b/test/Microsoft.AspNetCore.Server.KestrelTests/TestHelpers/MockSocket.cs
@@ -1,0 +1,21 @@
+﻿using System;
+using Microsoft.AspNetCore.Server.Kestrel.Infrastructure;
+using Microsoft.AspNetCore.Server.Kestrel.Networking;
+
+namespace Microsoft.AspNetCore.Server.KestrelTests.TestHelpers
+{
+    class MockSocket : UvStreamHandle
+    {
+        public MockSocket(Libuv uv, int threadId, IKestrelTrace logger) : base(logger)
+        {
+            CreateMemory(uv, threadId, IntPtr.Size);
+        }
+
+        protected override bool ReleaseHandle()
+        {
+            DestroyMemory(handle);
+            handle = IntPtr.Zero;
+            return true;
+        }
+    }
+}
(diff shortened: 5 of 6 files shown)


davidfowl:
:shipit:

halter73:
I still think we should find a way to turn the repro app into a test case somehow.

I know it the repro app depends on a lot dependencies on top of Kestrel, but maybe that would be OK if the test lived in the ServerTests repo. Were you only able to repro using the browser on the client?

The fix looks good though.

davidfowl:
@halter73 where is the repro app and what was failing before?

halter73:
You'd have to ask @cesarbs were the app is. The part I saw wasa form-based file upload that saved to disk. I know it took a while to get a repro initially, so there might have been done trick I'm leaving out. We already have large request tests that have never failed in this way, so it isn't trivial to test.

> what was failing before?

I don't know what you mean. #520 covers it pretty well. Kestrel prematurely closed the connection in the middle of the upload because it misinterpreted the 0 status in the read callback.

davidfowl:
Might be a PITA to repro. At least the docs are pretty clear about it.

halter73:
It's definitely a PITA to repro. I almost gave it a :shipit:, but zero regression tests for this kind of fix feels bad. If we can't do a functional test, we should do a unit test. We already have a MockLibuv class we use to test SocketOutput.

src/Microsoft.AspNetCore.Server.Kestrel/Http/Connection.cs
```diff
@@ -252,8 +252,16 @@ private static void ReadCallback(UvStreamHandle handle, int status, object state
 
         private void OnRead(UvStreamHandle handle, int status)
         {
+            if (status == 0)
+            {
+                // A zero status does not indicate an error or connection end. It indicates
+                // there is no data to read right now and can be ignored.
+                // See the note at http://docs.libuv.org/en/v1.x/stream.html#c.uv_read_cb.
+                return;
```
halter73: So I was wrong about being able to simply return here.

The reason libuv is calling our read callback without any data is to allow us to cleanup after our alloc callback. This explains the Debug.Assert failure we saw yesterday.

Since our alloc callback calls `SocketInput.IncomingStart`, we need to null out `SocketInput._pinned`. Before that, if `_pinned != _tail`, we should return `_pinned` to the pool.

I would just add another method to `SocketInput` that does this. Maybe call it `IncomingDeferred`.

cesarbs:
@davidfowl The sample app is any app that can accept large file uploads. The issue likely also repros by just sending a large request body to Kestrel, but I've had more consistency/success when using an Mvc app and uploading a large (>1GB) file.

I used this to get a repro:

https://github.com/JosephGarrone/ASPNetUploadError

It's an RC1 app, but I moved it to the latest stuff.

cesarbs:
Fixed the leak, now working on a test.

test/Microsoft.AspNetCore.Server.KestrelTests/TestHelpers/MockSocket.cs
```diff
@@ -0,0 +1,29 @@
...
+
+        public MockSocket(Libuv uv, IKestrelTrace logger) : base(logger)
+        {
+            CreateMemory(uv, Thread.CurrentThread.ManagedThreadId, uv.handle_size(Libuv.HandleType.STREAM));
+        }
+
+        protected override bool ReleaseHandle()
+        {
+            // No-op
+            return true;
```
halter73: Even if these are just tests, I don' like leaking memory. Copy UvReqeust's ReleaseHandle method:

        protected override bool ReleaseHandle()
        {
            DestroyMemory(handle);
            handle = IntPtr.Zero;
            return true;
        }

src/Microsoft.AspNetCore.Server.Kestrel/Http/SocketInput.cs
```diff
@@ -117,6 +117,19 @@ public void IncomingComplete(int count, Exception error)
             Complete();
         }
 
+        public void IncomingDeferred()
+        {
+            if (_pinned != null)
```
halter73: Add a `Debug.Assert(_pinned != null)`. I'd really want to know if that wasn't the case.

You can keep the if check to be extra safe in prod (though there's an argument for just crashing).

test/Microsoft.AspNetCore.Server.KestrelTests/TestHelpers/MockLibuv.cs
```diff
@@ -83,6 +86,14 @@ unsafe public MockLibuv()
             }
         }
 
+        public uv_read_cb ReadCallback { get; set; }
+
+        private int UvReadStart(UvStreamHandle handle, uv_alloc_cb allocCallback, uv_read_cb readCallback)
+        {
+            ReadCallback = readCallback;
```
halter73: With the `Debug.Assert` you are going to need the store the allocCallback too.

test/Microsoft.AspNetCore.Server.KestrelTests/TestHelpers/MockLibuv.cs
```diff
@@ -1,4 +1,5 @@
 using System;
+using System.Runtime.InteropServices;
```
halter73: Why do we need this using?

test/Microsoft.AspNetCore.Server.KestrelTests/ConnectionTests.cs
```diff
@@ -0,0 +1,36 @@
...
+                    FrameFactory = connectionContext => new Frame<HttpContext>(
+                        new DummyApplication(httpContext => TaskUtilities.CompletedTask), connectionContext),
+                    ServerAddress = ServerAddress.FromUrl("http://localhost:5000")
+                };
+                var socket = new MockSocket(mockLibuv, trace);
+                var connection = new Connection(context, socket);
+                connection.Start();
+
+                Libuv.uv_buf_t ignored = new Libuv.uv_buf_t();
+                mockLibuv.ReadCallback(socket.InternalGetHandle(), 0, ref ignored);
```
halter73: Can we contain the `InternalGetHandle` ugliness to MockLibuv like we do with OnWrite?

test/Microsoft.AspNetCore.Server.KestrelTests/ConnectionTests.cs
```diff
@@ -0,0 +1,36 @@
...
+        {
+            using (var memory = new MemoryPool2())
+            {
+                var trace = new TestKestrelTrace();
+                var mockLibuv = new MockLibuv();
+                var context = new ListenerContext(new TestServiceContext())
+                {
+                    FrameFactory = connectionContext => new Frame<HttpContext>(
+                        new DummyApplication(httpContext => TaskUtilities.CompletedTask), connectionContext),
+                    ServerAddress = ServerAddress.FromUrl("http://localhost:5000")
```
halter73: Out of curiosity, where is `ServerAddress` being used?

cesarbs: `Frame` ctor.

test/Microsoft.AspNetCore.Server.KestrelTests/TestHelpers/MockSocket.cs
```diff
@@ -0,0 +1,28 @@
...
+using Microsoft.AspNetCore.Server.Kestrel.Networking;
+
+namespace Microsoft.AspNetCore.Server.KestrelTests.TestHelpers
+{
+    class MockSocket : UvStreamHandle
+    {
+        public MockSocket(int threadId, IKestrelTrace logger) : base(logger)
+        {
+            var uv = new MockLibuv();
+            CreateMemory(uv, threadId, uv.handle_size(Libuv.HandleType.STREAM));
```
halter73: Why did setting handle to 1 like before not work? `uv.handle_size` always returns `IntPtr.Size` BTW since we're using MockLibuv.

cesarbs: I'll need to check again - at some point I got an exception about it being invalid. But that was when I was playing around with the code, maybe the way it is now it doesn't matter anymore.

cesarbs: This happens:

    System.NullReferenceException: Object reference not set to an instance of an object.
       at Microsoft.AspNetCore.Server.Kestrel.Networking.Libuv.unsafe_async_send(IntPtr handle) in D:\src\aspnet\KestrelHttpServer\src\Microsoft.AspNetCore.Server.Kestrel\Networking\Libuv.cs:line 226
       at Microsoft.AspNetCore.Server.Kestrel.Networking.UvAsyncHandle.ReleaseHandle() in D:\src\aspnet\KestrelHttpServer\src\Microsoft.AspNetCore.Server.Kestrel\Networking\UvAsyncHandle.cs:line 62

cesarbs:
Rebased & squashed.

test/Microsoft.AspNetCore.Server.KestrelTests/TestHelpers/MockSocket.cs
halter73: The `UvAsyncHandle` should not be finalized. This shows that the engine is not properly shutting down and is causing `UvAsyncHandle.Dispose` not to be called.

test/Microsoft.AspNetCore.Server.KestrelTests/TestHelpers/MockSocket.cs
```diff
@@ -0,0 +1,21 @@
...
+using Microsoft.AspNetCore.Server.Kestrel.Infrastructure;
+using Microsoft.AspNetCore.Server.Kestrel.Networking;
+
+namespace Microsoft.AspNetCore.Server.KestrelTests.TestHelpers
+{
+    class MockSocket : UvStreamHandle
+    {
+        public MockSocket(Libuv uv, int threadId, IKestrelTrace logger) : base(logger)
+        {
+            CreateMemory(uv, threadId, uv.handle_size(Libuv.HandleType.STREAM));
```
halter73: Would prefer that the last param to `CreateMemory` was `IntPtr.Size`.

halter73:
:shipit: Once you address my last comment.

test/Microsoft.AspNetCore.Server.KestrelTests/LifetimeNotImplemented.cs
```diff
@@ -35,7 +35,6 @@ public CancellationToken ApplicationStopping
 
         public void StopApplication()
         {
-            throw new NotImplementedException();
```
halter73: Whoa. Don't remove this. This should never be called, and this is used in all of our tests. If it is being called, I wan't to know about it.

cesarbs: I think that's another leftover from when I was playing around with the code. Will revert and see what happens.

PR #646, opened 2016-02-19
Put threads into background before starting

### LINKED ISSUE #644: System.Threading.ThreadStateException: Thread is dead; state cannot be accessed.
This only happens in non-dev builds since it is #ifdefed out.

    System.Threading.ThreadStateException: Thread is dead; state cannot be accessed.
       at System.Threading.Thread.SetBackgroundNative(Boolean isBackground)
       at Microsoft.AspNetCore.Server.Kestrel.KestrelThread.Stop(TimeSpan timeout)
       at Microsoft.AspNetCore.Server.Kestrel.KestrelEngine.Dispose()
       at Microsoft.AspNetCore.Server.Kestrel.KestrelServer.Dispose()
       at Microsoft.AspNetCore.Server.Kestrel.KestrelServer.Start[TContext](IHttpApplication`1 application)
       at Microsoft.AspNetCore.Hosting.Internal.WebHost.Start()
       at Microsoft.AspNetCore.Hosting.WebHostExtensions.Run(IWebHost host, CancellationToken token, String shutdownMessage)
       at Microsoft.AspNetCore.Hosting.WebHostExtensions.Run(IWebHost host)
       at HelloWorldMvc.Startup.Main(String[] args)
       at Microsoft.Dnx.Runtime.Common.EntryPointExecutor.Execute(Assembly assembly, String[] args, IServiceProvider serviceProvider)
       at Microsoft.Dnx.ApplicationHost.Program.<>c__DisplayClass3_0.<ExecuteMain>b__0()
       at System.Threading.Tasks.Task`1.InnerInvoke()
       at System.Threading.Tasks.Task.Execute()

benaadams:
Resolves #644

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Kestrel/Infrastructure/KestrelThread.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/Infrastructure/KestrelThread.cs
@@ -52,6 +52,11 @@ namespace Microsoft.AspNetCore.Server.Kestrel
             _post = new UvAsyncHandle(_log);
             _thread = new Thread(ThreadStart);
             _thread.Name = "KestrelThread - libuv";
+#if !DEBUG
+            // Mark the thread as being as unimportant to keeping the process alive.
+            // Don't do this for debug builds, so we know if the thread isn't terminating.
+            _thread.IsBackground = true;
+#endif
             QueueCloseHandle = PostCloseHandle;
             QueueCloseAsyncHandle = EnqueueCloseHandle;
         }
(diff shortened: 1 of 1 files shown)


src/Microsoft.AspNetCore.Server.Kestrel/Infrastructure/KestrelThread.cs
```diff
@@ -66,6 +66,11 @@ public KestrelThread(KestrelEngine engine)
 
         public Task StartAsync()
         {
+#if !DEBUG
+            // Mark the thread as being as unimportant to keeping the process alive.
+            // Don't do this for debug builds, so we know if the thread isn't terminating.
+            _thread.IsBackground = true;
+#endif
```
halter73: I verified this works by moving libuv.dll so it couldn't be loaded, but I think this makes more sense in the ctor after we set `_thread.Name`.

halter73:
I wonder if there is any way to test that a DllNotFoundException surfaces for `KestrelServer.Start` if libuv is missing. I'm not aware of any easy method of testing this, but if you have any ideas...

src/Microsoft.AspNetCore.Server.Kestrel/Infrastructure/KestrelThread.cs
benaadams: Makes sense

benaadams:
Moved in to .ctor

PR #648, opened 2016-02-21
Fix build issues

benaadams:
Fixes some loose ends that prevent the CI build

/cc @halter73 @cesarbs

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -5,5 +5,3 @@ build_script:
 clone_depth: 1
 test: off
 deploy: off
-environment:
-  NO_PARALLEL_TEST_PROJECTS: Microsoft.AspNetCore.Server.Kestrel.FunctionalTests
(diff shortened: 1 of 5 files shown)


benaadams:
Alas still has the `Unhandled Exception: System.Diagnostics.Debug+DebugAssertException: Block being garbage collected instead of returned to pool` issue on travis

Rest of the stuff works though

halter73:
That assertion was only recently enabled by https://github.com/aspnet/KestrelHttpServer/pull/623, and I haven't seen a Travis pass since that inclusion. I'll see if I can reproduce on my Ubuntu machine.

benaadams:
I think its on shutdown; seems to be after all the tests pass

cesarbs:
I'm seeing that assertion too. Trying to track down where it's coming from.

halter73:
All the components of the issues have been merged. Next time we see `System.Diagnostics.Debug+DebugAssertException: Block being garbage collected instead of returned to pool`, I'm filing an issue.

cesarbs:
@halter73 I've seen it and I'm investigating. I've finally managed to get a repro :grin:

PR #659, opened 2016-02-26
Handle uploads larger than 2GB.

cesarbs:
Fixes the problem described in https://github.com/aspnet/KestrelHttpServer/issues/520#issuecomment-188591242.

cc @halter73 @lodejard @DavidObando

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Kestrel/Http/MessageBody.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/Http/MessageBody.cs
@@ -122,8 +122,8 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Http
             var unparsedContentLength = headers.HeaderContentLength.ToString();
             if (unparsedContentLength.Length > 0)
             {
-                int contentLength;
-                if (!int.TryParse(unparsedContentLength, out contentLength) || contentLength < 0)
+                long contentLength;
+                if (!long.TryParse(unparsedContentLength, out contentLength) || contentLength < 0)
                 {
                     context.ReportCorruptedHttpRequest(new BadHttpRequestException("Invalid content length."));
                     return new ForContentLength(keepAlive, 0, context);
@@ -166,10 +166,10 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Http
 
         private class ForContentLength : MessageBody
         {
-            private readonly int _contentLength;
-            private int _inputLength;
+            private readonly long _contentLength;
+            private long _inputLength;
 
-            public ForContentLength(bool keepAlive, int contentLength, Frame context)
+            public ForContentLength(bool keepAlive, long contentLength, Frame context)
                 : base(context)
             {
                 RequestKeepAlive = keepAlive;
@@ -181,7 +181,8 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Http
             {
                 var input = _context.SocketInput;
 
-                var limit = buffer.Array == null ? _inputLength : Math.Min(buffer.Count, _inputLength);
+                var inputLengthLimit = (int)Math.Min(_inputLength, int.MaxValue);
+                var limit = buffer.Array == null ? inputLengthLimit : Math.Min(buffer.Count, inputLengthLimit);
                 if (limit == 0)
                 {
                     return 0;
--- a/test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/RequestTests.cs
+++ b/test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/RequestTests.cs
@@ -1,7 +1,7 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
-using System.Collections.Generic;
+using System;
 using System.Globalization;
 using System.Net.Http;
 using System.Net.Sockets;
@@ -11,7 +11,6 @@ using Microsoft.AspNetCore.Builder;
 using Microsoft.AspNetCore.Hosting;
 using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.Testing.xunit;
-using Microsoft.Extensions.Configuration;
 using Newtonsoft.Json;
 using Newtonsoft.Json.Linq;
 using Xunit;
@@ -24,16 +23,9 @@ namespace Microsoft.AspNetCore.Server.Kestrel.FunctionalTests
         public async Task LargeUpload()
         {
             var port = PortManager.GetPort();
-            var config = new ConfigurationBuilder()
-                .AddInMemoryCollection(new Dictionary<string, string>
-                {
-                    { "server.urls", $"http://localhost:{port}/" }
-                })
-                .Build();
-
             var builder = new WebHostBuilder()
-                .UseConfiguration(config)
                 .UseServer("Microsoft.AspNetCore.Server.Kestrel")
+                .UseUrls($"http://localhost:{port}/")
                 .Configure(app =>
                 {
                     app.Run(async context =>
@@ -76,6 +68,57 @@ namespace Microsoft.AspNetCore.Server.Kestrel.FunctionalTests
             }
         }
 
+        [ConditionalFact]
+        [FrameworkSkipCondition(RuntimeFrameworks.Mono, SkipReason = "Fails on Mono on Mac because it is not 64-bit.")]
+        public async Task LargeMultipartUpload()
+        {
+            var port = PortManager.GetPort();
+            var builder = new WebHostBuilder()
+                .UseServer("Microsoft.AspNetCore.Server.Kestrel")
+                .UseUrls($"http://localhost:{port}/")
+                .Configure(app =>
+                {
+                    app.Run(async context =>
+                    {
+                        long total = 0;
+                        var bytes = new byte[1024];
+                        var count = await context.Request.Body.ReadAsync(bytes, 0, bytes.Length);
+                        while (count > 0)
+                        {
+                            total += count;
+                            count = await context.Request.Body.ReadAsync(bytes, 0, bytes.Length);
+                        }
+                        await context.Response.WriteAsync(total.ToString(CultureInfo.InvariantCulture));
+                    });
+                });
+
+            using (var host = builder.Build())
+            {
+                host.Start();
+
+                using (var client = new HttpClient())
+                {
+                    using (var form = new MultipartFormDataContent())
+                    {
+                        const int oneMegabyte = 1024 * 1024;
+                        const int files = 2048;
+                        var bytes = new byte[oneMegabyte];
+
+                        for (int i = 0; i < files; i++)
+                        {
+                            var fileName = Guid.NewGuid().ToString();
+                            form.Add(new ByteArrayContent(bytes), "file", fileName);
+                        }
+
+                        var length = form.Headers.ContentLength.Value;
+                        var response = await client.PostAsync($"http://localhost:{port}/", form);
+                        response.EnsureSuccessStatusCode();
+                        Assert.Equal(length.ToString(CultureInfo.InvariantCulture), await response.Content.ReadAsStringAsync());
+                    }
+                }
+            }
+        }
+
         [Theory]
         [InlineData("127.0.0.1", "127.0.0.1")]
         [InlineData("localhost", "127.0.0.1")]
@@ -95,14 +138,9 @@ namespace Microsoft.AspNetCore.Server.Kestrel.FunctionalTests
         public async Task DoesNotHangOnConnectionCloseRequest()
         {
             var port = PortManager.GetPort();
-            var config = new ConfigurationBuilder().AddInMemoryCollection(
-                new Dictionary<string, string> {
-                    { "server.urls", $"http://localhost:{port}" }
-                }).Build();
-
             var builder = new WebHostBuilder()
-                .UseConfiguration(config)
                 .UseServer("Microsoft.AspNetCore.Server.Kestrel")
+                .UseUrls($"http://localhost:{port}")
                 .Configure(app =>
                 {
                     app.Run(async context =>
@@ -129,14 +167,9 @@ namespace Microsoft.AspNetCore.Server.Kestrel.FunctionalTests
         public void RequestPathIsNormalized()
         {
             var port = PortManager.GetPort();
-            var config = new ConfigurationBuilder().AddInMemoryCollection(
-                new Dictionary<string, string> {
-                    { "server.urls", $"http://localhost:{port}/\u0041\u030A" }
-                }).Build();
-
             var builder = new WebHostBuilder()
-                .UseConfiguration(config)
                 .UseServer("Microsoft.AspNetCore.Server.Kestrel")
+                .UseUrls($"http://localhost:{port}/\u0041\u030A")
                 .Configure(app =>
                 {
                     app.Run(async context =>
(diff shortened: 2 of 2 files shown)


test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/RequestTests.cs
```diff
@@ -76,6 +80,72 @@ public class RequestTests
             }
         }
 
+        [Fact]
+        public async Task LargeMultipartUpload()
```
cesarbs: Won't pass until #642 is merged.

DavidObando:
Looks good to me.

src/Microsoft.AspNetCore.Server.Kestrel/Http/MessageBody.cs
```diff
@@ -120,7 +120,7 @@ public Task Consume(CancellationToken cancellationToken = default(CancellationTo
             var contentLength = headers.HeaderContentLength.ToString();
             if (contentLength.Length > 0)
             {
-                return new ForContentLength(keepAlive, int.Parse(contentLength), context);
+                return new ForContentLength(keepAlive, long.Parse(contentLength), context);
```
benaadams: `>= 0` check?

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/RequestTests.cs
```diff
@@ -76,6 +80,72 @@ public class RequestTests
...
+        {
+            var port = PortManager.GetPort();
+            var config = new ConfigurationBuilder()
```
davidfowl: You don't need to do this anymore. There's a UseUrl extension method hanging off `WebHostBuilder`

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/RequestTests.cs
```diff
@@ -76,6 +80,72 @@ public class RequestTests
...
+                            var fileName = Guid.NewGuid().ToString();
+                            var bytes = new byte[oneGigabyte];
+                            new Random().NextBytes(bytes);
+                            var content = new ByteArrayContent(bytes);
+                            content.Headers.ContentDisposition = new System.Net.Http.Headers.ContentDispositionHeaderValue("form-data")
+                            {
+                                Name = "\"files\"",
+                                FileName = "\"" + fileName + "\""
+                            };
+                            content.Headers.ContentType = new System.Net.Http.Headers.MediaTypeHeaderValue("application/octet-stream");
```
halter73: Nit: Don't need fully qualified name?

src/Microsoft.AspNetCore.Server.Kestrel/Http/MessageBody.cs
halter73: @cesarbs is now going to have to rebase on top of your chunked requests changes, so it should be there as long as he doesn't remove it :wink:

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/RequestTests.cs
```diff
@@ -1,8 +1,11 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System;
```
halter73: Nit: I think this is the only new using you need. I tried seeing how this test would fail in dev, and this was the only using I had to add.

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/RequestTests.cs
```diff
@@ -76,6 +80,72 @@ public class RequestTests
...
+                    using (var form = new MultipartFormDataContent())
+                    {
+                        const int oneGigabyte = 1024 * 1024 * 1024;
+                        const int files = 2;
+
+                        for (int i = 0; i < files; i++)
+                        {
+                            var fileName = Guid.NewGuid().ToString();
+                            var bytes = new byte[oneGigabyte];
+                            new Random().NextBytes(bytes);
```
halter73: The addition of this test pushed the functional test project completion time from 23.4s to 38.9s on my machine. Just removing `new Random().NextBytes(bytes);` took the total time down to 26.1s.

If there is no reason we need random bytes, I suggest re remove this.

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/RequestTests.cs
cesarbs: Conflicts with `Microsoft.Net.Http.Headers`.

cesarbs: Which is an unused `using` :grin: Will fix.

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/RequestTests.cs
cesarbs: Not needed, removing.

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/RequestTests.cs
cesarbs: Nice, I'll update the other tests too.

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/RequestTests.cs
cesarbs: I'll have to disable this on Mono because on Mac it is 32-bit only and this test only really works in 64-bit.

DavidObando: @troydai @NTaylorMullen question: is there a way to conditionally disable a test based on 32-bit run vs 64-bit run?

benaadams: `IntPtr.Size == 8` for 64bit

cesarbs: I could add a new `ITestCondition` to `Microsoft.AspNetCore.Testing`. Something like `BitnessSkipCondition`. But is that really necessary? Skipping on Mono shouldn't be so bad since we're still running the test on CoreCLR.

DavidObando: Yeah, I meant something along the lines of [`[OSSkipCondition]`](https://github.com/aspnet/Testing/blob/dev/src/Microsoft.AspNetCore.Testing/xunit/OSSkipConditionAttribute.cs)

DavidObando: @cesarbs you're right. Skipping mono is enough.

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/RequestTests.cs
```diff
@@ -76,6 +71,65 @@ public class RequestTests
             }
         }
 
+        [ConditionalFact]
+        [FrameworkSkipCondition(RuntimeFrameworks.Mono, SkipReason = "Needs to run on 64-bit Mono, which isn't available on OSX.")]
```
halter73: Sorry if I'm missing something obvious, but why does this need to run on 64-bit mono? `long` is the same size on 32-bit mono right? Do you run out of address space or something?

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/RequestTests.cs
```diff
@@ -76,6 +71,65 @@ public class RequestTests
...
+                            var bytes = new byte[oneGigabyte];
+                            var content = new ByteArrayContent(bytes);
+                            content.Headers.ContentDisposition = new ContentDispositionHeaderValue("form-data")
+                            {
+                                Name = "\"files\"",
+                                FileName = "\"" + fileName + "\""
+                            };
+                            content.Headers.ContentType = new MediaTypeHeaderValue("application/octet-stream");
+
+                            form.Add(new ByteArrayContent(bytes), "file", fileName);
```
DavidObando: You should be adding `content` to `form` instead of creating a `new ByteArrayContent(bytes)`.

halter73: Yeah. I noticed this too. `var content` isn't being used at all right now. As this test is written currently, the test process balloons to about 2GB. I think this can be avoided, and I'm looking into how to best fix this test to be more memory efficient while still testing the same issue.

halter73: BTW. I already chaged this test to use 2048 1MB files. This still had memory problems and this seemed mostly to be an issue with HttpClient. Most of the memory was still used up by tons of massive byte arrays rooted by `HttpRequestMessage`.

    0:000> !DumpObj /d 0000020f9b752a60
    Name:        System.Byte[]
    MethodTable: 00007ffbd1f27e00
    EEClass:     00007ffbd16d6c98
    Size:        1048600(0x100018) bytes
    Array:       Rank 1, Number of elements 1048576, Type Byte (Print Array)
    Content:     ................................................................................................................................
    Fields:
    None
    0:000> !gcroot 0000020f9b752a60
    Thread 41bc:
    *** ERROR: Module load completed but symbols could not be loaded for Microsoft.AspNetCore.Server.Kestrel.dll
        000000e68813e630 00007ffb790d8d0c Microsoft.AspNetCore.Server.Kestrel.KestrelThread.ThreadStart(System.Object)
            rbp-80: 000000e68813e680
                ->  0000020f00118fa8 System.Action`2[[System.Action`1[[System.IntPtr, mscorlib]], mscorlib],[System.IntPtr, mscorlib]]
                ->  0000020f000f41b8 Microsoft.AspNetCore.Server.Kestrel.KestrelThread
                ->  0000020f000f3680 Microsoft.AspNetCore.Server.Kestrel.KestrelEngine
                ->  0000020f000f3360 Microsoft.AspNetCore.Server.Kestrel.Http.DateHeaderValueManager
                ->  0000020f000f34b0 System.Threading.Timer
                ->  0000020f000f3520 System.Threading.TimerHolder
                ->  0000020f000f34c8 System.Threading.TimerQueueTimer
                ->  0000020f002eea18 System.Threading.TimerQueueTimer
                ->  0000020f002ee898 System.Threading.CancellationTokenSource+LinkedCancellationTokenSource
                ->  0000020f002ef8c8 System.Threading.SparselyPopulatedArray`1[[System.Threading.CancellationCallbackInfo, mscorlib]][]
                ->  0000020f002ef900 System.Threading.SparselyPopulatedArray`1[[System.Threading.CancellationCallbackInfo, mscorlib]]
                ->  0000020f002ef920 System.Threading.SparselyPopulatedArrayFragment`1[[System.Threading.CancellationCallbackInfo, mscorlib]]
                ->  0000020f002ef950 System.Threading.CancellationCallbackInfo[]
                ->  0000020f002ef898 System.Threading.CancellationCallbackInfo
                ->  0000020f002eeb98 System.Net.Http.WinHttpRequestState
                ->  0000020f002eeb30 System.Threading.Tasks.TaskCompletionSource`1[[System.Net.Http.HttpResponseMessage, System.Net.Http]]
                ->  0000020f002eeb48 System.Threading.Tasks.Task`1[[System.Net.Http.HttpResponseMessage, System.Net.Http]]
                ->  0000020f002eece8 System.Threading.Tasks.StandardTaskContinuation
                ->  0000020f002eec98 System.Threading.Tasks.ContinuationTaskFromResultTask`1[[System.Net.Http.HttpResponseMessage, System.Net.Http]]
                ->  0000020f002eeaf0 System.Action`1[[System.Threading.Tasks.Task`1[[System.Net.Http.HttpResponseMessage, System.Net.Http]], mscorlib]]
                ->  0000020f002ee860 System.Net.Http.HttpClient+<>c__DisplayClass55_0
                ->  0000020f002ee818 System.Net.Http.HttpRequestMessage
                ->  0000020f0011d8c0 System.Net.Http.MultipartFormDataContent
                ->  0000020f0011dda8 System.Collections.Generic.List`1[[System.Net.Http.HttpContent, System.Net.Http]]
                ->  0000020f001f8a90 System.Net.Http.HttpContent[]
                ->  0000020f002e9058 System.Net.Http.ByteArrayContent
                ->  0000020f9b752a60 System.Byte[]

    Found 1 unique roots (run '!GCRoot -all' to see all roots).

Ignore the DateHeaderValueManager in the GCRoot chain, that's caused by the way .NET timers use linked lists.

A quick search seems to indicate you can disable request buffering if you have direct access to an HttpWebReqeust by setting the AllowWriteStreamBuffering property to false. Unfortunately, this doesn't appear to be accessible through the HttpClient API. Does anyone here know how to disable request buffering for HttpClient?

If not, we can probably change this test to use a Socket to make the request and file an issue against HttpClient to support a way to disable request buffering.

DavidObando: @halter73 That would be awesome. I'm new to this but from what I've seen, the client buffers the entire payload before sending it to the server. I asked about this [here](https://github.com/dotnet/corefx/issues/6479#issuecomment-190306795) but haven't heard from the corefx team.

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/RequestTests.cs
cesarbs: > Do you run out of address space or something?

Yes.

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/RequestTests.cs
cesarbs: Oops on not using `content` :grin:

I can make the change to use a socket, but writing those requests by hand will be a pain.

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/RequestTests.cs
halter73: I think this should be fixed by optimizing the test by either disabling request buffering on the HttpClient or using Sockets directly as suggested below.

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/RequestTests.cs
halter73: We already use Socket's to hand roll requests in our `EngineTests`, so hopefully this won't be much more difficult. I think it's worth it to avoid such high memory usage.

cesarbs: I wonder if this has to be a multipart request at all to repro. I'll try just sending a plain large request and see what happens.

SajayAntony: @cesarbs Incase you want to kick off your socket implementation - https://github.com/sajayAntony/httpsocketclient

cesarbs:
I ended up doing things a little differently. I kept the same test but used 2048 1MB files as suggested by Stephen. In order to keep memory usage down, I just used the same byte array for each file.

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/RequestTests.cs
```diff
@@ -1,9 +1,11 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System;
 using System.Collections.Generic;
 using System.Globalization;
 using System.Net.Http;
+using System.Net.Http.Headers;
```
halter73: Remove and sort. You don't need this and a few other usings.

halter73:
@cesarbs Nice change to reduce memory usage. That is what I was trying to achieve when I used a bunch of small files, but I guess I did something wrong because I still saw large memory usage. Maybe I did something stupid like put `new byte[oneMegabyte]` in the for loop. I hope it wasn't something that dumb, but I no longer have the code so we'll never know.

halter73:
:ship: Once you fix the usings and squash.

PR #668, opened 2016-03-02
Remove 2 from MemoryPool2 and related types

halter73:
- This was merely an artifact from when this was the second of 2 pools

@davidfowl Rejoice!


davidfowl:
FINALLY!

Issue #667, opened 2016-03-02
DebugAssertException: Block being garbage collected instead of returned to pool


halter73:
This DebugAssertException has been showing up in some of our CI runs since I enabled it [a little over a week ago](https://github.com/aspnet/KestrelHttpServer/pull/623). The issue doesn't affect most test runs, but I think this exception has mostly shown up in Microsoft.AspNetCore.Server.KestrelTests.

Here is a run on AppVeyor that failed due to this assertion: https://ci.appveyor.com/project/aspnetci/kestrelhttpserver/build/1.0.2396

It's possible that Kestrel is still not properly returning blocks in some circumstances, but this could also be a test issue (particularly if it is unit tests that are causing the issue).

I know @cesarbs has already started investigating this and has set up a small VM that repros the issue more reliably than most of our workstations, so I'm assigning it to him for now.

cesarbs:
Yep, and I've already found out that the culprits are `Http10KeepAliveContentLength` in `EngineTests` and `Http10TransferEncoding` in `ChunkedRequestTests` (which was previously in `EngineTests`. I haven't found any other leaky tests (yet :grin:). The problem is that the memory blocks allocated by `SocketOutput` instances are never returned when the `KestrelEngine` instance is disposed in those tests. I haven't yet found a way to make that happen.

halter73:
@cesarbs Can you add a [trace here](https://github.com/aspnet/KestrelHttpServer/blob/9c64a492615e2fa9153021438fd053301e9a3d4a/src/Microsoft.AspNetCore.Server.Kestrel/Infrastructure/KestrelThread.cs#L135) to see if any sockets are being disposed "rudely" when this happens? That might explain the issue.

cesarbs:
To double check: do you mean a trace inside the walk, or in `OnStopRude`? I do know that `OnStopRude` is being called.

halter73:
`OnStopRude` is always called as the code is written today. That might change once #641 is fixed. What matters is if any sockets are being disposed there.

benaadams:
If that is the issue, dispose the Slabs in `OnStopXXX` so `IsActive == false` which should resolve the issue?

halter73:
@benaadams Good point. When I first made the change to uncomment the `Debug.Assert`, I didn't have the `|| IsActive == false` so a rude shutdown could cause issues. Now it shouldn't...

cesarbs:
I found the problem.

[...]

PR #669, opened 2016-03-03
Prevent DebugAssertException from blocks not returned by non-graceful shutdowns (#667).

cesarbs:
#667

cc @halter73 @benaadams

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Kestrel/Http/Listener.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/Http/Listener.cs
@@ -98,9 +98,9 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Http
                     listener._closed = true;
 
                     listener.ConnectionManager.WalkConnectionsAndClose();
-                }, this);
+                }, this).ConfigureAwait(false);
 
-                await ConnectionManager.WaitForConnectionCloseAsync();
+                await ConnectionManager.WaitForConnectionCloseAsync().ConfigureAwait(false);
 
                 await Thread.PostAsync(state =>
                 {
@@ -109,7 +109,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Http
                     {
                         writeReqPool.Dequeue().Dispose();
                     }
-                }, this);
+                }, this).ConfigureAwait(false);
             }
 
             Memory.Dispose();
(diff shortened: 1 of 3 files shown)


src/Microsoft.AspNetCore.Server.Kestrel/Http/Listener.cs
```diff
@@ -84,6 +84,8 @@ protected virtual void DispatchConnection(UvStreamHandle socket)
 
         public virtual async Task DisposeAsync()
         {
+            Memory2.Dispose();
```
halter73: I makes sense that calling `Memory2.Dispose()` here will prevent the assert failures, but so would removing the assert :stuck_out_tongue:

Joking aside, this *is* better than removing the assert since the assert can still catch non-shutdown related memory leaks which are likely the most important. However, since the assert only throws for dev builds and most of our tests shutdown the server fairly quickly, even non-shutdown related leaks might not be caught.

The reason Memory2.Dispose is called after shutdown is to suppress finallization of the existing blocks, and prevent new slabs from being allocated for blocks leased during shutdown.

I would rather find out why DisposeAsync isn't completing within the timeout. If we can't the more targeted fix would be to dispose the pools [here in KestrelEngine only if graceful shutdown fails](https://github.com/aspnet/KestrelHttpServer/blob/bb2e76c7f1a23fe4fbe0a97b29a04bfe06b7efde/src/Microsoft.AspNetCore.Server.Kestrel/KestrelEngine.cs#L119).

But to reiterate, I would *much* rather find out what's going on with DisposeAsync timing out. If the `PostAsync` callback isn't getting invoked, I really want to know why.

halter73:
Ugh. Saw [this AppVeyor failure](https://ci.appveyor.com/project/aspnetci/kestrelhttpserver/build/1.0.2409) for the "branch" build of this PR. Has anyone else seen this particular failure previously?

          Microsoft.AspNetCore.Server.KestrelTests.EngineTests.ThrowingResultsIn500Response(testContext: TestServiceContext { App = null, AppLifetime = LifetimeNotImplemented { ApplicationStarted = (throws NotImplementedException), ApplicationStopped = (throws NotImplementedException), ApplicationStopping = (throws NotImplementedException) }, DateHeaderValueManager = TestDateHeaderValueManager { }, FrameFactory = null, Log = TestKestrelTrace { }, ... }) [FAIL]
      Assert.Equal() Failure
                 (pos 0)
      Expected: HTTP/1.1 500 Internal Server Error\r\n
      Actual:   
                 (pos 0)
      Stack Trace:
        C:\projects\kestrelhttpserver\test\Microsoft.AspNetCore.Server.KestrelTests\TestConnection.cs(85,0)

Issue #674, opened 2016-03-03
Log connection id in KestrelTrace.ApplicationError


halter73:
This should be fixed now since it doesn't look like #640 will be addressed in the short term.

PR #675, opened 2016-03-03
Log connection id in KestrelTrace.ApplicationError

halter73:
- I'm open to suggestions on how to most efficiently log connection IDs
  in LoggingThreadPool as well.

#674

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Kestrel/Infrastructure/IKestrelTrace.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/Infrastructure/IKestrelTrace.cs
@@ -38,8 +38,6 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Infrastructure
 
         void NotAllConnectionsClosedGracefully();
 
-        void ApplicationError(Exception ex);
-
         void ApplicationError(string connectionId, Exception ex);
     }
 }
--- a/src/Microsoft.AspNetCore.Server.Kestrel/Infrastructure/IKestrelTrace.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/Infrastructure/IKestrelTrace.cs
@@ -39,5 +39,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Infrastructure
         void NotAllConnectionsClosedGracefully();
 
         void ApplicationError(Exception ex);
+
+        void ApplicationError(string connectionId, Exception ex);
     }
 }
--- a/src/Microsoft.AspNetCore.Server.Kestrel/Infrastructure/KestrelTrace.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/Infrastructure/KestrelTrace.cs
@@ -22,6 +22,8 @@ namespace Microsoft.AspNetCore.Server.Kestrel
         private static readonly Action<ILogger, string, int, Exception> _connectionWroteFin;
         private static readonly Action<ILogger, string, Exception> _connectionKeepAlive;
         private static readonly Action<ILogger, string, Exception> _connectionDisconnect;
+        private static readonly Action<ILogger, Exception> _applicationError;
+        private static readonly Action<ILogger, string, Exception> _applicationErrorWithId;
         private static readonly Action<ILogger, string, Exception> _connectionError;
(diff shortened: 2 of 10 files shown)


src/Microsoft.AspNetCore.Server.Kestrel/Infrastructure/IKestrelTrace.cs
```diff
@@ -39,5 +39,7 @@ public interface IKestrelTrace : ILogger
         void NotAllConnectionsClosedGracefully();
 
         void ApplicationError(Exception ex);
```
davidfowl: Who calls this?

benaadams:
> I'm open to suggestions on how to most efficiently log connection IDs in LoggingThreadPool as well.

Create one per connection, and closure capture/curry the connection
or
Drop < 4.6 and use `RunContinuationsAsynchronously` in the tcs creations; also means can drop the `IsSync` param

src/Microsoft.AspNetCore.Server.Kestrel/Infrastructure/IKestrelTrace.cs
halter73: LoggingThreadPool. @benaadams worked hard to making logging lower allocation, so I wanted his feedback on how to do it. It looks like one LoggingThreadPool per connection might be the solution.

LoggingThreadPool is also used by KestrelThread so it might not be possible to log connection Id's there, but the queued callbacks should never throw all the way up to KestrelThread, so it's really just for safety. For those maybe we should add an UnexpectedKestrelError log since it's likely not an application error.

src/Microsoft.AspNetCore.Server.Kestrel/Infrastructure/KestrelTrace.cs
```diff
@@ -43,7 +45,8 @@ static KestrelTrace()
             _connectionDisconnect = LoggerMessage.Define<string>(LogLevel.Debug, 10, @"Connection id ""{ConnectionId}"" disconnecting.");
             // ConnectionWrite: Reserved: 11
             // ConnectionWriteCallback: Reserved: 12
-            // ApplicationError: Reserved: 13 - LoggerMessage.Define overload not present 
+            _applicationError = LoggerMessage.Define(LogLevel.Error, 13, "An unhandled exception thrown by the application.");
```
cesarbs: nit: *was* thrown (?)

halter73:
@benaadams @cesarbs 

Previously, I was under the impression that user-thrown exceptions could bubble up through the `TrySetResult/Canceled/Exception` methods, but I was unable to force this to happen in my manual testing. I tried doing:

    var ignore = context.Response.Body.WriteAsync(largeBuffer).ContinueWith(_ => { throw new Exception(); }

I verified that the exception was thrown with `ConnectionLoggingThreadPool.Complete` on the stack, but it didn't bubble up through the call to `TrySetResult`, so nothing was ever logged. I was also unable get an exception from user code to be logged by `ConnectionLoggingThreadPool.Run` since this is only used by `SocketInput`, and the action is always an async continuation.

Because of this, I think we should still try/catch and log just to be safe, but log the exception as a general error instead of an application error like the new `LoggingThreadPool` in this PR does. As far as I can tell, if an exception is ever caught, it's likely a bug in Kestrel and not the application. We also don't need to create a new thread pool with three new closures for each connection this way.

Thoughts?

cesarbs:
@halter73 I'm fine with that, but could you add a test that ensures this behavior is not changed?

halter73:
@cesarbs What type of test were you thinking about. Do you want to verify that nothing still gets logged when you have something like the `ContinueWith` in my last comment? Or something else?

cesarbs:
Yep, verify that nothing gets logged. So we establish that as a contract and can expect this to remain true.

benaadams:
The user code exception should be thrown+caught only at https://github.com/aspnet/KestrelHttpServer/blob/dev/src/Microsoft.AspNetCore.Server.Kestrel/Http/FrameOfT.cs#L71-L75

halter73:
There's also [FireOnStarting/Completed](https://github.com/aspnet/KestrelHttpServer/blob/0cffed76df2c33cb93173256b0744a585a958250/src/Microsoft.AspNetCore.Server.Kestrel/Http/Frame.cs#L411) but I think that's it.

Issue #637, opened 2016-02-17
Ssl Protocol should default to TLS1.2 only


blowdart:
In options -   SslProtocols = SslProtocols.Tls12 | SslProtocols.Tls11 | SslProtocols.Tls;

Tls is considered broken and definitely shouldn't be there by default. Tls1.1 uses weak hash algorithms by default and there's some concern around that, although its yet to be broken.

bgever:
Now this has been changed, if you try to connect with older versions of TLS you may see the below error message logged to the console:
> The client and server cannot communicate, because they do not possess a common algorithm

Hope this can help others debugging why that happens.
```
fail: Microsoft.AspNetCore.Server.Kestrel[0]
      ConnectionFilter.OnConnection
System.AggregateException: One or more errors occurred. ---> System.Security.Authentication.AuthenticationException: A call to SSPI failed, see inner exception. ---> System.ComponentModel.Win32Exception: The client and server cannot communicate, because they do not possess a common algorithm
   --- End of inner exception stack trace ---
   at System.Net.Security.SslState.InternalEndProcessAuthentication(LazyAsyncResult lazyResult)
   at System.Net.Security.SslState.EndProcessAuthentication(IAsyncResult result)
   at System.Net.Security.SslStream.EndAuthenticateAsServer(IAsyncResult asyncResult)
   at System.Threading.Tasks.TaskFactory`1.FromAsyncCoreLogic(IAsyncResult iar, Func`2 endFunction, Action`1 endAction, Task`1 promise, Boolean requiresSynchronization)
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at Microsoft.AspNetCore.Server.Kestrel.Https.HttpsConnectionFilter.<OnConnectionAsync>d__6.MoveNext()
   --- End of inner exception stack trace ---
---> (Inner Exception #0) System.Security.Authentication.AuthenticationException: A call to SSPI failed, see inner exception. ---> System.ComponentModel.Win32Exception: The client and server cannot communicate, because they do not possess a common algorithm
   --- End of inner exception stack trace ---
   at System.Net.Security.SslState.InternalEndProcessAuthentication(LazyAsyncResult lazyResult)
[...]

PR #676, opened 2016-03-04
Use TLS 1.1 or 1.2 only (#637).

cesarbs:
#637

cc @blowdart @halter73

### DIFF
--- a/test/Microsoft.AspNetCore.Server.KestrelTests/HttpsConnectionFilterTests.cs
+++ b/test/Microsoft.AspNetCore.Server.KestrelTests/HttpsConnectionFilterTests.cs
@@ -8,6 +8,7 @@ using System.Net;
 using System.Net.Http;
 using System.Net.Security;
 using System.Net.Sockets;
+using System.Security.Authentication;
 using System.Security.Cryptography.X509Certificates;
 using System.Text;
 using System.Threading.Tasks;
@@ -22,6 +23,16 @@ namespace Microsoft.AspNetCore.Server.KestrelTests
 {
     public class HttpsConnectionFilterTests
     {
+#if NET451
+        static HttpsConnectionFilterTests()
+        {
+            // SecurityProtocolType values below not available in Mono < 4.3 
+            const int SecurityProtocolTypeTls11 = 768;
+            const int SecurityProtocolTypeTls12 = 3072;
+            ServicePointManager.SecurityProtocol |= (SecurityProtocolType)(SecurityProtocolTypeTls12 | SecurityProtocolTypeTls11); 
+        }
+#endif
+
         private async Task App(HttpContext httpContext)
         {
             var request = httpContext.Request;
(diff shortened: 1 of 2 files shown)


test/Microsoft.AspNetCore.Server.KestrelTests/HttpsConnectionFilterTests.cs
```diff
@@ -22,6 +23,13 @@ namespace Microsoft.AspNetCore.Server.KestrelTests
 {
     public class HttpsConnectionFilterTests
     {
+#if NET451
+        static HttpsConnectionFilterTests()
+        {
+            ServicePointManager.SecurityProtocol |= SecurityProtocolType.Tls12 | SecurityProtocolType.Tls11;
```
halter73: Do all these tests fail trying to use SslProtocols.Tls on desktop if this isn't here?

cesarbs: Actually I need to revisit that. I added this code when I restricted it to TLS 1.2 (which needs to be explicitly enabled), but since we're allowing TLS 1.1 too this might not be needed.

cesarbs:
Comments addressed; rebased & squashed.

test/Microsoft.AspNetCore.Server.KestrelTests/HttpsConnectionFilterTests.cs
```diff
@@ -22,6 +23,16 @@ namespace Microsoft.AspNetCore.Server.KestrelTests
 {
     public class HttpsConnectionFilterTests
     {
+#if NET451
+        static HttpsConnectionFilterTests()
+        {
+            // SecurityProtocolType values below only available in Mono 4.3 
+            const int SecurityProtocolTypeTls11 = 768;
+            const int SecurityProtocolTypeTls12 = 3072;
+            ServicePointManager.SecurityProtocol |= (SecurityProtocolType)(SecurityProtocolTypeTls12 | SecurityProtocolTypeTls11);
```
cesarbs: @halter73 I have to keep this otherwise tests fail on Windows.

test/Microsoft.AspNetCore.Server.KestrelTests/HttpsConnectionFilterTests.cs
```diff
@@ -22,6 +23,16 @@ namespace Microsoft.AspNetCore.Server.KestrelTests
 {
     public class HttpsConnectionFilterTests
     {
+#if NET451
+        static HttpsConnectionFilterTests()
```
cesarbs: @halter73 I have to keep this, otherwise tests fail on Windows.

test/Microsoft.AspNetCore.Server.KestrelTests/HttpsConnectionFilterTests.cs
```diff
@@ -22,6 +23,16 @@ namespace Microsoft.AspNetCore.Server.KestrelTests
...
+        {
+            // SecurityProtocolType values below not available in Mono < 4.3 
+            const int SecurityProtocolTypeTls11 = 768;
+            const int SecurityProtocolTypeTls12 = 3072;
+            ServicePointManager.SecurityProtocol |= (SecurityProtocolType)(SecurityProtocolTypeTls12 | SecurityProtocolTypeTls11);
```
cesarbs: @halter73 I have to keep this, otherwise tests fail on Windows.

test/Microsoft.AspNetCore.Server.KestrelTests/HttpsConnectionFilterTests.cs
```diff
@@ -301,5 +312,59 @@ public class HttpsConnectionFilterTests
 #endif
             }
         }
+
+        // https://github.com/aspnet/KestrelHttpServer/issues/240
+        // This test currently fails on mono because of an issue with SslStream.
```
halter73: I think we should be consistent on whether we use ConditionalFact or Fact in these tests. Can we just skip mono? I assume we aren't running tests on mono right now, but if it's not too hard it's probably better to explicitly skip it if we know it doesn't work?

cesarbs: Eh, I should have deleted that comment. It came along when I copied another test. This new test works fine on Mono.

halter73: What about the other tests in this class? Can we remove all the skip conditions?

blowdart:
:shipit:

halter73:
:shipit:

Issue #672, opened 2016-03-03
Wrong HTTP version in test response


halter73:
ChunkedRequestTests.TrailingHeadersAreParsed failed with HTTP/1.0 in a response when HTTP/1.1 was expected. This could mean that SocketInput was corrupted somehow, but there are a lot of other possible causes. Test output:

    Microsoft.AspNetCore.Server.KestrelTests.ChunkedRequestTests.TrailingHeadersAreParsed(testContext: TestServiceContext { App = null, AppLifetime = LifetimeNotImplemented { ApplicationStarted = (throws NotImplementedException), ApplicationStopped = (throws NotImplementedException), ApplicationStopping = (throws NotImplementedException) }, DateHeaderValueManager = TestDateHeaderValueManager { }, FrameFactory = null, Log = TestKestrelTrace { }, ... }) [FAIL]
      Assert.Equal() Failure
                       ↓ (pos 7)
      Expected: HTTP/1.1 200 OK\r\nContent-Length: 11\r\n\r\nHello Wor···
      Actual:   HTTP/1.0 200 OK\r\nContent-Length: 11\r\n\r\nHello Wor···
                       ↑ (pos 7)
      Stack Trace:
           at Microsoft.AspNetCore.Server.KestrelTests.TestConnection.<Receive>d__8.MoveNext()
        --- End of stack trace from previous location where exception was thrown ---
           at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
           at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
           at System.Runtime.CompilerServices.TaskAwaiter.GetResult()
           at Microsoft.AspNetCore.Server.KestrelTests.TestConnection.<ReceiveEnd>d__10.MoveNext()
        --- End of stack trace from previous location where exception was thrown ---
           at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
           at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
           at System.Runtime.CompilerServices.TaskAwaiter.GetResult()
           at Microsoft.AspNetCore.Server.KestrelTests.ChunkedRequestTests.<TrailingHeadersAreParsed>d__7.MoveNext()
        --- End of stack trace from previous location where exception was thrown ---
[...]

PR #680, opened 2016-03-08
Wait for more input while request hasn't finished (#672).

cesarbs:
There's a race-ish situation in `TakeStartLine` where we may read past available input when reading the request path. If it just so happens that the remaining input is read into memory before we try to read the HTTP version, `HttpVersion` is incorrectly set with a space at the start of the string. This in turn causes the response to be sent as HTTP/1.0, even when responding to an HTTP/1.1 request. Not only that, but it may also mess up reads of subsequent requests on the same connection (e.g. while investigating this I saw parts of the request body being read as the request method).

It would be unnecessarily messy to write a test for this (would require passing a callback to `TakeStartLine` to accurately sync/time things), but to repro (using `ChunkedRequestTests.TrailingHeadersAreParsed` from the issue) you can do the following:

- Change `TestConnection.Send()` to sleep after writing the first 5 bytes (i.e. sleep after writing the request path and before writing the space that follows it)
- Sleep a little longer in `Frame.TakeStartLine` right before attempting to read the HTTP version

I'm reviewing `TakeMessageHeaders`, looking for the same issue there.

#672

cc @halter73 @mikeharder

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Kestrel/Http/Frame.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/Http/Frame.cs
@@ -731,10 +731,18 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Http
 
                 var needDecode = false;
                 var chFound = scan.Seek(ref _vectorSpaces, ref _vectorQuestionMarks, ref _vectorPercentages);
-                if (chFound == '%')
+                if (chFound == -1)
+                {
+                    return false;
+                }
+                else if (chFound == '%')
                 {
                     needDecode = true;
                     chFound = scan.Seek(ref _vectorSpaces, ref _vectorQuestionMarks);
+                    if (chFound == -1)
+                    {
+                        return false;
+                    }
                 }
 
                 var pathBegin = begin;
@@ -851,7 +859,10 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Http
                 while (!scan.IsEnd)
                 {
                     var beginName = scan;
-                    scan.Seek(ref _vectorColons, ref _vectorCRs);
+                    if (scan.Seek(ref _vectorColons, ref _vectorCRs) == -1)
+                    {
+                        return false;
+                    }
                     var endName = scan;
 
                     chFirst = scan.Take();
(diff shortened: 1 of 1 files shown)


cesarbs:
Found similar situtation in `TakeMessageHeaders`.

src/Microsoft.AspNetCore.Server.Kestrel/Http/Frame.cs
```diff
@@ -851,7 +859,10 @@ public bool TakeMessageHeaders(SocketInput input, FrameRequestHeaders requestHea
                 while (!scan.IsEnd)
                 {
                     var beginName = scan;
-                    scan.Seek(ref _vectorColons, ref _vectorCRs);
+                    if (scan.Seek(ref _vectorColons, ref _vectorCRs) == -1)
+                    {
+                        return false;
```
halter73: In this case, isn't this already taken care of by the following?

    if (chFirst == -1 || chSecond == -1)
    {
        return false;
    }

cesarbs: I don't think so - at least I was able to make it error out. If you seek when there's no input available, the iterator doesn't move forward. Then you call `Take()` and get whatever was there under the iterator.

halter73: > If you seek when there's no input available, the iterator doesn't move forward.

I don't think this is accurate if I'm understanding what you're saying. The way `Seek` reads to me is that it advances the iterator to either the first character that matches or the end of the end of the input. `_block` and `_index` are always updated before returning even when it returns -1 (unless `IsDefault` is true which it never would be).

I do think there could be an issue if more input is added as the `TakeStartLine` or `TakeMessageHeaders` methods are running. In this way `Seek` *could* return -1 while the following `Take()` might return an actual character. This breaks the assumptions that these methods make. Namely that if the complete input for the start line or a given header isn't available fore the entire duration of the method, the method should return false.

It should be safe to return false even if all the incoming data hasn't been processed, since if the `examined` argument to `ConsumingComplete` doesn't point to the end of the input, the next `await SocketInput` should complete immediately, and the `TakeStartLine/MessageHeaders` method should be called again.

So in the end, I'm now convinced that we should always check the return value of `Seek` and return false if it's -1, but not because it's not advancing the iterator.

benaadams: Looks like MessageBody is ok https://github.com/aspnet/KestrelHttpServer/blob/dev/src/Microsoft.AspNetCore.Server.Kestrel/Http/MessageBody.cs#L394-L399

cesarbs: @halter73 Now I see that I read it wrong. But as you point out it's still an issue not checking it's return value :grin:

cesarbs: @benaadams Yep, looks fine :+1:

halter73:
:shipit:

PR #686, opened 2016-03-10
Remove race condition from socket input that could stall reads

halter73:
Also remove flakiness from tests that are closed forcefully from the server


cesarbs:
:ship:

davidfowl:
Merge from the slopes!

Issue #684, opened 2016-03-09
Error -4082 EBUSY resource busy or locked


tstar:
I have minimal MVC application with one slow method Index in Home controller. I start dnx web and in browser run http://localhost:5000/. Then I press and hold F5 button for refresh page. After some time my page is not responded. After I trying to stop Kestrel via Ctrl+C I get error:
```
C:\WebApplication1\src\WebApplication1>dnx web
Hosting environment: Production
Application base path: C:\WebApplication1\src\WebApplication1
Now listening on: http://localhost:5000
Application started. Press Ctrl+C to shut down.
System.Threading.ThreadAbortException: The stream was in the process interruption.
   в System.Threading.Monitor.Enter(Object obj)
   в Microsoft.AspNetCore.Server.Kestrel.Http.Connection.Abort()
   в Microsoft.AspNetCore.Server.Kestrel.Http.SocketOutput.OnWriteCompleted(WriteContext writeContext)
   в Microsoft.AspNetCore.Server.Kestrel.Http.SocketOutput.WriteContext.CompleteWithContextLock()
   в Microsoft.AspNetCore.Server.Kestrel.Networking.UvWriteReq.UvWriteCb(IntPtr ptr, Int32 status)
   в Microsoft.AspNetCore.Server.Kestrel.Networking.Libuv.NativeMethods.uv_run(UvLoopHandle handle, Int32 mode)
   в Microsoft.AspNetCore.Server.Kestrel.Networking.Libuv.run(UvLoopHandle handle, Int32 mode)
   в Microsoft.AspNetCore.Server.Kestrel.KestrelThread.ThreadStart(Object parameter)
--- The end of the stack trace from the previous location where the exception occurred ---
   в Microsoft.AspNetCore.Server.Kestrel.KestrelThread.Stop(TimeSpan timeout)
   в Microsoft.AspNetCore.Server.Kestrel.KestrelEngine.Dispose()
   в Microsoft.AspNetCore.Server.Kestrel.KestrelServer.Dispose()
   в Microsoft.AspNetCore.Hosting.Internal.WebHost.Dispose()
   в Microsoft.AspNetCore.Hosting.WebHostExtensions.Run(IWebHost host, CancellationToken token, String shutdownMessage)
   в Microsoft.AspNetCore.Hosting.WebHostExtensions.Run(IWebHost host)
[...]

PR #690, opened 2016-03-10
Fix deadlock when connection is simultaneously aborted and ended (#684).

cesarbs:
#684 

cc @halter73 @mikeharder @lodejard @davidfowl @benaadams

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Kestrel/Http/Connection.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/Http/Connection.cs
@@ -152,23 +152,24 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Http
 
         public virtual void Abort()
         {
-            lock (_stateLock)
+            // Frame.Abort calls user code while this method is always
+            // called from a libuv thread.
+            System.Threading.ThreadPool.QueueUserWorkItem(state =>
             {
-                if (_connectionState == ConnectionState.CreatingFrame)
+                var connection = (Connection)state;
+
+                lock (connection._stateLock)
                 {
-                    _connectionState = ConnectionState.ToDisconnect;
-                }
-                else
-                {
-                    // Frame.Abort calls user code while this method is always
-                    // called from a libuv thread.
-                    System.Threading.ThreadPool.QueueUserWorkItem(state =>
+                    if (connection._connectionState == ConnectionState.CreatingFrame)
                     {
-                        var connection = (Connection)state;
+                        connection._connectionState = ConnectionState.ToDisconnect;
+                    }
+                    else
+                    {
                         connection._frame.Abort();
-                    }, this);
+                    }
                 }
-            }
+            }, this);
         }
 
         // Called on Libuv thread


src/Microsoft.AspNetCore.Server.Kestrel/Http/Connection.cs
```diff
@@ -198,12 +191,10 @@ public virtual void OnSocketClosed()
 
         private void ApplyConnectionFilter()
```
cesarbs: There was a very obvious deadlock here too: if `_connectionState != ConnectionState.CreatingFrame` we would call `ConnectionControl.End`, which would try to lock `_stateLock`.

I'm not sure if the lock is required only for synchronizing `_connectionState` or if everything else should be inside the lock, so to be safe I moved the `lock` statement to the `if` body after atomically updating `_connectionState`.

src/Microsoft.AspNetCore.Server.Kestrel/Http/Connection.cs
```diff
@@ -35,7 +35,7 @@ public class Connection : ConnectionContext, IConnectionControl
         private readonly SocketOutput _rawSocketOutput;
 
         private readonly object _stateLock = new object();
-        private ConnectionState _connectionState;
+        private int _connectionState; // Interlocked won't work with enum
```
cesarbs: Let it be clear that I loathe having had to do this.

benaadams: And int casts? Shudder

cesarbs: That would be much better, thanks.

halter73:
I think [my suggestion](https://github.com/aspnet/KestrelHttpServer/issues/684#issuecomment-194594910) to simply take the lock inside the callback is simpler and better.

cesarbs:
@halter73 But what about updating `_connectionState`? Would we do it inside the callback?

halter73:
Exactly. You need to update the state with the lock. Doing the state change with just Interlocked.CompareExchange creates a race in `ApplyConnectionFilter`.

cesarbs:
Updated.

I wasn't sure if you meant to move everything inside the callback, since that introduces a behavior change: whereas before `Abort()` would immediately update `_connectionState`, now it delays that until the callback is executed in another thread. I'm not sure what the implications of that are and I was afraid this could introduce more issues.

src/Microsoft.AspNetCore.Server.Kestrel/Http/Connection.cs
```diff
@@ -152,23 +152,21 @@ public Task StopAsync()
 
         public virtual void Abort()
         {
-            lock (_stateLock)
+            // Frame.Abort calls user code while this method is always
+            // called from a libuv thread.
+            System.Threading.ThreadPool.QueueUserWorkItem(state =>
             {
```
halter73: Nit: Move `var connection = (Connection)state;` to the top and do `lock (connection._stateLock)`, `connection._connectionState`, etc...

src/Microsoft.AspNetCore.Server.Kestrel/Http/Connection.cs
```diff
@@ -152,23 +152,21 @@ public Task StopAsync()
...
-                if (_connectionState == ConnectionState.CreatingFrame)
-                {
-                    _connectionState = ConnectionState.ToDisconnect;
-                }
-                else
```
halter73: Where did the else go!?

cesarbs: Oops, fixed.

src/Microsoft.AspNetCore.Server.Kestrel/Http/Connection.cs
cesarbs: Done.

halter73:
:shipit:

PR #692, opened 2016-03-11
Target dnx451

bricelam:
Reverts fb8cf86f715ff8f865ba077117558b22fdeffe50 on DNX

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Kestrel/project.json
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/project.json
@@ -31,6 +31,11 @@
         "System.Threading.Tasks": { "type": "build" }
       }
     },
+    "dnx451": {
+      "frameworkAssemblies": {
+        "System.Threading.Tasks": ""
+      }
+    },
     "netstandard1.3": {
       "dependencies": {
         "System.Collections": "4.0.11-*",


cesarbs:
@bricelam Can you give more context on why this is needed? I thought we weren't going to target `dnx451` anymore.

bricelam:
It's so people can continue using RC1 tools with our packages.

khellang:
> It's so people can continue using RC1 tools with our packages.

Shouldn't `net451` cover this?

davidfowl:
@bricelam I think we should just let people put this in their projects instead.

bricelam:
The commit referenced broke that. CLI pulls in *some* facades, but DNX doesn't get any.

@davidfowl I'm fine with that. @dougbu requested this change.

dougbu:
>  @dougbu requested this change

Yes but @Eilon was less concerned about me than the many people out in the real world trying to use `dev`.

muratg:
Where are we with these? Still needed?

davidfowl:
We shouldn;t make these temporary changes to add dnx anywhere. We're close to having what we want to ship ideally. I'd rather hack DNX to make it work than change the shipping packages.

Eilon:
@davidfowl I'm actually ok either way - the question is how desperate the need is. If adding it to Kestrel unblocks something we need right now, let's do that. But if it's not a pressing need and we can wait for a DNX hack, that's certainly a better option.

bricelam:
Recalling. After chatting with @Eilon, we decided that adding the facades to user's `project.json` is an acceptable workaround.

Issue #666, opened 2016-03-02
500 when abs_path present in http_URL - 3.2.2 http URL


SajayAntony:
Kestrel returns `HTTP/1.1 500 Internal Server Error` when the request has a complete request URI. 
Sample app - [HelloWorldMvc](https://github.com/aspnet/Performance/tree/dev/testapp/HelloWorldMvc)
#### Program.cs

``` csharp

using System;
using System.Net;
using System.Net.Sockets;
using System.Text;

namespace KestrelFullRequestUriIssue
{
    public class Program
    {
        public static void Main(string[] args)
        {
            var endpoint = new DnsEndPoint("localhost", 5000);
            Socket s = new Socket(SocketType.Stream, ProtocolType.Tcp);
            s.Connect(endpoint);
            var raw = "GET http://localhost:5000/ HTTP/1.1\r\n" +
                            "Host: localhost:5000\r\n" +
                            "Content-Length: 0\r\n" +
                            "Connection: close\r\n" +
                            "\r\n";

            var request = Encoding.ASCII.GetBytes(raw);
            Console.WriteLine(raw);
            s.Send(Encoding.ASCII.GetBytes(raw));
            byte[] buffer = new byte[1024];
            int count = 0;
            while ((count = s.Receive(buffer)) != 0)
                Console.WriteLine(Encoding.ASCII.GetChars(buffer, 0, count));
            s.Dispose();
        }
    }
}

```
#### project.json

``` json
{
  "version": "1.0.0-*",
  "description": "KestrelFullRequestUriIssue Console Application",
  "authors": [ "sajaya" ],
  "tags": [ "" ],
  "projectUrl": "",
  "licenseUrl": "",

  "compilationOptions": {
    "emitEntryPoint": true
  },

  "dependencies": {
  },

  "commands": {
    "KestrelFullRequestUriIssue": "KestrelFullRequestUriIssue"
  },

  "frameworks": {
    "dnx451": { },
    "dnxcore50": {
      "dependencies": {
        "Microsoft.CSharp": "4.0.1-beta-23516",
        "System.Collections": "4.0.11-beta-23516",
        "System.Console": "4.0.0-beta-23516",
        "System.Linq": "4.0.1-beta-23516",
        "System.Net.NetworkInformation": "4.0.0-beta-23019",
        "System.Net.Primitives": "4.0.11-beta-23516",
        "System.Private.Networking": "4.0.1-beta-23225",
[...]

Issue #709, opened 2016-03-22
Dealing with OPTIONS * requests


Tratcher:
Today Kestrel will take the \* from an OPTIONS \* request and put it in the Path. However, \* isn't a path and PathString throws.

When Kestrel sees a lone \* in the request line it should leave both Path and PathBase empty.

https://github.com/aspnet/ServerTests/issues/20#issuecomment-200049192

muratg:
Per triage meeting, this was done.

Issue #594, opened 2016-01-21
Expose raw request path/URL in HTTP context


cesarbs:
After normalization (see #573) we lose the raw request path/URL received from the client. Some applications might be interested in it, so we should expose the pre-normalized path/URL to them.

Tratcher:
This will likely be exposed via a standard feature interface.

DamianEdwards:
post RC2

cesarbs:
Done https://github.com/aspnet/KestrelHttpServer/pull/880

PR #711, opened 2016-03-23
Ignore everything but path and query in requests for absolute URIs (#666).

cesarbs:
cc @Tratcher @halter73 @SajayAntony 

#666

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Kestrel/Http/Frame.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/Http/Frame.cs
@@ -804,8 +804,21 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Http
                 QueryString = queryString;
                 HttpVersion = httpVersion;
 
-                bool caseMatches;
+                if (requestUrlPath.Length > 0 && requestUrlPath[0] != '/')
+                {
+                    Uri uri;
+                    if (Uri.TryCreate(requestUrlPath, UriKind.Absolute, out uri))
+                    {
+                        requestUrlPath = uri.PathAndQuery;
+                    }
+                    else
+                    {
+                        ReportCorruptedHttpRequest(new BadHttpRequestException($"Invalid request path: {requestUrlPath}"));
+                        return true;
+                    }
+                }
 
+                bool caseMatches;
                 if (!string.IsNullOrEmpty(_pathBase) &&
                     (requestUrlPath.Length == _pathBase.Length || (requestUrlPath.Length > _pathBase.Length && requestUrlPath[_pathBase.Length] == '/')) &&
                     RequestUrlStartsWithPathBase(requestUrlPath, out caseMatches))
--- a/src/Microsoft.AspNetCore.Server.Kestrel/Http/Frame.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/Http/Frame.cs
@@ -806,15 +806,37 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Http
 
                 if (requestUrlPath.Length > 0 && requestUrlPath[0] != '/')
                 {
-                    Uri uri;
-                    if (Uri.TryCreate(requestUrlPath, UriKind.Absolute, out uri))
+                    int hostIndex;
+                    if (requestUrlPath.StartsWith("http://"))
                     {
-                        requestUrlPath = uri.PathAndQuery;
+                        hostIndex = 7;
+                    }
+                    else if (requestUrlPath.StartsWith("https://"))
+                    {
+                        hostIndex = 8;
                     }
                     else
                     {
                         ReportCorruptedHttpRequest(new BadHttpRequestException($"Invalid request path: {requestUrlPath}"));
                         return true;
+                    }
+
+                    int pathIndex = requestUrlPath.IndexOf('/', hostIndex);
+                    if (pathIndex == -1)
+                    {
+                        int queryIndex = requestUrlPath.IndexOf('?', hostIndex);
+                        if (queryIndex == -1)
+                        {
+                            requestUrlPath = "/";
+                        }
+                        else
+                        {
+                            requestUrlPath = "/" + requestUrlPath.Substring(queryIndex);
+                        }
+                    }
+                    else
+                    {
+                        requestUrlPath = requestUrlPath.Substring(pathIndex);
                     }
                 }
 
--- a/src/Microsoft.AspNetCore.Server.Kestrel/Http/FrameOfT.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/Http/FrameOfT.cs
@@ -54,6 +54,12 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Http
 
                     InitializeHeaders();
 
+                    if (_corruptedRequest)
+                    {
+                        await ProduceEnd();
+                        return;
+                    }
+
                     while (!_requestProcessingStopping && !TakeMessageHeaders(SocketInput, FrameRequestHeaders))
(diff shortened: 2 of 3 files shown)


src/Microsoft.AspNetCore.Server.Kestrel/Http/Frame.cs
```diff
@@ -804,8 +804,21 @@ protected bool TakeStartLine(SocketInput input)
                 QueryString = queryString;
                 HttpVersion = httpVersion;
 
-                bool caseMatches;
+                if (requestUrlPath.Length > 0 && requestUrlPath[0] != '/')
+                {
+                    Uri uri;
+                    if (Uri.TryCreate(requestUrlPath, UriKind.Absolute, out uri))
```
cesarbs: Not optimal but I don't see the need to optimize for this case - requests for absolute URIs are not expected to be the norm. So most of the time we won't be running this.

Tratcher: System.Uri can't be used, it does lots of its own escaping and un-escaping.  I think your three interesting cases are:
- Starts with / if it's a path
- * for OPTIONS
- starts with http:// or https:// if its absolute, skip to the first / or ? after the scheme separator

cesarbs: Ok. I'll address `OPTIONS *` in a subsequent PR.

src/Microsoft.AspNetCore.Server.Kestrel/Http/Frame.cs
```diff
@@ -804,8 +804,43 @@ protected bool TakeStartLine(SocketInput input)
                 QueryString = queryString;
                 HttpVersion = httpVersion;
 
-                bool caseMatches;
+                if (requestUrlPath.Length > 0 && requestUrlPath[0] != '/')
```
cesarbs: Again, this is an edge case and I don't see the need to optimize this path.

Tratcher: what happens for requestUrlPath.Lenght == 0? That shouldn't be valid either.

Tratcher: Add a comment and RFC link for the scenario you're handling, and the fact that you're ignoring the scheme and host at the moment.

src/Microsoft.AspNetCore.Server.Kestrel/Http/Frame.cs
```diff
@@ -804,8 +804,43 @@ protected bool TakeStartLine(SocketInput input)
...
+                    {
+                        hostIndex = 7;
+                    }
+                    else if (requestUrlPath.StartsWith("https://"))
+                    {
+                        hostIndex = 8;
+                    }
+                    else
+                    {
+                        ReportCorruptedHttpRequest(new BadHttpRequestException($"Invalid request path: {requestUrlPath}"));
```
Tratcher: officially it's the request target, not path.

Tratcher:
The URL processing looks OK. @halter73 how about the 400 response?

src/Microsoft.AspNetCore.Server.Kestrel/Http/Frame.cs
```diff
@@ -804,8 +804,43 @@ protected bool TakeStartLine(SocketInput input)
...
+                        return true;
+                    }
+
+                    int pathIndex = requestUrlPath.IndexOf('/', hostIndex);
+                    if (pathIndex == -1)
+                    {
+                        int queryIndex = requestUrlPath.IndexOf('?', hostIndex);
```
halter73: Isn't the query string already removed from `requestUrlPath`?

src/Microsoft.AspNetCore.Server.Kestrel/Http/Frame.cs
```diff
@@ -804,8 +804,43 @@ protected bool TakeStartLine(SocketInput input)
                 QueryString = queryString;
                 HttpVersion = httpVersion;
 
-                bool caseMatches;
+                if (requestUrlPath.Length > 0 && requestUrlPath[0] != '/')
+                {
+                    int hostIndex;
+                    if (requestUrlPath.StartsWith("http://"))
```
SajayAntony: I'd be interested in the allocation profile on this. Just curious if we could just operate on the bytes before materializing to a string or all the operations here non-allocating?

SajayAntony: I'm guessing the normal path is a relative URI this path might not even showup. Is that correct?

---
In reply to: [57374946](https://github.com/aspnet/KestrelHttpServer/pull/711#discussion_r57374946) [](ancestors = 57374946)

src/Microsoft.AspNetCore.Server.Kestrel/Http/FrameOfT.cs
```diff
@@ -54,6 +54,12 @@ public class Frame<TContext> : Frame
 
                     InitializeHeaders();
 
+                    if (_corruptedRequest)
+                    {
+                        await ProduceEnd();
+                        return;
+                    }
+
```
halter73: This seems different than what we do for corrupted request headers. I would like to determine the correct behavior and consolidate this logic.

src/Microsoft.AspNetCore.Server.Kestrel/Http/Frame.cs
SajayAntony: Also do you want play back user input security wise?

---
In reply to: [57367957](https://github.com/aspnet/KestrelHttpServer/pull/711#discussion_r57367957) [](ancestors = 57367957)

halter73:
I wonder if we should take this opportunity to return a 400 for non HTTP/1.1 or HTTP/1.0 requests.

src/Microsoft.AspNetCore.Server.Kestrel/Http/Frame.cs
cesarbs: Totally overlooked that.

Tratcher: Hmm, looking at the code that brings up an interesting point. pathBegin and pathEnd are now the wrong name, it should be urlBegin or targetBegin. Similar for requestUrlPath.

cesarbs: Yeah, I'm renaming those.

src/Microsoft.AspNetCore.Server.Kestrel/Http/Frame.cs
cesarbs: It's what I'm expecting.

Tratcher:
Or 505 HTTP Version Not Supported ? https://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html

src/Microsoft.AspNetCore.Server.Kestrel/Http/Frame.cs
bbowyersmyth: Specify the StringComparison to avoid a CurrentCulture comparision. Is requestUrlPath expected to be lower case?

Tratcher: ordinal ignore case.

cesarbs:
We've decided this should be handled after RC2, since it's an edge case. We can more adequately handle this scenario, as well as #709, once we implement #594.

cesarbs:
I'll file an issue for the wrong HTTP version too.

PR #721, opened 2016-04-01
Don't run restore in #initialize

pranavkm:


makefile.shade
```diff
@@ -8,8 +8,5 @@ k-standard-goals
 custom-goals
 
 #initialize if='Directory.Exists("src")'
-  exec program='dotnet' commandline='restore ../../tools/Microsoft.AspNetCore.Server.Kestrel.GeneratedCode' workingdir='${kestrelSrcDir}'
```
pranavkm: This is taken care by `#repo-iniitialize` which runs before this

Issue #704, opened 2016-03-18
Race when consuming start line and headers sent immediately before client FIN


halter73:
Example test failure due to the bug:

    [Test collection for Microsoft.AspNetCore.Server.KestrelTests.EngineTests (1)] Microsoft.AspNetCore.Server.KestrelTests.EngineTests.ZeroContentLengthSetAutomaticallyAfterNoWrites(testContext: TestServiceContext { App = null, AppLifetime = LifetimeNotImplemented { ApplicationStarted = (throws NotImplementedException), ApplicationStopped = (throws NotImplementedException), ApplicationStopping = (throws NotImplementedException) }, DateHeaderValueManager = TestDateHeaderValueManager { }, FrameFactory = null, Log = TestKestrelTrace { }, ... })
    [16:48:45]
    [Microsoft.AspNetCore.Server.KestrelTests.EngineTests.ZeroContentLengthSetAutomaticallyAfterNoWrites(testContext: TestServiceContext { App = null, AppLifetime = LifetimeNotImplemented { ApplicationStarted = (throws NotImplementedException), ApplicationStopped = (throws NotImplementedException), ApplicationStopping = (throws NotImplementedException) }, DateHeaderValueManager = TestDateHeaderValueManager { }, FrameFactory = null, Log = TestKestrelTrace { }, ... })] Assert.Equal() Failure
                                           (pos 38)
    Expected: úúúontent-Length: 0\r\n\r\nHTTP/1.0 200 OK\r\nConnection: keep-alive\r\núúú
    Actual:   úúúontent-Length: 0\r\n\r\n
                                           (pos 38)
       at Microsoft.AspNetCore.Server.KestrelTests.TestConnection.<Receive>d__8.MoveNext() in D:\b\w\e70224f4ea5e88a1\KestrelHttpServer\test\Microsoft.AspNetCore.Server.KestrelTests\TestConnection.cs:line 85
    --- End of stack trace from previous location where exception was thrown ---
       at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
       at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
[...]

PR #722, opened 2016-04-01
Ensure entire request body is consumed before handling client FIN

halter73:
This completes the fix for #704 

This prevents `Request.ReadAsync` from throwing when it should complete successfully right after the client sends a FIN.

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Kestrel/Http/MessageBody.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/Http/MessageBody.cs
@@ -249,29 +249,46 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Http
                 {
                     while (_mode == Mode.Prefix)
                     {
+                        var fin = input.RemoteIntakeFin;
+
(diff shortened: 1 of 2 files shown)


cesarbs:
Looks good but I think you should also throw from https://github.com/aspnet/KestrelHttpServer/blob/dev/src/Microsoft.AspNetCore.Server.Kestrel/Http/FrameOfT.cs#L44 and https://github.com/aspnet/KestrelHttpServer/blob/fd70fb732d28d2c8a2ef871f6ef0d9487a6de46e/src/Microsoft.AspNetCore.Server.Kestrel/Http/FrameOfT.cs#L65

cesarbs:
Tests?

halter73:
I don't think it's wrong to close connections without a response if the client sends a FIN before completing the start line or headers. Right now, we only ever return a 400/500 response automatically if the app func has been called with the app failing without writing a response.

I think pushing a change to how we handle incomplete requests is a little to risky to do in ask mode for such an edge case.

@Tratcher

Issue #725, opened 2016-04-04
Remove the thread.Abort on the kestrel thread


davidfowl:
https://github.com/aspnet/KestrelHttpServer/blob/dev/src/Microsoft.AspNetCore.Server.Kestrel/Infrastructure/KestrelThread.cs#L120

We should try to avoid different behavior on .NET Core vs .NET Framework where possible. What happens on .NET Core when the timeout elapses and the UV thread still hasn't stopped running?

/cc @halter73 @cesarbs

benaadams:
Meh, can just drop these; the threads are background so they will just die anyway

cesarbs:
@benaadams Agreed. I had thought something similar right when I started working on Kestrel, but thought there was an important reason for those `Abort`s

benaadams:
@cesarbs wasn't till 19th Feb they went to background mode https://github.com/aspnet/KestrelHttpServer/pull/646

cesarbs:
@benaadams Ah, thanks for pointing that out.

PR #726, opened 2016-04-05
Never call Thread.Abort

halter73:
- This API isn't available on .NET Core so we should rely entirely on
  our other methods of terminating the libuv threads.

#725


benaadams:
:rocket:

davidfowl:
:shipit:

davidfowl:
I guess this needs to wait (ask mode?) /cc @Eilon

cesarbs:
:ship:

PR #727, opened 2016-04-05
Remove unused dependency on TraceSource

davidfowl:
- TraceSource is RID specific (see https://github.com/dotnet/corefx/issues/7480) and it
causes 2 dlls to end up in the output of every ASP.NET application. We don't even
use it anymore.

I want to get this in RC2. It reduces the output size by `109kb`

/cc @danroth27 @Eilon


DamianEdwards:
:shipit:

Issue #738, opened 2016-04-12
Server crashes due to null reference when connection is aborted before frame is created


yuwaMSFT:
InnerException:   <none>
StackTrace (generated):
    SP               IP               Function
    00007F0C7E7FB810 00007F0CD98A161A Microsoft.AspNetCore.Server.Kestrel.dll!Microsoft.AspNetCore.Server.Kestrel.Http.Connection+<>c.<Abort>b__17_0(System.Object)+0x7a
    00007F0C7E7FB840 00007F0CD7C7A061 mscorlib.ni.dll!System.Threading.ExecutionContext.Run(System.Threading.ExecutionContext, System.Threading.ContextCallback, System.Object)+0xe1
    00007F0C7E7FB890 00007F0CD7DEB613 mscorlib.ni.dll!System.Threading.QueueUserWorkItemCallbackDefaultContext.System.Threading.IThreadPoolWorkItem.ExecuteWorkItem()+0x43
    00007F0C7E7FB8B0 00007F0CD7CE5DA1 mscorlib.ni.dll!System.Threading.ThreadPoolWorkQueue.Dispatch()+0x1b1

StackTraceString: <none>
HResult: 80004003

Nested exception -------------------------------------------------------------
Exception object: 00007f0cb8e5d7f8
Exception type:   System.NullReferenceException
Message:          Object reference not set to an instance of an object.
InnerException:   <none>
StackTrace (generated):
    SP               IP               Function
    00007F0C7E7FB810 00007F0CD98A161A Microsoft.AspNetCore.Server.Kestrel.dll!Microsoft.AspNetCore.Server.Kestrel.Http.Connection+<>c.<Abort>b__17_0(System.Object)+0x7a
    00007F0C7E7FB840 00007F0CD7C7A061 mscorlib.ni.dll!System.Threading.ExecutionContext.Run(System.Threading.ExecutionContext, System.Threading.ContextCallback, System.Object)+0xe1
    00007F0C7E7FB890 00007F0CD7DEB613 mscorlib.ni.dll!System.Threading.QueueUserWorkItemCallbackDefaultContext.System.Threading.IThreadPoolWorkItem.ExecuteWorkItem()+0x43
    00007F0C7E7FB8B0 00007F0CD7CE5DA1 mscorlib.ni.dll!System.Threading.ThreadPoolWorkQueue.Dispatch()+0x1b1

(lldb) sos DumpObj 00007f0cb8e2ea48
[...]

PR #739, opened 2016-04-12
Prevent crash when aborting connection before Frame is created (#738).

cesarbs:
- Check that `_frame` is not null
- Queue work item via `LoggingThreadPool` to isolate future potential exceptions and prevent new crashes

#738

cc @halter73 @benaadams @yuwaMSFT


yuwaMSFT:
:shipit:

halter73:
:shipit:

Issue #720, opened 2016-04-01
Add IWebHostBuilder.UseKestrel(options) overload to configure Kestrel


mikeharder:
Copied from #713:

```C#
var host = new WebHostBuilder()
    .UseKestrel(options =>
     {
         options.NoDelay = true;
         options.ThreadCount = 4;
         options.UseHttps(...);
     })
    .UseStartup<StartupHelloWorld>()
    .Build();

host.Run();
```

We should also remove these as a result of this:

https://github.com/aspnet/KestrelHttpServer/blob/dev/src/Microsoft.AspNetCore.Server.Kestrel.Https/HttpsApplicationBuilderExtensions.cs

It clears up the fact that these aren't actually pieces of middleware but server configuration.

DamianEdwards:
YES PLS RIGHT NOW LOVE IT :shipit:

davidfowl:
I did a test implementation and there are some problems around extension methods that require resolved services to configure. `UseConnectionLogging()` requires an `ILoggerFactory`and `UseHttps()` and requires the base path to the x509cert.

davidfowl:
Kestrel
- Add `KestrelServerOptions` and remove `KestrelServerInformation`
- Remove `KestrelServerInformation` from server features
- Add `IServiceProvider` to `KestrelServerOptions`
- Change UseKestrel() to add the IServerFactory to the container directly
- Add `UseKestrel(Action<KestrelServerOptions>)` which adds 
- Change connection filter extension methods to extension methods on `KestrelServerOptions`

PR #741, opened 2016-04-13
Add IWebHostBuilder.UseKestrel(options) overload to configure Kestrel (#720)

mikeharder:
Src and samples have been changed.  Still working on tests.  Most diffs are just renames from "information" to "options".

@halter73, @davidfowl, @Tratcher, @DamianEdwards


davidfowl:
This looks good to me

src/Microsoft.AspNetCore.Server.Kestrel.Https/KestrelServerOptionsExtensions.cs
```diff
@@ -0,0 +1,33 @@
...
+using System.IO;
+using System.Security.Cryptography.X509Certificates;
+using Microsoft.AspNetCore.Hosting;
+using Microsoft.AspNetCore.Server.Kestrel.Filter;
+using Microsoft.AspNetCore.Server.Kestrel.Https;
+using Microsoft.Extensions.DependencyInjection;
+
+namespace Microsoft.AspNetCore.Server.Kestrel
+{
+    public static class KestrelServerOptionsExtensions
```
Tratcher: KestrelServerOptionsHttpsExtensions

benaadams:
Its less exotic that current, which can only be good

:rocket:

src/Microsoft.AspNetCore.Server.Kestrel/Filter/LoggingFilterKestrelOptionsExtensions.cs
```diff
@@ -0,0 +1,33 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.AspNetCore.Server.Kestrel.Filter
```
Tratcher: We can bring this into the Microsoft.AspNetCore.Server.Kestrel namespace with the options type. We just didn't want it in the Microsoft.AspNetCore.Builder namespace.

Tratcher:
This should be targeting Release, no?

davidfowl:
> This should be targeting Release, no?

Yep, we spoke about this and @mikeharder will move it to release.

Tratcher:
:+1:

davidfowl:
Appveyor seems like a random failure

src/Microsoft.AspNetCore.Server.Kestrel/Filter/LoggingFilterKestrelOptionsExtensions.cs
Tratcher: Now that I think about it, it doesn't matter if this is in the same namespace as the options type because it's used inside a lamda.

src/Microsoft.AspNetCore.Server.Kestrel/KestrelWebHostBuilderExtensions.cs
```diff
@@ -1,16 +1,31 @@
...
 {
     public static class KestrelWebHostBuilderExtensions
     {
         public static IWebHostBuilder UseKestrel(this IWebHostBuilder hostBuilder)
         {
-            return hostBuilder.UseServer(typeof(KestrelServer).GetTypeInfo().Assembly.FullName);
+            return hostBuilder.ConfigureServices(services => services.AddSingleton<IServerFactory, ServerFactory>());
+        }
+
+        public static IWebHostBuilder UseKestrel(this IWebHostBuilder hostBuilder, Action<KestrelServerOptions> options)
```
Tratcher: Is there some special reason we need to configure the options via WebHostBuilder rather than doing it in Start.ConfigureServices with a `ConfigureKestrel(Action<KestrelServerOptions> options)` extension like we'd use for middleware? This won't execute until after ConfigureServices anyways. And then devs can use services injected via Startup's ctor.

davidfowl: We want to split server configuration from application configuration. Of course the user can also configure options in their startup class but this is the default experience we want to preserve. It also keeps the server specific stuff in Program.Main instead of putting it in the user's Startup class by default.

mikeharder:
@davidfowl, @DamianEdwards, @Tratcher, @halter73 

What namespaces should we use for the new extension methods UseHttps(this KestrelServerOptions and UseConnectionLogging(this KestrelServerOptions)?

The typical convention is to put extension methods in the namespace of the object they extend, which would be Microsoft.AspNetCore.Server.Kestrel in this case.  However, this may be slightly confusing, since customers could call UseKestrel() without the Kestrel namespace, and may properties would be settable, but the extension methods UseHttps() and UseConnectionLogging() would not be available until they added the Kestrel namespace.

Alternatively, we could put the new extension methods in the same namespace as UseKestrel(), which is Microsoft.AspNetCore.Hosting.  However, this breaks the convention of having extension methods in the same namespace as the object they are extending.

Tratcher:
Put them in `Microsoft.AspNetCore.Server.Kestrel`. Trying to guess 2nd degree call sites is going to break down to quickly as there are multiple places you can configure the options.

halter73:
:+1: for Microsoft.AspNetCore.Hosting namespaced KestrelServerOptions*Extensions.

halter73:
:ship:

PR #744, opened 2016-04-13
Fix package metadata

Eilon:
@halter73 @Tratcher 

FYI this is for release/RC2.


Tratcher:
:shipit:

halter73:
:shipit:

Eilon:
@halter73 will push this, thanks for the reviews!

Issue #742, opened 2016-04-13
Remove DNX mono support for loading libuv


davidfowl:
It's time to get rid of this code

/cc @moozzyk

PR #745, opened 2016-04-13
Remove support for DNX on osx on mono

davidfowl:
- Remove mono osx specific PInvokes
- Use PlatformAbstractions to detect darwin instead of direct pinvoke

#742 

/cc @moozzyk @halter73

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Kestrel/Networking/PlatformApis.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/Networking/PlatformApis.cs
@@ -21,41 +21,13 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Networking
             if (!IsWindows)
             {
                 // When running on Mono in Darwin OSVersion doesn't return Darwin. It returns Unix instead.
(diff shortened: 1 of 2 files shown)


src/Microsoft.AspNetCore.Server.Kestrel/Networking/PlatformApis.cs
```diff
@@ -21,41 +21,13 @@ static PlatformApis()
             if (!IsWindows)
             {
                 // When running on Mono in Darwin OSVersion doesn't return Darwin. It returns Unix instead.
```
moozzyk: Should this comment be removed?

moozzyk:
Minor comment then :ship::it:

src/Microsoft.AspNetCore.Server.Kestrel/Networking/PlatformApis.cs
davidfowl: The comment is still valid. Though we could replace all of this code with platform abstractions.

PR #749, opened 2016-04-14
Remove IServerFactory

Tratcher:
https://github.com/aspnet/Hosting/pull/702/
@halter73 @davidfowl @JunTaoLuo @DamianEdwards @mikeharder @muratg


src/Microsoft.AspNetCore.Server.Kestrel/KestrelServer.cs
```diff
@@ -4,12 +4,14 @@
 using System;
 using System.Collections.Generic;
 using Microsoft.AspNetCore.Hosting;
+using Microsoft.AspNetCore.Hosting.Internal;
```
mikeharder: It seems wrong for KestrelServer to depend on Hosting.Internal.  I would move ServerAddressesFeature to a public namespace.

src/Microsoft.AspNetCore.Server.Kestrel/KestrelServer.cs
```diff
@@ -36,15 +34,19 @@ public KestrelServer(IFeatureCollection features, KestrelServerOptions options,
...
-                throw new ArgumentNullException(nameof(logger));
+                throw new ArgumentNullException(nameof(loggerFactory));
             }
 
+            Options = options.Value ?? new KestrelServerOptions();
             _applicationLifetime = applicationLifetime;
-            _logger = logger;
-            Features = features;
-            Options = options;
+            _logger = loggerFactory.CreateLogger("Microsoft.AspNetCore.Server.Kestrel");
```
mikeharder: I would use `typeof(KestrelServer).GetTypeInfo().Assembly.FullName` instead of hardcoding `"Microsoft.AspNetCore.Server.Kestrel"`.

src/Microsoft.AspNetCore.Server.Kestrel/KestrelServer.cs
```diff
@@ -94,13 +96,7 @@ public void Start<TContext>(IHttpApplication<TContext> application)
...
                 var atLeastOneListener = false;
 
-                var addressesFeature = Features.Get<IServerAddressesFeature>();
-                if (addressesFeature == null)
-                {
-                    throw new InvalidOperationException($"{nameof(IServerAddressesFeature)} is missing.");
-                }
-
-                foreach (var address in addressesFeature.Addresses)
+                foreach (var address in _serverAddresses.Addresses)
```
mikeharder: Would it be more correct to get the `IServerAddressesFeature` from `Features.Get()`, rather than caching it in the ctor?  Is it possible for something else to change the IServerAddressFeature between KestrelServer.ctor() and KestrelServer.Start()?

src/Microsoft.AspNetCore.Server.Kestrel/KestrelServer.cs
```diff
@@ -36,15 +34,19 @@ public KestrelServer(IFeatureCollection features, KestrelServerOptions options,
...
             _applicationLifetime = applicationLifetime;
-            _logger = logger;
-            Features = features;
-            Options = options;
+            _logger = loggerFactory.CreateLogger("Microsoft.AspNetCore.Server.Kestrel");
+            Features = new FeatureCollection();
+            var componentFactory = new HttpComponentFactory(Options);
+            Features.Set<IHttpComponentFactory>(componentFactory);
+            _serverAddresses = new ServerAddressesFeature();
+            Features.Set<IServerAddressesFeature>(_serverAddresses);
```
halter73: I thought Hosting was going to set the `IServerAddressesFeature` and the servers could merely replace it if they want to?

@davidfowl

src/Microsoft.AspNetCore.Server.Kestrel/KestrelServer.cs
Tratcher: I consciously prevented that. The server owns the FeatureCollection and may map the IServerAddressesFeature to some internal structure. We don't want apps to replace it and have it work on some servers and not others.

src/Microsoft.AspNetCore.Server.Kestrel/KestrelServer.cs
Tratcher: See my comment below.

src/Microsoft.AspNetCore.Server.Kestrel/KestrelServer.cs
davidfowl: Agreed

src/Microsoft.AspNetCore.Server.Kestrel/KestrelServer.cs
```diff
@@ -36,15 +34,19 @@ public KestrelServer(IFeatureCollection features, KestrelServerOptions options,
...
 
+            Options = options.Value ?? new KestrelServerOptions();
             _applicationLifetime = applicationLifetime;
-            _logger = logger;
-            Features = features;
-            Options = options;
+            _logger = loggerFactory.CreateLogger("Microsoft.AspNetCore.Server.Kestrel");
+            Features = new FeatureCollection();
+            var componentFactory = new HttpComponentFactory(Options);
+            Features.Set<IHttpComponentFactory>(componentFactory);
```
davidfowl: Why is this a feature?

Tratcher:
Updated

src/Microsoft.AspNetCore.Server.Kestrel/KestrelServer.cs
Tratcher: It seems to be replaceable and will then be used for stream and header pooling. I agree this is a strange place to put it. Why not just resolve it from DI?

Regardless, this is unrelated to the current changes. I just moved it from ServerFactory.

mikeharder:
:shipit:

halter73:
ex post facto :ship:

Issue #747, opened 2016-04-13
Got exception on accessing disposed object "SslStream" when running concurrent clients on https


yuwaMSFT:
I was running some simple https scenarios with load. occasionally I saw the exception error printed out:

```
fail: Microsoft.AspNetCore.Server.Kestrel[13]
      Connection id "0HKR3RK20AN4V": An unhandled exception was thrown by the application.
**System.ObjectDisposedException: Cannot access a disposed object.
Object name: 'SslStream'.**
   at System.Net.Security.SslState.CheckThrow(Boolean authSucessCheck)
   at System.Net.Security.SslState.get_SecureStream()
   at System.Net.Security.SslStream.Write(Byte[] buffer, Int32 offset, Int32 count)
   at Microsoft.AspNetCore.Server.Kestrel.Filter.StreamSocketOutput.ProducingComplete(MemoryPoolIterator end)
   at Microsoft.AspNetCore.Server.Kestrel.Http.Frame.CreateResponseHeader(Byte[] statusBytes, Boolean appCompleted)
   at Microsoft.AspNetCore.Server.Kestrel.Http.Frame.<ProduceStartAndFireOnStartingAwaited>d__127.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at Microsoft.AspNetCore.Server.Kestrel.Http.Frame.<WriteAsyncAwaited>d__120.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at Microsoft.AspNetCore.WebUtilities.HttpResponseStreamWriter.<FlushInternalAsync>d__26.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
[...]

Issue #737, opened 2016-04-12
Exception at System.Net.Security.SslStreamInternal.EndRead when proxy request from Nginx to Kestrel through HTTPS


yuwaMSFT:
I was trying out the SSL feature on Kestrel. When putting Nginx in front to proxy request to Kestrel server via https, I always got the following exception for each client request:

\\funfile\Scratch\yuwa\Failures\041216\kestrelhttps\kestrellog.log

```
fail: Microsoft.AspNetCore.Server.Kestrel[0]
      FilteredStreamAdapter.CopyToAsync
System.AggregateException: One or more errors occurred. (The read operation failed, see inner exception.) ---> System.IO.IOException: The read operation failed, see inner exception. ---> System.Threading.Tasks.TaskCanceledException: The request was aborted
   at Microsoft.AspNetCore.Server.Kestrel.Http.SocketInput.GetResult()
   at Microsoft.AspNetCore.Server.Kestrel.Http.SocketInputExtensions.<ReadAsyncAwaited>d__1.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at System.Threading.Tasks.ValueTask`1.get_Result()
   at Microsoft.AspNetCore.Server.Kestrel.Filter.LibuvStream.Read(Byte[] buffer, Int32 offset, Int32 count)
   at System.Threading.Tasks.Task`1.InnerInvoke()
   at System.Threading.Tasks.Task.Execute()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at System.IO.StreamAsyncHelper.EndRead(IAsyncResult asyncResult)
   at System.Net.FixedSizeReader.ReadCallback(IAsyncResult transportResult)
   --- End of inner exception stack trace ---
   at System.Net.Security.SslStreamInternal.EndRead(IAsyncResult asyncResult)
   at System.Net.Security.SslStream.<>c.<ReadAsync>b__85_1(IAsyncResult iar)
[...]

PR #750, opened 2016-04-14
Fix connection termination issues when using connection filters (#737, #747).

cesarbs:
This PR addresses two issues that surface when using connection filters (namely our https connection filter):

#747) If the client closes the connection while we're still writing a response, we see exceptions in the log because we're trying to write to a diposed streams. Specifically with https, if a client sends an https request and closes the connection while we're still writing output, we get an exception from `SslStream.Write()` because the stream has already been disposed.

Not capturing that exception is undesired because we end up leaking memory blocks and leaving them up to be collected by the GC, while ideally we should properly return all of them to the memory pool.

This is not a completely unexpected situation, so the fix is just to log that it happened and prevent further attempts to write to the disposed stream.

#737) If a client takes too long to send a FIN and in the mean time we've already consumed all input and written a response, we get an exception when trying to close the connection because we still have code awaiting socket input. This is what is happening in this issue: for whatever reason nginx takes too long to send a FIN and we attempt to close the connection before receiving it. However, the task started by `FilteredStreamAdapter.ReadInput()` is still running and awaiting for input (which given enough time would be the FIN). Since we close the connection before receiving the FIN we see the reported stack trace because `Connection._rawSocketInput` is disposed and aborts any consumers awaiting on it.

The fix is to force a FIN into `Connection._rawSocketInput` an `Abort()` method in `FilteredStreamAdapter` that sets a flag that we check after the task that copies data from the wrapped stream to the outer `SocketInput` finishes. The previous call to `IncomingData()` on `_rawSocketInput` causes the `CopyToAsync` task to end gracefully and by calling `FilteredStreamAdapter.Abort()` we ensure the continuation for the `CopyToAsync` task will just attempt to abort awaiting consumers but not log any errors.

Another change is to return the continuation Task from `FilteredStreamAdapter.ReadInput()` so we can set another continuation on it to dispose other `Connection` members. By disposing those objects on that second continuation we ensure the disposal is really happening after the `CopyToAsync` task has finished AND any awaiters on `SocketInput` have been aborted.

cc @halter73 @yuwaMSFT @Eilon

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Kestrel/Filter/StreamSocketOutput.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/Filter/StreamSocketOutput.cs
@@ -16,33 +16,57 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Filter
         private static readonly byte[] _endChunkBytes = Encoding.ASCII.GetBytes("\r\n");
         private static readonly byte[] _nullBuffer = new byte[0];
 
+        private readonly string _connectionId;
         private readonly Stream _outputStream;
         private readonly MemoryPool _memory;
+        private readonly IKestrelTrace _logger;
         private MemoryPoolBlock _producingBlock;
+
+        private bool _canWrite = true;
 
         private object _writeLock = new object();
 
-        public StreamSocketOutput(Stream outputStream, MemoryPool memory)
+        public StreamSocketOutput(string connectionId, Stream outputStream, MemoryPool memory, IKestrelTrace logger)
         {
+            _connectionId = connectionId;
             _outputStream = outputStream;
             _memory = memory;
+            _logger = logger;
         }
 
         public void Write(ArraySegment<byte> buffer, bool chunk)
         {
             lock (_writeLock)
             {
-                if (chunk && buffer.Array != null)
+                if (buffer.Count == 0 )
                 {
-                    var beginChunkBytes = ChunkWriter.BeginChunkBytes(buffer.Count);
-                    _outputStream.Write(beginChunkBytes.Array, beginChunkBytes.Offset, beginChunkBytes.Count);
+                    return;
                 }
 
-                _outputStream.Write(buffer.Array ?? _nullBuffer, buffer.Offset, buffer.Count);
-
-                if (chunk && buffer.Array != null)
+                try
                 {
-                    _outputStream.Write(_endChunkBytes, 0, _endChunkBytes.Length);
+                    if (!_canWrite)
+                    {
+                        return;
+                    }
+
+                    if (chunk && buffer.Array != null)
+                    {
+                        var beginChunkBytes = ChunkWriter.BeginChunkBytes(buffer.Count);
+                        _outputStream.Write(beginChunkBytes.Array, beginChunkBytes.Offset, beginChunkBytes.Count);
+                    }
+
+                    _outputStream.Write(buffer.Array ?? _nullBuffer, buffer.Offset, buffer.Count);
+
+                    if (chunk && buffer.Array != null)
+                    {
+                        _outputStream.Write(_endChunkBytes, 0, _endChunkBytes.Length);
+                    }
+                }
+                catch (Exception ex)
+                {
+                    _canWrite = false;
+                    _logger.ConnectionError(_connectionId, ex);
                 }
             }
         }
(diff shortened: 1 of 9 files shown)


src/Microsoft.AspNetCore.Server.Kestrel/Filter/StreamSocketOutput.cs
```diff
@@ -65,14 +70,39 @@ public void ProducingComplete(MemoryPoolIterator end)
...
-                _outputStream.Write(block.Data.Array, block.Data.Offset, block.Data.Count);
+                // Writes can fail if the client has closed the connection but we're still sending a response.
+                // If we don't handle an exception from _outputStream.Write() here, we'll leak memory blocks.
+                if (_canWrite)
+                {
+                    try
+                    {
+                         _outputStream.Write(block.Data.Array, block.Data.Offset, block.Data.Count);
+                    }
+                    catch (Exception e)
```
Tratcher: This exception is lost

cesarbs:
A few more changes to come.

src/Microsoft.AspNetCore.Server.Kestrel/Filter/StreamSocketOutput.cs
yuwaMSFT: Just to confirm: do we want to suppress all types of exceptions and prevent future write?
The logged msg may not be accurate as the exceptions according to API could be:
        // Exceptions:
        //   T:System.ArgumentException:
        //     The sum of offset and count is greater than the buffer length.
        //
        //   T:System.ArgumentNullException:
        //     buffer is null.
        //
        //   T:System.ArgumentOutOfRangeException:
        //     offset or count is negative.
        //
        //   T:System.IO.IOException:
        //     An I/O error occured, such as the specified file cannot be found.
        //
        //   T:System.NotSupportedException:
        //     The stream does not support writing.
        //
        //   T:System.ObjectDisposedException:
        //     System.IO.Stream.Write(System.Byte[],System.Int32,System.Int32) was called after
        //     the stream was closed.

cesarbs:
And not only because I have to fix the build :stuck_out_tongue:

src/Microsoft.AspNetCore.Server.Kestrel/Filter/StreamSocketOutput.cs
```diff
@@ -65,14 +71,50 @@ public void ProducingComplete(MemoryPoolIterator end)
...
-                _outputStream.Write(block.Data.Array, block.Data.Offset, block.Data.Count);
+                // Writes can fail if the client has closed the connection but we're still sending a response.
+                // If we don't handle an exception from _outputStream.Write() here, we'll leak memory blocks.
+                if (_canWrite)
+                {
+                    try
+                    {
+                         _outputStream.Write(block.Data.Array, block.Data.Offset, block.Data.Count);
+                    }
+                    catch (ObjectDisposedException ex)
```
halter73: I wouldn't expect an ODE anymore, so we can do just the `ConnectionError` fix.

src/Microsoft.AspNetCore.Server.Kestrel/Http/Connection.cs
```diff
@@ -177,11 +178,9 @@ public virtual void OnSocketClosed()
         {
             _rawSocketInput.Dispose();
 
-            // If a connection filter was applied there will be two SocketInputs.
-            // If a connection filter failed, SocketInput will be null.
-            if (SocketInput != null && SocketInput != _rawSocketInput)
+            if (_filteredStreamAdapter != null)
             {
-                SocketInput.Dispose();
```
halter73: `?.`

src/Microsoft.AspNetCore.Server.Kestrel/Filter/StreamSocketOutput.cs
```diff
@@ -65,14 +71,50 @@ public void ProducingComplete(MemoryPoolIterator end)
...
+                }
 
                 var returnBlock = block;
                 block = block.Next;
                 returnBlock.Pool.Return(returnBlock);
             }
+            
+            if (_canWrite)
+            {
+                var count = end.Index - end.Block.Data.Offset;
```
halter73: Nit: this can be moved inside of the `try` once you only log using only `ConnectionError`

src/Microsoft.AspNetCore.Server.Kestrel/Filter/StreamSocketOutput.cs
```diff
@@ -65,14 +71,50 @@ public void ProducingComplete(MemoryPoolIterator end)
             var block = _producingBlock;
             while (block != end.Block)
             {
-                _outputStream.Write(block.Data.Array, block.Data.Offset, block.Data.Count);
+                // Writes can fail if the client has closed the connection but we're still sending a response.
```
halter73: This part of the comment can be removed as this should no longer be accurate.

src/Microsoft.AspNetCore.Server.Kestrel/Filter/StreamSocketOutput.cs
```diff
@@ -16,16 +16,22 @@ public class StreamSocketOutput : ISocketOutput
...
-        public StreamSocketOutput(Stream outputStream, MemoryPool memory)
+        public StreamSocketOutput(string connectionId, Stream outputStream, MemoryPool memory, IKestrelTrace logger)
         {
+            _connectionId = connectionId;
             _outputStream = outputStream;
             _memory = memory;
+            _logger = logger;
         }
 
         public void Write(ArraySegment<byte> buffer, bool chunk)
```
halter73: I think this method needs to be changed too. We should log and swallow exceptions from _outputStream.Write and we should ignore zero length writes.

While we're at it ChunkedResponseTests should be run with the NoOpConnectionFilter like the EngineTests are.

cesarbs: I guess you meant `PassThroughConnectionFilter` :grin:

cesarbs:
@Eilon @halter73 Updated first comment in this PR explaining the issues and fixes.

cesarbs:
Rebased.

test/Microsoft.AspNetCore.Server.KestrelTests/TestHelpers/TestApplicationErrorLogger.cs
```diff
@@ -32,12 +34,17 @@ public void Log<TState>(LogLevel logLevel, EventId eventId, TState state, Except
             Console.WriteLine($"Log {logLevel}[{eventId}]: {formatter(state, exception)} {exception?.Message}");
 #endif
 
-            TotalErrorsLogged++;
+            TotalMessagesLogged++;
```
halter73: Just remove this property, and use the now correct `TotalErrorsLogged` in the test that was using it.

halter73:
:shipit:

Issue #755, opened 2016-04-18
Calling UseKestrel() (no args) then calling Configure<KestrelServerOptions> ends up with null service provider


davidfowl:
The following won't work if UseKestrel doesn't use the `Action<KestrelServerOptions>` overload. We need to register the default 

```C#
builder.UseKestrel()
```

```C#
services.Configure<KestrelServerOptions>(o =>
{
    System.Console.WriteLine(o.ApplicationServices == null);
});
```

This line https://github.com/aspnet/KestrelHttpServer/blob/dev/src/Microsoft.AspNetCore.Server.Kestrel/WebHostBuilderKestrelExtensions.cs#L45 needs to be moved to `UseKestrel()`

/cc @mikeharder @Tratcher

PR #757, opened 2016-04-18
Add KestrelServerOptionsSetup before KestrelServerOptions (#755)

mikeharder:
- Required to ensure that options.ApplicationServices is available after during UseKestrel(options)

@davidfowl, @halter73, @Tratcher, @cesarbs


davidfowl:
:shipit:

Issue #758, opened 2016-04-19
Random Port is not updated


brunolauze:
**dotnet run -c Release -- --server.urls=http://*:0**

In the context where we want to fire kestrel with a random port ; where consul or others will receive registration and update load balancing according to new process/port.

the command above is working, and we know Tcp Listener created with port 0 will get a random usable port from the kernel on any operating system. Win32, Linux and BSD

But still Kestrel doesn't check the assigned port and lies in the output.

PS D:\dev\n3.acronyms\ui\web\home> dotnet run -c Release -- --server.urls=http://*:0
Project N3.Acronyms.UI.Web.Home (.NETCoreApp,Version=v1.0) was previously compiled. Skipping compilation.
dbug: Microsoft.AspNetCore.Hosting.Internal.WebHost[3]
      Hosting starting
dbug: Microsoft.AspNetCore.Hosting.Internal.WebHost[4]
      Hosting started
Hosting environment: Production
Content root path: D:\dev\n3.acronyms\ui\web\home

**Now listening on: http://*:0**

Application started. Press Ctrl+C to shut down.
Application started. Press Ctrl+C to shut down.

Where we should have seen the new port assigned:

**Now listening on: http://*:19246**

I am asking either to fix this error or report a correct way to assign free random port in Kestrel.

ayende:
+1 o that, this is something that we would need for running tests

PR #760, opened 2016-04-20
More robust port assignment for tests.

cesarbs:

cc @pakrym for `project.json` review

cc @Eilon need to get into release to unblock CI

### DIFF
--- a/test/Microsoft.AspNetCore.Server.Kestrel.TestCommon/PortManager.cs
+++ b/test/Microsoft.AspNetCore.Server.Kestrel.TestCommon/PortManager.cs
@@ -1,0 +1,38 @@
+using System.Net;
+using System.Net.Sockets;
+
+namespace Microsoft.AspNetCore.Server.Kestrel.TestCommon
+{
+    public static class PortManager
+    {
+        public static int _nextPort = 8001;
+        public static object _portLock = new object();
+        
+        public static int GetNextPort()
+        {
+            lock (_portLock)
+            {
+                using (var socket = new Socket(AddressFamily.InterNetwork, SocketType.Stream, ProtocolType.Tcp))
+                {
+                    while (true)
+                    {
+                        try
+                        {
+                            var port = _nextPort++;
+                            socket.Bind(new IPEndPoint(IPAddress.Loopback, port));
+                            return port;
+                        }
+                        catch (SocketException)
+                        {
+                            // Retry unless exhausted
+                            if (_nextPort == 65536)
+                            {
+                                throw;
+                            }
+                        }
+                    }
+                }
+            }
+        }
+    }
+}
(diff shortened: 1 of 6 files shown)


pakrym:
We have https://github.com/aspnet/Hosting/blob/dev/src/Microsoft.AspNetCore.Server.Testing/Common/TestUriHelper.cs can it be reused?

test/Microsoft.AspNetCore.Server.Kestrel.TestCommon/PortManager.cs
```diff
@@ -0,0 +1,20 @@
+using Microsoft.AspNetCore.Server.Testing.Common;
+
+namespace Microsoft.AspNetCore.Server.Kestrel.TestCommon
+{
+    public static class PortManager
+    {
+        public static int _nextPort = 8001;
+        public static object _portLock = new object();
+        
+        public static int GetNextPort()
```
mikeharder: Maybe keep the old name GetPort(), to avoid changing all the callers of this API?  I don't think GetNextPort() is a particularly better name.

cesarbs:
@pakrym I don't think so. There's a delay between finding a free port and a test binding to it. Multiple tests might end up finding the same free port but it will blow up by the time they actually try to bind to it.

test/Microsoft.AspNetCore.Server.Kestrel.TestCommon/PortManager.cs
cesarbs: Just realized there's a problem with this logic. `TestUriHelper.FindFreePort` won't search from `_nextPort` and above if the initial port is not available. This defeats the purpose of this class.

Will submit an update soon.

pakrym:
I just don't like same functionality in multiple places, can you fix `Hosting`  implementation not to have this problem and use it?

mikeharder:
I see this is needed immediately to unblock CI.

In parallel I am working on #758 for post-RC2.  If we can make it work, we should change tests to tell Kestrel to bind to port 0, then let the OS choose the port and expose it from Kestrel.

cesarbs:
Yeah, I actually wanted to just change the port range to unblock the CI ASAP. Anyone give me a go on this? I can work on the more robust solution next, but actually I wonder if it's worth it since @mikeharder is working on the port 0 thing as he said.

mikeharder:
If we want a minimal fix to unblock the CI ASAP, I would just fix PortManager.GetPort() and TestServer.GetNextPort() to check for free ports, and not change any of the calling code.

halter73:
:shipit:

mikeharder:
:shipit:

mikeharder:
@cesarbs: We need to add an xproj for the new TestCommon folder.  Otherwise every time you open KestrelHttpServer.sln it creates the xproj and wants you to check it in.

mikeharder:
Also need to add xproj to KestrelHttpServer.sln.

cesarbs:
:+1:

mikeharder:
@cesarbs: I also noticed that TestCommon\PortManager.cs is missing the copyright header.

PR #761, opened 2016-04-21
Update addresses in IServerAddressesFeature with assigned dynamic ports

mikeharder:
Addresses #758.

Making ServerAddress.Port a settable property was the easiest way to implement this.  If we want to keep ServerAddress immutable, we can find another way to flow the assigned port from TcpListenerPrimary to KestrelServer.

@halter73, @cesarbs, @Tratcher, @davidfowl


src/Microsoft.AspNetCore.Server.Kestrel/KestrelServer.cs
```diff
@@ -96,7 +97,7 @@ public void Start<TContext>(IHttpApplication<TContext> application)
                 engine.Start(threadCount);
                 var atLeastOneListener = false;
 
-                foreach (var address in _serverAddresses.Addresses)
+                foreach (var address in _serverAddresses.Addresses.ToArray())
```
cesarbs: Why?

cesarbs:
Have you tried using port 0 in the tests?

src/Microsoft.AspNetCore.Server.Kestrel/ServerAddress.cs
```diff
@@ -13,7 +13,7 @@ public class ServerAddress
     {
         public string Host { get; private set; }
         public string PathBase { get; private set; }
-        public int Port { get; private set; }
+        public int Port { get; set; }
```
brunolauze: IHMO, internal would probably be better here

src/Microsoft.AspNetCore.Server.Kestrel/KestrelServer.cs
Tratcher: Because the original collection is being modified down below.

src/Microsoft.AspNetCore.Server.Kestrel/KestrelServer.cs
```diff
@@ -108,6 +109,10 @@ public void Start<TContext>(IHttpApplication<TContext> application)
                         atLeastOneListener = true;
                         _disposables.Push(engine.CreateServer(
                             parsedAddress));
+
+                        // If requested port was "0", replace with assigned dynamic port.
+                        _serverAddresses.Addresses.Remove(address);
+                        _serverAddresses.Addresses.Add(parsedAddress.ToString());
```
Tratcher: If I register http://localhost:0/, will I get back http://127.0.01:12345/?

mikeharder: No.  I updated samples\SampleApp to call

```
.UseUrls("http://localhost:0", "https://localhost:0")
```

and it prints the following:

```
Now listening on: http://localhost:34934
Now listening on: https://localhost:34935
```

cesarbs: Have you tested with

* hostname
* IPv4 address
* IPv6 address

Wanna make sure this covers all bases (which it should, but need to see it working).

cesarbs: Also make sure your changes don't impact listening on a unix socket (shouldn't but again, needs testing).

src/Microsoft.AspNetCore.Server.Kestrel/ServerAddress.cs
cesarbs: Agreed.

mikeharder:
Updated tests to use dynamic ports.  All tests pass on Windows, still need to test Linux and Mac.

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/AddressRegistrationTests.cs
```diff
@@ -40,58 +56,110 @@ public class AddressRegistrationTests
...
+                        (sender, cert, chain, sslPolicyErrors) => true;
+
+                try
                 {
-                    foreach (var testUrl in testUrls)
+#if NET451
+                    var handler = new HttpClientHandler();
+                    ServicePointManager.ServerCertificateValidationCallback += validationCallback;
+#else
+                    var handler = new WinHttpHandler();
```
halter73: We shouldn't do this for non-HTTPS addreses. This fails on Linux and OS X (see the Travis runs), so we need to use OSSkipConditions like we do for CanReadAndWriteWithHttpsConnectionFilter to use WinHttpHandler.

I think we should split the RegisterAddresses* Theories into HTTP/HTTPS versions so we don't need to skip more tests than necessary.

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/AddressRegistrationTests.cs
```diff
@@ -40,58 +56,110 @@ public class AddressRegistrationTests
...
-                var dataset = new TheoryData<string, string[]>();
-                dataset.Add($"{port1}", new[] { $"http://localhost:{port1}/" });
-                dataset.Add($"{port1};{port2}", new[] { $"http://localhost:{port1}/", $"http://localhost:{port2}/" });
-                dataset.Add($"http://127.0.0.1:{port1}/", new[] { $"http://127.0.0.1:{port1}/", });
-                dataset.Add($"http://localhost:{port1}/base/path", new[] { $"http://localhost:{port1}/base/path" });
+                var dataset = new TheoryData<string, Func<IServerAddressesFeature, string[]>>();
+
+                // Default host and port
+                dataset.Add(null, _ => new[] { "http://localhost:5000/" });
+                dataset.Add(string.Empty, _ => new[] { "http://localhost:5000/" });
```
halter73: It looks dangerous to have to different pieces of theory data ultimately binding to the same port. I don't think xunit will run tests from the same class in parallel, but I'm not a 100% sure.

Maybe this is a good time [disable parallelism](https://github.com/aspnet/ServerTests/blob/0f5fc8122b72eaf262dd5702e4821e6e9419069e/test/ServerComparison.FunctionalTests/Properties/AssemblyInfo.cs) in our tests. They run pretty quickly anyway.

muratg: Or port number can also be passed in as data, no?

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/project.json
```diff
@@ -36,5 +38,8 @@
   "compilationOptions": {
     "allowUnsafe": true
   },
-  "testRunner": "xunit"
+  "testRunner": "xunit",
+  "content": [
+    "TestResources/testCert.pfx"
```
muratg: Will this "expire" by any chance? Or is it good for foreseeable future? :)

test/Microsoft.AspNetCore.Server.KestrelTests/TestServer.cs
```diff
@@ -58,10 +58,5 @@ public void Dispose()
             _server.Dispose();
             _engine.Dispose();
         }
-
-        public static int GetNextPort()
-        {
-            return PortManager.GetNextPort();
-        }
```
muratg: :+1: :100:

muratg:
:watch:

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/AddressRegistrationTests.cs
mikeharder: ServerCertificateValidationCallback was recently added to CurlHandler (see https://github.com/dotnet/corefx/pull/7060).    I will attempt to modify our tests to use CurlHandler on Linux and Mac OS.

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/AddressRegistrationTests.cs
mikeharder: I don't think this is related to my changes, since multiple theories where using the same port before my changes.  By default, tests in the same class are not run in parallel by XUnit, so these tests should be fine.

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/project.json
mikeharder: I copied the existing certificate file from Microsoft.AspNetCore.Server.KestrelTests.  Our tests already disable certificate validation since it's self-signed, and I suspect this would disable expiration validation as well.

halter73: It expires 9/15/2016 at 11:21PM. Our test `validationCallback`s always return true, so it might not matter. The test is already issued by "localhost".

muratg: Sounds good.

muratg:
Looks good to me, though I'll let @cesarbs or @halter73 give the squirrel :)

cesarbs:
LGTM. One question though: the remaining tests that still need the `GetNextPort()` stuff, do they reside in only one of the test projects? If so, I would move that logic there and remove the TestCommon project I begrudgingly introduced a few days ago.

muratg:
:shipit: (Not sure if you'd like to consider moving the GetNextPort suggestion above in this PR or another one.)

mikeharder:
I'm moving GetNextPort() into `FunctionalTests\AddressRegistrationTests.cs` since it's the only class that needs it.  I'm also deleting the TestCommon project, and copying the helper extension method `IWebHost.GetPort()` into both `FunctionalTests` and `KestrelTests`, since I don't think its worth sharing a common project for a single extension method.

I also need to add `CurlHandler` to our HTTPS tests so they can pass on Mac and Linux.

mikeharder:
I have removed the HTTPS tests from `FunctionalTests\AddressRegistrationTests.cs` for the time being.  To make these work xplat, we need the HttpClientHandler.ServerCertificateValidationCallback property, which was added in RC3 builds of CoreFx (https://github.com/dotnet/corefx/pull/7060).

src/Microsoft.AspNetCore.Server.Kestrel/KestrelServer.cs
mikeharder: I added tests for hostname, IPv4 address, and IPv6 address to `FunctionalTests\AddressRegistrationTests.cs`.  The test currently fails on the IPv6 link-local address which ends with `%18`.  The value returned by context.Request.GetDisplayUrl() is missing the `%18`.  I still need to determine whether this is a test issue or a product issue.

cesarbs:
:shipit: After Travis passes.

mikeharder:
@halter73: Please review my latest commit which impacts unix pipes (12219733ddc1f6cb3b2892268a9617516cd25852).

I'm pretty sure ToString() should exclude the `:Port` for unix pipes.

I'm less certain about whether ToString() should include `:PathBase` if PathBase is empty, but non-unix-pipe addresses only include `/PathBase` if it's not empty, so I did the same.

mikeharder:
@cesarbs: Travis is failing because I added tests which attempt to bind to port 80.  I'll move these to a separate test method, with a condition that verifies whether port 80 is free.

cesarbs:
Needs rebase.

mikeharder:
Rebased to latest.  I'm still investigating the IPv6 with zone index failure on Mac.

mikeharder:
@halter73: AppVeyor and Travis are finally passing.  Can you please review the commit which impacts unix pipes, as described above?

muratg:
Latest commit looks good to me.

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/IWebHostPortExtensions.cs
```diff
@@ -0,0 +1,25 @@
...
+    {
+        public static int GetPort(this IWebHost host)
+        {
+            return host.GetPorts().First();
+        }
+
+        public static IEnumerable<int> GetPorts(this IWebHost host)
+        {
+            return host.ServerFeatures.Get<IServerAddressesFeature>().Addresses
+                .Select(a => a.StartsWith("http://+") ? a.Replace("http://+", "http://localhost") : a)
```
khellang: Might be a stupid question, but what about the scheme here? Can't you have addresses with `https://+`?

mikeharder: Good catch.  We currently don't have any tests which use `https://+`, but I've improved this code to handle both cases for future tests.

mikeharder: Issue for adding HTTPS tests: https://github.com/aspnet/KestrelHttpServer/issues/794

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/AddressRegistrationTests.cs
```diff
@@ -16,20 +21,36 @@ namespace Microsoft.AspNetCore.Server.Kestrel.FunctionalTests
...
-        public async Task RegisterAddresses_IPv6_Success(string addressInput, string[] testUrls)
+        public async Task RegisterAddresses_IPv6_Success(string addressInput, Func<IServerAddressesFeature, string[]> testUrls)
         {
             await RegisterAddresses_Success(addressInput, testUrls);
         }
 
-        public async Task RegisterAddresses_Success(string addressInput, string[] testUrls)
+        [ConditionalTheory, MemberData(nameof(AddressRegistrationDataIPv6ScopeId))]
+        [OSSkipCondition(OperatingSystems.Linux, SkipReason = "HttpClient does not support IPv6 with scope ID on Linux.")]
+        [OSSkipCondition(OperatingSystems.MacOSX, SkipReason = "HttpClient does not support IPv6 with scope ID on Mac.")]
```
cesarbs: add link to issue in skip message

cesarbs:
Beautiful. I've left a minor comment; address it then :ship: :rocket:

cesarbs:
:shipit:

mikeharder:
Merged to dev in 14bbba06418234f1f6d41eb2540ca26f8598fd23

Issue #753, opened 2016-04-15
Remove SocketInputStream


halter73:
Now that we implement `CopyToAsync` ourselves (which takes `SocketInputStream` as the destination as well as a `MemoryPoolBlock` as an intermediate buffer, it would be simpler to have `CopyToAsync` write to `SocketInput` directly.

PR #762, opened 2016-04-21
Remove SocketInputStream (#753).

cesarbs:
#753

cc @halter73 @mikeharder


src/Microsoft.AspNetCore.Server.Kestrel/Http/Connection.cs
```diff
@@ -31,7 +31,7 @@ public class Connection : ConnectionContext, IConnectionControl
         private ConnectionFilterContext _filterContext;
         private LibuvStream _libuvStream;
         private FilteredStreamAdapter _filteredStreamAdapter;
-        private Task _readInputContinuation;
+        private Task _readInputTask;
```
muratg: Why rename?  Just curious...

cesarbs: Having "continuation" is the name is needlessly leaking information about what's happening inside `ReadInput()`.

cesarbs: A call to `ContinueWith()` that is.

halter73:
Makes you wonder why it wasn't written this way in the first place. I promise there was a reason!

Basically before SslStream was really working, @lodejard and I were hoping that SslStream would smartly override CopyToAsync to call SocketInputStream with its own byte buffers to avoid the double buffer problem we now have. (Notice we now have to use a MemoryPoolBlock in FilterInputAsync to copy data from a byte buffer SslStream already allocated).

Those supposed benefits are completely worthless with the default Stream implementation of CopyToAsync though.

:shipit:

Issue #649, opened 2016-02-22
ECONNRESET constant should be platform dependent


jak137:
We observed ``IOException`` errors from Kestrel server (1.0.0-rc1 with keep_alive patch from #418) running on Ubuntu 14.04 Linux platform when it was responding to health check requests from HAProxy:

```
option httpchk GET /ping HTTP/1.1\r\nHost:\ batalion\r\nConnection:\ close
```

```
2016-02-22 19:52:36.1378|INFO|Microsoft.AspNet.Hosting.Internal.HostingEngine|Request starting HTTP/1.1 GET http://batalion/ping  
2016-02-22 19:52:36.1378|INFO|NGmu.Startup|Connection: close
2016-02-22 19:52:36.1378|INFO|NGmu.Startup|Host: batalion
2016-02-22 19:52:36.1378|INFO|Microsoft.AspNet.Hosting.Internal.HostingEngine|Request finished in 0ms 200 
2016-02-22 19:52:36.1382|WARN|Microsoft.AspNet.Server.Kestrel|Connection processing ended abnormally
System.IO.IOException: Error -104 ECONNRESET connection reset by peer ---> Microsoft.AspNet.Server.Kestrel.Networking.UvException: Error -104 ECONNRESET connection reset by peer
  --- End of inner exception stack trace ---
  at Microsoft.AspNet.Server.Kestrel.Http.SocketInput.GetResult () <0x42822ea0 + 0x000a3> in <filename unknown>:0 
  at Microsoft.AspNet.Server.Kestrel.Http.SocketInputExtensions+<ReadAsync>d__0.MoveNext () <0x4281ff40 + 0x00147> in <filename unknown>:0 
--- End of stack trace from previous location where exception was thrown ---
  at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw () <0x7fc130001860 + 0x00029> in <filename unknown>:0 
  at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess (System.Threading.Tasks.Task task) <0x7fc12ffff840 + 0x000a7> in <filename unknown>:0 
  at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification (System.Threading.Tasks.Task task) <0x7fc12ffff7c0 + 0x0006b> in <filename unknown>:0 
[...]

PR #764, opened 2016-04-21
Make ECONNRESET value platform-specific (#649).

cesarbs:
Gross, but we don't have platform-specific `#define`s.

#649 

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Kestrel/Infrastructure/Constants.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/Infrastructure/Constants.cs
@@ -1,6 +1,9 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System;
+using Microsoft.Extensions.PlatformAbstractions;
+
 namespace Microsoft.AspNetCore.Server.Kestrel.Infrastructure
 {
     internal class Constants
@@ -8,7 +11,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Infrastructure
         public const int ListenBacklog = 128;
 
         public const int EOF = -4095;
-        public const int ECONNRESET = -4077;
+        public static readonly int? ECONNRESET = GetECONNRESET();
 
         /// <summary>
         /// Prefix of host name used to specify Unix sockets in the configuration.
@@ -20,5 +23,20 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Infrastructure
         /// for info on the format.
         /// </summary>
         public const string RFC1123DateFormat = "r";
+
+        private static int? GetECONNRESET()
+        {
+            switch (PlatformServices.Default.Runtime.OperatingSystemPlatform)
+            {
+                case Platform.Windows:
+                    return -4077;
+                case Platform.Linux:
+                    return -104;
+                case Platform.Darwin:
+                    return -54;
+                default:
+                    return null;
+            }
+        }
     }
 }
--- a/src/Microsoft.AspNetCore.Server.Kestrel/Http/Connection.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/Http/Connection.cs
@@ -40,6 +40,8 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Http
         private ConnectionState _connectionState;
         private TaskCompletionSource<object> _socketClosedTcs;
 
+        bool _eConnResetChecked = false;
+
         public Connection(ListenerContext context, UvStreamHandle socket) : base(context)
         {
             _socket = socket;
@@ -270,6 +272,12 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Http
                 // We need to clean up whatever was allocated by OnAlloc.
                 _rawSocketInput.IncomingDeferred();
                 return;
+            }
+
+            if (!_eConnResetChecked && !Constants.ECONNRESET.HasValue)
+            {
+                Log.LogWarning("Unable to determine ECONNRESET value on this platform.");
+                _eConnResetChecked = true;
             }
 
             var normalRead = status > 0;


src/Microsoft.AspNetCore.Server.Kestrel/Infrastructure/Constants.cs
```diff
@@ -1,14 +1,33 @@
...
+
 namespace Microsoft.AspNetCore.Server.Kestrel.Infrastructure
 {
     internal class Constants
     {
         public const int ListenBacklog = 128;
 
         public const int EOF = -4095;
-        public const int ECONNRESET = -4077;
+        public static int ECONNRESET
```
mikeharder: Fetching the value would be slightly faster if you make the field a `public static readonly int` and initialize it once:

```C#
public static readonly int ECONNRESET = GetECONNRESET();

private static int GetECONNRESET() {
    switch (PlatformServices.Default.Runtime.OperatingSystemPlatform)
}
```

src/Microsoft.AspNetCore.Server.Kestrel/Infrastructure/Constants.cs
```diff
@@ -1,14 +1,33 @@
...
+                switch (PlatformServices.Default.Runtime.OperatingSystemPlatform)
+                {
+                    case Platform.Windows:
+                        return -4077;
+                    case Platform.Linux:
+                        return -104;
+                    case Platform.Darwin:
+                        return -54;
+                    default:
+                        throw new Exception("Unable to determine ECONNRESET value on this platform.");
```
smbecker: Is it wise to throw here? There is absolutely nothing that a user can do about it.

mikeharder:
Unless we add a lot more platform-specific code, I doubt we would ever have platform-specific #defines, since this means we would need to build and ship a DLL for each platform.

cesarbs:
@mikeharder Yep. Alternatives to my approach would be to code gen the constant or move the platform check logic to the place where `ECONNRESET` is used. I like this one better because it keeps everything in one place, but I might be convinced otherwise.

cesarbs:
@smbecker I don't expect anyone to hit the default case in the switch statement, since those are our currently supported platforms. If it is hit though, it's hard to decide what to do other throw: I could return a know value like `EOF`, but then the code would be lying. I'd rather become aware that there's a platform we need to pay attention to and set `ECONNRESET` appropriately for it.

mikeharder:
:shipit:

src/Microsoft.AspNetCore.Server.Kestrel/Infrastructure/Constants.cs
```diff
@@ -20,5 +23,20 @@ internal class Constants
...
+            switch (PlatformServices.Default.Runtime.OperatingSystemPlatform)
+            {
+                case Platform.Windows:
+                    return -4077;
+                case Platform.Linux:
+                    return -104;
+                case Platform.Darwin:
+                    return -54;
+                default:
+                    throw new Exception("Unable to determine ECONNRESET value on this platform.");
```
davidfowl: PlatformNotSupportedException?

halter73: I'm not sure this should be throwing at all. Maybe we just log that we couldn't determine the ECONNRESET constants.

This really only affects the noisiness of the logs, and it is possible that that someone could bring their own libuv.

davidfowl: True. Should we have a default return value?

smbecker: Do you just default it to the Linux constant? More than likely, if it hits that default block, it is some variant of Linux that isn't being detected properly.

cesarbs:
Made `ECONNRESET` nullable per @davidfowl's suggestion. Logging one per connection when the value cannot be determined on a given platform.

src/Microsoft.AspNetCore.Server.Kestrel/Infrastructure/Constants.cs
```diff
@@ -1,14 +1,17 @@
...
+
 namespace Microsoft.AspNetCore.Server.Kestrel.Infrastructure
 {
     internal class Constants
     {
         public const int ListenBacklog = 128;
 
         public const int EOF = -4095;
-        public const int ECONNRESET = -4077;
+        public static readonly int? ECONNRESET = GetECONNRESET();
```
mikeharder: This value is read on every call to Connection.OnRead().  Would performance be better if we set this to a sentinel value like int.MinValue if we can't determine the platform?  The perf overhead of nullable int may be small, but I hate to pay an unnecessary cost on every call to OnRead().

halter73: Probably should measure with your fancy new BenchmarksDriver :smile:

davidfowl: I'd be surprised if this cost anything.

mikeharder: I ran a micro-benchmark using BenchmarkDotNet:

```
                        Method |    Median |    StdDev |
------------------------------ |---------- |---------- |
                  CompareConst | 0.0005 ns | 0.0110 ns |
         CompareStaticReadonly | 0.0000 ns | 0.0000 ns |
 CompareStaticReadonlyNullable | 0.2871 ns | 0.0015 ns |
```

Nullable is slower, but it might not matter in a real-world benchmark like Plaintext.

cesarbs: Just ran some plaintext benchmarks. No difference.

halter73: :shipit:

PR #771, opened 2016-04-26
Minor Stream improvements

justinvp:
- Unsupported members should throw `NotSupportedException` instead of
`NotImplementedException` per MSDN docs for [`CanRead`](https://msdn.microsoft.com/en-us/library/system.io.stream.canread(v=vs.110).aspx)/[`CanSeek`](https://msdn.microsoft.com/en-us/library/system.io.stream.canseek(v=vs.110).aspx)/[`CanWrite`](https://msdn.microsoft.com/en-us/library/system.io.stream.canwrite(v=vs.110).aspx).

- Position should throw `NotSupportedException` when `CanSeek` is false.

- `FrameRequestStream.Flush` and `FlushAsync` should not throw
`NotImplementedException`.

- Use expression-bodied members for `CanRead`/`CanSeek`/`CanWrite` on
`FrameRequestStream` to match `FrameResponseStream`.

- Provide no-op override of `LibuvStream.FlushAsync` to match `Flush`.

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Kestrel/Http/FrameRequestStream.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/Http/FrameRequestStream.cs
@@ -19,35 +19,51 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Http
             _state = FrameStreamState.Closed;
         }
 
-        public override bool CanRead { get { return true; } }
+        public override bool CanRead => true;
 
-        public override bool CanSeek { get { return false; } }
+        public override bool CanSeek => false;
 
-        public override bool CanWrite { get { return false; } }
+        public override bool CanWrite => false;
 
         public override long Length
         {
             get
             {
-                throw new NotImplementedException();
+                throw new NotSupportedException();
             }
         }
 
-        public override long Position { get; set; }
+        public override long Position
+        {
+            get
+            {
+                throw new NotSupportedException();
+            }
+            set
+            {
+                throw new NotSupportedException();
+            }
+        }
 
         public override void Flush()
         {
-            throw new NotImplementedException();
+            // No-op.
+        }
+
+        public override Task FlushAsync(CancellationToken cancellationToken)
+        {
+            // No-op.
+            return TaskUtilities.CompletedTask;
         }
 
         public override long Seek(long offset, SeekOrigin origin)
         {
-            throw new NotImplementedException();
+            throw new NotSupportedException();
         }
 
         public override void SetLength(long value)
         {
-            throw new NotImplementedException();
+            throw new NotSupportedException();
         }
 
         public override int Read(byte[] buffer, int offset, int count)
@@ -113,7 +129,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Http
 
         public override void Write(byte[] buffer, int offset, int count)
         {
-            throw new NotImplementedException();
+            throw new NotSupportedException();
         }
 
         public Stream StartAcceptingReads(MessageBody body)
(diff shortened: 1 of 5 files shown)


test/Microsoft.AspNetCore.Server.KestrelTests/FrameRequestStreamTests.cs
```diff
@@ -0,0 +1,54 @@
...
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Server.Kestrel.Http;
+using Xunit;
+
+namespace Microsoft.AspNetCore.Server.KestrelTests
+{
+    public class FrameRequestStreamTests
+    {
+        [Fact]
+        public void CanPropertiesReturnExpectedValues()
```
cesarbs: Break this up into 3 tests, one for each property.

test/Microsoft.AspNetCore.Server.KestrelTests/FrameRequestStreamTests.cs
```diff
@@ -0,0 +1,54 @@
...
+            var stream = new FrameRequestStream();
+
+            Assert.True(stream.CanRead);
+            Assert.False(stream.CanSeek);
+            Assert.False(stream.CanWrite);
+        }
+
+        [Fact]
+        public async Task UnsupportedMembersThrow()
+        {
```
cesarbs: Same here.

test/Microsoft.AspNetCore.Server.KestrelTests/FrameRequestStreamTests.cs
```diff
@@ -0,0 +1,54 @@
...
+            Assert.Throws<NotSupportedException>(() => stream.BeginWrite(new byte[1], 0, 1, null, null));
+#endif
+        }
+
+        [Fact]
+        public async Task FlushDoesNotThrow()
+        {
+            var stream = new FrameRequestStream();
+
+            stream.Flush();
```
cesarbs: Use `Assert.DoesNotThrow<>`.

test/Microsoft.AspNetCore.Server.KestrelTests/FrameResponseStreamTests.cs
```diff
@@ -0,0 +1,45 @@
...
+
+using System;
+using System.IO;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Server.Kestrel.Http;
+using Xunit;
+
+namespace Microsoft.AspNetCore.Server.KestrelTests
+{
+    public class FrameResponseStreamTests
```
cesarbs: Break up the tests in this class too.

cesarbs:
LGTM, but please address feedback on unit tests.

test/Microsoft.AspNetCore.Server.KestrelTests/FrameRequestStreamTests.cs
justinvp: xUnit 2.0 does not have `Assert.DoesNotThrow()`.

test/Microsoft.AspNetCore.Server.KestrelTests/FrameResponseStreamTests.cs
justinvp: Done.

justinvp:
Thanks for the review. I broke up the unit tests.

src/Microsoft.AspNetCore.Server.Kestrel/Http/FrameRequestStream.cs
```diff
@@ -19,35 +19,51 @@ public FrameRequestStream()
             _state = FrameStreamState.Closed;
         }
 
-        public override bool CanRead { get { return true; } }
+        public override bool CanRead => true;
 
-        public override bool CanSeek { get { return false; } }
+        public override bool CanSeek => false;
```
davidfowl: super nit: These would probably be better as:

```C#
public override bool CanSeek { get; } = false;
 
```

justinvp: Why? Your suggestion adds a private backing field, which increases the cost in memory. The expression-bodied member just returns false, without adding a private field to the type.

davidfowl: Yea, I guess it does add a field.

cesarbs:
Sweet :shipit:

PR #772, opened 2016-04-26
Use Task.FromCanceled<TResult>() on NETSTANDARD1_3

justinvp:
`Task.FromCanceled<int>(cancellationToken)` can be used on `NETSTANDARD1_3` in `TaskUtilities.GetCancelledZeroTask()`.

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Kestrel/Http/FrameRequestStream.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/Http/FrameRequestStream.cs
@@ -165,7 +165,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Http
                 case FrameStreamState.Open:
                     if (cancellationToken.IsCancellationRequested)
                     {
-                        return TaskUtilities.GetCancelledZeroTask();
+                        return TaskUtilities.GetCancelledZeroTask(cancellationToken);
                     }
                     break;
                 case FrameStreamState.Closed:
--- a/src/Microsoft.AspNetCore.Server.Kestrel/Infrastructure/TaskUtilities.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/Infrastructure/TaskUtilities.cs
@@ -26,12 +26,15 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Infrastructure
 #endif
         }
 
-        public static Task<int> GetCancelledZeroTask()
+        public static Task<int> GetCancelledZeroTask(CancellationToken cancellationToken = default(CancellationToken))
         {
-            // Task<int>.FromCanceled doesn't return Task<int>
+#if NETSTANDARD1_3
+            return Task.FromCanceled<int>(cancellationToken);
+#else
             var tcs = new TaskCompletionSource<int>();
             tcs.TrySetCanceled();
             return tcs.Task;
+#endif
         }
     }
 }


cesarbs:
Not sure what the benefit of this is. I'm inclined not to take it since it only adds code with no clear improvement.

justinvp:
Benefits:
- Avoids an unnecessary `TaskCompletionSource<int>` allocation on Net Standard.
- The implementation of `GetCancelledZeroTask()` now matches `GetCancelledTask()` (the method above it).
- This comment: `// Task<int>.FromCanceled doesn't return Task<int>` is misguided/confusing. There is a method that returns `Task<int>` from a cancellation token: `Task.FromCanceled<int>()`

justinvp:
Also, `FrameRequestStream.ValidateState` now matches `FrameResponseStream`, passing along the cancellation token to the canceled task when `_state` is `FrameStreamState.Open` and the token has been canceled.

src/Microsoft.AspNetCore.Server.Kestrel/Http/FrameRequestStream.cs
```diff
@@ -165,13 +165,13 @@ private Task<int> ValidateState(CancellationToken cancellationToken)
...
                     {
-                        return TaskUtilities.GetCancelledZeroTask();
+                        return TaskUtilities.GetCancelledZeroTask(cancellationToken);
                     }
                     break;
                 case FrameStreamState.Closed:
                     throw new ObjectDisposedException(nameof(FrameRequestStream));
                 case FrameStreamState.Aborted:
-                    return TaskUtilities.GetCancelledZeroTask();
+                    return TaskUtilities.GetCancelledZeroTask(default(CancellationToken));
```
justinvp: Unrelated to this change, but I noticed that `FrameResponseStream.ValidateState` only returns a canceled task for the `Aborted` case when `cancellationToken.IsCancellationRequested` is true, whereas here it is always returning a canceled task for `Aborted` regardless of `cancellationToken.IsCancellationRequested`.

`FrameResponseStream.ValidateState` ([src](https://github.com/aspnet/KestrelHttpServer/blob/b0d5cfcf47a8de83afe930056c87f193e4f955c6/src/Microsoft.AspNetCore.Server.Kestrel/Http/FrameResponseStream.cs#L160-L166)):

```csharp
                    if (cancellationToken.IsCancellationRequested)
                    {
                        // Aborted state only throws on write if cancellationToken requests it
                        return TaskUtilities.GetCancelledTask(cancellationToken);
                    }
                    break;
```

Is this by design, or should `FrameRequestStream` be modified to behave the same as `FrameResponseStream`?

benaadams: Is by design https://github.com/aspnet/KestrelHttpServer/pull/567#discussion_r50049886 and the reading and writing are different.

mikeharder:
@cesarbs, @halter73: I do think we should merge this for the reasons listed above by @justinvp.

src/Microsoft.AspNetCore.Server.Kestrel/Infrastructure/TaskUtilities.cs
```diff
@@ -26,12 +26,15 @@ public static Task GetCancelledTask(CancellationToken cancellationToken)
 #endif
         }
 
-        public static Task<int> GetCancelledZeroTask()
+        public static Task<int> GetCancelledZeroTask(CancellationToken cancellationToken)
```
mikeharder: I would add an overload `GetCancelledZeroTask()` which calls `GetCancelledZeroTask(CancellationToken.None)`.  This is cleaner than requiring the caller to pass `default(CancellationToken)` or `CancellationToken.None`.

justinvp: How about making it an optional parameter with a default value of `default(CancellationToken)`?

mikeharder: Sure, either would be fine.

cesarbs:
@justinvp @mikeharder Yeah, the point about avoiding an allocation made me change my mind :smile: It's sad thought that we can't have this on all frameworks.

justinvp:
Sorry for not making the benefits more clear from the start. I'll update the PR, addressing @mikeharder's feedback later today.

justinvp:
Feedback addressed. Thanks for the review.

cesarbs:
:shipit:

PR #776, opened 2016-04-27
Use type FullName instead of assembly FullName

khellang:
Fixes #775

PR #777, opened 2016-04-27
Use type's namespace instead of assembly FullName

### LINKED ISSUE #775: Kestrel Logs print full assembly name as category name
Makes it harder to filter kestrel logs:

````
dbug: Microsoft.AspNetCore.Server.Kestrel, Version=1.0.0.0, Culture=neutral, PublicKeyToken=adb9793829ddae60[9]
      Connection id "0HKREDSFMRPFN" completed keep alive response.
```

This regressed here https://github.com/aspnet/KestrelHttpServer/commit/c48353f4ef1150b3197246de94a343ae8d9b29d4#diff-2a47c9b99ecab7d9ea77789228f7e771R44

khellang:
Same as #776, but targeting release. Fixes #775

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Kestrel/KestrelServer.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/KestrelServer.cs
@@ -41,7 +41,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel
 
             Options = options.Value ?? new KestrelServerOptions();
             _applicationLifetime = applicationLifetime;
-            _logger = loggerFactory.CreateLogger(typeof(KestrelServer).GetTypeInfo().Assembly.FullName);
+            _logger = loggerFactory.CreateLogger(typeof(KestrelServer).GetTypeInfo().Namespace);
             Features = new FeatureCollection();
             var componentFactory = new HttpComponentFactory(Options);
             Features.Set<IHttpComponentFactory>(componentFactory);


mikeharder:
@davidfowl: I can merge this to release and dev if you give it a :shipit: and get it approved for RC2.

halter73:
:shipit:

mikeharder:
@davidfowl: The previous value was the namespace `Microsoft.AspNetCore.Server.Kestrel` rather than the type `Microsoft.AspNetCore.Server.Kestrel.KestrelServer`.  Which do we want going forward?

davidfowl:
The former one looks nicer

khellang:
Actually, why not just use `typeof(KestrelServer).Name`? What is `CreateLogger <T>` using?

halter73:
@khellang I agree that `Microsoft.AspNetCore.Server.Kestrel` looks nicer. It also makes the most sense to me, because the `KestrelServer` class isn't doing any logging directly.

khellang:
Aight. Changed to namespace. Or did you just want med to hard code `Microsoft.AspNetCore.Server.Kestrel`?

Anything looks better than `Microsoft.AspNetCore.Server.Kestrel, Version=1.0.0.0, Culture=neutral, PublicKeyToken=adb9793829ddae60` :wink:

davidfowl:
Yah, hardcode it

khellang:
Hardcoded it is :grin:

davidfowl:
@Eilon suggested we do `typeof(KestrelServer).Namespace`

khellang:
Fffffffuuuuu. That's what I had before hard coding it :stuck_out_tongue_closed_eyes:

Eilon:
Yeah don't listen to @davidfowl 😄

benaadams:
Is `typeof(KestrelServer).Namespace` a compile time constant? (and if not why not...)

Eilon:
@benaadams it's not, but I think in this case it's once per app.

benaadams:
@Eilon raised issue https://github.com/dotnet/roslyn/issues/10972

mikeharder:
We decided to take this change for post-RC2.  Manually merged to dev in 8a9840216c0793789e444f490009afe0a72acbb1.

Issue #679, opened 2016-03-08
Use IOCP on secondary Windows threads


halter73:
@geoffkizer did some perf analysis and discovered that Kestrel's secondary threads weren't using IOCP on Windows. I [brought this up on the libuv mailing list](https://groups.google.com/forum/#!topic/libuv/5DAW1hViw0Q). Here is an excerpt of what I wrote:

> After doing some profiling, we noticed that the connections accepted on the duplicated socket on the secondary threads are using event handles instead of IOCP for async completion on windows, e.g.:

    int uv_tcp_write(uv_loop_t* loop,
                     uv_write_t* req,
                     uv_tcp_t* handle,
                     const uv_buf_t bufs[],
                     unsigned int nbufs,
                     uv_write_cb cb) {
      int result;
      DWORD bytes;

      uv_req_init(loop, (uv_req_t*) req);
      req->type = UV_WRITE;
      req->handle = (uv_stream_t*) handle;
      req->cb = cb;

     

      /* Prepare the overlapped structure. */
      memset(&(req->u.io.overlapped), 0, sizeof(req->u.io.overlapped));
      if (handle->flags & UV_HANDLE_EMULATE_IOCP) {
        req->event_handle = CreateEvent(NULL, 0, 0, NULL);
        if (!req->event_handle) {
          uv_fatal_error(GetLastError(), "CreateEvent");
        }
        req->u.io.overlapped.hEvent = (HANDLE) ((ULONG_PTR) req->event_handle | 1);
        req->wait_handle = INVALID_HANDLE_VALUE;
      }

      ...

      if (handle->flags & UV_HANDLE_EMULATE_IOCP &&
            !RegisterWaitForSingleObject(&req->wait_handle,
              req->event_handle, post_write_completion, (void*) req,
              INFINITE, WT_EXECUTEINWAITTHREAD | WT_EXECUTEONLYONCE)) {
          SET_REQ_ERROR(req, GetLastError());
          uv_insert_pending_req(loop, (uv_req_t*)req);
      }

> This seems to occur because we accept the socket on both the primary loop and secondary loop. The first call to uv_tcp_accept associates the socket with the primary loop's completion port which causes the second call to CreateIoCompletionPort (via uv_pipe_accept->uv_tcp_import) to fail on the duplicated socket on the secondary thread. This sequence of events is what causes the UV_HANDLE_EMULATE_IOCP flag to be set.

    static int uv_tcp_set_socket(uv_loop_t* loop,
[...]

PR #778, opened 2016-04-27
Use IOCP on secondary listener threads on Windows

moozzyk:
Addresses #679

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Kestrel/Http/ListenerPrimary.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/Http/ListenerPrimary.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Runtime.InteropServices;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Server.Kestrel.Infrastructure;
 using Microsoft.AspNetCore.Server.Kestrel.Networking;
@@ -11,7 +12,7 @@ using Microsoft.Extensions.Logging;
 namespace Microsoft.AspNetCore.Server.Kestrel.Http
 {
     /// <summary>
-    /// A primary listener waits for incoming connections on a specified socket. Incoming 
+    /// A primary listener waits for incoming connections on a specified socket. Incoming
     /// connections may be passed to a secondary listener to handle.
     /// </summary>
     public abstract class ListenerPrimary : Listener
@@ -19,8 +20,10 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Http
         private readonly List<UvPipeHandle> _dispatchPipes = new List<UvPipeHandle>();
         private int _dispatchIndex;
         private string _pipeName;
+        private IntPtr _fileCompletionInfoPtr;
+        private bool _tryDetachFromIOCP = PlatformApis.IsWindows;
 
-        // this message is passed to write2 because it must be non-zero-length, 
+        // this message is passed to write2 because it must be non-zero-length,
         // but it has no other functional significance
         private readonly ArraySegment<ArraySegment<byte>> _dummyMessage = new ArraySegment<ArraySegment<byte>>(new[] { new ArraySegment<byte>(new byte[] { 1, 2, 3, 4 }) });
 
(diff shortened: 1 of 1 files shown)


halter73:
@davidfowl @DamianEdwards 

Don't want to risk putting this in release?

src/Microsoft.AspNetCore.Server.Kestrel/Http/ListenerPrimary.cs
```diff
@@ -101,12 +112,59 @@ protected override void DispatchConnection(UvStreamHandle socket)
...
+            const int FileReplaceCompletionInformation = 61;
+            // https://msdn.microsoft.com/en-us/library/cc704588.aspx
+            const uint STATUS_INVALID_INFO_CLASS = 0xC0000003;
+
+            var statusBlock = new IO_STATUS_BLOCK();
+            var socket = IntPtr.Zero;
+            Thread.Loop.Libuv.uv_fileno(handle, ref socket);
+
+            if (NtSetInformationFile(socket, out statusBlock, _fileCompletionInfoPtr,
+                (uint)Marshal.SizeOf<FILE_COMPLETION_INFORMATION>(), FileReplaceCompletionInformation) == STATUS_INVALID_INFO_CLASS)
```
halter73: Out of curiosity, what's returned for pre-8.1 versions of windows and linux? I'd kinda expect a `DllNotFoundException` on linux.

DamianEdwards:
We already decided not to make this fix for RC2 weeks ago (as much as I'd really like to). Gotta :shipit:

src/Microsoft.AspNetCore.Server.Kestrel/Http/ListenerPrimary.cs
```diff
@@ -101,12 +112,59 @@ protected override void DispatchConnection(UvStreamHandle socket)
...
+            ulong information;
+        }
+
+        private struct FILE_COMPLETION_INFORMATION
+        {
+            public IntPtr Port;
+            public IntPtr Key;
+        }
+
+        [DllImport("NtDll.dll")]
```
davidfowl: Does this work on nano? Or do we need to use API sets

moozzyk: We require forwarders on Nano because of libuv.

src/Microsoft.AspNetCore.Server.Kestrel/Http/ListenerPrimary.cs
moozzyk: Windows: No, the `ZwSetInformationFile`/`NtSetInformationFile` were available since Windows 2000. The `FILE_COMPLETION_INFORMATION` struct and `FileReplaceCompletionInformation` were introduced in Win 8.1 (https://msdn.microsoft.com/en-us/library/windows/hardware/ff728840(v=vs.85).aspx). Hence no DllNotFound/MethodNotFound. I did verify on windows 7 and the returned value was `STATUS_INVALID_INFO_CLASS` 
Linux: I actually should initialize `_tryDetachFromIOCP` to `PlatformApis.IsWindows` to not even try this.

src/Microsoft.AspNetCore.Server.Kestrel/Http/ListenerPrimary.cs
davidfowl: reverse forwarders you mean

moozzyk: Yes. I will however try to find api-set needed. It is very likely that this api set is not included in dotnet cli nor is installed on win7 by default. If this is the case we could ask cli folks to add it and we would not introduce another dependency on reverse forwarders.

moozzyk: We will have to rely on reverse forwarders for now.

moozzyk: Turns out that ntdll.dll actually is on Nano so it should just work.

moozzyk:
:up::date:

src/Microsoft.AspNetCore.Server.Kestrel/Http/ListenerPrimary.cs
halter73: Nice. I missed the `_tryDetachFromIOCP = PlatformApis.IsWindows;` part.

halter73:
:shipit:

src/Microsoft.AspNetCore.Server.Kestrel/Http/ListenerPrimary.cs
moozzyk: No, you did not. It was originally not there. I changed after your comment.

src/Microsoft.AspNetCore.Server.Kestrel/Http/ListenerPrimary.cs
davidfowl: Kewl

davidfowl:
:shipit:

Issue #780, opened 2016-04-28
reallocating new Headers and Streams on Frames is slow


nathana1:
Additionally the existing pooling support doesn't scale well so it is actually slower than not pooling.  Proposal is to add an option to cache the Headers/Streams on the Frame object and remove current pooling scheme.

benaadams:
Related: https://github.com/aspnet/KestrelHttpServer/issues/754

muratg:
@cesarbs Could you dupe this?

cesarbs:
Dupe of #754.

Issue #754, opened 2016-04-17
Remove first class pooling support from KestrelOptions


davidfowl:
- Remove first class support for pooling parameters in `KestrelOptions`
- Expose `IHttpComponentFactory` as something that can be configured from the outside
- Remove `IHttpComponentFactory` from server features

We should decide if we want to add `IHttpComponentFactory` as part of `KestrelOptions` or have it be a service in the DI container.

benaadams:
With the new Options is it possible to add extension methods that add to the config options?

benaadams:
Go generic?

```csharp
interface IComponentFactory<T>
{
   T CreateComponent();
   void DisposeComponent(T component);
}
```

Not sure how the init params would work; maybe


```csharp
interface IComponentFactory<T, TInit>
{
   T CreateComponent(TInit init);
   void DisposeComponent(T component);
}
```

davidfowl:
Nope, too generic is always bad. That looks like a generic `IFactory<T>` thats trying to replace `new()`

muratg:
@DamianEdwards @Eilon 

If we decide to do this, we'll need to do design and make the changes in RC2. Thoughts?

DamianEdwards:
@davidfowl can you provide some more detail here? As it stands, this doesn't seem like a an area that will be exposed enough to warrant making this change now, i.e. if we change it RTM we're going to break hardly anybody as almost nobody will set this option on `KestrelOptions`.

benaadams:
Hardly anybody... :stuck_out_tongue_winking_eye:

benaadams:
Related: #780

benaadams:
Resolved by  #815 ?

cesarbs:
@benaadams Did you reference the right PR?

benaadams:
@cesarbs lol no, should be https://github.com/aspnet/KestrelHttpServer/pull/791

PR #781, opened 2016-04-28
cache headers and streams on frame

nathana1:
Addresses #780


src/Microsoft.AspNetCore.Server.Kestrel/Http/Frame.cs
```diff
@@ -176,16 +176,28 @@ public bool HasResponseStarted
...
+            if(_frameHeaders == null)
+            {
+                _frameHeaders = new Headers();
+                _frameHeaders.Initialize(DateHeaderValueManager);
+                RequestHeaders = _frameHeaders.RequestHeaders;
+                ResponseHeaders = _frameHeaders.ResponseHeaders;
+            }
+            else
+            {
+                _frameHeaders.Initialize(DateHeaderValueManager);
```
mikeharder: Can the call to `Initialize()` be moved after the `if` block, instead of being duplicated in both blocks?

src/Microsoft.AspNetCore.Server.Kestrel/Http/Frame.cs
```diff
@@ -250,21 +262,36 @@ public void Reset()
 
         protected void ResetComponents()
         {
-            var frameHeaders = Interlocked.Exchange(ref _frameHeaders, null);
-            if (frameHeaders != null)
+            if(ServerOptions?.CacheHeadersAndStreams ?? false)
```
mikeharder: Can just use `if (ServerOptions.CacheHeadersAndStreams)`.  Most Kestrel code assumes `ServerOptions` is not null.

nathana1: This is null in some of the tests so it fails.  Let me know if changing tests is a better option.

halter73:
Can we copy the perf numbers from the email thread to this PR?

src/Microsoft.AspNetCore.Server.Kestrel/Http/Frame.cs
halter73: I would fix the tests

benaadams:
It does mean sharing them between user connections when behind a proxy

halter73:
@benaadams True. Most of the benchmarking we've done hasn't involved handling new connections. I would at least like to know the performance characteristics of this when handling a bunch of new connections compared to what we had before. I don't think it's a simple as adding a flag to wrk, but it shouldn't be that difficult.

benaadams:
@halter73 I think you can just turn off keep alives?

How about pool on connection close; get from pool on connection start; cache on connection during rest (as per this) - though would mean using them during keep-alive idle. Could be resolved with as part of timeouts (e.g. idle > 5 sec => pool)

nathana1:
We should get about 15% win on the plaintext scenario with this change.  A couple of other fixes we have been discussing give similar sized improvements (like #778).  These improvements are largely additive when combined.  With the current pooling implementation there is a 50% reduction in throughput.

|Scenario|Reqs/sec|
|---|---|---|---|---|
|Baseline|1009999|
|IOCP|1114899|
|This change|1157813|
|Known header removal|1189857|
|All 3|1449341|

benaadams:
@halter73 yeah ok, don't think its a flag; could try adding `Connection: close` header in the wrk script

halter73:
I like the keep alive idea. You might just be able to add a `Connection: close` header using a wrk script assuming wrk will quickly create a new connection.

The initial reason for looking into this logic was that `ConcurrentQueue` was not scaling as well as `ConcurrentBag` when there is no upper bound for pooled objects, but the `ConcurrentQueue` performs better in real world where there needs to be a pool limit.

This PR solves the problem neatly in the context of the techempower benchmarks where we don't handle many new connections by avoiding pools altogether. The pooling logic did feel a little yucky, especially considering the default pool size is zero, but if pooling helps perf by a ton when there a bunch of new connections that's probably more imported. We definitely need to measure.

src/Microsoft.AspNetCore.Server.Kestrel/KestrelServerOptions.cs
```diff
@@ -14,21 +14,12 @@ public class KestrelServerOptions
...
-
-        /// <summary>
-        /// Gets or sets value that instructs <seealso cref="KestrelServer"/> whether it is safe to 
-        /// pool the Request and Response headers
-        /// for another request after the Response's OnCompleted callback has fired. 
-        /// When this values is greater than zero, it is not safe to retain references to feature components after this event has fired.
-        /// Value is zero by default.
-        /// </summary>
-        public int MaxPooledHeaders { get; set; }
+        public bool CacheHeadersAndStreams { get; set; }
```
davidfowl: Naw, this should be a first class thing. We need to have an interface that somebody can implement that will give them control over the lifetime of these things. There should be no first class boolean that says if we're caching headers or streams.

halter73: @benaadams already referenced it in the issue, but #754 describes the current idea for how we want to support this. That design doesn't really support per-connection "caching" as it's done in this PR though.

benaadams:
@halter73 issue with `ConcurrentBag` is it allocates like crazy https://github.com/dotnet/corefx/issues/7917

src/Microsoft.AspNetCore.Server.Kestrel/KestrelServerOptions.cs
davidfowl: I don't understand why that wouldn't be addressed by the same issue. The idea is to get rid of a first class notion of caching 2 things that handle "the lifetime of things kestrel allocates via some extensibility point". This is the first 2 of many things we'll want to enable pooling for. Are we going to keep adding flags:

```C#
public bool CacheHeadersAndStreams { get; set; }
public bool CacheWriteContexts { get; set; }
public bool CacheSomeotherStuff { get; set; }

```

src/Microsoft.AspNetCore.Server.Kestrel/Http/Frame.cs
```diff
@@ -176,16 +176,28 @@ public bool HasResponseStarted
...
+                _frameHeaders.Initialize(DateHeaderValueManager);
+            }
         }
 
 
         public void InitializeStreams(MessageBody messageBody)
         {
-            _frameStreams = HttpComponentFactory.CreateStreams(this);
-
+            if(_frameStreams == null)
```
cesarbs: nit: space after if

src/Microsoft.AspNetCore.Server.Kestrel/Http/Frame.cs
cesarbs: nit: space after `if`

moozzyk:
FYI IOCP change is in the dev branch

src/Microsoft.AspNetCore.Server.Kestrel/KestrelServerOptions.cs
halter73: I absolutely agree with the idea behind #754. At the rate we're going Kestrel configuration is going to be far crazier than IIS :grimacing: 

It's just that an `IHttpComponentFactory` singleton likely requires putting components back into a collection in between each request or allocating a new component each time. This PR is trying to optimize further than that by keeping the components in a field for the duration of the connection.

How would the `IHttpComponentFactory` be designed so it could communicate to `Frame.cs` that it should keep reusing the component until the the connection is closed instead of returning it? Some sort of flag/bool/configuration seems to be the most obvious way if we what to make component per request or per connection configurable.

mikeharder:
If we're caching objects per `Frame`, we should be able to use a `ConcurrentDictionary<Frame, object>` rather than a `ConcurrentBag` which shares objects across all frames.  We should compare the performance of `ConcurrentDictionary<Frame, object>` against fields on Frame.

nathana1:
I'm working on a methodology that lets us test a lot of connection churn without running out of ephemeral ports on the client so I can get reliable numbers.  I'll add ConcurrentDictionary to my matrix as well.

mikeharder:
With `CacheHeadersAndStreams=false`, this change appears to regress Plaintext by a few percent.

Branch | RPS (millions)
------------ | -------------
dev | 1.080
nathana1/connection-cache | 1.044

mikeharder:
Ignore my previous comment.  The perf difference is caused by this branch missing the improvement 75adbc18a2b6fda212322e3068b6e9f99d64cdd8.

halter73:
@davidfowl So you basically want this PR without `ServerOptions.CacheHeadersAndStreams`? Just always cache and hope users don't corrupt the connection?

halter73:
Closing since this is pretty much the equivalent of #851 with the option to disable caching, and we decided to remove the option. Thanks!

PR #783, opened 2016-04-29
Fix race causing request corruption/data loss

### LINKED ISSUE #782: Race causing request corruption/data loss
This issue was first caught by this sometimes-failing test:

       xUnit.net Runner (64-bit win10-x64)
          Stack Trace:
            C:\Users\shalter\dev\Universe\KestrelHttpServer\test\Microsoft.AspNetCore.Server.Kestrel.FunctionalTests\RequestTests.cs(115,0): at Microsoft.AspNetCore.Server.Kestrel.FunctionalTests.RequestTests.<LargeMultipartUpload>d__1.MoveNext()

Basically what's happening here is that the final call to  `MessageBody.ForContentLength.ReadAsyncImplementation` stalls because even though the client sent the full request body, `SocketInputExtensions.ReadAsync` did not return it all throughout the course of the request. When the `HttpClient` eventually reset the connection after hitting the default 100 second timeout, `ReadAsyncImplementation` would incorrectly throw a `BadHttpRequestException: Unexpected end of request content`.

Ultimately the reason `SocketInputExtensions.ReadAsync` lost some request data was due to a race condition in `MemoryPoolIterator.CopyTo`. In between calculation `following` and checking if `block.Next == null`, request data was both written after `block.End` *and* more data was then added into `block.Next`. Here is a fix to `CopyTo` that we ultimately probably aren't going to take because we would have to also fix `Seek`, `Peek`, `Skip`, `Take` and possibly other not yet considered methods.

    --- a/src/Microsoft.AspNetCore.Server.Kestrel/Infrastructure/MemoryPoolIterator.cs
    +++ b/src/Microsoft.AspNetCore.Server.Kestrel/Infrastructure/MemoryPoolIterator.cs
    @@ -679,6 +679,11 @@ public MemoryPoolIterator CopyTo(byte[] array, int offset, int count, out int ac
                 var remaining = count;
                 while (true)
                 {
    +                // Determine if we might attempt to copy data from block.Next before
    +                // calculating "following" so we don't risk skipping data that could
    +                // be added after block.End when we decide to copy from block.Next.
    +                // block.End will always be advanced before block.Next is set.
    +                var isLastBlock = block.Next == null;
                     if (remaining <= following)
                     {
                         actual = count;
                         if (array != null)
                         {
                             Buffer.BlockCopy(block.Array, index, array, offset, remaining);
                         }
                         return new MemoryPoolIterator(block, index + remaining);
                     }
    -                else if (block.Next == null)
    +                else if (isLastBlock)
                     {
                         actual = count - remaining + following;
                         if (array != null)

Instead we are likely going to stop reading new request data into "unoccupied tail-space".

halter73:
Always put new request data in new memory pool blocks
- This can prevent races in CopyTo, Seek, etc...

Fixes #782

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Kestrel/Http/SocketInput.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/Http/SocketInput.cs
@@ -43,17 +43,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Http
 
         public MemoryPoolBlock IncomingStart()
         {
-            const int minimumSize = 2048;
-
-            if (_tail != null && minimumSize <= _tail.Data.Offset + _tail.Data.Count - _tail.End)
-            {
-                _pinned = _tail;
-            }
-            else
-            {
-                _pinned = _memory.Lease();
-            }
-
+            _pinned = _memory.Lease();
             return _pinned;
         }
 
@@ -63,17 +53,17 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Http
             {
                 if (count > 0)
                 {
-                    if (_tail == null)
-                    {
-                        _tail = _memory.Lease();
-                    }
-
-                    var iterator = new MemoryPoolIterator(_tail, _tail.End);
+                    var startBlock = _memory.Lease();
+                    var iterator = new MemoryPoolIterator(startBlock);
                     iterator.CopyFrom(buffer, offset, count);
 
                     if (_head == null)
                     {
-                        _head = _tail;
+                        _head = startBlock;
+                    }
+                    else
+                    {
+                        _tail.Next = startBlock;
                     }
 
                     _tail = iterator.Block;
@@ -98,10 +88,6 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Http
                     if (_head == null)
                     {
                         _head = _tail = _pinned;
-                    }
-                    else if (_tail == _pinned)
-                    {
-                        // NO-OP: this was a read into unoccupied tail-space
                     }
                     else
                     {


halter73:
> dotnet run -- -n plaintext -s http://adx-3u-perfsvr:5001 -c http://adx-3u-svr02:5002 -o KestrelHttpServer@release
    Project Repository (.NETStandardApp,Version=v1.5) was previously compiled. Skipping compilation.
    Project Benchmarks.ClientJob (.NETStandardApp,Version=v1.5) was previously compiled. Skipping compilation.
    Project Benchmarks.ServerJob (.NETStandardApp,Version=v1.5) was previously compiled. Skipping compilation.
    Project BenchmarksDriver (.NETStandardApp,Version=v1.5) was previously compiled. Skipping compilation.
    [05:31:01.208] Starting scenario Plaintext on benchmark server...
    [05:31:51.255] Warmup
    [05:31:51.269] Starting scenario Plaintext on benchmark client...
    [05:32:02.943] Scenario Plaintext completed on benchmark client
    [05:32:02.943] RPS: 1017722.69
    [05:32:02.943] Stopping scenario Plaintext on benchmark client...
    [05:32:02.952] Benchmark
    [05:32:02.953] Starting scenario Plaintext on benchmark client...
    [05:32:14.022] Scenario Plaintext completed on benchmark client
    [05:32:14.022] RPS: 1030647.49
    [05:32:14.022] Stopping scenario Plaintext on benchmark client...
    [05:32:14.025] Stopping scenario Plaintext on benchmark server...
    > dotnet run -- -n plaintext -s http://adx-3u-perfsvr:5001 -c http://adx-3u-svr02:5002 -o KestrelHttpServer@halter73/no-occupancy
    Project Repository (.NETStandardApp,Version=v1.5) was previously compiled. Skipping compilation.
    Project Benchmarks.ClientJob (.NETStandardApp,Version=v1.5) was previously compiled. Skipping compilation.
    Project Benchmarks.ServerJob (.NETStandardApp,Version=v1.5) was previously compiled. Skipping compilation.
    Project BenchmarksDriver (.NETStandardApp,Version=v1.5) was previously compiled. Skipping compilation.
    [05:32:25.925] Starting scenario Plaintext on benchmark server...
    [05:32:44.219] Warmup
    [05:32:44.234] Starting scenario Plaintext on benchmark client...
    [05:32:54.646] Scenario Plaintext completed on benchmark client
    [05:32:54.646] RPS: 1007649.22
    [05:32:54.646] Stopping scenario Plaintext on benchmark client...
    [05:32:54.650] Benchmark
    [05:32:54.650] Starting scenario Plaintext on benchmark client...
    [05:33:05.087] Scenario Plaintext completed on benchmark client
    [05:33:05.087] RPS: 1027027.92
    [05:33:05.087] Stopping scenario Plaintext on benchmark client...
[05:33:05.090] Stopping scenario Plaintext on benchmark server...

davidfowl:
post RC2, we should unify the libuv and connection filter code paths. We can just do the IncomingStart(), IncomingComplete thing instead of having IncomingData. Writing to the block would always happen from the outside.

cesarbs:
Is the perf drop consistent across benchmark runs, or it it statistically insignificant?

Whatever the case :shipit: Much better doing things this way :smile:

mikeharder:
I suspect the perf drop is just noise, but even if this change does reduce perf by 0.3%, it's worth it.

halter73:
Replace by #800

Issue #683, opened 2016-03-09
Improve request line parsing


cesarbs:
Check for:

- Invalid/missing method
- Invalid/missing target
- Invalid/missing HTTP version
- Missing CR LF

Should return 400 for all except invalid HTTP version, which should be 505 (https://github.com/aspnet/KestrelHttpServer/issues/712).

cesarbs:
Updated the issue to reflect new reality.

Issue #785, opened 2016-04-29
Limit request target length


cesarbs:
Currently there's no limit; add configurable limit and reject requests for targets that are too long.

Related: #683 #784

cesarbs:
cc @muratg putting in 1.0

Tratcher:
I think this is redundant with #784. The only other things on the request line are the method and version, both very short.

cesarbs:
I'd prefer to keep the distinction because there is a specific status for too long targets (414), whereas a line that is too long would be rejected with a 400. Besides, the method and version are only very short when coming from well behaved clients :smile:

benaadams:
Related https://github.com/aspnet/KestrelHttpServer/issues/475

Tratcher:
True, but we want to keep the number of nobs down to a sane level.

cesarbs:
Agreed that it's redundant with #784. Closing.

Issue #784, opened 2016-04-29
Limit request line length


cesarbs:
Right now there's no set limit on the request line. This should be configurable and requests with request lines longer than the limit should be rejected.

Related: #683

cesarbs:
@muratg putting in 1.0.

benaadams:
Related https://github.com/aspnet/KestrelHttpServer/issues/475

PR #786, opened 2016-04-29
Reject requests with invalid request lines (#683)

cesarbs:
#683

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Kestrel/Http/Frame.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/Http/Frame.cs
@@ -35,6 +35,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Http
         private static readonly byte[] _bytesContentLengthZero = Encoding.ASCII.GetBytes("\r\nContent-Length: 0");
         private static readonly byte[] _bytesSpace = Encoding.ASCII.GetBytes(" ");
         private static readonly byte[] _bytesEndHeaders = Encoding.ASCII.GetBytes("\r\n\r\n");
+        private static readonly int _httpVersionLength = "HTTP/1.*".Length;
 
         private static Vector<byte> _vectorCRs = new Vector<byte>((byte)'\r');
         private static Vector<byte> _vectorColons = new Vector<byte>((byte)':');
@@ -45,7 +46,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Http
         private readonly object _onStartingSync = new Object();
         private readonly object _onCompletedSync = new Object();
 
-        protected bool _corruptedRequest = false;
+        private bool _requestRejected;
         private Headers _frameHeaders;
         private Streams _frameStreams;
 
@@ -60,7 +61,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Http
         protected CancellationTokenSource _abortedCts;
         protected CancellationToken? _manuallySetRequestAbortToken;
 
-        protected bool _responseStarted;
+        protected RequestProcessingStatus _requestProcessingStatus;
         protected bool _keepAlive;
         private bool _autoChunk;
         protected Exception _applicationException;
@@ -96,7 +97,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Http
                 {
                     return "HTTP/1.0";
                 }
-                return "";
+                return string.Empty;
             }
             set
             {
@@ -167,9 +168,10 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Http
                 return cts;
             }
         }
+
         public bool HasResponseStarted
         {
-            get { return _responseStarted; }
+            get { return _requestProcessingStatus == RequestProcessingStatus.ResponseStarted; }
         }
 
         protected FrameRequestHeaders FrameRequestHeaders => _frameHeaders.RequestHeaders;
@@ -216,7 +218,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Http
             _onStarting = null;
             _onCompleted = null;
 
-            _responseStarted = false;
+            _requestProcessingStatus = RequestProcessingStatus.RequestPending;
             _keepAlive = false;
             _autoChunk = false;
             _applicationException = null;
@@ -446,7 +448,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Http
 
         public Task WriteAsync(ArraySegment<byte> data, CancellationToken cancellationToken)
         {
-            if (!_responseStarted)
+            if (!HasResponseStarted)
             {
                 return WriteAsyncAwaited(data, cancellationToken);
             }
@@ -506,7 +508,10 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Http
 
         public void ProduceContinue()
         {
-            if (_responseStarted) return;
+            if (HasResponseStarted)
+            {
+                return;
+            }
 
             StringValues expect;
             if (_httpVersion == HttpVersionType.Http1_1 &&
@@ -519,7 +524,10 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Http
 
         public Task ProduceStartAndFireOnStarting()
         {
-            if (_responseStarted) return TaskUtilities.CompletedTask;
+            if (HasResponseStarted)
+            {
+                return TaskUtilities.CompletedTask;
+            }
 
             if (_onStarting != null)
             {
@@ -554,30 +562,49 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Http
 
         private void ProduceStart(bool appCompleted)
         {
-            if (_responseStarted) return;
-            _responseStarted = true;
+            if (HasResponseStarted)
+            {
+                return;
+            }
+
+            _requestProcessingStatus = RequestProcessingStatus.ResponseStarted;
 
             var statusBytes = ReasonPhrases.ToStatusBytes(StatusCode, ReasonPhrase);
 
             CreateResponseHeader(statusBytes, appCompleted);
         }
 
+        protected Task TryProduceInvalidRequestResponse()
+        {
+            if (_requestProcessingStatus == RequestProcessingStatus.RequestStarted && _requestRejected)
+            {
+                if (_frameHeaders == null)
+                {
+                    InitializeHeaders();
+                }
+
+                return ProduceEnd();
+            }
+
+            return TaskUtilities.CompletedTask;
+        }
+
         protected Task ProduceEnd()
         {
-            if (_corruptedRequest || _applicationException != null)
+            if (_requestRejected || _applicationException != null)
             {
-                if (_corruptedRequest)
+                if (_requestRejected)
                 {
                     // 400 Bad Request
                     StatusCode = 400;
-                } 
+                }
                 else
                 {
                     // 500 Internal Server Error
                     StatusCode = 500;
                 }
 
-                if (_responseStarted)
+                if (HasResponseStarted)
                 {
                     // We can no longer respond with a 500, so we simply close the connection.
                     _requestProcessingStopping = true;
@@ -601,7 +628,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Http
                 }
             }
 
-            if (!_responseStarted)
+            if (!HasResponseStarted)
             {
                 return ProduceEndAwaited();
             }
@@ -709,31 +736,50 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Http
             SocketOutput.ProducingComplete(end);
         }
 
-        protected bool TakeStartLine(SocketInput input)
+        protected RequestLineStatus TakeStartLine(SocketInput input)
         {
             var scan = input.ConsumingStart();
             var consumed = scan;
+
             try
             {
+                // We may hit this when the client has stopped sending data but
+                // the connection hasn't closed yet, and therefore Frame.Stop()
+                // hasn't been called yet.
+                if (scan.Peek() == -1)
+                {
+                    return RequestLineStatus.Empty;
+                }
+
+                _requestProcessingStatus = RequestProcessingStatus.RequestStarted;
+
                 string method;
                 var begin = scan;
-                if (!begin.GetKnownMethod(ref scan, out method))
+                if (!begin.GetKnownMethod(out method))
                 {
                     if (scan.Seek(ref _vectorSpaces) == -1)
                     {
-                        return false;
+                        return RequestLineStatus.MethodIncomplete;
                     }
+
                     method = begin.GetAsciiString(scan);
-                    scan.Take();
+                    if (method == null)
+                    {
+                        RejectRequest("Missing method.");
+                    }
+                }
+                else
+                {
+                    scan.Skip(method.Length);
                 }
 
+                scan.Take();
                 begin = scan;
-
                 var needDecode = false;
                 var chFound = scan.Seek(ref _vectorSpaces, ref _vectorQuestionMarks, ref _vectorPercentages);
                 if (chFound == -1)
                 {
-                    return false;
+                    return RequestLineStatus.TargetIncomplete;
                 }
                 else if (chFound == '%')
                 {
@@ -741,7 +787,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Http
                     chFound = scan.Seek(ref _vectorSpaces, ref _vectorQuestionMarks);
                     if (chFound == -1)
                     {
-                        return false;
+                        return RequestLineStatus.TargetIncomplete;
                     }
                 }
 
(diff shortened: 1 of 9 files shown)


cesarbs:
cc @moozzyk

cesarbs:
I've filed #784 and #785 to address further improvements to request line parsing.

src/Microsoft.AspNetCore.Server.Kestrel/Http/Frame.cs
```diff
@@ -1024,5 +1087,23 @@ private enum HttpVersionType
...
+            Empty,
+            MethodIncomplete,
+            TargetIncomplete,
+            VersionIncomplete,
+            MissingCR,
+            MissingLF,
+            Done
+        }
+
+        protected enum RequestRejectReason
```
muratg: `RequestRejectionReason` instead?

src/Microsoft.AspNetCore.Server.Kestrel/Http/Frame.cs
```diff
@@ -118,6 +119,7 @@ public string HttpVersion
         public IHeaderDictionary RequestHeaders { get; set; }
         public Stream RequestBody { get; set; }
 
+        public bool RejectedRequest => _requestRejectReason != RequestRejectReason.None;
```
muratg: nit: `RequestRejected` instead?

src/Microsoft.AspNetCore.Server.Kestrel/Http/Frame.cs
```diff
@@ -994,12 +1051,18 @@ public bool StatusCanHaveBody(int statusCode)
                    statusCode != 304;
         }
 
-        public void ReportCorruptedHttpRequest(BadHttpRequestException ex)
+        public void ReportBadHttpRequest(BadHttpRequestException ex)
```
Tratcher: Can you simplify the calling pattern by passing a string and constructing the exception internally?

Tratcher: For that matter, why bother constructing the exception if it's never thrown, only logged? Just log the message directly.

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/AddressRegistrationTests.cs
```diff
@@ -96,7 +97,32 @@ private void ConfigureEchoAddress(IApplicationBuilder app)
...
-                return context.Response.WriteAsync(context.Request.GetDisplayUrl());
+                const string SchemeDelimiter = "://";
+                
+                var request = context.Request;
+                var host = request.Host.Value;
+                var pathBase = request.PathBase.Value;
+                var path = request.Path.Value;
+                var queryString = request.QueryString.Value;
+
+                // PERF: Calculate string length to allocate correct buffer size for StringBuilder.
```
Tratcher: perf is irrelevant in tests. Simplify

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/AddressRegistrationTests.cs
```diff
@@ -96,7 +97,32 @@ private void ConfigureEchoAddress(IApplicationBuilder app)
...
+
+                var sb = new StringBuilder(length)
+                    .Append(request.Scheme)
+                    .Append(SchemeDelimiter)
+                    .Append(host)
+                    .Append(pathBase)
+                    .Append(path)
+                    .Append(queryString)
+                    .ToString();
+                //return context.Response.WriteAsync(context.Request.GetDisplayUrl());
```
Tratcher: remove

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/AddressRegistrationTests.cs
```diff
@@ -96,7 +97,32 @@ private void ConfigureEchoAddress(IApplicationBuilder app)
...
             app.Run(context =>
             {
-                return context.Response.WriteAsync(context.Request.GetDisplayUrl());
+                const string SchemeDelimiter = "://";
+                
+                var request = context.Request;
+                var host = request.Host.Value;
+                var pathBase = request.PathBase.Value;
+                var path = request.Path.Value;
+                var queryString = request.QueryString.Value;
```
Tratcher: Why these values instead of GetDisplayUrl?

src/Microsoft.AspNetCore.Server.Kestrel/Http/Frame.cs
cesarbs: There's a place in `MessageBody` where we construct the exception, report it and then throw it.

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/AddressRegistrationTests.cs
cesarbs: Need to revert this. This was copied from elsewhere when I was investigating a null ref a test was throwing.

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/AddressRegistrationTests.cs
cesarbs: See above. I was investigating something and forgot to revert :grin:

src/Microsoft.AspNetCore.Server.Kestrel/Http/Frame.cs
benaadams: Add overload?

src/Microsoft.AspNetCore.Server.Kestrel/Http/Frame.cs
```diff
@@ -752,35 +779,66 @@ protected bool TakeStartLine(SocketInput input)
...
+
                 scan.Take();
                 begin = scan;
+                chFound = scan.Seek(ref _vectorCRs);
+                var scanned = begin.GetLength(scan);
+                if (chFound == -1 && scanned < _httpVersionLength)
+                {
+                    return RequestLineStatus.VersionIncomplete;
+                }
+                else if (chFound == '\r' && scanned == 0)
```
muratg: Could it be plain '\n' instead of "\r\n" per the RFC?

Tratcher: no

cesarbs:
@Tratcher @benaadams Added `Frame.ReportBadHttpRequest()` overload taking a string.

src/Microsoft.AspNetCore.Server.Kestrel/Http/Frame.cs
```diff
@@ -709,39 +718,57 @@ private Task WriteSuffix()
...
-                if (!begin.GetKnownMethod(ref scan, out method))
+                if (!begin.GetKnownMethod(out method))
                 {
                     if (scan.Seek(ref _vectorSpaces) == -1)
                     {
-                        return false;
+                        // We may hit this when the client has stopped sending data but
+                        // the connection hasn't closed yet, and therefore Frame.Stop()
+                        // hasn't been called yet.
+                        if (begin.GetLength(scan) == 0)
```
halter73: `if (begin.Peek() == -1)`

src/Microsoft.AspNetCore.Server.Kestrel/Http/Frame.cs
```diff
@@ -752,35 +779,66 @@ protected bool TakeStartLine(SocketInput input)
...
-                    if (scan.Seek(ref _vectorSpaces) != ' ')
+                    if (scan.Seek(ref _vectorSpaces) == -1)
                     {
-                        return false;
+                        return RequestLineStatus.TargetIncomplete;
                     }
                     queryString = begin.GetAsciiString(scan);
                 }
 
+                if (pathBegin.GetLength(scan) == 0)
```
halter73: I don't like calling `GetLength` this in the normal start line parsing path. I think we should use BenchmarkDotNet to compare this code path before and after for common start lines.

This is happening a lot, and we've already iterated through all this to get scan variable. I think it might be better to just make this an out param to Seek. I know we do this every time we call GetAsciiString, so maybe that should take a length parameter too.

src/Microsoft.AspNetCore.Server.Kestrel/Http/Frame.cs
```diff
@@ -752,35 +779,66 @@ protected bool TakeStartLine(SocketInput input)
...
+                if (chFound == -1 && scanned < _httpVersionLength)
+                {
+                    return RequestLineStatus.VersionIncomplete;
+                }
+                else if (chFound == '\r' && scanned == 0)
+                {
+                    ReportBadHttpRequest("Missing HTTP version.");
+                    return RequestLineStatus.Done;
+                }
+                else if ((chFound == '\r' &&  scanned < _httpVersionLength) || scanned > _httpVersionLength)
```
halter73: `else if (scanned < _httpVersionLength || scanned > _httpVersionLength)` is easier to read.

You've already returned `RequestLineStatus.VersionIncomplete` `if (chFound == -1 && scanned < _httpVersionLength)`.

halter73: Or better yet `else if (scanned != _httpVersionLength)`. Boolean logic is hard :grin:

src/Microsoft.AspNetCore.Server.Kestrel/Http/Frame.cs
```diff
@@ -246,6 +248,8 @@ public void Reset()
 
             _manuallySetRequestAbortToken = null;
             _abortedCts = null;
+
+            _requestRejectReason = RequestRejectionReason.None;
```
mikeharder: Why does `_requestRejectReason` need to be reset, when `_corruptedRequest` was not?

src/Microsoft.AspNetCore.Server.Kestrel/Http/Frame.cs
cesarbs: I can actually move this outside of this block and check it at the very start.

src/Microsoft.AspNetCore.Server.Kestrel/Http/Frame.cs
cesarbs: I don't like it either. I think I'll move this check to after we parse the request target, so I can just check against `string.Empty`.

src/Microsoft.AspNetCore.Server.Kestrel/Http/Frame.cs
cesarbs: At some point when I was making theses changes this was necessary, but now it's just a leftover from a previous state of the code.

src/Microsoft.AspNetCore.Server.Kestrel/Http/Frame.cs
```diff
@@ -118,6 +119,7 @@ public string HttpVersion
         public IHeaderDictionary RequestHeaders { get; set; }
         public Stream RequestBody { get; set; }
 
+        public bool RequestRejected => _requestRejectionReason != RequestRejectionReason.None;
```
mikeharder: The previous field `_corruptedRequest` was `protected` rather than `public`.  Should `RequestRejected` also be protected?

src/Microsoft.AspNetCore.Server.Kestrel/Http/FrameOfT.cs
```diff
@@ -33,27 +33,35 @@ public class Frame<TContext> : Frame
...
                         if (SocketInput.RemoteIntakeFin)
                         {
                             // We need to attempt to consume start lines and headers even after
                             // SocketInput.RemoteIntakeFin is set to true to ensure we don't close a
                             // connection without giving the application a chance to respond to a request
                             // sent immediately before the a FIN from the client.
-                            if (TakeStartLine(SocketInput))
+                            var requestLineStatus = TakeStartLine(SocketInput);
+                            if (requestLineStatus != RequestLineStatus.Empty &&
+                                requestLineStatus != RequestLineStatus.Done)
```
halter73: It it's done parsing, you should break. If the request line is empty you should return.

src/Microsoft.AspNetCore.Server.Kestrel/Http/FrameOfT.cs
```diff
@@ -33,27 +33,35 @@ public class Frame<TContext> : Frame
...
                             {
-                                break;
+                                ReportBadHttpRequest($"Invalid request line: {requestLineStatus}");
```
halter73: It should be simple to have `ReportBadHttpRequest($"Incomplete headers");` before you return in the following while loop.

src/Microsoft.AspNetCore.Server.Kestrel/Http/Frame.cs
cesarbs: Yes, thanks for pointing that out.

src/Microsoft.AspNetCore.Server.Kestrel/Http/Frame.cs
cesarbs: I figured I can just do `if (pathBegin.Peek() == ' ')` - if there's a space there we don't even care what comes next, since there should only be one space between the method and the target.

cesarbs:
@halter73 Addressed your suggestion to return 400 on invalid headers. There's still more work to be done on headers but that should happen as part of separate PRs.

cesarbs:
Note how `EngineTests.ConnectionClosesWhenFinReceivedBeforeRequestCompletes` looks more correct with this change :grin:

cesarbs:
Not attempting to identify unsupported HTTP versions anymore to send 505 response; sending 400 instead.

The logic that was in place to send a 505 only made things more complicated and added little value. RFC 7230 is not even too strong on sending a 505:

>  A server can send a 505 (HTTP Version Not Supported) response if it wishes, for any reason, to refuse service of the client's major protocol version.

IIS is quite inconsistent in sending 505s or 400s when a request has a version different than HTTP/1.0 or HTTP/1.1.

IMHO being concerned about sending a 505 would only make sense if Kestrel was only an HTTP/1.0 server and wanted to reject HTTP/1.1 requests.

cesarbs:
`MemoryPoolIterator.Skip()` is broken and I'm fixing it in this PR.

cesarbs:
Pending: `BadHttpRequestTests` should check headers too.

src/Microsoft.AspNetCore.Server.Kestrel/Infrastructure/MemoryPoolIterator.cs
```diff
@@ -98,38 +98,30 @@ public int Take()
 
         public void Skip(int bytesToSkip)
```
halter73: If I were to change this, I would have it throw if there aren't enough following bytes instead of advancing to the end.

Attempting to skip more bytes than there actually are is indicative of a bug in the caller, and none of our existing code would ever skip too many bytes.

cesarbs: Agreed as discussed offline. I'll revert these changes.

cesarbs:
* Rebased
* Reverted changes to `MemoryPoolIterator.Skip()`
* Check headers in `BadHttpRequestTests`

src/Microsoft.AspNetCore.Server.Kestrel/Infrastructure/MemoryPoolIterator.cs
```diff
@@ -124,28 +124,22 @@ public void Skip(int bytesToSkip)
                 {
                     return;
```
halter73: throw

test/Microsoft.AspNetCore.Server.KestrelTests/BadHttpRequestTests.cs
```diff
@@ -0,0 +1,64 @@
...
+        [InlineData("GET / HTTP/3.0\r\n\r\n")]
+        [InlineData("GET / H\r\n\r\n")]
+        [InlineData("GET / HTTP/1.\r\n\r\n")]
+        [InlineData("GET / hello\r\n\r\n")]
+        [InlineData("GET / 8charact\r\n\r\n")]
+        public async Task TestBadRequests(string request)
+        {
+            using (var server = new TestServer(context => { return Task.FromResult(0); }))
+            {
+                using (var socket = new Socket(AddressFamily.InterNetwork, SocketType.Stream, ProtocolType.Tcp))
```
halter73: Use TestConnection

halter73:
:shipit:

src/Microsoft.AspNetCore.Server.Kestrel/Http/FrameOfT.cs
```diff
@@ -158,6 +154,8 @@ public class Frame<TContext> : Frame
             {
                 try
                 {
+                    await TryProduceEnd();
```
halter73: This might actually break server shutdown as written. We might return an empty 200 response when before we would close the connection. Closing the connection immediately in the middle of reading the start line or headers may have been wrong, but returning a 200 without calling application code would be worse.

halter73:
Don't :ship: until addressing my last comment.

src/Microsoft.AspNetCore.Server.Kestrel/Http/Frame.cs
```diff
@@ -554,30 +562,49 @@ public Task ProduceStartAndFireOnStarting()
...
+            _requestProcessingStatus = RequestProcessingStatus.ResponseStarted;
 
             var statusBytes = ReasonPhrases.ToStatusBytes(StatusCode, ReasonPhrase);
 
             CreateResponseHeader(statusBytes, appCompleted);
         }
 
+        protected Task TryProduceEnd()
+        {
+            if (_requestProcessingStatus == RequestProcessingStatus.RequestStarted)
```
halter73: Maybe call this TrySend400 and make this `if (_requestProcessingStatus == RequestProcessingStatus.RequestStarted && _requestRejected)`

cesarbs:
@halter73 Updated.

PR #787, opened 2016-04-30
Optimize Memorypool block use

benaadams:
Only use pooled blocks in tests (tests only place not using pooled blocks)
Remove `GCHandle` from blocks used for no longer present temp blocks
Remove now redundant use of nooping `pin`+`unpin` functions w/ memory blocks
Remove unnecessary `fixed` blocks as block pointer is fixed for lifetime

From https://github.com/aspnet/KestrelHttpServer/pull/519

/cc @halter73 @cesarbs @davidfowl


davidfowl:
@halter73 and I were discussing this yesterday 😄

halter73:
:shipit:

PR #789, opened 2016-04-30
[Exploritory] Relayer

benaadams:
Still needs work


benaadams:
I think next step would be ServiceContext being a Kestrel thing so coming out of Connection

davidfowl:
`Frame` : `ServiceContext` looks problematic,

benaadams:
In ServiceConext is a lot of http specific stuff rather than networking

These might be valid for networking
```csharp
public IApplicationLifetime AppLifetime { get; set; }
public IConnectionTrace Log { get; set; }
public IThreadPool ThreadPool { get; set; }
```
This is probably split (e.g. `NoDelay`)
```csharp
public KestrelServerOptions ServerOptions { get; set; }
```
These are very much not Connection related
```csharp
public DateHeaderValueManager DateHeaderValueManager { get; set; }
public IHttpComponentFactory HttpComponentFactory { get; set; }
public Func<IConnectionContext, Frame> FrameFactory { get; set; }
public Action<IFeatureCollection> PrepareRequest { get; set; }
```

benaadams:
Reverting the ServiceConext thing but FrameFactory can be `Func<IConnectionContext, IFrameControl>` if the methods Start, Stop and Abort are added to it.

benaadams:
Better?

benaadams:
Better? ComponentFactory crosses the line; but its an idea

benaadams:
Going to leave it alone now :)

benaadams:
Closing :)

PR #790, opened 2016-05-01
Pool only when next request not ready

benaadams:
Also simpler alternative to `ConcurrentQueue` based on `ArrayPool`

Alternate to #781 with returning to pool when next request is not ready

Addresses #780

/cc @nathana1 @halter73 @mikeharder


benaadams:
This one is probably better https://github.com/aspnet/KestrelHttpServer/pull/791

Issue #631, opened 2016-02-14
[Needs Design] SocketOutput can be pooled between requests


benaadams:
At the end of a request whether there are any outstanding writes can be checked:
* if none, it can be returned to a pool. 
* if some a final "write" can be added to its queue that returns it to pool.

At the start of a request reinitialized with `UvStreamHandle` `Connection` etc

This will mean only "in process" requests are consuming memory for the `SocketOutput`, and not open connections.

PR #791, opened 2016-05-01
Move pooling to seperate package, add use to samples

### LINKED ISSUE #754: Remove first class pooling support from KestrelOptions

### LINKED ISSUE #780: reallocating new Headers and Streams on Frames is slow

benaadams:
Alternate to #781 with returning to pool when next request is *not* ready

Resolves #780 reallocating new Headers and Streams on Frames is slow
Resolves #754 Remove first class pooling support from KestrelOptions

The entire `Headers` and `Streams` lifetime is controlled by `options` `.StreamFactory` and `.HeaderFactory`. 

*Note:* While the `Microsoft.AspNetCore.Server.Kestrel.Pools` package adds pooling via the factories; the factory extensibility point could also be easily utilized to add monitoring, telemetry etc; so its use now is available to more than just pooling.

No pooling (default)
```csharp
.UseKestrel()
```
Pooling can be set by including the `"Microsoft.AspNetCore.Server.Kestrel.Pools": "0.1.0-*"` package
```csharp
.UseKestrel(options =>
{
    options.StreamFactory = new StreamFactory();
    options.HeaderFactory = new HeaderFactory();
})
```
```csharp
.UseKestrel(options =>
{
    options.StreamFactory = new StreamFactory(MaxPooled: 256);
    options.HeaderFactory = new HeaderFactory(MaxPooled: 256);
})
```
```csharp
.UseKestrel(options =>
{
    options.StreamFactory = new StreamFactory() { MaxPooled: 512 };
    options.HeaderFactory = new HeaderFactory() { MaxPooled: 512 };
})
```
Or alternately using a user defined factory that implements the interface

`IComponentFactory<T> where T : class, IComponent` 

Currently the only factories supported are where `T` is `Streams` or `Headers`, however this is easy extensible for future types e.g. https://github.com/aspnet/KestrelHttpServer/issues/631 `SocketOuput` during idle etc

**In Microsoft.AspNetCore.Server.Kestrel\Http\Frame.cs**

Request start
* In InitalizeHeaders/Streams *if* the component is `null`, factory`.Create()` or component .ctor if factory `null` called to create the component
* The component is then initialized

Request end
* If factory is `null` new headers/streams are discarded and streams & headers are set to `null`
* If factory is `not null` `Dispose` is called on the factory with the component passed by `ref` and
 * If the next request *is* ready `requestImmediateReuse == true`.
 * If the next request *is not* ready `requestImmediateReuse == false`.

If factory is `null` new headers/streams are *always* recreated for each request and *always* discarded at the end of the request and streams & headers are set to `null` - so there is no pooling.

**In Microsoft.AspNetCore.Server.Kestrel.Pools**

`T Create`
* If `MaxPooled == 0` new component is always created
* If `MaxPooled > 0` component is attempted to be retrieved from the pool, if not a new one is created

`void Dispose(ref T component, bool requestImmediateReuse)`
* If `MaxPooled == 0` passed `byref` component is set to null
* If `MaxPooled > 0 && requestImmediateReuse == true` component`.Reset()` is called
* If `MaxPooled > 0 && requestImmediateReuse == false` 
 * component`.Uninitalize()` is called
 * component is returned to pool
 *  `byref` component is set to null

**Breaking Public Api Changes**
* `DateHeaderValueManager`  moved from property of `ServiceContext` to property of `Frame`
* `HttpComponentFactory` removed from `ServiceContext`
*  `MaxPooledStreams` removed from `KestrelServerOptions`
*  `MaxPooledHeaders` removed from `KestrelServerOptions`
* `IHttpComponentFactory` interface removed
* `HttpComponentFactory` class removed

**Api Changes**
* `FrameDuplexStream` public rather than private
* `FrameResponseStream` public rather than private
* `FrameResponseStream` initializes with `IFrameControl` rather than `FrameContext`
* `IComponent` with `Reset()` + `Uninitalize()` added
* `IComponentFactory` with `T Create()` + `Dispose(ref T component, bool requestImmediateReuse)` added

/cc @davidfowl @nathana1 @halter73 @mikeharder  @Eilon @muratg @DamianEdwards


nathana1:
I haven't done detailed analysis yet but I'm seeing mixed results so far on our techempower hardware:

|Scenario|Million Reqs/sec|
|-------|---------|
|Baseline|4.25|
|#781 on|4.95|
|#791 default|3.81|
|#791 pool size 120|4.65|

This is the default plaintext scenario, baseline is current kestrel dev branch.  I'll get some profiles and try this out with some connection churn as well.

benaadams:
I'm having issue compiling the benchmarks repo for coreclr in release; will make a small app and see if it can be improved.

cesarbs:
@mikeharder

mikeharder:
What problem are you having compiling benchmarks?  A new CLI was just accepted into the dev branches (1.0.0-rc2-002543), so you might need to delete the `.build` directory and run `build.cmd` to pickup the latest CLI.

benaadams:
not sure, build works but only creates debug and when I use dotnet build/publish it only makes full framework. I'm have a general issue with the dotnet tools recognising the core app/framework types - I assuming its some flux.

mikeharder:
Have you tried this?

```
set Configuration=Release
build.cmd
```

benaadams:
@mikeharder that's just cheating :) I wish I'd known that months ago!

nathana1:
It looks like a lot more time is being spent in DateHeaderValueManager.PrepareDateValues.  It's at 13.2% inclusive time in the default case for this pr vs. 1% in the baseline.

benaadams:
O_o suggests I broke something in the refactor, good spot wasn't looking there.

benaadams:
@nathana1 was calling it at start and end of request

benaadams:
@nathana1 hopefully this should be behaving better?

nathana1:
For the default techempower scenario this looks better, no pooling is the same as the baseline and with pooling is unchanged from before.

I've been doing some more testing with a pipeline depth of 1 (instead of 16) and depth=1 with some connections being closed.  In these cases with pooling on I'm only able to get to 30% cpu, the other scenarios saturate.  I get about 50-75% of the throughput of the baseline with pooling in these cases.  I'm still investigating the bottleneck here.

benaadams:
@nathana1 its the spinlock sleeping due to contention :( - will mitigate

benaadams:
Broken pipe on OSX :-/

benaadams:
Fixed lock contention, but broken pipe on OSX :-/

benaadams:
@nathana1 not proud, but the contention should now be gone

nathana1:
I'm seeing good results with this change, it is within 3% of #781 on the default techempower scenario and is the same as #781 for the 2 scenarios I've been testing with pipelining depth of 1.

I did some experimenting around breaking the object pool array into buckets and then protecting each bucket with a separate lock (https://github.com/nathana1/KestrelHttpServer/commit/b184d3bf025f8fa262ac44f7d2435669ab5e0d4e).  I get equivalent perf as your most recent change.  One issue with this is the way I'm distributing objects into buckets (massaging the thread id) seems hacky, so we would need a way to get fairly even distribution without having to take lock and defeat the purpose.

src/Microsoft.AspNetCore.Server.Kestrel/Infrastructure/Components/StreamFactory.cs
```diff
@@ -0,0 +1,14 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.AspNetCore.Server.Kestrel.Http;
+
+namespace Microsoft.AspNetCore.Server.Kestrel.Infrastructure
+{
+    public class StreamFactory : DefaultComponentFactory<Streams>
```
halter73: Why are the default factories so complicated if the default pool size is zero? I don't think our default factories should have any collections at all.

I think it may be better to move the pooling factories into another project to show it can be easily done and to simplify the main code base.

benaadams: Should the default factory throw if you try to set `MaxPooled` to anything other than zero?

halter73: I'm not sure IComponentFactory needs a MaxPooled property. Certainly not a settable one. That could be a property of your non-default PoolingComponentFactory or whatever that ends up being called.

I'm also unsure about the need for the IComponentFactory.Reset method, since Frame(OfT) can easily reset it's own components. I don't think the IComponent interface needs to exist at all. Once again, Frame is best to decide when to reset/uninitialize components before it reuses/returns them. The settable CorrelationId seems like a super crazy and leaky abstraction. There has to be a better way to reduce contention.

benaadams:
Added pooling samples 
* `PoolingSimpleApp` just switches on pooling using the default pooling
* `PoolingAlternativeApp` replaces the default pooling for a custom stack based MCS lock

@nathana1 currently it uses binning based on `Interlocked.Increment & mask`; it retains distribution by returning the object back to the pool it came from. Sort of setting thread affinity/checking the core the thread is currently running on/thread local/thread static which all have downsides it was the best I could come up with.

benaadams:
Removed all pooling and added it to samples.

Removed `MaxPooled` and `Reset` from `IComponentFactory`

`StreamFactory` and `HeaderFactory` now default to `null`; Frame does not call `Reset` if they are `null` as there is no pooling.

`CorrelationId` has been moved to samples on an inherited sub class of `Headers` and `Streams`

Haven't removed `Uninitalize` from the `IComponent` interface as its up to the factory whether its uninitializing and retaining or throwing away not the `Frame`.

With pooling only being in samples and not a first class component/package any benefits it gives won't really count towards the benchmarks; as it will be "take aspnet add a large amount of bespoke code" rather than "turn on these switches"

@halter73 better?

halter73:
@benaadams Better!

I'm not against putting the component pooling in a separate package. @davidfowl What do you think about us shipping a component pooling package separate from Kestrel proper like we do for HTTPS?

I do wonder if we can't get the caching benefits back somehow. Like I mentioned in [this comment](https://github.com/aspnet/KestrelHttpServer/pull/781/files#r61622390), it would probably require some sort of flag unless we made it default. I didn't like (ab)using the MaxPooled property on IComponentFactory for that purpose.

samples/PoolingAlternativeApp/ComponentFactory.cs
```diff
@@ -0,0 +1,115 @@
...
+using System;
+using System.Runtime.CompilerServices;
+using System.Threading;
+using Microsoft.AspNetCore.Server.Kestrel.Infrastructure;
+
+namespace PoolingAlternativeApp
+{
+    public abstract class ComponentFactory<T> : IComponentFactory<T> where T : class, IComponent
+    {
+        private static readonly int _poolCount = CalcuatePoolCount();
```
halter73: Spelling

src/Microsoft.AspNetCore.Server.Kestrel/Http/Frame.cs
```diff
@@ -248,23 +259,43 @@ public void Reset()
             _abortedCts = null;
         }
 
-        protected void ResetComponents()
+        protected void ResetComponents(bool immediateReuse)
         {
-            var frameHeaders = Interlocked.Exchange(ref _frameHeaders, null);
-            if (frameHeaders != null)
+            if (immediateReuse && HeaderFactory != null)
```
halter73: I'm not in love with the `HeaderFactory != null` heuristic for caching headers. I would prefer this to be a first class option if it should configurable at all. Even if @davidfowl hates flags.

halter73:
> Haven't removed Uninitalize from the IComponent interface as its up to the factory whether its uninitializing and retaining or throwing away not the Frame.

I'm not sure I understand the design. Why woudn't frame always uninitialize before calling Factory.Dispose?

nathana1:
@halter73 Definitely agree we should find a way to keep the caching benefits.  It's such a big improvement even with some connection churn.

src/Microsoft.AspNetCore.Server.Kestrel/Http/Frame.cs
halter73: Yeah. So now I see we *are* still caching, but this way of configuring it seems bad.

benaadams:
> Why woudn't frame always uninitialize before calling Factory.Dispose?

No point in Uninitializing if its just going to GC.

src/Microsoft.AspNetCore.Server.Kestrel/Http/Frame.cs
benaadams: I think the Reset method should go via the factory and the factory should return the "new" component; as it is still making an assumption that when there is a factory it is pooling here; whereas it might be doing something else like tracking, monitoring, counting etc.

benaadams:
@halter73 Better? 

New package `Microsoft.AspNetCore.Server.Kestrel.Pools` v `0.1.0`
Only one sample app `PoolingSimpleApp` with then uses it

Whether to reset, uninitalize and store, or throw away is controlled by the Factory via changed `Dispose` method 
```csharp
void Dispose(ref T component, bool requestImmediateReuse);
```

benaadams:
@stephentoub this locking seem ok? It preforms well, but it is also forming inter-thread stack space queues, which I don't know if its sensible/protected by `finally`
https://github.com/benaadams/KestrelHttpServer/blob/pooling-to-options/src/Microsoft.AspNetCore.Server.Kestrel.Pools/Infrastructure/ComponentPool.cs#L14-L159

geoffkizer:
I just tested the DateHeaderValueManager changes by themselves and they show a nice win on TechEmpower, 5.01 -> 5.34.  I'm going to play around with it and see if I can get further wins.

Note that this means that some part of the wins from this change are due solely to the DateHeader optimization.  We should probably pull this out and measure each separately.

benaadams:
@geoffkizer by themselves https://github.com/aspnet/KestrelHttpServer/pull/810

geoffkizer:
Thanks.  FYI I'm playing with this code now (DateHeaderValueManager) to see if there are further wins to be had.  Let me know if you're looking at it also.

nathana1:
I did some more testing this morning and numbers seem in line with what we were seeing previously (I'm missing some of @geoffkizer's optimizations so they a little lower):

|Scenario|Million Reqs/sec|
|----|----|
|Baseline|4.33|
|#791 default|4.44|
|#791 max pool 256|4.75|

One thing to note though is when I set the pool size to zero:

options.StreamFactory = new StreamFactory(MaxPooled: 0); 

It doesn't appear the headers/streams get reset properly, they are read only so throw when attempting to write to them again.  I'm seeing an exception thrown here:  https://github.com/benaadams/KestrelHttpServer/blob/pooling-to-options/src/Microsoft.AspNetCore.Server.Kestrel/Http/FrameHeaders.cs#L32

benaadams:
@nathana1 should be fixed; rebased on #810 rather than including Date changes in this PR

benaadams:
Rebased on newer #810; updated summary

benaadams:
Closed re: #851

PR #793, opened 2016-05-03
Don't reset connection info via interface

benaadams:
A not insignificant time in frame reset is spent resetting the connection addresses via interface when in all likelyhood they have not changed.

reset connection info

This resets them via class variable and only does the reset if they have changed

reset connection after

Resolves https://github.com/aspnet/KestrelHttpServer/issues/819

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Kestrel/Http/Frame.FeatureCollection.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/Http/Frame.FeatureCollection.cs
@@ -11,7 +11,6 @@ using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.Http.Features;
-using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Primitives;
 
 namespace Microsoft.AspNetCore.Server.Kestrel.Http
@@ -265,15 +264,35 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Http
 
         int IFeatureCollection.Revision => _featureRevision;
 
-        IPAddress IHttpConnectionFeature.RemoteIpAddress { get; set; }
+        IPAddress IHttpConnectionFeature.RemoteIpAddress
+        {
+            get { return RemoteIpAddress; }
+            set { RemoteIpAddress = value; }
+        }
 
-        IPAddress IHttpConnectionFeature.LocalIpAddress { get; set; }
+        IPAddress IHttpConnectionFeature.LocalIpAddress
+        {
+            get { return LocalIpAddress; }
+            set { LocalIpAddress = value; }
+        }
 
-        int IHttpConnectionFeature.RemotePort { get; set; }
+        int IHttpConnectionFeature.RemotePort
+        {
+            get { return RemotePort; }
+            set { RemotePort = value; }
+        }
 
-        int IHttpConnectionFeature.LocalPort { get; set; }
+        int IHttpConnectionFeature.LocalPort
+        {
+            get { return LocalPort; }
+            set { LocalPort = value; }
+        }
 
-        string IHttpConnectionFeature.ConnectionId { get; set; }
+        string IHttpConnectionFeature.ConnectionId
+        {
+            get { return ConnectionIdFeature; }
+            set { ConnectionIdFeature = value; }
+        }
 
         object IFeatureCollection.this[Type key]
         {
--- a/src/Microsoft.AspNetCore.Server.Kestrel/Http/Frame.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/Http/Frame.cs
@@ -5,6 +5,7 @@ using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
+using System.Net;
 using System.Numerics;
 using System.Text;
 using System.Threading;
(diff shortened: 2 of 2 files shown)


halter73:
I don't think the 2.5% gain on just the Reset method doesn't seem worth the extra complexity even if the complexity is minor

 If it's the interface that makes it inefficient, I think it would be more straightforward to put the Remot/Local properties in `Frame.cs` itself like we already do for almost all `IHttp(Request|Response)Feature`s. I'm not sure why  IHttpConnectionFeature.ConnectionId isn't already implemented as a pass through property.

benaadams:
Its so it can be user overriden, then restored from the underlying context values on reset

halter73:
you can still do that if you put the properties in `Frame.cs`.

benaadams:
@halter73 more like now?

src/Microsoft.AspNetCore.Server.Kestrel/Http/Frame.FeatureCollection.cs
```diff
@@ -264,15 +264,15 @@ bool IHttpUpgradeFeature.IsUpgradableRequest
...
+        IPAddress LocalIpAddress { get; set; }
 
-        int IHttpConnectionFeature.RemotePort { get; set; }
+        int RemotePort { get; set; }
 
-        int IHttpConnectionFeature.LocalPort { get; set; }
+        int LocalPort { get; set; }
 
-        string IHttpConnectionFeature.ConnectionId { get; set; }
+        new string ConnectionId { get; set; }
```
halter73: Yikes. I would duplicate the property and have the new properties in Frame.cs pass through to here like we do for `IHttp(Request|Response)Features`.

This class only contains explicit interface implementations.

src/Microsoft.AspNetCore.Server.Kestrel/Http/Frame.cs
```diff
@@ -234,14 +234,13 @@ public void Reset()
...
-            httpConnectionFeature.RemotePort = RemoteEndPoint?.Port ?? 0;
+            RemoteIpAddress = RemoteEndPoint?.Address;
+            RemotePort = RemoteEndPoint?.Port ?? 0;
 
-            httpConnectionFeature.LocalIpAddress = LocalEndPoint?.Address;
-            httpConnectionFeature.LocalPort = LocalEndPoint?.Port ?? 0;
+            LocalIpAddress = LocalEndPoint?.Address;
+            LocalPort = LocalEndPoint?.Port ?? 0;
 
-            httpConnectionFeature.ConnectionId = ConnectionId;
```
halter73: I now see what you're saying about it needing the interface properties to be settable. I think it only really messes with `ConnectionId` since we have the other original value in `RemoteEndPoint` and `LocalEndPoint`.

We could have a private field like `_originalConnectionId`, but I don't like it. Do you know how the perf of this iteration of the PR compares to the first iteration? I'm not really sure we need to change this code at all. 2.5% of `Reset` doesn't seem that bad.

benaadams: Its 2.5 x 5 so 12% of reset

benaadams: Its 2.5 x 5 so 12% of reset and the reset with the "bad" pooling

halter73: Ok. Then I think the way to avoid the the virtual set, is to put all the `IHttpConnectionFeature` related auto getter-setter properties in Frame.cs, and put pass through interface-specific properties in Frame.FeatureCollection.cs.

We will need a new property in `Frame.cs` called something like `ConnectionIdFeature`. E.g:

        string IHttpConnectionFeature.ConnectionId
        {
            get
            {
                return ConnectionIdFeature ?? ConnectionId;
            }

            set
            {
                ConnectionIdFeature = value;
            }
        }
 
Then you can just set `ConnectionIdFeature` to null in `Reset()`.

halter73:
I still don't like `_connectionInfoChanged`. If it's the virtual invocation of the setter that's slow, just make all the fields you added properties with the same name the have on the interface then always reset them.

`ConnectionId` is the only naming conflict, so come up with some other name for the one we use to back the feature collection like the `ConnectionIdFeature` I suggested before. That's certainly no less confusing that the `ConnectionId` vs `_connectionId` this PR currently has. That's bound to mess up someone in the future who wants to use the connection id inside Frame.

This is exactly what we do for IHttpRequestFeature/IHttpResponseFeature properties. If assigning to auto properties and using pass-through logic in the explicit interface implementation of the properties is too slow, we should look at fixing just about everything in Frame.FeatureCollection.cs. Surely 4 regular assignments in reset can't be that expensive, right?

benaadams:
Was checking out if they could all the nulls could moved to a struct and reset all at once with `= default(struct)`; got distracted by trying to work out what the GC was doing.

But... getting quite weird, probably not worth it.

halter73:
Yeah. My main desire for this change is not to do anything weird. I like picking a simple and efficient pattern and using that for all cases unless there is a big perf impact. Assigning via an interface was already weird, so I don't mind changing it, but I don't want anything remotely exotic.

benaadams:
@halter73 updated

halter73:
Thanks!

PR #800, opened 2016-05-04
Always check if block is last in linked list before consuming it

### LINKED ISSUE #782: Race causing request corruption/data loss

halter73:
- This prevents a race where we could read into the "Next" Block without
  completely consuming the previous one.

Another attempt to fix #782 that doesn't risk leasing a new block per byte in a request like #783 could in the pathological case.

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Kestrel/Infrastructure/MemoryPoolIterator.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/Infrastructure/MemoryPoolIterator.cs
@@ -4,6 +4,7 @@
 using System;
 using System.Diagnostics;
 using System.Numerics;
+using System.Threading;
 
 namespace Microsoft.AspNetCore.Server.Kestrel.Infrastructure
 {
@@ -767,7 +768,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Infrastructure
                 {
                     var nextBlock = pool.Lease();
                     block.End = blockIndex;
-                    block.Next = nextBlock;
+                    Volatile.Write(ref block.Next, nextBlock);
                     block = nextBlock;
 
                     blockIndex = block.Data.Offset;
@@ -820,7 +821,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Infrastructure
                     {
                         var nextBlock = pool.Lease();
                         block.End = blockIndex;
-                        block.Next = nextBlock;
+                        Volatile.Write(ref block.Next, nextBlock);
                         block = nextBlock;
 
                         blockIndex = block.Data.Offset;
--- a/src/Microsoft.AspNetCore.Server.Kestrel/Infrastructure/MemoryPoolIterator.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/Infrastructure/MemoryPoolIterator.cs
@@ -68,6 +68,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Infrastructure
             }
 
             var index = _index;
+            var wasLastBlock = block.Next == null;
 
             if (index < block.End)
             {
@@ -77,7 +78,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Infrastructure
 
             do
             {
-                if (block.Next == null)
+                if (wasLastBlock)
                 {
                     return -1;
                 }
@@ -86,6 +87,8 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Infrastructure
                     block = block.Next;
                     index = block.Start;
                 }
+
+                wasLastBlock = block.Next == null;
 
                 if (index < block.End)
                 {
@@ -102,7 +105,10 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Infrastructure
             {
                 return;
             }
+
+            var wasLastBlock = _block.Next == null;
             var following = _block.End - _index;
+
             if (following >= bytesToSkip)
             {
                 _index += bytesToSkip;
@@ -113,7 +119,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Infrastructure
             var index = _index;
             while (true)
             {
-                if (block.Next == null)
+                if (wasLastBlock)
                 {
                     return;
                 }
@@ -123,7 +129,10 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Infrastructure
                     block = block.Next;
                     index = block.Start;
                 }
+
+                wasLastBlock = block.Next == null;
                 following = block.End - index;
+
                 if (following >= bytesToSkip)
                 {
                     _block = block;
@@ -141,6 +150,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Infrastructure
                 return -1;
             }
 
+            var wasLastBlock = _block.Next == null;
             var index = _index;
 
             if (index < block.End)
@@ -150,7 +160,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Infrastructure
 
             do
             {
-                if (block.Next == null)
+                if (wasLastBlock)
                 {
                     return -1;
                 }
@@ -159,6 +169,8 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Infrastructure
                     block = block.Next;
                     index = block.Start;
                 }
+
+                wasLastBlock = block.Next == null;
 
                 if (index < block.End)
                 {
@@ -173,14 +185,17 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Infrastructure
             {
                 return -1;
             }
-            else if (_block.End - _index >= sizeof(long))
+
+            var wasLastBlock = _block.Next == null;
+
+            if (_block.End - _index >= sizeof(long))
             {
                 fixed (byte* ptr = &_block.Array[_index])
                 {
                     return *(long*)(ptr);
                 }
             }
-            else if (_block.Next == null)
+            else if (wasLastBlock)
             {
                 return -1;
             }
@@ -219,6 +234,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Infrastructure
 
             var block = _block;
             var index = _index;
+            var wasLastBlock = block.Next == null;
             var following = block.End - index;
             byte[] array;
             var byte0 = byte0Vector[0];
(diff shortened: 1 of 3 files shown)


halter73:
> dotnet run -- -n plaintext -s http://adx-3u-perfsvr:5001 -c http://adx-3u-svr02:5002 -o KestrelHttpServer@release
    Project Repository (.NETStandard,Version=v1.0) was previously compiled. Skipping compilation.
    Project Benchmarks.ClientJob (.NETStandard,Version=v1.0) was previously compiled. Skipping compilation.
    Project Benchmarks.ServerJob (.NETStandard,Version=v1.0) was previously compiled. Skipping compilation.
    Project BenchmarksDriver (.NETCoreApp,Version=v1.0) was previously compiled. Skipping compilation.
    [04:58:02.797] Starting scenario Plaintext on benchmark server...
    [04:58:26.821] Warmup
    [04:58:26.841] Starting scenario Plaintext on benchmark client...
    [04:58:38.032] Scenario Plaintext completed on benchmark client
    [04:58:38.032] RPS: 989547.57
    [04:58:38.032] Stopping scenario Plaintext on benchmark client...
    [04:58:38.035] Benchmark
    [04:58:38.035] Starting scenario Plaintext on benchmark client...
    [04:58:49.082] Scenario Plaintext completed on benchmark client
    [04:58:49.082] RPS: 1008641.1
    [04:58:49.082] Stopping scenario Plaintext on benchmark client...
    [04:58:49.085] Stopping scenario Plaintext on benchmark server...
    > dotnet run -- -n plaintext -s http://adx-3u-perfsvr:5001 -c http://adx-3u-svr02:5002 -o KestrelHttpServer@halter73/the-first-shall-be-last
    Project Repository (.NETStandard,Version=v1.0) was previously compiled. Skipping compilation.
    Project Benchmarks.ClientJob (.NETStandard,Version=v1.0) was previously compiled. Skipping compilation.
    Project Benchmarks.ServerJob (.NETStandard,Version=v1.0) was previously compiled. Skipping compilation.
    Project BenchmarksDriver (.NETCoreApp,Version=v1.0) was previously compiled. Skipping compilation.
    [04:59:10.461] Starting scenario Plaintext on benchmark server...
    [04:59:33.296] Warmup
    [04:59:33.346] Starting scenario Plaintext on benchmark client...
    [04:59:44.461] Scenario Plaintext completed on benchmark client
    [04:59:44.461] RPS: 965932.7
    [04:59:44.461] Stopping scenario Plaintext on benchmark client...
    [04:59:44.464] Benchmark
    [04:59:44.464] Starting scenario Plaintext on benchmark client...
    [04:59:55.537] Scenario Plaintext completed on benchmark client
    [04:59:55.537] RPS: 1004270.23
    [04:59:55.537] Stopping scenario Plaintext on benchmark client...
    [04:59:55.540] Stopping scenario Plaintext on benchmark server...

Eilon:
@halter73 I don't suppose there's a way to add a unit test for this? At least one that might crash if it's broken, and otherwise "do nothing bad" and thus succeed?

davidfowl:
You would have to add test hooks to write something useful.

davidfowl:
@halter73 did you run it in release mode?

jaredpar:
I do not think this fixes the problem.  I'm not 100% confident on the state of the .Net memory model but if memory serves it does not guarantee ordering between writes to different locations (CLI spec certainly does not give this guarantee).  Hence it's possible to see writes is one thread which have an order in a completely different order on another thread.  

This means it's possible for situations like the following to occur:

- Producer moves `End` from 13 to 42, then sets `Next` to a non-null value.
- Consumer sees `End` as 13 and `Next` as non-null.  

If this happens the code in `Take` will lose data.  

I believe the only way to solve this is to use a memory barrier to give ordering guarantees between threads.  For example add a full barrier on write of the `Next` value.  Then on the consumer you can make guarantees like

> If Next is not null then End has a sane value.

It's been a while since I dealt with the low level details of the memory model though.  I would defer to @stephentoub opinion over mine here.

src/Microsoft.AspNetCore.Server.Kestrel/Infrastructure/MemoryPoolIterator.cs
```diff
@@ -68,6 +68,7 @@ public int Take()
             }
 
             var index = _index;
+            var isLastBlock = block.Next == null;
```
jaredpar: Nit: you should rename this variable to "wasLastBlock".  The prefix "is" implies something which is true / false at this very moment.  That is not the case here.  At best you can say that at a previous point in time this was the last block.

stephentoub:
Yeah, this definitely needs some volatile love. https://github.com/dotnet/corefxlab/blob/master/src/System.Threading.Tasks.Channels/System/Collections/Concurrent/SingleProducerConsumerQueue.cs is used in a few places in coreclr/corefx where a fast single-producer single-consumer queue is needed, and you can see how it takes advantage of memory barriers.

halter73:
@stephentoub @jaredpar Thanks for taking a look and for the link to SingleProducerConsumerQueue. We'll certainly take a look to see if we can reuse some of that. SegmentState and the way it's used is particularly interesting.

You can see the most recent commit makes MemoryPoolBlock.End volatile and uses Volatile.Write for MemoryPoolBlock.Next to create a memory barrier. I think this is sufficient to prevent the consumer from potentially skipping data. The perf impact also seems negligible as judged by our plaintext benchmark.

Does this look good?

@davidfowl As I mentioned in person, the benchmarks are run in release mode.

@Eilon @davidfowl Is correct that to force a race would require adding test hooks that would allow you to block the consumer at a the critical moment (which doesn't exist anymore in theory). If you look at #782, you'll see a test that was failing about ~5% of the time due to this bug.

halter73:
> dotnet run -- -n plaintext -s http://adx-3u-perfsvr:5001 -c http://adx-3u-svr02:5002 -o KestrelHttpServer@release
    Project Repository (.NETStandard,Version=v1.0) was previously compiled. Skipping compilation.
    Project Benchmarks.ClientJob (.NETStandard,Version=v1.0) was previously compiled. Skipping compilation.
    Project Benchmarks.ServerJob (.NETStandard,Version=v1.0) was previously compiled. Skipping compilation.
    Project BenchmarksDriver (.NETCoreApp,Version=v1.0) was previously compiled. Skipping compilation.
    [03:08:05.815] Starting scenario Plaintext on benchmark server...
    [03:08:28.604] Warmup
    [03:08:28.628] Starting scenario Plaintext on benchmark client...
    [03:08:39.416] Scenario Plaintext completed on benchmark client
    [03:08:39.416] RPS: 985183.15
    [03:08:39.416] Stopping scenario Plaintext on benchmark client...
    [03:08:39.428] Benchmark
    [03:08:39.428] Starting scenario Plaintext on benchmark client...
    [03:08:49.809] Scenario Plaintext completed on benchmark client
    [03:08:49.809] RPS: 1008523.2
    [03:08:49.809] Stopping scenario Plaintext on benchmark client...
    [03:08:49.827] Stopping scenario Plaintext on benchmark server...
    > dotnet run -- -n plaintext -s http://adx-3u-perfsvr:5001 -c http://adx-3u-svr02:5002 -o KestrelHttpServer@halter73/the-first-shall-be-last
    Project Repository (.NETStandard,Version=v1.0) was previously compiled. Skipping compilation.
    Project Benchmarks.ClientJob (.NETStandard,Version=v1.0) was previously compiled. Skipping compilation.
    Project Benchmarks.ServerJob (.NETStandard,Version=v1.0) was previously compiled. Skipping compilation.
    Project BenchmarksDriver (.NETCoreApp,Version=v1.0) was previously compiled. Skipping compilation.
    [03:09:15.077] Starting scenario Plaintext on benchmark server...
    [03:09:38.317] Warmup
    [03:09:38.400] Starting scenario Plaintext on benchmark client...
    [03:09:48.733] Scenario Plaintext completed on benchmark client
    [03:09:48.733] RPS: 990907.41
    [03:09:48.733] Stopping scenario Plaintext on benchmark client...
    [03:09:48.736] Benchmark
    [03:09:48.736] Starting scenario Plaintext on benchmark client...
    [03:09:59.794] Scenario Plaintext completed on benchmark client
    [03:09:59.794] RPS: 1008054.57
    [03:09:59.795] Stopping scenario Plaintext on benchmark client...
    [03:09:59.798] Stopping scenario Plaintext on benchmark server...

Eilon:
@halter73 thanks, I'm not looking at add a test at all costs, so if there's no way to do it without completely artificial hooks let's not do it.

src/Microsoft.AspNetCore.Server.Kestrel/Infrastructure/MemoryPoolIterator.cs
```diff
@@ -605,24 +624,21 @@ public bool Put(byte data)
             {
                 return false;
             }
-            else if (_index < _block.End)
```
davidfowl: Explanation?

davidfowl: Ah I see it

PR #801, opened 2016-05-05
Shrink SocketInput lock

benaadams:
@halter73 can this be done or will it cause problems?

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Kestrel/Http/SocketInput.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/Http/SocketInput.cs
@@ -82,9 +82,9 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Http
                 {
                     RemoteIntakeFin = true;
                 }
-
-                Complete();
             }
+
+            Complete();
         }
 
         public void IncomingComplete(int count, Exception error)
@@ -120,9 +120,9 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Http
                 {
                     _awaitableError = error;
                 }
-
-                Complete();
             }
+
+            Complete();
         }
 
         public void IncomingDeferred()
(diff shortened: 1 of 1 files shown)


src/Microsoft.AspNetCore.Server.Kestrel/Http/SocketInput.cs
```diff
@@ -120,9 +120,9 @@ public void IncomingComplete(int count, Exception error)
                 {
                     _awaitableError = error;
                 }
-
-                Complete();
             }
+
+            Complete();
```
halter73: Complete cannot be moved outside of the lock because it sets `_awaitableState` to `_awaitableIsCompleted`. Somewhat paradoxically setting this state outside the lock can cause the consuming thread to stall.

The above code inside the lock advances the tail which can change the result of `examined.IsEnd` (which must also be checked inside the lock). This may cause `ConsumingComplete` to incorrectly set the `_awaitableState` to `_awaitableIsNotCompleted` when there is really more data to consume which in turn causes the following `await SocketInput` to stall.

This is [exactly what happened](https://github.com/aspnet/KestrelHttpServer/pull/686/commits/84a68208d0b32d227a547daa32d2b25cad48ee92) the [first time](https://github.com/aspnet/KestrelHttpServer/pull/525/commits/afe944c053c6c127623f5996c19a99e9fdd1f4ad) we tried to remove the locks in SocketInput. I'm now very weary of reducing the locking in this class.

There are some things that could obviously be removed from the lock with a little refactoring like actual calls to `_threadPool.Run`, but does that actually have a measurable perf impact making it worth bifurcating `SocketInput.Complete`?

benaadams: Thought it couldn't :( 

I'm having trouble maxing CPU and `IncomingData` and `ConsumingComplete` are taking a much larger proportion of time then I remember - but its been a while; which would suggest locks.

Related then in `ConsumingComplete` can the interlocked `_consumingState` and block return be moved out of the lock?

benaadams: @halter73 which would be https://github.com/aspnet/KestrelHttpServer/pull/807

PR #803, opened 2016-05-05
narrow lock scope when connections end

### LINKED ISSUE #802: lock in Connection.End is potentially too broad
This lock:  https://github.com/aspnet/KestrelHttpServer/blob/dev/src/Microsoft.AspNetCore.Server.Kestrel/Http/Connection.cs#L331
doesn't appear necessary in the keep alive case.  Scoping it to just the disconnecting case gives a small bump on our techempower machines:

|Scenario|Million Reqs/sec|
|---|---|
|Baseline|4.5|
|Optimized|4.6|

nathana1:
Resolves #802.

Locks just the disconnect case.

/cc @halter73 @mikeharder @cesarbs

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Kestrel/Http/Connection.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/Http/Connection.cs
@@ -328,20 +328,20 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Http
 
         void IConnectionControl.End(ProduceEndType endType)
         {
-            lock (_stateLock)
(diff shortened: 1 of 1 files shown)


halter73:
This seems safe considering that this only really affects calling `End` with `ProduceEndType.ConnectionKeepAlive`. And we only check the `_connectionState` to determine whether or not to log "Connection id ""{ConnectionId}"" completed keep alive response."

I'm surprised there is a measurable perf distance as there should be no contention aside from during connection initialization/abort.

PR #807, opened 2016-05-06
Shrink SocketInput lock

benaadams:
Reduce work in lock.

Follow up to #801


benaadams:
Makes LargeFileUpload even more grumpy :(

PR #808, opened 2016-05-06
Strong locking in SocketInput

benaadams:
Not happy with this but it should fix the corruption issues.

Reviewing to see if can consolidate things and improve perf

Might work as an RC2 stop gap?

/cc @halter73

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Kestrel/Http/Connection.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/Http/Connection.cs
@@ -270,7 +270,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Http
                 // there is no data to be read right now.
                 // See the note at http://docs.libuv.org/en/v1.x/stream.html#c.uv_read_cb.
                 // We need to clean up whatever was allocated by OnAlloc.
-                _rawSocketInput.IncomingDeferred();
+                _rawSocketInput.ReturnSocketBlock();
                 return;
             }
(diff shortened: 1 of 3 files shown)


halter73:
I don't think this PR is a good candidate for RC2 because I don't think this is actually any safer than #800. At least #800 is a clear improvement over what we had in RC1, whereas this is basically a rewrite of SocketInput with all the risk that entails.

Why do you think this is an improvement over #800? If that doesn't address the corruptions issues we should address those shortcomings in a more targeted way.

If I'm understanding this correctly, this PR seems to add a bunch complexity to SocketInput just to ensure that the linked list being consumed doesn't get appended to between ConsumingStart and ConsumingComplete. It's an interesting approach, but I think that in and of itself is likely to cause at least a minor perf hit since it's bound to cause more calls to ConsumingStart and ConsumingComplete which take a high contention lock. ConsumingStart didn't even use to take a lock, but now it has to.

Considering all the code that traverses the linked list exists in MemoryPoolIterator, I think it's better just fix the traversing logic. Using a volatile End pointer and Volatile.Write for MemoryPoolBlock.Next has a negligible perf impact.

halter73:
This is what's causing the AppVeyor failure for this PR:

This is the exact same test failure that lead me to filing #782. I can't be certain that it's the same root cause, but it is worrisome.

benaadams:
@halter73 I think its worse than #800; just was trying a less exotic approach :)

PR #810, opened 2016-05-07
Only call PrepareDateValues if not already called for tick

benaadams:
From https://github.com/aspnet/KestrelHttpServer/pull/791

Resolves https://github.com/aspnet/KestrelHttpServer/issues/818
Resolves https://github.com/aspnet/KestrelHttpServer/issues/821


benaadams:
@geoffkizer also added a manual inline test

davidfowl:
?? Is this really better? Why are the 2 older methods left over?

benaadams:
From https://github.com/aspnet/KestrelHttpServer/pull/791#issuecomment-217589238
> I just tested the DateHeaderValueManager changes by themselves and they show a nice win on TechEmpower, 5.01 -> 5.34. I'm going to play around with it and see if I can get further wins.

> Why are the 2 older methods left over?

Backwards compat?

geoffkizer:
Yeah, it's definitely better.  All the volatile reads/writes are expensive and reducing this to one call avoids a bunch of them.

That said, we're still doing 4 volatile reads and 1 volatile write here.  I prototyped reducing this to a single volatile read and it basically made the time spent here go to zero and gave another small win (2-3%).

I still need to fix up the timer/dispose logic.

The older methods can go away.

benaadams:
@geoffkizer bear in mind it doesn't need to be super accurate; it needs to not be broken e.g. using a half written byte array. For its use if the date header is off by a few seconds it doesn't hugely matter for how its used. Obviously I'm not suggesting slowing down the timer to under a second.

src/Microsoft.AspNetCore.Server.Kestrel/Http/DateHeaderValueManager.cs
```diff
@@ -70,6 +70,21 @@ public byte[] GetDateHeaderValueBytes()
             return _activeDateBytes ? _dateBytes0 : _dateBytes1;
         }
```
halter73: Remove the old methods. There should only be one true way.

src/Microsoft.AspNetCore.Server.Kestrel/Http/DateHeaderValueManager.cs
```diff
@@ -70,6 +70,21 @@ public byte[] GetDateHeaderValueBytes()
...
 
+        public DateHeaderValues GetDateHeaderValues()
+        {
+            _hadRequestsSinceLastTimerTick = true;
+            if (!_timerIsRunning)
```
halter73: Doesn't `PrepareDateValues` check this already?

halter73:
Impressive how much of a perf impact this has. It really shows how expensive volatile read/writes can be.

benaadams:
@geoffkizer I assume you are continuing this patch? Or should I remove the old methods?

benaadams:
Ok, cleaning it up...

Have to run cmd line tools to see if tests don't compile :-/

benaadams:
Cut down the checks. Could someone check the logic? May have gone too far...

geoffkizer:
This looks quite good performance-wise.  Removing the volatile write in the common case seems to resolve the major perf issue here.

I think we should merge this.

src/Microsoft.AspNetCore.Server.Kestrel/Http/DateHeaderValueManager.cs
```diff
@@ -55,19 +55,19 @@ public DateHeaderValueManager()
...
+            }
 
-        public byte[] GetDateHeaderValueBytes()
-        {
-            _hadRequestsSinceLastTimerTick = true;
-            PrepareDateValues();
-            return _activeDateBytes ? _dateBytes0 : _dateBytes1;
+            return new DateHeaderValues()
+            {
+                Bytes = _activeDateBytes ? _dateBytes0 : _dateBytes1,
```
stephentoub: Not specific to this change, but I'm unclear how the current usage of _dateBytes0 and _dateBytes1 is safe.  It appears that the code is trying to use one of the arrays as the current known-good and the other array as a scratch, and they then get switched when the scratch is updated.  But what's to prevent one thread from getting handed one of the arrays, that thread getting swapped out, lots of stuff happening, and then that thread getting resumed as another thread is encoding a new value into that same array?  Do we know that the encoding process is going to leave the data in the array in a non-corrupted state from the perspective of having valid data?

stephentoub: Assuming it's fine, you might be able to eliminate a read and a branch here by having the updater set an _activeBytes field to either _dateBytes0 or _dateBytes1, such that the reader can then just do ```Bytes = _activeBytes```.  Assuming reading is super hot and updating is rare, avoiding branches and extra reads on the super hot path could be beneficial.

benaadams: Reading is very hot; updating is once per second. As writing is very slow; probably not saving much from not allocating as its only a short array - though at high throughput each one is likely to hit a higher GC generation. 

Might be safest to allocate on each write? As some requests may take seconds or minutes depending on db access etc.

benaadams: Also its doing it for the string anyway

benaadams:
@stephentoub latest commit address your concerns?

src/Microsoft.AspNetCore.Server.Kestrel/Http/DateHeaderValueManager.cs
```diff
@@ -17,10 +17,9 @@ public class DateHeaderValueManager : IDisposable
         private readonly TimeSpan _timeWithoutRequestsUntilIdle;
         private readonly TimeSpan _timerInterval;
 
-        private volatile string _dateValue;
-        private volatile bool _activeDateBytes;
-        private readonly byte[] _dateBytes0 = Encoding.ASCII.GetBytes("\r\nDate: DDD, dd mmm yyyy hh:mm:ss GMT");
-        private readonly byte[] _dateBytes1 = Encoding.ASCII.GetBytes("\r\nDate: DDD, dd mmm yyyy hh:mm:ss GMT");
+        private readonly byte[] _datePreambleBytes = Encoding.ASCII.GetBytes("\r\nDate: ");
```
stephentoub: static?

src/Microsoft.AspNetCore.Server.Kestrel/Http/DateHeaderValueManager.cs
```diff
@@ -65,7 +64,7 @@ public DateHeaderValues GetDateHeaderValues()
 
             return new DateHeaderValues()
             {
-                Bytes = _activeDateBytes ? _dateBytes0 : _dateBytes1,
+                Bytes = _dateBytes,
```
stephentoub: Previously _activeDateBytes was volatile, so there was a volatile read prior to reading _dateBytes0 or _dateBytes1 and a volatile write after updating them.  Does _dateBytes need to be volatile?

src/Microsoft.AspNetCore.Server.Kestrel/Http/DateHeaderValueManager.cs
```diff
@@ -17,10 +17,9 @@ public class DateHeaderValueManager : IDisposable
...
         private readonly TimeSpan _timerInterval;
 
-        private volatile string _dateValue;
-        private volatile bool _activeDateBytes;
-        private readonly byte[] _dateBytes0 = Encoding.ASCII.GetBytes("\r\nDate: DDD, dd mmm yyyy hh:mm:ss GMT");
-        private readonly byte[] _dateBytes1 = Encoding.ASCII.GetBytes("\r\nDate: DDD, dd mmm yyyy hh:mm:ss GMT");
+        private readonly byte[] _datePreambleBytes = Encoding.ASCII.GetBytes("\r\nDate: ");
+        private string _dateValue;
+        private byte[] _dateBytes;
         private object _timerLocker = new object();
```
stephentoub: readonly?

src/Microsoft.AspNetCore.Server.Kestrel/Http/DateHeaderValueManager.cs
benaadams: It should always read a value? For its use case (browser/proxy cache age) it doesn't matter if its stale for up to 2 seconds

stephentoub: Ok.  The write to it should probably at least be made volatile, e.g. ```Volatile.Write(ref _dateBytes, dateBytes)```.

benaadams: Wouldn't that be the same issue with the string? Should that be made a `Volatile.Write`?

stephentoub: In theory.  In practice the volatile write(s) probably doesn't matter, as I've no doubt there are other operations that are going to force coherency.

benaadams: Updated both anyway :)

benaadams: Though the second would probably protect the first

stephentoub:
LGTM

benaadams:
Squashed

src/Microsoft.AspNetCore.Server.Kestrel/Http/DateHeaderValueManager.cs
geoffkizer: It seems like it's possible (though very unlikely) that you could get mismatching values for the bytes and the string, no?

Why don't we make DateHeaderValues a class to deal with this?  Slightly less work in GetDateHeaderValues too.

stephentoub: > Why don't we make DateHeaderValues a class to deal with this?

Makes sense to me.  Then GetDateHeaderValues can just ```return _headerValues;```

benaadams: > It seems like it's possible (though very unlikely) that you could get mismatching values for the bytes and the string, no?

They could differ by 1 sec every 1 sec for a window of ~100ns. Only the byte representation is actually used; except if someone creates a middleware to read it internally for some reason. 

However for timings and diagnostics you would much more likely use the more accurate [timestamps](https://github.com/aspnet/Hosting/blob/b7bdc9c40494f8e0f0eac22db91b8d5c58811ee2/src/Microsoft.AspNetCore.Hosting/Internal/HostingApplication.cs#L37-L44) which give sub millisecond accuracy rather than 1sec +/- 1sec accuracy.

Use of the date header for caching/proxy/age normally is for time periods of between 5 min and 1 year so the accuracy isn't too huge a factor.

Changing it to a class would add an extra level of indirection per request; and an extra allocation (though only once per second!).

Happy to change it if there is a strong opinion on it.

benaadams: It could be important for digest header signing I suppose?

benaadams: It could be important for digest header signing I suppose? Though then you are also at the mercy of the output order with some of the headers being in a dictionary can coming out in an unknown order.

benaadams: Re less work in `GetDateHeaderValues` could class struct it to copy rather than create? Not sure costs https://github.com/benaadams/KestrelHttpServer/commit/3a992bd55b2a56e9e8ab81a4dabbf31b579e217a

benaadams: @geoffkizer @stephentoub long story short.. it is now a class and should now update them both simultaneously :grinning:

halter73:
LGTM. Can we get benchmark results on the latest incarnation of just this PR?

src/Microsoft.AspNetCore.Server.Kestrel/Http/DateHeaderValueManager.cs
halter73: Nice.

benaadams:
Switched DateHeaderValues from struct to class to allow simultaneous update of sting and byte versions

PR #804, opened 2016-05-05
Use Unsafe.Read<> for faster Vector ctors

benaadams:
Needs `System.Runtime.CompilerServices.Unsafe` added to feed

Also using Vector for validation of longer headers which is good for `Content-Security-Policy` and `Cookie`

Issue #795, opened 2016-05-03
improve performance of known headers


nathana1:
When I remove most of the headers from KnownHeaders.cs we see a good throughput increase:

|Scenario|Reqs/sec|
|---|---|
|Baseline| 1009999 |
|IOCP| 1114899 |
|Cache Headers/Streams| 1157813 |
|Known header removal| 1189857 |
|All 3| 1449341 |

Need to find the right headers to remove or other techniques to realize this gain.  See https://github.com/aspnet/KestrelHttpServer/commit/f40f4fbf069ecc7fc2bf2ddfbae3f6a3f4ff8757 for prototype code.

benaadams:
wrk only includes Host header, which wouldn't match real world use.

Would outputting the headers in a particular order help? E.g. always have Host as the first check for request?

geoffkizer:
There are a ton of known headers, and it's not clear we really should predefine these, since there's some cost to it.  We should at least look at removing some of the pretty uncommon ones.

nathana1:
I was able to taking a closer at the profiles and there's 2 primary issues:

1. FrameHeaders objects are huge so they take time to allocate, we are also addressing this with the pooling/caching work.
2. CopyToFast isn't that fast when we are creating response headers probably because there are so many different headers it checks for.

One approach is to be slightly less draconian than my prototype linked above (reinclude the ones @benaadams listed and the ones kestrel actually uses).  I'll also take a look for other approaches.

geoffkizer:
I think there's one other potential issue too:  Even with pooling, Reset on the Headers objects are expensive because they are so large.  Nathan do you see that in your profiles?

nathana1:
Yeah, it does seem to be smaller than the other 2 though.

nathana1:
I hoped to get an additional improvement from #812 since CopyToFast spends half its time doing bitwise comparisons and the other half block copies but I'm not seeing any perf difference between safe and unsafe whether I include the header trim or not.

benaadams:
@nathana1 if you are reducing them, I might have something...

benaadams:
Known headers have known offsets; have a look at /MemoryPoolIterator2.Generated.cs https://github.com/benaadams/KestrelHttpServer/commit/f09696445822c2d0c6337735ccb8a2f57ba29d0a

benaadams:
nvm CopyToFast is output not input

muratg:
[...]

PR #812, opened 2016-05-07
[Blocked] Use faster Unsafe.Read/CopyBlock methods

benaadams:
Needs `System.Runtime.CompilerServices.Unsafe` added to feed

Supersedes #804

Blocked on https://github.com/aspnet/Home/issues/1428

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Kestrel/Infrastructure/MemoryPoolIterator.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/Infrastructure/MemoryPoolIterator.cs
@@ -4,6 +4,7 @@
 using System;
 using System.Diagnostics;
 using System.Numerics;
+using System.Runtime.CompilerServices;
 
 namespace Microsoft.AspNetCore.Server.Kestrel.Infrastructure
 {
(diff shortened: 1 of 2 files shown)


geoffkizer:
Per @nathana1 #795, this doesn't seem to be a win.

I tried something similar with Buffer.MemoryCopy (unsafe version of BufferCopy.BlockCopy) and didn't see a win either -- actually seemed to make things slightly worse.

This is unfortunate.  I wish we could do better than BlockCopy, but not sure how to at this point...

benaadams:
`Unsafe.Read<Vector<byte>>(block.DataFixedPtr + index);` is faster than `new Vector<byte>(array, index)`

Did assume rather than check `Unsafe.CopyBlock` :(

benaadams:
`Unsafe.Read<Vector<byte>>(block.DataFixedPtr + index).CopyTo(array)` may therefore be better https://github.com/IllyriadGames/ByteArrayExtensions

However `BlockCopy` has been improved since then

geoffkizer:
Definitely worth a look...

benaadams:
Closing for now

PR #816, opened 2016-05-10
Queue SocketInput action to threadpool out of lock

benaadams:
Reduce chance of contention/cpu spin when lock hit

/cc @halter73 

Resolves https://github.com/aspnet/KestrelHttpServer/issues/817


halter73:
I don't think the logic here is wrong, but it does make calling `Complete` more involved. I don't plan to merge this unless there is a measurable impact.

halter73:
Closing for reasoning above.

PR #829, opened 2016-05-11
trim the number of known headers

### LINKED ISSUE #795: improve performance of known headers

nathana1:
Resolves #795.

There's a fairly large perf improvement by removing some of the current known headers that are probably not that common (from asp.net lab automation):

|Scenario|Million reqs/sec|
|----|----|
|Baseline|1.09|
|This PR|1.23|

Primarily this reduces the amount of memory allocated for FrameRequestHeaders and FrameResponseHeaders and shrinks FrameResponseHeaders.CopyToFast so it doesn't have to do as many comparisons.

/cc @halter73 @cesarbs @geoffkizer @benaadams

### DIFF
--- a/test/Microsoft.AspNetCore.Server.KestrelTests/ChunkedRequestTests.cs
+++ b/test/Microsoft.AspNetCore.Server.KestrelTests/ChunkedRequestTests.cs
@@ -372,10 +372,10 @@ namespace Microsoft.AspNetCore.Server.KestrelTests
                     await connection.Receive(
                         "HTTP/1.1 400 Bad Request",
                         "Connection: close",
+                        "Content-Length: 0",
                         "");
                     await connection.ReceiveStartsWith("Date:");
                     await connection.ReceiveForcedEnd(
-                        "Content-Length: 0",
                         "Server: Kestrel",
                         "",
                         "");
@@ -417,10 +417,10 @@ namespace Microsoft.AspNetCore.Server.KestrelTests
                     await connection.Receive(
                         "HTTP/1.1 400 Bad Request",
                         "Connection: close",
+                        "Content-Length: 0",
                         "");
                     await connection.ReceiveStartsWith("Date:");
                     await connection.ReceiveForcedEnd(
-                        "Content-Length: 0",
                         "Server: Kestrel",
                         "",
                         "");
--- a/test/Microsoft.AspNetCore.Server.KestrelTests/EngineTests.cs
+++ b/test/Microsoft.AspNetCore.Server.KestrelTests/EngineTests.cs
@@ -555,19 +555,20 @@ namespace Microsoft.AspNetCore.Server.KestrelTests
                         "");
                     await connection.Receive(
                         "HTTP/1.1 500 Internal Server Error",
+                        "Content-Length: 0",
                         "");
                     await connection.ReceiveStartsWith("Date:");
                     await connection.Receive(
-                        "Content-Length: 0",
                         "Server: Kestrel",
                         "",
                         "HTTP/1.1 500 Internal Server Error",
                         "");
-                    await connection.Receive("Connection: close",
+                    await connection.Receive(
+                        "Connection: close",
+                        "Content-Length: 0",
                         "");
                     await connection.ReceiveStartsWith("Date:");
                     await connection.ReceiveEnd(
-                        "Content-Length: 0",
                         "Server: Kestrel",
                         "",
                         "");
@@ -708,9 +709,9 @@ namespace Microsoft.AspNetCore.Server.KestrelTests
                         "HTTP/1.1 400 Bad Request",
                         "Connection: close",
                         "");
+                    await connection.Receive("Content-Length: 0", "");
                     await connection.ReceiveStartsWith("Date:");
                     await connection.ReceiveForcedEnd(
-                        "Content-Length: 0",
                         "Server: Kestrel",
                         "",
                         "");
@@ -729,10 +730,10 @@ namespace Microsoft.AspNetCore.Server.KestrelTests
                         "",
                         "HTTP/1.1 400 Bad Request",
                         "Connection: close",
+                        "Content-Length: 0",
                         "");
                     await connection.ReceiveStartsWith("Date:");
                     await connection.ReceiveForcedEnd(
-                        "Content-Length: 0",
                         "Server: Kestrel",
                         "",
                         "");
@@ -789,18 +790,18 @@ namespace Microsoft.AspNetCore.Server.KestrelTests
                         "");
                     await connection.Receive(
                         "HTTP/1.1 500 Internal Server Error",
+                        "Content-Length: 0",
                         "");
                     await connection.ReceiveStartsWith("Date:");
                     await connection.Receive(
-                        "Content-Length: 0",
                         "Server: Kestrel",
                         "",
                         "HTTP/1.1 500 Internal Server Error",
                         "Connection: close",
+                        "Content-Length: 0",
                         "");
                     await connection.ReceiveStartsWith("Date:");
                     await connection.ReceiveEnd(
-                        "Content-Length: 0",
                         "Server: Kestrel",
                         "",
                         "");
--- a/tools/Microsoft.AspNetCore.Server.Kestrel.GeneratedCode/KnownHeaders.cs
+++ b/tools/Microsoft.AspNetCore.Server.Kestrel.GeneratedCode/KnownHeaders.cs
@@ -77,58 +77,20 @@ namespace Microsoft.AspNetCore.Server.Kestrel.GeneratedCode
         {
             var commonHeaders = new[]
             {
-                "Cache-Control",
                 "Connection",
-                "Date",
-                "Keep-Alive",
-                "Pragma",
-                "Trailer",
                 "Transfer-Encoding",
-                "Upgrade",
-                "Via",
-                "Warning",
-                "Allow",
                 "Content-Length",
-                "Content-Type",
-                "Content-Encoding",
-                "Content-Language",
-                "Content-Location",
-                "Content-MD5",
-                "Content-Range",
-                "Expires",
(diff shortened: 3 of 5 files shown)


cesarbs:
How do you determine what a "not that common" header is? I'd argue `Set-Cookie` is a pretty common response header.

I think @benaadams will not be happy about removing `Upgrade` from the list since his app uses websockets extensively IIRC.

benaadams:
And CORSs headers - but looking at 90% of requests `Upgrade` won't be included; e.g. is for a request that may last an hour or http/2 that will cover many requests. 

`Set-Cookie` is only per N response not every response - normally just the initial connect; so not common.

Changes I'd make

`Transfer-Encoding` is iffy on request as will only be for posts and GET to POST ratio will be high; perhaps only include it on Reposnse? As MVC chunks quite a lot.
`Date` again move only to Response rather than having on both.
`Content-Length` again move only to Response rather than having on both.
`Content-Type` again move only to Response rather than having on both.

Looking at a request by Chrome `Referrer` is probably one that is missing on Request

```
GET /images/spinners/octocat-spinner-128.gif HTTP/1.1
Host: assets-cdn.github.com
Connection: keep-alive
Accept: image/webp,image/*,*/*;q=0.8
User-Agent: Mozilla/5.0 (Windows NT 10.0; WOW64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/50.0.2661.94 Safari/537.36
DNT: 1
Referer: https://github.com/dotnet/corefx/blob/master/src/System.Runtime.CompilerServices.Unsafe/src/System.Runtime.CompilerServices.Unsafe.xml
Accept-Encoding: gzip, deflate, sdch
Accept-Language: en-US,en;q=0.8
Cookie: ....
```
It would be nice to override the pre-cached list so CORS for an asset server/cdn; but that could maybe be picked up as part of the header factories in https://github.com/aspnet/KestrelHttpServer/pull/791 ?

benaadams:
`Etag` might be one to drop also as is only file serving and special pipeline/caching

Tratcher:
There's several trade-offs here:
- The size of the header structures / the number of known headers.
- What request headers are commonly received?
- What request headers are commonly read? Is it worth special casing a header that is commonly read even if it's not commonly received? E.g. Content-Type?
- What response headers are commonly sent?

All told, this is a sliding scale problem where we rank the headers of interest and then draw a cutline. Don't recommend individual headers, rank your top 10-20 and we'll see how many we can include.

Example ranked lists:

Request:
Host
User-Agent
Content-Length
Accept
Referer
Accept-Encoding
Accept-Language
Cookie
Cache-Control
Connection

Response:
Server
Date
Content-Type
Content-Length
Transfer-Encoding
Cache-Control
Connection
Expires
Last-Modified
Set-Cookie
Pragma
Accept-Ranges
ETag
Link
Content-Encoding
Vary
Access-Control-Allow-Origin
Access-Control-Allow-Methods
Timing-Allow-Origin

benaadams:
@Tratcher with custom header factories though you can inherit the headers and special case specifics for the common server use case?

benaadams:
So for webapi focused server you might not need anything caching related; for an asset focus server you'd want all the caching and if modified headers; but wouldn't likely want anything POST related etc

halter73:
So whoever is implementing a custom header factory would have to generate the code themselves?

Also the way factories are implemented today, they custom header classes would have to derive from FrameRequestHeaders/FrameResponseHeaders instead of just implement an interface so there would be no way to remove fields, override some methods like `SetRawConnection` etc...

benaadams:
@halter73 yes, move FrameRequestHeaders/FrameResponseHeaders to an abstract class then offer standard variants or "server-profiles" - with the one that ships as Kestrel standard being a narrow subset.

I'll put a proposal together when RC2 tooling drops? Might be a crazy idea...

Tratcher:
Tweaking these on a per workload basis seems extreme, especially in the content server where this will be completely overshadowed by disk IO.

That said, how's this for a customization approach: Make the known headers lists configurable at startup, then generate optimized structures accordingly.

benaadams:
@Tratcher compile on the fly? As it is a bunch of `if`s

Tratcher:
I'll let you explore implementation ideas, I was more concerned about the surface area of the excitability model.

Tratcher:
And as for content, even if the disk access is fast, it's still going to take cycles to push all that content down the pipe. I expect cycles spent on real request and response bodies to overshadow fine tuning optimizations on headers.

aL3891:
Maybe this is a something generators can be used for? in my app i'd specify what headers I use, a generator would then make a custom Knownheaders type and plug that in via DI or something

nathana1:
I did a little more testing with a few different mixes of headers:  
- host header only (wrk default)
- requests with 40 of the headers I removed
- requests with the headers ie/edge pass to the techempower plaintext endpoint
- requests with what chrome passes to the plaintext endpoint

All of the browsers passed these headers (these are all still known headers in this PR):
- Connection
- User-Agent
- Accept, Accept-Encoding, Accept-Language

Chrome also passed these (though the caching ones vary when I refresh etc):
- Upgrade-Insecure-Request
- Cache-Control
- Pragma

Then I ran these against the current branch, this pr, and this pr with the 3 Chrome headers added (these are in millions of requests/sec on the techempower hardware):

|Scenario|Baseline|This PR|This PR+ Chrome headers|
|---|---|---|---|
|host only|4.21|4.59|4.59|
|40 headers|1.36|0.87|skipped|
|Edge headers (6)|2.92|3.05|3.08|
|Chrome headers (9)|2.45|2.45|2.58|

So few thoughts on the tradeoffs:
- adding a few (~3) known header doesn't make a measurable difference, but adding a lot does.
- having a few unknown headers does make a measureable difference, at about 3 there are no gains from this PR.
- with a few unknown headers the hotspots tend to be in the MemoryPoolIterator, especially FindFirstEqualByte and string allocations.
- with a lot of unknown headers, Dictionary operations become the hotspot

benaadams:
Any unknowns will cause a dictionary alloc in a non-header reuse situation

benaadams:
4 that absolutely *need* to be there on Request are Host, Connection, ContentLength and TransferEncoding.

Host is checked for invalid requests and the other 3 are checked as part of MessageBody https://github.com/aspnet/KestrelHttpServer/blob/dev/src/Microsoft.AspNetCore.Server.Kestrel/Http/MessageBody.cs#L110-L122

tools/Microsoft.AspNetCore.Server.Kestrel.GeneratedCode/KnownHeaders.cs
```diff
@@ -106,7 +105,8 @@ public static string GeneratedFile()
             };
             var responseHeaders = commonHeaders.Concat(new[]
             {
-                "ETag",
+                "Content-Type",
```
halter73: Isn't this used for POST requests?

test/Microsoft.AspNetCore.Server.KestrelTests/EngineTests.cs
```diff
@@ -810,17 +811,17 @@ public void ConnectionCanReadAndWrite(TestServiceContext testContext)
                     await connection.Receive(
                         "HTTP/1.1 500 Internal Server Error",
                         "");
+                    await connection.Receive("Content-Length: 0", "");
```
cesarbs: nit: move this into the previous call to `Receive()`, like above.

test/Microsoft.AspNetCore.Server.KestrelTests/EngineTests.cs
```diff
@@ -810,17 +811,17 @@ public void ConnectionCanReadAndWrite(TestServiceContext testContext)
...
                     await connection.ReceiveStartsWith("Date:");
                     await connection.Receive(
-                        "Content-Length: 0",
                         "Server: Kestrel",
                         "",
                         "HTTP/1.1 500 Internal Server Error",
                         "Connection: close",
                         "");
+                    await connection.Receive("Content-Length: 0", "");
```
cesarbs: nit: move this into the previous call to `Receive()`, like above.

test/Microsoft.AspNetCore.Server.KestrelTests/ChunkedRequestTests.cs
```diff
@@ -416,9 +416,9 @@ public static TheoryData<ServiceContext> ConnectionFilterData
                     await connection.Receive(
                         "HTTP/1.1 400 Bad Request",
                         "");
+                    await connection.Receive("Content-Length: 0", "");
```
cesarbs: nit: move this into the previous call to `Receive()`.

test/Microsoft.AspNetCore.Server.KestrelTests/ChunkedRequestTests.cs
```diff
@@ -372,9 +372,9 @@ public static TheoryData<ServiceContext> ConnectionFilterData
...
```
cesarbs: nit: move this into the previous call to `Receive()`.

halter73:
Now that we are caching header dictionaries for the lifetime of the connection, the perf gain per request (assuming a bunch of requests per connections) is quite minimized. Considering this PR still removes some headers we've identified as being pretty common (e.g. Cache-Control), I don't feel great merging this. I also don't think it's worth it right now to spend time collecting a representative sample of requests and responses Kestrel might serve to come up with a better list.

PR #832, opened 2016-05-12
reuse typeof results in generated Frame code

### LINKED ISSUE #831: generated code for Frame doesn't always reuse result of typeof
The types of various features gets cached in Frame.Generated.cs: https://github.com/aspnet/KestrelHttpServer/blob/dev/src/Microsoft.AspNetCore.Server.Kestrel/Http/Frame.Generated.cs#L9

But that cached value doesn't get reused in FastFeatureGet/Set: https://github.com/aspnet/KestrelHttpServer/blob/dev/src/Microsoft.AspNetCore.Server.Kestrel/Http/Frame.Generated.cs#L66

There is a small win (2.5%) on the TechEmpower hardware when reusing those values instead of calling typeof again:

|Scenario|Million Reqs/Sec|
|---|---|
|Baseline|5.57|
|Reuse|5.71|

nathana1:
Resolves #831.

There is a small win (2.5%) on the TechEmpower hardware when reusing these values instead of calling typeof again:

|Scenario|Million Reqs/Sec|
|---|---|
|Baseline| 5.57 |
|Reuse| 5.71 |

/cc @halter73 @cesarbs

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Kestrel/Http/Frame.Generated.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/Http/Frame.Generated.cs
@@ -63,67 +63,67 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Http
 
         private object FastFeatureGet(Type key)
         {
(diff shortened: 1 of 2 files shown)


benaadams:
:rocket:

Issue #827, opened 2016-05-11
NullReferenceException during shutdown


DavidObando:
Ran into this, but haven't been able to reproduce:
```
    MainToConfigureOverhead [Variation: Kestrel] [Stage: Collection] [Iteration: 47/100] [FAIL]
      System.NullReferenceException : Object reference not set to an instance of an object.
      Stack Trace:
           at Microsoft.AspNetCore.Server.Kestrel.KestrelThread.<OnStopRude>b__33_0(IntPtr ptr)
           at Microsoft.AspNetCore.Server.Kestrel.Networking.Libuv.NativeMethods.uv_walk(UvLoopHandle loop, uv_walk_cb walk_cb, IntPtr arg)
           at Microsoft.AspNetCore.Server.Kestrel.KestrelThread.OnStopRude()
           at Microsoft.AspNetCore.Server.Kestrel.KestrelThread.DoPostWork()
           at Microsoft.AspNetCore.Server.Kestrel.KestrelThread.OnPost()
           at Microsoft.AspNetCore.Server.Kestrel.Networking.Libuv.NativeMethods.uv_run(UvLoopHandle handle, Int32 mode)
           at Microsoft.AspNetCore.Server.Kestrel.Networking.Libuv.run(UvLoopHandle handle, Int32 mode)
           at Microsoft.AspNetCore.Server.Kestrel.KestrelThread.ThreadStart(Object parameter)
        --- End of stack trace from previous location where exception was thrown ---
           at Microsoft.AspNetCore.Server.Kestrel.KestrelThread.Stop(TimeSpan timeout)
           at Microsoft.AspNetCore.Server.Kestrel.KestrelEngine.Dispose()
           at Microsoft.AspNetCore.Server.Kestrel.KestrelServer.Dispose()
           at Microsoft.Extensions.DependencyInjection.ServiceProvider.Dispose()
           at Microsoft.AspNetCore.Hosting.Internal.WebHost.Dispose()
        HostingTests.cs(43,0): at Microbenchmarks.Tests.HostingTests.MainToConfigureOverhead(String variationServer)
```

This was from the Performance repository, while doing:
```
cd .\test\Microbenchmarks.Tests
dotnet test
```

A fix would be to check if handle != null before calling Dispose() [here](https://github.com/aspnet/KestrelHttpServer/blob/dev/src/Microsoft.AspNetCore.Server.Kestrel/Infrastructure/KestrelThread.cs#L133).

PR #833, opened 2016-05-12
Prevent NullReferenceException in KestrelThread.OnStopRude

halter73:
- UvMemory.FromIntPtr looks up a weak reference so we need to null check.

#827


cesarbs:
:shipit:

halter73:
Merged: 185ad00

Issue #792, opened 2016-05-03
Always send HTTP/1.1 responses


cesarbs:
Per RFC 7230, Section 2.6 http://tools.ietf.org/html/rfc7230#section-2.6:

> A server SHOULD send a response version equal to the highest version to which the server is conformant that has a major version less than or equal to the one received in the request.

Further:

> A server MAY send an HTTP/1.0 response to a request if it is known or suspected that the client incorrectly implements the HTTP specification and is incapable of correctly processing later version
responses, such as when a client fails to parse the version number correctly or when an intermediary is known to blindly forward the
HTTP-version even when it doesn't conform to the given minor version of the protocol.  Such protocol downgrades SHOULD NOT be performed
unless triggered by specific client attributes, such as when one or more of the request header fields (e.g., User-Agent) uniquely match the values sent by a client known to be in error.

About HTTP/1.0 clients:

> When an HTTP message is received with a major version number that the recipient implements, but a higher minor version number than what the recipient implements, the recipient SHOULD process the message as if it were in the highest minor version within that major version to which the recipient is conformant.  A recipient can assume that a message with a higher minor version, when sent to a recipient that
has not yet indicated support for that higher version, is sufficiently backwards-compatible to be safely processed by any implementation of the same major version.

We should always send HTTP/1.1 responses since that's the highest version we implement.

IIS and WebListener conform to the standard in this regard.

cesarbs:
@muratg I want this change in 1.0.0.

PR #835, opened 2016-05-16
Always send HTTP/1.1 responses (#792).

cesarbs:
RFC compliance. See #792 for details.

### DIFF
--- a/test/Microsoft.AspNetCore.Server.KestrelTests/ChunkedRequestTests.cs
+++ b/test/Microsoft.AspNetCore.Server.KestrelTests/ChunkedRequestTests.cs
@@ -77,7 +77,7 @@ namespace Microsoft.AspNetCore.Server.KestrelTests
                         "0",
                          "");
                     await connection.ReceiveEnd(
-                        "HTTP/1.0 200 OK",
+                        "HTTP/1.1 200 OK",
                         "",
                         "Hello World");
                 }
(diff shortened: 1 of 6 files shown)


test/Microsoft.AspNetCore.Server.KestrelTests/ChunkedRequestTests.cs
```diff
@@ -77,7 +77,7 @@ public static TheoryData<ServiceContext> ConnectionFilterData
                         "0",
                          "");
                     await connection.ReceiveEnd(
-                        "HTTP/1.0 200 OK",
+                        "HTTP/1.1 200 OK",
```
cesarbs: This is an odd test (expand above to see all) because HTTP/1.0 clients would hardly support chunked encoding. In any case, no reason to reject such a request. Plus, we're sending the right response. RFC 7230 section 3.3.1 says:

> A server MUST NOT send a response containing Transfer-Encoding unless the corresponding request indicates HTTP/1.1 (or later).

We implement this exact behavior in Frame.cs. You can see in this PR I've added a comment remarking that should never change.

test/Microsoft.AspNetCore.Server.KestrelTests/ChunkedRequestTests.cs
```diff
@@ -105,13 +105,13 @@ public static TheoryData<ServiceContext> ConnectionFilterData
                         "",
                         "Goodbye");
                     await connection.Receive(
-                        "HTTP/1.0 200 OK",
+                        "HTTP/1.1 200 OK",
```
cesarbs: My comment above applied here too.

mikeharder:
:shipit:

PR #836, opened 2016-05-17
Send 'Connection: close' in all 400 responses to HTTP/1.1 requests (#840).

cesarbs:
#840

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Kestrel/Http/Frame.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/Http/Frame.cs
@@ -30,7 +30,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Http
         private static readonly byte[] _bytesConnectionClose = Encoding.ASCII.GetBytes("\r\nConnection: close");
         private static readonly byte[] _bytesConnectionKeepAlive = Encoding.ASCII.GetBytes("\r\nConnection: keep-alive");
         private static readonly byte[] _bytesTransferEncodingChunked = Encoding.ASCII.GetBytes("\r\nTransfer-Encoding: chunked");
-        private static readonly byte[] _bytesHttpVersion1_1 = Encoding.ASCII.GetBytes("HTTP/1.1 ");
+        private static readonly byte[] _bytesHttpVersion11 = Encoding.ASCII.GetBytes("HTTP/1.1 ");
         private static readonly byte[] _bytesContentLengthZero = Encoding.ASCII.GetBytes("\r\nContent-Length: 0");
         private static readonly byte[] _bytesSpace = Encoding.ASCII.GetBytes(" ");
         private static readonly byte[] _bytesEndHeaders = Encoding.ASCII.GetBytes("\r\n\r\n");
@@ -87,29 +87,30 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Http
         {
             get
             {
-                if (_httpVersion == HttpVersionType.Http1_1)
+                if (_httpVersion == HttpVersionType.Http11)
                 {
                     return "HTTP/1.1";
                 }
-                if (_httpVersion == HttpVersionType.Http1_0)
+                if (_httpVersion == HttpVersionType.Http10)
                 {
                     return "HTTP/1.0";
                 }
+
                 return string.Empty;
             }
             set
             {
                 if (value == "HTTP/1.1")
                 {
-                    _httpVersion = HttpVersionType.Http1_1;
+                    _httpVersion = HttpVersionType.Http11;
                 }
                 else if (value == "HTTP/1.0")
                 {
-                    _httpVersion = HttpVersionType.Http1_0;
+                    _httpVersion = HttpVersionType.Http10;
                 }
                 else
                 {
-                    _httpVersion = HttpVersionType.Unknown;
+                    _httpVersion = HttpVersionType.Unset;
                 }
             }
         }
@@ -229,7 +230,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Http
             PathBase = null;
             Path = null;
             QueryString = null;
-            _httpVersion = HttpVersionType.Unknown;
+            _httpVersion = HttpVersionType.Unset;
             StatusCode = 200;
             ReasonPhrase = null;
 
@@ -512,7 +513,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Http
             }
 
             StringValues expect;
-            if (_httpVersion == HttpVersionType.Http1_1 &&
+            if (_httpVersion == HttpVersionType.Http11 &&
                 RequestHeaders.TryGetValue("Expect", out expect) &&
                 (expect.FirstOrDefault() ?? "").Equals("100-continue", StringComparison.OrdinalIgnoreCase))
             {
@@ -595,6 +596,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Http
                 {
                     // 400 Bad Request
                     StatusCode = 400;
+                    _keepAlive = false;
                 }
                 else
                 {
@@ -712,7 +714,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Http
                     //
                     // A server MUST NOT send a response containing Transfer-Encoding unless the corresponding
                     // request indicates HTTP/1.1 (or later).
-                    if (_httpVersion == HttpVersionType.Http1_1)
+                    if (_httpVersion == HttpVersionType.Http11)
                     {
                         _autoChunk = true;
                         responseHeaders.SetRawTransferEncoding("chunked", _bytesTransferEncodingChunked);
@@ -724,16 +726,16 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Http
                 }
             }
 
-            if (_keepAlive == false && responseHeaders.HasConnection == false && _httpVersion == HttpVersionType.Http1_1)
+            if (!_keepAlive && !responseHeaders.HasConnection && _httpVersion != HttpVersionType.Http10)
             {
                 responseHeaders.SetRawConnection("close", _bytesConnectionClose);
             }
-            else if (_keepAlive && responseHeaders.HasConnection == false && _httpVersion == HttpVersionType.Http1_0)
+            else if (_keepAlive && !responseHeaders.HasConnection && _httpVersion == HttpVersionType.Http10)
             {
                 responseHeaders.SetRawConnection("keep-alive", _bytesConnectionKeepAlive);
             }
 
-            end.CopyFrom(_bytesHttpVersion1_1);
+            end.CopyFrom(_bytesHttpVersion11);
             end.CopyFrom(statusBytes);
             responseHeaders.CopyTo(ref end);
             end.CopyFrom(_bytesEndHeaders, 0, _bytesEndHeaders.Length);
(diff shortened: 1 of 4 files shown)


test/Microsoft.AspNetCore.Server.KestrelTests/BadHttpRequestTests.cs
```diff
@@ -1,72 +1,99 @@
...
     {
-        [Theory(Skip = "This test fails intermittently and needs to be investigated.")]
-        [InlineData("/ HTTP/1.1\r\n\r\n")]
-        [InlineData(" / HTTP/1.1\r\n\r\n")]
-        [InlineData("  / HTTP/1.1\r\n\r\n")]
-        [InlineData("GET  / HTTP/1.1\r\n\r\n")]
-        [InlineData("GET  /  HTTP/1.1\r\n\r\n")]
-        [InlineData("GET  HTTP/1.1\r\n\r\n")]
+        // Don't send more data than necessary to fail, otherwise the test might throw if
+        // it's still trying to send characters when the server had already shut down the connection
```
mikeharder: It sounds like we should have tests for this behavior as well, rather than excluding such tests.

halter73: I think that is what `ReceiveBadRequestRespose` is verifying.

test/Microsoft.AspNetCore.Server.KestrelTests/BadHttpRequestTests.cs
```diff
@@ -1,72 +1,99 @@
...
+            {
+                using (var connection = new TestConnection(server.Port))
+                {
+                    await connection.Send(request);
+                    await ReceiveBadRequestRespose(connection);
                 }
             }
         }
+
+        private async Task ReceiveBadRequestRespose(TestConnection connection)
```
mikeharder: Typo: Respose -> Response

test/Microsoft.AspNetCore.Server.KestrelTests/BadHttpRequestTests.cs
cesarbs: The other test I added ensures we fail as soon as a bad request is seen, it kinda covers this. The problem is that if we send more data after the connection has already been closed and swallow the exceptions, we also fail to receive the response. This is not expected and I wonder if it has something to do with how sockets are implemented in .NET. Good time to page @stephentoub I guess :smile:

cesarbs: @mikeharder Also, the other test I added specifically ensures that Kestrel closes the connection as soon as a bad request line is detected (notice it uses `Send` instead of `SendEnd`).

mikeharder: It sounds like we need to understand this better at the `Socket` layer.  If a client sends too much data and the server closes the connection, the client still needs to receive the response so it can understand what it did wrong.

mikeharder: This could also be a bug in our `TestConnection` class rather than `Socket`.  Perhaps it shouldn't call `_socket.Shutdown()` until after the response has been read.  Maybe this works in normal cases, but fails if the server closed the connection.

cesarbs: @mikeharder I just did a quick test:

    const int serverPort = 8080;

    new Thread(_ =>
    {
        using (var server = new Socket(AddressFamily.InterNetwork, SocketType.Stream, ProtocolType.Tcp))
        {
            server.Bind(new IPEndPoint(IPAddress.Loopback, serverPort));
            server.Listen(128);

            while (true)
            {
                using (var client = server.Accept())
                {
                    Console.WriteLine("Server: sending response");
                    client.Send(Encoding.ASCII.GetBytes("response"));
                    Console.WriteLine("Server: shutting down connection");
                    client.Shutdown(SocketShutdown.Both);
                }
            }
        }
    }).Start();

    {
        Console.WriteLine("Client: connecting");
        socket.Connect(IPAddress.Loopback, serverPort);
        Console.WriteLine("Client: sending request");
        socket.Send(Encoding.ASCII.GetBytes("request"));
        Thread.Sleep(1000);
        var responseBytes = new byte[1024];
        Console.WriteLine("Client: receiving response");
        socket.Receive(responseBytes);
        Console.WriteLine($"Response: {Encoding.ASCII.GetString(responseBytes)}");
    }

Output:

    Client: connecting
    Client: sending request
    Server: sending response
    Server: shutting down connection
    Client: receiving response

    Unhandled Exception: System.Net.Sockets.SocketException: An existing connection was forcibly closed by the remote host
       at System.Net.Sockets.Socket.Receive(Byte[] buffer, Int32 offset, Int32 size, SocketFlags socketFlags)
       at System.Net.Sockets.Socket.Receive(Byte[] buffer)
       at SocketTests.Program.Main(String[] args)

The `socket.Receive()` call at the end throws. I'm not 100% sure, but I think we should be able to read data available on a socket even if the remote end has closed the connection.

cesarbs: I did some testing I figured out this is a Windows-only issue. What happens is that `Socket.Shutdown(SocketShutdown.Both)` sends an RST on Windows, essentially killing the connection and preventing the client from reading any data that had been sent previously.

image


I did the same test on Linux and noticed the same call sends a FIN. tcpdump output:

    08:42:14.193331 IP 127.0.0.1.8080 > 127.0.0.1.40060: Flags [F.], seq 9, ack 8, win 342, options [nop,nop,TS val 902865888 ecr 902865887], length 0

cesarbs: It's not really a `Socket` issue in .NET. This is the behavior in winsock.

cesarbs: Turns out it was not caused by the server shutting down the socket in both directions, since libuv shuts down the socket with `SD_SEND`.

The original issue was happening because Kestrel closed the connection while the client was still trying to write request bytes. This caused an exception to be thrown, which I though could be just swallowed so I could move on to trying to read the 400 response.

The problem seems to be that once we try to send data on a closed connection, winsock's (this only happens on Windows) error handling frees *everything* used by the client socket, *including* the buffered response from the server. So the test failed with an exception when we tried to read the response because it was not there anymore.

@mikeharder found an explanation here:

https://www.chilkatsoft.com/p/p_299.asp

test/Microsoft.AspNetCore.Server.KestrelTests/BadHttpRequestTests.cs
```diff
@@ -1,72 +1,119 @@
...
             }
         }
+
+        private async Task ReceiveBadRequestResponse(TestConnection connection, bool expectedConnectionCloseHeader)
+        {
+            await connection.Receive(
+                "HTTP/1.1 400 Bad Request",
+                "");
+
+            if (expectedConnectionCloseHeader)
```
cesarbs: This is super ugly. Could be mitigated by always setting `Connection: close` on error responses - which wouldn't be wrong at all. What do you guys think?

mikeharder: According to https://tools.ietf.org/html/rfc7230#section-6.6:

> The Connection header field (Section 6.1) provides a "close" connection option that a sender SHOULD send when it wishes to close the connection after the current request/response pair.

So if the server is closing the connection on an error response, it should set `Connection: close`.

A related question is whether the server should close the connection on an error response, or reuse the connection for subsequent requests.  I couldn't find anything in RFC 7230, but https://www.w3.org/Protocols/rfc2616/rfc2616-sec8.html#sec8.1.2 seems to indicate the server should keep the connection open:

> A significant difference between HTTP/1.1 and earlier versions of HTTP is that persistent connections are the default behavior of any HTTP connection. That is, unless otherwise indicated, the client SHOULD assume that the server will maintain a persistent connection, even after error responses from the server.

It might be reasonable to have different behavior for different errors.  For example, an invalid request could close the connection, while something like `404 Not Found` could keep the connection alive.

halter73: Could we use the `Connection: close` in conjunction with `Content-Length: 0`? FWIW we do keep the connection open for 500s, 404s, etc... It's mostly bad requests we close the connection for.

cesarbs: The RFC only specifies that 400s should be accompanied by a `Connection: close` in situations dealing with transfer encoding (http://tools.ietf.org/html/rfc7230#section-3.3.3). It doesn't specify that it should remain open otherwise, but I'd say keeping it open is more HTTP/1.1-ish.

So now I'm going against what I said before :grin: We *could* change the code to indicate in calls to `Frame.RejectRequest()` whether to keep the connection open or not.

benaadams: Bad requests should terminate the connection as there maybe more data in the input at the time the request is detected as being bad; which means where the next request begins cannot be determined.

cesarbs:
Have a look at the latest commit. It does two things:

* Sets `Frame._keepAlive` to false before sending a 400 response
* Always sends `Connection: close` responses to non-keepalive requests

Had to change a few HTTP/1.0 tests to check for the `Connection: close` header in the response.

cesarbs:
Nah, that wasn't good. Thinking.

cesarbs:
I think I'm happy with the latest commit.

* Replace `HttpVersionType` with a flag indicating whether it's an HTTP/1.0 request
* Always send `Connection: close` when `_keepAlive == false` AND the request HTTP version is NOT HTTP/1.0

Tratcher:
There's nothing wrong with sending `Connection: close` in response to a normal 1.0 request. You're just stating the default. Keep the logic simple.

test/Microsoft.AspNetCore.Server.KestrelTests/EngineTests.cs
```diff
@@ -727,21 +727,6 @@ public void ConnectionCanReadAndWrite(TestServiceContext testContext)
                 using (var connection = new TestConnection(server.Port))
                 {
                     await connection.SendEnd(
-                        "GET /");
```
cesarbs: This is covered in `BadHttpRequestTests`.

src/Microsoft.AspNetCore.Server.Kestrel/Http/Frame.cs
```diff
@@ -1097,13 +1079,6 @@ protected void ReportApplicationError(Exception ex)
             Log.ApplicationError(ConnectionId, ex);
         }
 
-        private enum HttpVersionType
```
halter73: I prefer the more "normalized" version of having the enum instead of a string and a bool.

This makes it clear that there are only 2 or 3 valid states.

halter73:
:ship:

halter73:
We don't actually send "Send 'Connection: close' in 400 responses to HTTP/1.1 requests.", right? I thought we just send 'Connection: close' for HTTP/1.0 requests and requests which don't properly include a valid HTTP versioin

cesarbs:
We currently send it on some situations, but not in the bad request line cases from `BadRequestTests`.

PR #841, opened 2016-05-18
Fast-path response header connection check

benaadams:
responseHeaders.HasConnection is a direct bool rather then calling GetEnumerator + MoveNext on StringValue struct


halter73:
:ship:

geoffkizer:
I was just looking at this in the TechEmpower trace.  Small win but definitely worth doing.

geoffkizer:
Would be even better to only call HasConnection once (we call it again later in the same function)

benaadams:
@geoffkizer have lifted the multi test vars to locals, that better?

geoffkizer:
I doubt it matters for _keepAlive and _httpVersion, since they are just member vars.  But HasConnection is a method call, so caching it seems worth it.

benaadams:
just changed it to hasconnection

geoffkizer:
Looks great!

PR #842, opened 2016-05-19
Remove extra FrameContext class

### LINKED ISSUE #837: FrameContext is an unneeded class
Everywhere that uses it, uses it to resolves `IFrameControl` only, which `Frame` could happily directly implement so it doesn't seem to add anything.

benaadams:
It doesn't particularly do anything

Resolves #837


src/Microsoft.AspNetCore.Server.Kestrel/Http/FrameResponseStream.cs
```diff
@@ -11,7 +11,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Http
 {
     class FrameResponseStream : Stream
     {
-        private FrameContext _context;
+        private IFrameControl _context;
```
cesarbs: Please rename it to `_frameControl`.

benaadams: Done

cesarbs: Forgot to push? :smile:

src/Microsoft.AspNetCore.Server.Kestrel/Http/FrameResponseStream.cs
```diff
@@ -134,14 +134,14 @@ public void Abort()
             }
         }
 
-        public void Initialize(FrameContext context)
+        public void Initialize(IFrameControl context)
         {
-            _context = context;
```
halter73: We could do the same rename in MessageBody.cs and TestInput.cs (though the test stuff is less important to me).

halter73:
Other than the finish of the context->frame rename, LGTM.

src/Microsoft.AspNetCore.Server.Kestrel/Http/FrameResponseStream.cs
benaadams: MessageBody is dependent on a few more bits, SocketInput, RejectRequest, TakeMessageHeaders. Could add to IFrameControl?

benaadams:
I think I've got all the renames now?

cesarbs:
:ship:

halter73:
:shipit:

PR #843, opened 2016-05-19
Remove Frame._requestProcessingStarted field.

cesarbs:
As agreed with @halter73, this is unnecessarily defensive.


halter73:
:shipit:

PR #845, opened 2016-05-19
Remove dead code in ReasonPhrases.

cesarbs:
More cleanup? Not sure if we want to keep as a public API though.

We should probably unify this with https://github.com/aspnet/HttpAbstractions/blob/dev/src/Microsoft.AspNetCore.WebUtilities/ReasonPhrases.cs.


halter73:
:shipit:

PR #846, opened 2016-05-19
Move ECONNRESET value check to server initialization.

cesarbs:
As suggested by @halter73.

cc @mikeharder @davidfowl


halter73:
:shipit:

Issue #830, opened 2016-05-12
Client certificates are disposed before request can use them


shanewalters:
### Functional impact
When using Kestrel with HTTPS and requiring client certificates, the client certificate added to context feature collection is disposed after the callback to `userCertificateValidationCallback()` has completed. In a subsequent controller action, a request to `HttpContext.Connection.ClientCertificate` returns an instance of the disposed certificate.

### Minimal repro steps
1. Setup a basic/empty web project
2. In the `Startup` class, in `UseKestrel()`, call `options.UseHttps()` with `ClientCertificateMode = ClientCertificateMode.RequireCertificate`, and `ClientCertificateValidation = (x, y, z) => x.Verify()`
3. Setup a basic controller and action, say `Test/Index()`, that derives from `Microsoft.AspNetCore.Mvc.Controller`.
4. In the sample controller and action, make a call to `HttpContext.Connection.ClientCertificate.Verify()`

### Expected result
The client certificate should not be disposed, and can be examined.

### Actual result
An exception is thrown when trying to use the certificate:
`System.Security.Cryptography.CryptographicException: m_safeCertContext is an invalid handle.`

### Further technical details
In the `userCertificateValidationCallback()` method, at the very end of the call, a reference is made to the valid certificate (`clientCertificate = certificate2`); however, because the caller of `userCertificateValidationCallback()` calls `certificate.Reset()` after validation, the local variable (`clientCertificate`) is now a reference to a disposed certificate. This disposed certificate is then added to the feature collection: `features.Set<ITlsConnectionFeature>(new TlsConnectionFeature { ClientCertificate = clientCertificate });`

### Observations
I observed that simply cloning the certificate did not appear to resolve the issue (such as a call to `new X509Certificate2(clientCertificate)` from inside `userCertificateValidationCallback()`); however, if I stored the `RawData` byte array (from inside `userCertificateValidationCallback()`), then call `features.Set<ITlsConnectionFeature>(new TlsConnectionFeature { ClientCertificate = new X509Certificate2(rawData) });`, the certificate was available, but I don't know if this is the correct solution or not.

### Related File
https://github.com/aspnet/KestrelHttpServer/blob/c828fafe1b39279d590ad1a18b62844bae76d77b/src/Microsoft.AspNetCore.Server.Kestrel.Https/HttpsConnectionFilter.cs

shanewalters:
If it would help, I could put together a PR.

halter73:
[...]

PR #848, opened 2016-05-19
Use SslStream.RemoteCertificate to recreate disposed certificate

shanewalters:
Updated IFeatureCollection to use new X509Certificate2, rather than the disposed instance from `userCertificateValidationCallback()`.

Addresses #830.

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Kestrel.Https/HttpsConnectionFilter.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel.Https/HttpsConnectionFilter.cs
@@ -40,7 +40,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Https
 
             if (string.Equals(context.Address.Scheme, "https", StringComparison.OrdinalIgnoreCase))
             {
-                X509Certificate2 clientCertificate = null;
+                var certificateHandle = IntPtr.Zero;
                 SslStream sslStream;
                 if (_options.ClientCertificateMode == ClientCertificateMode.NoCertificate)
                 {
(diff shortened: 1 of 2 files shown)


src/Microsoft.AspNetCore.Server.Kestrel.Https/HttpsConnectionFilter.cs
```diff
@@ -98,9 +98,9 @@ public HttpsConnectionFilter(HttpsConnectionFilterOptions options, IConnectionFi
                 {
                     previousPrepareRequest?.Invoke(features);
 
-                    if (clientCertificate != null)
+                    if (certificateHandle != IntPtr.Zero)
                     {
-                        features.Set<ITlsConnectionFeature>(new TlsConnectionFeature { ClientCertificate = clientCertificate });
```
halter73: Could this be changed to the following?

    var clientCertificate = ConvertToX509Certificate2(sslStream.RemoteCertificate);
    if (clientCertificate != null)
    {
        features.Set<ITlsConnectionFeature>(new TlsConnectionFeature { ClientCertificate = clientCertificate });
    }

    // ...

    private X509Certificate2 ConvertToX509Certificate2(X509Certificate certificate)
    {
        if (certificate == null)
        {
            return null;
        }

        X509Certificate2 certificate2 = certificate as X509Certificate2;
        if (certificate2 != null)
        {
            return certificate2;
        }

    #if NETSTANDARD1_3
        // conversion X509Certificate to X509Certificate2 not supported
        // https://github.com/dotnet/corefx/issues/4510
        return null;
    #else
        return new X509Certificate2(certificate);
    #endif
    }

This way we avoid copying `clientCertificate` from the `userCertificateValidationCallback` which I think is cleaner. We can also reuse the `ConvertToX509Certificate2` method in the `userCertificateValidationCallback`.

halter73:
Can you add a test verifying your PR fixes #830? Thanks!

shanewalters:
The added test is `CertificatePassedToHttpContextIsNotDisposed()`. The important line is `Assert.NotNull(context.Connection.ClientCertificate.PublicKey);` which will fail after a call to reset/dispose.

halter73:
@shanewalters Awesome. Thanks again! :shipit:

PR #851, opened 2016-05-19
Always cache headers and streams across frames in the same connection (#754).

cesarbs:
Addresses #754. Final decision was to always cache i.e. not make it configurable.

Previous PR addressing this was #781. I basically took that and removed the options stuff. Also added test to check that we're reusing headers (we were only testing stream reuse).

cc @halter73 @mikeharder @davidfowl @Tratcher @nathana1


benaadams:
Can we also cache `DefaultHttpContext` also then? Purdy please... :smile_cat:

cesarbs:
@benaadams There are some per-request things that are set by `HttpContextFactory` in the context: https://github.com/aspnet/HttpAbstractions/blob/dev/src/Microsoft.AspNetCore.Http/HttpContextFactory.cs

benaadams:
@cesarbs yeah, but PooledHttpContextFactory could also be the default Factory https://github.com/aspnet/HttpAbstractions/blob/dev/samples/SampleApp/PooledHttpContextFactory.cs

davidfowl:
@benaadams we're not going to do that for RTM. We'll look at more options post RTM.

cesarbs:
Ping.

src/Microsoft.AspNetCore.Server.Kestrel/Http/Frame.cs
```diff
@@ -174,16 +174,25 @@ public bool HasResponseStarted
...
         {
-            _frameStreams = HttpComponentFactory.CreateStreams(this);
+            if (_frameStreams == null)
+            {
+                _frameStreams = new Streams();
+            }
 
+            _frameStreams.Initialize(this);
             RequestBody = _frameStreams.RequestBody.StartAcceptingReads(messageBody);
             ResponseBody = _frameStreams.ResponseBody.StartAcceptingWrites();
```
halter73: Both of the above `Start*` methods return `this`. I would make these void and set `RequestBody`, `ResponseBody` and `DuplexStream` in the if statement above.

src/Microsoft.AspNetCore.Server.Kestrel/Http/Frame.cs
```diff
@@ -248,22 +257,8 @@ public void Reset()
...
-            var frameStreams = Interlocked.Exchange(ref _frameStreams, null);
-            if (frameStreams != null)
-            {
-                RequestBody = null;
-                ResponseBody = null;
-                DuplexStream = null;
-                HttpComponentFactory.DisposeStreams(frameStreams);
-            }
+            _frameHeaders?.Uninitialize();
+            _frameStreams?.Uninitialize();
```
halter73: Uninitialized does nothing for streams since StopStreams should have already been called.

src/Microsoft.AspNetCore.Server.Kestrel/Http/Frame.cs
```diff
@@ -174,16 +174,25 @@ public bool HasResponseStarted
...
 
         public void InitializeStreams(MessageBody messageBody)
         {
-            _frameStreams = HttpComponentFactory.CreateStreams(this);
+            if (_frameStreams == null)
+            {
+                _frameStreams = new Streams();
+            }
 
+            _frameStreams.Initialize(this);
```
halter73: Initialize doesn't do anything for streams that couldn't be done in the constructor.

src/Microsoft.AspNetCore.Server.Kestrel/Http/Frame.cs
```diff
@@ -248,22 +257,8 @@ public void Reset()
...
-
-            var frameStreams = Interlocked.Exchange(ref _frameStreams, null);
-            if (frameStreams != null)
-            {
-                RequestBody = null;
-                ResponseBody = null;
-                DuplexStream = null;
-                HttpComponentFactory.DisposeStreams(frameStreams);
-            }
+            _frameHeaders?.Uninitialize();
```
halter73: I would just call this directly in Reset() now.

src/Microsoft.AspNetCore.Server.Kestrel/Infrastructure/Streams.cs
```diff
@@ -0,0 +1,31 @@
...
+            DuplexStream = new FrameDuplexStream(RequestBody, ResponseBody);
+        }
+
+        public FrameRequestStream RequestBody { get; }
+        public FrameResponseStream ResponseBody { get; }
+        public FrameDuplexStream DuplexStream { get; }
+
+        public void Initialize(IFrameControl frameControl)
+        {
+            ResponseBody.Initialize(frameControl);
```
halter73: I would also move this to the FrameResponseStream ctor.

src/Microsoft.AspNetCore.Server.Kestrel/Infrastructure/Streams.cs
```diff
@@ -0,0 +1,31 @@
...
+        }
+
+        public void Uninitialize()
+        {
+            ResponseBody.Uninitialize();
```
halter73: This method can be removed from here and FrameResponseStream.

src/Microsoft.AspNetCore.Server.Kestrel/Http/Frame.cs
halter73: I want to remove ResetComponents().

src/Microsoft.AspNetCore.Server.Kestrel/Infrastructure/Headers.cs
```diff
@@ -0,0 +1,29 @@
...
+        {
+            var dateHeaderValues = dateValueManager.GetDateHeaderValues();
+            ResponseHeaders.SetRawDate(dateHeaderValues.String, dateHeaderValues.Bytes);
+            ResponseHeaders.SetRawServer("Kestrel", BytesServer);
+        }
+
+        public FrameRequestHeaders RequestHeaders { get; } = new FrameRequestHeaders();
+        public FrameResponseHeaders ResponseHeaders { get; } = new FrameResponseHeaders();
+
+        public void Uninitialize()
```
halter73: I would like to rename this to Reset() for consistency.

halter73:
:shipit:

Issue #541, opened 2015-12-29
Skip Null check and zero length checks of Pointers


Anderman:
pointers to byte arrays are use as 
`fixed(byte* ptr arr)` 
the code generated will check if this is an null or zero length array

To skip these tests the statement can be written as
`fixed(byte* ptr &arr[0])`

see generated code

`fixed(byte* ptr arr)` 
```
000007FF622857AB  mov         rax,rcx  
000007FF622857AE  test        rax,rax  
000007FF622857B1  je          000007FF622857BA  
000007FF622857B3  mov         ecx,dword ptr [rax+8]  
000007FF622857B6  test        ecx,ecx  
000007FF622857B8  jne         000007FF622857C3  
000007FF622857BA  xor         ecx,ecx  
000007FF622857BC  mov         qword ptr [rsp+20h],rcx  
000007FF622857C1  jmp         000007FF622857D1  
000007FF622857C3  cmp         ecx,0  
000007FF622857C6  jbe         000007FF622857FE  
000007FF622857C8  add         rax,10h  
000007FF622857CC  mov         qword ptr [rsp+20h],rax 
```

`fixed(byte* ptr &arr[0])`
```
000007FF6228582B  cmp         dword ptr [rcx+8],0  
000007FF6228582F  jbe         000007FF62285867  
000007FF62285831  add         rcx,10h  
000007FF62285835  mov         qword ptr [rsp+20h],rcx 
```

benaadams:
Interesting :)

Anderman:
This one is also nice to know [Optimize generated code for Fixed statement](https://github.com/dotnet/coreclr/issues/2480). Looks like a clr or roslyn issue

clrjunkie:
+1

halter73:
@Anderman did #542 address all the places where we should use the `fixed(byte* ptr &arr[0])` pattern?

Anderman:
Yes, all that I could found. I skipped 1 or 2 because they didn't have much impact.

Anderman:
@halter73  I think @benaadams  did already change some in his PR

halter73:
I also noticed @benaadams used the same pattern in at least one of his PRs.

Thanks! Just wanted to double check before closing.

benaadams:
@Anderman is making me up my game :)

Anderman:
Ha, Thx there still is many things that can be improved

halter73:
[...]

PR #853, opened 2016-05-20
add optimized FastByteCopy

geoffkizer:
Buffer.BlockCopy is not particularly efficient for small blocks.

For blocks < 64 bytes, do the copy ourselves in unsafe code.

This gives something like 4% on TechEmpower plaintext (when combined with other proposed fixes).

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Kestrel/Infrastructure/MemoryPoolIterator.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/Infrastructure/MemoryPoolIterator.cs
@@ -663,6 +663,109 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Infrastructure
             }
         }
 
+        public static unsafe void FastByteCopy(byte[] srcArray, uint srcOffset, byte[] destArray, uint destOffset, uint byteCount)
+        {
+            Debug.Assert(srcOffset + byteCount <= srcArray.Length);
+            Debug.Assert(destOffset + byteCount <= destArray.Length);
+            
+            if (byteCount < 64)
+            {
+                // Buffer.BlockCopy is expensive for small copies.  Do it manually ourselves.
+                
+                fixed (byte* pSrc = srcArray)
+                fixed (byte* pDest = destArray)
+                {
+                    var src = pSrc + srcOffset;
+                    var dest = pDest + destOffset;
+
+                    while (byteCount >= 16)
+                    {
+                        *(long*)dest = *(long*)src;
+                        *(long*)(dest + 8) = *(long*)(src + 8);
+                        
+                        src += 16;
+                        dest += 16;
+                        byteCount -= 16;
+                    }
+                    
+                    switch (byteCount)
+                    {
+                    case 0:
+                        return;
+                    case 1:
+                        *dest = *src;
+                        return;
+                    case 2:
+                        *(short *)dest = *(short *)src;
+                        return;
+                    case 3:
+                        *(short *)dest = *(short *)src;
+                        *(dest + 2) = *(src + 2);
+                        return;
+                    case 4:
+                        *(int *)dest = *(int *)src;
+                        return;
+                    case 5:
+                        *(int*)dest = *(int*)src;
+                        *(dest + 4) = *(src + 4);
+                        return;
+                    case 6:
+                        *(int*)dest = *(int*)src;
+                        *(short*)(dest + 4) = *(short*)(src + 4);
+                        return;
+                    case 7:
+                        *(int*)dest = *(int*)src;
+                        *(short*)(dest + 4) = *(short*)(src + 4);
+                        *(dest + 6) = *(src + 6);
+                        return;
+                    case 8:
+                        *(long*)dest = *(long*)src;
+                        return;
+                    case 9:
+                        *(long*)dest = *(long*)src;
+                        *(dest + 8) = *(src + 8);
+                        return;
+                    case 10:
+                        *(long*)dest = *(long*)src;
+                        *(short*)(dest + 8) = *(short*)(src + 8);
+                        return;
+                    case 11:
+                        *(long*)dest = *(long*)src;
+                        *(short*)(dest + 8) = *(short*)(src + 8);
+                        *(dest + 10) = *(src + 10);
+                        return;
+                    case 12:
+                        *(long*)dest = *(long*)src;
+                        *(int*)(dest + 8) = *(int*)(src + 8);
+                        return;
+                    case 13:
+                        *(long*)dest = *(long*)src;
+                        *(int*)(dest + 8) = *(int*)(src + 8);
+                        *(dest + 12) = *(src + 12);
+                        return;
+                    case 14:
+                        *(long*)dest = *(long*)src;
+                        *(int*)(dest + 8) = *(int*)(src + 8);
+                        *(short*)(dest + 12) = *(short*)(src + 12);
+                        return;
+                    case 15:
+                        *(long*)dest = *(long*)src;
+                        *(int*)(dest + 8) = *(int*)(src + 8);
+                        *(short*)(dest + 12) = *(short*)(src + 12);
+                        *(dest + 14) = *(src + 14);
+                        return;
+                    default:
+                        Debug.Assert(false);
+                        break;
+                    }
+                }
+            }
+            else
+            {
+                Buffer.BlockCopy(srcArray, (int)srcOffset, destArray, (int)destOffset, (int)byteCount);
+            }
+        }
+        
         public MemoryPoolIterator CopyTo(byte[] array, int offset, int count, out int actual)
         {
             if (IsDefault)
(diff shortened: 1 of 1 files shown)


geoffkizer:
@benaadams, @nathana1, @halter73

benaadams:
Vectors are better > `Vector<byte>.Count` with some unrolling; though may be better when the `Unsafe` library is available. Not sure much in the techempower it that large?

:rocket:

geoffkizer:
Yeah, Vectors are better, at least in theory.  Unsafe library would be needed to make this work well, I think.

That said, vector size can vary based on hardware, so that would make the code a bit more involved; not sure it's worth it for these scenarios.  

BTW, I played around with this a bit, but it's certainly worth looking at more.  There may be tweaks we could do to make this better.

Ideally we could have the JIT do smart things here, e.g. do appropriate vector optimization, generate fixed size copies, etc.  That's probably a longer term thing.

Largest copy in TechEmpower is 37 bytes, for the Date header.

benaadams:
Actually do have a change can drop one of the fixed blocks in all cases

src/Microsoft.AspNetCore.Server.Kestrel/Infrastructure/MemoryPoolIterator.cs
```diff
@@ -687,7 +790,7 @@ public MemoryPoolIterator CopyTo(byte[] array, int offset, int count, out int ac
                     actual = count;
                     if (array != null)
                     {
-                        Buffer.BlockCopy(block.Array, index, array, offset, remaining);
+                        FastByteCopy(block.Array, (uint)index, array, (uint)offset, (uint)remaining);
```
benaadams: can use `block.DataFixedPtr` rather than `fixed`

src/Microsoft.AspNetCore.Server.Kestrel/Infrastructure/MemoryPoolIterator.cs
```diff
@@ -663,6 +663,109 @@ public int GetLength(MemoryPoolIterator end)
...
+        {
+            Debug.Assert(srcOffset + byteCount <= srcArray.Length);
+            Debug.Assert(destOffset + byteCount <= destArray.Length);
+            
+            if (byteCount < 64)
+            {
+                // Buffer.BlockCopy is expensive for small copies.  Do it manually ourselves.
+                
+                fixed (byte* pSrc = srcArray)
+                fixed (byte* pDest = destArray)
```
benaadams: oddly `fixed (byte* pDest = &destArray[0])` is faster

geoffkizer:
I think fixed() here is basically a no-op.

src/Microsoft.AspNetCore.Server.Kestrel/Infrastructure/MemoryPoolIterator.cs
```diff
@@ -663,6 +663,109 @@ public int GetLength(MemoryPoolIterator end)
...
         }
 
+        public static unsafe void FastByteCopy(byte[] srcArray, uint srcOffset, byte[] destArray, uint destOffset, uint byteCount)
+        {
+            Debug.Assert(srcOffset + byteCount <= srcArray.Length);
+            Debug.Assert(destOffset + byteCount <= destArray.Length);
+            
+            if (byteCount < 64)
+            {
+                // Buffer.BlockCopy is expensive for small copies.  Do it manually ourselves.
```
davidfowl: Why doesn't `Buffer.BlockCopy` do this?

geoffkizer: Buffer.BlockCopy has to do type checks regardless, so it's better to have an unsafe version that doesn't require this.

Perhaps we should have something like Buffer.UnsafeByteCopy, but that's a larger discussion.

davidfowl: I mean, we have other code like this in kestrel but it's not great that we need to do it.

geoffkizer: I agree, I'm just saying that figuring out a better way is probably a longer discussion.  Do you want to have that discussion before we decide whether to take this?

davidfowl: I rather do it in parallel.

benaadams: https://github.com/dotnet/coreclr/issues/2430 ;-)

src/Microsoft.AspNetCore.Server.Kestrel/Infrastructure/MemoryPoolIterator.cs
halter73: @geoffkizer Why is fixed() basically a no-op? I get that it's probably just marks the array and has no major impact on gc except in the relatively rare case when the collector would want to move the array in the middle of `FastByteCopy`.

I think the bigger issue is that the code generated for `fixed (byte* pDest = destArray)` will check if `destArray` is an null or zero length array while `fixed (byte* pDest = &destArray[0])` wont. At least according to #541 which is why we use this pattern in a few other places like `GetAsciiStringAsLong`.

geoffkizer: Oddly, I get an IndexOutOfRange exception during unit tests with "byte* pDest = &destArray[0]", whereas I don't with "byte* pDest = destArray".

Are you sure the info from #541 is correct?

benaadams: will be length 0 arrays

benaadams: Will be length 0 arrays; when you probably just want to skip the whole function anyway

geoffkizer: Yeah, I understand, but I don't want to add a test for 0 length.  The function handles that case fine.

benaadams: `if (byteCount == 0) return;`

src/Microsoft.AspNetCore.Server.Kestrel/Infrastructure/MemoryPoolIterator.cs
```diff
@@ -663,6 +663,109 @@ public int GetLength(MemoryPoolIterator end)
             }
         }
 
+        public static unsafe void FastByteCopy(byte[] srcArray, uint srcOffset, byte[] destArray, uint destOffset, uint byteCount)
+        {
+            Debug.Assert(srcOffset + byteCount <= srcArray.Length);
```
benaadams: Asserts on blocks should be testing `block.Data.Count` rather than `block.Array.Length`

benaadams:
@geoffkizer this any good? https://github.com/benaadams/KestrelHttpServer/commit/cc11660005381824242f0eca9b58f9f7b28f058e

benaadams:
@geoffkizer this any good? https://github.com/benaadams/KestrelHttpServer/commit/201139ae2437c6d4fb01b3e2ab5af75a9fbb2449

benaadams:
Actually this is probably better https://github.com/benaadams/KestrelHttpServer/commit/757f7e48420c592d7785699385ce70d986ed8b2b?w=1

halter73:
I'm going to merge this as is. @davidfowl Sound good to you?

halter73:
@geoffkizer For small byte copies do you think the cost of `fixed` is significant enough to warrant taking advantage of `block.DataFixedPtr` like @benaadams suggested?

geoffkizer:
I tried using DataFixedPtr and it didn't seem to matter.

benaadams:
The Debug.Asserts are still testing the wrong things

benaadams:
The Debug.Asserts are still testing the wrong things; they are testing against the entire slab array size rather than the individual block size

geoffkizer:
As it stands, FastByteCopy is intentionally ignorant of Blocks, so that it can be a general purpose copy mechanism.  That's why the asserts are the way they are.

benaadams:
Which is why I added the asserts to CopyTo at the block level so it would assert if any of the copies were going outside of the block. 

As the block is a small window over a much larger array the asserts in the FastByteCopy are unlikely to ever trigger even if CopyTo is being asked to inadvertently overwrite adjacent blocks; as is far as its concerned its a valid request.

benaadams:
Ofc, they are additional asserts that weren't there previously and the behaviour is not changing on BlockCopy - so might not matter

benaadams:
Ignore me; likely out of scope for this PR

benaadams:
Added PR https://github.com/aspnet/KestrelHttpServer/pull/900 for other asserts

src/Microsoft.AspNetCore.Server.Kestrel/Infrastructure/MemoryPoolIterator.cs
benaadams: @halter73 turns out &var is a deopt on ryujit x64 https://gist.github.com/benaadams/11123e9785481e07d8ca483bef798689

halter73:
@geoffkizer @davidfowl Closing since we want  #897 for 1.0.1, and we didn't think this kind of change was worth merging temporarily for 1.0.0.

Issue #805, opened 2016-05-05
Response allows setting StatusCode and ReasonPhrase after response was started


Tratcher:
_From @pakrym on May 5, 2016 21:36_

This works:
```
await context.Response.WriteAsync("Hello");
context.Response.StatusCode = 2;
```
It should prevent setting status code the same way `Headers`, `Cookies`, `ContentType` do.

/cc @Tratcher 

_Copied from original issue: aspnet/HttpAbstractions#626_

PR #857, opened 2016-05-20
Throw when setting Frame.StatusCode after response has already started (#805).

cesarbs:
Note this doesn't change the response. Since the response has already started we can't send a 500. At least we get a log message.

#805 

cc @halter73 @mikeharder @Tratcher @pakrym


src/Microsoft.AspNetCore.Server.Kestrel/Http/Frame.cs
```diff
@@ -115,7 +115,24 @@ public string HttpVersion
...
+                if (HasResponseStarted)
+                {
+                    throw new InvalidOperationException("Status code cannot be set, response has already started.");
+                }
+
+                _statusCode = value;
+            }
+        }
+
         public string ReasonPhrase { get; set; }
```
Tratcher: ReasonPhrase too.

src/Microsoft.AspNetCore.Server.Kestrel/Http/Frame.cs
```diff
@@ -115,7 +115,24 @@ public string HttpVersion
         public IHeaderDictionary RequestHeaders { get; set; }
         public Stream RequestBody { get; set; }
 
-        public int StatusCode { get; set; }
+        private int _statusCode;
+        public int StatusCode
```
Tratcher: What about https://github.com/aspnet/KestrelHttpServer/blob/c6865f0e447b99bb5b56a943d155d986a4120bcb/src/Microsoft.AspNetCore.Server.Kestrel/Http/Frame.cs#L614-L617 ?

test/Microsoft.AspNetCore.Server.KestrelTests/FrameTests.cs
```diff
@@ -50,5 +52,46 @@ public void EmptyHeaderValuesCanBeParsed(string rawHeaders, int numHeaders)
...
+
+            // Act
+            frame.Reset();
+
+            // Assert
+            Assert.Equal("http", ((IFeatureCollection)frame).Get<IHttpRequestFeature>().Scheme);
+        }
+
+        [Fact]
+        public void ThrowsWhenStatusCodeIsSetAfterResponseStarted()
```
Tratcher: Are there similar tests for setting Header values?

cesarbs: I don't think so. Will add.

src/Microsoft.AspNetCore.Server.Kestrel/Http/Frame.cs
halter73: I would probably just change that to something like:


    if (HasResponseStarted)
    {
        // We can no longer respond with a 500, so we simply close the connection.
        _requestProcessingStopping = true;
    }
    else
    {
        if (_requestRejected)
        {
            // 400 Bad Request
            StatusCode = 400;
            _keepAlive = false;
        }
        else
        {
            // 500 Internal Server Error
            StatusCode = 500;
        }

        // ...

test/Microsoft.AspNetCore.Server.KestrelTests/FrameTests.cs
halter73: Add it to `FrameResponseHeadersTests`.

src/Microsoft.AspNetCore.Server.Kestrel/Http/Frame.cs
cesarbs: Yeah, it's what I did.

test/Microsoft.AspNetCore.Server.KestrelTests/FrameTests.cs
cesarbs: Done.

cesarbs:
Ping.

halter73:
:shipit:

Issue #814, opened 2016-05-09
Don't bind if only port is specified.


Tratcher:
You can pass just a port integer into Kestrel as the address to bind to. Currently Kestrel assumes this means to bind to the given port publicly for all IPs. https://github.com/aspnet/KestrelHttpServer/blob/dev/src/Microsoft.AspNetCore.Server.Kestrel/ServerAddress.cs#L87

Kestrel should only bind to localhost unless explicitly specified. Note WebListener already behaves this way.

Tratcher:
https://github.com/aspnet/WebListener/blob/dev/src/Microsoft.Net.Http.Server/UrlPrefix.cs#L128

Tratcher:
Change of plans (as reflected in the title). Remove support for only specifying the port. People make bad assumptions about the rest of the prefix.

mikeharder:
Since this is a breaking change, I plan to create an issue under https://github.com/aspnet/Announcements after this is merged.

PR #861, opened 2016-05-21
Don't bind if only port is specified (#814)

mikeharder:
- Also renamed ServerAddressFacts to ServerAddressTests to match existing test classes

@Tratcher, @cesarbs, @halter73


cesarbs:
After your change, what happens if you specify only the port number? Add a test for whatever the expected/resulting behavior is.

halter73:
:shipit:

@cesarbs This does add a unit test verifying that ServerAddress.FromUrl returns null. This is what we also do for missing scheme which results in [A FormatException "Unrecognized listening address: ..."](https://github.com/aspnet/KestrelHttpServer/blob/60988801324cc918c9f589fe08b7a4bbc491ac79/src/Microsoft.AspNetCore.Server.Kestrel/KestrelServer.cs#L106)

cesarbs:
:ship:

PR #862, opened 2016-05-21
Set Date and Server headers on response start

### LINKED ISSUE #223: Move the response date header later
Date is a required response header. If it's set at the beginning of a request it's too easy to clear, e.g. the exception page will clear response headers before applying its own.

Probably here where your serializing the response headers:
https://github.com/aspnet/KestrelHttpServer/blob/b466c3b7d7aba238867c2e57f54b9d44cf5b3ead/src/Microsoft.AspNet.Server.Kestrel/Http/Frame.cs#L524

khellang:
Closes #223 

// @Tratcher


src/Microsoft.AspNetCore.Server.Kestrel/Infrastructure/Headers.cs
```diff
@@ -1,34 +1,14 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
-using System.Text;
 using Microsoft.AspNetCore.Server.Kestrel.Http;
 
 namespace Microsoft.AspNetCore.Server.Kestrel.Infrastructure
 {
     class Headers
```
khellang: Is it worth keeping this class at this point?

test/Microsoft.AspNetCore.Server.KestrelTests/ChunkedRequestTests.cs
```diff
@@ -78,6 +78,8 @@ public static TheoryData<ServiceContext> ConnectionFilterData
                          "");
                     await connection.ReceiveEnd(
                         "HTTP/1.1 200 OK",
+                        "Date: " + testContext.DateHeaderValue,
+                        "Server: Kestrel",
```
khellang: Is it worth setting `AddServerHeader = false` in `TestServiceContext` to get rid of all of these `Server` headers in the tests?

khellang: Also, should I go through and remove redundant calls to `response.Headers.Clear()` (used to clear the `Date` and `Server` headers)?

test/Microsoft.AspNetCore.Server.KestrelTests/ChunkedRequestTests.cs
```diff
@@ -107,12 +109,16 @@ public static TheoryData<ServiceContext> ConnectionFilterData
                     await connection.Receive(
                         "HTTP/1.1 200 OK",
                         "Connection: keep-alive",
+                        "Date: " + testContext.DateHeaderValue,
```
khellang: Do you think this is an OK solution to assert on the `Date` header? I see `connection.ReceiveStartsWith("Date:")` is used a few places. Would you rather I used that instead?

cesarbs: If there's no chance of a mismatch, I'd prefer that (the new one).

cesarbs: nit: can you use `$"Date: {testContext.DateHeaderValue}"` instead?

test/Microsoft.AspNetCore.Server.KestrelTests/FrameResponseHeadersTests.cs
```diff
@@ -14,73 +14,6 @@ namespace Microsoft.AspNetCore.Server.KestrelTests
     public class FrameResponseHeadersTests
     {
         [Theory]
-        [InlineData(true)]
-        [InlineData(false)]
-        public void InitialDictionaryContainsServerAndDate(bool addServerHeader)
```
cesarbs: I'd rather have this converted into a test that checks those headers are *not* there.

test/Microsoft.AspNetCore.Server.KestrelTests/FrameResponseHeadersTests.cs
```diff
@@ -14,73 +14,6 @@ namespace Microsoft.AspNetCore.Server.KestrelTests
...
-            Assert.True(headers.TryGetValue("Date", out dateHeader));
-            Assert.Equal(1, dateHeader.Count);
-            Assert.True(DateTime.TryParse(dateHeader[0], out date));
-            Assert.True(DateTime.Now - date <= TimeSpan.FromMinutes(1));
-
-            Assert.False(headers.IsReadOnly);
-        }
-
-        [Fact]
-        public void InitialEntriesCanBeCleared()
```
cesarbs: I'd rather have this converted into a test that checks the new initial header count.

src/Microsoft.AspNetCore.Server.Kestrel/Infrastructure/Headers.cs
cesarbs: Seems rather pointless indeed 😄

test/Microsoft.AspNetCore.Server.KestrelTests/ChunkedRequestTests.cs
cesarbs: Sounds good to me, will make the tests more concise. I don't think there are any tests where it actually matters that the `Server` header is present or that `response.Headers.Clear()` is called, but please be mindful of that.

test/Microsoft.AspNetCore.Server.KestrelTests/FrameResponseHeadersTests.cs
khellang: This will be covered by the test that verifies that the initial dictionary is empty.

khellang:
⌚ 👀

khellang:
🆙📅 - Ready for another review! // @cesarbs

test/Microsoft.AspNetCore.Server.KestrelTests/BadHttpRequestTests.cs
```diff
@@ -103,10 +103,9 @@ public class BadHttpRequestTests
             await connection.Receive(
                 "Connection: close",
                 "");
-            await connection.ReceiveStartsWith("Date: ");
             await connection.ReceiveEnd(
+                $"Date: {connection.Server.Context.DateHeaderValue}",
                 "Content-Length: 0",
-                "Server: Kestrel",
```
halter73: Are there any remaining tests the verify that the "Server: Kestrel" header is sent by default?

src/Microsoft.AspNetCore.Server.Kestrel/Http/Frame.cs
```diff
@@ -170,21 +170,23 @@ public bool HasResponseStarted
...
 
         public void InitializeHeaders()
         {
-            if (_frameHeaders == null)
+            if (FrameRequestHeaders == null)
             {
-                _frameHeaders = new Headers(ServerOptions);
-                RequestHeaders = _frameHeaders.RequestHeaders;
-                ResponseHeaders = _frameHeaders.ResponseHeaders;
+                RequestHeaders = FrameRequestHeaders = new FrameRequestHeaders();
```
halter73: This looks kind of gross. Can we just change the type of the `RequestHeaders` and `ResponseHeaders` properties from `IHeaderDictionary` to `FrameRequestHeaders` and `FrameResponseHeaders` and use those properties everywhere?

khellang: I'm not sure it gets much better. You'd get into issues with the explicit implementations of `IDictionary<string, StringValues>` in `FrameHeaders` and casting in Frame.FeatureCollection. Example:

```csharp
IHeaderDictionary IHttpRequestFeature.Headers
{
    get
    {
        return RequestHeaders;
    }

    set
    {
        // This won't work if RequestHeaders is FrameRequestHeaders. Cast?
        RequestHeaders = value;
    }
}
```

halter73: Settable features. What a pain! It makes sense now though.

Is `RequestHeaders` even reset between requests? It doesn't look like it, and I think it needs to be. This seems to be a preexisting issue, but we might as well fix it now that we're looking at it.

It looks like `ResponseHeaders` is reset in `ProduceEnd` which works I guess.

@Tratcher Why is `IHttpResponseFeature.Headers` settable to begin with? It's not like Kestrel would send those headers to begin with.

halter73:
I really think we need a test that verifies the "Server: Kestrel" header is sent by default.

khellang:
@halter73 Yep. Added a test that checks the default headers :smile:

test/Microsoft.AspNetCore.Server.KestrelTests/DefaultHeaderTests.cs
```diff
@@ -0,0 +1,41 @@
...
+                {
+                    await connection.SendEnd(
+                        "GET / HTTP/1.0",
+                        "",
+                        "");
+
+                    await connection.ReceiveEnd(
+                        "HTTP/1.1 200 OK",
+                        $"Date: {testContext.DateHeaderValue}",
+                        "Server: Kestrel",
```
halter73: Awesome!

halter73:
I guess resetting `IHttpRequestFeature.Headers` between requests can be done in another PR. :ship:

src/Microsoft.AspNetCore.Server.Kestrel/Http/Frame.cs
Tratcher: @halter73 largely because everything else is. in theory you could wrap the headers collection and apply filters/transforms. No, I would not expect Kestrel to send headers from a new dictionary.

khellang:
Rebased and resolved conflicts.

> I guess resetting IHttpRequestFeature.Headers between requests can be done in another PR. 🚢

Filed #879

src/Microsoft.AspNetCore.Server.Kestrel/Http/Frame.cs
khellang: @halter73 Uuuh. Isn't [this](https://github.com/khellang/KestrelHttpServer/blob/4f036a04353640f60703370dea45748d600d2b30/src/Microsoft.AspNetCore.Server.Kestrel/Http/Frame.cs#L264) resetting `RequestHeaders`? `FrameRequestHeaders` and `RequestHeaders` are the same, [remember](https://github.com/khellang/KestrelHttpServer/blob/4f036a04353640f60703370dea45748d600d2b30/src/Microsoft.AspNetCore.Server.Kestrel/Http/Frame.cs#L221)? :wink: Which in turn [is the same as](https://github.com/khellang/KestrelHttpServer/blob/4f036a04353640f60703370dea45748d600d2b30/src/Microsoft.AspNetCore.Server.Kestrel/Http/Frame.FeatureCollection.cs#L159) `IHttpRequestFeature.Headers`...

cesarbs:
Looks good :shipit:

src/Microsoft.AspNetCore.Server.Kestrel/Http/Frame.cs
halter73: @khellang I was referring specifically to the issue you filed in #879 which is that the `RequestHeaders` reference should be reset to `FrameRequestHeaders` like the `ResponseHeaders` reference is reset to `FrameResponseHeaders` in `ProduceEnd`. Not that `Reset()` needs to be called on the `RequestHeaders` object itself.

halter73:
LGTM too. Can you squash the commits?

khellang:
> Can you squash the commits?

Done 👍

halter73:
Thanks!

PR #863, opened 2016-05-21
Fix the sync-completing async fast paths

### LINKED ISSUE #864: Async sync fast-paths aren't fast pathing
ProduceStartAndFireOnStarting only goes async if the user has set an `OnStarting` task - which may also complete sync. 

In normal circumstances `_onStarting != null` which means it returns a `CompletedTask` and doesn't need to go async or await.

The mechanism for fast-pathing this is already there, however its checking the wrong thing (whether a function needs to be called, rather than whether the task the function returns  `RanToCompletion`)

The red below shows the unnecessary call stack this creates (5% of WriteAsync):

too much await

benaadams:
ProduceStartAndFireOnStarting only goes async if the user has set an `OnStarting` task - which may also complete sync. 

In normal circumstances `_onStarting != null` which means it returns a `CompletedTask` and doesn't need to go async or await.

The mechanism for fast-pathing this is already there, however its checking the wrong thing (whether a function needs to be called, rather than whether the task the function returns  `RanToCompletion`)

The red below shows the unnecessary call stack this creates (drops 5% of WriteAsync perf on the extras)

too much await

Post changes the call chains are much reduced

post-await-fastpath

Resolves  #864

/cc @halter73 @cesarbs @rynowak

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
@@ -455,9 +455,27 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Http
             SocketOutput.Write(_emptyData);
         }
 
-        public async Task FlushAsync(CancellationToken cancellationToken)
+        public Task FlushAsync(CancellationToken cancellationToken)
         {
-            await ProduceStartAndFireOnStarting();
+            var produceStartTask = ProduceStartAndFireOnStarting();
+            // ProduceStartAndFireOnStarting normally returns a CompletedTask
+            if (produceStartTask.Status != TaskStatus.RanToCompletion)
+            {
+                // If the Task was not completed go async and await the task
+                // to surface any errors, cancellation or wait for the Task 
+                // to complete before calling SocketOutput.WriteAsync
+                return FlushAsyncAwaited(produceStartTask, cancellationToken);
+            }
+
+            // ProduceStartAndFireOnStarting completed sync 
+            // fast-path by not constructing an async statemachine, 
+            // examining the various contexts and just return the final Write Task
+            return SocketOutput.WriteAsync(_emptyData, cancellationToken: cancellationToken);
+        }
+
+        public async Task FlushAsyncAwaited(Task produceStartTask, CancellationToken cancellationToken)
+        {
+            await produceStartTask;
             await SocketOutput.WriteAsync(_emptyData, cancellationToken: cancellationToken);
         }
 
@@ -483,9 +501,19 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Http
         {
             if (!HasResponseStarted)
             {
-                return WriteAsyncAwaited(data, cancellationToken);
+                var produceStartTask = ProduceStartAndFireOnStarting();
+                // ProduceStartAndFireOnStarting normally returns a CompletedTask
+                if (produceStartTask.Status != TaskStatus.RanToCompletion)
+                {
+                    // If the Task was not completed go async and await the task
+                    // to surface any errors, cancellation or wait for the Task 
+                    // to complete before calling SocketOutput.WriteAsync
+                    return WriteAsyncAwaited(produceStartTask, data, cancellationToken);
+                }
             }
 
+            // Otherwise fast-path by not constructing an async statemachine, 
+            // examining the various contexts and just return the final Write Task
             if (_autoChunk)
             {
                 if (data.Count == 0)
@@ -500,9 +528,9 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Http
             }
         }
 
-        public async Task WriteAsyncAwaited(ArraySegment<byte> data, CancellationToken cancellationToken)
+        public async Task WriteAsyncAwaited(Task produceStartTask, ArraySegment<byte> data, CancellationToken cancellationToken)
         {
-            await ProduceStartAndFireOnStarting();
+            await produceStartTask;
 
             if (_autoChunk)
             {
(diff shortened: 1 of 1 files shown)


benaadams:
Same with WriteAutoChunkSuffix

benaadams:
Also FlushAsync

benaadams:
Also ProduceEnd

benaadams:
Post changes the call chains are much reduced

post-await-fastpath

src/Microsoft.AspNetCore.Server.Kestrel/Http/Frame.cs
```diff
@@ -615,18 +639,21 @@ protected Task ProduceEnd()
 
             if (!HasResponseStarted)
             {
-                return ProduceEndAwaited();
+                ProduceStart(appCompleted: true);
+                // Force flush
+                var flushTask = SocketOutput.WriteAsync(_emptyData);
```
halter73: `WriteAsync` never needs to be awaited in this case.

src/Microsoft.AspNetCore.Server.Kestrel/Http/Frame.cs
```diff
@@ -400,19 +400,39 @@ public void OnCompleted(Func<object, Task> callback, object state)
 
         public void Flush()
         {
-            ProduceStartAndFireOnStarting().GetAwaiter().GetResult();
+            var produceStartTask = ProduceStartAndFireOnStarting();
+            if (produceStartTask.Status != TaskStatus.RanToCompletion)
+            {
+                produceStartTask.GetAwaiter().GetResult();
```
halter73: Is this pattern measurably faster? The awaiter is a single field stack allocated struct and GetResult() no-ops after a single if condition like the one above.

src/Microsoft.AspNetCore.Server.Kestrel/Http/Frame.cs
```diff
@@ -400,19 +400,39 @@ public void OnCompleted(Func<object, Task> callback, object state)
...
+            var produceStartTask = ProduceStartAndFireOnStarting();
+            if (produceStartTask.Status != TaskStatus.RanToCompletion)
+            {
+                return FlushAsyncAwaited(produceStartTask, cancellationToken);
+            }
+
+            return SocketOutput.WriteAsync(_emptyData, cancellationToken: cancellationToken);
+        }
+
+        public async Task FlushAsyncAwaited(Task produceStartTask, CancellationToken cancellationToken)
```
halter73: So is this basically the same change you made to MessageBody where you used `ValueTask<int>`? I assume you're using normal tasks here since the Tasks here aren't generic.

src/Microsoft.AspNetCore.Server.Kestrel/Http/Frame.cs
benaadams: Prevents an unobserved task exception or having to use `#pragma warning disable 4014`

src/Microsoft.AspNetCore.Server.Kestrel/Http/Frame.cs
benaadams: Will drop the sync changes

src/Microsoft.AspNetCore.Server.Kestrel/Http/Frame.cs
benaadams: Its delaying invoking the `async` part and calling into a method that the compiler has created as a state machine unless the task has not completed. Similar to directly returning the task where there is only one await; rather than `async`ing the method and awaiting it.

Some places there is no value in this optimization, for example `RequestProcessingAsync` as there is already a statemachine and will always be one and within that context `await` already does this optimization. However in these paths the statemachine creation can be avoided all together.

src/Microsoft.AspNetCore.Server.Kestrel/Http/Frame.cs
benaadams: Removed them

src/Microsoft.AspNetCore.Server.Kestrel/Http/Frame.cs
benaadams: `WriteAsync`+`WriteAsyncAwaited` are already there; however currently it will always call into `WriteAsyncAwaited`, construct a statemachine and `await` as the request hasn't started.

What it should be checking is if `ProduceStartAndFireOnStarting` produces an incompleted Task; if so then construct a statemachine and `await` it; otherwise just move onto the next step; which it will almost always do.

cesarbs: I don't know whether we will take this change or not, but please add comments documenting it so in the future we know why `FlushAsync` is the way it is. Otherwise all this knowledge gets buried in GitHub discussions.

benaadams: Added comments

cesarbs:
@benaadams Thanks!

benaadams:
Rebased

benaadams:
Rebased

benaadams:
> log  : Installing Microsoft.Extensions.Logging.Abstractions 1.1.0-alpha1-21645.
error: Unable to resolve 'Microsoft.Extensions.Configuration.Abstractions (>= 1.1.0-alpha1-21645)' for '.NETFramework,Version=v4.5.1'.
error: Unable to resolve 'Microsoft.Extensions.Configuration.Abstractions (>= 1.1.0-alpha1-21645)' for '.NETCoreApp,Version=v1.0'.

benaadams:
Re-rebased

cesarbs:
Will benchmark.

cesarbs:
dev:

    [04:22:14.678] Starting scenario Plaintext on benchmark server...
    [04:22:31.114] Warmup
    [04:22:31.121] Starting scenario Plaintext on benchmark client...
    [04:22:42.235] Scenario Plaintext completed on benchmark client
    [04:22:42.235] RPS: 1076079.28
    [04:22:42.235] Stopping scenario Plaintext on benchmark client...
    [04:22:42.236] Benchmark
    [04:22:42.237] Starting scenario Plaintext on benchmark client...
    [04:22:53.328] Scenario Plaintext completed on benchmark client
    [04:22:53.328] RPS: 1140837.3
    [04:22:53.329] Stopping scenario Plaintext on benchmark client...
    [04:22:53.330] Stopping scenario Plaintext on benchmark server...

This PR:

    [04:22:58.733] Starting scenario Plaintext on benchmark server...
    [04:23:20.307] Warmup
    [04:23:20.307] Starting scenario Plaintext on benchmark client...
    [04:23:31.477] Scenario Plaintext completed on benchmark client
    [04:23:31.477] RPS: 1097788.21
    [04:23:31.477] Stopping scenario Plaintext on benchmark client...
    [04:23:31.477] Benchmark
    [04:23:31.477] Starting scenario Plaintext on benchmark client...
    [04:23:42.607] Scenario Plaintext completed on benchmark client
    [04:23:42.607] RPS: 1142243.09
    [04:23:42.607] Stopping scenario Plaintext on benchmark client...
    [04:23:42.607] Stopping scenario Plaintext on benchmark server...

I'm being such a buzzkill today...

benaadams:
1405 more; that's got to be at least 10x php or something? :stuck_out_tongue_winking_eye: 

I think I'm off my game...

PR #868, opened 2016-05-23
Remove Console.WriteLine() from MemoryPoolBlockTests

mikeharder:
- Causes noise in test results

@cesarbs, @halter73


cesarbs:
:shipit:

halter73:
I removed a bunch of `WriteLines` a few months back, but these were the ones that I thought could potentially useful. Now that we are testing the `Vector` path even if `Vector.IsHardwareAccelerated == False`, I guess I'm OK with removing it.

On my machine, .NET 4.5.1 still doesn't have hardware accelerated vectors unlike AppVeyor which is good to know.

mikeharder:
@halter73: I'm not opposed to tests doing something different based on `Vector.IsHardwareAccelerated`, but callilng `Console.WriteLine()` from a test is usually the wrong approach, since there's no guarantee anyone is actually looking at the output, or knows what it's supposed to mean.  Should we instead have a skip condition or something similar based on `Vector.IsHardwareAccelerated`?

PR #869, opened 2016-05-24
Cache MessageBody across frames in same connection

benaadams:
Saves 1M object allocations and 48MB per million requests

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Kestrel/Http/MessageBody.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/Http/MessageBody.cs
@@ -2,12 +2,10 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
-using System.IO;
 using System.Numerics;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Server.Kestrel.Infrastructure;
-using Microsoft.AspNetCore.Server.Kestrel.Exceptions;
 
 namespace Microsoft.AspNetCore.Server.Kestrel.Http
 {
@@ -99,6 +97,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Http
         public abstract ValueTask<int> ReadAsyncImplementation(ArraySegment<byte> buffer, CancellationToken cancellationToken);
 
         public static MessageBody For(
+            MessageBody lastMessageBody,
             string httpVersion,
             FrameRequestHeaders headers,
             Frame context)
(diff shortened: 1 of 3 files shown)


benaadams:
Would need more tests; but I doubt it will be merged any time soon...

src/Microsoft.AspNetCore.Server.Kestrel/Http/MessageBody.cs
```diff
@@ -129,15 +134,32 @@ public Task Consume(CancellationToken cancellationToken = default(CancellationTo
...
                     return new ForContentLength(keepAlive, contentLength, context);
                 }
             }
 
             if (keepAlive)
             {
+                var messageBody = lastMessageBody as ForContentLength;
+                if (messageBody != null)
+                {
+                    messageBody.Reset(keepAlive, 0);
```
khellang: I don't get it... What's the point of resetting this when you're returning a new instance below anyway?

benaadams: Because I'm an idiot... :disappointed:

benaadams: Fixed idiocy... :smile:

src/Microsoft.AspNetCore.Server.Kestrel/Http/MessageBody.cs
```diff
@@ -129,15 +135,34 @@ public Task Consume(CancellationToken cancellationToken = default(CancellationTo
...
+                    return messageBody;
+                }
+
                 return new ForContentLength(true, 0, context);
             }
 
+            if (lastMessageBody != null && lastMessageBody is ForRemainingData)
```
halter73: `lastMessageBody is ForRemainingData`? isn't that impossible?

halter73:
I don't think we're ready to take this kind of change for 1.0.0.

src/Microsoft.AspNetCore.Server.Kestrel/Http/MessageBody.cs
benaadams: Probably

benaadams:
> I don't think we're ready to take this kind of change for 1.0.0.

I assume not as it is a behavioural change

halter73:
@benaadams Why is this a behavioral change? We're already caching `RequestBody`.

benaadams:
Its reusing an object that's wasn't reused before; its not a great leap but it is a change

halter73:
I'm closing this then. I don't think the impact is that significant. Post 1.0 we might revisit.

Issue #766, opened 2016-04-24
Error -4089 EAFNOSUPPORT address family not supported on Azure App Service deployment


TsengSR:
I am using the rc1-update1 to deploy an application to Azure App Service (free plan for testing). 

Using dnx-clr-win-x86.1.0.0-rc1-update1. After deploying I can't connect to the url, it just times out. 

Executing web.cmd from the Azure Commandline returns

```
Microsoft.AspNet.Server.Kestrel.Networking.UvException: Error -4089 EAFNOSUPPORT address family not supported
   at Microsoft.AspNet.Server.Kestrel.Networking.Libuv.Check(Int32 statusCode)
   at Microsoft.AspNet.Server.Kestrel.Networking.Libuv.tcp_bind(UvTcpHandle handle, sockaddr& addr, Int32 flags)
   at Microsoft.AspNet.Server.Kestrel.Networking.UvTcpHandle.Bind(ServerAddress address)
   at Microsoft.AspNet.Server.Kestrel.Http.TcpListener.CreateListenSocket()
   at Microsoft.AspNet.Server.Kestrel.Http.Listener.<>c__DisplayClass5_0.<StartAsync>b__0(Object _)
```

Not sure what's causing it, I've tried changing the ports of kestrel and just using one url, but nothing helped.

TsengSR:
Got it. I used http://*:8081 as port. This seems to want to bind IPv6 too which the app service instance do not seem to support. http://0.0.0.0:8081 worked though

halter73:
Thanks for the report. This does seem like a bug semi-related to https://github.com/aspnet/KestrelHttpServer/issues/231. If we can't bind to `::` (which also binds to `0.0.0.0`), we should try binding to `0.0.0.0` explicitly in [UvTcpHandle.Bind](https://github.com/aspnet/KestrelHttpServer/blob/179b57b01f73b7b10ddce87cd0e57d196922462b/src/Microsoft.AspNetCore.Server.Kestrel/Networking/UvTcpHandle.cs#L27).

I'm glad you found a workaround for now.

mikeharder:
I'm unable to reproduce this in either Azure App Service or self-host.

For Azure App Service, in RC2 the configuration code was changed so `UseIISIntegration()` overrides `UseUrls()` (if it is called *after*).  So applications configured with `UseUrls("http://*:5000").UseIISIntegration()` work fine in App Service, regardless of IPv6.

[...]

PR #870, opened 2016-05-24
Bind to both IPv4 and IPv6 when localhost is specified (#231).

cesarbs:
#231

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Kestrel/KestrelServer.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/KestrelServer.cs
@@ -4,6 +4,7 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Net;
 using System.Reflection;
 using Microsoft.AspNetCore.Hosting;
 using Microsoft.AspNetCore.Hosting.Server;
@@ -11,6 +12,7 @@ using Microsoft.AspNetCore.Hosting.Server.Features;
 using Microsoft.AspNetCore.Http.Features;
 using Microsoft.AspNetCore.Server.Kestrel.Http;
 using Microsoft.AspNetCore.Server.Kestrel.Infrastructure;
+using Microsoft.AspNetCore.Server.Kestrel.Networking;
 using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Options;
 
@@ -94,6 +96,10 @@ namespace Microsoft.AspNetCore.Server.Kestrel
                 {
                     _logger.LogWarning("Unable to determine ECONNRESET value on this platform.");
                 }
+                if (!Constants.EADDRINUSE.HasValue)
+                {
+                    _logger.LogWarning("Unable to determine EADDRINUSE value on this platform.");
+                }
 
                 engine.Start(threadCount);
                 var atLeastOneListener = false;
@@ -108,8 +114,69 @@ namespace Microsoft.AspNetCore.Server.Kestrel
                     else
                     {
                         atLeastOneListener = true;
-                        _disposables.Push(engine.CreateServer(
-                            parsedAddress));
+
+                        if (!parsedAddress.Host.Equals("localhost", StringComparison.OrdinalIgnoreCase))
+                        {
+                            _disposables.Push(engine.CreateServer(
+                                parsedAddress));
+                        }
+                        else
+                        {
+                            if (parsedAddress.Port == 0)
+                            {
+                                throw new InvalidOperationException("Dynamic port binding is not supported when binding to localhost. You must either bind to 127.0.0.1:0 or [::1]:0, or both.");
+                            }
+
+                            var ipv4Address = parsedAddress.WithHost("127.0.0.1");
+                            var exceptions = new List<UvException>();
+
+                            try
+                            {
+                                _disposables.Push(engine.CreateServer(ipv4Address));
+                            }
+                            catch (AggregateException ex)
+                            {
+                                var uvException = ex.InnerException as UvException;
+
+                                if (uvException != null && uvException.StatusCode != Constants.EADDRINUSE)
+                                {
+                                    _logger.LogWarning(0, ex, $"Unable to bind to {parsedAddress.ToString()} on the IPv4 loopback interface.");
+                                    exceptions.Add(uvException);
+                                }
+                                else
+                                {
+                                    throw;
+                                }
+                            }
+
+                            var ipv6Address = parsedAddress.WithHost("[::1]");
+
+                            try
+                            {
+                                _disposables.Push(engine.CreateServer(ipv6Address));
+                            }
+                            catch (AggregateException ex)
+                            {
+                                var uvException = ex.InnerException as UvException;
+
+                                if (uvException != null && uvException.StatusCode != Constants.EADDRINUSE)
+                                {
+                                    _logger.LogWarning(0, ex, $"Unable to bind to {parsedAddress.ToString()} on the IPv6 loopback interface.");
+                                    exceptions.Add(uvException);
+                                }
+                                else
+                                {
+                                    throw;
+                                }
+                            }
+
+                            if (exceptions.Count == 2)
+                            {
+                                var ex = new AggregateException(exceptions);
+                                _logger.LogError(0, ex, $"Unable to bind to {parsedAddress.ToString()} on any loopback interface.");
+                                throw ex;
+                            }
+                        }
 
                         // If requested port was "0", replace with assigned dynamic port.
                         _serverAddresses.Addresses.Remove(address);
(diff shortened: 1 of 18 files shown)


src/Microsoft.AspNetCore.Server.Kestrel/KestrelServer.cs
```diff
@@ -108,8 +114,45 @@ public void Start<TContext>(IHttpApplication<TContext> application)
                     else
                     {
                         atLeastOneListener = true;
-                        _disposables.Push(engine.CreateServer(
-                            parsedAddress));
+
+                        if (parsedAddress.Host.Equals("localhost", StringComparison.OrdinalIgnoreCase))
+                        {
+                            var exceptions = new List<Exception>();
```
halter73: Creating and adding to a list just to get a count? Why not just use a counter?

src/Microsoft.AspNetCore.Server.Kestrel/KestrelServer.cs
```diff
@@ -108,8 +114,45 @@ public void Start<TContext>(IHttpApplication<TContext> application)
...
+                                    exceptions.Add(ex);
+                                }
+                            }
+
+                            if (exceptions.Count == 2)
+                            {
+                                throw new InvalidOperationException("Binding to localhost failed to bind to any loopback interface.");
+                            }
+
+                            parsedAddress.Host = originalHost;
```
halter73: Just new up new server addresses instead of doing all this swapping/mutating.

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/AddressRegistrationTests.cs
```diff
@@ -84,6 +84,45 @@ public class AddressRegistrationTests
...
+        }
+
+        private void ThrowsWhenBindingLocalhostToAddressInUse(AddressFamily addressFamily, IPAddress address)
+        {
+            using (var socket = new Socket(addressFamily, SocketType.Stream, ProtocolType.Tcp))
+            {
+                var port = GetNextPort();
+                socket.Bind(new IPEndPoint(address, port));
+
+                var config = new ConfigurationBuilder()
```
davidfowl: Remove this and just use `UseUrls` on WebHostBuilder

src/Microsoft.AspNetCore.Server.Kestrel/KestrelServer.cs
```diff
@@ -108,8 +114,45 @@ public void Start<TContext>(IHttpApplication<TContext> application)
...
+                            {
+                                try
+                                {
+                                    parsedAddress.Host = loopbackAddress.Address;
+                                    _disposables.Push(engine.CreateServer(
+                                        parsedAddress));
+                                }
+                                catch (UvException ex) when (ex.StatusCode != Constants.EADDRINUSE)
+                                {
+                                    _logger.LogWarning(0, ex, $"Unable to bind {address} to loopback {loopbackAddress.Family} address.");
```
Tratcher: Shouldn't this just say Unable to bind to {address}?

src/Microsoft.AspNetCore.Server.Kestrel/KestrelServer.cs
cesarbs: Yeah that was different before and I didn't revert this. I was throwing an `AggregateException`.

cesarbs: Do you think that might be better?

src/Microsoft.AspNetCore.Server.Kestrel/KestrelServer.cs
cesarbs: `address` would be `http://localhost:someport/something`. The idea is to inform that we failed to bind to a given address when the user asked to bind to localhost.

src/Microsoft.AspNetCore.Server.Kestrel/KestrelServer.cs
cesarbs: I'm using the same object to get `Port` updated in case it was 0. Otherwise binding to `localhost:0` keeps the port as 0 in the end.

halter73:
Travis is producing a bunch of `Error -99 EADDRNOTAVAIL address not available` errors in travis tests on both OS X and Linux: https://travis-ci.org/aspnet/KestrelHttpServer/jobs/132455305

src/Microsoft.AspNetCore.Server.Kestrel/KestrelServer.cs
```diff
@@ -108,8 +114,51 @@ public void Start<TContext>(IHttpApplication<TContext> application)
...
+                            catch (AggregateException ex) when (ex.InnerException is UvException &&
+                                                                (ex.InnerException as UvException).StatusCode != Constants.EADDRINUSE)
+                            {
+                                _logger.LogWarning(0, ex, $"Unable to bind to {parsedAddress.ToString()} on the IPv4 loopback interface.");
+                                exceptions.Add(ex.InnerException as UvException);
+                            }
+
+                            // Make sure we bind to the same port on port 0.
+                            var port = ipv4Address.Port;
+                            var ipv6Address = parsedAddress.WithHost("[::1]").WithPort(port);
```
halter73: Much more clear now :+1:

src/Microsoft.AspNetCore.Server.Kestrel/KestrelServer.cs
```diff
@@ -108,8 +114,51 @@ public void Start<TContext>(IHttpApplication<TContext> application)
...
+                            parsedAddress.Port = port;
+
+                            try
+                            {
+                                _disposables.Push(engine.CreateServer(ipv6Address));
+                            }
+                            catch (AggregateException ex) when (ex.InnerException is UvException &&
```
halter73: Just use `if` and `throw

cesarbs:
No more filters.

halter73:
:shipit:

cesarbs:
Updated to disallow binding to `localhost:0`. cc @davidfowl

src/Microsoft.AspNetCore.Server.Kestrel/KestrelServer.cs
```diff
@@ -108,8 +114,72 @@ public void Start<TContext>(IHttpApplication<TContext> application)
...
+                                    exceptions.Add(uvException);
+                                }
+                                else
+                                {
+                                    throw;
+                                }
+                            }
+
+                            // Make sure we bind to the same port on port 0.
+                            var port = ipv4Address.Port;
```
mikeharder: You can remove this logic since we no longer support `localhost:0`.

cesarbs:
:bell:

src/Microsoft.AspNetCore.Server.Kestrel/Networking/UvTcpHandle.cs
```diff
@@ -81,7 +81,6 @@ public void NoDelay(bool enable)
         /// </summary>
         public static IPEndPoint CreateIPEndpoint(ServerAddress address)
         {
-            // TODO: IPv6 support
```
mikeharder: I would leave this comment in for now.  I can remove it when I fix #766.

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/ThreadCountTests.cs
```diff
@@ -16,19 +15,12 @@ public class ThreadCountTests
...
-                })
-                .Build();
-
             var hostBuilder = new WebHostBuilder()
-                .UseConfiguration(config)
                 .UseKestrel(options =>
                 {
                     options.ThreadCount = threadCount;
                 })
+                .UseUrls("http://localhost:0/")
```
mikeharder: Should this be changed to `127.0.0.1:0`?  How is it even passing with `localhost:0`?

mikeharder:
You can :shipit: after addressing my comments.  Or, you could wait until I figure out the design for #766, since it might impact this change as well.

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/ThreadCountTests.cs
cesarbs: It's passing because it calls `KestrelEngine.CreateServer` directly. Should I move the check there? I wouldn't want it to have it in two places._

cesarbs: Other tests we doing that. This one was simply not marked as a test :laughing:

PR #871, opened 2016-05-24
Init date values in ctor

### LINKED ISSUE #866: System.NullReferenceException: Object reference not set to an instance of an object at .Kestrel.Infrastructure.Headers.Initialize
I am running the following exception when running MusicStore app

```warn: Microsoft.AspNetCore.Server.Kestrel[0]
      Connection processing ended abnormally
warn: Microsoft.AspNetCore.Server.Kestrel[0]
      Connection processing ended abnormally
warn: Microsoft.AspNetCore.Server.Kestrel[0]
      Connection processing ended abnormally
System.NullReferenceException: Object reference not set to an instance of an object.
   at Microsoft.AspNetCore.Server.Kestrel.Infrastructure.Headers.Initialize(DateHeaderValueManager dateValueManager)
   at Microsoft.AspNetCore.Server.Kestrel.Http.Frame`1.<RequestProcessingAsync>d__2.MoveNext()
System.NullReferenceException: Object reference not set to an instance of an object.
   at Microsoft.AspNetCore.Server.Kestrel.Infrastructure.Headers.Initialize(DateHeaderValueManager dateValueManager)
   at Microsoft.AspNetCore.Server.Kestrel.Http.Frame`1.<RequestProcessingAsync>d__2.MoveNext()
System.NullReferenceException: Object reference not set to an instance of an object.
   at Microsoft.AspNetCore.Server.Kestrel.Infrastructure.Headers.Initialize(DateHeaderValueManager dateValueManager)
   at Microsoft.AspNetCore.Server.Kestrel.Http.Frame`1.<RequestProcessingAsync>d__2.MoveNext()```

benaadams:
This change sets the object in the DateHeader .ctor rather than waiting for the timer tick to initialize it.

Start up race that seemed to work prior to https://github.com/aspnet/KestrelHttpServer/pull/810 by using an null array for the date header value which `FrameResponseHeaders.CopyToFast` [would skip](https://github.com/aspnet/KestrelHttpServer/blob/dev/src/Microsoft.AspNetCore.Server.Kestrel/Http/FrameHeaders.Generated.cs#L9038-L9050) therefore not outputting a date header. So an invalid response, however not a failure.

With the date bytes and date string now in a object; that object is null - which causes the null reference when resolving the bytes at start-up.

Resolves  #866

/cc @halter73

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Kestrel/Http/DateHeaderValueManager.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/Http/DateHeaderValueManager.cs
@@ -49,6 +49,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Http
             _timeWithoutRequestsUntilIdle = timeWithoutRequestsUntilIdle;
             _timerInterval = timerInterval;
             _dateValueTimer = new Timer(TimerLoop, state: null, dueTime: Timeout.Infinite, period: Timeout.Infinite);
+            SetDateValues(systemClock.UtcNow);
         }
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


halter73:
I prefer #872 as this leaves the race when the timer is restarting and if the `DateHeaderValueManager` isn't used for a little while after being constructed.

PR #874, opened 2016-05-24
ForChunkedEncoding vector as static rather than instance

benaadams:
`_vectorCRs` only needs to be allocated once rather than per object

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Kestrel/Http/MessageBody.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/Http/MessageBody.cs
@@ -2,12 +2,10 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
-using System.IO;
 using System.Numerics;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Server.Kestrel.Infrastructure;
-using Microsoft.AspNetCore.Server.Kestrel.Exceptions;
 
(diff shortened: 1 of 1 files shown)


halter73:
There be dragons hiding behind even the safest seeming changes. Look at all the ChunkedRequestsTests that failed on AppVeyor: https://ci.appveyor.com/project/aspnet-ci/kestrelhttpserver/build/1.0.1646/tests

It turns out that your change causes an `InvalidProgramException` to be thrown when `ForChunkedEncoding` is initialized on desktop CLR when `Vector.IsHardwareAccelerated == True`.

Interestingly if you add a second static vector to the `ForChunkedEncoding` everything seems to work correctly. This might be why we haven't seen this issue in `Frame`. This doesn't repro on my laptop since I don't get hardware acceleration of the desktop framework, but @cesarbs can repro it on his machine. We're looking into filing an issue with a simplified repo to dotnet.

benaadams:
@halter73 hmm that rings a bell... https://github.com/aspnet/KestrelHttpServer/commit/dab1a1fa8df6383c35777c27f0bb3ead68bdb0d8

benaadams:
Probably needs a comment on that line it; I git blame myself :disappointed:

benaadams:
Can't find any issue I raised for it either :-1:

cesarbs:
It's super easy to repro. Filed https://github.com/dotnet/corefx/issues/8825.

halter73:
While we could do something adventurous like adding another static Vector to ForChunkedEncoding, I think we just add a comment for now. #877

PR #880, opened 2016-05-25
Implement IHttpRequestFeature.RawTarget (aspnet/HttpAbstractions#596)

cesarbs:
aspnet/HttpAbstractions#596

Needs https://github.com/aspnet/HttpAbstractions/pull/639 to be merged to work.


test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/RequestTests.cs
```diff
@@ -204,6 +205,58 @@ public void RequestPathIsNormalized()
             }
         }
 
+        [Theory]
+        [InlineData("/")]
```
Tratcher: `*`?

Tratcher:
:+1:

cesarbs:
Added test for paths not starting with `'/'`.

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/RequestTests.cs
```diff
@@ -202,6 +203,108 @@ public void RequestPathIsNormalized()
...
+                    Assert.StartsWith("HTTP/1.1 200 OK", response.ToString());
+                }
+            }
+        }
+
+        [Theory]
+        [InlineData("*")]
+        [InlineData("DoesNotStartWith/")]
+        [InlineData("*?arg=value")]
+        [InlineData("*/?arg=value")]
```
halter73: `[InlineData("*%2F")]`?

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/RequestTests.cs
```diff
@@ -202,6 +203,108 @@ public void RequestPathIsNormalized()
...
+        [Theory]
+        [InlineData("/")]
+        [InlineData("/path")]
+        [InlineData("/path?foo=1&bar=2")]
+        [InlineData("/hello%20world")]
+        [InlineData("/hello%20world?foo=1&bar=2")]
+        [InlineData("/base/path")]
+        [InlineData("/base/path?foo=1&bar=2")]
+        [InlineData("/base/hello%20world")]
+        [InlineData("/base/hello%20world?foo=1&bar=2")]
```
halter73: We should also check URLs with dot segments.

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/RequestTests.cs
```diff
@@ -202,6 +203,108 @@ public void RequestPathIsNormalized()
...
+                        Assert.Equal(requestTarget, context.Features.Get<IHttpRequestFeature>().RawTarget);
+                        await context.Response.WriteAsync("hello, world");
+                    });
+                });
+
+            using (var host = builder.Build())
+            {
+                host.Start();
+
+                using (var socket = TestConnection.CreateConnectedLoopbackSocket(host.GetPort()))
```
mikeharder: Is there a reason we directly use `Socket`, rather than a higher-level abstraction like `HttpClient` or `TcpClient`?

src/Microsoft.AspNetCore.Server.Kestrel/Http/Frame.cs
```diff
@@ -902,21 +905,25 @@ protected RequestLineStatus TakeStartLine(SocketInput input)
...
                 QueryString = queryString;
+                RawTarget = rawTarget;
                 HttpVersion = httpVersion;
 
                 bool caseMatches;
-
-                if (!string.IsNullOrEmpty(_pathBase) &&
-                    (requestUrlPath.Length == _pathBase.Length || (requestUrlPath.Length > _pathBase.Length && requestUrlPath[_pathBase.Length] == '/')) &&
-                    RequestUrlStartsWithPathBase(requestUrlPath, out caseMatches))
+                if (RequestUrlStartsWithPathBase(requestUrlPath, out caseMatches))
```
mikeharder: Are these changes related to RawTarget, or just bugs we found when testing RawTarget?

cesarbs: It's a refactoring, that `if` was horrible.

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/RequestTests.cs
cesarbs: Done :+1:

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/RequestTests.cs
cesarbs: We're using `Socket` to make sure we're sending exactly what we want to send to the server. `HttpClient` changes things.

halter73: Yeah. This type of test would normally be done in the other test project despite the name. This is why this version of `TestConnection` is so limited.

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/RequestTests.cs
cesarbs: Done in another test.

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/RequestTests.cs
mikeharder: OK, `TcpClient` might make things a little simpler, but `Socket` is not too complicated in this case.

cesarbs:
Checked in some unintended changes, reverting.

cesarbs:
Updated with a small perf detail.

cesarbs:
Squirrel?

cesarbs:
:bell: I'd like to merge this tonight.

halter73:
Would it be so bad to merge tomorrow? The build's broken.

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/RequestTests.cs
halter73: I think this should be in the other test project for consistency's sake. Also, why not verify the entire response?

cesarbs:
Squashed.

halter73:
This make only the second usage of `FunctionalTests.TestConnection`. I want to remove that class, so can we please not use it?

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/RequestTests.cs
```diff
@@ -202,6 +203,122 @@ public void RequestPathIsNormalized()
...
+                        var length = socket.Receive(buffer);
+                        if (length == 0)
+                        {
+                            break;
+                        }
+
+                        response.Append(Encoding.ASCII.GetString(buffer, 0, length));
+                    }
+
+                    Assert.StartsWith("HTTP/1.1 200 OK", response.ToString());
```
halter73: Testing only `StartsWith` seems wrong. In most of our tests, we verify the entire response is what we expect.

cesarbs:
@halter73 Removed `TestConnection` (the `FunctionalTests` one).

cesarbs:
:bell:

test/Microsoft.AspNetCore.Server.KestrelTests/RequestTargetProcessingTests.cs
```diff
@@ -0,0 +1,131 @@
...
+        [InlineData("DoesNotStartWith/")]
+        [InlineData("DoesNotStartWith/?arg=value")]
+        [InlineData("DoesNotStartWithSlash?arg=value")]
+        [InlineData("./")]
+        [InlineData("../")]
+        [InlineData("../.")]
+        [InlineData(".././")]
+        [InlineData("../..")]
+        [InlineData("../../")]
+        public async Task NonPathRequestTargetSetInRawTarget(string requestTarget)
```
halter73: Would these test cases pass with WebListenter after https://github.com/aspnet/WebListener/pull/199 I mainly wondering whether Path, PathBase and QueryString would all be an empty string.

Should this be added to ServerTests?

cesarbs: WebListener actually rejects those requests. Maybe we should do the same in Kestrel? The only request I was able to make to a target not starting with `/` was `OPTIONS *`.

cesarbs: Also, WebListener turned `*` into `/*` in `Path`.

halter73:
:shipit:

src/Microsoft.AspNetCore.Server.Kestrel/Http/Frame.cs
```diff
@@ -907,28 +914,43 @@ protected RequestLineStatus TakeStartLine(SocketInput input)
...
+                        // decoding and there's no query string following it.
+                        rawTarget = requestUrlPath;
+                    }
+                    else
+                    {
+                        rawTarget = pathBegin.GetAsciiString(queryEnd);
+                    }
                 }
 
                 requestUrlPath = PathNormalizer.RemoveDotSegments(requestUrlPath);
```
halter73: New variable please.

Issue #860, opened 2016-05-21
ServerAddress.FromUrl("http://") throws ArgumentOutOfRangeException


mikeharder:
## Repro
```C#
ServerAddress.FromUrl("http://")
```

## Expected
Return either `null` or a valid `ServerAddress`

## Actual
```
Unhandled Exception: System.ArgumentOutOfRangeException: Length cannot be less than zero.
Parameter name: length
   at System.String.Substring(Int32 startIndex, Int32 length)
   at Microsoft.AspNetCore.Server.Kestrel.ServerAddress.FromUrl(String url)
```

khellang:
Don't you think it should throw *something* for invalid URLs? Returning `null` could lead to an equally vague `NullReferenceException`.

mikeharder:
@khellang: Thanks for the suggestion, we agree ServerAddress.FromUrl() should throw instead of returning null.  Tracked by https://github.com/aspnet/KestrelHttpServer/issues/875.

PR #884, opened 2016-05-27
ServerAddress.FromUrl("http://") throws ArgumentOutOfRangeException (#860)

mikeharder:
- Code change in `ServerAddress.cs` prevents the `ArgumentOutOfRangeException`, and causes `FromUrl("http://")` to return a `ServerAddress` with `Host=""`.
- I considered throwing `FormatException` if `String.IsNullOrEmpty(Host)`, but this raises further questions of what we consider a valid `Host`:
 - Is whitespace a valid host?
 - Is a special character like `!` or `.` a valid host?
 - Is a string containing whitespace (like `foo bar`) a valid host?
- So I think we should allow any host until we decide to start actually validating the host per the RFCs.

@halter73, @cesarbs, @davidfowl, @Tratcher

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Kestrel/ServerAddress.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/ServerAddress.cs
@@ -142,7 +142,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel
             }
 
             // Path should not end with a / since it will be used as PathBase later
-            if (url[url.Length - 1] == '/')
+            if (url[url.Length - 1] == '/' && url.Length > pathDelimiterEnd)
             {
                 serverAddress.PathBase = url.Substring(pathDelimiterEnd, url.Length - pathDelimiterEnd - 1);
             }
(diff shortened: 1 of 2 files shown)


halter73:
You convinced me wrt the empty host. :shipit:

davidfowl:
Why is travis sad?

cesarbs:
LGTM.

I don't like the current situation where you can specify anything you want. Since we're not validating hosts I'd much rather only allow:

* `localhost`
* `*`
* IP addresses
* UNIX socket path

And block everything else. But we probably don't have the time for that at this point.

cesarbs:
@davidfowl Need to fix `travis.yml` to install the openssl dependency on OS X. I'll send a PR in a moment.

cesarbs:
Travis should be good now but this PR needs to be rebased.

mikeharder:
@Tratcher has convinced me we should throw if `Host` is null or empty, but allow anything else.  The reason is we may want to validate the `Host` header in the future, and we should require `ServerAddress.Host` to be non-empty, to avoid things like considering a request valid if it's missing the `Host` header.  Other cases like `Host=" "` or `Host="."` are less problematic, since any non-empty string will be valid to compare to a host header.

mikeharder:
Code now throws if `Host` is null or empty.  Waiting for signoff.

cesarbs:
:ship:

Issue #879, opened 2016-05-24
Reset Frame's IHttpRequestFeature.Headers and IHttpResponseFeature.Headers between requests


khellang:
See https://github.com/aspnet/KestrelHttpServer/pull/862#discussion_r64489390

Also, look into consolidating `RequestHeaders` and `FrameRequestHeaders`. See https://github.com/aspnet/KestrelHttpServer/pull/862#discussion_r64484432

cesarbs:
@muratg I think this is important enough for 1.0.0. @halter73 do you agree?

muratg:
CC @mikeharder & @davidfowl for their thoughts as well.

davidfowl:
Lets do it.

halter73:
This change should be very simple even if it might not affect (m)any real apps.

PR #885, opened 2016-05-27
Reset Frame's IHttpRequestFeature.Headers and IHttpResponseFeature.Headers between requests (#879)

cesarbs:
#879 

cc @halter73 @mikeharder @khellang

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Kestrel/Http/Frame.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/Http/Frame.cs
@@ -11,7 +11,6 @@ using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Http;
-using Microsoft.AspNetCore.Http.Features;
 using Microsoft.AspNetCore.Server.Kestrel.Infrastructure;
 using Microsoft.AspNetCore.Server.Kestrel.Exceptions;
 using Microsoft.Extensions.Logging;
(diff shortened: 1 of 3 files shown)


src/Microsoft.AspNetCore.Server.Kestrel/Http/Frame.cs
```diff
@@ -218,9 +218,11 @@ public void InitializeHeaders()
         {
             if (FrameRequestHeaders == null)
             {
-                RequestHeaders = FrameRequestHeaders = new FrameRequestHeaders();
+                FrameRequestHeaders = new FrameRequestHeaders();
             }
 
+            RequestHeaders = FrameRequestHeaders;
```
khellang: Question; why is this change only for request headers? Response headers are only reset like this [here](https://github.com/aspnet/KestrelHttpServer/blob/8df6fbc5009a6963864f042d552f3c4fdea4f2d7/src/Microsoft.AspNetCore.Server.Kestrel/Http/Frame.cs#L652), which means it'll only happen `if (_requestRejected || _applicationException != null && !HasResponseStarted)`. Am I not seeing something?

halter73: Yep. Resetting in two separate places was silly to begin with because it can allow for oversights like the one I made here. Just reset them both here.

I don't think there is any need to reset it in ProduceEnd. So I would remove that. I would also update IHttpUpgradeFeature.UpgradeAsync to use FrameResponseHeaders while you're at it.

cesarbs:
Now resetting `ResponseHeaders` too.

halter73:
:shipit:

test/Microsoft.AspNetCore.Server.KestrelTests/FrameTests.cs
```diff
@@ -115,5 +116,51 @@ public void ThrowsWhenReasonPhraseIsSetAfterResponseStarted()
             Assert.True(frame.HasResponseStarted);
             Assert.Throws<InvalidOperationException>(() => ((IHttpResponseFeature)frame).ReasonPhrase = "Reason phrase");
         }
+
+        [Fact]
+        public void InitializeHeadersResetsRequestHeaders()
+        {
```
halter73: I would probably combine these tests, but it's up to you.

PR #886, opened 2016-05-27
Test code nitpicks.

cesarbs:
cc @halter73 @mikeharder


halter73:
I guess, why not? :ship:

PR #887, opened 2016-05-27
Remove Frame.RequestUri.

cesarbs:
Nothing uses it.


khellang:
:shipit:

mikeharder:
:shipit:

halter73:
:shipit:

PR #888, opened 2016-05-27
Fix disabled tests in MultipleLoopTests.

cesarbs:
cc @halter73 @mikeharder

### DIFF
--- a/test/Microsoft.AspNetCore.Server.KestrelTests/MultipleLoopTests.cs
+++ b/test/Microsoft.AspNetCore.Server.KestrelTests/MultipleLoopTests.cs
@@ -1,5 +1,7 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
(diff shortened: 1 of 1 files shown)


test/Microsoft.AspNetCore.Server.KestrelTests/MultipleLoopTests.cs
```diff
@@ -97,16 +101,16 @@ public void ServerPipeListenForConnections()
...
                     connect.Dispose();
+
                     clientConnectionPipe.ReadStart(
-                        (_3, cb, _4) => buf,
-                        (_3, status2, _4) =>
+                        (handle2, cb, state2) => buf,
+                        (handle2, status2, state2) =>
                         {
-                            if (status2 == 0)
+                            if (status2 == Constants.EOF)
```
cesarbs: This is the actual fix.

mikeharder:
:shipit:

halter73:
:shipit:

halter73:
Thanks for looking into this!

PR #890, opened 2016-05-27
Replace PlatformAbstractions with RuntimeInformation

BrennanConroy:
Part of https://github.com/aspnet/PlatformAbstractions/issues/34

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Kestrel/Infrastructure/Constants.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/Infrastructure/Constants.cs
@@ -1,7 +1,7 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
-using Microsoft.Extensions.PlatformAbstractions;
+using System.Runtime.InteropServices;
 
 namespace Microsoft.AspNetCore.Server.Kestrel.Infrastructure
 {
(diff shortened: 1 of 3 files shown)


pranavkm:
:shipit: once Travis passes.

mikeharder:
Please wait, I'm still reviewing this.

src/Microsoft.AspNetCore.Server.Kestrel/Infrastructure/Constants.cs
```diff
@@ -28,32 +28,36 @@ internal class Constants
...
-                    return -4077;
-                case Platform.Linux:
-                    return -104;
-                case Platform.Darwin:
-                    return -54;
-                default:
-                    return null;
+               return -4077;
             }
+            if (RuntimeInformation.IsOSPlatform(OSPlatform.Linux))
```
mikeharder: Nit: Change to `else if` with a final `else`.

src/Microsoft.AspNetCore.Server.Kestrel/Infrastructure/Constants.cs
```diff
@@ -28,32 +28,36 @@ internal class Constants
...
+            {
+                return -54;
+            }
+            return null;
         }
 
         private static int? GetEADDRINUSE()
         {
-            switch (PlatformServices.Default.Runtime.OperatingSystemPlatform)
+            if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
```
mikeharder: Nit: Change to else if with a final else.

mikeharder:
:shipit: after my feedback.

src/Microsoft.AspNetCore.Server.Kestrel/Infrastructure/Constants.cs
moozzyk: why?

mikeharder: It's clearer that exactly one condition will match.  It's logically the same if every condition has a `return` statement, but it's still clearer to use us `if / else if / else` rather than `if / if / if`.

moozzyk: In this case I don't have any preference just wanted to know what your thoughts are. In more complicated cases I often times find `else if`s tricky - especially when they introduce new variables and suddenly you need to work on multiple set intersections many of which are not covered by any of the clauses/conditions. In such cases I prefer (if possible and not too nested) an `if` nested in `else` than `else if`. .

halter73:
:shipit:

davidfowl:
:shipit:

PR #896, opened 2016-05-31
Fix NRE when aborting connection or client disconects

pakrym:
@cesarbs @halter73

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Kestrel/Http/SocketInput.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/Http/SocketInput.cs
@@ -27,8 +27,10 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Http
         private MemoryPoolBlock _tail;
         private MemoryPoolBlock _pinned;
 
-        private int _consumingState;
         private object _sync = new object();
+
+        private bool _consuming;
+        private bool _disposed;
 
         public SocketInput(MemoryPool memory, IThreadPool threadPool)
         {
@@ -163,12 +165,15 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Http
 
         public MemoryPoolIterator ConsumingStart()
         {
-            if (Interlocked.CompareExchange(ref _consumingState, 1, 0) != 0)
+            lock (_sync)
             {
-                throw new InvalidOperationException("Already consuming input.");
+                if (_consuming)
+                {
+                    throw new InvalidOperationException("Already consuming input.");
+                }
+                _consuming = true;
+                return new MemoryPoolIterator(_head);
             }
-
-            return new MemoryPoolIterator(_head);
         }
 
         public void ConsumingComplete(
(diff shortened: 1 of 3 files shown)


src/Microsoft.AspNetCore.Server.Kestrel/Http/SocketInput.cs
```diff
@@ -180,25 +180,29 @@ public MemoryPoolIterator ConsumingStart()
 
             lock (_sync)
             {
-                if (!consumed.IsDefault)
+                // object was disposed
```
cesarbs: nit: make this comment more informative e.g. "Make sure object was not disposed", "Check if object was disposed" or "Object may have been disposed"

src/Microsoft.AspNetCore.Server.Kestrel/Infrastructure/TaskUtilities.cs
```diff
@@ -29,6 +29,10 @@ public static Task GetCancelledTask(CancellationToken cancellationToken)
         public static Task<int> GetCancelledZeroTask(CancellationToken cancellationToken = default(CancellationToken))
         {
 #if NETSTANDARD1_3
+            if (!cancellationToken.IsCancellationRequested)
```
cesarbs: Add a comment explaining why you're doing this.

src/Microsoft.AspNetCore.Server.Kestrel/Http/SocketInput.cs
```diff
@@ -180,25 +180,29 @@ public MemoryPoolIterator ConsumingStart()
...
+                if (_head != null)
                 {
-                    returnStart = _head;
-                    returnEnd = consumed.Block;
-                    _head = consumed.Block;
-                    _head.Start = consumed.Index;
-                }
+                    if (!consumed.IsDefault)
```
cesarbs: You only really care whether `_head` is null or not in this first `if`. I'm not sure, but I think you should leave the subsequent `if` block outside of the null check to avoid potential hangs. I'll wait for @halter73's input on this.

cesarbs:
Please benchmark this change against the plaintext scenario.

src/Microsoft.AspNetCore.Server.Kestrel/Http/SocketInput.cs
pakrym: Why would it hang?

cesarbs: You're right.

pakrym:
Both this PR and clear run is  `1,200,000 +- 50,000`

cesarbs:
Sweet.

src/Microsoft.AspNetCore.Server.Kestrel/Infrastructure/TaskUtilities.cs
```diff
@@ -29,6 +29,11 @@ public static Task GetCancelledTask(CancellationToken cancellationToken)
         public static Task<int> GetCancelledZeroTask(CancellationToken cancellationToken = default(CancellationToken))
         {
 #if NETSTANDARD1_3
+            // Make sure cancellationToken is cancelled before passing to Task.FromCanceled
+            if (!cancellationToken.IsCancellationRequested)
```
halter73: Nice catch.

src/Microsoft.AspNetCore.Server.Kestrel/Http/SocketInput.cs
```diff
@@ -286,20 +290,23 @@ public void GetResult()
...
+                // Return all blocks
+                var block = _head;
+                while (block != null)
+                {
+                    var returnBlock = block;
+                    block = block.Next;
 
-            _head = null;
-            _tail = null;
+                    returnBlock.Pool.Return(returnBlock);
```
halter73: If the blocks are currently being consumed (i.e. We are between calls to ConsumingStart and ConsumingComplete). I don't think it's safe to return them to the pool yet. I think we need to wait until the following call to ConsumingComplete.

halter73: We'll also likely have to put a lock in ConsumingStart to ensure the safety of this.

pakrym:
:up::date:

src/Microsoft.AspNetCore.Server.Kestrel/Http/SocketInput.cs
```diff
@@ -167,8 +169,10 @@ public MemoryPoolIterator ConsumingStart()
             {
                 throw new InvalidOperationException("Already consuming input.");
             }
-
-            return new MemoryPoolIterator(_head);
+            lock (_sync)
```
halter73: Update `_consumingState` inside the `_sync` lock here and in `ConsumingComplete` to avoid a race in dispose causing blocks to not get returned.

halter73: Make it a bool too.

src/Microsoft.AspNetCore.Server.Kestrel/Http/SocketInput.cs
```diff
@@ -27,9 +27,11 @@ public class SocketInput : ICriticalNotifyCompletion, IDisposable
         private MemoryPoolBlock _tail;
         private MemoryPoolBlock _pinned;
 
-        private int _consumingState;
+        private bool _consuming;
         private object _sync = new object();
 
+        private bool _disposed;
```
halter73: Nit: put _consuming and _disposed together. I think they're related.

src/Microsoft.AspNetCore.Server.Kestrel/Http/SocketInput.cs
pakrym: Done.

halter73:
:shipit:

PR #897, opened 2016-05-31
Improve response generation perf by emitting known strings directly, instead of using Buffer.BlockCopy etc

geoffkizer:
There are a lot of small, known strings we emit into response output (e.g. "HTTP/1.1", "Content-Type", etc).

Optimize these by adding or generating code that emits them directly into the buffers, when possible.

This change would supersede #853, since it greatly reduces the reliance on small copies via BlockCopy.

The win on TechEmpower is around 5%.

Note there are tests that need to be fixed before we can merge this...

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Kestrel/Http/Frame.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/Http/Frame.cs
@@ -3,10 +3,12 @@
 
 using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.IO;
 using System.Linq;
 using System.Net;
 using System.Numerics;
+using System.Runtime.CompilerServices;
 using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
@@ -593,9 +595,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Http
 
             _requestProcessingStatus = RequestProcessingStatus.ResponseStarted;
 
-            var statusBytes = ReasonPhrases.ToStatusBytes(StatusCode, ReasonPhrase);
-
-            CreateResponseHeader(statusBytes, appCompleted);
+            CreateResponseHeader(appCompleted);
         }
 
         protected Task TryProduceInvalidRequestResponse()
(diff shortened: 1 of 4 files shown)


halter73:
Very interesting. @geoffkizer You would close #853 and take this? Then we just wait for a faster Buffer.BlockCopy for the general case?

What tests need to be fixed? Looks like everything passed on AppVeyor and Travis.

src/Microsoft.AspNetCore.Server.Kestrel/Http/Frame.cs
```diff
@@ -696,8 +696,129 @@ private Task WriteSuffix()
...
         }
 
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        private unsafe void EmitHttp11(ref MemoryPoolIterator end)
+        {
+            byte* b;
+            
+            if (end.GetRawBuffer(9, out b))
+            {
+                 // Emit directly
```
halter73: Can we include the exact string we're emitting in the comment. i.e: `// Emit "HTTP/1.1 " directly`

src/Microsoft.AspNetCore.Server.Kestrel/Http/Frame.cs
```diff
@@ -761,21 +887,26 @@ private Task WriteSuffix()
...
-                responseHeaders.SetRawServer(Constants.ServerName, _bytesServer);
-            }
+            EmitHttp11(ref end);
+            WriteStatus(ref end);
+            responseHeaders.CopyTo(ref end);
 
             if (!responseHeaders.HasDate)
             {
                 var dateHeaderValues = DateHeaderValueManager.GetDateHeaderValues();
                 responseHeaders.SetRawDate(dateHeaderValues.String, dateHeaderValues.Bytes);
```
halter73: I don't think that the date needs to be added to the response headers any more.

On the same note, I don't think we need to generate the string anymore, so GetDateHeaderValues can just return a byte array instead of DateHeaderValues. This also means we don't need to make Date a "raw" header anymore.

src/Microsoft.AspNetCore.Server.Kestrel/Http/Frame.cs
```diff
@@ -696,8 +696,129 @@ private Task WriteSuffix()
             }
         }
 
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
```
mikeharder: How important is `AggressiveInlining`?  We currently don't use it anywhere in Kestrel.

src/Microsoft.AspNetCore.Server.Kestrel/Http/Frame.cs
```diff
@@ -696,8 +696,129 @@ private Task WriteSuffix()
...
+        private unsafe void EmitHttp11(ref MemoryPoolIterator end)
+        {
+            byte* b;
+            
+            if (end.GetRawBuffer(9, out b))
+            {
+                 // Emit directly
+                *(long*)b = 0x312E312F50545448;
```
mikeharder: Instead of hardcoding these numbers, could we generate them from the source string in code, and store the result in a `static readonly` field?  If the perf is the same, I would prefer this for clarity.

src/Microsoft.AspNetCore.Server.Kestrel/Http/FrameHeaders.Generated.cs
```diff
@@ -9027,7 +9043,23 @@ protected void CopyToFast(ref MemoryPoolIterator output)
...
-                                output.CopyFrom(_headerBytes, 17, 14);
+                                 
+            byte* b;
+            
+            if (output.GetRawBuffer(14, out b))
+            {
+                 // Emit directly
+*(long*)(b + 0) = 0x63656E6E6F430A0D;
+*(int*)(b + 8) = 0x6E6F6974;
+*(short*)(b + 12) = 0x203A;
```
halter73: Nit: fix indentation when generating.

benaadams:
Exotic, but I quite like it :+1:

src/Microsoft.AspNetCore.Server.Kestrel/Http/Frame.cs
benaadams: Could function local const them? e.g.
```csharp
private unsafe void EmitHttp11(ref MemoryPoolIterator end)
{
    const long bytesHttp1_1 = 0x312E312F50545448;
    const byte bytesSpace = 0x20;
    
    byte* b;
            
    if (end.GetRawBuffer(9, out b))
    {
        // Emit directly
        *(long*)b = bytesHttp1_1;
        *(b + 8) = bytesSpace;
````

File: src/Microsoft.AspNetCore.Server.Kestrel/Http/FrameHeaders.Generated.cs [hunk 6]
```diff
@@ -9010,7 +9010,23 @@ protected void CopyToFast(ref MemoryPoolIterator output)
...
                             if (value != null)
                             {
-                                output.CopyFrom(_headerBytes, 0, 17);
+                                 
+            byte* b;
+            
+            if (output.GetRawBuffer(17, out b))
+            {
+                 // Emit directly
+*(long*)(b + 0) = 0x2D65686361430A0D;
```
mikeharder: Nit: Missing leading whitespace.

src/Microsoft.AspNetCore.Server.Kestrel/Infrastructure/MemoryPoolIterator.cs
```diff
@@ -836,5 +827,44 @@ public unsafe void CopyFromAscii(string data)
             _block = block;
             _index = blockIndex;
         }
+        
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public unsafe bool GetRawBuffer(int bytesRequired, out byte* rawBuffer)
```
davidfowl: nit: TryGetRawBuffer

halter73:
I talked to @davidfowl, and we think this is too risky for 1.0.0.

@geoffkizer I'm taking #853 for now.

geoffkizer:
Sounds good.  We can revisit down the road.

Re AggressiveInlining, yes this can matter.  The JIT isn't terribly aggressive about inlining, and in a case like this where there are a bunch of small functions, forcing them to be inlined can make a real difference.  Depends on the specific scenario, though, of course.

Longer term it would be nice to have a cleaner way to accomplish this sort of thing, but that's another discussion.

benaadams:
Rebase?

muratg:
cc @davidfowl please take a look

muratg:
@halter73 will fix some issues with this.

src/Microsoft.AspNetCore.Server.Kestrel/Http/Frame.cs
JeffCyr:
```c#
*(long*)b = 0x312E312F50545448;
```
Isn't this undefined behavior with possibly misaligned write?

halter73: @JeffCyr. In what sense is this undefined? Are you worried about big endian architectures? Is the concern about misaligned writes simply a perf issue? In that case, wouldn't `Buffer.BlockCopy` also take a hit due to the misalignment?

I think there's the issue in the generating code where sometimes in the last assignment before `UpdateEnd` it's casting `b` to a `short*` when it should remain a `byte*`, causing the assignment to write one byte past the `End` index. In practice this isn't an issue since there is padding at the end of each block, but we would fix that before merging ofc.

JeffCyr: @halter73 I mean in the sense that depending on the processor architecture, the instruction could fail or result in a silent invalid memory access.

https://www.kernel.org/doc/Documentation/unaligned-memory-access.txt

> Why unaligned access is bad
> 
> The effects of performing an unaligned memory access vary from architecture
> to architecture. It would be easy to write a whole document on the differences
> here; a summary of the common scenarios is presented below:
> 
>  - Some architectures are able to perform unaligned memory accesses
>    transparently, but there is usually a significant performance cost.
>  - Some architectures raise processor exceptions when unaligned accesses
>    happen. The exception handler is able to correct the unaligned access,
>    at significant cost to performance.
>  - Some architectures raise processor exceptions when unaligned accesses
>    happen, but the exceptions do not contain enough information for the
>    unaligned access to be corrected.
>  - Some architectures are not capable of unaligned memory access, but will
>    silently perform a different memory access to the one that was requested,
>    resulting in a subtle code bug that is hard to detect!

benaadams: Mostly modern x86/x64 don't care as long as you don't cross a page boundary (when you can have a performance hit). All the memory blocks are page aligned so no writes will cross page boundaries.

x86 doesn't like writing longs though :)

halter73:
This conflicts pretty heavily with https://github.com/aspnet/KestrelHttpServer/pull/1277. Once that's merged we can consider making similar changes on top of Pipelines.

PR #901, opened 2016-06-01
Fixing "Error -4092 EACCES permission denied" in tests

moozzyk:
A test is trying to bind to port 80 which fails if IIS is running on the machine


moozzyk:
@cesarbs

cesarbs:
:ship:

halter73:
:shipit: 

We have [tests](https://github.com/aspnet/KestrelHttpServer/blob/a3d0bd0ec413d5a6563807925ba0bd11ad3eaba2/test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/AddressRegistrationTests.cs#L31) that specifically check whether we can bind on port 80 for both IPv4 and IPv6.

moozzyk:
Yes, these tests are skipped on my machine because I have IIS installed.

PR #907, opened 2016-06-02
Avoid Null ReferenceException in Frame.RequestAbortedSource

halter73:
- Avoid unneeded resetting of _abortedCts and _manuallySetRequestAbortToken


cesarbs:
:ship:

PR #908, opened 2016-06-03
Fix ThreadCountTests so they pass on OSX.

cesarbs:
@halter73 @mikeharder @BrennanConroy

### DIFF
--- a/test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/ThreadCountTests.cs
+++ b/test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/ThreadCountTests.cs
@@ -35,20 +35,22 @@ namespace Microsoft.AspNetCore.Server.Kestrel.FunctionalTests
             {
                 host.Start();
 
-                using (var client = new HttpClient())
+                // Send 20 requests just to make sure we don't get any failures
+                var requestTasks = new List<Task<string>>();
+                for (int i = 0; i < 20; i++)
                 {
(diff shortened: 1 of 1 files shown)


pakrym:
Does this mean that `HttpClient` leaks handles?

cesarbs:
@pakrym not sure.

@stephentoub any insight on what could be happening here? Before the change we were getting curl exceptions saying it got an empty response from the server.

stephentoub:
> Before the change we were getting curl exceptions saying it got an empty response from the server.

Are you sure its not an issue with the server? From a quick glance at the change, the biggest difference is the former is using a single HttpClient whereas the latter is using one per request, which means the former is using connection pooling whereas the latter isn't.

cesarbs:
Debugging this and got this:

    Microsoft.AspNetCore.Server.Kestrel.Networking.UvException: Error -40 EMSGSIZE message too long
    fail: Microsoft.AspNetCore.Server.Kestrel[0]
          DispatchPipe.ReadStart

Seems like a problem in Kestrel indeed.

mikeharder:
So this isn't related to the open file limit at all?

cesarbs:
@mikeharder I don't want to say it's 100% not at this point, but I don't think so. I was watching the number of file descriptors used by the process and it never went above the limit.

halter73:
@cesarbs Do you think we should close this?

cesarbs:
@halter73 Yeah.

PR #911, opened 2016-06-03
Fixing a test failing on Mac and Linux

moozzyk:
Replacing non-ASCII character with equivalent Unicode character escape sequences


anurse:
:shipit: when builds are green (and ask mode approval)

moozzyk:
Mac is red due to a failure @cesarbs is looking at. I coded the fix on Mac and tested it there. It's been approved by @Eilon

PR #912, opened 2016-06-04
Limit size of memory buffer when reading request (#304)

mikeharder:
Creating PR to start getting feedback and run tests on Travis/AppVeyor.

Addresses #304.

## Tasks
- [x] Implement changes in `Connection`, `SocketInput`, and `KestrelServerOptions`
- [x] Unit tests
- [x] Functional tests for HTTP
- [x] Fix bug where `Connection.OnRead()` and `IConnectionControl.Pause()` can both call `_socket.ReadStop()` and throw.
 - Occasionally repros during functional test which sets max buffer to the exact same size as the uploaded data.  So Connection can see the buffer become full and EOF around the same time.
 - Made `UvStreamHandle.ReadStop()` idempotent
- [x] Fix bug where `UvStreamHandle.ReadStart()` can throw a UvException in some cases (e.g. socket is no longer connected).
- [x] Change functional test from `Socket` to `NetworkStream`, so the same test can be reused with `SslStream`.
- [x] Test HTTPS
- [x] Functional tests sometimes fail with ` Block being garbage collected instead of returned to pool` assertion.
 - Happens very occasionally on Windows, but almost always on Mac and Linux
 - Only happens when testing SSL with Content-Length.  Non-SSL and non-Content-Length doesn't repro.
 - Owner: @cesarbs 
- [x] Functional tests fail on Mac and Linux with unexpected client behavior, writing either too many or too few bytes before being paused
 - [x] Too few bytes appears to be caused by a deadlock in xunit when VM only has a single core.
 - [x] Too many bytes caused by larger OS buffer on Linux (10MB) compared to Windows (700kb). 
 - [x] Still getting too few bytes on OSX Travis, may need to increase timeout.
- [x] Test performance of TechEmpower scenarios
- [ ] Test performance of large uploads.  Worst-case is when the server is reading slightly slower than the client is sending, which will cause frequent calls to `Pause()/Resume()`.
- [x] Test real-world application (e.g. MVC) that writes uploaded file to a (simulated) slow disk.
 - Owner: @DavidObando 
- [x] Change type from `int?` to `long?` to support buffers larger than `Int32.MaxValue`.  Future-proof without breaking change.
- [x] Add sentence to doc comment explaining that `null` disables the buffer length checking.
- [x] Decide final name of option: `MaxRequestBufferSize`
- [x] Decide default value for `MaxRequestBufferSize`.  Reasonable options:
 - `null`: No limit.  Least likely to cause issues since execution flow is nearly identical to before this change.  However, few customers will change the default, so we are least likely to find issues since the code is not being executed very often, and customers may be surprised if they increase request limit in nginx/IIS but don't change `MaxInputBufferLength` and see Kestrel using a lot of memory.
 - `1 MB`: Matches default limit in nginx.  Most requests will be under the limit, unless they are large uploads.
- [x] Decide threshold to call resume.  Reasonable options:
 - `MaxInputBufferLength - 1`: If the client is sending faster than the server can consume, Pause()/Resume() may be called as often as every packet, which might degrade perf.  However, it might also be the most efficient way to do things, since the buffers are kept full and the client is never blocked very long.  Another advantage is the client is less likely to get a false-positive timeout.
 - `MaxInputBufferLength * 0.5`: Blocks the client until the server has drained at least half the buffer.  The client will be paused less often, but each pause will be for a longer time.
- [ ] Add timeouts to async calls in functional test, so test fails fast if an async call fails to return
- [ ] Consider adding unit tests for new code in `Connection`.

## Performance
No observable change in Plaintext or Json RPS.

## Notes
- Default request body is 1MB in nginx and 28.6MB in IIS7
 - http://nginx.org/en/docs/http/ngx_http_core_module.html#client_max_body_size
 - https://www.iis.net/configreference/system.webserver/security/requestfiltering/requestlimits


src/Microsoft.AspNetCore.Server.Kestrel/KestrelServerOptions.cs
```diff
@@ -8,16 +8,33 @@ namespace Microsoft.AspNetCore.Server.Kestrel
...
+        private int _maxInputBufferLength = 1024 * 1024;
+        public int MaxInputBufferLength
+        {
+            get
+            {
+                return _maxInputBufferLength;
+            }
+            set
+            {
+                if (value < -1 || value == 0)
```
Tratcher: Please don't use magic values. Use `int?` instead.

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/MaxInputBufferLengthTests.cs
```diff
@@ -0,0 +1,173 @@
...
+        [InlineData(1024 * 1024, true, true)]
+        [InlineData(1024 * 1024, false, true)]
+        [InlineData(5 * 1024 * 1024, true, true)]
+        [InlineData(5 * 1024 * 1024, false, true)]
+        [InlineData(10 * 1024 * 1024, true, false)]
+        [InlineData(10 * 1024 * 1024, false, false)]
+        [InlineData(Int32.MaxValue, true, false)]
+        [InlineData(Int32.MaxValue, false, false)]
+        [InlineData(-1, true, false)]
+        [InlineData(-1, false, false)]
```
halter73: Do you have any tests for `MaxInputBufferLength = 0`?

mikeharder: It's not allowed.  Valid values are currently `>1` or `-1` (which will soon be replaced by `null`).  I will add a test for a tiny buffer like `1`, which should pause the connection after each packet.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/SocketInput.cs
```diff
@@ -321,5 +341,73 @@ private static void ReturnBlocks(MemoryPoolBlock block, MemoryPoolBlock end)
...
+                    {
+                        _connectionPaused = false;
+                        _connectionThread.Post(
+                            (connectionControl) => ((IConnectionControl)connectionControl).Resume(),
+                            _connectionControl);
+                    }
+                }
+            }
+        }
+
```
halter73: Nit: remove newline.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/SocketInput.cs
```diff
@@ -189,10 +203,16 @@ public MemoryPoolIterator ConsumingStart()
...
                     {
+                        var lengthConsumed = new MemoryPoolIterator(_head).GetLength(consumed);
+
                         returnStart = _head;
                         returnEnd = consumed.Block;
                         _head = consumed.Block;
                         _head.Start = consumed.Index;
+
+                        // Must call Subtract() after bytes have been freed, to avoid producer starting too early and growing
+                        // buffer beyond max length.
```
halter73: This comment implies that the bytes should already be returned to the pool which they're not. Maybe "after _head as been advanced" would be clearer? Even that I don't think is too important as long as `Subtract` is called with the `_sync` lock.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/SocketInput.cs
```diff
@@ -321,5 +341,73 @@ private static void ReturnBlocks(MemoryPoolBlock block, MemoryPoolBlock end)
...
+            private bool _connectionPaused;
+
+            public BufferLengthConnectionController(int maxLength, IConnectionControl connectionControl, KestrelThread connectionThread)
+            {
+                _maxLength = maxLength;
+                _connectionControl = connectionControl;
+                _connectionThread = connectionThread;
+            }
+
+            public int Length
```
halter73: Can this be private?

src/Microsoft.AspNetCore.Server.Kestrel/KestrelServerOptions.cs
```diff
@@ -8,16 +8,33 @@ namespace Microsoft.AspNetCore.Server.Kestrel
...
+        /// </summary>
+        public bool AddServerHeader { get; set; } = true;
+
         public IServiceProvider ApplicationServices { get; set; }
 
         public IConnectionFilter ConnectionFilter { get; set; }
 
-        public bool NoDelay { get; set; } = true;
+        private int _maxInputBufferLength = 1024 * 1024;
+        public int MaxInputBufferLength
```
halter73: Not that we are consistent with this, but we should add a doc comment. Probably something along the lines of `Per-request buffer limit measured in bytes.`

halter73:
Can you include the before and after plaintext benchmark results the description?

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/MaxInputBufferLengthTests.cs
```diff
@@ -0,0 +1,173 @@
...
+                        // 
+                        // The minimum is (maxInputBufferLength - maxSendSize + 1), since if bytesWritten is
+                        // (maxInputBufferLength - maxSendSize) or smaller, the client should be able to
+                        // complete another send.
+                        // 
+                        // The maximum is harder to determine, since there can be OS-level buffers in both the client
+                        // and server, which allow the client to send more than maxInputBufferLength before getting
+                        // paused.  We assume the combined buffers are smaller than the difference between
+                        // data.Length and maxInputBufferLength.                          
+                        Assert.InRange(bytesWritten, maxInputBufferLength.Value - maxSendSize + 1, data.Length - 1);
```
halter73: This assertion is failing on AppVeyor and Travis. It seems that `bytesWritten` is usually (but not always) 0. E.g. [On AppVeyor:](https://ci.appveyor.com/project/aspnet-ci/kestrelhttpserver/build/1.0.1754#L530)

    Microsoft.AspNetCore.Server.Kestrel.FunctionalTests.MaxInputBufferLengthTests.LargeUpload(maxInputBufferLength: 16384, sendContentLengthHeader: True, expectPause: True) [FAIL] 
      Assert.InRange() Failure 
      Range:  (12289 - 10485759) 
      Actual: 0 
      Stack Trace: 
        MaxInputBufferLengthTests.cs(79,0): at Microsoft.AspNetCore.Server.Kestrel.FunctionalTests.MaxInputBufferLengthTests.<LargeUpload>d__0.MoveNext() 
           at System.Runtime.CompilerSer

mikeharder: Will try a larger timeout, but 100ms is reliable on my dev machine.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/SocketInput.cs
mikeharder: Would it be more correct to call `Subtract()` after `ReturnBlocks()`?  My goal was to call `Subtract()` at the proper time, so the buffer count matches the actual consumed bytes as closely as possible.  If `Subtract()` is called before bytes have been returned to the pool, it's possible for the producer to start again and grow the buffer larger than it should be allowed.  Unlikely to matter much in practice, but I wanted to be as close as possible to correct.

mikeharder: Specifically, would change code to:

```
_bufferLengthControl?.Subtract(
    new MemoryPoolIterator(returnStart).GetLength(new MemoryPoolIterator(returnEnd)));
ReturnBlocks(returnStart, returnEnd);
```

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/SocketInput.cs
```diff
@@ -321,5 +341,73 @@ private static void ReturnBlocks(MemoryPoolBlock block, MemoryPoolBlock end)
                 returnBlock.Pool.Return(returnBlock);
             }
         }
+
+        private class BufferLengthConnectionController
```
davidfowl: I still don't think this component should pause and resume. Leave that up to the caller. This couples too many pieces together (especially knowing the changes that are coming). If this is the easiest thing to do for 1.0 then it's fine but this really shouldn't know that much about pausing and resuming the connection and the fact that it has to be posted to the libuv thread.

mikeharder: This design was also problematic when using a connection filter, since this involves two instances of `SocketInput`.  Refactoring will be pushed shortly.

davidfowl: Ok, we'll have to redo this when the refactoring happens.

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/MaxInputBufferLengthTests.cs
```diff
@@ -0,0 +1,233 @@
...
+
+namespace Microsoft.AspNetCore.Server.Kestrel.FunctionalTests
+{
+    public class MaxInputBufferLengthTests
+    {
+        private const int _dataLength = 10 * 1024 * 1024;
+
+        public static IEnumerable<object[]> LargeUploadData
+        {
+            get
```
cesarbs: Please include some rationale for the test data you're creating here i.e. why those values and what you're trying to achieve with each one (or each group).

I'm fine with having the test like this (since we already follow this pattern in other places), but I'm not a fan of this approach. I prefer separate, descriptive tests (e.g. `UploadExpectsPause`, `BufferLengthControlWorksWithSsl`, etc.) that stand as a spec of what the code is supposed to do.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/SocketInput.cs
halter73: I don't think it matters if you call subtract before or after returning the blocks. I only commented because your comment makes it seem like it does. I think this is just me misconstruing what we mean by the bytes being "freed".

halter73: So I think it's fine leaving the code as it is. I was just trying to suggest a clearer comment. I think the comment you have now is fine.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Connection.cs
```diff
@@ -316,7 +324,16 @@ void IConnectionControl.Pause()
...
             Log.ConnectionResume(ConnectionId);
-            _socket.ReadStart(_allocCallback, _readCallback, this);
+            try
+            {
+                _socket.ReadStart(_allocCallback, _readCallback, this);
+            }
+            catch (UvException ex)
+            {
+                // ReadStart() can throw a UvException in some cases (e.g. socket is no longer connected).
+                // This should be treated the same as OnRead() seeing a "normalDone" condition.
```
halter73: This is actually the same as OnRead() seeing a *"errorDone"* condition since you're passing the exception. I'm not sure how safe this is since we normally treat a connection reset as "normalDone". It might be better to just log the exception and not pass it to IncomingComplete().

test/Microsoft.AspNetCore.Server.KestrelTests/project.json
```diff
@@ -21,7 +21,8 @@
         "System.Net.Http": "4.1.0-*",
         "System.Net.Http.WinHttpHandler": "4.0.0-*",
         "System.Net.Sockets": "4.1.0-*",
-        "System.Runtime.Handles": "4.0.1-*"
+        "System.Runtime.Handles": "4.0.1-*",
+        "moq.netcore": "4.4.0-beta8"
```
halter73: Ain't this convenient :smile:

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Connection.cs
```diff
@@ -384,5 +401,69 @@ private enum ConnectionState
             Disconnecting,
             SocketClosed
         }
+
+        private class BufferLengthControl : IBufferLengthControl
```
halter73: Since this is effectively used by multiple classes (even if it is via an interface), I think I like this in it's own file. I think already Kestrel uses nested classes too heavily.

src/Microsoft.AspNetCore.Server.Kestrel/KestrelServerOptions.cs
```diff
@@ -8,16 +8,33 @@ namespace Microsoft.AspNetCore.Server.Kestrel
...
+        /// </summary>
+        public bool AddServerHeader { get; set; } = true;
+
         public IServiceProvider ApplicationServices { get; set; }
 
         public IConnectionFilter ConnectionFilter { get; set; }
 
-        public bool NoDelay { get; set; } = true;
+        private int? _maxInputBufferLength = 1024 * 1024;
+        public int? MaxInputBufferLength
```
halter73: Add doc comment. I think we might try to add this for all the KestrelServerOptions properties soon. https://github.com/aspnet/KestrelHttpServer/issues/922

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Connection.cs
```diff
@@ -316,7 +324,16 @@ void IConnectionControl.Pause()
         void IConnectionControl.Resume()
         {
             Log.ConnectionResume(ConnectionId);
-            _socket.ReadStart(_allocCallback, _readCallback, this);
+            try
+            {
+                _socket.ReadStart(_allocCallback, _readCallback, this);
```
halter73: You might be able to avoid an exception in some cases by checking that `_rawSocketInput.RemoteIntakeFin == false` before making the call.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Connection.cs
mikeharder: OK, I will stop passing the exception, but I don't think logging the exception is correct either, since this is a perfectly normal thing to happen, and logging an exception will confuse customers.

`Connection.OnRead()` just calls `Log.ConnectionReadFin()` for `normalDone`.  Should I do the same in Resume()?

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Connection.cs
mikeharder: I would rather not add extra code just to avoid an exception.  It would be like checking `File.Exists()` before you call `File.Open()`, which usually isn't worth it.

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/MaxInputBufferLengthTests.cs
mikeharder: I will add a comment for each buffer size explaining why it's worth testing.

I do think the buffer sizes (and whether we expect the client to pause) should be passed as theory data.  Creating a new test for each buffer size would be a lot of extra work.

For the 4 combinations of `sendContentLengthHeader` and `ssl`, it would be reasonable to create 4 tests like:

1. `LargeUploadWithContentLength`
2. `LargeUploadWithoutContentLength`
3. `LargeUploadWithContentLengthOverSsl`
4. `LargeUploadWithoutContentLengthOverSsl`

But I still think passing these as boolean parameters is cleaner.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/SocketInput.cs
mikeharder: I agree the current code seems to work fine, and in practice there will be a very short time window between the two places to call Subtract().  But which do you think is the most correct, given that the actual bytes used by the connection should not exceed the limit?  I think it's waiting until after ReturnBlocks(), no?

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Connection.cs
halter73: I think I disagree with you about checking `File.Exists`, at least when there's a good chance that the file in fact doesn't exist. This is all happening on the libuv thread so there is no race going on like there might be with files. The only time ReadStart may throw as far as I'm aware is in the somewhat rare case that we called ReadStop after receiving the last bit of data but before the FIN/reset.

It might be worth asking the libuv mailing list to see if there is a way to listen for a connection closed event without relying on the read callback (or somehow still applying back pressure).

If there is no other way, we could arguably have ReadStart return the status. We would likely have to check it and throw it it fails during the first call in Connection.Start, but here we could just check it in an if instead of using try/catch for control flow.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Connection.cs
halter73: `Log.ConnectionReadFin()` is definitely a good call.

We don't log for ECONNRESET. Presumably we will log/throw given an incomplete request somewhere else (At least most of the time. `Connection: close` makes this tricky, but that's already an issue with ECONNRESET received via the read callback.)

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/SocketInput.cs
```diff
@@ -189,10 +197,16 @@ public MemoryPoolIterator ConsumingStart()
                 {
                     if (!consumed.IsDefault)
                     {
+                        var lengthConsumed = new MemoryPoolIterator(_head).GetLength(consumed);
```
cesarbs: Only compute this if `_bufferLengthControl` is not `null`.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Connection.cs
mikeharder: Even though libuv uses a single thread, I assume ReadStart() calls into some OS function which itself can fail at any time.  So we need to either try/catch or check a status code.  Status code might have better perf but .NET convention for I/O is exceptions, and this particular exception will not be thrown often enough where it should impact real-world perf.

I think the current code is fine, and adding more code is just increasing complexity with no real benefit.

src/Microsoft.AspNetCore.Server.Kestrel/KestrelServerOptions.cs
```diff
@@ -8,16 +8,37 @@ namespace Microsoft.AspNetCore.Server.Kestrel
...
+
         public IServiceProvider ApplicationServices { get; set; }
 
         public IConnectionFilter ConnectionFilter { get; set; }
 
-        public bool NoDelay { get; set; } = true;
-
         /// <summary>
-        /// Gets or sets whether the <c>Server</c> header should be included in each response.
+        /// Maximum number of bytes used to buffer input for each connection.
```
halter73: I like when doc comments include the default value.

src/Microsoft.AspNetCore.Server.Kestrel/KestrelServerOptions.cs
```diff
@@ -8,16 +8,37 @@ namespace Microsoft.AspNetCore.Server.Kestrel
 {
     public class KestrelServerOptions
     {
+        private int? _maxInputBufferLength = 1024 * 1024;
```
halter73: Did you base this on a limit on the limit of another server? It might be nice to add a non-doc comment explaining how we chose the default.

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/MaxInputBufferLengthTests.cs
```diff
@@ -0,0 +1,248 @@
...
+            socket.Connect(IPAddress.Loopback, port);
+
+            return socket;
+        }
+
+        private static void WritePostRequestHeaders(Stream stream, int? contentLength)
+        {
+            using (var writer = new StreamWriter(stream, Encoding.ASCII, bufferSize: 1024, leaveOpen: true))
+            {
+                writer.WriteLine("POST / HTTP/1.0");
```
cesarbs: The reason this is failing on non-Windows is that WriteLine is writing `\n` at the end instead of `\r\n`. Change these calls to `Write` and include `\r\n` in the strings.

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/MaxInputBufferLengthTests.cs
mikeharder: Currently, the only issue we're seeing on AppVeyor and Travis is the `Block being garbage collected instead of returned to pool` assertion, which @cesarbs is investigating.  We'll need to either fix the root cause or disable this assertion for 1.0.0, to unblock the tests for my change.

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/MaxInputBufferLengthTests.cs
```diff
@@ -89,24 +89,26 @@ public static IEnumerable<object[]> LargeUploadData
                 using (var socket = CreateSocket(port))
                 using (var stream = await CreateStreamAsync(socket, ssl, host.GetHost()))
                 {
-                    WritePostRequestHeaders(stream, sendContentLengthHeader ? (int?)data.Length : null);
+                    await WritePostRequestHeaders(stream, sendContentLengthHeader ? (int?)data.Length : null);
 
                     var bytesWritten = 0;
 
-                    var sendTask = Task.Run(() =>
+                    Func<Task> sendFunc = async () =>
```
cesarbs: I like this better the way it was before.

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/MaxInputBufferLengthTests.cs
```diff
@@ -0,0 +1,250 @@
...
+            {
+                var port = host.GetPort(ssl ? "https" : "http");
+                using (var socket = CreateSocket(port))
+                using (var stream = await CreateStreamAsync(socket, ssl, host.GetHost()))
+                {
+                    await WritePostRequestHeaders(stream, sendContentLengthHeader ? (int?)data.Length : null);
+
+                    var bytesWritten = 0;
+
+                    Func<Task> sendFunc = async () =>
```
cesarbs: I liked this better the way it was before.

mikeharder: Which part?  I wanted to switch from sync to async writes, in part to improve test performance, and also because async IO is generally better and what we recommend customers use.  This requires creating an async delegate.

This can then be invoked either directly or by calling `Task.Run(Func<Task>)`, but I don't see any benefit to `Task.Run()`.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Connection.cs
```diff
@@ -49,7 +52,12 @@ public Connection(ListenerContext context, UvStreamHandle socket) : base(context
 
             ConnectionId = GenerateConnectionId(Interlocked.Increment(ref _lastConnectionId));
 
-            _rawSocketInput = new SocketInput(Memory, ThreadPool);
+            if (ServerOptions.MaxInputBufferLength.HasValue)
```
SajayAntony: This  default  will result in large uploads with slow consumers to crash the process. Is that ok?

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/project.json
```diff
@@ -8,7 +8,8 @@
     "Microsoft.AspNetCore.Server.Kestrel.Https": "1.0.0-*",
     "Microsoft.AspNetCore.Testing": "1.0.0-*",
     "Newtonsoft.Json": "9.0.1-beta1",
-    "xunit": "2.1.0"
+    "xunit": "2.1.0",
+    "Microsoft.Extensions.Logging.Console": "1.0.0-rc3-21270"
```
SajayAntony: Do you just want *?

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Connection.cs
mikeharder: Can you be more specific?  I don't understand what you mean.

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/project.json
mikeharder: Thanks, fixed.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Connection.cs
SajayAntony: I'm not sure how the default behaves  `private int? _maxInputBufferLength = 1024 * 1024; `
Does this mean 1MB is the default and in that case will HasValue every be false?

Basically is there a scenario where we never use the BufferLenghtControl which would cause large post to kill the server.

Tratcher: The user could turn it off by setting it to null

halter73: @SajayAntony Is the server process crashing?

mikeharder: The default is 1MB, user can set any value between 1 byte and Int32.MaxValue, plus `null` to completely disable all code related to this change, for both perf and also in case this code has a bug.

benaadams: Do you mean slow loris'esk body non-timeout? Was suggested in https://github.com/aspnet/KestrelHttpServer/pull/485#discussion_r55264843 that that was out of the preview of the server

SajayAntony: @halter73 nope. the run is actually looking good. 
@Tratcher thanks, that's what I was missing. The API seems bit clunky (setting something to null to disable 😨 )

cesarbs:
LGTM. Squash then 🚢

PR #917, opened 2016-06-06
Make Bad Request tests more reliable

halter73:
- Avoid calling write again after the request is already rejected

This avoids sporadic failures like the following:

    Microsoft.AspNetCore.Server.KestrelTests.BadHttpRequestTests.BadRequestIfPathContainsNullCharacters(path: "/%F3%85%82%00") [FAIL]
      System.IO.IOException : Unable to write data to the transport connection: An established connection was aborted by the software in your host machine.
      ---- System.Net.Sockets.SocketException : An established connection was aborted by the software in your host machine
      Stack Trace:
           at System.Net.Sockets.NetworkStream.BeginWrite(Byte[] buffer, Int32 offset, Int32 size, AsyncCallback callback, Object state)
           at System.Net.Sockets.NetworkStream.<>c.<WriteAsync>b__59_0(Byte[] bufferArg, Int32 offsetArg, Int32 sizeArg, AsyncCallback callback, Object state)
           at System.Threading.Tasks.TaskFactory`1.FromAsyncImpl[TArg1,TArg2,TArg3](Func`6 beginMethod, Func`2 endFunction, Action`1 endAction, TArg1 arg1, TArg2 arg2, TArg3 arg3, Object state, TaskCreationOptions creationOptions)
           at System.Threading.Tasks.TaskFactory.FromAsync[TArg1,TArg2,TArg3](Func`6 beginMethod, Action`1 endMethod, TArg1 arg1, TArg2 arg2, TArg3 arg3, Object state)
           at System.Net.Sockets.NetworkStream.WriteAsync(Byte[] buffer, Int32 offset, Int32 size, CancellationToken cancellationToken)
           at System.IO.StreamWriter.<FlushAsyncInternal>d__67.MoveNext()
        --- End of stack trace from previous location where exception was thrown ---
           at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
           at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
           at System.Runtime.CompilerServices.TaskAwaiter.GetResult()
        C:\Users\shalter\dev\Universe\KestrelHttpServer\test\Microsoft.AspNetCore.Server.KestrelTests\TestConnection.cs(69,0): at Microsoft.AspNetCore.Server.KestrelTests.TestConnection.<Send>d__11.MoveNext()
        --- End of stack trace from previous location where exception was thrown ---
[...]

PR #918, opened 2016-06-06
Fixing BadRequest tests to not try sending data after Bad Request

moozzyk:


cesarbs:
:ship: when Travis goes green.

halter73:
I think #917 is the more comprehensive/reliable fix.

moozzyk:
Not sure if I like swallowing the exception - can it hide random failures?

halter73:
This conversation might be better in #917, but I think my change is perfectly safe.

It's only catching an exception from a call to `Shutdown`. All tests that use `SendAllTryEnd` expect the server to respond with a 400 and reset the connection so it's not unexpected that the `Shutdown` will fail.

Initially I tried changing all the tests to just call `SendAll` instead of `SendEnd`. This would have been fine except for some of the `Theories` expect the request to be rejected because they're incomplete.

I could refactor the Theories and split up the test cases into ones that should call `SendAll` and those that should call `SendAllEnd`. Maybe this is better, but `Shutdown` failing isn't an issue, since we are always verifying we get the response that we expect.

moozzyk:
We agreed that it would be good to have both fixes.

PR #925, opened 2016-06-13
Added "308 Permanent Redirect" to ReasonPhrases

khellang:
From RFC7538 - https://tools.ietf.org/html/rfc7538


khellang:
Also, are you interested in a PR for "451 Unavailable For Legal Reasons" ([RFC7725](https://tools.ietf.org/html/rfc7725))? I see it's missing from the HTTP abstractions as well...

cesarbs:
:ship:

cesarbs:
@muratg @Eilon @davidfowl What's your take on 451?

halter73:
:shipit: 308 seems like a pretty safe. I would take 451 too because why not?

khellang:
Submitted 451 as well... #930

muratg:
Seems safe and nice to me too. Though bar for 1.0.0 is very high now, so we'll get this in once we branch for RTM.

cesarbs:
Let's merge this. :ship:

PR #927, opened 2016-06-13
Combine LargeMultipartUpload and LargeUpload into one test.

cesarbs:
Resolves the CI issues we've been seeing.

I've had to remove the `Assert`s in the test app, otherwise the test was too slow. I'll think of something new and faster to assert.

I've ensured this tests what we want to test by reverting https://github.com/aspnet/KestrelHttpServer/commit/41f77ee3fd14df9d666efd3662803ae4294b4947 and checking that the test fails without those changes.

cc @pranavkm @mikeharder @halter73

### DIFF
--- a/test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/RequestTests.cs
+++ b/test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/RequestTests.cs
@@ -1,9 +1,11 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
-using System;
 using System.Globalization;
+using System.Net;
 using System.Net.Http;
+using System.Net.Sockets;
+using System.Text;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Builder;
 using Microsoft.AspNetCore.Hosting;
@@ -17,9 +19,19 @@ namespace Microsoft.AspNetCore.Server.Kestrel.FunctionalTests
 {
     public class RequestTests
     {
-        [Fact]
-        public async Task LargeUpload()
+        [Theory]
+        [InlineData(10 * 1024 * 1024, true)]
+        // In the following dataset, send at least 2GB.
+        // Never change to a lower value, otherwise regression testing for
+        // https://github.com/aspnet/KestrelHttpServer/issues/520#issuecomment-188591242
+        // will be lost.
+        [InlineData((long)int.MaxValue + 1, false)]
+        public void LargeUpload(long contentLength, bool checkBytes)
         {
+            const int bufferLength = 1024 * 1024;
+            Assert.True(contentLength % bufferLength == 0, $"{nameof(contentLength)} sent must be evenly divisible by {bufferLength}.");
+            Assert.True(bufferLength % 256 == 0, $"{nameof(bufferLength)} must be evenly divisible by 256");
+
             var builder = new WebHostBuilder()
                 .UseKestrel()
                 .UseUrls("http://127.0.0.1:0/")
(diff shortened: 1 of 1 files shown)


test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/RequestTests.cs
```diff
@@ -49,68 +47,32 @@ public class RequestTests
             {
                 host.Start();
 
-                using (var client = new HttpClient())
+                using (var socket = new Socket(AddressFamily.InterNetwork, SocketType.Stream, ProtocolType.Tcp))
                 {
-                    var bytes = new byte[1024 * 1024];
-                    for (int i = 0; i < bytes.Length; i++)
-                    {
-                        bytes[i] = (byte)i;
```
Tratcher: Hmm, this no longer checks for data corruption in the body... that's unfortunate.

mikeharder: I would test uploads of both `10MB` and `> 2GB`, and only validate the bytes in the `10MB` case.  A parameterized test would be good for this.

cesarbs: @Tratcher Yeah, I won't leave it like that.

@mikeharder I like that idea.

cesarbs:
Updated.

mikeharder:
Test might be cleaner if you used `TcpClient` plus `StreamReader/StreamWriter`, but :shipit: if you want to keep using raw `Socket`.

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/RequestTests.cs
```diff
@@ -28,17 +40,20 @@ public class RequestTests
...
+                        var receivedBytes = new byte[bufferLength];
+                        var received = 0;
+                        while ((received = await context.Request.Body.ReadAsync(receivedBytes, 0, receivedBytes.Length)) > 0)
                         {
-                            for (int i = 0; i < count; i++)
+                            if (checkBytes)
                             {
-                                Assert.Equal(total % 256, bytes[i]);
-                                total++;
+                                for (int i = 0; i < received; i++)
```
pranavkm: Could you generate the expected sequence up front and do an `Assert.Equal` instead? 

```
var contentBytes = Enumerable.Range(0, bufferLength).ToArray();
...

Assert.Equal(contentBytes, receivedBytes);
```

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/RequestTests.cs
```diff
@@ -49,68 +64,35 @@ public class RequestTests
...
-
-            using (var host = builder.Build())
-            {
-                host.Start();
+                    }
 
-                using (var client = new HttpClient())
-                {
-                    using (var form = new MultipartFormDataContent())
+                    for (int i = 0; i < contentLength / contentBytes.Length; i++)
```
pranavkm: `var` all day every day

pranavkm:
:shipit:

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/RequestTests.cs
mikeharder: If you make this change, be sure to check perf.  I validate a 20MB upload in https://github.com/aspnet/KestrelHttpServer/pull/912, and I found large (10x) performance differences between ways to do the check.

mikeharder: I found a simple `for` loop to perform quite well.

cesarbs: I'll leave it as is.

Issue #934, opened 2016-06-20
Considering treating ECONNRESET as an error case


halter73:
Kestrel has treated an `ECONNRESET` read the same as `EOF` [pretty much since the beginning](https://github.com/aspnet/KestrelHttpServer/commit/21778f631a4d8a976c19f32655d2c9f6dd1d7074), and I don't think this decision ha really been questioned before. Like how we were improperly handling 0 status back then, I'm starting to think that treating `ECONNRESET` different from other read errors might also be a mistake.

First, the fact that we don't even log `ECONNRESET` differently than `EOF`, makes it difficult to diagnose issues like #932. I believe that issue stems from a connection reset, but I can't really be certain. I don't think the problem is just logging though.

Right now, if `ECONNRESET` happens mid request, this will often lead to a `BadHttpRequestException` due to invalid headers, content length, chunking, etc... If the request becomes invalid, the issue is somewhat mitigated, since it at least it leads to some sort of read error. It would still be nicer if a `ECONNRESET UvException/IOException` was surfaced instead when that is the root cause of the bad request.

`HTTP/1.0` and `Connection: close` requests are scarier. Instead of there being an error, the request is simply truncated. This could cause apps to misbehave in subtle ways.

It shouldn't be a concern that clients using `ECONNRESET` as a "normal" means of closing the request, since this would prevent the client from ever receiving the response. If an `ECONNRESET` is received after the request is completed but before the response is written, the application won't see any exceptions unless it passes the `RequestAborted` token to calls to Write (#566).

If we make this change, we can also remove the platform specific logic to determine the error code. (#764).

halter73:
@cesarbs @mikeharder

PR #953, opened 2016-07-05
Treat ECONNRESET as a connection error (#934).

cesarbs:
#934, #959

cc @halter73 @mikeharder

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/SocketInput.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/SocketInput.cs
@@ -22,7 +22,6 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Http
         private readonly ManualResetEventSlim _manualResetEvent = new ManualResetEventSlim(false, 0);
 
         private Action _awaitableState;
-        private Exception _awaitableError;
 
         private MemoryPoolBlock _head;
         private MemoryPoolBlock _tail;
@@ -33,6 +32,8 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Http
         private bool _consuming;
         private bool _disposed;
 
+        private TaskCompletionSource<object> _tcs = new TaskCompletionSource<object>();
+
         public SocketInput(MemoryPool memory, IThreadPool threadPool, IBufferSizeControl bufferSizeControl = null)
         {
             _memory = memory;
@@ -41,9 +42,15 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Http
             _awaitableState = _awaitableIsNotCompleted;
         }
 
-        public bool RemoteIntakeFin { get; set; }
-
         public bool IsCompleted => ReferenceEquals(_awaitableState, _awaitableIsCompleted);
+
+        private bool ReadingInput => _tcs.Task.Status == TaskStatus.WaitingForActivation;
+
+        public bool CheckFinOrThrow()
+        {
+            CheckConnectionError();
+            return _tcs.Task.Status == TaskStatus.RanToCompletion;
+        }
 
         public MemoryPoolBlock IncomingStart()
         {
@@ -87,7 +94,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Http
                 }
                 else
                 {
-                    RemoteIntakeFin = true;
+                    FinReceived();
                 }
 
                 Complete();
@@ -122,13 +129,13 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Http
                     _pinned = null;
                 }
 
-                if (count == 0)
-                {
-                    RemoteIntakeFin = true;
-                }
                 if (error != null)
                 {
-                    _awaitableError = error;
+                    SetConnectionError(error);
+                }
+                else if (count == 0)
+                {
+                    FinReceived();
                 }
 
                 Complete();
@@ -216,8 +223,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Http
 
                     if (!examined.IsDefault &&
                         examined.IsEnd &&
-                        RemoteIntakeFin == false &&
-                        _awaitableError == null)
+                        ReadingInput)
                     {
                         _manualResetEvent.Reset();
 
@@ -252,8 +258,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Http
 
         public void AbortAwaiting()
         {
-            _awaitableError = new TaskCanceledException("The request was aborted");
-
+            SetConnectionError(new TaskCanceledException("The request was aborted"));
             Complete();
         }
 
@@ -279,7 +284,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Http
             }
             else
             {
-                _awaitableError = new InvalidOperationException("Concurrent reads are not supported.");
+                SetConnectionError(new InvalidOperationException("Concurrent reads are not supported."));
 
                 Interlocked.Exchange(
                     ref _awaitableState,
@@ -303,15 +308,8 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Http
             {
                 _manualResetEvent.Wait();
             }
-            var error = _awaitableError;
-            if (error != null)
-            {
-                if (error is TaskCanceledException || error is InvalidOperationException)
-                {
-                    throw error;
-                }
-                throw new IOException(error.Message, error);
-            }
+
+            CheckConnectionError();
         }
 
         public void Dispose()
@@ -338,6 +336,29 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Http
                 block = block.Next;
 
                 returnBlock.Pool.Return(returnBlock);
+            }
+        }
+
+        private void SetConnectionError(Exception error)
+        {
+            _tcs.TrySetException(error);
+        }
+
+        private void FinReceived()
+        {
+            _tcs.TrySetResult(null);
+        }
+
+        private void CheckConnectionError()
+        {
+            var error = _tcs.Task.Exception?.InnerException;
+            if (error != null)
+            {
+                if (error is TaskCanceledException || error is InvalidOperationException)
+                {
+                    throw error;
+                }
+                throw new IOException(error.Message, error);
             }
         }
     }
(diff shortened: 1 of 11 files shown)


cesarbs:
Rebased.

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/RequestTests.cs
```diff
@@ -184,6 +187,51 @@ public Task RemoteIPv6Address()
...
+
+            var testSink = new TestSink(write => write.EventId.Id == connectionErrorEventId);
+            var requestAbortedAcquired = new CancellationTokenSource();
+            var builder = new WebHostBuilder()
+                .UseLoggerFactory(new TestLoggerFactory(testSink, true))
+                .UseKestrel()
+                .UseUrls($"http://127.0.0.1:0")
+                .Configure(app => app.Run(context =>
+                {
+                    return Task.FromResult(0);
```
halter73: We should also assert that `context.Request.Body.ReadAsync` fails with an `IOException`.

I think that `SocketInputExtensions.ReadAsync` and possibly a few other places will have to be updated to check if there is an error (possibly by calling `SocketInput.GetResult()`) in a thread-safe way before exiting due to `RemoteIntakeFin` being set.

cesarbs: Shouldn't that be the case for any connection error? Why aren't we doing that already?

If that's further work not specific to ECONNRESET I'll create a separate task to address this.

cesarbs: I've opened https://github.com/aspnet/KestrelHttpServer/issues/959 to address this.

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/RequestTests.cs
```diff
@@ -184,6 +187,51 @@ public Task RemoteIPv6Address()
...
+                {
+                    socket.Connect(new IPEndPoint(IPAddress.Loopback, host.GetPort()));
+                    socket.LingerState = new LingerOption(true, 0);
+                }
+
+                // Give it some time for the connection error log message to be logged
+                var filteredWrites = testSink.Writes.Where(write => write.EventId.Id == connectionErrorEventId);
+                for (var i = 0; i < 10; i++)
+                {
+                    System.Console.WriteLine(i);
```
mikeharder: Remove call to `Console.WriteLine()`?

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/RequestTests.cs
```diff
@@ -184,6 +187,51 @@ public Task RemoteIPv6Address()
...
+
+                using (var socket = new Socket(AddressFamily.InterNetwork, SocketType.Stream, ProtocolType.Tcp))
+                {
+                    socket.Connect(new IPEndPoint(IPAddress.Loopback, host.GetPort()));
+                    socket.LingerState = new LingerOption(true, 0);
+                }
+
+                // Give it some time for the connection error log message to be logged
+                var filteredWrites = testSink.Writes.Where(write => write.EventId.Id == connectionErrorEventId);
+                for (var i = 0; i < 10; i++)
```
mikeharder: I would make the test code itself wait forever (rather than only 500ms), then wrap the whole test in a call to `Task.Run() .TimeoutAfter()` with a large timeout (say 10 seconds), so it fails if it exceeds the timeout for any reason.

cesarbs:
@halter73 Now handling error when checking for FIN.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/SocketInput.cs
```diff
@@ -340,5 +349,15 @@ private static void ReturnBlocks(MemoryPoolBlock block, MemoryPoolBlock end)
                 returnBlock.Pool.Return(returnBlock);
             }
         }
+
+        private void ThrowOnConnectionError()
```
davidfowl: 😞

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/SocketInput.cs
```diff
@@ -45,6 +45,19 @@ public SocketInput(MemoryPool memory, IThreadPool threadPool, IBufferSizeControl
 
         public bool IsCompleted => ReferenceEquals(_awaitableState, _awaitableIsCompleted);
 
+        public bool CheckFinOrThrow()
```
halter73: Should `RemoteIntakeFin` be public now? Can we make it tri-state to avoid the additional locking?

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/SocketInput.cs
halter73: @davidfowl You're now well-versed in this code. Do you have a better alternative?

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/SocketInput.cs
cesarbs: I was thinking of making it internal. But why make it tri-state since you're either getting a bool from `CheckFinOrThrow` or an exception?

cesarbs: Ooh I see what you mean, tri-state so we avoid races between `RemoteIntakeFin` and `_awaitableError` being inconsistent with each other, right?

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/SocketInput.cs
davidfowl: Can we just use a Task instead?

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/SocketInput.cs
davidfowl: Make it `Task Completion` (which is basically a tri-state)

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/SocketInput.cs
halter73: I also like this idea @cesarbs.

cesarbs:
Updated to use `TaskCompletionSource` to combine FIN and error states, as suggested by @davidfowl.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/SocketInput.cs
```diff
@@ -33,6 +32,8 @@ public class SocketInput : ICriticalNotifyCompletion, IDisposable
         private bool _consuming;
         private bool _disposed;
 
+        private TaskCompletionSource<bool> _tcs = new TaskCompletionSource<bool>();
```
halter73: `TaskCompletionSource<bool>` implies `Result` is useful and might be `false`. Since it never will be, `TaskCompletionSource<object>` is probably better.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/SocketInput.cs
```diff
@@ -41,10 +42,19 @@ public SocketInput(MemoryPool memory, IThreadPool threadPool, IBufferSizeControl
             _awaitableState = _awaitableIsNotCompleted;
         }
 
-        public bool RemoteIntakeFin { get; set; }
-
         public bool IsCompleted => ReferenceEquals(_awaitableState, _awaitableIsCompleted);
 
+        internal bool ReadingInput => _tcs.Task.Status == TaskStatus.WaitingForActivation;
```
halter73: private?

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/SocketInput.cs
```diff
@@ -41,10 +42,19 @@ public SocketInput(MemoryPool memory, IThreadPool threadPool, IBufferSizeControl
...
+
+        public bool CheckFinOrThrow()
+        {
+            lock (_sync)
```
davidfowl: Remind me why we need to lock?

halter73: I think this was only required when we stored `_awaitableError` and `RemoteIntakeFin` in separate fields.

We can't just remove the lock with the code as-is, because there is the risk that this would return true after the connection is faulted. I think the lock could be removed if the code changed to:

    CheckConnectionError();
    return _tcs.Task.Status == TaskStatus.RanToCompletion;

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/RequestTests.cs
```diff
@@ -184,6 +188,96 @@ public Task RemoteIPv6Address()
...
+                        await context.Request.Body.ReadAsync(new byte[1], 0, 1);
+                    }
+                    catch (BadHttpRequestException)
+                    {
+                        // We need this here because BadHttpRequestException derives from IOException,
+                        // and we're looking for an actual IOException.
+                    }
+                    catch (IOException)
+                    {
+                        ioExceptionThrown = true;
```
halter73: You could capture the exception and check that the inner exception is a `TaskCanceledException` with a "The request was aborted" message.

halter73:
:shipit:

cesarbs:
Updated tests:

* Check type of `InnerException` inside caught `IOException` (@halter73 we get a `UvException` here, not a `TaskCanceledException`)
* Add timeouts to `WaitAsync` calls on semaphores
* Use custom test sink to look for expected log message on connection reset. With the previous approach there would sometimes be errors when calling `Any()` on the writes collection while it was being modified

cesarbs:
Had to change `ThrowsOnReadAfterConnectionError` to handle two possible exception types thrown on an attempt to read from an errored connection:

* `IOException`: this was in the original test. We get one if `ReadAsync` is awaiting on `SocketInput` when the connection is aborted.
* `TaskCanceledException`: this is the new one. We get one if `ReadAsync` is called by the time the connection has already been aborted after an error.

This is quite ugly. We should probably address this and have a single behavior, but I'd like it to be addressed as a separate task.

Issue #175, opened 2015-08-26
Handle response content-length mis-match errors


Tratcher:
If the application sets the content-length response header and then does not write that exact number of bytes then this corrupts the connection state. There are three scenarios:
A) The application writes more bytes than specified. The client will truncate at the specified content-length and then fail trying to parse the remainder as the start of the next response. The client should then abort the connection.
B) The applications writes fewer bytes than specified. The client will hang waiting for the remaining data. Eventually it should time out and disconnect.
C) The application specifies a non-zero content length but doesn't write any data. The client will hang just as in (B)

For (A) it would be nice to track the content length and throw from Write if too many bytes are written. The connection should be aborted.

For (B) we need to track the content length and disconnect if the response ends without writing enough data. There's no way to raise this error to the application except in the logs. This scenario should be treated the same as if there were an unhandled exception after the response body had started.

(C) we could detect even without tracking content written because we know if the response has started or not. The response should be changed to a 500 internal server error and logged as an error.

muratg:
@lodejard What do you think about this one?

cesarbs:
I'll chime in. It's a nice feature but I wonder if it's worth the added complexity to the code.

* Why/when would any of those 3 scenarios take place?
* How often are them expected to occur?

All 3 scenarios manifest their symptoms in the client, which would eventually lead to an investigation on the server side. Unless Kestrel itself could be the source of these issues, I'd say it's up to the application to ensure it is sending the right data.

Tratcher:
Application developers make mistakes. This is more about how readily can they diagnose those mistakes. B and C are difficult to diagnose as they manifest as hangs on the client side (or data corruption in a pipelining scenario). The average browser has built in mitigations for A where they just drain any remaining data.

muratg:
Moving this to Backlog as we will be in RC2 ask mode very soon.  If you feel strongly about this issue, please ping me.

lodejard:
Catching up on notification backlog - we talked about this one the other day actually. This could be part of a middleware in the diagnostics package, actually, where response length and numerous other "application correctness" checks could be done by wrapping many features of the http request/response

cesarbs:
@lodejard I like that idea. I don't think Kestrel should be trying to account for potential application mistakes. It would make every correct application pay a price for whatever applications out there that are not doing the right thing regarding Content-Length.

Tratcher:
[...]

PR #962, opened 2016-07-09
Don't emit TE header or body for non-body responses

### LINKED ISSUE #952: Random 502 cgi gateway error when response 304 then flush response
_From @303248153 on July 2, 2016 22:58_

First I found this problem with my project, sometime static files will return 502 with cgi gateway error.
Today I do a diggup, to reproduce this problem you needs:

- Host asp.net core application on IIS, if you run with kestrel only, this problem won't happen.
- Response 304, only 304 cause this problem.
- Explict flush the response
- Do the request parallel

I uploaded a minimal reproduce project, please check it.
It maybe a thread race condition problem because only parallel request will cause this problem.
[WebApplication1.zip](https://github.com/aspnet/IISIntegration/files/344769/WebApplication1.zip)

I don't known if it's same with other 502.3 issues,
and I already using asp.net core 1.0.0 as you can see in the project.json.

_Copied from original issue: aspnet/IISIntegration#219_

benaadams:
Now doesn't emit TE or body for non body responses

Resolves #952

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
@@ -58,6 +58,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Http
 
         private RequestProcessingStatus _requestProcessingStatus;
         protected bool _keepAlive;
+        private bool _canHaveBody;
         private bool _autoChunk;
         protected Exception _applicationException;
 
@@ -135,7 +136,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Http
             {
                 if (HasResponseStarted)
                 {
-                    ThrowResponseAlreadyStartedException(nameof(StatusCode));
+                    BadHttpResponse.ThrowException(ResponseRejectionReasons.ValueCannotBeSetResponseStarted, ResponseRejectionParameter.StatusCode);
                 }
 
                 _statusCode = value;
@@ -153,7 +154,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Http
             {
                 if (HasResponseStarted)
                 {
-                    ThrowResponseAlreadyStartedException(nameof(ReasonPhrase));
+                    BadHttpResponse.ThrowException(ResponseRejectionReasons.ValueCannotBeSetResponseStarted, ResponseRejectionParameter.ReasonPhrase);
                 }
 
                 _reasonPhrase = value;
@@ -388,7 +389,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Http
             {
                 if (HasResponseStarted)
                 {
-                    ThrowResponseAlreadyStartedException(nameof(OnStarting));
+                    BadHttpResponse.ThrowException(ResponseRejectionReasons.OnStartingCannotBeSetResponseStarted, ResponseRejectionParameter.OnStarting);
                 }
 
                 if (_onStarting == null)
@@ -475,17 +476,24 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Http
         {
             ProduceStartAndFireOnStarting().GetAwaiter().GetResult();
 
-            if (_autoChunk)
+            if (_canHaveBody)
             {
-                if (data.Count == 0)
+                if (_autoChunk)
                 {
-                    return;
+                    if (data.Count == 0)
+                    {
+                        return;
+                    }
+                    WriteChunked(data);
                 }
-                WriteChunked(data);
+                else
+                {
+                    SocketOutput.Write(data);
+                }
             }
             else
             {
-                SocketOutput.Write(data);
+                HandleNonBodyResponseWrite(data.Count);
             }
         }
 
@@ -496,17 +504,25 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Http
                 return WriteAsyncAwaited(data, cancellationToken);
             }
 
-            if (_autoChunk)
+            if (_canHaveBody)
             {
-                if (data.Count == 0)
+                if (_autoChunk)
                 {
-                    return TaskUtilities.CompletedTask;
+                    if (data.Count == 0)
+                    {
+                        return TaskUtilities.CompletedTask;
+                    }
+                    return WriteChunkedAsync(data, cancellationToken);
                 }
-                return WriteChunkedAsync(data, cancellationToken);
+                else
+                {
+                    return SocketOutput.WriteAsync(data, cancellationToken: cancellationToken);
+                }
             }
             else
             {
-                return SocketOutput.WriteAsync(data, cancellationToken: cancellationToken);
+                HandleNonBodyResponseWrite(data.Count);
+                return TaskUtilities.CompletedTask;
             }
         }
 
@@ -514,18 +530,27 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Http
         {
             await ProduceStartAndFireOnStarting();
 
-            if (_autoChunk)
+            if (_canHaveBody)
             {
-                if (data.Count == 0)
+                if (_autoChunk)
                 {
-                    return;
+                    if (data.Count == 0)
+                    {
+                        return;
+                    }
+                    await WriteChunkedAsync(data, cancellationToken);
                 }
-                await WriteChunkedAsync(data, cancellationToken);
+                else
+                {
+                    await SocketOutput.WriteAsync(data, cancellationToken: cancellationToken);
+                }
             }
             else
             {
-                await SocketOutput.WriteAsync(data, cancellationToken: cancellationToken);
+                HandleNonBodyResponseWrite(data.Count);
+                return;
             }
+
         }
 
         private void WriteChunked(ArraySegment<byte> data)
@@ -640,28 +665,14 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Http
 
                 if (_requestRejected)
                 {
-                    // 400 Bad Request
-                    StatusCode = 400;
                     _keepAlive = false;
+                    // 400 Bad Request
+                    ErrorResetHeadersToDefaults(statusCode: 400);
                 }
                 else
                 {
                     // 500 Internal Server Error
-                    StatusCode = 500;
-                }
-
-                ReasonPhrase = null;
-
-                var responseHeaders = FrameResponseHeaders;
-                responseHeaders.Reset();
-                var dateHeaderValues = DateHeaderValueManager.GetDateHeaderValues();
-
-                responseHeaders.SetRawDate(dateHeaderValues.String, dateHeaderValues.Bytes);
-                responseHeaders.SetRawContentLength("0", _bytesContentLengthZero);
-
-                if (ServerOptions.AddServerHeader)
-                {
-                    responseHeaders.SetRawServer(Constants.ServerName, _bytesServer);
+                    ErrorResetHeadersToDefaults(statusCode: 500);
                 }
             }
 
@@ -715,9 +726,11 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Http
             bool appCompleted)
         {
             var responseHeaders = FrameResponseHeaders;
-            responseHeaders.SetReadOnly();
 
             var hasConnection = responseHeaders.HasConnection;
+
+            // Set whether response can have body
+            _canHaveBody = StatusCanHaveBody(StatusCode) && Method != "HEAD";
 
             var end = SocketOutput.ProducingStart();
             if (_keepAlive && hasConnection)
@@ -725,40 +738,49 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Http
                 var connectionValue = responseHeaders.HeaderConnection.ToString();
                 _keepAlive = connectionValue.Equals("keep-alive", StringComparison.OrdinalIgnoreCase);
             }
-
-            if (!responseHeaders.HasTransferEncoding && !responseHeaders.HasContentLength)
+            
+            if (_canHaveBody)
             {
-                if (appCompleted)
+                if (!responseHeaders.HasTransferEncoding && !responseHeaders.HasContentLength)
                 {
-                    // Don't set the Content-Length or Transfer-Encoding headers
-                    // automatically for HEAD requests or 101, 204, 205, 304 responses.
-                    if (Method != "HEAD" && StatusCanHaveBody(StatusCode))
+                    if (appCompleted)
                     {
                         // Since the app has completed and we are only now generating
                         // the headers we can safely set the Content-Length to 0.
                         responseHeaders.SetRawContentLength("0", _bytesContentLengthZero);
                     }
-                }
-                else if(_keepAlive)
-                {
-                    // Note for future reference: never change this to set _autoChunk to true on HTTP/1.0
-                    // connections, even if we were to infer the client supports it because an HTTP/1.0 request
-                    // was received that used chunked encoding. Sending a chunked response to an HTTP/1.0
-                    // client would break compliance with RFC 7230 (section 3.3.1):
-                    //
-                    // A server MUST NOT send a response containing Transfer-Encoding unless the corresponding
-                    // request indicates HTTP/1.1 (or later).
-                    if (_httpVersion == Http.HttpVersion.Http11)
-                    {
-                        _autoChunk = true;
-                        responseHeaders.SetRawTransferEncoding("chunked", _bytesTransferEncodingChunked);
-                    }
                     else
                     {
-                        _keepAlive = false;
+                        // Note for future reference: never change this to set _autoChunk to true on HTTP/1.0
+                        // connections, even if we were to infer the client supports it because an HTTP/1.0 request
+                        // was received that used chunked encoding. Sending a chunked response to an HTTP/1.0
+                        // client would break compliance with RFC 7230 (section 3.3.1):
+                        //
+                        // A server MUST NOT send a response containing Transfer-Encoding unless the corresponding
+                        // request indicates HTTP/1.1 (or later).
+                        if (_httpVersion == Http.HttpVersion.Http11)
+                        {
+                            _autoChunk = true;
+                            responseHeaders.SetRawTransferEncoding("chunked", _bytesTransferEncodingChunked);
+                        }
+                        else
+                        {
+                            _keepAlive = false;
+                        }
                     }
                 }
             }
+            else
+            {
+                // Don't set the Content-Length or Transfer-Encoding headers
+                // automatically for HEAD requests or 101, 204, 205, 304 responses.
+                if (responseHeaders.HasTransferEncoding)
+                {
+                    RejectNonBodyTransferEncodingResponse(appCompleted);
+                }
+            }
+
+            responseHeaders.SetReadOnly();
 
             if (!_keepAlive && !hasConnection)
             {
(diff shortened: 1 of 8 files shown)


Tratcher:
So this allows 304 responses with bodies & connection: close? How well do browsers handle that?

benaadams:
Is progress anyway... Will add an IO execption on write if status code doesn't allow body?

benaadams:
Is progress anyway... Will add an IO exception on write if status code doesn't allow body? Or just dump the writes?

benaadams:
Probably should dump the body then HEAD requests will work

benaadams:
Now doesn't emit TE or body for non body responses

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
```diff
@@ -486,6 +490,9 @@ public Task WriteAsync(ArraySegment<byte> data, CancellationToken cancellationTo
                 return WriteAsyncAwaited(data, cancellationToken);
             }
 
+            // Don't write to body for HEAD requests or 101, 204, 205, 304 responses.
+            if (!_canHaveBody) return TaskUtilities.CompletedTask;
```
Tratcher: What about WriteAsyncAwaited?

Tratcher:
This is an interesting way to enforce the protocol. The downside is that it's silent and the user doesn't know what they've done wrong.  E.g. they would keep trying to copy a 5gb file not realizing they're responding to a HEAD request. Something like this should at least be logged as Debug.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
benaadams: Added

cesarbs:
@benaadams Can you add tests?

benaadams:
@Tratcher Added logging as per `ConnectionDisconnectedWrite`
@cesarbs will do

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
```diff
@@ -465,6 +466,13 @@ public void Write(ArraySegment<byte> data)
         {
             ProduceStartAndFireOnStarting().GetAwaiter().GetResult();
 
+            if (!_canHaveBody)
+            {
+                // Don't write to body for HEAD requests or 101, 204, 205, 304 responses.
+                LogNonBodyRequestWrite(data.Count);
```
halter73: I think this should throw if there is any actual data. Flush seems fine (though unnecessary), but not anything else.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
```diff
@@ -727,39 +759,49 @@ private Task WriteSuffix()
...
                 }
             }
+            else
+            {
+                // Don't set the Content-Length or Transfer-Encoding headers
+                // automatically for HEAD requests or 101, 204, 205, 304 responses.
+                if (responseHeaders.HasTransferEncoding)
+                {
+                    // Remove Transfer-Encoding encoding header if it exists
+                    responseHeaders.Remove("Transfer-Encoding");
```
halter73: I prefer an error/exception if the user set the `Transfer-Encoding` header instead of silently removing it.

benaadams:
@halter73 re: execptions, depends how you want HEAD requests to work; do they work automatically out of the box or does the user need to code a specific response for HEAD?

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
benaadams: User generally wouldn't and it would be handled automatically? So this would be user error.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
benaadams: Question on whether GET responses can serve HEAD requests?

Can differentiate the response codes from method, so 101, 204, 205, 304 throw and HEAD eats?

benaadams:
@cesarbs added tests - however it still needs test to check the socket output (e.g. that its sending 500s and not doing something strange when it errors)

cesarbs:
@benaadams Can you rebase?

test/Microsoft.AspNetCore.Server.KestrelTests/TestFrameProtectedMembers.cs
```diff
@@ -0,0 +1,23 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
```
cesarbs: nit: remove extra blank line

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
```diff
@@ -1198,14 +1220,68 @@ public bool StatusCanHaveBody(int statusCode)
...
+            }
+            else
+            {
+                ReportApplicationError(ex);
+                // 500 Internal Server Error
+                ErrorResetHeadersToDefaults(statusCode: 500);
+            }
+        }
+
+        private void ErrorResetHeadersToDefaults(int statusCode)
```
cesarbs: nit: I'd prefer a different name. I'm thinking of `PrepareErrorResponse` or `SetErrorResponse`.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
```diff
@@ -1198,14 +1220,68 @@ public bool StatusCanHaveBody(int statusCode)
...
+                // Throw Exception for 101, 204, 205, 304 responses.
+                throw BadHttpResponseException.GetException(ResponseRejectionReasons.WriteToNonBodyResponse, StatusCode.ToString());
+            }
         }
 
         private void ThrowResponseAbortedException()
         {
             throw new ObjectDisposedException(
+                    "The response has been aborted due to an unhandled application exception.",
+                    _applicationException);
```
cesarbs: nit: single indent level

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
```diff
@@ -1198,14 +1220,68 @@ public bool StatusCanHaveBody(int statusCode)
...
+        }
+
+        public void RejectRequest(string message)
```
cesarbs: Wouldn't that be `RejectResponse`?

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/ResponseRejectionReasons.cs
```diff
@@ -0,0 +1,13 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Http
+{
+    public enum ResponseRejectionReasons
+    {
+        HeadersReadonlyResponseStarted,
+        ValueCannotBeSetResponseStarted,
```
cesarbs: I'd flip these two names around:

`ResponseStartedHeadersReadonly`
`ResponseStartedValueCannotBeSet`

test/Microsoft.AspNetCore.Server.KestrelTests/FrameTests.cs
```diff
@@ -520,5 +521,182 @@ public void InitializeStreamsResetsStreams()
...
+                SocketOutput = new MockSocketOuptut(),
+                Log = new TestKestrelTrace()
+            };
+            var frame = new TestFrameProtectedMembers<object>(application: null, context: connectionContext);
+            frame.InitializeHeaders();
+            frame.KeepAlive = true;
+            frame.HttpVersion = "HTTP/1.1";
+            ((IHttpRequestFeature)frame).Method = "HEAD";
+
+            //Act
```
cesarbs: nit: space after `//`

test/Microsoft.AspNetCore.Server.KestrelTests/FrameTests.cs
```diff
@@ -520,5 +521,182 @@ public void InitializeStreamsResetsStreams()
...
+                SocketOutput = new MockSocketOuptut(),
+                Log = new TestKestrelTrace()
+            };
+            var frame = new TestFrameProtectedMembers<object>(application: null, context: connectionContext);
+            frame.InitializeHeaders();
+            frame.KeepAlive = true;
+            frame.HttpVersion = "HTTP/1.1";
+            ((IHttpResponseFeature)frame).StatusCode = 304;
+
+            //Act
```
cesarbs: nit: space after `//`

src/Microsoft.AspNetCore.Server.Kestrel/BadHttpResponseException.cs
```diff
@@ -0,0 +1,56 @@
...
+                    break;
+                default:
+                    ex = new BadHttpResponseException("Bad response.");
+                    break;
+            }
+
+            return ex;
+        }
+
+        internal static BadHttpResponseException GetException(ResponseRejectionReasons reason, string value)
```
cesarbs: You could make `value` a generic arg here, so you wouldn't have to call `ToString()` on `StatusCode` below.

src/Microsoft.AspNetCore.Server.Kestrel/BadHttpResponseException.cs
```diff
@@ -0,0 +1,56 @@
...
+                case ResponseRejectionReasons.WriteToNonBodyResponse:
+                    ex = new BadHttpResponseException($"Write to non-body {value} response.");
+                    break;
+                default:
+                    ex = new BadHttpResponseException("Bad response.");
+                    break;
+            }
+
+            return ex;
+
```
cesarbs: nit: remove blank line

cesarbs:
@benaadams I've added a few more comments. Address those and I think we're good to go :+1:

benaadams:
Broke the new tests :(

benaadams:
Ah catching wrong exceptions :)

benaadams:
@cesarbs probably want to do another round on the exceptions when best approach in coreclr becomes more apparent.

benaadams:
Microsoft.AspNetCore.Server.KestrelTests.SocketOutputTests.WritesDontCompleteImmediatelyWhenTooManyBytesAreAlreadyPreCompleted [FAIL] https://github.com/aspnet/KestrelHttpServer/issues/1002

src/Microsoft.AspNetCore.Server.Kestrel/BadHttpResponseException.cs
```diff
@@ -0,0 +1,71 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using Microsoft.AspNetCore.Server.Kestrel.Internal.Http;
+
+namespace Microsoft.AspNetCore.Server.Kestrel
+{
+    public sealed class BadHttpResponseException : InvalidOperationException
```
halter73: I think this exception type would only be appropriate for attempting to write a body when it's not allowed by the spec. In the other cases, like modifying the headers/statuscode/callbacks too late, I think a `BadHttpResponseException` is confusing. I would prefer an `InvalidOperationException` in all cases. It's not like this is an exception type you would have want to handle specifically. It should always be avoidable.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
```diff
@@ -1214,9 +1236,56 @@ public bool StatusCanHaveBody(int statusCode)
...
+            }
+        }
+
+        public void HandleNonBodyResponseWrite(int count)
         {
-            throw new InvalidOperationException(value + " cannot be set, response had already started.");
+            if (Method == "HEAD")
+            {
+                // Don't write to body for HEAD requests.
+                if (Log.IsEnabled(LogLevel.Debug))
```
halter73: `Log.IsEnabled` should already be checked by the action returned by `LoggerMessage.Define`.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
```diff
@@ -1214,9 +1236,56 @@ public bool StatusCanHaveBody(int statusCode)
...
+                {
+                    Log.ConnectionHeadResponseBodyWrite(ConnectionId, count);
```
halter73: So we're not throwing here because a lot of real apps will write responses without looking at the request method?

benaadams: Yeah, normally for GET type requests don't bother checking if GET/POST/HEAD etc. So is getting HEAD behaviour for free(ish)

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
```diff
@@ -1214,9 +1236,56 @@ public bool StatusCanHaveBody(int statusCode)
...
         }
 
-        private void ThrowResponseAlreadyStartedException(string value)
+        private void RejectNonBodyTransferEncodingResponse(bool appCompleted)
+        {
+            var ex = BadHttpResponseException.GetException(ResponseRejectionReason.NonBodyResponseTransferEncodingSet, StatusCode);
+            if (!appCompleted)
+            {
+                // Back out of header creation surface exeception in user code
+                _requestProcessingStatus = RequestProcessingStatus.RequestStarted;
```
halter73: If we're logging an not throwing for HEAD requests in `HandleNonBodyResponseWrite`, shouldn't we just remove the header and do the same here? I think we should be consistent.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
```diff
@@ -728,39 +741,48 @@ private Task WriteSuffix()
...
+                            _keepAlive = false;
+                        }
                     }
                 }
             }
+            else
+            {
+                // Don't set the Content-Length or Transfer-Encoding headers
+                // automatically for HEAD requests or 101, 204, 205, 304 responses.
+                if (responseHeaders.HasTransferEncoding)
```
halter73: Why not reject if the `Content-Length` header is set while we're at it?

benaadams: Content-Length can be part of the HEAD request; as long as it is what GET would say.

https://tools.ietf.org/html/rfc7230#section-3.3.2
> A server MAY send a Content-Length header field in a response to a
   HEAD request (Section 4.3.2 of [RFC7231]); a server MUST NOT send
   Content-Length in such a response unless its field-value equals the
   decimal number of octets that would have been sent in the payload
   body of a response if the same request had used the GET method.

> A server MAY send a Content-Length header field in a 304 (Not
   Modified) response to a conditional GET request (Section 4.1 of
   [RFC7232]); a server MUST NOT send Content-Length in such a response
   unless its field-value equals the decimal number of octets that would
   have been sent in the payload body of a 200 (OK) response to the same
   request.

benaadams: e.g. don't set it; but don't remove it or error it

src/Microsoft.AspNetCore.Server.Kestrel/BadHttpResponseException.cs
benaadams: Differentiate server created exceptions from user code created exceptions?

halter73: I wouldn't consider these server created exceptions. These are application created exceptions. Most APIs will throw plain `InvalidOperationException`s if called incorrectly.

I actually don't mind `BadHttpResponseException` for the caller attempting to do things that would break the HTTP spec like setting `Transfer-Encoding` and a non-body status. I personally wouldn't do it, but I don't mind.

Throwing a `BadHttpResponseException` for calling `OnStarted` too late or attempting to add response headers too late seems confusing and wrong though.

test/Microsoft.AspNetCore.Server.KestrelTests/FrameTests.cs
```diff
@@ -540,5 +541,182 @@ public void InitializeStreamsResetsStreams()
...
+            ((IHttpRequestFeature)frame).Method = "HEAD";
+
+            // Assert
+            frame.Flush(); // Does not throw
+
+            frame.Write(new ArraySegment<byte>(new byte[1]));
+
+            frame.Flush(); // Does not throw
+        }
+
```
cesarbs: nit: remove extra blank line

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/ResponseTests.cs
```diff
@@ -159,7 +159,7 @@ public class ResponseTests
                     {
                         await context.Response.WriteAsync("hello, world");
                         await context.Response.Body.FlushAsync();
-                        ex = Assert.Throws<InvalidOperationException>(() => context.Response.OnStarting(_ => TaskUtilities.CompletedTask, null));
+                        ex = Assert.Throws<BadHttpResponseException>(() => context.Response.OnStarting(_ => TaskUtilities.CompletedTask, null));
```
cesarbs: I agree with @halter73 that this should continue throwing `InvalidOperationException`.

benaadams:
Rebased, some merge clash bugs still

cesarbs:
I have a fix for the keep alive test failure here: https://github.com/aspnet/KestrelHttpServer/pull/1090

cesarbs:
@benaadams Can you rebase?

benaadams:
I think rebase is going to be messy...

cesarbs:
One thing I've noticed: `StatusCanHaveBody` should be changed to allow 101 to have a body. Otherwise WebSockets are broken because `Write` and `WriteAsync` will throw on writes.

I've pulled this PR to my machine and I'm working on rebasing it and fixing any test failures. It has useful stuff for #175.

benaadams:
Ok works for me 👍

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
```diff
@@ -715,50 +726,61 @@ private Task WriteSuffix()
...
-                    // automatically for HEAD requests or 101, 204, 205, 304 responses.
-                    if (Method != "HEAD" && StatusCanHaveBody(StatusCode))
+                    if (appCompleted)
                     {
                         // Since the app has completed and we are only now generating
                         // the headers we can safely set the Content-Length to 0.
                         responseHeaders.SetRawContentLength("0", _bytesContentLengthZero);
                     }
-                }
-                else if(_keepAlive)
```
cesarbs: @benaadams I'm curious why you removed this condition in your change. I had to re-introduce the keep-alive check, otherwise some tests would fail.

benaadams: Is it relevant? Unless you are saying only keep-alive can be chunked?

halter73: That's the current logic. Why bother chunking on `Connection: close` responses?

benaadams: Haven't looked into it too closely; but couldn't you be returning a 200, chunked data and a `Connection: close`?

cesarbs: We have a test that specifically checks that we don't chunk HTTP/1.1 non-keepalive responses. Is that in the spec?

benaadams: You'd have to buffer all the data then to work out the size?

halter73: Not part of the spec for HTTP/1.1. It obviously is for HTTP/1.0 requests.

Chunked or not, the client won't work out the total size of the response until the connection closes. Why go through the extra cycles doing chunking for no apparent benefit?

benaadams: Oh, you mean because the connection is closing it doesn't need a `Content-Length` or `Transfer-Encoding` header because the FIN indicates that? How does the client detect a partial response?

cesarbs: Agreed with @benaadams - if we don't chunk, the client has no way to tell whether the response ended or the connection closed abruptly. The client could in theory verify if it was a clean connection close or a reset, but we shouldn't count on that.

If it's not in the spec I say we should chunk HTTP/1.1 even on `Connection: close`.

halter73: Good point about application errors. That's a fact of life for HTTP/1.0, but reliably knowing that a response has gracefully closed is a benefit of chunked encoding even if it's a non keep-alive connection.

I'm now convinced about this changed. Just make sure we don't regress the behavior for HTTP/1.0 requests. It's easier to mess up now that all responses are HTTP/1.1.

cesarbs:
Closing in favor of #1134.

PR #964, opened 2016-07-11
Remove skip condition for IPv6 with scope ID on Linux and Mac

mikeharder:
- Skip was required to workaround CoreFx issue which is now fixed


cesarbs:
yay :ship:

Issue #794, opened 2016-05-03
Add tests for HTTPS urls to AddressRegistrationTests


mikeharder:
**Test Cases**
`https://+`
`https://localhost`
`https://localhost:GetNextPort()`
`https://localhost:0`

**Notes**
Relative path to the certificate in `UseSsl()` was not working on Travis/Linux.

PR #965, opened 2016-07-12
Add tests for HTTPS urls to AddressRegistrationTests (#794)

mikeharder:


halter73:
So the plan is to test this with `Socket/SslStream` instead of `HttpClient` so it can pass on all OS X machines?

mikeharder:
@halter73: Correct, I'm writing an HttpClientSlim which uses Socket+SslStream.

Tratcher:
https://github.com/Tratcher/HttpClient/blob/dev/src/Microsoft.Net.Http.Client/ManagedHandler.cs

mikeharder:
@halter73, @cesarbs, @Tratcher: Travis is now passing, ready for review.

halter73:
:shipit:

muratg:
Nice! 

:shipit:

PR #973, opened 2016-07-15
Remove state lock from Connection

davidfowl:
- Remove the state lock by creating the frame earlier
- Set properties on the frame that need to be set before calling start.
- Don't rely on the copy ctor of the *Context hierarchy to set
*everything*

/cc @cesarbs @halter73 @mikeharder

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Connection.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Connection.cs
@@ -35,7 +35,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Http
         private FilteredStreamAdapter _filteredStreamAdapter;
         private Task _readInputTask;
 
-        private TaskCompletionSource<object> _socketClosedTcs;
+        private TaskCompletionSource<object> _socketClosedTcs = new TaskCompletionSource<object>();
         private BufferSizeControl _bufferSizeControl;
 
         public Connection(ListenerContext context, UvStreamHandle socket) : base(context)
@@ -127,13 +127,8 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Http
 
         public Task StopAsync()
         {
-            if (_socketClosedTcs == null)
-            {
-                _socketClosedTcs = new TaskCompletionSource<object>();
-
-                _frame.Stop();
-                _frame.SocketInput.CompleteAwaiting();
-            }
+            _frame.Stop();
+            _frame.SocketInput.CompleteAwaiting();
 
             return _socketClosedTcs.Task;
         }
@@ -168,7 +163,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Http
                 SocketInput.Dispose();
             }
 
-            _socketClosedTcs?.TrySetResult(null);
+            _socketClosedTcs.TrySetResult(null);
         }
 
         private void ApplyConnectionFilter()
--- a/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Connection.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Connection.cs
@@ -35,13 +35,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Http
         private FilteredStreamAdapter _filteredStreamAdapter;
         private Task _readInputTask;
 
-        private readonly SocketInput _rawSocketInput;
-        private readonly SocketOutput _rawSocketOutput;
-
-        private readonly object _stateLock = new object();
-        private ConnectionState _connectionState;
         private TaskCompletionSource<object> _socketClosedTcs;
-
         private BufferSizeControl _bufferSizeControl;
 
         public Connection(ListenerContext context, UvStreamHandle socket) : base(context)
@@ -57,8 +51,17 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Http
                 _bufferSizeControl = new BufferSizeControl(ServerOptions.MaxRequestBufferSize.Value, this, Thread);
             }
 
-            _rawSocketInput = new SocketInput(Memory, ThreadPool, _bufferSizeControl);
-            _rawSocketOutput = new SocketOutput(Thread, _socket, Memory, this, ConnectionId, Log, ThreadPool, WriteReqPool);
+            SocketInput = new SocketInput(Memory, ThreadPool, _bufferSizeControl);
+            SocketOutput = new SocketOutput(Thread, _socket, Memory, this, ConnectionId, Log, ThreadPool, WriteReqPool);
+
+            var tcpHandle = _socket as UvTcpHandle;
+            if (tcpHandle != null)
+            {
+                RemoteEndPoint = tcpHandle.GetPeerIPEndPoint();
+                LocalEndPoint = tcpHandle.GetSockIPEndPoint();
+            }
+
+            _frame = FrameFactory(this);
         }
 
         // Internal for testing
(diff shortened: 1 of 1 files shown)


src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Connection.cs
```diff
@@ -168,18 +140,7 @@ public virtual void Abort()
...
-                    if (connection._connectionState == ConnectionState.CreatingFrame)
-                    {
-                        connection._connectionState = ConnectionState.ToDisconnect;
-                    }
-                    else
-                    {
-                        connection._frame?.Abort();
-                    }
-                }
+                connection._frame.Abort();
```
halter73: nit: `_frame.Abort()`

cesarbs:
There seems to be a slight perf increase:

dev:

    [02:31:02.477] Starting scenario Plaintext on benchmark server...
    [02:31:19.961] Warmup
    [02:31:19.961] Starting scenario Plaintext on benchmark client...
    [02:31:31.065] Scenario Plaintext completed on benchmark client
    [02:31:31.065] RPS: 1085215.49
    [02:31:31.065] Stopping scenario Plaintext on benchmark client...
    [02:31:31.065] Benchmark
    [02:31:31.065] Starting scenario Plaintext on benchmark client...
    [02:31:42.122] Scenario Plaintext completed on benchmark client
    [02:31:42.122] RPS: 1134092
    [02:31:42.122] Stopping scenario Plaintext on benchmark client...
    [02:31:42.124] Stopping scenario Plaintext on benchmark server...

PR:

    [02:31:56.848] Starting scenario Plaintext on benchmark server...
    [02:32:19.227] Warmup
    [02:32:19.235] Starting scenario Plaintext on benchmark client...
    [02:32:30.326] Scenario Plaintext completed on benchmark client
    [02:32:30.326] RPS: 1115500.55
    [02:32:30.326] Stopping scenario Plaintext on benchmark client...
    [02:32:30.326] Benchmark
    [02:32:30.326] Starting scenario Plaintext on benchmark client...
    [02:32:41.419] Scenario Plaintext completed on benchmark client
    [02:32:41.419] RPS: 1180015.06
    [02:32:41.420] Stopping scenario Plaintext on benchmark client...
    [02:32:41.421] Stopping scenario Plaintext on benchmark server...

davidfowl:
MOARR PERFF!

mikeharder:
@cesarbs: Due to noise in perf results, I suggest running 3 samples of before and after the change, as close as possible in time (i.e. before, after, before, after, before, after).  If all the "after" are faster than the "before", it's a good indication there's a real perf benefit.  Otherwise it might just be noise.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Connection.cs
```diff
@@ -57,8 +51,17 @@ public Connection(ListenerContext context, UvStreamHandle socket) : base(context
...
+            SocketOutput = new SocketOutput(Thread, _socket, Memory, this, ConnectionId, Log, ThreadPool, WriteReqPool);
+
+            var tcpHandle = _socket as UvTcpHandle;
+            if (tcpHandle != null)
+            {
+                RemoteEndPoint = tcpHandle.GetPeerIPEndPoint();
+                LocalEndPoint = tcpHandle.GetSockIPEndPoint();
+            }
+
+            _frame = FrameFactory(this);
```
halter73: Make the `ConnectionContext` copy constructor not copy `SocketInput` and `SocketOuput` and add a comment that the contract is that `Connection` sets this on `Frame` explicitely before calling `Frame.Start()`.

You could even make these two arguments to `Start`, but I'm not sure that's necessary.

davidfowl:
@mikeharder buzz kill 😆

benaadams:
I was seeing lock action with VTune https://github.com/dotnet/coreclr/issues/6144

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Connection.cs
davidfowl: I'm going to be doing more refactoring to remove the copy ctors for Frame so I don't want to do that as part of this PR. Will add a comment saying that the copy ctor values are replaced later. That sound ok?

halter73: Oh. And I think you can make `_frame` readonly for extra clarity that we reuse the object for each request.

halter73:
`+46 −129` -1 lock. Super awesome! Certainly isn't hurting perf and is way simpler! :+1:

cesarbs:
@mikeharder I should've mentioned I did exactly that. Numbers were pretty consistent.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Connection.cs
halter73: Fine with the comment. The comment could point out `SocketInput` and `SocketOutput` will always be null initially.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Connection.cs
```diff
@@ -141,24 +127,10 @@ public void Start()
...
-                    case ConnectionState.CreatingFrame:
-                        _connectionState = ConnectionState.ToDisconnect;
-                        break;
-                    case ConnectionState.Open:
-                        _frame.Stop();
-                        SocketInput.CompleteAwaiting();
-                        break;
-                }
+            _frame.Stop();
+            _frame.SocketInput.CompleteAwaiting();
```
halter73: Gotta be careful :stuck_out_tongue: 

`_frame.SocketInput?.CompleteAwaiting();`

halter73: Might actually want to add a test for this. Could be done with a custom connection filter.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Connection.cs
```diff
@@ -189,65 +150,45 @@ public virtual void OnSocketClosed()
...
-                    else
-                    {
-                        SocketInput = _rawSocketInput;
-                        SocketOutput = _rawSocketOutput;
-                    }
+            _frame.PrepareRequest = _filterContext.PrepareRequest;
 
-                    PrepareRequest = _filterContext.PrepareRequest;
+            // Reset needs to be called here so prepare request gets applied
+            _frame.Reset();
```
halter73: I think `Frame.Start` should call `Frame.Reset` then, and we shouldn't call it in the `Frame` ctor.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Connection.cs
```diff
@@ -342,28 +283,13 @@ void IConnectionControl.End(ProduceEndType endType)
...
-                        {
-                            return;
-                        }
-                        _connectionState = ConnectionState.Disconnecting;
-
-                        Log.ConnectionDisconnect(ConnectionId);
-                        _rawSocketOutput.End(endType);
-                        break;
-                    }
+                    Log.ConnectionDisconnect(ConnectionId);
```
halter73: Might be better to log this in `OnSocketClosed` now. It was arguably the better place to begin with, and we shouldn't risk duplicate logs when it's so easily avoidable.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Connection.cs
davidfowl: It won't be null though.

halter73: It would have been if you changed the copy constructor! I guess this means https://github.com/aspnet/KestrelHttpServer/pull/973/files#r71045330 is unnecessary.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Connection.cs
```diff
@@ -73,35 +76,18 @@ public void Start()
...
-                    SocketOutput = _rawSocketOutput;
-
-                    _frame = CreateFrame();
-                    _frame.Start();
-                }
             }
             else
             {
-                _libuvStream = new LibuvStream(_rawSocketInput, _rawSocketOutput);
+                _libuvStream = new LibuvStream(SocketInput, SocketOutput);
```
halter73: I'm surprised at your restraint in not renaming this.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Connection.cs
```diff
@@ -189,65 +150,45 @@ public virtual void OnSocketClosed()
...
             }
 
-            lock (_stateLock)
-            {
-                _connectionState = ConnectionState.SocketClosed;
-
-                if (_socketClosedTcs != null)
-                {
-                    // This is always waited on synchronously, so it's safe to
-                    // call on the libuv thread.
```
halter73: Might want to keep this comment. Maybe even add that it's only `KestrelThread.Stop()` that uses it. (or `Listener` until your other PR is merged)

halter73:
:shipit: pending comments.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Connection.cs
davidfowl: Impressive I know

PR #974, opened 2016-07-15
Allocate things per KestrelThread instead of per listener

davidfowl:
- Moved things that have loop affinity to KestrelThread like
WriteReqPool, MemoryPool and the ConnectionManager
- Changed on the listeners to only kill the ListenSocket, not the
connections on dispose
- Moved connection disposal to KestrelThread.Stop
- Simplify the connection manager logic so it's possible to walk and
wait in a single call

(Note: Builds on no-connection-locks but that doesn't really matter)

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/ConnectionManager.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/ConnectionManager.cs
@@ -3,30 +3,34 @@
 
 using System;
 using System.Collections.Generic;
+using System.Threading;
 using System.Threading.Tasks;
+using Microsoft.AspNetCore.Server.Kestrel.Internal.Infrastructure;
 using Microsoft.AspNetCore.Server.Kestrel.Internal.Networking;
 
 namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Http
 {
     public class ConnectionManager
     {
-        private KestrelThread _thread;
-        private List<Task> _connectionStopTasks;
+        private readonly KestrelThread _thread;
+        private readonly IThreadPool _threadPool;
 
-        public ConnectionManager(KestrelThread thread)
+        public ConnectionManager(KestrelThread thread, IThreadPool threadPool)
         {
             _thread = thread;
+            _threadPool = threadPool;
         }
 
-        // This must be called on the libuv event loop
         public void WalkConnectionsAndClose()
         {
-            if (_connectionStopTasks != null)
-            {
-                throw new InvalidOperationException($"{nameof(WalkConnectionsAndClose)} cannot be called twice.");
-            }
+            var wh = new ManualResetEventSlim();
+            _thread.Post(state => ((ConnectionManager)state).WalkConnectionsAndCloseCore(wh), this);
+            wh.Wait();
+        }
 
-            _connectionStopTasks = new List<Task>();
+        private void WalkConnectionsAndCloseCore(ManualResetEventSlim wh)
+        {
+            var connectionStopTasks = new List<Task>();
 
             _thread.Walk(ptr =>
             {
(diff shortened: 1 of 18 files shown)


cesarbs:
Nice. But what's the purpose of this change? Does it improve perf?

cesarbs:
No perf gain but no impact either, so :+1:

halter73:
@cesarbs Since the number of listeners (primary and secondary) is `numKestrelThreads * numListeningAddresses`, having a pool per listener was wasteful. Since the only reason for multiple pools in the first place was for some degree of thread affinity, might as well store the pool directly with a thread.

Functionally, nothing changes if you are only listening to one address, but this is better if you are listening to multiple addresses.

It's also cleaner, since I wrote the Listener shutdown logic as if there was one listener per thread. Each listener would walk all libuv handles on the thread and close the during ungraceful shutdown.  This resulted in listeners closing handles that really didn't belong to them. It's not a huge deal, since those handles should be closed any during ungraceful shutdown, but it makes more sense for the thread to walk the handles.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/ConnectionManager.cs
```diff
@@ -3,30 +3,34 @@
...
 
-        // This must be called on the libuv event loop
         public void WalkConnectionsAndClose()
         {
-            if (_connectionStopTasks != null)
-            {
-                throw new InvalidOperationException($"{nameof(WalkConnectionsAndClose)} cannot be called twice.");
-            }
+            var wh = new ManualResetEventSlim();
+            _thread.Post(state => ((ConnectionManager)state).WalkConnectionsAndCloseCore(wh), this);
```
halter73: But the closure! Jk.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Infrastructure/KestrelThread.cs
```diff
@@ -18,6 +19,8 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal
     /// </summary>
     public class KestrelThread
     {
+        public const int MaxPooledWriteReqs = 1024;
```
halter73: You could unconst this and set it to zero before disposing all the `UvWriteReq`s in `Stop`. This will prevent new `UvWriteReq`s from potentially being added to the pool afterwards and finalized.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Infrastructure/KestrelThread.cs
```diff
@@ -85,6 +97,22 @@ public void AllowStop()
 
         public void Stop(TimeSpan timeout)
         {
+            // Close and wait for all connections
+            ConnectionManager.WalkConnectionsAndClose();
+
+            // REVIEW: Should we use the timeout?
```
halter73: Not timing out seems risky. Better for `Stop` to throw than hang here IMHO. If we change this, `WalkConnectionsAndClose`  would need to change for the same reason.

benaadams:
Tangentially... you could have two memory pool types, a regular (non-concurrent) stack for libuv reads with no contention (concurrent collection for user code block-chain returns?) With current concurrent one for writes.

Though don't think a lot of time is spent in memory pool; ymmv.

halter73:
@benaadams This is something that we/ve discussed, but we would want to measure before making a change. There's also some transport related refactoring that we would like to do first.

benaadams:
@halter73 just throwing it out there :wink:

halter73:
@benaadams Appreciate it. Just want to make sure you don't waste your time on a PR we probably won't merge quickly :wink:

benaadams:
@halter73 you went and shipped; I'm all (mostly) upstream now, have to ship too! Will still do bugs/blockers I come across though.

Also I've seen the wonder of @davidfowl's new async, waiting to see how that plays out...

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Infrastructure/KestrelThread.cs
davidfowl: I'll go further here and make this private and expose a method.

davidfowl:
It's coming! The refactoring is underway!

davidfowl:
Moved this PR to https://github.com/aspnet/KestrelHttpServer/pull/979

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Infrastructure/KestrelThread.cs
davidfowl: Timing out WalkConnectionsAndClose seems fine, not sure about the disposable of UvWriteReq

Issue #975, opened 2016-07-15
Read after connection error can throw two different exception types


cesarbs:
* `IOException`: thrown if `ReadAsync` is awaiting on `SocketInput` when the connection is aborted.
* `TaskCanceledException`: thrown if `ReadAsync` is called by the time the connection has already been aborted after an error.

Need to make this consistent and return only one type of exception.

PR #977, opened 2016-07-16
Always throw same exception type on read after connection error (#975).

cesarbs:
#975 

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/FrameRequestStream.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/FrameRequestStream.cs
@@ -13,6 +13,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Http
     {
         private MessageBody _body;
         private FrameStreamState _state;
+        private Exception _error;
 
         public FrameRequestStream()
         {
@@ -163,13 +164,15 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Http
             _body = null;
         }
 
-        public void Abort()
+        public void Abort(Exception error = null)
         {
             // We don't want to throw an ODE until the app func actually completes.
-            // If the request is aborted, we throw an TaskCanceledException instead.
+            // If the request is aborted, we throw an TaskCanceledException instead,
+            // unless error is not null, in which case we throw it.
             if (_state != FrameStreamState.Closed)
             {
                 _state = FrameStreamState.Aborted;
+                _error = error;
             }
         }
 
@@ -186,7 +189,9 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Http
                 case FrameStreamState.Closed:
                     throw new ObjectDisposedException(nameof(FrameRequestStream));
                 case FrameStreamState.Aborted:
-                    return TaskUtilities.GetCancelledZeroTask();
+                    return _error != null ?
+                        TaskUtilities.GetFaultedTask(_error) :
+                        TaskUtilities.GetCancelledZeroTask();
             }
             return null;
         }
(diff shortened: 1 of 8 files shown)


cesarbs:
Ping.

mikeharder:
Why is it important to always throw the same exception type?

davidfowl:
@mikeharder Makes catch statements less annoying to write.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/SocketInput.cs
```diff
@@ -354,11 +353,7 @@ private void CheckConnectionError()
             var error = _tcs.Task.Exception?.InnerException;
             if (error != null)
             {
-                if (error is TaskCanceledException || error is InvalidOperationException)
```
halter73: Nice side effect. :+1:

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/RequestTests.cs
```diff
@@ -243,26 +241,10 @@ public Task RemoteIPv6Address()
...
-                    {
-                        // We need this here because BadHttpRequestException derives from IOException,
-                        // and we're looking for an actual IOException.
-                    }
                     catch (IOException ex)
                     {
-                        // This is one of two exception types that might thrown in this scenario.
-                        // An IOException is thrown if ReadAsync is awaiting on SocketInput when
-                        // the connection is aborted.
                         expectedExceptionThrown = ex.InnerException is UvException;
```
halter73: Can we check that the `UvException` has the expected error code?

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/FrameRequestStream.cs
```diff
@@ -180,13 +183,17 @@ private Task<int> ValidateState(CancellationToken cancellationToken)
                 case FrameStreamState.Open:
                     if (cancellationToken.IsCancellationRequested)
                     {
-                        return TaskUtilities.GetCancelledZeroTask(cancellationToken);
+                        return _error != null ?
+                            TaskUtilities.GetFaultedTask(_error, cancellationToken) :
+                            TaskUtilities.GetCancelledZeroTask(cancellationToken);
```
halter73: This seems unnecessary since `_error` cannot be set when this `_state` is `FrameStreamState.Open`.

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/RequestTests.cs
cesarbs: Checking `StatusCode` is not optimal because 1) we lose the advantage of not needing platform specific code anymore to determine the correct value of `ECONNRESET` and 2) I'd have to expose internals to the functional tests project.

I'm going to check for `"ECONNRESET"` in the exception message like I did in the other test.

cesarbs:
Updated & rebased.

cesarbs:
Ping.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Infrastructure/TaskUtilities.cs
```diff
@@ -41,5 +42,16 @@ public static Task<int> GetCancelledZeroTask(CancellationToken cancellationToken
             return tcs.Task;
 #endif
         }
+
+        public static Task<int> GetFaultedTask(Exception error, CancellationToken cancellationToken = default(CancellationToken))
```
khellang: I can't see the CancellationToken being used?

cesarbs: Thanks, I had copied the method above it and forgot to remove the unused arg :smile:

davidfowl:
:shipit:

PR #979, opened 2016-07-16
Allocate things per KestrelThread instead of per listener

davidfowl:
- Moved things that have loop affinity to KestrelThread like
WriteReqPool, MemoryPool and the ConnectionManager
- Changed on the listeners to only kill the ListenSocket, not the
connections on dispose
- Moved connection disposal to KestrelThread.Stop
- Simplify the connection manager logic so it's possible to walk and
wait in a single call

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Infrastructure/KestrelThread.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Infrastructure/KestrelThread.cs
@@ -7,6 +7,7 @@ using System.Runtime.ExceptionServices;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Hosting;
+using Microsoft.AspNetCore.Server.Kestrel.Internal.Http;
 using Microsoft.AspNetCore.Server.Kestrel.Internal.Infrastructure;
 using Microsoft.AspNetCore.Server.Kestrel.Internal.Networking;
 using Microsoft.Extensions.Logging;
@@ -60,9 +61,18 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal
 #endif
             QueueCloseHandle = PostCloseHandle;
             QueueCloseAsyncHandle = EnqueueCloseHandle;
+            Memory = new MemoryPool();
+            WriteReqPool = new WriteReqPool(this, _log);
+            ConnectionManager = new ConnectionManager(this, _threadPool);
         }
 
         public UvLoopHandle Loop { get { return _loop; } }
+
+        public MemoryPool Memory { get; }
+
+        public ConnectionManager ConnectionManager { get; }
+
+        public WriteReqPool WriteReqPool { get; }
 
         public ExceptionDispatchInfo FatalError { get { return _closeError; } }
 
@@ -77,12 +87,6 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal
             return tcs.Task;
         }
 
-        // This must be called from the libuv event loop.
-        public void AllowStop()
-        {
-            _post.Unreference();
-        }
-
         public void Stop(TimeSpan timeout)
         {
             lock (_startSync)
@@ -95,6 +99,10 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal
 
             if (_thread.IsAlive)
             {
+                // These operations need to run on the libuv thread so it only makes
+                // sense to attept execution if it's still running
+                DisposeConnections(timeout);
+
                 var stepTimeout = (int)(timeout.TotalMilliseconds / 2);
                 try
                 {
@@ -122,6 +130,34 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal
             if (_closeError != null)
             {
                 _closeError.Throw();
+            }
+        }
+
+        private void DisposeConnections(TimeSpan timeout)
+        {
+            try
+            {
+                // Close and wait for all connections
+                if (!ConnectionManager.WalkConnectionsAndClose(timeout))
+                {
+                    _log.LogError(0, null, "Waiting for connections timed out");
+                }
+
+                var result = PostAsync(state =>
+                {
+                    var listener = (KestrelThread)state;
+                    listener.WriteReqPool.Dispose();
+                },
+                this).Wait(timeout);
+
+                if (!result)
+                {
+                    _log.LogError(0, null, "Disposing write requests failed");
+                }
+            }
+            finally
+            {
+                Memory.Dispose();
             }
         }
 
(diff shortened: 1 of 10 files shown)


davidfowl:
Opened a new PR /cc @halter73 @cesarbs

cesarbs:
What's the difference to the previous PR?

davidfowl:
I had to rebase since it was based on top of my other pr. I also added the a new type WriteReqPool and used the timeout passed into stop for waiting on connections to close.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Infrastructure/KestrelThread.cs
```diff
@@ -85,6 +95,22 @@ public void AllowStop()
...
+            // Close and wait for all connections
+            if (!ConnectionManager.WalkConnectionsAndClose(timeout))
+            {
+                _log.LogError(0, null, "Waiting for connections timed out");
+            }
+
+            PostAsync(state =>
+            {
+                var listener = (KestrelThread)state;
+                var writeReqPool = listener.WriteReqPool;
```
cesarbs: Why not just `listener.WriteReqPool.Dispose()`?

Also nit: remove blank line at the end of the block.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Infrastructure/WriteRequestPool.cs
```diff
@@ -0,0 +1,73 @@
+﻿using System;
+using System.Collections.Generic;
+using Microsoft.AspNetCore.Server.Kestrel.Internal.Networking;
+
+namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Infrastructure
+{
+    public class WriteRequestPool
```
cesarbs: Since we use `WriteReq`* everywhere else, rename this to `WriteReqPool`.

davidfowl:
🆙 📅

davidfowl:
> fail: Block being garbage collected instead of returned to pool
fail:    at System.Environment.GetStackTrace(Exception e, Boolean needFileInfo)
fail:    at System.Environment.get_StackTrace()
fail:    at System.Diagnostics.Debug.Assert(Boolean condition, String message, String detailMessage)
fail:    at System.Diagnostics.Debug.Assert(Boolean condition, String message)
fail:    at Microsoft.AspNetCore.Server.Kestrel.Internal.Infrastructure.MemoryPoolBlock.Finalize() in 

errrr

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Infrastructure/KestrelThread.cs
```diff
@@ -85,6 +95,8 @@ public void AllowStop()
 
         public void Stop(TimeSpan timeout)
         {
+            DisposeConnections(timeout);
```
halter73: I would delete `KestrelThread.AllowStop()` (nothing calls it) now that this needs to be called for gracefull shutdown, and this requires the thread to be running.

halter73: This should probably be after `_thread.IsAlive`. I would expect both the `WalkConnectionsAndClose` and `PostAsync` in `DisposeConnections` to hang if the thread is dead.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Infrastructure/KestrelThread.cs
```diff
@@ -125,6 +137,29 @@ public void Stop(TimeSpan timeout)
...
+                {
+                    _log.LogError(0, null, "Waiting for connections timed out");
+                }
+
+                PostAsync(state =>
+                {
+                    var listener = (KestrelThread)state;
+                    listener.WriteReqPool.Dispose();
+                },
+                this).Wait();
```
halter73: This will cause `KestrelThread.Stop` to hang if the thread is unresponsive. This should never happen if the thread is still alive, but it's still possible if something else is blocking the event loop.

It might be OK to block the caller to `Stop()` in this case, but it's a change in behavior. Before, only an error would be logged.

davidfowl:
🆙 📅

davidfowl:
Ping @cesarbs @halter73

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Infrastructure/KestrelThread.cs
```diff
@@ -95,6 +99,10 @@ public void Stop(TimeSpan timeout)
 
             if (_thread.IsAlive)
             {
+                // These operations need to run on the libuv thread so it only makes
+                // sense to attept execution if it's still running
```
cesarbs: nit: attempt

cesarbs:
LGTM.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Infrastructure/KestrelThread.cs
```diff
@@ -77,12 +87,6 @@ public Task StartAsync()
             return tcs.Task;
         }
 
-        // This must be called from the libuv event loop.
-        public void AllowStop()
```
halter73: If you want to be nice, you can go back to keeping this, since #997 will call it again.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Infrastructure/KestrelThread.cs
```diff
@@ -95,6 +99,10 @@ public void Stop(TimeSpan timeout)
 
             if (_thread.IsAlive)
             {
+                // These operations need to run on the libuv thread so it only makes
+                // sense to attept execution if it's still running
+                DisposeConnections(timeout);
```
halter73: `timeout` should be `KestrelOptions.ShutdownTimeout` which is "The amount of time after the server begins shutting down before connections will be forcefully closed. Kestrel will wait for the duration of the timeout for any ongoing request processing to complete before terminating the connection. No new connections or requests will be accepted during this time. Defaults to 5 seconds."

`DisposeListeners` should should now probably use a shorter timout (Maybe 1 second?)

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Infrastructure/KestrelThread.cs
```diff
@@ -125,6 +133,34 @@ public void Stop(TimeSpan timeout)
...
         }
 
+        private void DisposeConnections(TimeSpan timeout)
+        {
+            try
+            {
+                // Close and wait for all connections
+                if (!ConnectionManager.WalkConnectionsAndClose(timeout))
+                {
+                    _log.LogError(0, null, "Waiting for connections timed out");
```
halter73: `_log.NotAllConnectionsClosedGracefully()` There's a risk where this could get logged multiple times where it wouldn't before, but I don't think that's too big of an issue.

davidfowl: Yea, I was going to log once but hopefully nobody ever sees 😄

Issue #978, opened 2016-07-16
RequestTests.ConnectionResetAbortsRequest() sometimes fails on Travis/OSX


mikeharder:
RequestTests.ConnectionResetAbortsRequest() failed on a Travis/OSX run with the following exception.  Perhaps the timeout needs to be longer than 2500ms?

```
info:       Assert.True() Failure
info:       Expected: True
info:       Actual:   False
fail:     Microsoft.AspNetCore.Server.Kestrel.FunctionalTests.RequestTests.ConnectionResetAbortsRequest [FAIL]
info:       Stack Trace:
info:         /Users/travis/build/aspnet/KestrelHttpServer/test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/RequestTests.cs(218,0): at Microsoft.AspNetCore.Server.Kestrel.FunctionalTests.RequestTests.<ConnectionResetAbortsRequest>d__5.MoveNext()
info:         --- End of stack trace from previous location where exception was thrown ---
info:            at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
info:            at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
info:         --- End of stack trace from previous location where exception was thrown ---
info:            at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
info:            at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
info:         --- End of stack trace from previous location where exception was thrown ---
info:            at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
info:            at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
```

PR #983, opened 2016-07-17
Wait for connection in RequestTests.ConnectionResetAbortsRequest (#978).

cesarbs:
The problem was that the client socket was being disposed even before the server would "see" the connection.

#978 

cc @mikeharder

### DIFF
--- a/test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/RequestTests.cs
+++ b/test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/RequestTests.cs
@@ -5,7 +5,6 @@ using System;
 using System.Collections.Generic;
 using System.Globalization;
 using System.IO;
-using System.Linq;
 using System.Net;
 using System.Net.Http;
 using System.Net.Sockets;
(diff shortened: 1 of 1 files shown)


cesarbs:
Ping.

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/RequestTests.cs
```diff
@@ -211,11 +213,14 @@ public Task RemoteIPv6Address()
                 using (var socket = new Socket(AddressFamily.InterNetwork, SocketType.Stream, ProtocolType.Tcp))
                 {
                     socket.Connect(new IPEndPoint(IPAddress.Loopback, host.GetPort()));
+                    // Wait until connection is established
+                    Assert.True(await connectionStarted.WaitAsync(semaphoreWaitTimeout));
```
cesarbs: This is what fixes the test.

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/RequestTests.cs
```diff
@@ -27,6 +25,8 @@ namespace Microsoft.AspNetCore.Server.Kestrel.FunctionalTests
 {
     public class RequestTests
     {
+        const int semaphoreWaitTimeout = 2500;
```
mikeharder: Nit: `private const int`.  I would also prefix the name with `_`, but I'm not sure if this is a project convention or just my personal convention.

mikeharder:
:shipit: after reviewing feedback

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/RequestTests.cs
halter73: I'd also go with underscore. We aren't the most consistent with this unfortunately, but that ssems to be the most popular for `private const`.

halter73:
:shipit:

PR #985, opened 2016-07-18
[Discuss] Jitter offset to reduce cpu cache set collisions

benaadams:
Don't want every block offset to start on a 4096 boundary or the blocks will continuously evict each other in cache when used (strong line on 4096 below): 

4096 boundary cache
(from http://igoro.com/archive/gallery-of-processor-cache-effects/)

This jitters the start offset by a word/`sizeof(int)`/4 bytes using the 64 block unused section to give 16 different cache set values rather than 1 and reduce cache collisions between blocks.

As the blocks are both full pages and page aligned the cpu won't prefetch the next block so the 64 buffer can be used for this purpose.

/cc @davidfowl @halter73 @cesarbs

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Infrastructure/MemoryPool.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Infrastructure/MemoryPool.cs
@@ -15,8 +15,17 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Infrastructure
         private const int _blockStride = 4096;
 
         /// <summary>
-        /// The last 64 bytes of a block are unused to prevent CPU from pre-fetching the next 64 byte into it's memory cache. 
+        /// If the block size was not an exact page and page aligned we would want to leave the last 64 bytes 
+        /// of a block unused to prevent CPU from pre-fetching the next 64 bytes of the next block into it's 
+        /// memory cache and potentially causing cache false sharing. 
         /// See https://github.com/aspnet/KestrelHttpServer/issues/117 and https://www.youtube.com/watch?v=L7zSU9HI-6I
+        /// 
+        /// However, cpu will only pre-fetch within a page, so this doesn't apply, but since the blocks *are* 
+        /// all page aligned a secondary cache effect kicks where the 4096 alignment will cause each block to 
+        /// evict each other as they will share the same cache slot. 
+        /// 
+        /// To prevent this we still use a 64 byte buffer and then jitter the start offset of each block by
+        /// word multiples (4 bytes) to maintain memory alignment and avoid cache collisions.
         /// </summary>
         private const int _blockUnused = 64;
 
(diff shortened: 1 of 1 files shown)


benaadams:
MacOS errors: Cannot bind to port 80 on the host. Cannot bind to port 443 on the host.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Infrastructure/MemoryPool.cs
```diff
@@ -15,8 +15,17 @@ public class MemoryPool : IDisposable
         private const int _blockStride = 4096;
 
         /// <summary>
-        /// The last 64 bytes of a block are unused to prevent CPU from pre-fetching the next 64 byte into it's memory cache. 
+        /// If the block size was not an exact page and page aligned we would want to leave the last 64 bytes 
+        /// of a block unused to prevent CPU from pre-fetching the next 64 bytes of the next block into it's 
+        /// memory cache and paotentially causing cache false sharing.
```
davidfowl: typo: paotentially

halter73:
Very interesting. I just finished reading the article.

I don't think this makes sense for kestrel since we would likely evict the entire L2 cache before any given block gets reused even with optimal caching. From the blog post:

> `3.` Why the blue triangle in upper left? In the triangle area, we cannot hold all necessary data in cache simultaneously … not due to the associativity, but simply because of the L2 cache size limit.
For example, consider the array length 16MB with step 128. We are repeatedly updating every 128th byte in the array, which means that we touch every other 64-byte memory chunk. To store every other cache line of a 16MB array, we’d need 8MB cache. But, my machine only has 4MB of cache.
Even if the 4MB cache on my machine was fully associative, it still wouldn’t be able to hold 8MB of data.
> `4.` Why does the triangle fade out in the left? Notice that the gradient goes from 0 to 64 bytes – one cache line! As explained in examples 1 and 2, additional accesses to same cache line are nearly free. For example, when stepping by 16 bytes, it will take 4 steps to get to the next cache line. So, we get four memory accesses for the price of one.
Since the number of steps is the same for all cases, a cheaper step results in a shorter running time.

I think the first part (`3.`) is the most relevant. Since the "step size" for Kestrel will almost always be 1, let alone 128, it seems very unlikely that a block would be reused before the entire L2 cache is evicted even if optimally used. We can try measuring it though. If the plaintext benchmark doesn't show an improvement, I doubt we would ever see one.

cesarbs:
No significant difference in benchmarks, unfortunately.

dev:

    [12:25:46.770] Starting scenario Plaintext on benchmark server...
    [12:26:02.212] Warmup
    [12:26:02.212] Starting scenario Plaintext on benchmark client...
    [12:26:13.371] Scenario Plaintext completed on benchmark client
    [12:26:13.371] RPS: 1089361.22
    [12:26:13.371] Stopping scenario Plaintext on benchmark client...
    [12:26:13.371] Benchmark
    [12:26:13.371] Starting scenario Plaintext on benchmark client...
    [12:26:24.505] Scenario Plaintext completed on benchmark client
    [12:26:24.505] RPS: 1113479.41
    [12:26:24.505] Stopping scenario Plaintext on benchmark client...
    [12:26:24.505] Stopping scenario Plaintext on benchmark server...

This PR:

    [12:26:38.566] Starting scenario Plaintext on benchmark server...
    [12:27:00.092] Warmup
    [12:27:00.092] Starting scenario Plaintext on benchmark client...
    [12:27:11.229] Scenario Plaintext completed on benchmark client
    [12:27:11.229] RPS: 1083539.05
    [12:27:11.229] Stopping scenario Plaintext on benchmark client...
    [12:27:11.229] Benchmark
    [12:27:11.229] Starting scenario Plaintext on benchmark client...
    [12:27:22.389] Scenario Plaintext completed on benchmark client
    [12:27:22.389] RPS: 1122281.04
    [12:27:22.390] Stopping scenario Plaintext on benchmark client...
    [12:27:22.392] Stopping scenario Plaintext on benchmark server...

halter73:
I think we should close this then.

benaadams:
I think the pool would need to move to a stack to reap any benefits, but ConcurrentStack is very alloc heavy and also would have high contention on it (which is why ConcurrentQueue).

However... if for example the pool was per thread (e.g. https://github.com/aspnet/KestrelHttpServer/pull/979) and the libuv read allocations could be single threaded and a generic Stack used, when it might repeat benefits. (Returns would need to be done by libuv so ConcurrentQueue might be needed for the handover)

benaadams:
> I think the first part (3.) is the most relevant. Since the "step size" for Kestrel will almost always be 1, let alone 128, it seems very unlikely that a block would be reused before the entire L2 cache is evicted even if optimally used. 

Issue is every block aliases to the cache location so another connection on same core will conflict as the cache is still in use as its advances slowly. However, as cache is 16-way associative would likely need 16 conflicts to cause issue - so high connection counts. Close for now, then?

benaadams:
Ok, measured CPU - no real caching issues (Backend-bound underlined in red below) - so closing

kestrel-cpu-latency

benaadams:
Backend-bound-drilldown

PR #990, opened 2016-07-19
Removed unneeded things from Connection

davidfowl:
- Removed re-assignment of SocketInput and SocketOutput if the connection filter is ineffective
- Removed CreateFrame()
- Clean up the code in the readInputTask continuation


cesarbs:
:ship:

PR #993, opened 2016-07-19
Set StatusCode before disposing HttpContext (#876)

mikeharder:
Recommend ignoring whitespace when viewing the diff: https://github.com/aspnet/KestrelHttpServer/pull/993/files?w=0

Addresses #876 

@halter73, @cesarbs, @JunTaoLuo

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/FrameOfT.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/FrameOfT.cs
@@ -92,55 +92,69 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Http
                         var context = _application.CreateContext(this);
                         try
                         {
-                            await _application.ProcessRequestAsync(context).ConfigureAwait(false);
-                            VerifyResponseContentLength();
-                        }
-                        catch (Exception ex)
-                        {
-                            ReportApplicationError(ex);
+                            try
+                            {
+                                await _application.ProcessRequestAsync(context).ConfigureAwait(false);
+                                VerifyResponseContentLength();
+                            }
+                            catch (Exception ex)
+                            {
+                                ReportApplicationError(ex);
+                            }
+                            finally
+                            {
+                                // Trigger OnStarting if it hasn't been called yet and the app hasn't
+                                // already failed. If an OnStarting callback throws we can go through
+                                // our normal error handling in ProduceEnd.
+                                // https://github.com/aspnet/KestrelHttpServer/issues/43
+                                if (!HasResponseStarted && _applicationException == null && _onStarting != null)
+                                {
+                                    await FireOnStarting();
+                                }
+
+                                PauseStreams();
+
+                                if (_onCompleted != null)
+                                {
+                                    await FireOnCompleted();
+                                }
+
+                                // If _requestAbort is set, the connection has already been closed.
+                                if (Volatile.Read(ref _requestAborted) == 0)
+                                {
+                                    ResumeStreams();
+
+                                    if (_keepAlive)
+                                    {
+                                        // Finish reading the request body in case the app did not.
+                                        await messageBody.Consume();
+                                    }
+
+                                    // ProduceEnd() must be called before _application.DisposeContext(), to ensure
+                                    // HttpContext.Response.StatusCode is correctly set when
+                                    // IHttpContextFactory.Dispose(HttpContext) is called.
+                                    await ProduceEnd();
+                                }
+                                else if (!HasResponseStarted)
+                                {
+                                    // If the request was aborted and no response was sent, there's no
+                                    // meaningful status code to log.
+                                    StatusCode = 0;
+                                }
+                            }
                         }
                         finally
                         {
-                            // Trigger OnStarting if it hasn't been called yet and the app hasn't
-                            // already failed. If an OnStarting callback throws we can go through
-                            // our normal error handling in ProduceEnd.
-                            // https://github.com/aspnet/KestrelHttpServer/issues/43
-                            if (!HasResponseStarted && _applicationException == null && _onStarting != null)
-                            {
-                                await FireOnStarting();
-                            }
-
-                            PauseStreams();
-
-                            if (_onCompleted != null)
-                            {
-                                await FireOnCompleted();
-                            }
-
                             _application.DisposeContext(context, _applicationException);
                         }
+                    }
 
-                        // If _requestAbort is set, the connection has already been closed.
-                        if (Volatile.Read(ref _requestAborted) == 0)
-                        {
-                            ResumeStreams();
+                    StopStreams();
 
-                            if (_keepAlive)
-                            {
-                                // Finish reading the request body in case the app did not.
-                                await messageBody.Consume();
-                            }
-
-                            await ProduceEnd();
-                        }
-
-                        StopStreams();
-
-                        if (!_keepAlive)
-                        {
-                            // End the connection for non keep alive as data incoming may have been thrown off
-                            return;
-                        }
+                    if (!_keepAlive)
+                    {
+                        // End the connection for non keep alive as data incoming may have been thrown off
+                        return;
                     }
 
                     // Don't reset frame state if we're exiting the loop. This avoids losing request rejection
--- a/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/FrameOfT.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/FrameOfT.cs
@@ -100,6 +100,11 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Http
                             catch (Exception ex)
                             {
                                 ReportApplicationError(ex);
+
+                                if (ex is BadHttpRequestException)
+                                {
+                                    throw;
+                                }
                             }
                             finally
                             {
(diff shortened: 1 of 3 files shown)


src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/FrameOfT.cs
```diff
@@ -113,21 +113,24 @@ public class Frame<TContext> : Frame
...
+                                }
+
+                                // ProduceEnd() must be called before _application.DisposeContext(), to ensure
+                                // HttpContext.Response.StatusCode is correct set when
+                                // IHttpContextFactory.Dispose(HttpContext) is called.
+                                await ProduceEnd();
                             }
 
-                            await ProduceEnd();
+                            _application.DisposeContext(context, _applicationException);
```
halter73: I think this should be called even if `_requestAborted` is true. This can be the case if the app calls `HttpContext.Abort()`. Might want to add a test for that.

It will also probably require another finally block because `messageBody.Consume();` can throw too.

mikeharder: Thanks, I will investigate further.  This was changed recently in https://github.com/aspnet/KestrelHttpServer/commit/3186e1bd72df577274f0e8bd63aa50c511bff4b4.

CC: @cesarbs

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/FrameOfT.cs
```diff
@@ -113,21 +113,24 @@ public class Frame<TContext> : Frame
...
+                                ResumeStreams();
+
+                                if (_keepAlive)
+                                {
+                                    // Finish reading the request body in case the app did not.
+                                    await messageBody.Consume();
+                                }
+
+                                // ProduceEnd() must be called before _application.DisposeContext(), to ensure
+                                // HttpContext.Response.StatusCode is correct set when
```
cesarbs: nit: correctly

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/FrameOfT.cs
mikeharder: Currently, if the app calls `HttpContext.Abort()`, then `HttpContext.Response.StatusCode` will be `200` when it's disposed (which appears in the logs).  I agree this seems wrong.  However, if the app aborts the request then no status code is sent to the client, so the status of the request is technically undefined.

What should `HttpContext.Response.StatusCode` be in this case?  We can set it to `500` which is arguably better than `200`, but something like `null` or `-1` (property is an `int` rather than `HttpStatusCode`) might be more appropriate.

benaadams: 503? Service Unavailable

mikeharder: Perhaps `HttpContext.Response.StatusCode` should just be ignored in this case, and the logging code should check `HttpContext.RequestAborted`, and display something other than the status code (e.g. `"aborted"`).

mikeharder: @benaadams: I think `HttpContext.Response.StatusCode` should remain whatever it was before the request was aborted, which is already working with my current changes.  This seems like a separate issue of whether logging should differentiate between completed and aborted requests.

mikeharder: One edge case currently not handled, is an app which calls `HttpContext.Abort()` and then throws an exception, which should probably result in `HttpContext.Response.StatusCode = 500`.  We can try to call `ProduceEnd()` even if the request has been aborted, which will correct set the status code, but I'm not sure if `ProduceEnd()` assumes the request was not aborted.

benaadams: Not changing it to something related to it being aborted might confuse automated log parsing?

cesarbs:
Check the failures on Travis - you might need to update those tests that failed.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/FrameOfT.cs
```diff
@@ -113,21 +113,25 @@ public class Frame<TContext> : Frame
...
-                            if (_keepAlive)
-                            {
-                                // Finish reading the request body in case the app did not.
-                                await messageBody.Consume();
                             }
 
+                            // ProduceEnd() must be called before _application.DisposeContext(), to ensure
+                            // HttpContext.Response.StatusCode is correctly set when
+                            // IHttpContextFactory.Dispose(HttpContext) is called.
                             await ProduceEnd();
```
halter73: `EngineTests.NoResponseSentWhenConnectionIsClosedByServerBeforeClientFinishesSendingRequest` seems to be failing because `ProduceEnd` was moved outside the if block.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/FrameOfT.cs
```diff
@@ -113,21 +113,25 @@ public class Frame<TContext> : Frame
...
-                        }
+                            // If _requestAbort is set, the connection has already been closed.
+                            if (Volatile.Read(ref _requestAborted) == 0)
+                            {
+                                ResumeStreams();
 
-                        // If _requestAbort is set, the connection has already been closed.
-                        if (Volatile.Read(ref _requestAborted) == 0)
-                        {
-                            ResumeStreams();
```
halter73: We should be sure to call ` _application.DisposeContext` if `messageBody.Consume()` throws. The best way to test this is to have an app func that doesn't read the request body and then send a request with malformed chunking.

I mentioned this in a previous iteration, but want to make sure it didn't get lost.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/FrameOfT.cs
cesarbs: I'm not sure that's the case. That test calls `Abort()`, which will close the connection, so anything written after that (the error response written by `ProduceEnd()` should not be received by the client.

cesarbs:
Ping.

mikeharder:
:shipit:

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/FrameOfT.cs
```diff
@@ -115,21 +115,29 @@ public class Frame<TContext> : Frame
...
+                                        await messageBody.Consume();
+                                    }
+                                }
+
+                                // ProduceEnd() must be called before _application.DisposeContext(), to ensure
+                                // HttpContext.Response.StatusCode is correctly set when
+                                // IHttpContextFactory.Dispose(HttpContext) is called.
+                                await ProduceEnd();
+                            }
+                            finally
```
halter73: I think we can get rid of the finally above. I've wanted to clean up the control flow of this method for a while. It's not exactly buggy, but i's confusing because it's not immediately clear what we expect can throw.

halter73: To be clear, not this finally, but the one after catch { ReportApplicationError }.

halter73: My suggestion is that we have a new `try` immediately after `_application.CreateContext(this);`. This `try` block would continue all the way until this finally.

Yes. This means you would have `try { try { await _application.ProcessRequestAsync(context).ConfigureAwait(false); } catch { ...`, but at least that's easier to follow.

cesarbs:
Updated. @halter73 I'm setting status code to 0 when the request is aborted and no response has started.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/FrameOfT.cs
```diff
@@ -125,32 +126,43 @@ public class Frame<TContext> : Frame
...
-                        if (!_keepAlive)
+                        finally
                         {
-                            // End the connection for non keep alive as data incoming may have been thrown off
-                            return;
+                            _application.DisposeContext(context, _applicationException);
                         }
                     }
 
+                    StopStreams();
```
halter73: Do you think StopStreams() should go in the finally?

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/FrameOfT.cs
```diff
@@ -125,32 +126,43 @@ public class Frame<TContext> : Frame
...
+                                // ProduceEnd() must be called before _application.DisposeContext(), to ensure
+                                // HttpContext.Response.StatusCode is correctly set when
+                                // IHttpContextFactory.Dispose(HttpContext) is called.
+                                await ProduceEnd();
+                            }
+                            else if (!HasResponseStarted)
+                            {
+                                // If the request was aborted and no response was sent, there's no
+                                // meaningful status code to log.
+                                StatusCode = 0;
```
halter73: How does this interact with TryProduceInvalidRequestResponse()?

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/ResponseTests.cs
```diff
@@ -185,6 +188,123 @@ public class ResponseTests
...
+        public Task ResponseStatusCodeSetBeforeHttpContextDisposedRequestMalformed()
+        {
+            return ResponseStatusCodeSetBeforeHttpContextDispose(
+                context =>
+                {
+                    return TaskCache.CompletedTask;
+                },
+                expectedClientStatusCode: null,
+                expectedServerStatusCode: HttpStatusCode.BadRequest,
+                sendMalformedRequest: true);
```
halter73: We should have a test for when the app actual reads the request body, and the swallows the exception or doesn't.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/FrameOfT.cs
cesarbs: I'm thinking I should move `await TryProduceInvalidRequestResponse();` inside the check for `_requestAborted == 0`.

cesarbs:
@halter73 https://github.com/aspnet/KestrelHttpServer/pull/993/commits/5b03f68efd707cb90fa3630230231adaf51f080a splits request rejection and setting bad request state in `Frame`, as we dicussed yesterday.

cesarbs:
I'll add a test that validates that an app can write it's own 400 response if it chooses to handle a `BadHttpRequestException`.

halter73:
SingleErrorResponseSentWhenAppSwallowsBadRequestException is failed on the last commit on [AppVeyor](https://ci.appveyor.com/project/aspnetci/kestrelhttpserver/build/1.0.3645#L428). It was testing [this commit](https://github.com/aspnet/KestrelHttpServer/commits/3f7738811a8689d08ec96a8eb964886690acfd34) where the TCP loopback fast path code was already removed.

cesarbs:
`TestConnection` is failing to receive data when the FIN handshake isn't completed and the server sends an RST. My guess is that the shutdown callback is called before the FIN handshake is done, and the RST is sent when we dispose the socket.

It's a weird issue though since the response data was sent by the server (can see it in Message Analyzer), but it looks as if you can't read buffered data if the connection was reset. The problem is probably all the way down in winsock - we've seen a similar issue in the past.

cesarbs:
The test failure is not related to this PR. Any more feedback?

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/FrameOfT.cs
```diff
@@ -92,57 +92,80 @@ public class Frame<TContext> : Frame
...
                         }
-
-                        StopStreams();
-
-                        if (!_keepAlive)
+                        catch (BadHttpRequestException ex)
                         {
-                            // End the connection for non keep alive as data incoming may have been thrown off
-                            return;
+                            SetBadRequestState(ex);
```
halter73: Add comment that this is for bad request data in the request body (not start line/headers) and SetBadRequestState can't be called later because _application.DisposeContext logs the status code.

cesarbs: Added comments.

halter73:
:shipit:

Issue #988, opened 2016-07-18
Assert "Block being garbage collected instead of returned to pool" on Travis/Linux


mikeharder:
The following assert was raised on Travis/Linux (https://travis-ci.org/aspnet/KestrelHttpServer/jobs/145638862):

```
Unhandled Exception: System.Diagnostics.Debug+DebugAssertException: Block being garbage collected instead of returned to pool
   at System.Environment.GetStackTrace(Exception e, Boolean needFileInfo)
   at System.Environment.get_StackTrace()
   at System.Diagnostics.Debug.Assert(Boolean condition, String message, String detailMessage)
   at System.Diagnostics.Debug.Assert(Boolean condition, String message)
   at Microsoft.AspNetCore.Server.Kestrel.Internal.Infrastructure.MemoryPoolBlock.Finalize() in /home/travis/build/aspnet/KestrelHttpServer/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Infrastructure/MemoryPoolBlock.cs:line 76
   at System.Diagnostics.Debug.UnixDebugLogger.ShowAssertDialog(String stackTrace, String message, String detailMessage)
   at System.Diagnostics.Debug.Assert(Boolean condition, String message, String detailMessage)
   at System.Diagnostics.Debug.Assert(Boolean condition, String message)
   at Microsoft.AspNetCore.Server.Kestrel.Internal.Infrastructure.MemoryPoolBlock.Finalize()
```

davidfowl:
On dev?

mikeharder:
Yes, on dev.

davidfowl:
The randomness is killing me

benaadams:
Caught one in Kestrel allocation stacktrace

```
Unhandled Exception: System.Diagnostics.Debug+DebugAssertException: Block being garbage collected instead of returned to pool: IncomingStart, I:\Work\GitHub\KestrelHttpServer\src\Microsoft.AspNetCore.Server.Kestrel\Internal\Http\SocketInput.cs, 65
   at System.Environment.GetStackTrace(Exception e, Boolean needFileInfo)
   at System.Environment.get_StackTrace()
[...]

PR #994, opened 2016-07-19
Ensure libuv handles get disposed properly after binding errors

### LINKED ISSUE #972: "ObjectDisposedException: Safe handle has been closed" when running FunctionalTests on OSX
The following exception was thrown once when running FunctionalTests on OSX.  No test failures were reported before the exception:

```
Unhandled Exception: System.ObjectDisposedException: Safe handle has been closed
   at System.Runtime.InteropServices.SafeHandle.DangerousAddRef(Boolean& success)
   at System.StubHelpers.StubHelpers.SafeHandleAddRef(SafeHandle pHandle, Boolean& success)
   at Microsoft.AspNetCore.Server.Kestrel.Internal.Networking.Libuv.NativeMethods.uv_async_send(UvAsyncHandle handle)
   at Microsoft.AspNetCore.Server.Kestrel.Internal.Networking.Libuv.async_send(UvAsyncHandle handle)
   at Microsoft.AspNetCore.Server.Kestrel.Internal.Networking.UvAsyncHandle.Send()
   at Microsoft.AspNetCore.Server.Kestrel.Internal.KestrelThread.PostCloseHandle(Action`1 callback, IntPtr handle)
   at Microsoft.AspNetCore.Server.Kestrel.Internal.Networking.UvHandle.ReleaseHandle()
   at Microsoft.AspNetCore.Server.Kestrel.Internal.Networking.UvStreamHandle.ReleaseHandle()
```

halter73:
This should completely fix #972. I've been running the functional tests for a few hours on Windows and Linux and I haven't seen this issue which I saw about once every 20 runs before.

I added some tracing in `KestrelThread.OnStopRude`, and this also gets rid of most non-post handles getting walked in that method. The only handles that *sometimes* get walked now (maybe once every 5 functional test runs on my machine) seem to be the `UvPipeHandle`s created [here](https://github.com/aspnet/KestrelHttpServer/blob/bdf9f8dd4e27621c05fbfa302d96b719afc18ebd/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/ListenerPrimary.cs#L58) and [here](https://github.com/aspnet/KestrelHttpServer/blob/bdf9f8dd4e27621c05fbfa302d96b719afc18ebd/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/ListenerPrimary.cs#L72) in `ListenerPrimary`. My tracing doesn't indicate there are any timeouts disposing the listeners that would explain this. Fortunately, since these handles stay rooted until the `KestrelServer` is disposed and no longer referenced, you don't get the ODE on the finalizer thread in this case.

In the hours of running the functional tests on my Windows and Linux machines, I haven't yet seen #988. I might try a smaller VM to better emulate the conditions on Travis.

@davidfowl @mikeharder @cesarbs

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/PipeListener.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/PipeListener.cs
@@ -1,7 +1,6 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
-using Microsoft.AspNetCore.Server.Kestrel.Internal.Infrastructure;
 using Microsoft.AspNetCore.Server.Kestrel.Internal.Networking;
 using Microsoft.Extensions.Logging;
 
@@ -22,9 +21,18 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Http
         protected override UvStreamHandle CreateListenSocket()
         {
             var socket = new UvPipeHandle(Log);
-            socket.Init(Thread.Loop, Thread.QueueCloseHandle, false);
-            socket.Bind(ServerAddress.UnixPipePath);
-            socket.Listen(Constants.ListenBacklog, (stream, status, error, state) => ConnectionCallback(stream, status, error, state), this);
+
+            try
+            {
+                socket.Init(Thread.Loop, Thread.QueueCloseHandle, false);
+                socket.Bind(ServerAddress.UnixPipePath);
+            }
+            catch
+            {
+                socket.Dispose();
+                throw;
+            }
+
             return socket;
         }
 
--- a/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/SocketOutput.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/SocketOutput.cs
@@ -583,7 +583,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Http
 
                 _writeReq = Self._writeReqPool.Allocate();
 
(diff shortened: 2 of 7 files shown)


src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/SocketOutput.cs
```diff
@@ -594,7 +594,7 @@ public void DoWriteIfNeeded()
                     _writeReq.Init(Self._thread.Loop);
                 }
 
-                _writeReq.Write(Self._socket, _lockedStart, _lockedEnd, _bufferCount, (_writeReq, status, error, state) =>
+                _writeReq.Write(Self._socket, _lockedStart, _lockedEnd, _bufferCount, (req, status, error, state) =>
```
mikeharder: Is this just code cleanup unrelated to the bug?

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/PipeListener.cs
```diff
@@ -19,13 +19,15 @@ public PipeListener(ServiceContext serviceContext) : base(serviceContext)
         /// <summary>
         /// Creates the socket used to listen for incoming connections
         /// </summary>
-        protected override UvStreamHandle CreateListenSocket()
+        protected override void CreateListenSocket()
```
mikeharder: Did you consider the following?
```
try
{
    socket.Init(...);
    socket.Bind(...);
    socket.Listen(...);
}
catch
{
    socket.Dispose();
    throw;
}
return socket;
```
I think this would be just as safe, and a smaller change overall.

halter73: Yeah. I think I rejected it too hastily because I wanted everything to be disposed as normal in the `DisposeAsync` path. Not having `ListenSocket` prevents most of `DisposeAsync` from running, but that should be fine. I'll change it.

PR #997, opened 2016-07-20
Wait for graceful shutdown now that it usually works

halter73:
- Unreference the async handle to allow the loop to stop without walking first
- Wait before walking open handles to allow earlier uv_close calls to complete

Turns out that the issue I was seeing with `UvPipeHandle`s reference in #994 was caused by `uv_close` not completely before calling `OnStopRude()`. Same thing happened more frequently on linux with `UvTcpHandles` as well. Waiting for the loop to stop naturally fixes the issue.

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Infrastructure/KestrelThread.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Infrastructure/KestrelThread.cs
@@ -30,6 +30,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal
         private readonly KestrelEngine _engine;
         private readonly IApplicationLifetime _appLifetime;
         private readonly Thread _thread;
+        private readonly TaskCompletionSource<object> _threadTcs = new TaskCompletionSource<object>();
         private readonly UvLoopHandle _loop;
         private readonly UvAsyncHandle _post;
         private Queue<Work> _workAdding = new Queue<Work>(1024);
@@ -89,13 +90,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal
             return tcs.Task;
         }
 
-        // This must be called from the libuv event loop.
-        public void AllowStop()
-        {
-            _post.Unreference();
-        }
-
-        public void Stop(TimeSpan timeout)
+        public async Task StopAsync(TimeSpan timeout)
         {
             lock (_startSync)
             {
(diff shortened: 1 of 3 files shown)


src/Microsoft.AspNetCore.Server.Kestrel/Internal/Infrastructure/KestrelThread.cs
```diff
@@ -93,28 +93,33 @@ public void Stop(TimeSpan timeout)
                 }
             }
 
+            var stepTimeout = (int)(timeout.TotalMilliseconds / 3);
```
mikeharder: While we're changing this method, I would improve the timeout handling:

```
private static TimeSpan RemainingTimeout(TimeSpan totalTimeout, TimeSpan elapsed)
{
    var remainingTimeout = totalTimeout - elapsed;
    return (remainingTimeout > TimeSpan.Zero) ? remainingTimeout : TimeSpan.Zero;
}

{
    var sw = Stopwatch.StartNew();
    ...
    if (!_thread.Join(RemainingTimeout(timeout, sw.Elapsed)))
```

mikeharder:
:shipit: after reviewing my comment about improving the timeout logic.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Infrastructure/KestrelThread.cs
mikeharder: Actually, this probably isn't what we want, since the whole timeout could be exhausted by the first call to `Join()`.  I would probably just pass the `timeout` parameter directly to each call to `Join()`, and reduce the timeout passed by the caller if you want to lower the total maximum time.

davidfowl:
This is going to mess with my change so I wish push before so mine change will mess with yours 😈

benaadams:
It currently shuts the threads down one by one waiting after each; can that be changed so it signals all of them; then waits?

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Infrastructure/KestrelThread.cs
```diff
@@ -93,28 +93,33 @@ public void Stop(TimeSpan timeout)
...
-                catch (ObjectDisposedException)
-                {
-                    // REVIEW: Should we log something here?
-                    // Until we rework this logic, ODEs are bound to happen sometimes.
-                    if (!_thread.Join(stepTimeout))
+                    catch (ObjectDisposedException)
                     {
-                        _log.LogError(0, null, "KestrelThread.Stop failed to terminate libuv thread.");
+                        // REVIEW: Should we log something here?
+                        // Until we rework this logic, ODEs are bound to happen sometimes.
```
cesarbs: Is the comment still relevant?

cesarbs:
:ship:

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Infrastructure/KestrelThread.cs
halter73: Yes

halter73:
@davidfowl This has been updated for parallel/async shutdown.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Infrastructure/KestrelThread.cs
```diff
@@ -141,22 +141,22 @@ public void Stop(TimeSpan timeout)
...
+        private async Task DisposeConnectionsAsync()
         {
             try
             {
                 // Close and wait for all connections
-                if (!ConnectionManager.WalkConnectionsAndClose(_shutdownTimeout))
+                if (!await ConnectionManager.WalkConnectionsAndCloseAsync(_shutdownTimeout).ConfigureAwait(false))
                 {
-                    _log.LogError(0, null, "Waiting for connections timed out");
+                    _log.NotAllConnectionsClosedGracefully();
```
halter73: @davidfowl Why didn't you change this before merging?

davidfowl: Change what

PR #1003, opened 2016-07-21
Throw if SocketInput is used after dispose

davidfowl:


src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/SocketInput.cs
```diff
@@ -72,6 +74,8 @@ public void IncomingData(byte[] buffer, int offset, int count)
         {
             lock (_sync)
             {
+                CheckDisposed();
```
halter73: If we see any failures I would expect to see it here. While `IncomingStart/Complete` shouldn't ever be called after the socket is closed, there's no telling what a ConnectionFilter might do.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/SocketInput.cs
```diff
@@ -316,15 +324,26 @@ public void Dispose()
...
                 }
-                _disposed = true;
+            }
+        }
+
+        private void CheckDisposed()
+        {
+            if (_disposed)
+            {
+                throw new ObjectDisposedException(GetType().Name);
```
halter73: `#if DEBUG`? While this shouldn't throw (but might in the case of ConnectionFilters), we probably shouldn't crash the entire server. If we should, we might as well put a `Trace.Assert` in the `MemoryPoolBlock` finalizer.

At least in the case of `ConnectionFilter`s, `IncommingData` will likely be called inside a `LoggingThreadPool` try/catch.

davidfowl: CRASH all the THINGS!

halter73: Just don't merge it lol

davidfowl:
Doing this in debug mode seems fine

PR #1005, opened 2016-07-21
Do not allocate large write task queues by default

pakrym:
Does not change benchmark performance but significantly decreases allocations per connection.


pakrym:
cc @cesarbs @halter73

cesarbs:
:ship:

davidfowl:
:shipit:

PR #1006, opened 2016-07-21
Known strings fast compare

benaadams:
Use allows `ReferenceEquals` rather than string compare (without assuming anything about interning)
Also use http version enum in `MessageBody.For` rather than http version string


cesarbs:
@benaadams Why did you close it?

benaadams:
Not sure about the known strings; it does intern them - and don't have metrics to prove anything.

Opened the enum as https://github.com/aspnet/KestrelHttpServer/pull/1010

benaadams:
Not sure about the known strings; it does intern them - so its only going to be slow for HTTP/1.0 as that's going to compare 7 chars before deciding its not 1.1

Opened the enum as https://github.com/aspnet/KestrelHttpServer/pull/1010

benaadams:
Added intern checks in https://github.com/aspnet/KestrelHttpServer/pull/1015 as string compare does an ReferenceEquals as first check

PR #1007, opened 2016-07-21
Fast-path header dictionaries

benaadams:
Don't allocate a key string (which then isn't used) for known keys. (Cuts out 42% of string allocations)

Before
unnecessary-allocations
After
removed-unnecessary-allocations
String allocs are now only x4 higher than other things
total-allocations

Currently the header dictionary is randomly? ordered. This moves things around so some items are in a particular order:
* Host first in FrameRequestHeaders' Append/AppendValueFast/SetValueFast 
* Content-Length, Content-Type first in FrameResponseHeaders' AppendValueFast/SetValueFast 

Enumerator and CopyToFast in FrameResponseHeaders bail if nothing after: Connection, Date, Server, Content-Length, Transfer-Encoding (for MVC), Content-Type headers - rather than running through the full list of headers.

Full order probably needs work on the rest with frequency analysis.

@nathana1 @halter73 @cesarbs how does this compare to #829 ?

ReponseCookiesFeature and FormFeature moved to delayed allocation in https://github.com/aspnet/HttpAbstractions/pull/675

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/FrameHeaders.Generated.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/FrameHeaders.Generated.cs
@@ -4197,7 +4197,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Http
         
         public unsafe void Append(byte[] keyBytes, int keyOffset, int keyLength, string value)
         {
-            var key = new string('\0', keyLength);
+            string key;
             fixed (byte* ptr = &keyBytes[keyOffset])
             {
                 var pUB = ptr;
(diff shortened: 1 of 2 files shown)


halter73:
The first commit seems like an obvious win. Nice catch.

The rest is a little complicated/questionable without benchmarks for the reordering changes specifically showing a nice improvement. I think we're just going to merge the first commit and close the rest.

benaadams:
> The first commit seems like an obvious win. 

3rd commit :)

> The rest is a little complicated/questionable without benchmarks for the reordering changes specifically showing a nice improvement.

Its basically to get the wins of cutting down the header dictionary in #829.

For input FrameRequestHeaders Host is basically the only header set, so previously it switches to the 3rd case statement runs string compare ignore-case first for Date, then for From, then does Host. Now it switches to first case and runs Host string compare ignore-case then exits.

For output FrameResponseHeaders only Content-Length, Content-Type, Date and Server are set and Transfer-Encoding for MVC; so after outputting those CopyToFast checks whether it can finish then rather than checking the other 29 flags. (I also added Connection so I didn't need to change the order in all the tests)

Its still in a randomish order, just a more beneficial one.

benaadams:
@halter73 reordered the commits so the allocation is the first one; will get some metrics on the others.

benaadams:
I think you might be right; looking a dubious difference for request; will close and eval response.

PR #1008, opened 2016-07-22
Consolidate BadHttpRequestException messages

benaadams:
Easier to review consistent language in exception messages seeing them all together.
Add and update translations/resource look ups if that is done. 

Should work with non-returning functions Jit update: https://github.com/dotnet/coreclr/pull/6103

Not sure if you want to take this?

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
@@ -814,7 +814,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Http
 
                     if (method == null)
                     {
-                        RejectRequest(RequestRejectionReasons.MissingMethod);
+                        RejectRequest(RequestRejectionReason.MissingMethod);
                     }
 
                     // Note: We're not in the fast path any more (GetKnownMethod should have handled any HTTP Method we're aware of)
@@ -823,7 +823,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Http
                     {
                         if (!IsValidTokenChar(method[i]))
                         {
-                            RejectRequest(RequestRejectionReasons.InvalidMethod);
+                            RejectRequest(RequestRejectionReason.InvalidMethod);
                         }
                     }
                 }
@@ -868,7 +868,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Http
 
                 if (pathBegin.Peek() == ' ')
                 {
-                    RejectRequest(RequestRejectionReasons.MissingRequestTarget);
+                    RejectRequest(RequestRejectionReason.MissingRequestTarget);
                 }
 
                 scan.Take();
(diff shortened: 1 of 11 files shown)


src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
```diff
@@ -567,30 +569,28 @@ public Task ProduceStartAndFireOnStarting()
...
-                    "The response has been aborted due to an unhandled application exception.",
-                    _applicationException);
+                ProduceStart(appCompleted: false);
+                return TaskUtilities.CompletedTask;
             }
 
-            ProduceStart(appCompleted: false);
-
+            ThrowResponseAbortedException();
+            // Line can never be reached but compiler doesn't know
```
halter73: Why did you invert all these ifs?

benaadams: Legacy jit testing; inverted back

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
```diff
@@ -1203,9 +1198,28 @@ public bool StatusCanHaveBody(int statusCode)
...
+        private void ThrowResponseAlreadyStartedException(string value)
+        {
+            throw new InvalidOperationException(value + " cannot be set, response had already started.");
+        }
+
+        private void ThrowResponseAbortedException()
+        {
+            throw new ObjectDisposedException(
+                    "The response has been aborted due to an unhandled application exception.",
+                    _applicationException);
```
cesarbs: gigantic nit: there are 8 spaces in these 2 lines - please indent with 4 spaces.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/RequestRejectionReasons.cs
```diff
@@ -0,0 +1,31 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Http
+{
+    public enum RequestRejectionReasons
```
cesarbs: Is there a convention to name enums in plural? Personally I'd prefer `RequestRejectionReason`.

cesarbs:
I like it. Address my (suuuper important :smile: comments) and we can :ship:, unless @halter73 or @mikeharder have any objections?

benaadams:
Added some missing . in the exception messages too

Is there anyway to get the GeneratedCode running as part of a VS build step again? Only runs on command line build.

cesarbs:
:ship:

halter73:
@benaadams I think it was #581, the PR that transitioned from dnx to dotnet, that broke it.

> Putting [code generation] in the makefile is a temporary solution until we have prerestore and postrestore in nuget.

We should take a second look to see if we can call the code generation a script in the main project.json again so it builds with VS.

PR #1009, opened 2016-07-22
Only allocate key string when unknown

benaadams:
Don't allocate a key string (which then isn't used) for known keys. (Cuts out 42% of string allocations - will be lower % with more headers as the value is still allocated)

Before
unnecessary-allocations
After
removed-unnecessary-allocations
String allocs are now only x4 higher than other things
total-allocations

Resolves https://github.com/aspnet/KestrelHttpServer/issues/1011

/cc @halter73


halter73:
LGTM

mikeharder:
:shipit:

cesarbs:
:ship: but let me benchmarks first :grin:

cesarbs:
No visible gain in plaintext but no loss either, but anyways the less allocations the merrier!

I'll merge this.

PR #1010, opened 2016-07-22
Use HttpVersionType enum rather than string compare

benaadams:
Resolves https://github.com/aspnet/KestrelHttpServer/issues/1014

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
@@ -61,7 +61,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Http
         private bool _autoChunk;
         protected Exception _applicationException;
 
-        private HttpVersionType _httpVersion;
+        protected HttpVersion _httpVersion;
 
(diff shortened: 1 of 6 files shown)


src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/HttpVersionType.cs
```diff
@@ -0,0 +1,12 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Http
+{
+    public enum HttpVersionType
```
cesarbs: Can we name it `HttpVersion` or will that conflict with something else?

benaadams: Renamed

benaadams: Conflicts with the attribute though... is prefixed

halter73:
LGTM

cesarbs:
:ship:

PR #1015, opened 2016-07-22
Tests to makes sure known strings are interned

benaadams:
To ensure the fast-path reference equals return from https://github.com/aspnet/KestrelHttpServer/pull/1006 isn't needed

### DIFF
--- a/test/Microsoft.AspNetCore.Server.KestrelTests/MemoryPoolIteratorTests.cs
+++ b/test/Microsoft.AspNetCore.Server.KestrelTests/MemoryPoolIteratorTests.cs
@@ -381,5 +381,58 @@ namespace Microsoft.AspNetCore.Server.KestrelTests
 
             _pool.Return(block);
         }
+
+        [Theory]
+        [InlineData("HTTP/1.0\r", "HTTP/1.0")]
+        [InlineData("HTTP/1.1\r", "HTTP/1.1")]
+        public void KnownVersionsAreInterned(string input, string expected)
+        {
+            TestKnownStringsInterning(input, expected, MemoryPoolIteratorExtensions.GetKnownVersion);
+        }
+
+        [Theory]
+        [InlineData("CONNECT / HTTP/1.1", "CONNECT")]
+        [InlineData("DELETE / HTTP/1.1", "DELETE")]
+        [InlineData("GET / HTTP/1.1", "GET")]
+        [InlineData("HEAD / HTTP/1.1", "HEAD")]
+        [InlineData("PATCH / HTTP/1.1", "PATCH")]
+        [InlineData("POST / HTTP/1.1", "POST")]
+        [InlineData("PUT / HTTP/1.1", "PUT")]
+        [InlineData("OPTIONS / HTTP/1.1", "OPTIONS")]
+        [InlineData("TRACE / HTTP/1.1", "TRACE")]
+        public void KnownMethodsAreInterned(string input, string expected)
+        {
+            TestKnownStringsInterning(input, expected, MemoryPoolIteratorExtensions.GetKnownMethod);
+        }
+
+        private delegate bool GetKnownString(MemoryPoolIterator iter, out string result);
+
+        private void TestKnownStringsInterning(string input, string expected, GetKnownString action)
+        {
+            // Arrange
+            var chars = input.ToCharArray().Select(c => (byte)c).ToArray();
+            var block1 = _pool.Lease();
+            var block2 = _pool.Lease();
+            Buffer.BlockCopy(chars, 0, block1.Array, block1.Start, chars.Length);
+            Buffer.BlockCopy(chars, 0, block2.Array, block2.Start, chars.Length);
+            block1.End += chars.Length;
+            block2.End += chars.Length;
+            var begin1 = block1.GetIterator();
+            var begin2 = block2.GetIterator();
+
+            // Act
+            string knownString1, knownString2;
+            var result1 = action(begin1, out knownString1);
+            var result2 = action(begin2, out knownString2);
+
+            _pool.Return(block1);
+            _pool.Return(block2);
+
+            // Assert
+            Assert.True(result1);
+            Assert.True(result2);
+            Assert.Equal(knownString1, expected);
+            Assert.Same(knownString1, knownString2);
+        }
     }
 }


test/Microsoft.AspNetCore.Server.KestrelTests/MemoryPoolIteratorTests.cs
```diff
@@ -381,5 +381,76 @@ public void GetsKnownVersion(string input, char endChar, bool expectedResult, st
 
             _pool.Return(block);
         }
+
+        [Theory]
+        [InlineData("HTTP/1.0\r", '\r')]
+        [InlineData("HTTP/1.1\r", '\r')]
+        public void KnownVersionsAreInterned(string input, char endChar)
```
mikeharder: Is endChar used anywhere?  I don't see it.

test/Microsoft.AspNetCore.Server.KestrelTests/MemoryPoolIteratorTests.cs
```diff
@@ -381,5 +381,76 @@ public void GetsKnownVersion(string input, char endChar, bool expectedResult, st
...
+            Buffer.BlockCopy(chars, 0, block2.Array, block2.Start, chars.Length);
+            block1.End += chars.Length;
+            block2.End += chars.Length;
+            var begin1 = block1.GetIterator();
+            var begin2 = block2.GetIterator();
+            string knownString1;
+            string knownString2;
+
+            // Act
+            var result1 = begin1.GetKnownVersion(out knownString1);
```
mikeharder: Should this be GetKnownMethod()?

test/Microsoft.AspNetCore.Server.KestrelTests/MemoryPoolIteratorTests.cs
```diff
@@ -381,5 +381,76 @@ public void GetsKnownVersion(string input, char endChar, bool expectedResult, st
...
+        [InlineData("HEAD / HTTP/1.1", ' ')]
+        [InlineData("PATCH / HTTP/1.1", ' ')]
+        [InlineData("POST / HTTP/1.1", ' ')]
+        [InlineData("PUT / HTTP/1.1", ' ')]
+        [InlineData("OPTIONS / HTTP/1.1", ' ')]
+        [InlineData("TRACE / HTTP/1.1", ' ')]
+        public void KnownMethodsAreInterned(string input, char endChar)
+        {
+            // Arrange
+            var chars = input.ToCharArray().Select(c => (byte)c).ToArray();
```
mikeharder: I would create a helper method to avoid duplicate code in these tests.

test/Microsoft.AspNetCore.Server.KestrelTests/MemoryPoolIteratorTests.cs
benaadams: eep null == null will fix

benaadams:
@mikeharder resolved

test/Microsoft.AspNetCore.Server.KestrelTests/MemoryPoolIteratorTests.cs
```diff
@@ -381,5 +381,66 @@ public void GetsKnownVersion(string input, char endChar, bool expectedResult, st
...
+        [Theory]
+        [InlineData("HTTP/1.0\r", "HTTP/1.0")]
+        [InlineData("HTTP/1.1\r", "HTTP/1.1")]
+        public void KnownVersionsAreInterned(string input, string expected)
+        {
+            TestKnownStringsInterning(input, expected,
+                (iter, expect) => {
+                    string output;
+                    iter.GetKnownVersion(out output);
+                    Assert.True(output == expect);
```
cesarbs: `Assert.Same(expect, output)` would be better here since you're testing that they're using the same reference.

test/Microsoft.AspNetCore.Server.KestrelTests/MemoryPoolIteratorTests.cs
```diff
@@ -381,5 +381,66 @@ public void GetsKnownVersion(string input, char endChar, bool expectedResult, st
...
+            var knownString1 = action(begin1, expected);
+            var knownString2 = action(begin2, expected);
+
+            _pool.Return(block1);
+            _pool.Return(block2);
+
+            // Assert
+            Assert.True(knownString1 == knownString2);
+            Assert.Equal(knownString1, knownString2);
+            Assert.True(ReferenceEquals(knownString1, knownString2));
```
cesarbs: `Assert.Same(knownString1, knownString2)`

cesarbs: That actually makes the previous 2 tests unnecessary.

test/Microsoft.AspNetCore.Server.KestrelTests/MemoryPoolIteratorTests.cs
benaadams: Testing different references, but making sure its returning the right thing (not null)

benaadams:
@cesarbs changed to Equals and Same

test/Microsoft.AspNetCore.Server.KestrelTests/MemoryPoolIteratorTests.cs
```diff
@@ -381,5 +381,66 @@ public void GetsKnownVersion(string input, char endChar, bool expectedResult, st
...
             _pool.Return(block);
         }
+
+        [Theory]
+        [InlineData("HTTP/1.0\r", "HTTP/1.0")]
+        [InlineData("HTTP/1.1\r", "HTTP/1.1")]
+        public void KnownVersionsAreInterned(string input, string expected)
+        {
+            TestKnownStringsInterning(input, expected,
+                (iter, expect) => {
```
cesarbs: I think there's a way to make this simpler. You could declare a delegate like

    delegate bool GetKnownString(MemoryPoolIterator iter, out string result);

Then have `TestKnownStringsInterning` take a `GetKnownString` and call it and check against it's `expected` param. That way you don't need this lambda here and IMO the test looks clearer.

To be honest though, we already check for that the output is correct in `GetsKnownMethod` and `GetsKnownVersion`, so this check is actually redundant. You're already making the desired check for interning in `TestKnownStringsInterning`.

benaadams: Added the check as I got the test wrong first time and was comparing null == null for the returns; so its just to ensure the test is testing the right thing.

benaadams: You can use an extension method as a delegate - that's quite funky; I just learnt something new, thank you @cesarbs :+1:

test/Microsoft.AspNetCore.Server.KestrelTests/MemoryPoolIteratorTests.cs
```diff
@@ -381,5 +381,58 @@ public void GetsKnownVersion(string input, char endChar, bool expectedResult, st
...
+        [InlineData("POST / HTTP/1.1", "POST")]
+        [InlineData("PUT / HTTP/1.1", "PUT")]
+        [InlineData("OPTIONS / HTTP/1.1", "OPTIONS")]
+        [InlineData("TRACE / HTTP/1.1", "TRACE")]
+        public void KnownMethodsAreInterned(string input, string expected)
+        {
+            TestKnownStringsInterning(input, expected, MemoryPoolIteratorExtensions.GetKnownMethod);
+        }
+
+        delegate bool GetKnownString(MemoryPoolIterator iter, out string result);
```
cesarbs: make it `private`

cesarbs:
:ship: after last comment.

benaadams:
Privatised it

PR #1022, opened 2016-07-29
Limit request line length (#784)

cesarbs:
#784

In order to address this, I've added new `MemoryPoolIterator.Seek` overloads. One allows setting a limit on how many bytes can be scanned, and returns (via an `out` parameter) the number of bytes scanned; the other one allows passing in another `MemoryPoolIterator` to server as a limit on the seek operation.

This change changes the behavior of `Frame.TakeStartLine` to wait until the final LF has been received before parsing the line. If LF is not found before the line limit, the request is rejected. Because of this change, `BadHttpRequestTests.ServerClosesConnectionAsSoonAsBadRequestLineIsDetected` is not a meaningful test anymore.

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Infrastructure/MemoryPoolIterator.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Infrastructure/MemoryPoolIterator.cs
@@ -217,6 +217,123 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Infrastructure
 
         public unsafe int Seek(ref Vector<byte> byte0Vector)
         {
+            int bytesScanned;
+            return Seek(ref byte0Vector, out bytesScanned);
+        }
+
+        public unsafe int Seek(
+            ref Vector<byte> byte0Vector,
+            out int bytesScanned,
+            int limit = int.MaxValue)
+        {
+            bytesScanned = 0;
+
+            if (IsDefault || limit <= 0)
+            {
+                return -1;
+            }
+
+            var block = _block;
+            var index = _index;
+            var wasLastBlock = block.Next == null;
+            var following = block.End - index;
+            byte[] array;
+            var byte0 = byte0Vector[0];
+
+            while (true)
+            {
+                while (following == 0)
+                {
+                    if (bytesScanned >= limit || wasLastBlock)
+                    {
+                        _block = block;
+                        _index = index;
+                        return -1;
+                    }
+
+                    block = block.Next;
+                    index = block.Start;
+                    wasLastBlock = block.Next == null;
+                    following = block.End - index;
+                }
+                array = block.Array;
+                while (following > 0)
+                {
+                    // Need unit tests to test Vector path
+#if !DEBUG
+                    // Check will be Jitted away https://github.com/dotnet/coreclr/issues/1079
+                    if (Vector.IsHardwareAccelerated)
+                    {
+#endif
+                    if (following >= _vectorSpan)
+                    {
+                        var byte0Equals = Vector.Equals(new Vector<byte>(array, index), byte0Vector);
+
+                        if (byte0Equals.Equals(Vector<byte>.Zero))
+                        {
+                            if (bytesScanned + _vectorSpan >= limit)
+                            {
+                                _block = block;
+                                // Ensure iterator is left at limit position
+                                _index = index + (limit - bytesScanned);
+                                bytesScanned = limit;
+                                return -1;
+                            }
+
+                            bytesScanned += _vectorSpan;
+                            following -= _vectorSpan;
+                            index += _vectorSpan;
+                            continue;
+                        }
+
+                        _block = block;
+
+                        var firstEqualByteIndex = FindFirstEqualByte(ref byte0Equals);
+                        var vectorBytesScanned = firstEqualByteIndex + 1;
+
+                        if (bytesScanned + vectorBytesScanned > limit)
+                        {
+                            // Ensure iterator is left at limit position
+                            _index = index + (limit - bytesScanned);
+                            bytesScanned = limit;
+                            return -1;
+                        }
+
+                        _index = index + firstEqualByteIndex;
+                        bytesScanned += vectorBytesScanned;
+
+                        return byte0;
+                    }
+                    // Need unit tests to test Vector path
+#if !DEBUG
+                    }
+#endif
+
+                    var pCurrent = (block.DataFixedPtr + index);
+                    var pEnd = pCurrent + Math.Min(following, limit - bytesScanned);
+                    do
+                    {
+                        bytesScanned++;
+                        if (*pCurrent == byte0)
+                        {
+                            _block = block;
+                            _index = index;
+                            return byte0;
+                        }
+                        pCurrent++;
+                        index++;
+                    } while (pCurrent < pEnd);
+
+                    following = 0;
+                    break;
+                }
+            }
+        }
+
+        public unsafe int Seek(
+            ref Vector<byte> byte0Vector,
+            ref MemoryPoolIterator limit)
+        {
             if (IsDefault)
             {
                 return -1;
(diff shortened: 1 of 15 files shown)


src/Microsoft.AspNetCore.Server.Kestrel/KestrelServerOptions.cs
```diff
@@ -57,13 +60,35 @@ public class KestrelServerOptions
...
+        /// </remarks>
+        public int MaxRequestLineSize
+        {
+            get
+            {
+                return _maxRequestLineSize;
+            }
+            set
+            {
+                if (value <= 0)
```
Tratcher: What's the shortest practical request line? `GET / HTTP/1.1`? Any reason to allow the limit to be lower than 16?

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/MaxRequestLineSizeTests.cs
```diff
@@ -0,0 +1,66 @@
...
+                    var response = await client.GetAsync($"http://127.0.0.1:{host.GetPort()}");
+                    Assert.Equal(HttpStatusCode.BadRequest, response.StatusCode);
+                }
+            }
+        }
+
+        [Theory]
+        [InlineData(1, 2)]
+        [InlineData(int.MaxValue - 1, int.MaxValue)]
+        public void ServerFailsToStartWhenMaxRequestLineSizeIsLessThanMaxRequestBufferSize(long maxRequestBufferSize, int maxRequestLineSize)
```
Tratcher: Isn't this description backwards?

src/Microsoft.AspNetCore.Server.Kestrel/KestrelServerOptions.cs
halter73: Any rule saying a method has to be at least 3 characters (even if they always are in practice)? You could argue there is no reason for even 16 to be valid. I don't see any reason to be too smart with this limit.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Infrastructure/MemoryPoolIterator.cs
```diff
@@ -277,6 +328,16 @@ public unsafe int Seek(ref Vector<byte> byte0Vector)
                     var pEnd = pCurrent + following;
                     do
                     {
+                        bytesScanned++;
+
+                        if ((bytesScanned > limitBytes) ||
+                            (!limitIterator.IsDefault && block == limitIterator.Block && index > limitIterator.Index))
```
halter73: instead of doing this complicated comparison for every byte, couldn't you adjust `pEnd` to match the limit index?

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Infrastructure/MemoryPoolIterator.cs
```diff
@@ -261,11 +293,30 @@ public unsafe int Seek(ref Vector<byte> byte0Vector)
                             {
                                 following -= _vectorSpan;
                                 index += _vectorSpan;
+                                bytesScanned += _vectorSpan;
+
+                                if ((bytesScanned > limitBytes) ||
+                                    (!limitIterator.IsDefault && block == limitIterator.Block && index > limitIterator.Index))
```
halter73: I think all these `!limitIterator.IsDefault` checks are unnecessary considering `block` will never be null.

src/Microsoft.AspNetCore.Server.Kestrel/KestrelServerOptions.cs
```diff
@@ -57,13 +60,35 @@ public class KestrelServerOptions
...
             }
         }
 
         /// <summary>
+        /// Gets or sets the maximum allowed size for the HTTP request line.
+        /// </summary>
+        /// <remarks>
+        /// Defaults to 8,192 bytes (8 KB).
+        /// </remarks>
+        public int MaxRequestLineSize
```
davidfowl: I think we should move these settings to a sub object so the top level class isn;t polluted.

halter73: "Limits" maybe? For both max lengths and timeouts.

davidfowl: Yep

cesarbs:
Updated:

* Removed unnecessary `IsDefault` checks
* Set `pEnd` based on limit instead of checking for limit on each iteration of the `do while` loops
* Fixed inconsistencies with the iterator state when `Seek()` fails. It's now always at the expected position: one byte past the byte limit in seeks with byte limits, and exactly at the limit iterator in seeks with an iterator limit.

cesarbs:
Also, I removed the private overload for `Seek()` which took both a byte and an iterator limit and tried to handle those together. There's now a separate `Seek()` for each option. I'm thinking of reducing the resulting code duplication by adding a `checkLimit` callback, but I need to make sure that doesn't negatively impact perf.

cesarbs:
Another update: moves `MaxRequestLineSize` to the new class `KestrelServerLimits`, accessible via the `Limits` property of `KestrelServerOptions`.

cesarbs:
Rebased.

cesarbs:
@davidfowl I'm thinking of moving `MaxRequestBufferSize` to the new `KestrelServerLimits` class and marking the one in `KestrelServerOptions` as deprecated, and keeping it as a proxy to the new one so we don't break anything. What do you think?

davidfowl:
We should talk to @Eilon about using `[Obsolete]` to deprecate APIs.

test/Microsoft.AspNetCore.Server.KestrelTests/MemoryPoolIteratorTests.cs
```diff
@@ -406,6 +411,223 @@ public void KnownMethodsAreInterned(string input, string expected)
...
         }
 
+        [Theory]
+        [MemberData(nameof(SeekByteLimitData))]
+        public void TestSeekByteLimitWithinSameBlock(string input, char seek, int limit, int expectedBytesScanned, int expectedReturnValue)
+        {
+            // Arrange
+            var seekVector = new Vector<byte>((byte)seek);
+
+            var block = _pool.Lease();
```
davidfowl: put the lease return in a try finally

Eilon:
@cesarbs you can follow the guidelines here to mark things as obsolete: https://github.com/aspnet/Home/wiki/Engineering-guidelines#breaking-changes

Then log a new bug to remove the old property and @muratg can triage into the 2.0.0 milestone and mark as breaking-change.

test/Microsoft.AspNetCore.Server.KestrelTests/MemoryPoolIteratorTests.cs
cesarbs: Ugh :( Ok.

cesarbs:
@davidfowl Put tests in try/finally.

@Eilon @davidfowl Moved property and marked old one as obsolete. Please review the last commit to check that I did everything right.

muratg:
https://github.com/aspnet/KestrelHttpServer/issues/1027 to track the future removal of it

src/Microsoft.AspNetCore.Server.Kestrel/KestrelServerOptions.cs
```diff
@@ -41,29 +34,37 @@ public class KestrelServerOptions
...
+        /// </para>
+        /// <para>
+        /// Gets or sets the maximum size of the request buffer.
+        /// </para>
         /// </summary>
         /// <remarks>
+        /// When set to null, the size of the request buffer is unlimited.
         /// Defaults to 1,048,576 bytes (1 MB).
         /// </remarks>
+        [Obsolete]
```
Eilon: Please follow the guidelines: Include the obsoletion message here.

cesarbs: https://github.com/aspnet/KestrelHttpServer/pull/1029

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Infrastructure/MemoryPoolIterator.cs
```diff
@@ -259,13 +379,33 @@ public unsafe int Seek(ref Vector<byte> byte0Vector)
...
                                 following -= _vectorSpan;
                                 index += _vectorSpan;
                                 continue;
                             }
 
                             _block = block;
-                            _index = index + FindFirstEqualByte(ref byte0Equals);
+
+                            var firstEqualByteIndex = FindFirstEqualByte(ref byte0Equals);
+                            var vectorBytesScanned = firstEqualByteIndex + 1;
```
benaadams: @cesarbs The vectorBytesScanned variable isn't used here

PR #1025, opened 2016-08-02
Move TestConnection to shared directory

cesarbs:
So we can use it in functional tests. I want to use it in #1022.

@mikeharder @davidfowl @halter73


mikeharder:
:shipit:

PR #1032, opened 2016-08-04
Move server start failure test in MaxRequestLineSizeTests to KestreServerTests

cesarbs:
cc @mikeharder @halter73 @davidfowl


mikeharder:
:shipit:

Issue #475, opened 2015-12-13
Add header limit parameters


benaadams:
Failure state: `431 Request Header Fields Too Large` *(ngnix returns `400 Bad Request`)*

*"The server is unwilling to process the request because either an individual header field, or all the header fields collectively, are too large."* [RFC 6585](https://tools.ietf.org/html/rfc6585)

**http.sys** [default](https://support.microsoft.com/en-us/kb/820129) `MaxRequestBytes` 16384
**ngnix** [default](http://nginx.org/en/docs/http/ngx_http_core_module.html#large_client_header_buffers) `large_client_header_buffers` 4 * 8k = 32768
**apache.tomcat** [default](http://tomcat.apache.org/tomcat-9.0-doc/config/http.html) `maxHttpHeaderSize` 8192 

Will currently be enforced by reverse proxy config (e.g. IIS/nginx) when deployed in recommended configuration - but not in pure Kestrel deployment.

Tratcher:
See https://github.com/aspnet/KestrelHttpServer/issues/479#issuecomment-164325646

benaadams:
Question on whether not having a max url/start length (beyond total header size) would have any upstream issues?

Guessing it wouldn't have a great impact beyond anything else (like lots of headers)

benaadams:
Probably simpler code wise if there was only one param.

Tratcher:
@blowdart ?

blowdart:
Work for me. We used to have this with IIS and UrlScan

muratg:
@blowdart Is this required for RC2/RTM?   Or can we punt it to post RTM?

blowdart:
Yes

DamianEdwards:
post-RC2

cesarbs:
@muratg Moving to 1.0.0 and assigning to myself.

cesarbs:
So do we want just a single parameter to limit request line + header length? No checking request line length individually and individual header lengths?

Ping @Tratcher @halter73 @blowdart @davidfowl

blowdart:
No, we want all the things.

cesarbs:
So, rounding up the specific things we want:

* Request line length
* Request target length (I know @Tratcher would like this to be part of the previous one)
* Request query length? I'd rather have it as part of target length.
* Max single header length
* Max total headers length
* Max header count

What do you guys think?

Tratcher:
There's a lot of overlap and a lot of knobs here. I think we only need:
[...]

PR #1033, opened 2016-08-05
Add header limit options (#475)

cesarbs:
New options in `KestrelServerLimits`:

* `MaxRequestHeadersTotalSize`: defaults to 32 KB.
* `MaxRequestHeaderCount`: defaults to 100.

Applies to normal request headers and trailing headers combined.

#475

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Kestrel/BadHttpRequestException.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/BadHttpRequestException.cs
@@ -49,8 +49,8 @@ namespace Microsoft.AspNetCore.Server.Kestrel
                 case RequestRejectionReason.WhitespaceIsNotAllowedInHeaderName:
                     ex = new BadHttpRequestException("Whitespace is not allowed in header name.");
                     break;
-                case RequestRejectionReason.HeaderLineMustEndInCRLFOnlyCRFound:
-                    ex = new BadHttpRequestException("Header line must end in CRLF; only CR found.");
+                case RequestRejectionReason.HeaderValueMustNotContainCR:
+                    ex = new BadHttpRequestException("Header value must not contain CR characters.");
                     break;
                 case RequestRejectionReason.HeaderValueLineFoldingNotSupported:
                     ex = new BadHttpRequestException("Header value line folding not supported.");
@@ -90,6 +90,15 @@ namespace Microsoft.AspNetCore.Server.Kestrel
                     break;
                 case RequestRejectionReason.MissingCrAfterVersion:
                     ex = new BadHttpRequestException("Missing CR in request line.");
+                    break;
+                case RequestRejectionReason.HeadersExceedMaxTotalSize:
+                    ex = new BadHttpRequestException("Request headers too long.");
+                    break;
+                case RequestRejectionReason.MissingCRInHeaderLine:
+                    ex = new BadHttpRequestException("No CR character found in header line.");
+                    break;
+                case RequestRejectionReason.TooManyHeaders:
+                    ex = new BadHttpRequestException("Request contains too many headers.");
                     break;
                 default:
                     ex = new BadHttpRequestException("Bad request.");
--- a/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
@@ -66,6 +66,9 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Http
 
         private readonly string _pathBase;
 
+        private int _remainingRequestHeadersBytesAllowed;
+        private int _requestHeadersParsed;
+
         public Frame(ConnectionContext context)
             : base(context)
         {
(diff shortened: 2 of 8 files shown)


cesarbs:
@blowdart

src/Microsoft.AspNetCore.Server.Kestrel/KestrelServerLimits.cs
```diff
@@ -14,6 +14,13 @@ public class KestrelServerLimits
         // Matches the default large_client_header_buffers in nginx.
         private int _maxRequestLineSize = 8 * 1024;
 
+        // Matches the default large_client_header_buffers in nginx.
+
```
davidfowl: nit:space

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
```diff
@@ -305,6 +308,9 @@ public void Reset()
 
             _manuallySetRequestAbortToken = null;
             _abortedCts = null;
+
+            _maxRequestHeadersTotalSize = ServerOptions.Limits.MaxRequestHeadersTotalSize;
```
halter73: nit: `_remainingRequestHeadersBytesAllowed` or something like that since it changes as headers come in.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
```diff
@@ -1078,33 +1084,46 @@ public bool TakeMessageHeaders(SocketInput input, FrameRequestHeaders requestHea
...
                     {
                         RejectRequest(RequestRejectionReason.NoColonCharacterFoundInHeaderLine);
                     }
+                    var endName = scan;
+
+                    scan.Take();
 
                     var validateName = beginName;
-                    if (validateName.Seek(ref _vectorSpaces, ref _vectorTabs, ref _vectorColons) != ':')
+                    if (validateName.Seek(ref _vectorSpaces, ref _vectorTabs, ref endName) != -1)
```
halter73: I like how this improves clarity slightly.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
```diff
@@ -1124,30 +1143,21 @@ public bool TakeMessageHeaders(SocketInput input, FrameRequestHeaders requestHea
...
+                    if (ch != '\n')
                     {
-                        RejectRequest(RequestRejectionReason.HeaderLineMustEndInCRLFOnlyCRFound);
+                        RejectRequest(RequestRejectionReason.HeaderLineMustEndInCRLF);
                     }
 
                     var next = scan.Peek();
-                    if (next == -1)
-                    {
-                        return false;
```
halter73: I don't think we reject linear whitespace if the input stalls right beforehand. Probably should add a test.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
```diff
@@ -1124,30 +1143,21 @@ public bool TakeMessageHeaders(SocketInput input, FrameRequestHeaders requestHea
                     }
```
halter73: ch = scan.Peek();
    if (ch == -1)	
    {
        return false;
    }

Can be removed from above, right?

src/Microsoft.AspNetCore.Server.Kestrel/BadHttpRequestException.cs
```diff
@@ -49,8 +49,8 @@ internal static BadHttpRequestException GetException(RequestRejectionReason reas
                 case RequestRejectionReason.WhitespaceIsNotAllowedInHeaderName:
                     ex = new BadHttpRequestException("Whitespace is not allowed in header name.");
                     break;
-                case RequestRejectionReason.HeaderLineMustEndInCRLFOnlyCRFound:
-                    ex = new BadHttpRequestException("Header line must end in CRLF; only CR found.");
+                case RequestRejectionReason.HeaderLineMustEndInCRLF:
+                    ex = new BadHttpRequestException("Header line must end in CRLF.");
```
halter73: Maybe "Header value must not contain a CR"?

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
cesarbs: Yep, just did it.

cesarbs: Actually need to keep the `Peek()`.

cesarbs: I rewrote it a bit differently.

cesarbs:
@halter73 Added test for the linear whitespace situation where the next character is not yet available: https://github.com/aspnet/KestrelHttpServer/pull/1033/files#diff-78adb33626b9b8aa43b7d4529f4f0becR214

halter73:
:shipit:

Issue #1002, opened 2016-07-21
Intermittent CI failure WritesDontCompleteImmediatelyWhenTooManyBytesIncludingNonImmediateAreAlreadyPreCompleted


benaadams:
https://ci.appveyor.com/project/aspnetci/kestrelhttpserver/build/1.0.3225

```
Microsoft.AspNetCore.Server.KestrelTests.SocketOutputTests.WritesDontCompleteImmediatelyWhenTooManyBytesIncludingNonImmediateAreAlreadyPreCompleted [FAIL]
      Assert.True() Failure
      Expected: True
      Actual:   False
      Stack Trace:
        SocketOutputTests.cs(201,0): at Microsoft.AspNetCore.Server.KestrelTests.SocketOutputTests.WritesDontCompleteImmediatelyWhenTooManyBytesIncludingNonImmediateAreAlreadyPreCompleted()
```
More info
```
System.ObjectDisposedException: Cannot access a disposed object. 
fail: Object name: 'WriteReqPool'.
```

benaadams:
Bit more info from a Mono run
```
info:   Starting:    Microsoft.AspNetCore.Server.KestrelTests
fail:     Microsoft.AspNetCore.Server.KestrelTests.SocketOutputTests.WritesDontCompleteImmediatelyWhenTooManyBytesIncludingNonImmediateAreAlreadyPreCompleted [FAIL]
info:       Assert.True() Failure
info:       Expected: True
info:       Actual:   False
info:       Stack Trace:
info:         /home/travis/build/aspnet/KestrelHttpServer/test/Microsoft.AspNetCore.Server.KestrelTests/SocketOutputTests.cs(197,0): at Microsoft.AspNetCore.Server.KestrelTests.SocketOutputTests.WritesDontCompleteImmediatelyWhenTooManyBytesIncludingNonImmediateAreAlreadyPreCompleted()
fail: Unhandled Exception: System.ObjectDisposedException: Cannot access a disposed object.
fail: Object name: 'WriteReqPool'.
fail:    at Microsoft.AspNetCore.Server.Kestrel.Internal.Infrastructure.WriteReqPool.Return(UvWriteReq req)
fail:    at Microsoft.AspNetCore.Server.Kestrel.Internal.Http.SocketOutput.WriteContext.PoolWriteReq(UvWriteReq writeReq)
[...]

PR #1036, opened 2016-08-06
Use Buffer.MemoryCopy when available

benaadams:
MemoryCopy has improved a lot (and is faster than BlockCopy) https://github.com/dotnet/coreclr/pull/6627

And is undergoing further improvements https://github.com/dotnet/coreclr/pull/6638

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Infrastructure/MemoryPoolIterator.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Infrastructure/MemoryPoolIterator.cs
@@ -1015,7 +1015,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Infrastructure
 
         public void CopyFrom(byte[] data, int offset, int count)
         {
-            if (IsDefault)
+            if (count == 0 || IsDefault)
             {
                 return;
             }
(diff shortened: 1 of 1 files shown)


nietras:
Have you checked the benchmarks I did for .NET 4.6 RyuJIT 64-bit, not sure `MemoryCopy` is the best approach yet ;)
image

benaadams:
Well its currently using BlockCopy so MemoryCopy is currently better :smile: 

And it looks like it will improve

benaadams:
Updated https://github.com/aspnet/KestrelHttpServer/issues/1002 with further failure info

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Infrastructure/MemoryPoolIterator.cs
```diff
@@ -870,12 +870,30 @@ public int GetLength(MemoryPoolIterator end)
...
                 actual = 0;
                 return this;
             }
 
+            if (array == null)
+            {
+                return Consume(count, out actual);
+            }
+
+            Debug.Assert(count + offset <= array.Length);
```
cesarbs: Add message.

benaadams: Maybe should change it to actual throw?

benaadams: Also get the test the right way round XD

benaadams:
I'd hold on this - movement in coreclr on BlockCopy https://github.com/dotnet/coreclr/pull/7198

cesarbs:
@benaadams Any updates on this? https://github.com/dotnet/coreclr/pull/7198 was merged almost a month ago.

benaadams:
@cesarbs that change is in post 1.1.0 😞 

Could take this now and add issue to remeasure post 1.1.0?

halter73:
@benaadams Since this missed the 1.1 train and dotnet/coreclr#7198 is merged, should we close this?

benaadams:
Yeah, think so - landscape has changed. Might be worth a revalidation at some point; but don't think its worth keeping this PR open

Issue #806, opened 2016-05-05
Response.OnStarting should throw when response already started


Tratcher:
_From @pakrym on May 5, 2016 21:52_

This would prevent hard to debug issues with callbacks that never ran.

_Copied from original issue: aspnet/HttpAbstractions#627_

PR #1038, opened 2016-08-08
Throw when setting OnStarting after response has already started (#806).

cesarbs:
#806 

@halter73 @mikeharder @pakrym

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
@@ -382,6 +382,11 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Http
         {
             lock (_onStartingSync)
             {
+                if (HasResponseStarted)
+                {
+                    ThrowResponseAlreadyStartedException(nameof(OnStarting));
+                }
+
                 if (_onStarting == null)
                 {
                     _onStarting = new List<KeyValuePair<Func<object, Task>, object>>();
--- a/test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/ResponseTests.cs
+++ b/test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/ResponseTests.cs
@@ -9,6 +9,7 @@ using System.Threading.Tasks;
 using Microsoft.AspNetCore.Builder;
 using Microsoft.AspNetCore.Hosting;
 using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Server.Kestrel.Internal.Infrastructure;
 using Microsoft.Extensions.Primitives;
 using Xunit;
 
@@ -140,6 +141,39 @@ namespace Microsoft.AspNetCore.Server.Kestrel.FunctionalTests
                     Assert.Equal(HttpStatusCode.InternalServerError, response.StatusCode);
                     Assert.False(onStartingCalled);
                     Assert.True(onCompletedCalled);
+                }
+            }
+        }
+
+        [Fact]
+        public async Task OnStartingThrowsWhenSetAfterResponseHasAlreadyStarted()
+        {
+            InvalidOperationException ex = null;
+
+            var hostBuilder = new WebHostBuilder()
+                .UseKestrel()
+                .UseUrls("http://127.0.0.1:0/")
+                .Configure(app =>
+                {
+                    app.Run(async context =>
+                    {
+                        await context.Response.WriteAsync("hello, world");
+                        await context.Response.Body.FlushAsync();
+                        ex = Assert.Throws<InvalidOperationException>(() => context.Response.OnStarting(_ => TaskUtilities.CompletedTask, null));
+                    });
+                });
+
+            using (var host = hostBuilder.Build())
+            {
+                host.Start();
+
+                using (var client = new HttpClient())
+                {
+                    var response = await client.GetAsync($"http://localhost:{host.GetPort()}/");
+
+                    // Despite the error, the response had already started
+                    Assert.Equal(HttpStatusCode.OK, response.StatusCode);
+                    Assert.NotNull(ex);
                 }
             }
         }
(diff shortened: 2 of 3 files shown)


mikeharder:
Would it be worthwhile to add a functional test as well, showing how customer code would set `OnStarting()`?

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
```diff
@@ -382,6 +382,11 @@ public void OnStarting(Func<object, Task> callback, object state)
         {
             lock (_onStartingSync)
             {
+                if (HasResponseStarted)
+                {
+                    ThrowResponseAlreadyStartedException(nameof(OnStarting));
```
benaadams: Use an enum rather than a string; string version is additional calls, better to push it out to the infrequently called function. (related background info https://github.com/dotnet/coreclr/pull/6634#issuecomment-238325004)

cesarbs: You did that in #962, right? I can change this if we take it :smile:

mikeharder:
:shipit: from me, though you might want to review @benaadams comment.

davidfowl:
:shipit:

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
benaadams: I'll get a move on :wink:

cesarbs: @benaadams Could you elaborate on your first comment about using an enum? Not sure what's going on there.

benaadams: So with the string it inserts a load string call then calls the function
```asm
mov      ecx, 0x498F
call     CORINFO_HELP_STRCNS_CURRENT_MODULE
mov      rcx, rax
call     Frame:ThrowResponseAlreadyStartedException(ref)
```
With the enum it just sets a value
```asm
mov      ecx, 3
call     Frame:ThrowResponseAlreadyStartedException(int)
```
It does push the load string off to `Frame:ThrowResponseAlreadyStartedException` as the string still needs to get loaded; but that does really get called, so the the main function is more trim and the instruction cache hotter.

benaadams: Also hopefully it will also soon consistently move those calls to the bottom of the function as seen in this example (from https://github.com/dotnet/coreclr/pull/6634#issuecomment-238379006)

Jit inline decisions (decides not to inline the functions that throw [no-return])
```
Inlines into 06002A67 ArraySegment`1:.ctor(ref,int,int):this
  [0 IL=0004 TR=000067 06000C7B] [FAILED: does not return] ThrowHelper:ThrowArgumentNullException(int)
  [0 IL=0016 TR=000061 06000C69] [FAILED: does not return] ThrowHelper:ThrowArgumentOutOfRangeException(int,int)
  [0 IL=0028 TR=000054 06000C69] [FAILED: does not return] ThrowHelper:ThrowArgumentOutOfRangeException(int,int)
  [0 IL=0043 TR=000048 06000C77] [FAILED: does not return] ThrowHelper:ThrowArgumentException(int)
Budget: initialTime=270, finalTime=270, initialBudget=2700, currentBudget=2700
Budget: initialSize=1727, finalSize=1727
```
Moves the calling of the throw functions out of the flow, even though they were at the start to a section called "cold code"
```asm
; Assembly listing for method ArraySegment`1:.ctor(ref,int,int):this
; Emitting BLENDED_CODE for X64 CPU with SSE2
; optimized code
; rsp based frame
; partially interruptible
; Final local variable assignments
;
;  V00 this         [V00,T00] (  5,   5  )   byref  ->  rsi         this
;* V01 TypeCtx      [V01    ] (  0,   0  )    long  ->  zero-ref   
;  V02 arg1         [V02,T01] (  5,   5  )     ref  ->   r8        
;  V03 arg2         [V03,T02] (  5,   5  )     int  ->  rdi        
;  V04 arg3         [V04,T03] (  3,   3  )     int  ->  rbx        
;  V05 loc0         [V05    ] (  1,   1  )  lclBlk (32) [rsp+0x00]  
;
; Lcl frame size = 32
G_M41510_IG01:
       push     rdi
       push     rsi
       push     rbx
       sub      rsp, 32
       mov      rsi, rcx
       mov      edi, r9d
       mov      ebx, dword ptr [rsp+60H]
G_M41510_IG02:
       test     r8, r8
       je       G_M41510_IG08
G_M41510_IG03:
       test     edi, edi
       jl       G_M41510_IG09
G_M41510_IG04:
       test     ebx, ebx
       jl       G_M41510_IG10
G_M41510_IG05:
       mov      edx, dword ptr [r8+8]
       sub      edx, edi
       cmp      edx, ebx
       jl       G_M41510_IG11
G_M41510_IG06:
       mov      rcx, rsi
       mov      rdx, r8
       call     CORINFO_HELP_CHECKED_ASSIGN_REF
       mov      dword ptr [rsi+8], edi
       mov      dword ptr [rsi+12], ebx
G_M41510_IG07:
       add      rsp, 32
       pop      rbx
       pop      rsi
       pop      rdi
       ret      
************** Beginning of cold code **************
G_M41510_IG08:
       mov      ecx, 3
       call     ThrowHelper:ThrowArgumentNullException(int)
G_M41510_IG09:
       mov      ecx, 41
       mov      edx, 4
       call     ThrowHelper:ThrowArgumentOutOfRangeException(int,int)
G_M41510_IG10:
       mov      ecx, 16
       mov      edx, 4
       call     ThrowHelper:ThrowArgumentOutOfRangeException(int,int)
G_M41510_IG11:
       mov      ecx, 23
       call     ThrowHelper:ThrowArgumentException(int)
       int3     
; Total bytes of code 132, prolog size 7 for method ArraySegment`1:.ctor(ref,int,int):this
```

Issue #1030, opened 2016-08-04
NRE in AppVeyor while running tests


cesarbs:
https://ci.appveyor.com/project/aspnetci/kestrelhttpserver/build/1.0.3307

    System.NullReferenceException: Object reference not set to an instance of an object.
       at Microsoft.AspNetCore.Server.Kestrel.Internal.Networking.Libuv.unsafe_async_send(IntPtr handle) in C:\projects\kestrelhttpserver\src\Microsoft.AspNetCore.Server.Kestrel\Internal\Networking\Libuv.cs:line 173
       at Microsoft.AspNetCore.Server.Kestrel.Internal.Networking.UvAsyncHandle.ReleaseHandle() in C:\projects\kestrelhttpserver\src\Microsoft.AspNetCore.Server.Kestrel\Internal\Networking\UvAsyncHandle.cs:line 62
       at System.Runtime.InteropServices.SafeHandle.InternalFinalize()
       at System.Runtime.InteropServices.SafeHandle.Finalize()

halter73:
I think the only thing that could be null in that line of code is the `_uv_unsafe_async_send` Func. Surprising.

cesarbs:
Crazy. Will keep open for a while and if it turns out to have been a one-off thing I'll close.

halter73:
Thinking it might have something to do with `_uv_unsafe_async_send` being the only libuv function declared using `DllImportAttribute.EntryPoint`. It doesn't get called much, since the `UvAsyncHandle` doesn't usually get finalized.

halter73:
I just realized that it has to be a test using `MockLibuv`. That explains the func being null.

cesarbs:
Reproed with a StackTrace: https://ci.appveyor.com/project/aspnetci/kestrelhttpserver/build/1.0.3371

benaadams:
"System.Exception: Why is this getting called?" :smile:

halter73:
Could be fixed by https://github.com/aspnet/KestrelHttpServer/pull/1098. If we see this again, we can reopen with a link to the new failure.

PR #1039, opened 2016-08-08
Prevent NullReferenceException in test runs (#1030).

cesarbs:
#1030

cc @halter73 @mikeharder

### DIFF
--- a/test/Microsoft.AspNetCore.Server.KestrelTests/TestHelpers/MockLibuv.cs
+++ b/test/Microsoft.AspNetCore.Server.KestrelTests/TestHelpers/MockLibuv.cs
@@ -1,3 +1,6 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
 using System;
 using System.Threading;
 using Microsoft.AspNetCore.Server.Kestrel.Internal.Networking;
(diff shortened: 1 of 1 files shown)


test/Microsoft.AspNetCore.Server.KestrelTests/TestHelpers/MockLibuv.cs
```diff
@@ -76,6 +76,7 @@ unsafe public MockLibuv()
             _uv_strerror = errno => IntPtr.Zero;
             _uv_read_start = UvReadStart;
             _uv_read_stop = handle => 0;
+            _uv_unsafe_async_send = handle => 0;
```
halter73: This would leak memory if called :cry:

halter73: `=> _onPost(_postHandle.InternalGetHandle());` will probably work but it would be nice to know why this is getting called in the first place.

cesarbs:
Updated.

halter73:
Actually, Can we just update MockLibuv to capture the stack trace when constructed? Then we can throw an exception with the captured stack trace from `_uv_unsafe_async_send`. Then we can figure out which test is flaky.

cesarbs:
@halter73 Updated.

halter73:
:shipit:

Issue #1048, opened 2016-08-16
Kestrel applies chunking to "Connection: Upgrade" responses


anurse:
Discussed this with @halter73 so he already has most of the context, just filing this for tracking.

Recent changes to the way `Transfer-Encoding` is applied to the response now cause any response with `Connection: Upgrade` in the response headers to have HTTP chunking applied to the response. This is a Bad Thing(TM) because Kestrel shouldn't be messing with the response body of 101 responses because we've switched protocols.

This manifests in failures in our WebSockets code because the chunking metadata is being written out :).

/cc @Tratcher 

/fyi @pakrym @JunTaoLuo (this is causing the WebSockets CI failures)

JunTaoLuo:
Was this the reason for the hangs as well? It's good to know why it's causing failures but if it also causes hangs, it would be nice to change WebSockets to timeout and report a failure instead of hanging our builds.

anurse:
Yes, I plan to do that. As far as I can tell, it is the cause of the hangs (the hangs don't repro with 1.0.0 Kestrel either, which hints extremely strongly at the same cause :))

cesarbs:
@halter73 Shouldn't you have closed this?

anurse:
WebSockets appears to have been unbroken, so I'd also nominate this for closing :)

halter73:
@muratg  Considering this bug was never in a release, should we remove the bug label or milestone so it doesn't show up in the release notes.

anurse:
But leaving it in makes it look like we fixed more stuff :trollface: 

I'd agree, we shouldn't rel-note this, but @muratg or @Eilon can confirm that is the general approach :)

PR #1051, opened 2016-08-17
Disable automatic chunking for all non keep-alive requests

halter73:
- Fixes a bug where Upgrade requests (e.g. WebSockets) would be chunked
- Allows chunking to be disabled by setting "Connection: close" on the response

#1048

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
@@ -723,14 +723,8 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Http
             var end = SocketOutput.ProducingStart();
             if (_keepAlive && hasConnection)
             {
-                foreach (var connectionValue in responseHeaders.HeaderConnection)
-                {
-                    if (connectionValue.IndexOf("close", StringComparison.OrdinalIgnoreCase) != -1)
-                    {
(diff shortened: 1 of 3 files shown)


halter73:
@anurse @Tratcher @JunTaoLuo @pakrym @cesarbs

anurse:
:shipit:

pakrym:
:shipit:

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
```diff
@@ -723,13 +723,10 @@ private Task WriteSuffix()
             var end = SocketOutput.ProducingStart();
             if (_keepAlive && hasConnection)
             {
-                foreach (var connectionValue in responseHeaders.HeaderConnection)
+                var connectionValue = responseHeaders.HeaderConnection.ToString();
```
davidfowl: ToString()!?

halter73: Following the [pattern we use to check the request Connection header](https://github.com/aspnet/KestrelHttpServer/blob/dev/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/MessageBody.cs#L107) for whether or not it's been set to "keep-alive". Super cheap in the 99.9...% case where either it hasn't been set (and `hasConnection` is false) or the `Connection` header has a single value.

Issue #1052, opened 2016-08-17
Log when libuv binding fails


muratg:
Re issue reported in https://github.com/aspnet/KestrelHttpServer/issues/1020 

We decided to add logging to help customers when they hit this issue.

Add logging at https://github.com/aspnet/KestrelHttpServer/blob/dev/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Listener.cs#L47

PR #1054, opened 2016-08-18
Log all startup errors in KestrelServer

halter73:
#1052 

@cesarbs

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Kestrel/KestrelServer.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/KestrelServer.cs
@@ -56,17 +56,17 @@ namespace Microsoft.AspNetCore.Server.Kestrel
 
         public void Start<TContext>(IHttpApplication<TContext> application)
         {
-            ValidateOptions();
-
-            if (_disposables != null)
-            {
-                // The server has already started and/or has not been cleaned up yet
-                throw new InvalidOperationException("Server has already started.");
-            }
-            _disposables = new Stack<IDisposable>();
-
             try
             {
+                ValidateOptions();
+
+                if (_disposables != null)
+                {
+                    // The server has already started and/or has not been cleaned up yet
+                    throw new InvalidOperationException("Server has already started.");
+                }
+                _disposables = new Stack<IDisposable>();
+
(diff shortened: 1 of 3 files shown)


cesarbs:
:ship:

cwe1ss:
Startup exceptions in aspnet/hosting are logged as critical: https://github.com/aspnet/Hosting/blob/dev/src/Microsoft.AspNetCore.Hosting/Internal/HostingLoggerExtensions.cs#L66

Could you log this with level `Critical` as well? Seems to be the same severity and would make alerting a bit easier.

cwe1ss:
David once said that you try to avoid logging + re-throwing. Will this exception usually be handled by someone (hosting?) or does this blow up up the process anyway?

halter73:
@cwe1ss this blows up the process assuming that the caller to `WebHost.Start()` doesn't catch the exception.

When investigating #1052, we discovered that the windows event log doesn't show the message from the exception that crashes a .NET process (even though it shows up in the console) which is the primary motivation for this change. Unlike with an application error, it's not possible to show some sort of error page to communicate to the developer what's gone wrong, since the server failed to bind to the listening address.

I think making the error critical makes sense though.

cwe1ss:
Thx for your reply!

I guess Critical sounds good in that case. But will this even be usable for people? I guess most people only add logger providers in their Startup.Configure method (because that's in the template). Will this be called before this error is logged?
Or will this only work, if you add an existing ILoggerFactory to the `WebHostBuilder`?

cwe1ss:
I've just realized that the server needs the built pipeline, so startup is called before server.start - so just forget my previous comment :)

cwe1ss:
Last thing: it's a detail but I think it's weird to sometimes use logger methods with 0 as the explicit eventId and sometimes not. unfortunately, aspnet/logging doesn't offer extension methods for cases that also receive an inner exception, that's why KestrelServer.cs now contains both versions.

I've [created a PR](https://github.com/aspnet/Logging/pull/392) for this some time ago - maybe you can take a look at it someday if you think this would be useful.

halter73:
@cwe1ss `IServer.Start()` is called by `WebHost` after `Startup.Configure` completes, so Kestrel will log to whatever loggers are configured there.

PR #1055, opened 2016-08-18
Allocate DateHeaderValueManager byte array only once.

cesarbs:
Small thing but avoids an allocation every second. That repeated allocation is not necessary since the date always has the same number of characters (format is `ddd, dd MMM yyyy HH:mm:ss GMT`).


benaadams:
Torn reads? What happens if its read while being written?

cesarbs:
It's read via `_dateValues`, which is a new instance created after the data is in `_dateBytes`.

cesarbs:
@benaadams :sob: I guess I could lock but not sure we'd want that vs. allocating.

benaadams:
Is only 29 (+ 12?) bytes a second :smile:  

Isn't like its 88 bytes a request or 176 MB/s https://github.com/aspnet/HttpAbstractions/issues/677 at @ 2Mrps :trollface:

PR #1056, opened 2016-08-18
Add uv_timer_t wrappers to libuv layer

cesarbs:

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Networking/UvTimerHandle.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Networking/UvTimerHandle.cs
@@ -1,0 +1,57 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using Microsoft.AspNetCore.Server.Kestrel.Internal.Infrastructure;
+using Microsoft.Extensions.Logging;
+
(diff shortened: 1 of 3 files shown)


benaadams:
`System.Threading.Timer`? 

Has the advantage if they don't perform well you know where they work ;-)

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Networking/UvTimerHandle.cs
```diff
@@ -0,0 +1,57 @@
...
+
+using System;
+using Microsoft.AspNetCore.Server.Kestrel.Internal.Infrastructure;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Networking
+{
+    public class UvTimerHandle : UvHandle
+    {
+        private readonly static Libuv.uv_timer_cb _uv_timer_cb = (handle) => UvTimerCb(handle);
```
txdv: Why not _uv_timer_cb = UvTimerCb? It seems like the signatures match.

txdv:
What are the timers going to be used for?

davidfowl:
Timeouts

cesarbs:
@benaadams We've decided to go with uv timers to keep them in the event loop. With framework timers we'd have to post a lot (with the current plan for implementing timeouts).

halter73:
Looks reasonable. I'll hold off on reviewing until I see how it's used.

cesarbs:
Closing this since #1066 has the same changes.

Issue #1028, opened 2016-08-04
When HttpResponse.Body is replaced, the replacement is used for future requests


muratg:
Copied from https://github.com/aspnet/KestrelHttpServer/issues/940 to consider this for 1.0.1.
---

This is a contrived example to illustrate the point:

```
app.Use(next => context =>
{
    if (context.Response.Body is MemoryStream)
    {
        throw new Exception("Wrong stream type in context.Response.Body");
    }

    context.Response.Body = new MemoryStream();

    return context.Response.WriteAsync("Hello, world!");
});
```

Run the application and browse to it. The first request returns nothing (content was written to the MemoryStream). Subsequent requests will hit the Exception.

The real-world case is when the Body stream is decorated for a given request. The decorator stream will be re-used for a subsequent request, which could result in re-decorating the stream, and ultimately produce a long chain of decorators.

It appears that Kestrel is trying to be efficient by re-using its underlying Stream instances. But the reusable instances are being replaced, with unintended consequences for future requests.

cesarbs:
@muratg Should I close this? Or are we waiting for something before closing 1.0.1 issues?

muratg:
@cesarbs Sure, make sure to update the label to done.

PR #1060, opened 2016-08-22
Reset frame streams on each request (#1028).

cesarbs:
Fix going into 1.0.1.

#1028

cc @halter73 @mikeharder @pranavkm @muratg


halter73:
:shipit:

benaadams:
:rocket:

Issue #1061, opened 2016-08-25
Make Kestrel's response buffer limit configurable


halter73:
Kestrel will not return an incomplete task from Response.WriteAsync(), or block on a synchronous Write() call, until 64KB more are buffered. Right now this limit is an internal constant in `SocketOutput` called `_maxBytesPreCompleted`. This should be made configurable like `MaxRequestBufferSize`.

PR #1062, opened 2016-08-25
Make Kestrel's response buffer limit configurable

halter73:
#1061

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/SocketOutput.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/SocketOutput.cs
@@ -15,7 +15,6 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Http
     public class SocketOutput : ISocketOutput
     {
         private const int _maxPendingWrites = 3;
-        private const int _maxBytesPreCompleted = 65536;
         // Well behaved WriteAsync users should await returned task, so there is no need to allocate more per connection by default
         private const int _initialTaskQueues = 1;
         private const int _maxPooledWriteContexts = 32;
@@ -26,6 +25,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Http
         private readonly KestrelThread _thread;
         private readonly UvStreamHandle _socket;
         private readonly Connection _connection;
+        private readonly long? _maxBytesPreCompleted;
         private readonly string _connectionId;
         private readonly IKestrelTrace _log;
         private readonly IThreadPool _threadPool;
@@ -51,7 +51,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Http
         private bool _postingWrite = false;
 
         private bool _cancelled = false;
-        private int _numBytesPreCompleted = 0;
+        private long _numBytesPreCompleted = 0;
         private Exception _lastWriteError;
         private WriteContext _nextWriteContext;
         private readonly Queue<WaitingTask> _tasksPending;
@@ -75,6 +75,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Http
             _tasksPending = new Queue<WaitingTask>(_initialTaskQueues);
             _writeContextPool = new Queue<WriteContext>(_maxPooledWriteContexts);
             _writeReqPool = thread.WriteReqPool;
+            _maxBytesPreCompleted = connection.ServerOptions.Limits.MaxResponseBufferSize;
 
             _head = thread.Memory.Lease();
             _tail = _head;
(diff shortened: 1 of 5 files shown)


halter73:
@cesarbs @davidfowl

src/Microsoft.AspNetCore.Server.Kestrel/KestrelServerLimits.cs
```diff
@@ -21,6 +24,30 @@ public class KestrelServerLimits
...
 
         /// <summary>
+        /// Gets or sets the maximum size of the response buffer before calls
+        /// write calls begin to block or return incomplete tasks.
+        /// </summary>
+        /// <remarks>
+        /// When set to null, the size of the response buffer is unlimited.
+        /// Defaults to 65536 bytes (64 KB).
+        /// </remarks>
+        public long? MaxResponseBufferSize
```
davidfowl: nullable huh? With great power comes great responsibility. Netty and node both have a high and low watermark used for backpressure. Outgoing writes are halted when it hits the high watermark and continue once the low watermark is hit.

cesarbs: It's consistent with `MaxRequestBufferSize`.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/SocketOutput.cs
```diff
@@ -75,6 +75,7 @@ public class SocketOutput : ISocketOutput
             _tasksPending = new Queue<WaitingTask>(_initialTaskQueues);
             _writeContextPool = new Queue<WriteContext>(_maxPooledWriteContexts);
             _writeReqPool = thread.WriteReqPool;
+            _maxBytesPreCompleted = connection.ServerOptions.Limits.MaxResponseBufferSize;
```
cesarbs: What if it's null?

src/Microsoft.AspNetCore.Server.Kestrel/KestrelServerLimits.cs
```diff
@@ -21,6 +24,30 @@ public class KestrelServerLimits
         private int _maxRequestHeaderCount = 100;
 
         /// <summary>
+        /// Gets or sets the maximum size of the response buffer before calls
```
cesarbs: there's an extra "calls" here

src/Microsoft.AspNetCore.Server.Kestrel/KestrelServerLimits.cs
davidfowl: Sure, that has the same problem as this. When if you end up in a situation where the consumer is reading slowly, you'll end up pausing and resuming constantly. Maybe we can add it later...

test/Microsoft.AspNetCore.Server.KestrelTests/SocketOutputTests.cs
```diff
@@ -32,9 +68,9 @@ public class SocketOutputTests
                 var socket = new MockSocket(mockLibuv, kestrelThread.Loop.ThreadId, new TestKestrelTrace());
                 var trace = new KestrelTrace(new TestKestrelTrace());
                 var ltp = new SynchronousThreadPool();
-                var socketOutput = new SocketOutput(kestrelThread, socket, new MockConnection(), "0", trace, ltp);
+                var socketOutput = new SocketOutput(kestrelThread, socket, new MockConnection(options), "0", trace, ltp);
 
-                // I doubt _maxBytesPreCompleted will ever be over a MB. If it is, we should change this test.
+                // At least one run of this test should have a MaxResponseBufferSize < 1048576.
```
cesarbs: nit: 1 MB or 1,048,576 for easier reading

test/Microsoft.AspNetCore.Server.KestrelTests/SocketOutputTests.cs
```diff
@@ -52,10 +88,120 @@ public class SocketOutputTests
...
+
+                var kestrelThread = kestrelEngine.Threads[0];
+                var socket = new MockSocket(mockLibuv, kestrelThread.Loop.ThreadId, new TestKestrelTrace());
+                var trace = new KestrelTrace(new TestKestrelTrace());
+                var ltp = new SynchronousThreadPool();
+                var options = new KestrelServerOptions { Limits = { MaxResponseBufferSize = null } };
+                var socketOutput = new SocketOutput(kestrelThread, socket, new MockConnection(options), "0", trace, ltp);
+
+                // Don't want to allocate anything too huge for perf. This is at least larger than the default buffer.
+                var bufferSize = 1048576;
```
cesarbs: nit: `1024 * 1024`, again for easier reading

src/Microsoft.AspNetCore.Server.Kestrel/KestrelServerLimits.cs
```diff
@@ -21,6 +24,30 @@ public class KestrelServerLimits
...
+        /// write calls begin to block or return incomplete tasks.
+        /// </summary>
+        /// <remarks>
```
cesarbs: Add remark about 0 disabling response buffering.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/SocketOutput.cs
```diff
@@ -51,7 +51,7 @@ public class SocketOutput : ISocketOutput
         private bool _postingWrite = false;
 
         private bool _cancelled = false;
-        private int _numBytesPreCompleted = 0;
+        private long _numBytesPreCompleted = 0;
```
cesarbs: One thing to watch out for here is that `long` is not guaranteed to have atomic reads and writes. Do we ever read and write this concurrently?

davidfowl: It's updated under the `_contextLock`

PR #1063, opened 2016-08-25
Remove unneeded unsafe modifiers

benaadams:


txdv:
Should be a compiler feature. Throw warnings on safe keywords which are basically not used.

halter73:
:ship:

PR #1066, opened 2016-08-25
Add keep-alive timeout (#464)

cesarbs:
This includes the libuv timer stuff in #1056 plus the keep-alive code.

~~There's a ~5% perf hit. I'm trying to improve that.~~

#464

cc @halter73 @mikeharder @davidfowl @Tratcher


src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Connection.cs
```diff
@@ -156,6 +161,18 @@ public virtual void OnSocketClosed()
...
         }
 
+        // Called on Libuv thread
+        public void Tick(long now)
+        {
+            if (_frame.Status == Frame.RequestProcessingStatus.RequestPending)
+            {
+                if ((now - _requestEndTime) / NanosecondsPerSecond > ServerOptions.Limits.KeepAliveTimeout)
+                {
+                    SocketInput.IncomingComplete(0, null);
```
halter73: StopAsync()

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Networking/Libuv.cs
```diff
@@ -594,6 +628,18 @@ private static class NativeMethods
...
+            unsafe public static extern int uv_timer_init(UvLoopHandle loop, UvTimerHandle handle);
+
+            [DllImport("libuv", CallingConvention = CallingConvention.Cdecl)]
+            unsafe public static extern int uv_timer_start(UvTimerHandle handle, uv_timer_cb cb, long timeout, long repeat);
+
+            [DllImport("libuv", CallingConvention = CallingConvention.Cdecl)]
+            unsafe public static extern int uv_timer_stop(UvTimerHandle handle);
+
+            [DllImport("libuv", CallingConvention = CallingConvention.Cdecl)]
+            unsafe public static extern long uv_hrtime();
```
halter73: looks like `uv_now` will be a little more efficient since we don't need submillisecond granularity.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Infrastructure/KestrelThread.cs
```diff
@@ -320,6 +328,18 @@ private void OnPost()
             } while (wasWork && loopsRemaining > 0);
         }
 
+        private void OnHeartbeat(UvTimerHandle timer)
+        {
+            Walk(ptr =>
+            {
+                var now = _engine.Libuv.hrtime();
+                //var now = DateTime.UtcNow.Ticks;
```
halter73: Nit: remove commented code

benaadams: DateTimeHeaderValueManager thing updates its time object every second; could add an actual DateTime property to it (as well as byte and string) and use that?

benaadams: Might also mean you can use its interface (that's created for testing it), to jump time around and test the timeouts more easily - without actually timing out.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Infrastructure/KestrelThread.cs
```diff
@@ -320,6 +328,17 @@ private void OnPost()
             } while (wasWork && loopsRemaining > 0);
         }
 
+        private void OnHeartbeat(UvTimerHandle timer)
+        {
+            Walk(ptr =>
+            {
+                var now = _loop.Now();
```
cesarbs: This shouldn't be in the walk.

cesarbs:
Thanks @halter73 for the suggestion about counting seconds between ticks instead of querying time. That fixed the perf regression :+1:

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/IConnectionControl.cs
```diff
@@ -8,5 +8,8 @@ public interface IConnectionControl
         void Pause();
         void Resume();
         void End(ProduceEndType endType);
+
+        void NotifyRequestStarted();
```
davidfowl: Why are these on `IConnectionControl`

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Infrastructure/KestrelThread.cs
```diff
@@ -267,6 +269,8 @@ private void ThreadStart(object parameter)
                 {
                     _loop.Init(_engine.Libuv);
                     _post.Init(_loop, OnPost, EnqueueCloseHandle);
+                    _heartbeatTimer.Init(_loop, EnqueueCloseHandle);
+                    _heartbeatTimer.Start(OnHeartbeat, 1000, 1000);
```
davidfowl: 1000, 1000, use named parameters

src/Microsoft.AspNetCore.Server.Kestrel/KestrelServerLimits.cs
```diff
@@ -108,5 +111,27 @@ public int MaxRequestHeaderCount
...
             }
         }
+
+        /// <summary>
+        /// Gets or sets the keep-alive timeout, in seconds.
+        /// </summary>
+        /// <remarks>
+        /// Defaults to 5 seconds.
+        /// </remarks>
+        public int KeepAliveTimeout
```
davidfowl: Timespan

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/KeepAliveTimeoutTests.cs
```diff
@@ -0,0 +1,222 @@
...
+using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Testing;
+using Xunit;
+
+namespace Microsoft.AspNetCore.Server.Kestrel.FunctionalTests
+{
+    public class KeepAliveTimeoutTests
+    {
+        private const int KeepAliveTimeout = 1; // seconds
+        private const int LongDelay = 5000; // milliseconds
```
davidfowl: Timespans

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Connection.cs
```diff
@@ -156,6 +157,21 @@ public virtual void OnSocketClosed()
             _socketClosedTcs.TrySetResult(null);
         }
 
+        // Called on Libuv thread
+        public void Tick()
+        {
+            if (_frame.Status == Frame.RequestProcessingStatus.RequestPending && // we're in between requests and
+                !SocketInput.IsCompleted)                                        // we haven't just started a new request
```
davidfowl: This is meant to be checking the connection's SocketInput not the frame's right?

src/Microsoft.AspNetCore.Server.Kestrel/KestrelServerLimits.cs
cesarbs: Why? Every other server out there uses seconds.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Connection.cs
cesarbs: Yes.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/IConnectionControl.cs
cesarbs: What do you suggest? To me it makes sense to expose these here - you use these to notify the signal the connection about request processing events so it can react accordingly.

Do you think I should introduce a new interface? `IRequestStatusNotification` or something along those lines?

src/Microsoft.AspNetCore.Server.Kestrel/KestrelServerLimits.cs
benaadams: Removes the "what's unit" question.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/IConnectionControl.cs
benaadams: Would it be wrong to put them straight onto connection?

src/Microsoft.AspNetCore.Server.Kestrel/KestrelServerLimits.cs
cesarbs: But could lead to false expectations i.e. < 1 second wouldn't work, also e.g. 2.5 wouldn't be much different than 2 or 3 (good deal of "error" as to when timeout is checked).

benaadams: `(int)Math.Ceiling(timespan.TotalSeconds)` then add to remarks "rounds up to second"?

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/IConnectionControl.cs
cesarbs: Would require `Frame` to known about `Connection` - stronger coupling.

cesarbs:
⌚

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
```diff
@@ -1267,6 +1267,25 @@ protected void ReportApplicationError(Exception ex)
             Log.ApplicationError(ConnectionId, ex);
         }
 
+        public void Tick()
+        {
+            if (_requestProcessingStatus == RequestProcessingStatus.RequestPending && // we're in between requests and
```
halter73: We almost never put comments at the end of a line. I get why you did it, but just put the comment above the if and use "and" in the comment.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
```diff
@@ -1277,11 +1296,10 @@ protected enum RequestLineStatus
             Done
         }
 
-        protected enum RequestProcessingStatus
+        public enum RequestProcessingStatus
         {
             RequestPending,
-            RequestStarted,
-            ResponseStarted
```
halter73: Revert this until we have a reason to denormalize.

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/KeepAliveTimeoutTests.cs
```diff
@@ -0,0 +1,223 @@
...
+namespace Microsoft.AspNetCore.Server.Kestrel.FunctionalTests
+{
+    public class KeepAliveTimeoutTests
+    {
+        private static readonly TimeSpan KeepAliveTimeout = TimeSpan.FromSeconds(1);
+        private const int LongDelay = 5000; // milliseconds
+        private const int ShortDelay = 250; // milliseconds
+
+        [Fact]
+        public async Task ConnectionClosedWhenKeepAliveTimeoutExpires()
```
halter73: I don't think xunit will run tests in the same class in parallel by default. Since these are longer running tests, is there a way we can make these run concurrently?

test/Microsoft.AspNetCore.Server.KestrelTests/ConnectionTests.cs
```diff
@@ -29,13 +32,20 @@ public void DoesNotEndConnectionOnZeroRead()
...
-                var socket = new MockSocket(mockLibuv, Thread.CurrentThread.ManagedThreadId, trace);
-                var connection = new Connection(context, socket);
-                connection.Start();
 
-                Libuv.uv_buf_t ignored;
-                mockLibuv.AllocCallback(socket.InternalGetHandle(), 2048, out ignored);
-                mockLibuv.ReadCallback(socket.InternalGetHandle(), 0, ref ignored);
+                MockSocket socket = null;
+                Connection connection = null;
+                await context.Thread.PostAsync(_ =>
```
halter73: This certainly seems more correct, by why did this just start failing with this change?

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/KeepAliveTimeoutTests.cs
```diff
@@ -0,0 +1,223 @@
...
+            host.Start();
+            return host;
+        }
+
+        private async Task ReceiveResponse(TestConnection connection)
+        {
+            await connection.Receive(
+                "HTTP/1.1 200 OK",
+                "");
+            await connection.ReceiveStartsWith("Date: ");
```
halter73: Use a `MockSystemClock` like `TestServer`. It was nice to get rid of `ReceiveStartsWith`, and I don't want to see it back.

test/Microsoft.AspNetCore.Server.KestrelTests/UvTimerHandleTests.cs
```diff
@@ -0,0 +1,71 @@
...
+            loop.Init(new Libuv());
+
+            var timer = new UvTimerHandle(trace);
+            timer.Init(loop, (a, b) => { });
+
+            var callbackInvoked = false;
+            timer.Start(_ =>
+            {
+                callbackInvoked = true;
+            }, 50, 0);
```
halter73: Can we make these timeouts even faster?

cesarbs: I'll change it to 1ms.

cesarbs:
@halter73 Using `TestServer` and I'm tentatively addressing your concern about long standing tests in the last commit (d09230e).

cesarbs:
Travis test failure on OS X (https://travis-ci.org/aspnet/KestrelHttpServer/jobs/157181697) is not related to this change since it's showing up in other branches (https://travis-ci.org/aspnet/KestrelHttpServer/jobs/157157166, https://travis-ci.org/aspnet/KestrelHttpServer/jobs/155406648 - in the latter it was the other test in the class that failed).

Still waiting for a 🚢  or more feedback.

⌚

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
```diff
@@ -1277,7 +1295,7 @@ protected enum RequestLineStatus
             Done
         }
 
-        protected enum RequestProcessingStatus
+        public enum RequestProcessingStatus
```
halter73: Why is this now public when the field was made private?

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Infrastructure/KestrelThread.cs
```diff
@@ -327,6 +335,16 @@ private void OnPost()
             } while (wasWork && loopsRemaining > 0);
         }
 
+        private void OnHeartbeat(UvTimerHandle timer)
+        {
+            Walk(ptr =>
+            {
+                var handle = UvMemory.FromIntPtr<UvHandle>(ptr);
+                var connection = (handle as UvStreamHandle)?.Connection;
+                connection?.Tick();
```
halter73: Nit: probably just as clear combined with the line above.

halter73:
:shipit:

PR #1067, opened 2016-08-25
Test for categoryName passed to ILoggerFactory.CreateLogger() in KestrelServer.ctor() (#797)

cesarbs:
#797

cc @mikeharder @halter73

### DIFF
--- a/test/Microsoft.AspNetCore.Server.KestrelTests/KestrelServerTests.cs
+++ b/test/Microsoft.AspNetCore.Server.KestrelTests/KestrelServerTests.cs
@@ -2,14 +2,14 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.Reflection;
 using Microsoft.AspNetCore.Hosting.Server;
 using Microsoft.AspNetCore.Hosting.Server.Features;
-using Microsoft.AspNetCore.Http.Features;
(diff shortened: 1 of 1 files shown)


test/Microsoft.AspNetCore.Server.KestrelTests/KestrelServerTests.cs
```diff
@@ -75,6 +75,14 @@ public void StartWithMaxRequestBufferSizeLessThanMaxRequestLineSizeThrows(long m
             Assert.Equal(1, testLogger.CriticalErrorsLogged);
         }
 
+        [Fact]
+        public void LoggerCategoryNameIsKestrelServerNamespace()
+        {
+            var mockLoggerFactory = new Mock<ILoggerFactory>();
+            new KestrelServer(Options.Create<KestrelServerOptions>(null), new LifetimeNotImplemented(), mockLoggerFactory.Object);
+            mockLoggerFactory.Verify(factory => factory.CreateLogger(typeof(KestrelServer).GetTypeInfo().Namespace));
```
halter73: If we really want to test we don't break things by changing the category name, shouldn't we just have the namespace as the string literal "Microsoft.AspNetCore.Server.Kestrel". I don't think that anyone will change the namespace of KestrelServer anytime soon since it's public, but might as well be sure.

cesarbs:
Updated.

halter73:
:shipit:

Issue #1070, opened 2016-08-29
Tests receiving empty responses on OSX


cesarbs:
System.Net.Http.HttpRequestException : An error occurred while sending the request.
    ---- System.Net.Http.CurlException : Server returned nothing (no headers, no data)
       at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
       at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
       at System.Runtime.CompilerServices.ConfiguredTaskAwaitable`1.ConfiguredTaskAwaiter.GetResult()
       at System.Net.Http.HttpClient.<FinishSendAsync>d__58.MoveNext()
    --- End of stack trace from previous location where exception was thrown ---
       at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
       at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
       at Microsoft.AspNetCore.Server.Kestrel.FunctionalTests.AddressRegistrationTests.<RegisterAddresses_Success>d__5.MoveNext() in /Users/asplab/Documents/buildAgent/work/d6337ac2abf63a4b/.repositories/KestrelHttpServer/test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/AddressRegistrationTests.cs:line 76
    --- End of stack trace from previous location where exception was thrown ---
       at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
       at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
       at Microsoft.AspNetCore.Server.Kestrel.FunctionalTests.AddressRegistrationTests.<RegisterAddresses_IPv6_Success>d__2.MoveNext() in /Users/asplab/Documents/buildAgent/work/d6337ac2abf63a4b/.repositories/KestrelHttpServer/test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/AddressRegistrationTests.cs:line 41
    --- End of stack trace from previous location where exception was thrown ---
       at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
       at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
[...]

PR #1071, opened 2016-08-30
Run functional tests sequentially to mitigate flakiness on OS X

cesarbs:
@pranavkm This should mitigate the flakiness we're seeing on OS X with errors like `Server returned nothing (no headers, no data)`. It did when we were seeing the same error on `ThreadCountTests.OneToTenThreads`.

If this doesn't work, I'll revert the change and disable the offending tests (which potentially means all `AddressRegistrationTests`, all `PathBaseTests` and all `ResponseTests`, unfortunately).

The actual issue here seems to be that the process runs out of handles while running the tests - not sure why it manifests the way it does. Running sequentially reduces the amount of handles being allocated at the same time. That's the closest we got to an answer at the time we were investigating `ThreadCountTests`, but we ended up deciding to just disable those tests on OS X and move on.

~~@Eilon suggested I make this change in dev and port it to 1.0.1, but on a second thought I decided not to because that would impact dev work, since they take twice as long to run with this change. What we're really concerned with here is unblocking 1.0.1 builds.~~

I've filed https://github.com/aspnet/KestrelHttpServer/issues/1070 to have this properly investigated.

cc @halter73 @muratg

### DIFF
--- a/test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/Properties/AssemblyInfo.cs
+++ b/test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/Properties/AssemblyInfo.cs
@@ -1,0 +1,3 @@
+﻿using Xunit;
+
+[assembly: CollectionBehavior(CollectionBehavior.CollectionPerAssembly)]


benaadams:
OS X was very grumpy with cli and restores. I think they upped the `ulimit` to get round it??

pranavkm:
Might want to edit the commit message to say that it's running functional tests serially (and not parallel as it does by default).

pranavkm:
:shipit:

cesarbs:
@benaadams Our build script ups the ulimit but we see these issues nevertheless 😞

cesarbs:
Updated with correct commit message.

Issue #1049, opened 2016-08-16
Flaky test: DateHeaderValueManagerTests


JunTaoLuo:
I'm seeing this test fail in our CI: http://aspnetci/viewLog.html?buildId=126592&buildTypeId=Lite_UniverseTest&tab=buildLog#_

```
[Microsoft.AspNetCore.Server.KestrelTests.DateHeaderValueManagerTests.GetDateHeaderValue_ReturnsUpdatedValueAfterIdle] Assert.Equal() Failure
                                   (pos 23)
Expected: úúú, 16 Aug 2016 22:51:28 GMT
Actual:   úúú, 16 Aug 2016 22:51:18 GMT
                                   (pos 23)
   at Microsoft.AspNetCore.Server.KestrelTests.DateHeaderValueManagerTests.<GetDateHeaderValue_ReturnsUpdatedValueAfterIdle>d__2.MoveNext() in C:\BuildAgent\work\33bdfc1cae7b2a38\KestrelHttpServer\test\Microsoft.AspNetCore.Server.KestrelTests\DateHeaderValueManagerTests.cs:line 101
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
```

This is inconsistent but still would warrant some investigation.

JunTaoLuo:
http://aspnetci/viewLog.html?buildId=126746&tab=buildResultsDiv&buildTypeId=XPlat_Mac_MacUniverseCoreCLR
Also seeing this flaky test on mac
Microsoft.AspNetCore.Server.Kestrel.FunctionalTests.AddressRegistrationTests.RegisterAddresses_IPv4_Success
```
[...]

PR #1072, opened 2016-08-30
Make SocketOutputTests more reliable

halter73:
#1049 

@cesarbs


src/Microsoft.AspNetCore.Server.Kestrel/Internal/Infrastructure/LoggingThreadPool.cs
```diff
@@ -63,6 +63,11 @@ public void Run(Action action)
             ThreadPool.QueueUserWorkItem(_runAction, action);
         }
 
+        public void UnsafeRun(WaitCallback action, object state)
```
cesarbs: Why "unsafe"?

davidfowl: No try catch.

cesarbs:
LGTM 🚀

PR #1073, opened 2016-08-31
Serialize functional tests on OS X

cesarbs:
@pranavkm @halter73 

https://github.com/aspnet/KestrelHttpServer/issues/1070

Kinda gross, but `Environment.OSVersion.Platform` gives us "this is Unix" at best - even though there's a `MacOSX` value in `PlatformID`, it returns `Unix` on OS X. Since we get `Unix` on Linux too, that would affect Linux builds as well.

A better option might be to check for some specific file in /etc or whatever that we know is only there on OS X. Does anyone know of any such file?

### DIFF
--- a/makefile.shade
+++ b/makefile.shade
@@ -1,11 +1,53 @@
+use namespace="System.Diagnostics"
+use namespace="System.IO"
+
 var VERSION='0.1'
 var FULL_VERSION='0.1'
 var AUTHORS='Microsoft'
-var kestrelSrcDir='${Path.Combine(Directory.GetCurrentDirectory(), "src/Microsoft.AspNetCore.Server.Kestrel")}'
 
 use-standard-lifecycle
 k-standard-goals
 custom-goals
 
-#initialize if='Directory.Exists("src")'
-  exec program='dotnet' commandline='run -p ../../tools/Microsoft.AspNetCore.Server.Kestrel.GeneratedCode Internal/Http/FrameHeaders.Generated.cs Internal/Http/Frame.Generated.cs' workingdir='${kestrelSrcDir}'
+#initialize
+  @{
+    if (Directory.Exists("src"))
+    {
+      Exec("dotnet",
+        commandline: "run -p ../../tools/Microsoft.AspNetCore.Server.Kestrel.GeneratedCode Internal/Http/FrameHeaders.Generated.cs Internal/Http/Frame.Generated.cs",
+        workingdir: Path.Combine(Directory.GetCurrentDirectory(), "src/Microsoft.AspNetCore.Server.Kestrel"));
+    }
+
+    if (IsOSX())
+    {
+      var noParallelTestProjects = Environment.GetEnvironmentVariable("NO_PARALLEL_TEST_PROJECTS") ?? string.Empty;
+      noParallelTestProjects += ",Microsoft.AspNetCore.Server.Kestrel.FunctionalTests";
+      Environment.SetEnvironmentVariable("NO_PARALLEL_TEST_PROJECTS", noParallelTestProjects);
+    }
+  }
+
+functions @{
+  bool IsOSX()
+  {
+    if (!IsLinux)
+    {
+      return false;
+    }
+
+    var processStartInfo = new ProcessStartInfo
+    {
+      FileName = "uname",
+      RedirectStandardOutput = true,
+      UseShellExecute = false
+    };
+    var output = string.Empty;
+
+    using (var process = Process.Start(processStartInfo))
+    {
+      output = process.StandardOutput.ReadToEnd();
+      process.WaitForExit();
+    }
+
+    return output.StartsWith("Darwin");
+  }
+}


makefile.shade
```diff
@@ -9,3 +12,40 @@ custom-goals
...
+
+#initialize if='IsOSX()'
+  @{
+    var noParallelTestProjects = Environment.GetEnvironmentVariable("NO_PARALLEL_TEST_PROJECTS") ?? string.Empty;
+    noParallelTestProjects += ",Microsoft.AspNetCore.Server.Kestrel.FunctionalTests";
+    Environment.SetEnvironmentVariable("NO_PARALLEL_TEST_PROJECTS", noParallelTestProjects);
+  }
+
+functions @{
+  bool IsOSX()
```
pranavkm: Does `Environment.OSVersion` not work?

BrennanConroy: Read the PR summary

pranavkm:
Can you guys live with running it serially on Linux (until we switch from Sake)? Seems unnecessary to have to exec uname

makefile.shade
```diff
@@ -9,3 +12,40 @@ custom-goals
...
+  {
+    var output = string.Empty;
+
+    var processStartInfo = new ProcessStartInfo
+    {
+      UseShellExecute = false,
+      FileName = "uname"
```
pranavkm: Doesn't this fail on Windows (where it doesn't have `uname`)?

makefile.shade
```diff
@@ -9,3 +12,40 @@ custom-goals
 
 #initialize if='Directory.Exists("src")'
   exec program='dotnet' commandline='run -p ../../tools/Microsoft.AspNetCore.Server.Kestrel.GeneratedCode Internal/Http/FrameHeaders.Generated.cs Internal/Http/Frame.Generated.cs' workingdir='${kestrelSrcDir}'
+
+#initialize if='IsOSX()'
```
pranavkm: This will overwrite the previous `#initialize` statement.

mikeharder:
Why are we serializing the functional tests on OSX?  Is it because of the file descriptor limit again?

halter73:
@mikeharder The assumption is that the file descriptor limit is the problem again. It's the same tests that are failing in the same way as before (sometimes) on OSX.

makefile.shade
cesarbs: Oh, I thought they were cumulative 😞  Is there a way to run multiple things at the same target based on different conditions?

cesarbs:
@pranavkm I'm fine with having them run sequentially on Linux too, but we need consensus from @halter73 and @mikeharder.

mikeharder:
I would much rather limit this to OSX, so we at least run in parallel on one flavor of Unix.

makefile.shade
pranavkm: There're are ways, but its probably less change to tack this on to the existing target.

pranavkm:
:watch:

halter73:
I'm with @mikeharder. I don't want serial test runs on linux.

pranavkm:
I'm ok with that. Just waiting on the `#initialize` changes (and verifying this works on Windows where `uname` doesn't exist)

cesarbs:
Updated:

- Works on Windows
- Fixed `#initialize`

makefile.shade
```diff
@@ -1,11 +1,65 @@
...
+      {
+        UseShellExecute = false,
+        FileName = "uname"
+      };
+
+      processStartInfo.RedirectStandardOutput = true;
+
+      using (var process = Process.Start(processStartInfo))
+      {
+        process.EnableRaisingEvents = true;
```
pranavkm: Couldn't you `Output.ReadToEnd()` here instead of listening for events? `uname` doesn't have a lot of spew so this should be fine.

makefile.shade
```diff
@@ -1,11 +1,65 @@
...
+      var noParallelTestProjects = Environment.GetEnvironmentVariable("NO_PARALLEL_TEST_PROJECTS") ?? string.Empty;
+      noParallelTestProjects += ",Microsoft.AspNetCore.Server.Kestrel.FunctionalTests";
+      Environment.SetEnvironmentVariable("NO_PARALLEL_TEST_PROJECTS", noParallelTestProjects);
+    }
+  }
+
+functions @{
+  bool IsOSX()
+  {
+    try
```
pranavkm:
```
if (!IsLinux)
{
  return false;
}
```

pranavkm:
Could you make this in dev instead of `rel/1.0.1`? Let's not add more churn as part of creating a patch.

pranavkm:
:shipit:

cesarbs:
Updated.

cesarbs:
@pranavkm The problem is that doing this only on dev will introduce a discrepancy between dev and 1.0.1. If we merge 1.0.1 to dev without this change, we'll get the `AssemblyInfo.cs` stuff without the reversal of the change.

pranavkm:
:shipit: x2

cesarbs:
7d1a03f51d64f09371f0ae4dbfb6bc32089f1e37

PR #1075, opened 2016-08-31
Make the char array for utf8 decoding smaller

davidfowl:
/cc @troydai @Tratcher

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Infrastructure/MemoryPoolIteratorExtensions.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Infrastructure/MemoryPoolIteratorExtensions.cs
@@ -150,7 +150,8 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Infrastructure
             var decoder = _utf8.GetDecoder();
 
             var length = start.GetLength(end);
-            var charLength = length * 2;
+            var charLength = length;
+            // Worse case is 1 byte = 1 char
             var chars = new char[charLength];
             var charIndex = 0;
 


troydai:
:shipit: 

The conversion of the length here is from the length of a byte array to the length of a char array, in which case I can't find a case where the former will larger than the latter. The original case, which I wrote, mistakenly take the original length as the count of characters.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Infrastructure/MemoryPoolIteratorExtensions.cs
```diff
@@ -150,7 +150,8 @@ public static string GetUtf8String(this MemoryPoolIterator start, MemoryPoolIter
             var decoder = _utf8.GetDecoder();
 
             var length = start.GetLength(end);
-            var charLength = length * 2;
+            var charLength = length;
```
rynowak: make it `length * sizeof(byte) / 8` in case it changes

davidfowl: Not following.

rynowak: Bro it's just a prank :trollface:

davidfowl: I couldn't tell if you were trolling me. Internetz hard hard

benaadams: @davidfowl you need to brush up on your C/C++; that pattern is everywarez! But they also have `long long`s and an `int` of undefined size :disappointed:

Issue #699, opened 2016-03-15
Return unused block from SocketInput/Output when connection is idle 


halter73:
`SocketInput` and `SocketOutput` both have `_head` and `_tail` fields referencing `MemoryPoolBlock`s. In both cases, the bytes between `_head` and `_tail` haven't been consumed.

When all the bytes are fully consumed `_head == _tail` and the `Start` and `End` properties of the block are also equal. Still the fully consumed block is never returned to the `MemoryPool` until the connection is closed. This causes each idle connection to waste 8KB of memory even when the write buffers are completely flushed and the read allocation callback hasn't been called.

SajayAntony:
This works for high high throughput and pipelined requests. I'm trying to understand the memory impact for large set of non-idle connection. Do we need to configure receiveBufferSize and SendBufferSize on these sockets?

taoliseki:
The default value for Socket.ReceiveBufferSize and Socket.SendBufferSize is 8KB. That can explain why we saw 16-17 KB memory usage per connection with Kestrel. 

For now I am not too concerned about the memory usage when the connections are idle. We are more interested to bring down the memory usage per connection (active or not). That's why Sajay mentioned the buffer size config. 

A good buffer size depends on the average usage in real world and I believe 8KB was chosen for a good tradeoff. The scenario in my test was just a hello world MVC app and the payload size for request/response was around 200 bytes. So if we set the buffer size to 1KB, in theory we should be able to reduce the memory usage to 1/8. But we need to get a better knowledge of the real-world scenarios.

taoliseki:
Actually reducing the memory usage for idle connections may be a good idea to reduce memory usage. In my test, I was making each of 60k connection idle for 15 sec to mimic a real scenario (each wcat client sleeps 10 sec between 2 consecutive requests). As a result, at any point of time only ~ 1k connections are active (with a request on the fly) and most of the connections are idle. If we are able to reduce the memory usage to almost 0 for idle connections, then potentially we may reduce the memory usage by 90% for this scenario. There will be more GC load and we need to measure that.

muratg:
Let's consider this post RC2.

PR #1077, opened 2016-09-01
[WIP] Allocate Connection members on first read

cesarbs:
A little experiment I ran.

Right now we allocate a bunch of resources at the time a new connection is established: `SocketInput`, `SocketOuput`, `Frame`, etc. If that connection is never used, that's a bunch of wasted resources. This change waits until the first read to allocate those resources.

Some browsers (e.g. Chrome) might open more connections than they will actually use. For example, when I hit the sample app with Chrome, I see 4 connections, only 2 of which are actually used. If we extrapolate that to thousands of clients, memory usage on the server would be much higher than actually needed.

I tested this with 10,000 idle connections (after warming up with 10,000 active connections to get some slabs allocated). There's a ~17% reduction in memory usage. Results below.

## dev:

image

image

## This change:

image

image

No impact to plaintext benchmark.

**Edit**: updated with 10,000 idle connections opened after warming up the server with 10,000 active connections. Thanks @benaadams for pointing this out.

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Connection.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Connection.cs
@@ -2,7 +2,6 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
-using System.Diagnostics;
 using System.IO;
 using System.Threading;
 using System.Threading.Tasks;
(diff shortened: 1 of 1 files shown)


davidfowl:
The call to StopAsync can happen from a non libuv thread so you need to make sure that things still work if StopAsync happens before the frame starts and after.

benaadams:
Why is the `byte[]` so different between the two?

cesarbs:
@benaadams Memory pool slabs. Good call, this is why the difference is so dramatic. Need to check difference with slabs already allocated.

cesarbs:
A better test: open 10,000 connections, make a request on each, leave them all open. Close all. Force GC. Open 10,000 idle connections. Measure.

## dev:

image

image

## This change:

image

image

halter73:
This was fixed in the more general case by https://github.com/aspnet/KestrelHttpServer/issues/699

PR #1083, opened 2016-09-04
Slimmer locks

benaadams:
Slim some locks in SocketInput

/cc @halter73 I know you like me changing these things :wink:

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/SocketInput.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/SocketInput.cs
@@ -72,6 +72,8 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Http
 
         public void IncomingComplete(int count, Exception error)
         {
+            Action awaitableState;
+
             lock (_sync)
             {
                 // Must call Add() before bytes are available to consumer, to ensure that Length is >= 0
@@ -107,8 +109,13 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Http
                     FinReceived();
                 }
 
-                Complete();
+                awaitableState = Interlocked.Exchange(
+                                    ref _awaitableState,
+                                    _awaitableIsCompleted);
+
             }
+
+            Complete(awaitableState);
         }
 
         public void IncomingDeferred()
@@ -129,12 +136,8 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Http
             }
         }
 
-        private void Complete()
+        private void Complete(Action awaitableState)
         {
-            var awaitableState = Interlocked.Exchange(
-                ref _awaitableState,
-                _awaitableIsCompleted);
-
             _manualResetEvent.Set();
 
             if (!ReferenceEquals(awaitableState, _awaitableIsCompleted) &&
@@ -146,21 +149,29 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Http
 
         public MemoryPoolIterator ConsumingStart()
         {
+            MemoryPoolBlock head;
+            bool isAlreadyConsuming;
+
             lock (_sync)
             {
-                if (_consuming)
-                {
-                    throw new InvalidOperationException("Already consuming input.");
-                }
+                isAlreadyConsuming = _consuming;
+                head = _head;
                 _consuming = true;
-                return new MemoryPoolIterator(_head);
             }
+
+            if (isAlreadyConsuming)
+            {
+                throw new InvalidOperationException("Already consuming input.");
+            }
+
+            return new MemoryPoolIterator(head);
         }
 
         public void ConsumingComplete(
             MemoryPoolIterator consumed,
             MemoryPoolIterator examined)
         {
+            bool isConsuming;
             MemoryPoolBlock returnStart = null;
             MemoryPoolBlock returnEnd = null;
 
@@ -184,7 +195,6 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Http
                         {
                             // Everything has been consumed and no data is being written to the
                             // _tail block, so return all blocks between _head and _tail inclusive.
-                            returnEnd = null;
                             _head = null;
                             _tail = null;
                         }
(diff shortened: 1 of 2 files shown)


src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/SocketInput.cs
```diff
@@ -103,9 +103,9 @@ public void IncomingComplete(int count, Exception error)
                 {
                     FinReceived();
                 }
-
-                Complete();
             }
+
+            Complete();
```
halter73: The call to Complete() here (or at the very least the Interlocked.Exchange call inside), needs to share the same lock as:

                    if (!examined.IsDefault &&
                        examined.IsEnd &&
                        ReadingInput)
                    {
                        _manualResetEvent.Reset();

                        Interlocked.CompareExchange(
                            ref _awaitableState,
                            _awaitableIsNotCompleted,
                            _awaitableIsCompleted);
                    }

in ConsumingComplete(...). Otherwise, the consumer can get stuck forever awaiting data at the end of the response that was already there.

benaadams: Since you are in a lock could just change them to an if statement rather than `Interlocked.CompareExchange`?

benaadams:
@halter73 better?

halter73:
I think the new code is correct, but now that new locks are added where we were using interlocked before, I'm not 100% sure this is a net improvement.

benaadams:
Previously, though it was doing interlocks in the locks as well as method invocation and struct creation; whereas now its just swapping variables - other than `examined.IsEnd` which both can be expensive and is non-inlinable :cry: but I don't think its safe to move. So the locks should be held for a very short time.

The lock when uncontested is just interlocked; when contested spins for a time - however the queue to threadpool also has an interlocked spin and will be more contested than these locks which should only be between two threads; whereas the threadpool global queue is between all threads, so is best to move it out of the lock as it will cause it to be held longer. Equally the block return is between all the threads sharing the memory pool and will spin?

So hopefully it should reduce the lock contention and time spent in cpu spins overall, while adding value to the spins on these locks when they do happen (as the will be acquired fast)?

benaadams:
Also if you think some of them could be changed back to `Interlocked` but outside the locks while maintaining the structure and safety - that would be a bonus...

benaadams:
Threadpool-wise, since all the completion queues are happening on the libuv thread, could it collect them all in a list and then queue them as a single batch item to the threadpool? Either as a batch that runs sequentially on the threadpool or one that unwraps to multiple items and requeues them on the threadpool to distribute across threadpool threads (depending on how many libuv threads you have queuing).

That should alleviate some of the time in the threadpool queue spins.

halter73:
I think `CompleteAwaiting`, `AbortAwaiting` and `OnCompleted` don't need to take the `_sync` lock when modifying the `_awaitableState` using `Interlocked`. That's the way it is currently. It looks ugly to combine Interlocked with normal locks, but I think it's not too unreasonable in this code.

I'm wary of changing this code since it's burned us before. Maybe we could do something to move `_threadPool.Run(awaitableState)` out of the lock. Maybe by returning `awaitableState` if it should be run, and null otherwise. That and calling `ReturnBlocks` outside of the lock seems safe.

I would like some measurements showing removing Interlocked improves perf before going that far.

benaadams:
@halter73 methodology the same now?

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/SocketInput.cs
```diff
@@ -196,30 +207,32 @@ public MemoryPoolIterator ConsumingStart()
...
                 {
                     returnStart = _head;
-                    returnEnd = null;
                     _head = null;
                     _tail = null;
                 }
 
-                ReturnBlocks(returnStart, returnEnd);
-
-                if (!_consuming)
```
halter73: _consuming can be checked and modified outside of the lock. The consumer shouldn't be making concurrent calls anyway.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/SocketInput.cs
```diff
@@ -196,30 +207,32 @@ public MemoryPoolIterator ConsumingStart()
                 else
                 {
                     returnStart = _head;
```
halter73: Maybe add a comment indicating this returns all the blocks.

halter73: Never mind. I forgot we use _consuming in dispose. Good job continuing to checking this in the lock.

halter73:
LGTM

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/SocketInput.cs
benaadams: Added comment. The interplay is not immediately apparent.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/SocketInput.cs
```diff
@@ -195,31 +206,34 @@ public MemoryPoolIterator ConsumingStart()
...
+                    // Dispose won't have returned the blocks if we were consuming, so return them now
                     returnStart = _head;
-                    returnEnd = null;
                     _head = null;
                     _tail = null;
                 }
 
-                ReturnBlocks(returnStart, returnEnd);
-
-                if (!_consuming)
```
halter73: It looks like github lost my comment earlier where I realized that all access to `_consuming` really does need to be guarded by the lock because of the interaction with Dispose.

benaadams:
Adding to the justification https://github.com/dotnet/coreclr/issues/7401
> The lock helpers such as JIT_MonEnter/JIT_MonExit are in the top hotspots in TechEmpower Plaintext scenario (measured on 64 bits Ubuntu 14.04 LTS, Intel® Xeon® Processor E5-2699 v4).

halter73:
@benaadams Can you rebase? I'll merge once @cesarbs and @pakrym take a look.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/SocketInput.cs
```diff
@@ -236,7 +250,7 @@ public void OnCompleted(Action continuation)
 
             if (ReferenceEquals(awaitableState, _awaitableIsNotCompleted))
             {
-                return;
+                // nop;
```
cesarbs: Why not just remove the check?

cesarbs: Change `else` to `else if (!ReferenceEquals(awaitableState, _awaitableIsNotCompleted)`?

benaadams: Changed

benaadams:
Rebased

halter73:
https://github.com/aspnet/KestrelHttpServer/commit/310411075f74a5c26ea197252765b96d9189e0dc

PR #1088, opened 2016-09-07
Remove double initializing of variable _field3

DamirAinullin:
Remove double initializing of variable _field3.
Also I see that variable _field3 is never used. I think there is possible problem with it.
Can you check it please?

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Networking/SockAddr.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Networking/SockAddr.cs
@@ -1,7 +1,6 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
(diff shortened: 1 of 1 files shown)


halter73:
@DamirAinullin The two P/Invokes SockAddr is used for ([uv_tcp_getsockname](http://docs.libuv.org/en/v1.x/tcp.html#c.uv_tcp_getsockname) and [uv_tcp_getpeername](http://docs.libuv.org/en/v1.x/tcp.html#c.uv_tcp_getsockname)) both take a size parameter. The size of the struct passed in needs to be at least large enough to hold the address, but can be larger.

The comments in the SockAddr struct and its usage shows all the address information we consume is always in the first 24 bytes no matter which platform. I don't see a reason to shorten the struct from 32 bytes to 24 bytes unless we're 100% sure it's safe.

Thanks for the PR!

halter73:
@troydai You orignally wrote the SockAddr class. Any insights why you went for a 32bit instead of 24bit struct?

PR #1090, opened 2016-09-07
Fix failing keep-alive timeout tests

cesarbs:
@halter73 @pakrym

### DIFF
--- a/test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/KeepAliveTimeoutTests.cs
+++ b/test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/KeepAliveTimeoutTests.cs
@@ -97,8 +97,12 @@ namespace Microsoft.AspNetCore.Server.Kestrel.FunctionalTests
                         "GET / HTTP/1.1",
                         "",
                         "");
-                    await ReceiveResponse(connection, server.Context);
                     await Task.Delay(ShortDelay);
+                }
+
+                for (var i = 0; i < 10; i++)
+                {
+                    await ReceiveResponse(connection, server.Context);
                 }
             }
         }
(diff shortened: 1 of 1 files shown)


test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/KeepAliveTimeoutTests.cs
```diff
@@ -91,23 +91,27 @@ public class KeepAliveTimeoutTests
         {
             using (var connection = new TestConnection(server.Port))
             {
-                for (var i = 0; i < 10; i++)
+                for (var i = 0; i < 11; i++)
```
pakrym: 11? Why not 10?

benaadams: Goes to 11 https://www.youtube.com/watch?v=KOO5S4vxi0o :wink:

cesarbs: @pakrym Ha, that's not necessary. I had it in my head that the keep-alive timeout I was using for the tests was 30s, but it's 10s. With the short delay being 3s I'd need 11 iterations to make sure I was getting past the timeout.

cesarbs:
Updated.

cesarbs:
Travis failure is unrelated.

cesarbs:
:watch:

halter73:
:shipit:

mikeharder:
@cesarbs: I thought we decided the keepalive should apply to both requests and responses, so the original test is actually better.  Is this fix temporary until the keepalive behavior can be changed?

cesarbs:
That's right, it's temporary. The timeout implementation needs to be revisited, but this keeps the CI good for now.

PR #1094, opened 2016-09-08
Ensure ConsumingComplete() is always called with the furthest examined position (#1095)

cesarbs:
#1095


halter73:
I think we should add a unit test to `FrameTests` verifying that TakeStartLine/TakeMessageHeaders exit given incomplete data.

cesarbs:
Added tests.

halter73:
:shipit:

PR #1097, opened 2016-09-09
Implement CopyToAsync on FrameRequestStream

### LINKED ISSUE #1096: Implement CopyToAsync on FrameRequestStream
So we can avoid double buffering up stack.

benaadams:
Resolves #1096


davidfowl:
Nope, far from actually 😄 . Needs to expose the buffers directly in the memory pool.

benaadams:
And I thought this was messy :laughing: 

Will need to work with the fun logic of message body and its variants https://github.com/aspnet/KestrelHttpServer/blob/dev/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/MessageBody.cs#L24-L37

davidfowl:
Yahp 😄 shouldn;t be too bad though

PR #1098, opened 2016-09-09
Avoid blocking shutdown in tests by calling TrySetResult inline

halter73:
- This allows for a graceful shutdown with dotnet test -Parallel None
- By default, the xunit synccontext will dispatch automatically off
  the KestrelThread, but it's best not to rely on this behavior.
- Fix deadlock in HttpsTests

### DIFF
--- a/test/Microsoft.AspNetCore.Server.KestrelTests/TestHelpers/MockLibuv.cs
+++ b/test/Microsoft.AspNetCore.Server.KestrelTests/TestHelpers/MockLibuv.cs
@@ -69,7 +69,6 @@ namespace Microsoft.AspNetCore.Server.KestrelTests.TestHelpers
                 {
                     _loopWh.Wait();
                     KestrelThreadBlocker.Wait();
-                    TaskCompletionSource<object> onPostTcs = null;
 
                     lock (_postLock)
                     {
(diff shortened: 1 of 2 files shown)


cesarbs:
:rocket: once Travis is finally green

test/Microsoft.AspNetCore.Server.KestrelTests/TestHelpers/MockLibuv.cs
```diff
@@ -84,14 +83,14 @@ unsafe public MockLibuv()
                             // Ensure any subsequent calls to uv_async_send
                             // create a new _onPostTcs to be completed.
                             _completedOnPostTcs = true;
-                            onPostTcs = _onPostTcs;
+
+                            // Calling TrySetResult outside the lock to avoid deadlock
```
benaadams: Not sure this  comment is correct with this change?

halter73: I think the comment is still accurate since the actually call to `TrySetResult` is made without the lock. I added to the end of the comment to indicate that we don't want to block the run loop.

PR #1102, opened 2016-09-12
Improve keep-alive timeout

cesarbs:
* Track time more accurately
* Control timeout in `Connection` instead of `Frame`

@halter73 @mikeharder @davidfowl

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Connection.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Connection.cs
@@ -2,6 +2,7 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.Diagnostics;
 using System.IO;
 using System.Threading;
 using System.Threading.Tasks;
@@ -38,6 +39,9 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Http
         private TaskCompletionSource<object> _socketClosedTcs = new TaskCompletionSource<object>();
         private BufferSizeControl _bufferSizeControl;
 
+        private long _lastTimestamp;
+        private long _timeoutTimestamp = long.MaxValue;
+
         public Connection(ListenerContext context, UvStreamHandle socket) : base(context)
         {
             _socket = socket;
@@ -62,6 +66,8 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Http
             }
 
             _frame = FrameFactory(this);
+
+            _lastTimestamp = Thread.Loop.Now();
         }
 
         // Internal for testing
@@ -156,9 +162,14 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Http
         }
 
         // Called on Libuv thread
-        public void Tick()
+        public void Tick(long timestamp)
         {
-            _frame.Tick();
+            if (timestamp > _timeoutTimestamp)
+            {
+                StopAsync();
+            }
+
+            Interlocked.Exchange(ref _lastTimestamp, timestamp);
         }
 
         private void ApplyConnectionFilter()
@@ -282,9 +293,17 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Http
             }
         }
 
-        void IConnectionControl.Stop()
+        void IConnectionControl.SetTimeout(long milliseconds)
         {
-            StopAsync();
+            Debug.Assert(_timeoutTimestamp == long.MaxValue, "Concurrent timeouts are not supported");
+
+            // Add KestrelThread.HeartbeatMilliseconds extra milliseconds since this can be called right before the next heartbeat.
+            Interlocked.Exchange(ref _timeoutTimestamp, _lastTimestamp + milliseconds + KestrelThread.HeartbeatMilliseconds);
+        }
+
+        void IConnectionControl.CancelTimeout()
+        {
+            Interlocked.Exchange(ref _timeoutTimestamp, long.MaxValue);
         }
 
         private static unsafe string GenerateConnectionId(long id)
--- a/src/Microsoft.AspNetCore.Server.Kestrel/KestrelServerLimits.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/KestrelServerLimits.cs
@@ -23,8 +23,8 @@ namespace Microsoft.AspNetCore.Server.Kestrel
         // Matches the default LimitRequestFields in Apache httpd.
         private int _maxRequestHeaderCount = 100;
 
-        // Matches the default http.sys keep-alive timouet.
-        private TimeSpan _keepAliveTimeout = TimeSpan.FromMinutes(2);
+        // Matches the default http.sys connection timeout.
+        private TimeSpan _connectionTimeout = TimeSpan.FromMinutes(2);
 
         /// <summary>
         /// Gets or sets the maximum size of the response buffer before write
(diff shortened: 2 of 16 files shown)


src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Connection.cs
```diff
@@ -158,7 +161,15 @@ public virtual void OnSocketClosed()
...
         {
-            _frame.Tick();
+            if (_timeoutEnabled)
+            {
+                if (_secondsSinceLastConnectionEvent > ServerOptions.Limits.ConnectionTimeout.TotalSeconds)
+                {
+                    StopAsync();
+                }
+
+                _secondsSinceLastConnectionEvent++;
```
mikeharder: This relies on `Tick()` being called exactly once per second, which seems fragile.  Have you compared the perf of something like `System.Diagnostics.Stopwatch()`, which would basically track the precise time of the last connection event, rather than an integer number of seconds?

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/SocketOutput.cs
```diff
@@ -192,6 +192,10 @@ public class SocketOutput : ISocketOutput
                 {
                     _postingWrite = true;
                     _ongoingWrites++;
+                    // We don't know how long it will take to complete the write, but we don't want to
+                    // close the connection knowing there is a pending/ongoing write, since we know there will
+                    // be connection activity. Hence, disable the timeout until there are no ongoing writes.
+                    _connection.ConnectionControl.DisableTimeout();
```
mikeharder: Why disable the timeout when writing, rather than just resetting the timeout every time bytes are written?

cesarbs: Because a single write can be arbitrarily long. This is why I re-enable the timeout once all writes are completed.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Connection.cs
cesarbs: It hadn't occurred to me to use a `Stopwatch` to track the time between ticks - I'll try that out. When I first wrote this I did try 3 different ways to track time more accurately but they all had significant perf costs.

cesarbs: Looking at how `Stopwatch` is implemented (https://github.com/dotnet/corefx/blob/master/src/System.Runtime.Extensions/src/System/Diagnostics/Stopwatch.cs), I think it'll have the same perf costs as the other stuff I did try before.

mikeharder: In this case I would make a few changes:

* Track the time between connection events as milliseconds rather than seconds.
* Make the millseconds between ticks public rather than private (`KestrelThread._heartbeatMilliseconds`).
* In `Tick()`, increment the elapsed milliseconds by the public constant.

cesarbs:
I must have gotten it right this time, because there's virtually no perf impact with my time tracking changes (494ffde) :grin: 

Averages of five runs of dev vs this change:

dev: 1130819.484 RPS
this PR: 1121710.11 RPS

cesarbs:
@Tratcher

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/SocketOutput.cs
halter73: Can we move this call outside the lock and inside `if (scheduleWrite)` below. Not that this is an expensive call, but why not?

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Connection.cs
```diff
@@ -156,9 +164,19 @@ public virtual void OnSocketClosed()
         }
 
         // Called on Libuv thread
-        public void Tick()
+        public void Tick(long timestamp)
         {
-            _frame.Tick();
+            if (_timeoutEnabled)
+            {
+                if (_millisecondsSinceLastConnectionEvent > ServerOptions.Limits.ConnectionTimeout.TotalMilliseconds)
```
halter73: Store this so ms doesn't need to be recalculated. You could also add a second to the configuration since `_millisecondsSinceLastConnectionEvent` could be up to a second too large.

src/Microsoft.AspNetCore.Server.Kestrel/KestrelServerLimits.cs
```diff
@@ -143,20 +143,20 @@ public int MaxRequestHeaderCount
...
 
         /// <summary>
-        /// Gets or sets the keep-alive timeout.
+        /// Gets or sets the connection timeout.
         /// </summary>
         /// <remarks>
         /// Defaults to 2 minutes. Timeout granularity is in seconds. Sub-second values will be rounded to the next second.
         /// </remarks>
-        public TimeSpan KeepAliveTimeout
+        public TimeSpan ConnectionTimeout
```
halter73: I would rather keep this a timeout for only time spent between request. Having this timeout trigger because an app takes a long time doing business logic before writing a response or completing seems bad to me (even if it would never kill the connection mid request or response).

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Connection.cs
```diff
@@ -282,9 +292,15 @@ void IConnectionControl.End(ProduceEndType endType)
             }
         }
 
-        void IConnectionControl.Stop()
+        void IConnectionControl.SetTimeout(long milliseconds)
+        {
+            // Add KestrelThread.HeartbeatMilliseconds extra milliseconds since this can be called right before the next heartbeat.
+            Interlocked.Exchange(ref _timeoutTimestamp, _lastTimestamp + milliseconds + KestrelThread.HeartbeatMilliseconds);
```
halter73: Can we have a Debug.Assert or something verifying that the _timeoutTimestamp was long.MaxValue previously?

test/Microsoft.AspNetCore.Server.KestrelTests/ConnectionTests.cs
```diff
@@ -30,14 +34,19 @@ public void DoesNotEndConnectionOnZeroRead()
...
-                var socket = new MockSocket(mockLibuv, Thread.CurrentThread.ManagedThreadId, trace);
-                var connection = new Connection(context, socket);
-                connection.Start();
 
-                Libuv.uv_buf_t ignored;
-                mockLibuv.AllocCallback(socket.InternalGetHandle(), 2048, out ignored);
-                mockLibuv.ReadCallback(socket.InternalGetHandle(), 0, ref ignored);
-                Assert.False(connection.SocketInput.CheckFinOrThrow());
+                Connection connection = null;
+                await context.Thread.PostAsync(_ =>
```
halter73: Why does this need to be changed now?

halter73:
I like this sooo much better. Nice :+1:

test/Microsoft.AspNetCore.Server.KestrelTests/ConnectionTests.cs
cesarbs: Because `Connection`'s ctor calls `UvLoopHandle.Now()`, so it needs to run in the same thread the loop was created.

halter73: Ahhh

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Connection.cs
cesarbs: Done.

PR #1105, opened 2016-09-13
Copyless FrameRequestStream.CopyToAsync(...)

halter73:
I'm going to add a test for behavior when `destinationstream.WriteAsync(...)` throws.

There are already quite a few tests that call `FrameRequestStream.CopyToAsync()`, but have been using the default implementation up to this point.

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/FrameRequestStream.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/FrameRequestStream.cs
@@ -119,8 +119,26 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Http
             var task = ValidateState(cancellationToken);
             if (task == null)
             {
-                // Needs .AsTask to match Stream's Async method return types
-                return _body.ReadAsync(new ArraySegment<byte>(buffer, offset, count), cancellationToken).AsTask();
+                return _body.ReadAsync(new ArraySegment<byte>(buffer, offset, count), cancellationToken);
+            }
+            return task;
+        }
+
+        public override Task CopyToAsync(Stream destination, int bufferSize, CancellationToken cancellationToken)
+        {
+            if (destination == null)
+            {
+                throw new ArgumentNullException(nameof(destination));
+            }
+            if (bufferSize <= 0)
+            {
+                throw new ArgumentException($"{nameof(bufferSize)} must be positive.", nameof(bufferSize));
+            }
+
+            var task = ValidateState(cancellationToken);
+            if (task == null)
+            {
+                return _body.CopyToAsync(destination, cancellationToken);
             }
             return task;
         }
--- a/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/MessageBody.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/MessageBody.cs
@@ -2,9 +2,11 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.IO;
 using System.Numerics;
 using System.Threading;
 using System.Threading.Tasks;
+using Microsoft.AspNetCore.Server.Kestrel.Internal.Infrastructure;
 using Microsoft.Extensions.Internal;
 
 namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Http
@@ -12,7 +14,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Http
     public abstract class MessageBody
     {
         private readonly Frame _context;
-        private int _send100Continue = 1;
+        private bool _send100Continue = true;
 
         protected MessageBody(Frame context)
         {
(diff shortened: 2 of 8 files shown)


src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/FrameRequestStream.cs
```diff
@@ -119,8 +119,19 @@ public override Task<int> ReadAsync(byte[] buffer, int offset, int count, Cancel
...
-                return _body.ReadAsync(new ArraySegment<byte>(buffer, offset, count), cancellationToken).AsTask();
+                return _body.ReadAsync(new ArraySegment<byte>(buffer, offset, count), cancellationToken);
+            }
+            return task;
+        }
+
+        public override Task CopyToAsync(Stream destination, int bufferSize, CancellationToken cancellationToken)
+        {
+            // REVIEW: Is bufferSize a minimum, maximum or suggestion? Should we validate that it's <= 4096 or >= 4096?
+
```
davidfowl: Let's treat it like a maximum and cap it at 4096.

halter73: The default is 81920, so I don't think we can realistically reject calls with a bufferSize > 4096.

davidfowl: Cap = Math.Max(4096, bufferSize)

benaadams: Check its `> 0`? Though it will have no effect if its wrong anyway...

halter73: What does `Cap = Math.Max(4096, bufferSize)` do?

I think we should just ignore bufferSize since we will never call WriteAsync with more than 4096 bytes, and I don't feel the need to segment blocks into even smaller chunks unnecessarily.

We can verify it `> 0` 'cause, why not?

halter73: Maybe we could verify that bufferSize is `>= 4096` instead...

benaadams: You don't actually do anything with the parameter though?

So I think it make sense to either check its not nonsensical (`>0`) or do nothing.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/MessageBody.cs
```diff
@@ -21,80 +23,199 @@ protected MessageBody(Frame context)
...
                 {
-                    // Completed Task, end of stream
                     return TaskCache.CompletedTask;
                 }
 
-            } while (true);
+                Task destinationTask;
+                try
+                {
+                    destinationTask = destination.WriteAsync(segment.Array, segment.Offset, segment.Count, cancellationToken);
```
cesarbs: You should add a test that verifies this call. Call `CopyToAsync` on some mock stream and check that the address of the array is the same as what `SocketInput.PeekAsync()` gives you.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Infrastructure/MemoryPoolIteratorExtensions.cs
```diff
@@ -232,6 +232,32 @@ public static ArraySegment<byte> GetArraySegment(this MemoryPoolIterator start,
             return new ArraySegment<byte>(array, 0, length);
         }
 
+        public static ArraySegment<byte> PeekArraySegment(this MemoryPoolIterator iter)
```
cesarbs: Add unit tests to check the expected result for each condition in this method.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Infrastructure/MemoryPoolIteratorExtensions.cs
```diff
@@ -232,6 +232,32 @@ public static ArraySegment<byte> GetArraySegment(this MemoryPoolIterator start,
...
+            {
+                if (block.Start < block.End)
+                {
+                    return new ArraySegment<byte>(block.Array, block.Start, block.End - block.Start);
+                }
+                block = block.Next;
+            }
+
+            // The following should be unreachable due to the IsEnd check above
+            throw new InvalidOperationException("This should be unreachable!");
```
halter73: Do you want a shorter comment?

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/MessageBody.cs
```diff
@@ -164,60 +285,68 @@ public ForContentLength(bool keepAlive, long contentLength, Frame context)
...
+                    return new ArraySegment<byte>(segment.Array, segment.Offset, (int)_inputLength);
                 }
+            }
+
+            protected override void ConsumedBytes(int count)
+            {
+                _inputLength -= count;
 
-                return actual;
+                base.ConsumedBytes(count);
```
halter73: I don't see the problem with putting implementation code in the base virtual method, but I'm not attached to it.

What do you suggest naming the abstract method called by ConsumedBytes()? OnConsumedBytes()?

cesarbs: `OnConsumedBytes` sounds good.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Infrastructure/MemoryPoolIteratorExtensions.cs
cesarbs: No, I was just remarking that this is the classic "can't happen" scenario (http://www.catb.org/jargon/html/C/can-t-happen.html) 😄

halter73:
@cesarbs

mikeharder:
Is this supposed to improve perf?  For which scenario(s), and what are the before/after measurements?

test/Microsoft.AspNetCore.Server.KestrelTests/FrameRequestStreamTests.cs
```diff
@@ -126,5 +127,61 @@ public void AbortWithErrorCausesReadToCancel()
             Assert.True(task.IsFaulted);
             Assert.Same(error, task.Exception.InnerException);
         }
+
+        [Fact]
+        public void StopAcceptingReadsCausesReadToThrowObjectDisposedExcpetion()
```
cesarbs: Excpetion

test/Microsoft.AspNetCore.Server.KestrelTests/FrameRequestStreamTests.cs
```diff
@@ -126,5 +127,61 @@ public void AbortWithErrorCausesReadToCancel()
...
+        [Fact]
+        public void NullDestinationCausesCopyToAsyncToThrowArgumentNullException()
+        {
+            var stream = new FrameRequestStream();
+            stream.StartAcceptingReads(null);
+            Assert.Throws<ArgumentNullException>(() => { stream.CopyToAsync(null); });
+        }
+
+        [Fact]
+        public void ZeroBufferSizeCausesCopyToAsyncToThrowArgementException()
```
cesarbs: Argement

test/Microsoft.AspNetCore.Server.KestrelTests/MessageBodyTests.cs
```diff
@@ -68,23 +76,120 @@ public void Http10ConnectionClose()
...
+                })
+                .Returns(TaskCache.CompletedTask);
+
+            using (var input = new TestInput())
+            {
+                var body = MessageBody.For(HttpVersion.Http11, headers, input.FrameContext);
+
+                var copyToAsyncTask = body.CopyToAsync(mockDestination.Object);
+
+                // The block returned by IncomingStart always has at leas 2048 available bytes,
```
cesarbs: leas

test/Microsoft.AspNetCore.Server.KestrelTests/MessageBodyTests.cs
```diff
@@ -68,23 +76,120 @@ public void Http10ConnectionClose()
...
+                input.Add("Hello", fin: true);
+
+                var ms = new MemoryStream();
+
+                await stream.CopyToAsync(ms);
+                var requestArray = ms.ToArray();
+                Assert.Equal(8197, requestArray.Length);
+                AssertASCII(largeInput + "Hello", new ArraySegment<byte>(requestArray, 0, requestArray.Length));
+
+                var readBuffer = new byte[1];
```
cesarbs: You could just inline the `new byte[1]` in the `ReadAsync`call.

test/Microsoft.AspNetCore.Server.KestrelTests/MessageBodyTests.cs
```diff
@@ -68,23 +76,120 @@ public void Http10ConnectionClose()
...
+                writeTcs = new TaskCompletionSource<byte[]>();
+                bytes = Encoding.ASCII.GetBytes(data[1]);
+                block = socketInput.IncomingStart();
+                Buffer.BlockCopy(bytes, 0, block.Array, block.End, bytes.Length);
+                socketInput.IncomingComplete(bytes.Length, null);
 
-                var count2 = await stream.ReadAsync(readBuffer, 0, 8192);
-                Assert.Equal(5, count2);
-                AssertASCII("Hello", new ArraySegment<byte>(readBuffer, 0, 5));
+                Assert.Same(block.Array, await writeTcs.Task);
```
cesarbs: Nice 👍

test/Microsoft.AspNetCore.Server.KestrelTests/MessageBodyTests.cs
```diff
@@ -68,23 +76,120 @@ public void Http10ConnectionClose()
...
+        };
+
+        public static IEnumerable<object[]> CombinedData => 
+            from stream in StreamData
+            from request in RequestData
+            select new[] { stream[0], request[0], request[1] };
+
+        [Theory]
+        [MemberData(nameof(RequestData))]
+        public async Task CopyToAsyncDoesntCopyBlocks(FrameRequestHeaders headers, string[] data)
```
cesarbs: Major nit: IMO "DoesNot" reads better in test names.

test/Microsoft.AspNetCore.Server.KestrelTests/SocketInputTests.cs
```diff
@@ -154,6 +154,80 @@ public void ConsumingOutOfOrderFailsGracefully()
...
+                scan.Skip(2);
+                socketInput.ConsumingComplete(scan, scan);
+
+                // Everything has been consume so socketInput is no longer in the completed state
+                Assert.False(socketInput.IsCompleted);
+            }
+        }
+
+        [Fact]
+        public async Task CompleteAwaitingDoesntCause0LengthRead()
```
cesarbs: Same comment about DoestNot vs Doesnt.

test/Microsoft.AspNetCore.Server.KestrelTests/SocketInputTests.cs
```diff
@@ -154,6 +154,80 @@ public void ConsumingOutOfOrderFailsGracefully()
...
+                socketInput.CompleteAwaiting();
+                Assert.False(readTask.IsCompleted);
+
+                socketInput.IncomingData(new byte[5], 0, 5);
+                Assert.Equal(5, await readTask);
+            }
+        }
+
+        [Fact]
+        public async Task CompleteAwaitingDoesntCause0LengthPeek()
```
cesarbs: Same comments about "DoesNot" vs "Doesnt" and "Zero" vs "0".

cesarbs:
LGTM. I've left some comments, but it's all nits. Address those then 🚀

davidfowl:
@mikeharder 

> Is this supposed to improve perf? For which scenario(s), and what are the before/after measurements?

This change in itself won't improve the performance. Upstream callers will now be able take advantage of it to reduce allocations and double buffering in some of our core components.

mikeharder:
@davidfowl 

> Upstream callers will now be able take advantage of it to reduce allocations and double buffering in some of our core components.

We should benchmark this scenario then.  I can write the benchmark if you can explain a bit more what the upstream caller needs to do.

davidfowl:
> We should benchmark this scenario then. I can write the benchmark if you can explain a bit more what the upstream caller needs to do.

We'll need to rewrite some codez. But the first target is https://github.com/aspnet/HttpAbstractions/blob/dev/src/Microsoft.AspNetCore.WebUtilities/FormReader.cs

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/FrameRequestStream.cs
```diff
@@ -126,7 +126,14 @@ public override Task<int> ReadAsync(byte[] buffer, int offset, int count, Cancel
...
         public override Task CopyToAsync(Stream destination, int bufferSize, CancellationToken cancellationToken)
         {
-            // REVIEW: Is bufferSize a minimum, maximum or suggestion? Should we validate that it's <= 4096 or >= 4096?
+            if (destination == null)
+            {
+                throw new ArgumentNullException(nameof(destination));
+            }
+            if (bufferSize <= 0)
+            {
+                throw new ArgumentException("bufferSize must be positive.", nameof(bufferSize));
```
cesarbs: nit: use `nameof` in message too.

cesarbs:
New GitHub review stuff makes things super confusing to read.

Added a minor nit on a `nameof`, then :shipit:

PR #1111, opened 2016-09-16
Add request headers timeout (#1110)

cesarbs:
#1110

- [x] Request headers timeout 
- [x] Abort or 408 instead of graceful close

@halter73 @mikeharder @Tratcher @blowdart

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
@@ -71,6 +71,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Http
         private int _requestHeadersParsed;
 
         protected readonly long _keepAliveMilliseconds;
+        private readonly long _requestHeadersTimeoutMilliseconds;
 
         public Frame(ConnectionContext context)
         {
@@ -84,6 +85,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Http
 
             FrameControl = this;
             _keepAliveMilliseconds = (long)ServerOptions.Limits.KeepAliveTimeout.TotalMilliseconds;
+            _requestHeadersTimeoutMilliseconds = (long)ServerOptions.Limits.RequestHeadersTimeout.TotalMilliseconds;
         }
 
         public ConnectionContext ConnectionContext { get; }
@@ -372,6 +374,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Http
             {
                 _requestProcessingStopping = true;
             }
+
             return _requestProcessingTask ?? TaskCache.CompletedTask;
         }
 
@@ -648,7 +651,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Http
 
         protected Task TryProduceInvalidRequestResponse()
         {
-            if (_requestProcessingStatus == RequestProcessingStatus.RequestStarted && _requestRejected)
+            if (_requestRejected)
             {
                 if (FrameRequestHeaders == null || FrameResponseHeaders == null)
                 {
@@ -833,7 +836,10 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Http
                     return RequestLineStatus.Empty;
                 }
 
-                ConnectionControl.CancelTimeout();
+                if (_requestProcessingStatus == RequestProcessingStatus.RequestPending)
+                {
+                    ConnectionControl.ResetTimeout(_requestHeadersTimeoutMilliseconds, TimeoutAction.SendTimeoutResponse);
+                }
 
                 _requestProcessingStatus = RequestProcessingStatus.RequestStarted;
 
@@ -1102,6 +1108,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Http
                         }
                         else if (ch == '\n')
                         {
+                            ConnectionControl.CancelTimeout();
                             consumed = end;
                             return true;
                         }
@@ -1274,12 +1281,23 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Http
             throw ex;
         }
 
+        public void SetBadRequestState(RequestRejectionReason reason)
+        {
+            SetBadRequestState(BadHttpRequestException.GetException(reason));
+        }
+
         public void SetBadRequestState(BadHttpRequestException ex)
         {
-            StatusCode = ex.StatusCode;
+            // Setting status code will throw if response has already started
+            if (!HasResponseStarted)
+            {
+                StatusCode = ex.StatusCode;
+            }
+
             _keepAlive = false;
             _requestProcessingStopping = true;
             _requestRejected = true;
+
             Log.ConnectionBadRequest(ConnectionId, ex);
         }
 
--- a/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/FrameOfT.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/FrameOfT.cs
@@ -32,7 +32,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Http
             {
                 while (!_requestProcessingStopping)
                 {
-                    ConnectionControl.SetTimeout(_keepAliveMilliseconds);
+                    ConnectionControl.SetTimeout(_keepAliveMilliseconds, TimeoutAction.CloseConnection);
 
                     while (!_requestProcessingStopping && TakeStartLine(SocketInput) != RequestLineStatus.Done)
                     {
@@ -141,7 +141,11 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Http
                         }
                     }
 
-                    Reset();
+                    // Don't lose request rejection state
+                    if (!_requestRejected)
+                    {
+                        Reset();
+                    }
                 }
             }
             catch (BadHttpRequestException ex)
--- a/src/Microsoft.AspNetCore.Server.Kestrel/KestrelServerLimits.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/KestrelServerLimits.cs
@@ -23,8 +23,10 @@ namespace Microsoft.AspNetCore.Server.Kestrel
         // Matches the default LimitRequestFields in Apache httpd.
         private int _maxRequestHeaderCount = 100;
 
-        // Matches the default http.sys connection timeout.
-        private TimeSpan _connectionTimeout = TimeSpan.FromMinutes(2);
+        // Matches the default http.sys connectionTimeout.
+        private TimeSpan _keepAliveTimeout = TimeSpan.FromMinutes(2);
+
+        private TimeSpan _requestHeadersTimeout = TimeSpan.FromSeconds(30);
 
         /// <summary>
         /// Gets or sets the maximum size of the response buffer before write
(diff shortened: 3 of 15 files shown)


test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/RequestHeadersTimeoutTests.cs
```diff
@@ -0,0 +1,125 @@
...
+        private static readonly TimeSpan LongDelay = TimeSpan.FromSeconds(30);
+        private static readonly TimeSpan ShortDelay = TimeSpan.FromSeconds(LongDelay.TotalSeconds / 10);
+
+
+        [Fact]
+        public async Task TestRequestHeadersTimeout()
+        {
+            using (var server = CreateServer())
+            {
+                var tasks = new[]
```
mikeharder: I assume the purpose is to parallelize multiple tests in the same class, which xUnit doesn't support.  If so, I would split these into multiple classes (perhaps with a common base class) and let xUnit handle the parallelism, rather than explicitly running them in parallel this way.

cesarbs: The only issue I have with this suggestion is that they would run sequentially in OS X, since we've disabled parallelism there. These are long running tests and they can all run against the same server/app, that's why I wrote these tests this way (same with keep-alive tests).

test/Microsoft.AspNetCore.Server.KestrelTests/FrameTests.cs
```diff
@@ -29,6 +29,7 @@ public void CanReadHeaderValueWithoutLeadingWhitespace()
             {
                 var connectionContext = new ConnectionContext()
                 {
+                    ConnectionControl = new Mock<IConnectionControl>().Object,
```
cesarbs: Cool, thanks. I've actually just learned about `Mock.Of<T>` while reading @halter73's PR (#1105) 😄

test/shared/TestConnection.cs
```diff
@@ -139,6 +139,22 @@ public void Dispose()
...
+            var eventArgs = new SocketAsyncEventArgs();
+            eventArgs.SetBuffer(new byte[1], 0, 1);
+            eventArgs.Completed += (s, e) =>
+            {
+                if (e.BytesTransferred == 0)
+                {
+                    tcs.SetResult(null);
+                }
+            };
+            _socket.ReceiveAsync(eventArgs);
```
cesarbs: Good call, totally overlooked that.

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/RequestHeadersTimeoutTests.cs
mikeharder: The current implementation is OK.  It's unfortunate that xUnit doesn't make it easier to execute test methods on a single class in parallel.  Either workaround is ugly.

cesarbs:
I had thought of sending a 408 on timeout, but given that other server (e.g. nginx) don't do it and just close the connection, I don't think it's worth the added complexity.

I call this done unless someone disagrees.

cesarbs:
Updated to send 408 on headers timeout.

I had to add some code to make sure we only send 408 on this timeout but not on keep-alive (I don't think it makes much sense to send a response where no request started to come in). I think there are two approaches to this: the one in this PR, or adding a `sendTimeoutResponse` parameter to `Connection.StopAsync()` and `Frame.Stop()`. I'd be glad to hear your thoughts on this.

We also have the option to simply close the connection (nginx actually does that) and not send a response at all. That would avoid the added complexity.

benaadams:
I was wondering if it should send the more detailed exception message as the status on localhost (for all the errors)?

However; can that be detected vs going via a localhost proxy? (ngnix/IIS) As you want to give the more generic message/error codes then.

benaadams:
^--- is obv more in relation to https://github.com/aspnet/KestrelHttpServer/pull/1115 and can open separate issue for it; if you think the distinction is possible to make?

cesarbs:
@benaadams Agreed that it's a separate issue. Not sure we'd like to do that though. Why send the details in the response if you can look at the logs?

benaadams:
> Why send the details in the response if you can look at the logs?

Bit like the `UseDeveloperExceptionPage`; but this would occur before that and it would never run. Not strongly attached to the idea - would likely have too much risk of going wrong and exposing more detail errors than is should.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
```diff
@@ -616,7 +619,7 @@ private void ProduceStart(bool appCompleted)
 
         protected Task TryProduceInvalidRequestResponse()
         {
-            if (_requestProcessingStatus == RequestProcessingStatus.RequestStarted && _requestRejected)
```
halter73: Why did you change this? I'm surprised this hasn't caused any tests to fail.

This change could cause two 4XX responses to be written for a single request. Here's the scenario:

1. There's an invalid request body (e.g. BadChunkSuffix)
2. The app swallows the exception from ReadAsync and completes successfully without writing to the response body.
3. ProduceEnd() is called inside the while loop causing a 4XX to be written.
4. Reset() is called inside the while loop, resetting _requestProcessingStatus to RequestPending
5. The while loop exits because SetBadRequestState sets _requestProccessingStopping to true.
6. TryProduceInvalidRequestResponse() is called in the outer finally of RequestProcessingAsync() causing a 4XX to be written *again*.

Can you add a test to verify this isn't regressed and change it back?

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
```diff
@@ -1242,12 +1249,23 @@ public void RejectRequest(RequestRejectionReason reason, string value)
...
 
+        public void SetBadRequestState(RequestRejectionReason reason)
+        {
+            SetBadRequestState(BadHttpRequestException.GetException(reason));
+        }
+
         public void SetBadRequestState(BadHttpRequestException ex)
         {
-            StatusCode = ex.StatusCode;
+            // Setting status code will throw if response has already started
```
halter73: Yikes. Good thing you caught this now. That could affect bad request bodies too.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/FrameOfT.cs
```diff
@@ -141,7 +141,11 @@ public class Frame<TContext> : Frame
                         }
                     }
 
-                    Reset();
+                    // Don't lose request rejection state
+                    if (!_requestRejected)
+                    {
+                        Reset();
```
halter73: Now I see why my above comment was wrong. I'll keep it there in case there isn't a test for it yet. If there was a test that failed that caused you to add this if, just mention it.

src/Microsoft.AspNetCore.Server.Kestrel/KestrelServerLimits.cs
```diff
@@ -23,8 +23,10 @@ public class KestrelServerLimits
         // Matches the default LimitRequestFields in Apache httpd.
         private int _maxRequestHeaderCount = 100;
 
-        // Matches the default http.sys connection timeout.
-        private TimeSpan _connectionTimeout = TimeSpan.FromMinutes(2);
+        // Matches the default http.sys connectionTimeout.
+        private TimeSpan _keepAliveTimeout = TimeSpan.FromMinutes(2);
+
+        private TimeSpan _requestHeadersTimeout = TimeSpan.FromSeconds(30);
```
halter73: Where did we take this default from?

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/RequestHeadersTimeoutTests.cs
```diff
@@ -0,0 +1,136 @@
...
+
+namespace Microsoft.AspNetCore.Server.Kestrel.FunctionalTests
+{
+    public class RequestHeadersTimeoutTests
+    {
+        private static readonly TimeSpan RequestHeadersTimeout = TimeSpan.FromSeconds(10);
+        private static readonly TimeSpan LongDelay = TimeSpan.FromSeconds(30);
+        private static readonly TimeSpan ShortDelay = TimeSpan.FromSeconds(LongDelay.TotalSeconds / 10);
+
+
```
halter73: nit: extra newline

test/shared/TestConnection.cs
```diff
@@ -139,6 +139,31 @@ public void Dispose()
             }
         }
 
+        public Task WaitForConnectionClose()
```
halter73: What calls this?

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/FrameOfT.cs
cesarbs: Only my new tests failed because `Reset()` was dropping the error code. You raised a good point; I'll add a test for the scenario in your other comment.

src/Microsoft.AspNetCore.Server.Kestrel/KestrelServerLimits.cs
cesarbs: It's the midpoint for the header timeout part of Apache's `RequestReadTimeout`. Their default is 20-40 seconds (it increases if data is coming at a minimum rate). nginx uses 60s.

test/shared/TestConnection.cs
cesarbs: New tests where calling it before I added the 408 code. Right now nothing is calling it, but I'll change `KeepAliveTimeoutTests` to use it because it's cleaner than checking for an exception on a send/receive attempt after the timeout.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
cesarbs: I've added the test, but this wouldn't happen even without the change to `FrameOfT`. `SetBadRequestState` sets `_keepAlive` to `false`, which means the loop would exit here:

```
if (!_keepAlive)
{
    // End the connection for non keep alive as data incoming may have been thrown off
    return;
}
```

So `Reset()` would not get called in that situation.

The reason I had to add

```
// Don't lose request rejection state
if (!_requestRejected)
{
    Reset();
}
```

to `FrameOfT` is that the timeout happens before the app starts, and we never hit the check above.

cesarbs:
@halter73 I've addressed your comments.

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/KeepAliveTimeoutTests.cs
```diff
@@ -50,17 +49,7 @@ public class KeepAliveTimeoutTests
...
-
-                await Assert.ThrowsAsync<IOException>(async () =>
-                {
-                    await connection.Send(
-                        "GET / HTTP/1.1",
-                        "",
-                        "");
-                    await ReceiveResponse(connection, server.Context);
-                });
+                await connection.WaitForConnectionClose().TimeoutAfter(LongDelay);
```
halter73: Why were we calling Send before the last ReceiveResponse before. Is that needed to observe the connection close faster?

halter73:
:shipit:

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/KeepAliveTimeoutTests.cs
cesarbs: `TestConnection.Receive` would randomly hang without a `Send` first. I think it has to do with socket state. Actually when I added `Send`, it would throw there and never get to the subsequent `Receive`. I could have removed the second receive. But just waiting for the FIN is way cleaner anyways.

Issue #653, opened 2016-02-23
Return different status codes depending on condition


benaadams:
Surface different error types, possibles

400 Bad Request + close connection? https://github.com/aspnet/KestrelHttpServer/issues/181 (Resolved by https://github.com/aspnet/KestrelHttpServer/pull/619)
408 Request Timeout (Resolved by https://github.com/aspnet/KestrelHttpServer/pull/485)
413 Payload Too Large https://github.com/aspnet/KestrelHttpServer/issues/478
414 URI Too Long / 431 Request Header Fields Too Large https://github.com/aspnet/KestrelHttpServer/issues/475

muratg:
This will partially be covered by @cesarbs's other assigned items

cesarbs:
@muratg I'll actually do those changes first and then work on this one.

PR #1115, opened 2016-09-21
More specific response status codes for errors (#653)

cesarbs:
#653 

@halter73 @mikeharder @Tratcher @davidfowl


cesarbs:
Want to get this in before https://github.com/aspnet/KestrelHttpServer/pull/1111 so I can return 408 on timeouts.

halter73:
So the new status codes are 414, 431 and 505. Is that it? It would be nice to include that in the commit message.

cesarbs:
@halter73 Updated commit message.

cesarbs:
Since I had to touch them, I've changed `MaxRequestLineSizeTests` and `RequestHeaderLimitsTests` to use `TestServer`.

Issue #1116, opened 2016-09-21
Poor log message when trying to connect with https to an http endpoint


Tratcher:
If someone sends an https request to an http endpoint it fails as expected, but the log message does not help them understand why: `"No space character found after method in request line."`

This was found when a site had a redirect to https rule but the site had not yet set up https.

Proposal: If the request line fails to parse consider checking if it has the format of an SSL/TLS handshake and provide a more specific error like `'An invalid SSL/TLS handshake was attempted on an HTTP endpoint."`

@mikaelm12 @muratg @danroth27

davidfowl:
Yea I don't think we should do this as annoying as that error message is. What do other servers do?

cesarbs:
This should be a matter of checking if the first byte in the request is 0x16. Or am I missing something?

Tratcher:
Something like that, but it may vary across versions.

mikeharder:
Is there an RFC which specifies that the first byte of an HTTPS request will be 0x16?

cesarbs:
https://tools.ietf.org/html/rfc5246#page-39

cesarbs:
https://tools.ietf.org/html/rfc5246#section-6.2.1

khellang:
Or, a more visual breakdown of the handshake TLS record; https://en.wikipedia.org/wiki/Transport_Layer_Security#Handshake_protocol

Tratcher:
RFC links make good code comments

mikeharder:
Also listed here: http://www.iana.org/assignments/tls-parameters/tls-parameters.xhtml#tls-parameters-5

PR #1119, opened 2016-09-22
Detect and log TLS handshake attempts on HTTP endpoints (#1116)

cesarbs:
#1116

@Tratcher @halter73 @mikeharder @mikaelm12 @muratg @danroth27

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Kestrel/BadHttpRequestException.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/BadHttpRequestException.cs
@@ -103,7 +103,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel
                     ex = new BadHttpRequestException("Request contains too many headers.", 431);
                     break;
                 case RequestRejectionReason.HttpsOnHttp:
-                    ex = new BadHttpRequestException("An SSL/TLS handshake might have been attempted at an HTTP endpoint.", 400);
+                    ex = new BadHttpRequestException("An SSL/TLS handshake might have been attempted on an HTTP endpoint.", 400);
                     break;
                 default:
                     ex = new BadHttpRequestException("Bad request.", 400);
--- a/src/Microsoft.AspNetCore.Server.Kestrel/BadHttpRequestException.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/BadHttpRequestException.cs
@@ -102,6 +102,9 @@ namespace Microsoft.AspNetCore.Server.Kestrel
                 case RequestRejectionReason.TooManyHeaders:
                     ex = new BadHttpRequestException("Request contains too many headers.", 431);
                     break;
+                case RequestRejectionReason.HttpsOnHttp:
+                    ex = new BadHttpRequestException("An SSL/TLS handshake might have been attempted at an HTTP endpoint.", 400);
+                    break;
                 default:
                     ex = new BadHttpRequestException("Bad request.", 400);
                     break;
--- a/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
@@ -803,6 +803,11 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Http
 
                 ConnectionControl.CancelTimeout();
 
+                if (scan.Peek() == 0x16)
+                {
+                    RejectRequest(RequestRejectionReason.HttpsOnHttp);
+                }
+
                 _requestProcessingStatus = RequestProcessingStatus.RequestStarted;
 
                 int bytesScanned;
--- a/test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/HttpsTests.cs
+++ b/test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/HttpsTests.cs
@@ -179,7 +179,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.FunctionalTests
                     TotalErrorsLogged++;
                 }
 
-                if (exception?.Message == "An SSL/TLS handshake might have been attempted at an HTTP endpoint.")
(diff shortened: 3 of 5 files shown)


src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
```diff
@@ -808,6 +808,11 @@ public RequestLineStatus TakeStartLine(SocketInput input)
 
                 ConnectionControl.CancelTimeout();
 
+                if (scan.Peek() == 0x16)
+                {
+                    throw new IOException("SSL/TLS handshake was attempted on an HTTP endpoint.");
```
halter73: Why not keep it as a BadHttpRequestException?

cesarbs: Yeah, I'll do that. I had initially reasoned that this would end up with the server attempting to send a 400 response which make no sense to a TLS client, but actually if I throw before changing `_requestProcessingStatus` (like I'm doing here) that won't happen.

mikeharder: Since we're just checking the first byte, and a client could send 0x16 for any reason (though most likely it was TLS handshake), I would change the error message to say it *might* have been an SSL/TLS handshake.

cesarbs: Updated.

src/Microsoft.AspNetCore.Server.Kestrel/BadHttpRequestException.cs
```diff
@@ -102,6 +102,9 @@ internal static BadHttpRequestException GetException(RequestRejectionReason reas
                 case RequestRejectionReason.TooManyHeaders:
                     ex = new BadHttpRequestException("Request contains too many headers.", 431);
                     break;
+                case RequestRejectionReason.HttpsOnHttp:
+                    ex = new BadHttpRequestException("An SSL/TLS handshake might have been attempted at an HTTP endpoint.", 400);
```
mikaelm12: I think this should read `An SSL/TLS handshake might have been attempted **on** an HTTP endpoint`

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/HttpsTests.cs
```diff
@@ -133,12 +170,19 @@ private class ApplicationErrorLogger : ILogger
...
+            public TaskCompletionSource<object> InvalidHandshakeLogTcs { get; } = new TaskCompletionSource<object>();
+
             public void Log<TState>(LogLevel logLevel, EventId eventId, TState state, Exception exception, Func<TState, Exception, string> formatter)
             {
                 if (logLevel == LogLevel.Error)
                 {
                     TotalErrorsLogged++;
                 }
+
+                if (exception?.Message == "An SSL/TLS handshake might have been attempted at an HTTP endpoint.")
```
mikaelm12: Here too

test/Microsoft.AspNetCore.Server.KestrelTests/FrameTests.cs
```diff
@@ -953,6 +953,33 @@ public void TakeStartLineThrowsOnUnsupportedHttpVersion()
...
+                    Log = trace
+                };
+                var frame = new Frame<object>(application: null, context: connectionContext);
+                frame.Reset();
+
+                var requestLineBytes = new byte[] { 0x16 };
+                socketInput.IncomingData(requestLineBytes, 0, requestLineBytes.Length);
+
+                var exception = Assert.Throws<BadHttpRequestException>(() => frame.TakeStartLine(socketInput));
+                Assert.Equal("An SSL/TLS handshake might have been attempted at an HTTP endpoint.", exception.Message);
```
mikaelm12: Last one

cesarbs:
@mikaelm12 Fixed grammar 😄

davidfowl:
Can we not do this? What does http.sys and Nginx do?  We have https in a separate package, yet we have this check in the core of kestrel. While this is a nicer error, nobody else ran into it yet (besides us)

cesarbs:
nginx:

image

muratg:
@davidfowl Once Kestrel is an edge server, this would be easy to hit if you're playing with URL rewrite. Are you concerned about a potential perf implication?

cesarbs:
IIS only logs this:

`2016-09-23 16:42:39 ::1 GET /favicon.ico - 8082 - ::1 Mozilla/5.0+(Windows+NT+10.0;+WOW64)+AppleWebKit/537.36+(KHTML,+like+Gecko)+Chrome/53.0.2785.116+Safari/537.36 https://localhost:8082/ 404 0 2 1`

cesarbs:
@davidfowl An alternative to this would be to not special case 0x16 and simplify error reporting for the request line - something like what nginx does i.e. just log "invalid request line" followed by what we saw in the request line (print out hex codes for non-print chars).

The issue right now is that you get a cryptic error message ("no space after method"). I don't think we need to be this specific when reporting errors in the request line.

benaadams:
@cesarbs could do an if split for extra error detail when `if (scan.Seek(ref _vectorSpaces, ref end) == -1)` has already failed?

davidfowl:
> @davidfowl An alternative to this would be to not special case 0x16 and simplify error reporting for the request line - something like what nginx does i.e. just log "invalid request line" followed by what we saw in the request line (print out hex codes for non-print chars).

> The issue right now is that you get a cryptic error message ("no space after method"). I don't think we need to be this specific when reporting errors in the request line.


That's much better

davidfowl:
@muratg I don't like the light coupling to TLS/SSL in kestrel core. The SSL component is in a separate package, yet we have this special logic in the core for a better error message. IIS probably 404's because it binds to a prefix (http.sys) and that includes the scheme, host and path. The NGINX error is what I would expect.

cesarbs:
Closing this in favor of https://github.com/aspnet/KestrelHttpServer/pull/1122

PR #1121, opened 2016-09-23
Change context relationships from inheritance to composition

cesarbs:
We keep all contexts around while copying their fields at the same time. We can just point back to existing contexts.

This reduces memory usage per connection:

type | dev | this change | reduction
------|------|-------------|---------------
Connection | 216 bytes | 160 bytes | 26%
Frame | 560 bytes | 464 bytes | 17%

Listeners are a little slimmer too, but that doesn't matter much since there are only a few of them per Kestrel instance.

### dev

dev_connection

dev_frame

### this change

change_connection

change_frame

@davidfowl @halter73 @mikeharder

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
@@ -11,6 +11,7 @@ using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Http.Features;
 using Microsoft.AspNetCore.Server.Kestrel.Internal.Infrastructure;
 using Microsoft.Extensions.Internal;
 using Microsoft.Extensions.Logging;
@@ -20,7 +21,7 @@ using Microsoft.Extensions.Primitives;
 
 namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Http
 {
-    public abstract partial class Frame : ConnectionContext, IFrameControl
+    public abstract partial class Frame : IFrameControl
     {
         private static readonly ArraySegment<byte> _endChunkedResponseBytes = CreateAsciiByteArraySegment("0\r\n\r\n");
         private static readonly ArraySegment<byte> _continueBytes = CreateAsciiByteArraySegment("HTTP/1.1 100 Continue\r\n\r\n");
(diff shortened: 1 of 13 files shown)


cesarbs:
@mikeharder Updated with your suggestion to add more properties to reduce the amount of necessary code changes.

test/Microsoft.AspNetCore.Server.KestrelTests/FrameResponseHeadersTests.cs
```diff
@@ -61,14 +66,14 @@ public void InitialDictionaryIsEmpty()
...
         [InlineData("Unknown\u0080-Header", "Data")]
-        [InlineData("Ser�ver", "Data")]
-        [InlineData("Server", "Da�ta")]
-        [InlineData("Unknown�-Header", "Data")]
-        [InlineData("Ser�ver", "Data")]
-        [InlineData("�erver", "Data")]
-        [InlineData("Server", "Da�ta")]
-        [InlineData("Unknown�-Header", "Data")]
-        [InlineData("Ser�ver", "Data")]
+        [InlineData("Ser�ver", "Data")]
```
cesarbs: Definitely not. Will revert. Need to find out who's the culprit.

cesarbs: Fixed.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
```diff
@@ -72,14 +73,42 @@ public abstract partial class Frame : ConnectionContext, IFrameControl
...
+                ConnectionContext.PrepareRequest = value;
+            }
+        }
+
+        protected IConnectionControl ConnectionControl => ConnectionContext.ConnectionControl;
+        protected IKestrelTrace Log => ConnectionContext.ListenerContext.ServiceContext.Log;
+
+        private DateHeaderValueManager DateHeaderValueManager => ConnectionContext.ListenerContext.ServiceContext.DateHeaderValueManager;
+        private ServerAddress ServerAddress => ConnectionContext.ListenerContext.ServerAddress;
+        private KestrelServerOptions ServerOptions => ConnectionContext.ListenerContext.ServiceContext.ServerOptions;
```
halter73: We actually use this a lot in frame (e.g. MaxRequestLineSize, MaxHeaderCount). Can we make KestrelServerLimits a private field and set it in the ctor instead of having frame access this via `ConnectionContext.ListenerContext.ServiceContext.ServerOptions.Limits` every time? Even if the properties are inlined, that's a lot of indirection for a fairly hot path.

benaadams: Change to `{ get; } =` rather than `=>`?

cesarbs: Done.

PR #1122, opened 2016-09-24
Refactor request line validation and error reporting (#1116)

cesarbs:
An alternative to #1119.

Removed a number of `RequestRejectionReason` members in favor of a new one (`InvalidRequestLine`). Log the invalid line with control characters escaped.

This addresses #1116 by logging something like:

`Connection id "0HKV41SOQDMIQ" bad request data: "Invalid request line: <0x16><0x03><0x01><0x00><0xB9><0x01><0x00><0x00><0xB5><0x03><0x03>%<0xE8>"<0x8E>)<0xFA><0x04><0xA9><0xB5>o<0x9C><0x8E>#<0xF8>9<0xD8><0x08><0xD9>2<0x18><0x16><0x9F><0xE5>D<0x14><0xCE><0x0F>&<0xF6><0xBE>cg<0x00><0x00>"<0xC0>+<0xC0>/<0xC0>,<0xC0>0<0xCC><0xA9><0xCC><0xA8><0xCC><0x14><0xCC><0x13><0xC0><0x09><0xC0><0x13><0xC0><0x0A>"`

I can special-case lines starting with a control character to only log the first, say, 5 bytes or so.

@halter73 @mikeharder @Tratcher @davidfowl @mikaelm12


cesarbs:
Pending: check perf.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
```diff
@@ -818,20 +819,46 @@ public RequestLineStatus TakeStartLine(SocketInput input)
...
+
+                var methodStart = scan;
+                var methodEnd = methodStart;
+                if (methodEnd.Seek(ref _vectorSpaces, ref end) == -1)
                 {
-                    if (scan.Seek(ref _vectorSpaces, ref end) == -1)
-                    {
-                        RejectRequest(RequestRejectionReason.MissingSpaceAfterMethod);
-                    }
+                    RejectRequest(RequestRejectionReason.InvalidRequestLine, GetInvalidRequestLineString(start, end));
```
halter73: We might want to make sure that the logging level is high enough before calling GetInvalidRequestLineString.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
```diff
@@ -818,20 +819,46 @@ public RequestLineStatus TakeStartLine(SocketInput input)
...
                 }
 
-                string method;
-                var begin = scan;
-                if (!begin.GetKnownMethod(out method))
+                end.Take();
+
+                var methodStart = scan;
+                var methodEnd = methodStart;
+                if (methodEnd.Seek(ref _vectorSpaces, ref end) == -1)
```
halter73: It must be faster to call GetKnownMethod first in the normal case when there's a valid request with a known method right?

halter73:
We should get some micro benchmarks for TakeStartLine. I'm afraid changes like this will always look negligible perfwise if we just run plaintext because of the high variance.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
cesarbs: This is not a hot code path. Why bother with that?

cesarbs:
I've decided to go with less dramatic changes to `Frame.TakeStartLine()`, leaving only the new error reporting.

Tratcher:
Does this assume you've found the CRLF first? that's unlikely in the https scenario.

cesarbs:
@Tratcher We look for an LF. If one isn't found, either the request will be rejected as having too long a start line, or it will time out.

We could check if the first character is a control character and reject the request with something like "A non-HTTP request was attempted on an HTTP endpoint." As discussed in the previous PR, we don't want to special case SSL/TLS here.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
halter73: We make the effort using LoggerMessage.Define in KestrelTrace to not even call ILogger.Log if a given LogLevel isn't enabled.

I think it makes sense for this too. I'm not sure how common invalid request lines are, but a quick check using ILogger.IsEnabled(logLevel) doesn't seem too onerous.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
```diff
@@ -1019,6 +1019,20 @@ public RequestLineStatus TakeStartLine(SocketInput input)
             }
         }
 
+        private string GetInvalidRequestLineString(MemoryPoolIterator start, MemoryPoolIterator end)
+        {
+            var sb = new StringBuilder();
+            var scan = start;
+
+            while (scan.Block != end.Block || scan.Index != end.Index)
```
halter73: No max length? Having up to an 8KB string in the exception might be a bit much.

halter73:
We could use GetInvalidRequestLineString for request line that was too long errors, but we'd want to make sure to truncate.

I don't think the initial part of the TLS handshake would be long enough to trigger that though. I suspect the client would wait for the server to send a response if it wasn't for the server sending a response after it happens to find the `\n` character.

We could try to find the space after the http method before finding the `\n`. This would allow us to use a smaller seek limit I'm not sure what that should be. Maybe 32 bytes?

benaadams:
Could move the ascii validation up; SeekAscii -> bomb when you get non-ascii? Then don't validate on convert as is prevalidated.

cesarbs:
I've added a max chars limit to the log message.

Let's not optimize for the error case. Doing a full scan on the request line to check for non-ASCII before parsing is going to impact every legit request line. I also don't like the idea of seeking for the space first since it makes this code a bit more complex, plus I don't want to place an arbitrary limit on the method name since that's not in the spec.

The one scenario we can cheaply prematurely check is the reason that prompted this change - HTTPS on HTTP. It's just a matter of checking for a control char in the first byte and emitting the right log message ("attempted non-HTTP on HTTP"). Anything beyond that is guesswork as to what should be done IMO.

cesarbs:
Ping.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
```diff
@@ -939,21 +943,19 @@ public RequestLineStatus TakeStartLine(SocketInput input)
 
                     if (httpVersion == null)
                     {
-                        RejectRequest(RequestRejectionReason.MissingHTTPVersion);
+                        RejectRequest(RequestRejectionReason.InvalidRequestLine, start.GetAsciiStringEscaped(end, MaxInvalidRequestLineChars));
                     }
                     else if (httpVersion != "HTTP/1.0" && httpVersion != "HTTP/1.1")
                     {
-                        RejectRequest(RequestRejectionReason.UnrecognizedHTTPVersion);
+                        RejectRequest(RequestRejectionReason.UnrecognizedHTTPVersion, httpVersion);
```
halter73: Couldn't httpVersion be arbitrarily long? The above comment about it being possible for a valid version to span more than 2 blocks is also wrong.

cesarbs: Addressed.

cesarbs:
Updated.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
```diff
@@ -923,43 +933,40 @@ public RequestLineStatus TakeStartLine(SocketInput input)
...
                 scan.Take(); // consume CR
-                if (scan.Block != end.Block || scan.Index != end.Index)
+                if (scan.Take() != '\n')
                 {
-                    RejectRequest(RequestRejectionReason.MissingLFInRequestLine);
+                    RejectRequest(RequestRejectionReason.InvalidRequestLine,
+                        Log.IsEnabled(LogLevel.Information) ? start.GetAsciiStringEscaped(end, MaxInvalidRequestLineChars) : string.Empty);
                 }
-                scan.Take(); // consume LF
-                end = scan;
```
halter73: It doesn't look the end is being advanced passed the line feed any more. Doesn't this lead to a bad examined value in the call to ConsumingComplete?

test/Microsoft.AspNetCore.Server.KestrelTests/FrameTests.cs
```diff
@@ -1128,7 +1128,7 @@ public void TakeStartLineThrowsOnUnsupportedHttpVersion()
                 socketInput.IncomingData(requestLineBytes, 0, requestLineBytes.Length);
 
                 var exception = Assert.Throws<BadHttpRequestException>(() => frame.TakeStartLine(socketInput));
-                Assert.Equal("Unrecognized HTTP version.", exception.Message);
+                Assert.Equal("Unrecognized HTTP version: HTTP/1.2", exception.Message);
```
halter73: Can we test for an http version longer than 8 chars?

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
cesarbs: It is: https://github.com/aspnet/KestrelHttpServer/pull/1122/files#diff-4a391b137e1879654de7fd4febc3db67R861

test/Microsoft.AspNetCore.Server.KestrelTests/FrameTests.cs
cesarbs: Added. While adding it I realized it makes sense to limit the printed version to 9 chars instead of 8. Otherwise something like "HTTP/1.1a" appears as "HTTP/1.1..." in the log and that's not as helpful as showing at least one character that made the version "bad".

cesarbs:
Updated.

halter73:
Ever get benchmarks?

halter73:
Otherwise LGTM

cesarbs:
@halter73 Yeah, no perf hit.

cesarbs:
@halter73 Or did you mean microbenchmarks?

halter73:
Numbers? I meant /plaintext, but microbenchmarks would be awesome if not too difficult.

cesarbs:
### dev

```
[04:39:53.046] Starting scenario Plaintext on benchmark client...
[04:40:04.074] Scenario Plaintext completed on benchmark client
[04:40:04.074] RPS: 1173673.66
```

### PR

```
[04:40:46.510] Starting scenario Plaintext on benchmark client...
[04:40:57.537] Scenario Plaintext completed on benchmark client
[04:40:57.537] RPS: 1181716.67
```

Usual fluctuation seen across multiple runs.

PR #1123, opened 2016-09-25
Specify sequential layout to avoid warning

svick:
The previous version caused this warning:

> SockAddr.cs(17,22): warning CS0414: The field 'SockAddr._field3' is assigned but its value is never used

To avoid the warning, explicitly specify `[StructLayout(LayoutKind.Sequential)]`. This does not change any behavior (since `Sequential` is the default for `struct`s), but it lets the compiler and readers know that layout is important.


cesarbs:
Nice.

Issue #1104, opened 2016-09-13
Connection header "close" hangs


jdslavin:
Request hangs when doing an HTTP GET request that has a "Connection" header of "close".

If the request is made with "Connection" header "keep-alive" all works well.

This issue came up because we are using an apache HTTP client library that automatically adds the Connection Header "close" when making non connection pool based requests.

benaadams:
Dupe https://github.com/aspnet/KestrelHttpServer/issues/1041 ? Looks fixed in 1.1.0

jdslavin:
I am not sure this is a duplicate, the close is sent on the request header where 1041 possibly looks like it deals with the response headers.     Easy enough to test though, any GET request with Connection "close" in the request headers hangs.

cesarbs:
@jdslavin What version of Kestrel are you using?

jdslavin:
1.0.0

halter73:
@benaadams I agree that this looks like it could be a dupe of #1041, but even before the fix Kestrel would always close the connection given a "Connection: close" request header which you would think would prevent most clients from hanging. The issue was not setting the right/best response headers.

@jdslavin To be sure, can you provide us with a sample app? If this happens with any app or with our static file handler, you can just say so.

As far as the client goes, are you using the Apache's [Java HttpClient](http://hc.apache.org/httpcomponents-client-ga/)? Can you reproduce the issue with Fiddler/Postman or something like that?

If you want to see if this has been fixed yourself, you can try using the 1.1.0-* preview bits from [our nightly myget feed](https://dotnet.myget.org/gallery/aspnetcore-dev)

halter73:
One issue I've seen with "Connection: close" requests is that some clients won't send a FIN until a response has been received. This makes it impossible to ever complete a call to `Request.Body.ReadAsync(...)` for such a request.

In #406, which was fixed in rc2, we stopped trying to consume the entire body in Kestrel itself for these types of request, but it could be that the app is doing it.

jdslavin:
I reproduced this same issue via Advanced Rest Client by just adding the "Connection: close"  request header.    I will try the preview bits to see what happens.

halter73:
@jodavis Is the app reading the request body? Do you have an app it repros with? Or does it repro with the static file middleware (with no other middleware set up)?

jdslavin:
[...]

PR #1126, opened 2016-09-27
Assume zero length on non-keepalive requests without Content-Length or Transfer-Encoding header (#1104)

cesarbs:
Prevent hangs when the client sends a `Connection: close` request but waits for the server to close the connection first, while the server is hung waiting for data from the client when no length was specified. This happens when a `Connection: close` request does not include a `Content-Length` or `Transfer-Encoding: chunked` header. See #1104.

@mikeharder pointed out that this might be considered a breaking change. This could break existing apps that expect to be able to read from `Connection: close` requests that send a body of unspecified length. This is a bad idea though, and apps should not be doing that, since there's no way to differentiate between message end and a premature connection close in this situation. cc @Eilon for input on that.

@halter73

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/MessageBody.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/MessageBody.cs
@@ -234,6 +234,11 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Http
             var connection = headers.HeaderConnection.ToString();
             if (connection.Length > 0)
             {
+                if (connection.Equals("upgrade", StringComparison.OrdinalIgnoreCase))
+                {
+                    return new ForRemainingData(context);
+                }
+
                 keepAlive = connection.Equals("keep-alive", StringComparison.OrdinalIgnoreCase);
             }
 
@@ -257,12 +262,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Http
                 }
             }
 
-            if (keepAlive)
-            {
-                return new ForContentLength(true, 0, context);
-            }
-
-            return new ForRemainingData(context);
+            return new ForContentLength(keepAlive, 0, context);
         }
 
         private class ForRemainingData : MessageBody
--- a/test/Microsoft.AspNetCore.Server.KestrelTests/EngineTests.cs
+++ b/test/Microsoft.AspNetCore.Server.KestrelTests/EngineTests.cs
@@ -14,7 +14,6 @@ using Microsoft.AspNetCore.Http.Features;
 using Microsoft.AspNetCore.Server.Kestrel;
 using Microsoft.AspNetCore.Server.Kestrel.Internal;
 using Microsoft.AspNetCore.Server.Kestrel.Internal.Http;
-using Microsoft.AspNetCore.Server.Kestrel.Internal.Infrastructure;
 using Microsoft.AspNetCore.Testing;
 using Microsoft.Extensions.Internal;
 using Xunit;
@@ -108,7 +107,7 @@ namespace Microsoft.AspNetCore.Server.KestrelTests
             var started = engine.CreateServer(address);
 
             var socket = TestConnection.CreateConnectedLoopbackSocket(address.Port);
-            socket.Send(Encoding.ASCII.GetBytes("POST / HTTP/1.0\r\n\r\nHello World"));
+            socket.Send(Encoding.ASCII.GetBytes("POST / HTTP/1.0\r\nContent-Length: 11\r\n\r\nHello World"));
             socket.Shutdown(SocketShutdown.Send);
             var buffer = new byte[8192];
             while (true)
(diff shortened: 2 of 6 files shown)


halter73:
Nearly any bug fix that changes bad behavior is technically a breaking change. I think this fits squarely within that. If this PR does what we intend, I would be surprised if this breaks anyone's app.

halter73:
This fixes #1104

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/MessageBody.cs
```diff
@@ -257,25 +257,7 @@ protected virtual void OnConsumedBytes(int count)
...
 
-            if (keepAlive)
-            {
-                return new ForContentLength(true, 0, context);
-            }
-
-            return new ForRemainingData(context);
-        }
-
-        private class ForRemainingData : MessageBody
```
halter73: Like I mentioned in https://github.com/aspnet/KestrelHttpServer/issues/1104#issuecomment-247186700 We need this for Upgrade requests that usually don't have a `Content-Length` (e.g. WebSockets).

test/Microsoft.AspNetCore.Server.KestrelTests/EngineTests.cs
```diff
@@ -1336,6 +1345,7 @@ public void ConnectionCanReadAndWrite(TestServiceContext testContext)
                     await connection.SendEnd(
                         "GET / HTTP/1.1",
                         "Connection: Upgrade",
+                        "Content-Length: 11",
```
halter73: Websocket Upgrade requests don't have `Content-Length`. https://tools.ietf.org/html/rfc6455#section-1.2

mikeharder:
@halter73: While I agree this isn't a major breaking change, I do think it's more breaking than a typical bug fix.  It breaks HTTP 1.0 requests with a body but no content-length.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/MessageBody.cs
cesarbs: Fixed

halter73:
@mikeharder Correct. I don't know of any client that actually does that however. We verified that IIS will respond with a 4XX for HTTP 1.0 requests with a body, but without a content length. We can look at what nginx and nodejs do too.

Eilon:
Hmm, yes, very interesting. I suppose there are two questions:

1. What do other servers do?
2. Do any "normal" clients typically require this kind of behavior?

Also @DamianEdwards due to breaking change.

davidfowl:
IIS explods with a required 411 content length required. That's what lead to this change in the first place. There was a piece of middleware in the pipeline eagerly reading causing the deadlock. Though breaking this behavior shouldn't affect anyone doing sane things

test/Microsoft.AspNetCore.Server.KestrelTests/EngineTests.cs
```diff
@@ -481,8 +490,8 @@ public void ConnectionCanReadAndWrite(TestServiceContext testContext)
 
                 using (var connection = server.CreateConnection())
                 {
-                    await connection.SendEnd(
-                        "GET / HTTP/1.0",
+                    await connection.Send(
+                        "POST / HTTP/1.0",
```
halter73: Why POST?

halter73:
@cesarbs Can you check the behavior of nginx and node.js? Their behavior might change based on HTTP method if they're like IIS btw.

test/Microsoft.AspNetCore.Server.KestrelTests/EngineTests.cs
cesarbs: I had added a body to the message and for correctness sake changed the method to POST. Then I removed the body and forgot to change the method back to GET. Reverting.

cesarbs: Reverted.

cesarbs:
@halter73 nginx sends a 411. nodejs accepts the request but assumes 0 length and closes the connection afterwards, even on a keep-alive connection. That's smart, avoids corrupting the next request.

halter73:
Node.js would only close the connection for non-content-length keep-alive POST requests, right? I imagine it would keep the connection alive for GET requests without a content-length.

cesarbs:
@halter73 Ah, yes, it keeps the connection open for a keep-alive GET request. It does honor the `Connection: close` request if it's there and doesn't attempt to read anything from the body (mentioning this just to say you can't repro the Kestrel issue on node).

halter73:
I'm OK with not having any HTTP method specific logic in this PR.

It would be more user friendly to require content-length/transfer-encoding for some HTTP methods like other servers do https://github.com/aspnet/KestrelHttpServer/issues/1104#issuecomment-247185848, but that could be done in a separate PR. This is certainly better than what we have now IMHO.

We should create another issue for rejecting requests that should have a content length based on the HTTP-method but don't. It will be interesting to see if there is any feedback on that.

Issue #1103, opened 2016-09-13
"EPIPE broken pipe" and "ECANCELED operation canceled" on event stream requests (+ ObjectDisposedException with HTTPS)


SteveSandersonMS:
Is there a correct, supported way to implement long-running requests that are intended to be closed by the client, e.g., for `EventStream`?

See the (possibly naive) implementation of `ServerSentEventController` below. This produces a `text/event-stream` response. The client-side code keeps open a request to this until the user navigates away, which aborts the request.

### Issue 1: Client-initiated aborts

Expected: When the client aborts the request, the `HttpContext.RequestAborted` token should fire immediately, and the response closes gracefully.

Actual: When the client aborts the request, `HttpContext.RequestAborted` does *not* fire immediately. After 5-10 seconds, in the console I get an error like `Connection id "0HKUROLPTMNFF" communication error
Microsoft.AspNetCore.Server.Kestrel.Internal.Networking.UvException: Error -32 EPIPE broken pipe` on OS X, with no stack trace. On Windows, the error is `ECANCELED operation canceled`, again with no stack trace. Immediately after this error, `HttpContext.RequestAborted` does fire.

Note that the errors show up regardless of whether the cancellation token is passed to `WriteAsync` or not.

### Issue 2: Server-initiated aborts

If, while the request is ongoing, you ctrl+c in the console to terminate the server app, I'd expect it really to terminate (closing the TCP connection to the client at once). However, the app keeps running for another 5-10 seconds, sending more messages to the client during that period.

**Appendix: ServerSentEventController**

```csharp
    public class ServerSentEventController : Controller
    {
        [Route("/event-stream")]
        public async Task MyEventStream()
        {
            HttpContext.Response.Headers.Add("Content-Type", "text/event-stream");

            var ct = HttpContext.RequestAborted;
            while (!ct.IsCancellationRequested)
            {
                try
                {
                    await HttpContext.Response.WriteAsync($"data: Hello ({DateTime.Now})\r\r", ct);
                    await Task.Delay(1000, ct);
                }
                catch (TaskCanceledException)
                {
                    // If the client aborted the request, don't regard it as an error
                }
            }
        }
    }
```

BTW I know there have been other issues reported previously about `EPIPE broken pipe` (OS X) and `ECANCELED operation canceled` (Windows), but those are closed and the above is still a simple repro.

cesarbs:
[...]

PR #1127, opened 2016-09-28
Don't reset frame state when connection is aborted (#1103).

cesarbs:
See https://github.com/aspnet/KestrelHttpServer/issues/1103#issuecomment-250237677.

- [x] TODO: add tests.


cesarbs:
Wondering if this is the best fix. Setting `PrepareRequest` to null in `Frame.Abort()` might be better. Thoughts?

halter73:
We call Reset() at the end of the loop instead of the beginning to avoid rooting strings and maybe a few other things between requests and when the connection is idle.

I think the more general purpose change might be just to do `if (!_requestProcessingStopping) Reset()`. This covers both the aborted and request rejected cases.

cesarbs:
@halter73 Changed to `Reset()` on `!_requestProcessingStopping`.

cesarbs:
Added test.

halter73:
:shipit:

PR #1128, opened 2016-09-28
Refactored to use the fields from the HttpMethods class

mikaelm12:
@davidfowl @halter73 @muratg @Tratcher


test/Microsoft.AspNetCore.Server.KestrelTests/MemoryPoolIteratorTests.cs
```diff
@@ -432,15 +432,15 @@ public void SkipThrowsWhenSkippingMoreBytesThanAvailableInMultipleBlocks()
...
-        [InlineData("CONNECT / HTTP/1.1", ' ', true, MemoryPoolIteratorExtensions.HttpConnectMethod)]
-        [InlineData("DELETE / HTTP/1.1", ' ', true, MemoryPoolIteratorExtensions.HttpDeleteMethod)]
-        [InlineData("GET / HTTP/1.1", ' ', true, MemoryPoolIteratorExtensions.HttpGetMethod)]
-        [InlineData("HEAD / HTTP/1.1", ' ', true, MemoryPoolIteratorExtensions.HttpHeadMethod)]
-        [InlineData("PATCH / HTTP/1.1", ' ', true, MemoryPoolIteratorExtensions.HttpPatchMethod)]
-        [InlineData("POST / HTTP/1.1", ' ', true, MemoryPoolIteratorExtensions.HttpPostMethod)]
-        [InlineData("PUT / HTTP/1.1", ' ', true, MemoryPoolIteratorExtensions.HttpPutMethod)]
-        [InlineData("OPTIONS / HTTP/1.1", ' ', true, MemoryPoolIteratorExtensions.HttpOptionsMethod)]
-        [InlineData("TRACE / HTTP/1.1", ' ', true, MemoryPoolIteratorExtensions.HttpTraceMethod)]
+        [InlineData("CONNECT / HTTP/1.1", ' ', true,"CONNECT")]
```
mikeharder: Any reason to hardcode `"CONNECT"` rather than using `HttpMethods.Connect`, as is done in the product code?

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Infrastructure/MemoryPoolIteratorExtensions.cs
```diff
@@ -279,7 +270,7 @@ public static bool GetKnownMethod(this MemoryPoolIterator begin, out string know
 
             if ((value & _mask4Chars) == _httpGetMethodLong)
             {
-                knownMethod = HttpGetMethod;
+                knownMethod = HttpMethods.Get;
```
mikeharder: Might be a tiny performance degradation since the value is changing from `const` to `static readonly`.  Is there a reason the properties on `HttpMethods` are `static readonly` rather than `const`?  I see the `HttpMethods` type is new for 1.1, so we should be able to change the properties to `const` now.  After 1.1 ships it would be a breaking change.

I don't see any possible way the value of these properties would change in the future, so `const` should be safe.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Infrastructure/MemoryPoolIteratorExtensions.cs
```diff
@@ -12,16 +13,6 @@ public static class MemoryPoolIteratorExtensions
     {
         private static readonly Encoding _utf8 = Encoding.UTF8;
 
-        public const string HttpConnectMethod = "CONNECT";
```
mikeharder: Are breaking changes to public APIs on internal types allowed in 1.1?

test/Microsoft.AspNetCore.Server.KestrelTests/MemoryPoolIteratorTests.cs
mikaelm12: Can't use `HttpMethods.Connect` here because its not a const.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Infrastructure/MemoryPoolIteratorExtensions.cs
halter73: It's in an "Internal" namespace.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Infrastructure/MemoryPoolIteratorExtensions.cs
halter73: https://github.com/aspnet/HttpAbstractions/pull/712#issuecomment-250249313

halter73:
@mikaelm12 You know how to run the plaintext benchmarks, right? I really don't expect this change will make any difference, but it's easy enough to test using @mikeharder's BenchmarksDriver.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Infrastructure/MemoryPoolIteratorExtensions.cs
davidfowl: I doubt the performance will degrade

test/Microsoft.AspNetCore.Server.KestrelTests/MemoryPoolIteratorTests.cs
halter73: Fix the missing space though.

benaadams: @davidfowl Already do https://github.com/aspnet/KestrelHttpServer/blob/dev/test/Microsoft.AspNetCore.Server.KestrelTests/MemoryPoolIteratorTests.cs#L497-L518

src/Microsoft.AspNetCore.Server.Kestrel/project.json
```diff
@@ -22,7 +22,8 @@
     "Microsoft.Extensions.TaskCache.Sources": {
       "version": "1.1.0-*",
       "type": "build"
-    }
+    },
+    "Microsoft.AspNetCore.Http.Abstractions": "1.1.0-*"
```
Tratcher: Sort

davidfowl: Hmmmmm, this is a new dependency? Doesn't this come in transitively already?

muratg: Yeah, it should be coming from `Microsoft.AspNetCore.Hosting` transitively.

PR #1134, opened 2016-09-30
Handle non-body responses

cesarbs:
Rebased @benaadams's #962 and made some changes.

- [x] TODO: chunk HTTP/1.1 non-keepalive responses (see discussion at https://github.com/aspnet/KestrelHttpServer/pull/962/files#r81381973)

Like the original PR, this addresses #952.

@halter73 @benaadams @Tratcher


halter73:
Let's close #962 then.

cesarbs:
Updated.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
```diff
@@ -747,50 +758,61 @@ private Task WriteSuffix()
...
+
             if (_keepAlive && hasConnection)
             {
                 var connectionValue = responseHeaders.HeaderConnection.ToString();
                 _keepAlive = connectionValue.Equals("keep-alive", StringComparison.OrdinalIgnoreCase);
             }
 
-            if (!responseHeaders.HasTransferEncoding && !responseHeaders.HasContentLength)
+            // Set whether response can have body
+            _canHaveBody = StatusCanHaveBody(StatusCode) && Method != "HEAD";
```
halter73: Since 101 status can have a body of sorts, I would prefer `_canHaveBody` is true in that case. When deciding whether or not to set `Content-Length: 0` we can check for 101 specifically.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
```diff
@@ -747,50 +758,61 @@ private Task WriteSuffix()
...
+
+            if (_canHaveBody)
             {
-                if (appCompleted)
+                if (!responseHeaders.HasTransferEncoding && !responseHeaders.HasContentLength)
                 {
-                    // Don't set the Content-Length or Transfer-Encoding headers
-                    // automatically for HEAD requests or 101, 204, 205, 304 responses.
```
halter73: I would like to keep this comment somewhere.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
```diff
@@ -747,50 +758,61 @@ private Task WriteSuffix()
...
+                        {
+                            _keepAlive = false;
+                        }
                     }
                 }
             }
+            else
+            {
+                // Don't set the Content-Length or Transfer-Encoding headers
+                // automatically for HEAD requests or 204, 205, 304 responses.
```
halter73: This comment doesn't makes sense here since no header is being automatically set.

Also, why not reject 101 responses with a Transfer-Encoding?
And why not reject 101/204/205/304 responses with a content-length. It looks like that's also illegal.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
```diff
@@ -1257,12 +1279,76 @@ public bool StatusCanHaveBody(int statusCode)
...
+                responseHeaders.SetRawServer(Constants.ServerName, _bytesServer);
+            }
+        }
+
+        public void HandleNonBodyResponseWrite(int count)
+        {
+            if (Method == "HEAD")
+            {
+                // Don't write to body for HEAD requests.
+                Log.ConnectionHeadResponseBodyWrite(ConnectionId, count);
```
halter73: Can we have a flag so this is only logged once per request/response? Maybe log it at the end of the response so we can count the size of all the writes.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
benaadams: 101/204/205/304 can have a content-length header of 0 I think?

benaadams: Can 101/204/205/304 have a `Content-Length` of `0` (if set)?

halter73: https://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html#sec10.3.5

> If the conditional GET used a strong cache validator (see section 13.3.3), the response SHOULD NOT include other entity-headers. Otherwise (i.e., the conditional GET used a weak validator), the response MUST NOT include other entity-headers; this prevents inconsistencies between cached entity-bodies and updated headers.

Entity-headers include Content-Length, but I don't know anything about strong cache validators, so I guess there is some wiggle room. The other status codes don't seem to mandate anything.

Where is the spec requiring that `Transfer-Encoding` header can't be sent? Like `Content-Length` it doesn't really make sense, and we wouldn't set it automatically, but where is it explicitely prohibited?

benaadams: https://tools.ietf.org/html/rfc7230#section-3.3.1

> Transfer-Encoding MAY be sent in a response to a HEAD request or in a 304 (Not Modified) response (Section 4.1 of [RFC7232]) to a GET request, neither of which includes a message body, to indicate that the origin server would have applied a transfer coding to the message body if the request had been an unconditional GET.  This indication is not required, however, because any recipient on the response chain (including the origin server) can remove transfer codings when they are not needed.

> A server MUST NOT send a Transfer-Encoding header field in any response with a status code of 1xx (Informational) or 204 (No Content).  A server MUST NOT send a Transfer-Encoding header field in any 2xx (Successful) response to a CONNECT request (Section 4.3.6 of [RFC7231]).

Has lots more here
>  A sender MUST NOT send a Content-Length header field in any message that contains a Transfer-Encoding header field.

> A user agent SHOULD send a Content-Length in a request message when no Transfer-Encoding is sent and the request method defines a meaning for an enclosed payload body.  For example, a Content-Length header field is normally sent in a POST request even when the value is 0 (indicating an empty payload body).  A user agent SHOULD NOT send a Content-Length header field when the request message does not contain a payload body and the method semantics do not anticipate such a body.

> A server MAY send a Content-Length header field in a response to a HEAD request (Section 4.3.2 of [RFC7231]); a server MUST NOT send Content-Length in such a response unless its field-value equals the decimal number of octets that would have been sent in the payload body of a response if the same request had used the GET method.

> A server MAY send a Content-Length header field in a 304 (Not Modified) response to a conditional GET request (Section 4.1 of [RFC7232]); a server MUST NOT send Content-Length in such a response    unless its field-value equals the decimal number of octets that would have been sent in the payload body of a 200 (OK) response to the same request.

etc

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
cesarbs: Will move above `if (_canHaveBody)`.

cesarbs:
@halter73 I've addressed your feedback.

test/Microsoft.AspNetCore.Server.KestrelTests/EngineTests.cs
```diff
@@ -585,6 +579,49 @@ public void ConnectionCanReadAndWrite(TestServiceContext testContext)
...
+                        $"Date: {testContext.DateHeaderValue}",
+                        "",
+                        "hello, world");
+                }
+
+                using (var connection = server.CreateConnection())
+                {
+                    await connection.SendEnd(
+                        "GET / HTTP/1.0",
+                        "Connection: kee-alive",
```
halter73: `kee-alive`

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
```diff
@@ -75,6 +75,8 @@ public abstract partial class Frame : IFrameControl
         protected readonly long _keepAliveMilliseconds;
         private readonly long _requestHeadersTimeoutMilliseconds;
 
+        private int _responseBytesWritten;
```
halter73: This should be long. Maybe call this _responseBytesIgnored since we're only using it to log the number of bytes not written.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
```diff
@@ -542,6 +547,8 @@ public Task WriteAsync(ArraySegment<byte> data, CancellationToken cancellationTo
                 return WriteAsyncAwaited(data, cancellationToken);
             }
 
+            _responseBytesWritten += data.Count;
```
halter73: Could this be done just in the !_canHaveBody case?

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
cesarbs: I'll use it when fixing https://github.com/aspnet/KestrelHttpServer/issues/175, so less diff when I make that change :smile:

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
cesarbs: Same as above.

test/Microsoft.AspNetCore.Server.KestrelTests/EngineTests.cs
```diff
@@ -548,9 +548,6 @@ public void ConnectionCanReadAndWrite(TestServiceContext testContext)
                         "POST / HTTP/1.1",
                         "Content-Length: 3",
                         "",
-                        "101POST / HTTP/1.1",
-                        "Content-Length: 3",
-                        "",
```
cesarbs: 101 is now closed after response so that made this test fail. That's why I added the new test.

cesarbs:
Updated.

halter73:
I would also like to see the /plaintext benchmark results for this. :shipit: If there is no degradation.

cesarbs:
### dev

```
[04:39:53.046] Starting scenario Plaintext on benchmark client...
[04:40:04.074] Scenario Plaintext completed on benchmark client
[04:40:04.074] RPS: 1185406.81
```

### PR

```
[04:40:46.510] Starting scenario Plaintext on benchmark client...
[04:40:57.537] Scenario Plaintext completed on benchmark client
[04:40:57.537] RPS: 1194005.61
```

Usual fluctuation seen across multiple runs.

PR #1135, opened 2016-09-30
add header ordering and short circuits to known headers

nathana1:
I revisited the 'header trim' we explored before RTM (https://github.com/aspnet/KestrelHttpServer/issues/795) and it still seemed to be about a 3% performance win for techempower plaintext.

To avoid having to decide which headers to remove, this change allows us to put the most common headers first and can short circuit instead of having to check every single known header.  This results in almost the same perf as the header trim.  In million reqs/sec:

|Baseline|Header Trim|Short Circuit|
|---|---|---|
|5.18|5.42|5.39|

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/FrameHeaders.Generated.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/FrameHeaders.Generated.cs
@@ -3703,447 +3703,446 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Http
                 _bits = 0;
                 return;
             }
-
             
-                if (((_bits & 1048576L) != 0))
+            if (((_bits & 1048576L) != 0))
+            {
+                _headers._Accept = default(StringValues);
+                _bits &= ~1048576L;
+                if(_bits == 0)
                 {
-                    _headers._Accept = default(StringValues);
-                    _bits &= ~1048576L;
-                    if(_bits == 0)
-                    {
-                        return;
-                    }
+                    return;
                 }
+            }
             
-                if (((_bits & 268435456L) != 0))
+            if (((_bits & 268435456L) != 0))
+            {
+                _headers._Host = default(StringValues);
+                _bits &= ~268435456L;
+                if(_bits == 0)
                 {
-                    _headers._Host = default(StringValues);
-                    _bits &= ~268435456L;
-                    if(_bits == 0)
-                    {
-                        return;
-                    }
+                    return;
                 }
+            }
             
-                if (((_bits & 1099511627776L) != 0))
+            if (((_bits & 1099511627776L) != 0))
+            {
+                _headers._UserAgent = default(StringValues);
+                _bits &= ~1099511627776L;
+                if(_bits == 0)
                 {
-                    _headers._UserAgent = default(StringValues);
-                    _bits &= ~1099511627776L;
-                    if(_bits == 0)
-                    {
-                        return;
-                    }
+                    return;
                 }
+            }
             
-                if (((_bits & 1L) != 0))
+            if (((_bits & 1L) != 0))
+            {
+                _headers._CacheControl = default(StringValues);
+                _bits &= ~1L;
+                if(_bits == 0)
                 {
-                    _headers._CacheControl = default(StringValues);
-                    _bits &= ~1L;
-                    if(_bits == 0)
-                    {
-                        return;
-                    }
+                    return;
                 }
+            }
             
-                if (((_bits & 2L) != 0))
+            if (((_bits & 2L) != 0))
+            {
+                _headers._Connection = default(StringValues);
+                _bits &= ~2L;
+                if(_bits == 0)
                 {
-                    _headers._Connection = default(StringValues);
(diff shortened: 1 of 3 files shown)


tools/Microsoft.AspNetCore.Server.Kestrel.GeneratedCode/KnownHeaders.cs
```diff
@@ -411,12 +439,83 @@ protected void CopyToFast(ref MemoryPoolIterator output)
...
         public unsafe void Append(byte[] keyBytes, int keyOffset, int keyLength, string value)
         {{
+            fixed (byte* ptr = &keyBytes[keyOffset])
+            {{
+                var pUB = ptr;
+                var pUL = (ulong*)pUB;
+                var pUI = (uint*)pUB;
+                var pUS = (ushort*)pUB;
+                switch (keyLength)
+                {{{Each(loop.Headers.Where(h => h.PrimaryHeader).GroupBy(x => x.Name.Length), byLength => $@"
```
halter73: Is this doing anything different than `AppendNonPrimaryHeaders`? Couldn't loop.Headers just be reordered to have put the PrimaryHeaders first?

benaadams: `GroupBy(x => x.Name.Length)` breaks that ordering

tools/Microsoft.AspNetCore.Server.Kestrel.GeneratedCode/KnownHeaders.cs
```diff
@@ -366,10 +384,19 @@ protected override bool RemoveFast(string key)
...
         protected override void ClearFast()
-        {{
-            _bits = 0;
-            _headers = default(HeaderReferences);
+        {{            
             MaybeUnknown?.Clear();
+            {Each(loop.Headers.OrderBy(h => !h.PrimaryHeader), header => $@"
+                if ({header.TestBit()})
+                {{
+                    _headers._{header.Identifier} = default(StringValues);
```
halter73: Is there a micro benchmark showing this is faster than `_headers = default(HeaderReferences);`? Is there any number of set headers where this is no longer the case?

tools/Microsoft.AspNetCore.Server.Kestrel.GeneratedCode/KnownHeaders.cs
```diff
@@ -395,7 +422,8 @@ protected override void CopyToFast(KeyValuePair<string, StringValues>[] array, i
         {(loop.ClassName == "FrameResponseHeaders" ? $@"
         protected void CopyToFast(ref MemoryPoolIterator output)
         {{
-            {Each(loop.Headers, header => $@"
+            var tempBits = _bits;
+            {Each(loop.Headers.Where(h => h.PrimaryHeader), header => $@"
```
halter73: Why not orderby and remove CopyToFastNonPrimary?

tools/Microsoft.AspNetCore.Server.Kestrel.GeneratedCode/KnownHeaders.cs
halter73: I would prefer it if the template could be parameterized with the list of headers as the argument instead of copy/pasted.

halter73:
Normally I would worried about over fitting to the plaintext benchmark, but this doesn't seem to make performance worse when there are more headers. I'm curious about whether ClearFast doesn't become slower with more headers though.

tools/Microsoft.AspNetCore.Server.Kestrel.GeneratedCode/KnownHeaders.cs
benaadams: The HeaderReferences is a pretty big struct - does the `= default()` change improve after this tweak in coreclr? https://github.com/dotnet/coreclr/pull/7198

Alas its not currently part of 1.1.0; was going to be added https://github.com/dotnet/coreclr/pull/7341 but..?

nathana1: I haven't tested that change yet, but moving to 1.1.0 looks like a fairly consistent 1.5-2% win.  Since JIT_Memset was showing up in the profiles for ClearFast that could be a win for the current implementation.

nathana1:
New commit rebases against dev, refactors Append to have template string and adds a bailout to ClearFast if there's too many (12) headers.

halter73:
LGTM. @cesarbs @pakrym Mind taking a quick look?

@nathana1 can you rebase again? Thanks.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/FrameHeaders.Generated.cs
```diff
@@ -4601,6 +5097,8 @@ public unsafe void Append(byte[] keyBytes, int keyOffset, int keyLength, string
                             }
                         }
                         break;
+
```
pakrym: nit: remove these linebreaks

pakrym:
:shipit:

Issue #1086, opened 2016-09-06
Include port in EADDRINUSE exception


DovydasNavickas:
It is not the first time I get this exception:
```Microsoft.AspNetCore.Server.Kestrel.Internal.Networking.UvException: Error -4091 EADDRINUSE address already in use```

But this time I think I set the settings right and I can't seem to find the actual port Kestrel tried to use, which makes it impossible (or I don't know how) to check which port Kestrel tried to use.

It would be awesome if an exception included the port number Kestrel tried to use.

khellang:
I really don't think this exception (or any UvException for that matter) should be surfaced to the user. What about a more specific exception here?

khellang:
I think https://github.com/aspnet/KestrelHttpServer/issues/299 might've been closed a bit prematurely. I'm still seeing ECONNRESET exceptions occasionally as well.

halter73:
@khellang Are you getting the ECONNRESET for reads or writes?

It's writes that should no longer fail for this reason. It's possible for reads of course if you don't get a complete request. In that case the UvException is an inner exception wrapped in an IOException.

khellang:
@halter73 I think I saw it while awaiting a `Task.Delay`. Are they supposed to be logged if the client aborts the request?

halter73:
@khellang ECONNRESETs should be logged as information like any other communication error.

DovydasNavickas:
Soooo.... There shouldn't be any UvException surfaced to the user in the future builds? Or what is the bottom line of this?

khellang:
I think there might be different interpretations on what "surfaced to the user" means.

People seem to mean two different things:

 1. A `UvException` is thrown and unhandled before reaching user-land.
 2. A `UvException` is throw, handled and logged as information.

I'm not really sure any of those should occur :confused:

DovydasNavickas:
I think `UvException` could be left as an `InnerException`, because throwing information out when you have it seems like a bad idea, but the user should see a friendly `IOException` telling everything one needs to take action and not something that makes one go "Whaaaat?".
In this case (`EADDRINUSE`), exception should be an `IOException` saying that this particular port is in use.

khellang:
@DovydasNavickas That's taken care of by #1143, but the `ECONNRESET` messages still gets logged.

DovydasNavickas:
Ok, got it. Missed the reference.

PR #1143, opened 2016-10-03
Include address in error message when address already in use (#1086)

cesarbs:
#1086

@halter73 @mikeharder

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Kestrel/KestrelServer.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/KestrelServer.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.IO;
 using System.Linq;
 using System.Reflection;
 using Microsoft.AspNetCore.Hosting;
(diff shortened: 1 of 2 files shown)


cesarbs:
@khellang

src/Microsoft.AspNetCore.Server.Kestrel/KestrelServer.cs
```diff
@@ -109,8 +110,19 @@ public void Start<TContext>(IHttpApplication<TContext> application)
...
-                            parsedAddress));
+                        try
+                        {
+                            _disposables.Push(engine.CreateServer(parsedAddress));
+                        }
+                        catch (AggregateException ex)
+                        {
+                            if ((ex.InnerException as UvException)?.StatusCode == Constants.EADDRINUSE)
+                            {
+                                throw new IOException($"Failed to bind to address {parsedAddress}: address already in use.", ex);
```
halter73: IOException? Would WebListener throw an IOException?

Tratcher: Startup is weird, there's really no standard exception you can throw. IO isn't terrible, at least it's related to the socket.

halter73: @Tratcher Do you know what WebListener would throw?

Tratcher: it's just a WebListenerException

halter73: It's technically a breaking change not to keep this a AggregateException/UvException, but AggregateExceptions are gross enough that I don't think anyone is catching that explicitly. I guess IOException is fine.

cesarbs: cc @Eilon @DamianEdwards

Eilon: Yeah I think this is fine. Doesn't seem like a case where someone would catch the specific exception anyway (or at all).

PR #1148, opened 2016-10-04
Fix 5% regression in techempower plaintext after change to SocketOutput

halter73:
The main improvement is that the _contextLock is never reacquired on the threadpool to return written blocks. This seems to make a significant difference on better hardware.

#1145 

The improvement on a 48 logical core machine is pretty significant. Unfortunately, I'm getting RPS numbers as high as those @nathana1 mentioned in the issue before or after the change. I wonder if there's been another regression since @nathana1, or if I'm doing something wrong measuring.

@nathana1 I left my session open on the server if you want to check it out. Thanks.

### dev

    $ ./multi-run.sh --depth 16
    ./run-wrk.sh -p 5000 -h 192.168.1.100 -t plaintext --script ../pipeline.lua -d 1m --depth 16 -c 512
    Warming up...
    4470347.84
    Running tests...
    4453736.01
    4453718.90
    4420181.20
    4434269.48
    4508274.92
    4488297.48
    4399945.67
    4466734.33
    4593534.31
    4531329.83
    4453736.01 4453718.90 4420181.20 4434269.48 4508274.92 4488297.48 4399945.67 4466734.33 4593534.31 4531329.83

### halter73/1145

    $ ./multi-run.sh --depth 16
    ./run-wrk.sh -p 5000 -h 192.168.1.100 -t plaintext --script ../pipeline.lua -d 1m --depth 16 -c 512
    Warming up...
    4699696.24
    Running tests...
    4758701.48
    4733957.38
    4707075.01
    4707825.09
    4753937.18
    4763750.12
    4673806.49
    4691102.53
    4703620.58
    4753383.33
    4758701.48 4733957.38 4707075.01 4707825.09 4753937.18 4763750.12 4673806.49 4691102.53 4703620.58 4753383.33

PR #1149, opened 2016-10-04
Improve perf by not taking SocketOutput._contextLock when copying buffers

halter73:
The downside is that calls to `SocketOutput.WriteAsync` are no longer "atomic" when called concurrently, but I would argue that `Response.WriteAsync` shouldn't be called concurrently.

After seeing the performance improvement that could be achieved in #1148 by taking the `_contextLock` less, I decided to try to improve perf by not taking SocketOutput._contextLock when copying buffers. I had this idea before, but the improvement now seems to be measurable on better hardware.
### halter73/1145

```
$ ./multi-run.sh --depth 16
./run-wrk.sh -p 5000 -h 192.168.1.100 -t plaintext --script ../pipeline.lua -d 1m --depth 16 -c 512
Warming up...
4699696.24
Running tests...
4758701.48
4733957.38
4707075.01
4707825.09
4753937.18
4763750.12
4673806.49
4691102.53
4703620.58
4753383.33
4758701.48 4733957.38 4707075.01 4707825.09 4753937.18 4763750.12 4673806.49 4691102.53 4703620.58 4753383.33
```
### halter73/1145 + halter73/dont-write-at-the-same-time

```
$ ./multi-run.sh --depth 16
./run-wrk.sh -p 5000 -h 192.168.1.100 -t plaintext --script ../pipeline.lua -d 1m --depth 16 -c 512
Warming up...
4706377.81
Running tests...
4812169.18
4818584.28
4848651.40
4780047.91
4721708.67
4759027.05
4829114.36
4827255.34
4830057.89
4798915.02
4812169.18 4818584.28 4848651.40 4780047.91 4721708.67 4759027.05 4829114.36 4827255.34 4830057.89 4798915.02
```

@nathana1 @davidfowl @cesarbs

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/SocketOutput.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/SocketOutput.cs
@@ -90,39 +90,43 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Http
             TaskCompletionSource<object> tcs = null;
             var scheduleWrite = false;
 
-            lock (_contextLock)
-            {
-                if (_socket.IsClosed)
-                {
-                    _log.ConnectionDisconnectedWrite(_connectionId, buffer.Count, _lastWriteError);
+            var chunkedBytesPreCompleted = 0;
 
+            if (_socket.IsClosed)
+            {
+                _log.ConnectionDisconnectedWrite(_connectionId, buffer.Count, _lastWriteError);
+
+                return TaskCache.CompletedTask;
+            }
+
+            if (buffer.Count > 0)
+            {
+                var tail = ProducingStart();
+                if (tail.IsDefault)
+                {
                     return TaskCache.CompletedTask;
                 }
 
-                if (buffer.Count > 0)
+                if (chunk)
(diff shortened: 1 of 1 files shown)


nathana1:
Looks like you were using the default threadCount, the optimal threadCount for plaintext on this machine is 8.  When I reran with threadCount=8 I got about 4.9M, so about the new baseline.  Is the benchmark repo on that machine with or without your changes?

halter73:
@nathana1 The benchmarks repo on the test server has a project reference to Kestrel. You can double check via git, but I think the halter73/1145 Kestrel branch is currently checked out with no further changes. I saw that halter73/1145 hat noticeably better RPS than dev.

nathana1:
Yep, it's halter73/1145.

benaadams:
@nathana1 are you seeing and difference going against coreclr 1.1.0 vs 1.0.1?

nathana1:
It looks like coreclr 1.1.0 is about 1.5%-2% faster.

halter73:
The perf improvement probably isn't dramatic enough for the lower safety (i.e. the possibility of really weird behavior given parallel writes).

PR #1151, opened 2016-10-05
Add CodeCoverage to tests

benaadams:


cesarbs:
I thought this was going to make code coverage work on VS, but it already works there. What does this enable exactly?

benaadams:
@cesarbs is that in an internal VS upgrade? Without it it currently outputs

codecoverage

cesarbs:
@benaadams Sorry, that's right, I should have cleared my package cache before checking the behavior in dev 😁

Issue #1147, opened 2016-10-04
TCP Loopback Fast Path causes occasional connection hangs


halter73:
_Edit by @cesarbs: original issue title was 'ZeroContentLengthSetAutomaticallyForNonKeepAliveRequests test failure'_

I've only seen this once so far.

    xUnit.net .NET CLI test runner (64-bit Desktop .NET win10-x64)
      Discovering: Microsoft.AspNetCore.Server.KestrelTests
      Discovered:  Microsoft.AspNetCore.Server.KestrelTests
      Starting:    Microsoft.AspNetCore.Server.KestrelTests
        Microsoft.AspNetCore.Server.KestrelTests.EngineTests.ZeroContentLengthSetAutomaticallyForNonKeepAliveRequests(testContext: TestServiceContext { App = null, AppLifetime = LifetimeNotImplemented { ApplicationStarted = (throws NotImplementedException), ApplicationStopped = (throws NotImplementedException), ApplicationStopping = (throws NotImplementedException) }, DateHeaderValue = "Tue, 04 Oct 2016 02:39:11 GMT", DateHeaderValueManager = DateHeaderValueManager { }, FrameFactory = null, ... }) [FAIL]
          System.IO.IOException : Unable to read data from the transport connection: An existing connection was forcibly closed by the remote host.
          ---- System.Net.Sockets.SocketException : An existing connection was forcibly closed by the remote host
          Stack Trace:
               at System.Net.Sockets.NetworkStream.EndRead(IAsyncResult asyncResult)
               at System.Threading.Tasks.TaskFactory`1.FromAsyncTrimPromise`1.Complete(TInstance thisRef, Func`3 endMethod, IAsyncResult asyncResult, Boolean requiresSynchronization)
            --- End of stack trace from previous location where exception was thrown ---
               at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
               at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
               at System.IO.StreamReader.<ReadBufferAsync>d__97.MoveNext()
            --- End of stack trace from previous location where exception was thrown ---
               at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
               at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
[...]

PR #1152, opened 2016-10-05
Remove TCP Loopback Fast Path code (#1147).

cesarbs:
See https://github.com/aspnet/KestrelHttpServer/issues/1147#issuecomment-251817618 for reason for this change.

Considering the issue seen in test code, I think we should remove this from server code too. If we want to keep it, we should add an option to enable/disable it.

@halter73 @mikeharder @davidfowl

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Networking/Libuv.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Networking/Libuv.cs
@@ -199,40 +199,6 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Networking
         {
             handle.Validate();
             ThrowIfErrored(_uv_tcp_bind(handle, ref addr, flags));
-            if (PlatformApis.IsWindows)
-            {
-                tcp_bind_windows_extras(handle);
-            }
-        }
-
-        private unsafe void tcp_bind_windows_extras(UvTcpHandle handle)
-        {
-            const int SIO_LOOPBACK_FAST_PATH = -1744830448; // IOC_IN | IOC_WS2 | 16;
-            const int WSAEOPNOTSUPP = 10000 + 45; // (WSABASEERR+45)
-            const int SOCKET_ERROR = -1;
-
-            var socket = IntPtr.Zero;
-            ThrowIfErrored(_uv_fileno(handle, ref socket));
-
-            // Enable loopback fast-path for lower latency for localhost comms, like HttpPlatformHandler fronting
-            // http://blogs.technet.com/b/wincat/archive/2012/12/05/fast-tcp-loopback-performance-and-low-latency-with-windows-server-2012-tcp-loopback-fast-path.aspx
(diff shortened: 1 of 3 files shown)


mikeharder:
I think we should understand this better before removing the optimization from server code.  Is your theory that loopback fast path is buggy?  If so, we should try to get an isolated repro and report it to Windows Networking.

Also, we should understand the performance impact of disabling this optimization.

Tratcher:
The perf impact should be minimal because it also required the client to opt into it, and we're not aware of any clients that have.

benaadams:
Netling does https://github.com/hallatore/Netling/blob/master/Netling.Core/Performance/HttpWorker.cs#L217-L218

cesarbs:
Repro:

### Server

```csharp
using System;
using System.Net;
using System.Net.Sockets;
using System.Text;

namespace Server
{
    class Program
    {
        static void Main(string[] args)
        {
            {
                socket.Bind(new IPEndPoint(IPAddress.Loopback, 5000));
                socket.Listen(128);

                const int SIO_LOOPBACK_FAST_PATH = -1744830448;
                var optionInValue = BitConverter.GetBytes(1);
                try
                {
                    socket.IOControl(SIO_LOOPBACK_FAST_PATH, optionInValue, null);
                }
                catch
                {
                }

                while (true)
                {
                    using (var client = socket.Accept())
                    {
                        var buffer = new byte[4];
                        var bytesReceived = 0;

                        while ((bytesReceived += client.Receive(buffer, bytesReceived, 4 - bytesReceived, SocketFlags.None)) < 4) ;
                        var request = Encoding.ASCII.GetString(buffer, 0, 4);
                        Console.WriteLine(request);
                        client.Send(Encoding.ASCII.GetBytes("pong"));
                    }
                }
            }
        }
    }
}
```

### Client

```csharp
using System;
using System.Net;
using System.Net.Sockets;
using System.Text;
using System.Threading;

namespace Client
{
    class Program
    {
        static void Main(string[] args)
        {
            while (true)
            {
                {
                    const int SIO_LOOPBACK_FAST_PATH = -1744830448;
                    var optionInValue = BitConverter.GetBytes(1);
                    try
                    {
                        socket.IOControl(SIO_LOOPBACK_FAST_PATH, optionInValue, null);
                    }
                    catch
                    {
                    }

                    socket.Connect(new IPEndPoint(IPAddress.Loopback, 5000));
                    socket.Send(Encoding.ASCII.GetBytes("ping"));

                    var buffer = new byte[4];
                    var bytesReceived = 0;

                    while ((bytesReceived += socket.Receive(buffer, bytesReceived, 4 - bytesReceived, SocketFlags.None)) < 4) ;
                    if (socket.Receive(buffer) != 0) throw new Exception();
                    var response = Encoding.ASCII.GetString(buffer, 0, 4);
                    Console.WriteLine(response);
                }
            }
        }
    }
}
```

halter73:
It seems far safer to remove it than to add it. Even netty should fallback to normal loopback. :shipit:

PR #1153, opened 2016-10-07
Interlocked.Read -> Volatile.Read

benaadams:
`Interlocked.Read` is a much more heavyweight operation than a `Volatile.Read`;  and may cause a cpu pipeline stall. However there may be a reason it was chosen in preference?

`Volatile.Read` is (1op on x64)

``` csharp
public static long Read(ref long location)
{
    // The VM will replace this with a more efficient implementation.
    var value = location;
    Thread.MemoryBarrier();
    return value;
}
```

`Interlocked.Read` is (22 ops on x64)

``` csharp
public static long Read(ref long location)
{
    return Interlocked.CompareExchange(ref location, 0, 0);
}
```

/cc @cesarbs @halter73

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Connection.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Connection.cs
@@ -171,7 +171,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Http
         // Called on Libuv thread
         public void Tick(long timestamp)
         {
-            if (timestamp > Interlocked.Read(ref _timeoutTimestamp))
+            if (timestamp > PlatformApis.VolatileRead(ref _timeoutTimestamp))
             {
                 ConnectionControl.CancelTimeout();
 
--- a/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/DateHeaderValueManager.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/DateHeaderValueManager.cs
@@ -6,6 +6,7 @@ using System.Text;
 using System.Threading;
 using Microsoft.AspNetCore.Server.Kestrel.Internal.Infrastructure;
 using Microsoft.Net.Http.Headers;
+using Microsoft.AspNetCore.Server.Kestrel.Internal.Networking;
 
 namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Http
 {
@@ -158,7 +159,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Http
             }
 
             // No requests since the last timer tick, we need to check if we're beyond the idle threshold
-            if ((now.Ticks - Interlocked.Read(ref _lastRequestSeenTicks)) >= _timeWithoutRequestsUntilIdle.Ticks)
+            if ((now.Ticks - PlatformApis.VolatileRead(ref _lastRequestSeenTicks)) >= _timeWithoutRequestsUntilIdle.Ticks)
             {
                 // No requests since idle threshold so stop the timer if it's still running
                 StopTimer();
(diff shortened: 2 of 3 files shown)


benaadams:
If you are worried about torn reads, a safer alternative could be
```csharp
long timeout;
if (IntPtr.Size == 8)
{
    timeout = Volatile.Read(ref _timeoutTimestamp);
}
else
{
    timeout = Interlocked.Read(ref _timeoutTimestamp);
}
if (timestamp > timeout)
```

benaadams:
Added 32bit function swap; should inline and discard other branch

halter73:
@benaadams Can you explain why Interlocked.Read is more expensive? The [Interlocked.Read MSDN docs](https://msdn.microsoft.com/en-us/library/system.threading.interlocked.read%28v=vs.110%29.aspx?f=255&MSPPError=-2147217396) go as far as to say that "The [Interlocked.]Read method is unnecessary on 64-bit systems" whereas the [Volatile.Read MSDN docs](https://msdn.microsoft.com/en-us/library/gg712759(v=vs.110).aspx) speaks of inserting memory barriers. Based on this, I would have though Volatile.Read to be more expensive.

I know you've looked at the implantation though, and I haven't.

benaadams:
`Volatile.Read` becomes just a read on x64 (though a forced read from memory rather than register)

`Interlocked.Read` stays as an `Interlocked.CompareExchange(ref location, 0, 0)` which likely fails but returns the in memory value. However it also has a 22 op latency; whereas Volatile read is 1 op.

i.e. the `Interlocked.Read` is unnecessary on x64; but it isn't replaced by anything more efficient, however `Volatile.Read` is.

I'm not sure if you could use a `Volatile.Write`/`Interlocked.Exchange` with the `Volatile.Read` on x86. There would be 2 memory barriers; but I'm not sure that would be enough to prevent a tear.

halter73:
@DamianEdwards I think someone should update those MSDN docs. Got anyone I should contact?

benaadams:
@halter73 it does say "On systems that require it, inserts a memory barrier"

halter73:
I'm OK with merging this. @cesarbs @pakrym thoughts?

cesarbs:
Wanna check perf first. Will benchmark in a minute.

cesarbs:
@benaadams Can you rebase?

cesarbs:
There appears to be a moderate perf hit from this change. Average of 5 plaintext runs on dev gives me 
1266933.98 RPS, while average of 5 runs on this PR is 1202942.508 RPS. Asked @benaadams to rebase to make sure it's not because of this PR not being based off the latest changes in dev.

halter73:
@cesarbs A microbenchmark would probably be more appropriate.

benaadams:
@cesarbs extra ref might not have help, so did a force inline; and rebased.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Connection.cs
```diff
@@ -171,7 +171,7 @@ public virtual void OnSocketClosed()
         // Called on Libuv thread
         public void Tick(long timestamp)
         {
-            if (timestamp > Interlocked.Read(ref _timeoutTimestamp))
+            if (timestamp > SystemClock.VolatileRead(ref _timeoutTimestamp))
```
pakrym: Can we move it to some other class? Usages outside `SystemClock` look very weird.

cesarbs:
@benaadams Perf looks better after rebase - average of 5 runs is 1246468.756 RPS. But I agree with @pakrym that the calls look weird in `SystemClock`.

benaadams:
Struggling on a name... obv `Volatile` and `Interlocked` are taken. Could change it to `TimestampRead` in clock.

A `VolatileRead` extension method on `long` would have been nice; but C# doesn't support ref extensions (though interestingly VB.NET does)

benaadams:
`NonTorn.Read`?

pakrym:
`PlatformApis.VolatileReadLong` ?

benaadams:
Done

pakrym:
LGTM

cesarbs:
:shipit:

PR #1155, opened 2016-10-07
Handle response content length mismatches (#175)

cesarbs:
#175 

- [x] TODO: `InvalidOperationException` might not be the right exception type for this
- [x] TODO: is logging this as application errors OK or should we add new event IDs for response length mismatches?

@halter73 @Tratcher @mikeharder @davidfowl

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
@@ -4,6 +4,7 @@
 using System;
 using System.Collections.Generic;
 using System.Diagnostics;
+using System.Globalization;
 using System.IO;
 using System.Linq;
 using System.Net;
@@ -75,7 +76,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Http
         protected readonly long _keepAliveMilliseconds;
         private readonly long _requestHeadersTimeoutMilliseconds;
 
-        private int _responseBytesWritten;
+        protected long _responseBytesWritten;
 
         public Frame(ConnectionContext context)
         {
@@ -516,8 +517,8 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Http
 
         public void Write(ArraySegment<byte> data)
         {
+            VerifyAndUpdateWrite(data.Count);
             ProduceStartAndFireOnStarting().GetAwaiter().GetResult();
-            _responseBytesWritten += data.Count;
 
             if (_canHaveBody)
             {
@@ -547,7 +548,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Http
                 return WriteAsyncAwaited(data, cancellationToken);
             }
 
-            _responseBytesWritten += data.Count;
+            VerifyAndUpdateWrite(data.Count);
 
             if (_canHaveBody)
             {
@@ -573,8 +574,9 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Http
 
         public async Task WriteAsyncAwaited(ArraySegment<byte> data, CancellationToken cancellationToken)
         {
+            VerifyAndUpdateWrite(data.Count);
+
             await ProduceStartAndFireOnStarting();
-            _responseBytesWritten += data.Count;
 
             if (_canHaveBody)
             {
@@ -596,6 +598,23 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Http
                 HandleNonBodyResponseWrite();
                 return;
             }
+        }
+
+        private void VerifyAndUpdateWrite(int count)
+        {
+            var responseHeaders = FrameResponseHeaders;
+
+            if (responseHeaders != null &&
+                !responseHeaders.HasTransferEncoding &&
+                responseHeaders.HasContentLength &&
+                _responseBytesWritten + count > responseHeaders.HeaderContentLengthValue.Value)
+            {
+                _keepAlive = false;
+                throw new InvalidOperationException(
+                    $"Response Content-Length mismatch: too many bytes written ({_responseBytesWritten + count} of {responseHeaders.HeaderContentLengthValue.Value}).");
+            }
+
+            _responseBytesWritten += count;
         }
 
         private void WriteChunked(ArraySegment<byte> data)
--- a/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/FrameHeaders.Generated.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/FrameHeaders.Generated.cs
@@ -3697,6 +3697,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Http
         {
             _bits = 0;
             _headers = default(HeaderReferences);
+            
             MaybeUnknown?.Clear();
         }
 
@@ -5670,6 +5671,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Http
             }
             set
             {
+                _contentLength = ParseContentLength(value);
                 _bits |= 2048L;
                 _headers._ContentLength = value; 
                 _headers._rawContentLength = null;
@@ -7384,6 +7386,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Http
                     {
                         if ("Content-Length".Equals(key, StringComparison.OrdinalIgnoreCase))
                         {
+                            _contentLength = ParseContentLength(value);
                             _bits |= 2048L;
                             _headers._ContentLength = value;
                             _headers._rawContentLength = null;
@@ -7809,6 +7812,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Http
                             {
                                 ThrowDuplicateKeyException();
                             }
+                            _contentLength = ParseContentLength(value);
                             _bits |= 2048L;
                             _headers._ContentLength = value;
                             _headers._rawContentLength = null;
@@ -8350,6 +8354,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Http
                         {
                             if (((_bits & 2048L) != 0))
                             {
+                                _contentLength = null;
                                 _bits &= ~2048L;
                                 _headers._ContentLength = StringValues.Empty;
                                 _headers._rawContentLength = null;
@@ -8601,6 +8606,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Http
         {
             _bits = 0;
             _headers = default(HeaderReferences);
+            _contentLength = null;
             MaybeUnknown?.Clear();
         }
 
--- a/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/FrameHeaders.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/FrameHeaders.cs
@@ -4,6 +4,7 @@
 using System;
 using System.Collections;
 using System.Collections.Generic;
+using System.Globalization;
 using System.Linq;
 using Microsoft.AspNetCore.Http;
 using Microsoft.Extensions.Primitives;
@@ -229,6 +230,18 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Http
                         ThrowInvalidHeaderCharacter(ch);
                     }
                 }
+            }
+        }
+
+        public static long ParseContentLength(StringValues value)
+        {
+            try
+            {
+                return long.Parse(value, NumberStyles.AllowLeadingWhite | NumberStyles.AllowTrailingWhite, CultureInfo.InvariantCulture);
+            }
+            catch (FormatException ex)
+            {
+                throw new InvalidOperationException("Content-Length value must be an integral number.", ex);
             }
         }
 
--- a/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/FrameOfT.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/FrameOfT.cs
@@ -92,6 +92,16 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Http
                         try
                         {
                             await _application.ProcessRequestAsync(context).ConfigureAwait(false);
+
+                            var responseHeaders = FrameResponseHeaders;
+                            if (!responseHeaders.HasTransferEncoding &&
+                                responseHeaders.HasContentLength &&
+                                _responseBytesWritten < responseHeaders.HeaderContentLengthValue.Value)
+                            {
+                                _keepAlive = false;
+                                ReportApplicationError(new InvalidOperationException(
+                                    $"Response Content-Length mismatch: too few bytes written ({_responseBytesWritten} of {responseHeaders.HeaderContentLengthValue.Value})."));
+                            }
                         }
                         catch (Exception ex)
                         {
--- a/test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/ResponseTests.cs
+++ b/test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/ResponseTests.cs
@@ -2,6 +2,7 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.IO;
 using System.Linq;
 using System.Net;
 using System.Net.Http;
@@ -14,6 +15,7 @@ using Microsoft.AspNetCore.Server.Kestrel.Internal.Http;
 using Microsoft.AspNetCore.Server.Kestrel.Internal.Infrastructure;
 using Microsoft.AspNetCore.Testing;
 using Microsoft.Extensions.Internal;
+using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Primitives;
 using Moq;
 using Xunit;
@@ -85,7 +87,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.FunctionalTests
                     app.Run(async context =>
                     {
                         context.Response.Headers.Add(headerName, headerValue);
-                        
+
                         await context.Response.WriteAsync("");
                     });
                 });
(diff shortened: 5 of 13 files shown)


src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
```diff
@@ -598,6 +604,34 @@ public Task WriteAsync(ArraySegment<byte> data, CancellationToken cancellationTo
...
+        }
+
+        protected void TryInitializeResponseContentLength()
+        {
+            if (!_responseContentLength.HasValue)
+            {
+                var responseHeaders = FrameResponseHeaders;
+                if (responseHeaders != null && responseHeaders.HasContentLength)
+                {
+                    _responseContentLength = int.Parse(responseHeaders.HeaderContentLength.ToString());
```
Tratcher: TryParse, culture, format

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
```diff
@@ -598,6 +604,34 @@ public Task WriteAsync(ArraySegment<byte> data, CancellationToken cancellationTo
...
+
+            if (_responseContentLength.HasValue && _responseBytesWritten > _responseContentLength.Value)
+            {
+                ThrowContentLengthMismatchException();
+            }
+        }
+
+        protected void TryInitializeResponseContentLength()
+        {
+            if (!_responseContentLength.HasValue)
```
Tratcher: If there's no content-length header this will re-run for every write. You need a flag to only run it once, or is HasContentLength really that fast?

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/FrameOfT.cs
```diff
@@ -92,6 +92,13 @@ public class Frame<TContext> : Frame
                         try
                         {
                             await _application.ProcessRequestAsync(context).ConfigureAwait(false);
+
+                            TryInitializeResponseContentLength();
+
+                            if (_responseContentLength.HasValue && _responseBytesWritten < _responseContentLength.Value)
+                            {
+                                ThrowContentLengthMismatchException();
```
Tratcher: This should be a different message "Too few bytes written".

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
```diff
@@ -598,6 +604,34 @@ public Task WriteAsync(ArraySegment<byte> data, CancellationToken cancellationTo
...
+                {
+                    _responseContentLength = int.Parse(responseHeaders.HeaderContentLength.ToString());
+                }
+            }
+        }
+
+        protected void ThrowContentLengthMismatchException()
+        {
+            _keepAlive = false;
+            throw new InvalidOperationException($"Response content length mismatch: wrote {_responseBytesWritten} bytes to {_responseContentLength.Value}-byte response.");
```
Tratcher: Too many bytes written

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
```diff
@@ -548,6 +551,7 @@ public Task WriteAsync(ArraySegment<byte> data, CancellationToken cancellationTo
             }
 
             _responseBytesWritten += data.Count;
+            VerifyWrite();
```
Tratcher: this calling pattern is misleading, _responseBytesWritten gets incremented regardless of if you've written the bytes out. Invert it:
```
             VerifyWrite(data.Count);
             _responseBytesWritten += data.Count;
```

and/or move `_responseBytesWritten += data.Count;` inside of VerifyWrite and make it VerifyAndUpdateWrite

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/FrameOfT.cs
Tratcher: Why even throw at this point? Why not just directly invoke ReportApplicationError?

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
Tratcher: Yes, Stream allows InvalidOperationException. 
https://msdn.microsoft.com/en-us/library/hh137799(v=vs.110).aspx

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/FrameOfT.cs
cesarbs: Why have two different messages when we can have a single one that reports the error in the same way?

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
cesarbs: I'll leave it as `Parse` since that will throw if the app does something like `                httpContext.Response.Headers["Content-Length"] = "foo";`

cesarbs:
@Tratcher I've addressed your comments.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
```diff
@@ -598,6 +602,32 @@ public Task WriteAsync(ArraySegment<byte> data, CancellationToken cancellationTo
...
+        }
+
+        protected void TryInitializeResponseContentLength()
+        {
+            if (!_responseContentLength.HasValue)
+            {
+                var responseHeaders = FrameResponseHeaders;
+                if (responseHeaders != null && responseHeaders.HasContentLength)
+                {
+                    _responseContentLength = long.Parse(responseHeaders.HeaderContentLength.ToString(), NumberStyles.None);
```
Tratcher: This will result in a FormatException being thrown from Write. Use TryParse and ignore values you can't parse.
When determining if you should chunk, do you check for the presence of a content-length or if you understand it?

cesarbs: We check for its presence but we don't try to parse it.

I disagree with you on this point. We should throw if the `Content-Length` header is set to an invalid value. The whole point of this change is to track errors related to response `Content-Length`, this would be one of them. Without a valid value, we can't track it.

Tratcher: Even if you want to throw, you can't throw a FormatException from Stream.Write. It needs to be wrapped in an InvalidOperationException

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
halter73: In that case you should throw from the header setter ideally.

I think we should make sure we have the [same logic we use in HttpAbstractions](https://github.com/aspnet/HttpAbstractions/blob/45696535048559188dce8b38a9fde82baa88893c/src/Microsoft.AspNetCore.Http/Internal/ParsingHelpers.cs#L407).  We should probably look into sharing this with HttpAbsractions somehow.

It's also a good opportunity to make sure we use the same logic for parsing the request header content-length. Right now, it seems like we're probably more lax than we should be.

halter73: Oh. And I see you changed this to a long 👍 Were there not any tests failing when this was int.Parse?

cesarbs:
Added validation to header setter. Not sure I covered all bases though. I'm thinking of adding a `long _contentLength` field to `FrameHeaders`, this would give fast access to that value instead of having to parse it from `HeaderContentLength`.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/FrameHeaders.Generated.cs
```diff
@@ -201,6 +202,7 @@ public StringValues HeaderContentLength
             }
             set
             {
+                ValidateLongValue(value);
```
Tratcher: can you cache the parsed value internally rather than re-parse it later?

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
halter73: Wow. I thought our LargeDownload test wrote over 2GB like the LargeUpload test. It looks like the LargeDownload test only has a 1MB body.

We should change the LargeDownload test to be like the LargeUpload test where we check every byte value in a few megabyte scenario, but still ensure that sending over 2GB works.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/FrameHeaders.cs
```diff
@@ -232,6 +233,11 @@ public static void ValidateHeaderCharacters(string headerCharacters)
             }
         }
 
+        public static void ValidateLongValue(StringValues value)
+        {
+            long.Parse(value, NumberStyles.AllowLeadingWhite | NumberStyles.AllowTrailingWhite, CultureInfo.InvariantCulture);
```
halter73: Might make things a little easier to debug if this was wrapped in an InvalidOperationException stating the Content-Length header must be a positive decimal number or something like this.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/FrameHeaders.Generated.cs
cesarbs: Yeah, it's what I was referring to in my comment above.

cesarbs:
Updated. Caching Content-Length and wrapping `FormatException` in `InvalidOperationException`.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/FrameHeaders.Generated.cs
```diff
@@ -12,6 +13,7 @@ public partial class FrameRequestHeaders
 
         private long _bits = 0;
         private HeaderReferences _headers;
+        private long? _contentLength;
```
Tratcher: Does this get cleared if I call `Headers.Remove("Content-Length")`?

cesarbs: Done.

cesarbs:
I'm applying the `FrameHeaders` change to `FrameResponseHeaders` only (https://github.com/aspnet/KestrelHttpServer/pull/1155/commits/1d3b71d6ab8442ce064879d2bbe6dac37192f34a)

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/ResponseTests.cs
```diff
@@ -324,7 +326,279 @@ public class ResponseTests
...
+                return TaskCache.CompletedTask;
+            }, serviceContext))
+            {
+                using (var connection = server.CreateConnection())
+                {
+                    await connection.Send(
+                        "GET / HTTP/1.1",
+                        "",
+                        "");
+                    await connection.ReceiveEnd(
```
Tratcher: Would this result in a RST? or just a FIN?

tools/Microsoft.AspNetCore.Server.Kestrel.GeneratedCode/KnownHeaders.cs
```diff
@@ -349,7 +361,8 @@ protected override bool RemoveFast(string key)
                         if (""{header.Name}"".Equals(key, StringComparison.OrdinalIgnoreCase))
                         {{
                             if ({header.TestBit()})
-                            {{
+                            {{{If(header.Identifier == "ContentLength", () => $@"
```
halter73: Missed `loop.ClassName == "FrameResponseHeaders" && ` here.

tools/Microsoft.AspNetCore.Server.Kestrel.GeneratedCode/KnownHeaders.cs
```diff
@@ -369,6 +382,7 @@ protected override void ClearFast()
         {{
             _bits = 0;
             _headers = default(HeaderReferences);
+            _contentLength = null;
```
halter73: Missed If here too.

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/ResponseTests.cs
cesarbs: FIN

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
```diff
@@ -684,7 +703,7 @@ private void ProduceStart(bool appCompleted)
 
         protected Task TryProduceInvalidRequestResponse()
         {
-            if (_requestRejected)
+            if (_requestRejected || _applicationException != null)
```
halter73: Why is this change necessary?

halter73: Obviously the application could swallow the exception, so we shouldn't be relying on it. And this change has little to do with invalid requests, right? I guess we're now parsing the request content-length more strictly...

cesarbs: If the response hasn't already started by the time we detect a mismatch in the content length, we can send a 500 response. `_applicationException` is set when there's a mismatch, so we need this check to run the error response logic.

halter73: That should be handled by the normal call to `ProduceEnd()`.

halter73:
I don't think we need a new event id for this. Treating it as a normal application error is fine with me.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
cesarbs: Indeed. My very first change (that I didn't even commit) was calling `Abort()` on a content length mismatch. This would prevent the call to `ProduceEnd()` in `FrameOfT`. But since I moved away from that decision, `ProduceEnd()` handles it as expected.

cesarbs:
Updated.

halter73:
:shipit:

tools/Microsoft.AspNetCore.Server.Kestrel.GeneratedCode/KnownHeaders.cs
```diff
@@ -228,12 +235,14 @@ public partial class {loop.ClassName}
...
             set
-            {{
+            {{{If(loop.ClassName == "FrameResponseHeaders" && header.Identifier == "ContentLength", () => $@"
+                _contentLength = ParseContentLength(value);")}
                 {header.SetBit()};
                 _headers._{header.Identifier} = value; {(header.EnhancedSetter == false ? "" : $@"
                 _headers._raw{header.Identifier} = null;")}
             }}
         }}")}
+        {(loop.ClassName == "FrameResponseHeaders" ? "public long? HeaderContentLengthValue => _contentLength;" : "")}
```
halter73: nit: I would probably put this in FrameResponseHeaders.cs

tools/Microsoft.AspNetCore.Server.Kestrel.GeneratedCode/KnownHeaders.cs
```diff
@@ -199,6 +204,7 @@ public static string GeneratedFile()
             return $@"
 using System;
 using System.Collections.Generic;
+using System.Globalization;
```
halter73: nit: I don't think this is being used.

tools/Microsoft.AspNetCore.Server.Kestrel.GeneratedCode/KnownHeaders.cs
```diff
@@ -304,7 +313,8 @@ protected override void SetValueFast(string key, StringValues value)
                 case {byLength.Key}:
                     {{{Each(byLength, header => $@"
                         if (""{header.Name}"".Equals(key, StringComparison.OrdinalIgnoreCase))
-                        {{
+                        {{{If(loop.ClassName == "FrameResponseHeaders" && header.Identifier == "ContentLength", () => $@"
```
halter73: nit: `$` is unnecessary.

tools/Microsoft.AspNetCore.Server.Kestrel.GeneratedCode/KnownHeaders.cs
```diff
@@ -328,7 +338,9 @@ protected override void AddValueFast(string key, StringValues value)
                             if ({header.TestBit()})
                             {{
                                 ThrowDuplicateKeyException();
-                            }}
+                            }}{
+                            If(loop.ClassName == "FrameResponseHeaders" && header.Identifier == "ContentLength", () => $@"
+                            _contentLength = ParseContentLength(value);")}
```
halter73: No `$` needed

tools/Microsoft.AspNetCore.Server.Kestrel.GeneratedCode/KnownHeaders.cs
```diff
@@ -228,12 +235,14 @@ public partial class {loop.ClassName}
                 return StringValues.Empty;
             }}
             set
-            {{
+            {{{If(loop.ClassName == "FrameResponseHeaders" && header.Identifier == "ContentLength", () => $@"
```
halter73: No `$`.

tools/Microsoft.AspNetCore.Server.Kestrel.GeneratedCode/KnownHeaders.cs
```diff
@@ -349,7 +361,8 @@ protected override bool RemoveFast(string key)
                         if (""{header.Name}"".Equals(key, StringComparison.OrdinalIgnoreCase))
                         {{
                             if ({header.TestBit()})
-                            {{
+                            {{{If(loop.ClassName == "FrameResponseHeaders" && header.Identifier == "ContentLength", () => $@"
```
halter73: ditto

cesarbs:
@halter73 nits addressed :)

PR #1158, opened 2016-10-08
Improve callstack for LoggingThreadPool

benaadams:
Output items in callstack as `<RunAction>b__7_0` rather than `ctor>b__4_0`

Before

After


halter73:
Sure. Why not? 🚢

cesarbs:
Will merge.

Issue #1157, opened 2016-10-07
TestKeepAliveTimeout (ConnectionKeptAliveBetweenRequests) failure


halter73:
I've seen this fail a few times on my debian 8.2 machine, but only when running the full functional test suite.

    Microsoft.AspNetCore.Server.Kestrel.FunctionalTests.KeepAliveTimeoutTests.TestKeepAliveTimeout [FAIL]
      System.IO.IOException : Unable to write data to the transport connection: Broken pipe.
      ---- System.Net.Sockets.SocketException : Broken pipe
      Stack Trace:
           at System.Net.Sockets.NetworkStream.BeginWrite(Byte[] buffer, Int32 offset, Int32 size, AsyncCallback callback, Object state)
           at System.Net.Sockets.NetworkStream.<>c.<WriteAsync>b__59_0(Byte[] bufferArg, Int32 offsetArg, Int32 sizeArg, AsyncCallback callback, Object state)
           at System.Threading.Tasks.TaskFactory`1.FromAsyncImpl[TArg1,TArg2,TArg3](Func`6 beginMethod, Func`2 endFunction, Action`1 endAction, TArg1 arg1, TArg2 arg2, TArg3 arg3, Object state, TaskCreationOptions creationOptions)
           at System.Threading.Tasks.TaskFactory.FromAsync[TArg1,TArg2,TArg3](Func`6 beginMethod, Action`1 endMethod, TArg1 arg1, TArg2 arg2, TArg3 arg3, Object state)
           at System.Net.Sockets.NetworkStream.WriteAsync(Byte[] buffer, Int32 offset, Int32 size, CancellationToken cancellationToken)
           at System.IO.StreamWriter.<FlushAsyncInternal>d__67.MoveNext()
        --- End of stack trace from previous location where exception was thrown ---
           at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
           at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
           at System.Runtime.CompilerServices.TaskAwaiter.GetResult()
        /home/shalter/KestrelHttpServer/test/shared/TestConnection.cs(77,0): at Microsoft.AspNetCore.Testing.TestConnection.<Send>d__8.MoveNext()
        --- End of stack trace from previous location where exception was thrown ---
           at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
[...]

PR #1160, opened 2016-10-11
Reduce delay variability in KeepAliveTimeoutTests.ConnectionKeptAliveBetweenRequests() (#1157)

cesarbs:
#1157

`Task.Delay()` times vary *a lot* when running all tests at once. `ConfigureAwait` does the trick.

### Before change

```
Slept for 5258
Slept for 4792
Slept for 5110
Slept for 3939
Slept for 8251
Slept for 13205 <-- blows up
```

### After change

```
Slept for 3000
Slept for 3002
Slept for 2999
Slept for 2998
Slept for 3001
Slept for 2997
Slept for 2999
Slept for 2999
Slept for 3002
Slept for 2997

```

@halter73

### DIFF
--- a/test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/KeepAliveTimeoutTests.cs
+++ b/test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/KeepAliveTimeoutTests.cs
@@ -21,6 +21,14 @@ namespace Microsoft.AspNetCore.Server.Kestrel.FunctionalTests
         [Fact]
         public async Task TestKeepAliveTimeout()
         {
+            // Delays in these tests cannot be much longer than expected.
+            // Call ConfigureAwait(false) to get rid of Xunit's synchronization context,
+            // otherwise it can cause unexpectedly longer delays when multiple tests
+            // are running in parallel. These tests becomes flaky on slower
+            // hardware because the continuations for the delay tasks might take too long to be
+            // scheduled if running on Xunit's synchronization context.
+            await Task.Delay(1).ConfigureAwait(false);
+
             var longRunningCancellationTokenSource = new CancellationTokenSource();
             var upgradeCancellationTokenSource = new CancellationTokenSource();
 


mikeharder:
Note there may be tradeoffs with setting `-maxthreads unlimited`.  I recommend reading the whole issue https://github.com/xunit/xunit/issues/864.  There is disagreement about the behavior xUnit should use by default.

halter73:
👍 for testing with `-maxthreads unlimited`. I can't add a reaction to the review summary for some reason.

cesarbs:
`-maxthreads unlimited` mostly helps, but there's still some variation:

```
Slept for 3047
Slept for 5417 <-- like here
Slept for 3013
Slept for 3001
Slept for 3032
Slept for 2998
Slept for 2999
Slept for 2998
Slept for 3002
Slept for 2998
```

Another sample:

```
Slept for 3015
Slept for 3048
Slept for 3021
Slept for 5278
Slept for 3173
Slept for 3068
Slept for 3007
Slept for 3117
Slept for 3139
Slept for 3002
```

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/KeepAliveTimeoutTests.cs
```diff
@@ -63,7 +63,7 @@ public class KeepAliveTimeoutTests
                         "GET / HTTP/1.1",
                         "",
                         "");
-                    await Task.Delay(ShortDelay);
```
halter73: I recommend putting `await Task.Delay(1).ConfigureAwait(false)` at the top of the `TestKeepAliveTimeout` to get everything of the xunit sync context.

cesarbs:
I'm back to my original solution.

1) `-maxthreads unlimited` helps a bit but there's still some wide variability in delay times
2) Separating each test into its own class per @mikeharder's suggestion didn't eliminate the eventual unexpectedly long delays
3) We're really only interested in not having a longer than expected delay in `ConnectionKeptAliveBetweenRequests`, so it makes sense to keep the `ConfigureAwait(false)` local to that single test

halter73:
I would like all the `Task.Delay`s in this class to be as accurate as they can be. Even if them running long wouldn't fail the test, it could make tests pass when the shouldn't (however unlikely). Also it could make the tests take longer than they should in general making the overall test run longer.

cesarbs:
Updated.

halter73:
You verified that this makes the Task.Delay in the sub-tests more reliable right? It definitely should, I just want to make sure you checked.

cesarbs:
@halter73 Yep, verified.

halter73:
:shipit:

PR #1161, opened 2016-10-11
Update cached Content-Length when setting raw header

cesarbs:
Missed this in https://github.com/aspnet/KestrelHttpServer/commit/f8813a600df30e018a6224048b31244be3d456b9 :(

@halter73 @Tratcher


halter73:
While this might not be strictly necessary, it keeps all the state in sync, so I'm a fan.

PR #1163, opened 2016-10-12
HEAD response can include Content-Length header

cesarbs:
StaticFiles tests are failing because of this.

@halter73 @mikeharder @Tratcher

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
@@ -617,6 +617,21 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Http
             _responseBytesWritten += count;
         }
 
+        protected void VerifyResponseContentLength()
+        {
+            var responseHeaders = FrameResponseHeaders;
+
+            if (!HttpMethods.IsHead(Method) &&
+                !responseHeaders.HasTransferEncoding &&
+                responseHeaders.HeaderContentLengthValue.HasValue &&
+                _responseBytesWritten < responseHeaders.HeaderContentLengthValue.Value)
+            {
+                _keepAlive = false;
+                ReportApplicationError(new InvalidOperationException(
+                    $"Response Content-Length mismatch: too few bytes written ({_responseBytesWritten} of {responseHeaders.HeaderContentLengthValue.Value})."));
+            }
+        }
+
         private void WriteChunked(ArraySegment<byte> data)
         {
             SocketOutput.Write(data, chunk: true);
--- a/test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/ResponseTests.cs
+++ b/test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/ResponseTests.cs
@@ -607,6 +607,62 @@ namespace Microsoft.AspNetCore.Server.Kestrel.FunctionalTests
             Assert.Equal(0, testLogger.ApplicationErrorsLogged);
         }
 
+        [Fact]
+        public async Task HeadResponseCanContainContentLengthHeader()
+        {
+            var testLogger = new TestApplicationErrorLogger();
+            var serviceContext = new TestServiceContext { Log = new TestKestrelTrace(testLogger) };
+
+            using (var server = new TestServer(httpContext =>
+            {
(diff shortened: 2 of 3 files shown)


src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
```diff
@@ -617,6 +617,21 @@ private void VerifyAndUpdateWrite(int count)
             _responseBytesWritten += count;
         }
 
+        protected void VerifyResponseContentLength()
+        {
+            var responseHeaders = FrameResponseHeaders;
+
+            if (!HttpMethods.IsHead(Method) &&
+                !responseHeaders.HasTransferEncoding &&
```
halter73: Do you have to check `!responseHeaders.HasTransferEncoding` since your are checking `responseHeaders.HasContentLength`?

halter73: And I prefer checking `responseHeaders.HeaderContentLengthValue` even though you fixed the set raw issue.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
```diff
@@ -617,6 +617,21 @@ private void VerifyAndUpdateWrite(int count)
             _responseBytesWritten += count;
         }
 
+        protected void VerifyResponseContentLength()
+        {
+            var responseHeaders = FrameResponseHeaders;
+
+            if (!HttpMethods.IsHead(Method) &&
```
halter73: Should you check that if `HttpMethods.IsHead(Method) && responseHeaders.ContentLengthValue.HasValue` then `_responseBytesWritten == 0 || _responseBytesWritten == responseHeaders.ContentLengthValue.Value`?

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
cesarbs: Yes. The spec allows both and `Transfer-Encoding` should take precedence.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
cesarbs: Even though we count those bytes, we ignore writes to HEAD responses. It would alert the app that there's an issue, but the error would surface anyways on a response to the same resource using a different method.

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/ResponseTests.cs
```diff
@@ -607,6 +607,34 @@ public class ResponseTests
...
 
+        [Fact]
+        public async Task HeadResponseCanContainContentLengthHeader()
+        {
+            var testLogger = new TestApplicationErrorLogger();
+            var serviceContext = new TestServiceContext { Log = new TestKestrelTrace(testLogger) };
+
+            using (var server = new TestServer(httpContext =>
+            {
+                httpContext.Response.ContentLength = 42;
```
halter73: We should have a HEAD test where we set the content-length and attempt to write a response body in the app. We have ResponseTests.ResponseBodyNotWrittenOnHeadResponseAndLoggedOnlyOnce, but it doesn't attempt to set a content-length.

cesarbs: Added.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
halter73: Yikes. I think the spec suggests removing the content-length header in this case which makes sense to me.

>        If a message is received with both a Transfer-Encoding and a
       Content-Length header field, the Transfer-Encoding overrides the
       Content-Length.  Such a message might indicate an attempt to
       perform request smuggling (Section 9.5) or response splitting
       (Section 9.4) and ought to be handled as an error.  A sender MUST
       remove the received Content-Length field prior to forwarding such
       a message downstream.

https://tools.ietf.org/html/rfc7230#section-3.3.3

@Tratcher thoughts?

halter73:
:shipit: b

PR #1166, opened 2016-10-14
Faster response Content-Length parsing.

cesarbs:
Fixes perf regression introduced by f8813a600df30e018a6224048b31244be3d456b9.

I ran benchmarks before the change, after it, on current dev and on this PR. Averages of 10 runs each:

Before change | After change | dev | PR
--------------- | -------------- | --- | ------
1170566.1 | 1081454.708 | 1089127.197 | 1153864.439

So most of the perf loss is recovered with this change.

@mikeharder @halter73 @davidfowl @DamianEdwards

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/FrameHeaders.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/FrameHeaders.cs
@@ -233,16 +233,41 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Http
             }
         }
 
-        public static long ParseContentLength(StringValues value)
+        public static unsafe long ParseContentLength(StringValues value)
         {
-            try
+            var input = value.ToString();
+            var parsed = 0L;
+
+            fixed (char* ptr = input)
             {
-                return long.Parse(value, NumberStyles.AllowLeadingWhite | NumberStyles.AllowTrailingWhite, CultureInfo.InvariantCulture);
+                var ch = (ushort*)ptr;
+                var end = ch + input.Length;
+
+                if (ch == end)
+                {
+                    ThrowInvalidContentLengthException(value);
+                }
+
+                ushort digit = 0;
+                while (ch < end && (digit = (ushort)(*ch - 0x30)) <= 9)
+                {
+                    parsed *= 10;
+                    parsed += digit;
+                    ch++;
+                }
+
+                if (ch != end)
+                {
+                    ThrowInvalidContentLengthException(value);
+                }
             }
-            catch (FormatException ex)
-            {
-                throw new InvalidOperationException("Content-Length value must be an integral number.", ex);
-            }
+
+            return parsed;
+        }
+
+        private static void ThrowInvalidContentLengthException(string value)
+        {
+            throw new InvalidOperationException($"Invalid Content-Length: \"{value}\". Value must be a positive integral number.");
         }
 
         private static void ThrowInvalidHeaderCharacter(char ch)
--- a/test/Microsoft.AspNetCore.Server.KestrelTests/FrameResponseHeadersTests.cs
+++ b/test/Microsoft.AspNetCore.Server.KestrelTests/FrameResponseHeadersTests.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Globalization;
 using System.Text;
 using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.Server.Kestrel;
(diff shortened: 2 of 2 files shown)


src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/FrameHeaders.cs
```diff
@@ -233,16 +233,57 @@ public static void ValidateHeaderCharacters(string headerCharacters)
...
+        public static unsafe long ParseContentLength(StringValues value)
         {
-            try
-            {
-                return long.Parse(value, NumberStyles.AllowLeadingWhite | NumberStyles.AllowTrailingWhite, CultureInfo.InvariantCulture);
-            }
-            catch (FormatException ex)
+            const string errorMessage = "Content-Length value must be an integral number.";
+            var input = value.ToString();
+            var parsed = (long)0;
```
halter73: nit: just use `long parsed = 0`

halter73:
@benaadams might also want to take a look at this.

Can we use ParseContentLength in MessageBody.For? This might have to be rewritten to TryParseContentLength so we can have different behavior in the failure cases.

Have you run any microbencharks comparing this to the previous long.Parse version? I suggest using https://github.com/PerfDotNet/BenchmarkDotNet.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/FrameHeaders.cs
```diff
@@ -233,16 +233,57 @@ public static void ValidateHeaderCharacters(string headerCharacters)
...
+                // Skip leading whitespace
+                while (ch < end && ((*ch >= 0x09 && *ch <= 0x0D) || *ch == 0x20))
+                {
+                    ch++;
+                }
+
+                // If done on end of input or char is not a number, input is invalid
+                if (ch == end || *ch < 0x30 || *ch > 0x39)
+                {
+                    throw new InvalidOperationException(errorMessage);
```
benaadams: Don't throw inline; move out out a separate method e.g. `void ThrowInvalidContentLength()` else it will bring in 3 `InvalidOperationException` ctors and 3 string loads

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/FrameHeaders.cs
moozzyk: 0L

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/FrameHeaders.cs
```diff
@@ -233,16 +233,57 @@ public static void ValidateHeaderCharacters(string headerCharacters)
...
+            var parsed = (long)0;
+
+            fixed (char* ptr = input)
             {
-                throw new InvalidOperationException("Content-Length value must be an integral number.", ex);
+                var ch = ptr;
+                var end = ptr + input.Length;
+
+                // Skip leading whitespace
+                while (ch < end && ((*ch >= 0x09 && *ch <= 0x0D) || *ch == 0x20))
```
benaadams: Could do
```
var char = (ushort*)ptr;
var end = char + input.Length;

while (ch < end && ((*ch - 0x09) <= (0x0D - 0x09)) || *ch == 0x20))
```

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/FrameHeaders.cs
```diff
@@ -233,16 +233,57 @@ public static void ValidateHeaderCharacters(string headerCharacters)
...
+                {
+                    ch++;
+                }
+
+                // If done on end of input or char is not a number, input is invalid
+                if (ch == end || *ch < 0x30 || *ch > 0x39)
```
benaadams: with ushort* above combine with parsing below?

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/FrameHeaders.cs
```diff
@@ -233,16 +233,57 @@ public static void ValidateHeaderCharacters(string headerCharacters)
...
+                {
+                    throw new InvalidOperationException(errorMessage);
+                }
+
+                // Parse number
+                while (ch < end && *ch >= 0x30 && *ch <= 0x39)
```
benaadams: Something like?
```csharp
ushort digit;
if (ch == end || (digit = (*ch - 0x30)) > 9)
{
    ThrowInvalidContentLength();
}

do
{
    parsed *= 10;
    parsed += digit;
} while (ch < end || (digit = (*ch - 0x30)) > 9);
```

benaadams:
Nice 👍

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/FrameHeaders.cs
benaadams: *edit* fixed while clause

halter73:
@benaadams Thanks for the feedback!

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/FrameHeaders.cs
cesarbs: I mean, the `ushort` made a different but not the rewritten `while` loop.

cesarbs:
@benaadams Thanks for the feedback! Got some very significant gains with your suggestions 😄

cesarbs:
@halter73 Didn't use BenchmarkDotNet but ran my own one. I'm parsing 10 million random positive longs. Numbers:

Initial `ParseContentLength` in PR: 200.1526ms
`ParseContentLength` after @benaadams' suggestions: 150.7072ms
long.Parse: 636.982ms

Give or take 10ms on each number on each run.

Code:

```csharp
using System;
using System.Diagnostics;
using System.Globalization;
using Microsoft.Extensions.Primitives;

namespace ParseContentLength
{
    public class Program
    {
        private const int loops = (int)1e7;

        {
            Console.WriteLine($"ParseContentLength: {Benchmark(value => ParseContentLength(value)).TotalMilliseconds}ms");
            Console.WriteLine($"long.Parse: {Benchmark(value => long.Parse(value, NumberStyles.AllowLeadingWhite | NumberStyles.AllowTrailingWhite, CultureInfo.InvariantCulture)).TotalMilliseconds}ms");
        }

        public static TimeSpan Benchmark(Func<string, long> parse)
        {
            var bytes = new byte[8];
            var random = new Random();
            long ticks = 0;

            for (var i = 0; i < loops; i++)
            {
                random.NextBytes(bytes);
                var input = Math.Abs(BitConverter.ToInt64(bytes, 0)).ToString();
                var sw = new Stopwatch();
                sw.Start();
                var result = parse(input);
                ticks += sw.ElapsedTicks;
                if (result.ToString() != input) throw new Exception();
            }

            return TimeSpan.FromTicks(ticks);
        }

        public static unsafe long ParseContentLength(StringValues value)
        {
            var input = value.ToString();
            var parsed = 0L;

            fixed (char* ptr = input)
            {
                var ch = (ushort*)ptr;
                var end = ch + input.Length;

                // Skip leading whitespace
                while (ch < end && ((*ch >= 0x09 && *ch <= 0x0D) || *ch == 0x20))
                {
                    ch++;
                }

                // Parse number
                ushort digit = 0;
                if (ch == end || (digit = (ushort)(*ch - 0x30)) > 9)
                {
                    ThrowInvalidContentLengthException();
                }

                do
                {
                    parsed *= 10;
                    parsed += digit;
                    ch++;
                } while (ch < end && (digit = (ushort)(*ch - 0x30)) <= 9);

                // If done and there's input and char is not whitespace, input is invalid 
                if (ch < end && (*ch < 0x09 || *ch > 0x0D || *ch != 0x20))
                {
                    ThrowInvalidContentLengthException();
                }

                // Skip trailing whitespace
                while (ch < end && ((*ch >= 0x09 && *ch <= 0x0D) || *ch == 0x20))
                {
                    ch++;
                }

                // If not at end of input, input is invalid
                if (ch != end)
                {
                    ThrowInvalidContentLengthException();
                }
            }

            return parsed;
        }

        private static void ThrowInvalidContentLengthException()
        {
            throw new InvalidOperationException("Content-Length value must be an integral number.");
        }
    }
}
```

halter73:
@cesarbs The nice thing about BenchmarkDotNet is that it does multiple test runs and gives you median and standard deviation. Once you've used it once, it's really easy to set up new scenarios and it gives you nice output (even markdown formatted) like shown [here](https://perfdotnet.github.io/BenchmarkDotNet/Overview.htm#benchmark-results).

cesarbs:
Plaintext still looks about the same: 1157141.668 RPS (average of 10 runs).

cesarbs:
@halter73 Alright I'll give it a shot.

halter73:
@cesarbs I was looking through our tests, and I don't see any that attempt to set `Response.Headers["Content-Length"]` to an invalid string. We should add that, especially now that we have our won parsing logic.

cesarbs:
```
Host Process Environment Information:
BenchmarkDotNet.Core=v0.9.9.0
OS=Windows
Processor=?, ProcessorCount=4
Frequency=2435876 ticks, Resolution=410.5299 ns, Timer=TSC
CLR=CORE, Arch=64-bit ? [RyuJIT]
GC=Concurrent Workstation
dotnet cli version: 1.0.0-preview2-003121

Type=Benchmarks  Mode=Throughput

             Method |      Median |     StdDev |
------------------- |------------ |----------- |
          ParseLong | 436.6269 ns | 10.3795 ns |
 ParseContentLength | 200.1591 ns | 30.6031 ns |
```

cesarbs:
@halter73 We do have tests for that in `FrameResponseHeaderTests`. Do you mean I should add functional tests?

halter73:
I remember the "bad" tests now. I was looking in the wrong place. Now that we implemented our own parsing, we should test more invalid content-lengths than just "bad" though. Control characters, whitespace in the middle of otherwise valid numbers, etc...

cesarbs:
Added more tests. Turns out `0x09` through `0x0D` don't work as whitespace because the header gets rejected by `FrameHeaders.ValidateHeaderCharacters()`. So either we:

1) Change `FrameHeaders.ValidateHeaderCharacters()` to allow `0x09` through` 0x0D`; or
2) Make `ParseContentLength()` even simpler by only allowing `0x20` for whitespace. It wouldn't be the same as `NumberStyles.AllowLeadingWhite | NumberStyles.AllowTrailingWhite` in `long.Parse` though.

Tratcher:
Don't allow control characters.

cesarbs:
@Tratcher Even if they're whitespace?

Tratcher:
Yes.

halter73:
Kestrel should already strip whitespace at the start and end of headers. Legal whitespace at least.

halter73:
@cesarbs We should compare the before/after perf of this PR on the big iron. I'll forward you an email on how to do it, and you can post the results.

cesarbs:
We strip whitespace when reading request headers. But what should we do when setting Content-Length from a string, and it contains leading or trailing whitespace? If we want to keep parity with https://github.com/aspnet/HttpAbstractions/blob/87cd79d6fc54bb4abf07c1e380cd7a9498a78612/src/Microsoft.AspNetCore.Http/Internal/ParsingHelpers.cs#L407, we should allow `0x09` through `0x0D`, as well as `0x20` (space). That's what `NumberStyles.AllowLeadingWhite` and `NumberStyles.AllowTrailingWhite` do.

halter73:
@cesarbs good point about request vs response. @Tratcher Do you think we should change ParsingHelpers to not strip whitespace? I don't mind being strict about setting a content-length string, but I would like consistency. But then there's back compat...

Tratcher:
Leave ParsingHelpers for now. Kestrel filters out everything for request headers, and Kestrel can be as strict as it wants to for response headers.

cesarbs:
### Big iron

dev | PR
---- | ---
4721943.71 RPS |  4866482.49

Considerably smaller difference.

test/Microsoft.AspNetCore.Server.KestrelTests/FrameResponseHeadersTests.cs
```diff
@@ -243,5 +257,27 @@ public void ContentLengthValueClearedWhenHeadersCleared()
...
+        {
+            "0",
+            "00",
+            "042",
+            "42",
+        };
+
+        public static TheoryData<string> BadContentLengths => new TheoryData<string>
+        {
+            "",
```
halter73: Can we throw in some invalid characters other than space? E.g. "!" and "ü".

cesarbs: "ü" will be rejected by `ValidateHeaderCharacters()`. But yeah I'll add a few more cases.

ulrichb:
It seems that this perf. improvement has been undone in https://github.com/aspnet/KestrelHttpServer/commit/4da06e8acd799b3e5a92fce00dc686eb2c55e831.

(See [HeaderUtilities.TryParseInt64](https://github.com/aspnet/HttpAbstractions/blob/cc295b1bd81c3cd54ed71eab87761e6acbf7a73e/src/Microsoft.Net.Http.Headers/HeaderUtilities.cs#L206-L209).)

Tratcher:
@ulrichb you're looking at the old version of TryParseInt64.
https://github.com/aspnet/HttpAbstractions/commit/d50a24145d15ac0e9f6d7ed919ae975a9ed9a646#diff-1d7c1559d9f2fc18636596536f94741cR440

ulrichb:
@Tratcher Oh, I checked the 'master'-branch version instead of 'dev'. Thanks.

Issue #1168, opened 2016-10-15
CreateResponseHeader control-flow simplification


cmckinsey:
This may be a trivial issue in the big scheme of things, however given this function is near the top of the profiles I thought I'd mention it. Feel free to close if not worthy.

I was looking at the JIT generated code through this method, when I noticed this section of code can be simplified. 

All possible combinations of variable values will result in shorter path length (i.e. fewer branches) by replacing this:
```

    if (!_keepAlive && !hasConnection)
    {
       responseHeaders.SetRawConnection("close", _bytesConnectionClose) ; 
    } 
    else if (_keepAlive && !hasConnection && _httpVersion == Http.HttpVersion.Http10) 
    { 
       responseHeaders.SetRawConnection("keep-alive", _bytesConnectionKeepAlive) ; 
    } 

```

with this sequence

```

    if (!hasConnection)
    {
        if (!_keepAlive)
        { 
           responseHeaders.SetRawConnection("close", _bytesConnectionClose) ; 
        } 
        else if (_httpVersion == Http.HttpVersion.Http10)
        { 
           responseHeaders.SetRawConnection("keep-alive", _bytesConnectionKeepAlive) ; 
        } 
    }

```

I haven't done any timings with this change so it may not be measurable. But it should never be slower than what's there now. Unfortunately the JIT will not do this automatically anytime soon. I also think it's easier to read so would suggest just changing the sources.

halter73:
If nothing else, it reads more easily this way.

PR #1177, opened 2016-10-24
Control-flow simplification in Frame.CreateResponseHeader() (#1168).

cesarbs:
Thanks @cmckinsey for suggesting this change (#1168).

There's no noticeable change in benchmarks, but the code reads a lot better this way.

cc @halter73


halter73:
:shipit:

Issue #1181, opened 2016-10-25
Handle non "chunked" Transfer-Encoding request header values.


halter73:
Kestrel currently assumes that the request is chunked if the Transfer-Encoding request header is present. This [may not be the case](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Transfer-Encoding).

PR #1185, opened 2016-10-27
Handle non "chunked" Transfer-Encoding header values (#1181)

cesarbs:
#1181

- Reject request if `chunked` is not final token in `Transfer-Encoding`
- Close connection if `chunked` is not final token in response `Transfer-Encoding`

Point for discussion: should we insert `chunked` and auto-chunk responses that don't have `chunked` in `Transfer-Encoding`? Or should we leave it up to the app to chunk the response in those cases? The problem with not auto-chunking is that Kestrel will now close the connection if such a response is sent.

@halter73 @davidfowl @Tratcher @mikeharder

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
@@ -811,13 +811,25 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Http
         {
             var responseHeaders = FrameResponseHeaders;
             var hasConnection = responseHeaders.HasConnection;
-            var connectionOptions = hasConnection ? FrameHeaders.ParseConnection(responseHeaders.HeaderConnection) : ConnectionOptions.None;
+            var connectionOptions = FrameHeaders.ParseConnection(responseHeaders.HeaderConnection);
+            var hasTransferEncoding = responseHeaders.HasTransferEncoding;
+            var transferCoding = FrameHeaders.GetFinalTransferCoding(responseHeaders.HeaderTransferEncoding);
 
             var end = SocketOutput.ProducingStart();
 
             if (_keepAlive && hasConnection)
             {
                 _keepAlive = (connectionOptions & ConnectionOptions.KeepAlive) == ConnectionOptions.KeepAlive;
+            }
+
+            // https://tools.ietf.org/html/rfc7230#section-3.3.1
+            // If any transfer coding other than
+            // chunked is applied to a response payload body, the sender MUST either
+            // apply chunked as the final transfer coding or terminate the message
+            // by closing the connection.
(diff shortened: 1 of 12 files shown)


cesarbs:
No perf impact:

dev: 1210317.84 RPS
PR: 1224226.38

benaadams:
For compression `Content-Encoding` would be used if Kestrel was source server; but maybe? `Transfer-Encoding` would be used if it was a proxy - but not necessarily chunked if length was known? (Trying to think of a scenerio where it would apply)

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
```diff
@@ -812,6 +812,8 @@ private Task WriteSuffix()
             var responseHeaders = FrameResponseHeaders;
             var hasConnection = responseHeaders.HasConnection;
             var connectionOptions = hasConnection ? FrameHeaders.ParseConnection(responseHeaders.HeaderConnection) : ConnectionOptions.None;
+            var hasTransferEncoding = responseHeaders.HasTransferEncoding;
+            var transferCoding = hasTransferEncoding ? FrameHeaders.GetFinalTransferCoding(responseHeaders.HeaderTransferEncoding) : TransferCoding.None;
```
halter73: Nit: HasConnection and HasContentLength both simply call StringValues.Count. It would be simpler to have ParseConnection and GetFinalTransferCoding check the Count themselves and return None rather than doing the Has checks here.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
```diff
@@ -820,14 +822,29 @@ private Task WriteSuffix()
...
+            // chunked is applied to a response payload body, the sender MUST either
+            // apply chunked as the final transfer coding or terminate the message
+            // by closing the connection.
+            if (hasTransferEncoding && transferCoding != TransferCoding.Chunked)
+            {
+                _keepAlive = false;
+
+                if (hasConnection)
+                {
+                    responseHeaders.SetRawConnection("close", _bytesConnectionClose);
```
halter73: I like setting _keepAlive to false, but I'm not sure about changing headers that the user as already explicitly set even to save them from violating the HTTP spec. Do we do that for other headers?

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
cesarbs: Just realized I don't even have to check `Count` - the `foreach` simply won't iterate over anything and those methods will return `None`.

cesarbs:
Updated.

cesarbs:
Ping.

halter73:
:shipit:

Issue #1167, opened 2016-10-15
Functional tests failing due to aborted connections


halter73:
Ex:

https://ci.appveyor.com/project/aspnetci/kestrelhttpserver/build/1.0.3648#L449
https://ci.appveyor.com/project/aspnetci/kestrelhttpserver/build/1.0.3645#L428

This reminds me of https://github.com/aspnet/KestrelHttpServer/issues/1147, but the error is coming from BeginRead instead of EndRead.

halter73:
I'm wondering if re-adding [a line like this](https://github.com/aspnet/KestrelHttpServer/blob/1.0.0-beta8/src/Microsoft.AspNet.Server.Kestrel/Http/Frame.cs#L214) to wait for a client FIN/RST prior to calling uv_close is going to be the fix.

I remember removing this because it caused server shutdown to always take the full 5 seconds. Chrome and IE (but not firefox) would take forever to send a FIN back when receiving a FIN on an idle connection. For this reason, we might want to set a pretty aggressive timeout.

halter73:
ZeroContentLengthAssumedOnNonKeepAliveRequestsWithoutContentLengthOrTransferEncodingHeader also failed on OS X now: https://travis-ci.org/aspnet/KestrelHttpServer/jobs/171495305

halter73:
This doesn't seem to be fully fixed. SingleErrorResponseSentWhenAppSwallowsBadRequestException is still failing.

1. https://ci.appveyor.com/project/aspnetci/kestrelhttpserver/build/1.0.3704
2. https://github.com/aspnet/KestrelHttpServer/pull/993#issuecomment-253887152 (From well before)

This is presumably the issue with sending extra data after the request was rejected like we used to in ZeroContentLengthAssumedOnNonKeepAliveRequestsWithoutContentLengthOrTransferEncodingHeader.

muratg:
@halter73 good to close?

PR #1186, opened 2016-10-29
Fix test failures caused by connection resets (#1167).

cesarbs:
#1167

@halter73

### DIFF
--- a/test/Microsoft.AspNetCore.Server.KestrelTests/EngineTests.cs
+++ b/test/Microsoft.AspNetCore.Server.KestrelTests/EngineTests.cs
@@ -580,6 +580,7 @@ namespace Microsoft.AspNetCore.Server.KestrelTests
         {
             using (var server = new TestServer(async httpContext =>
             {
+                // This will hang if 0 content length is not assumed by the server
                 Assert.Equal(0, await httpContext.Request.Body.ReadAsync(new byte[1], 0, 1).TimeoutAfter(TimeSpan.FromSeconds(10)));
             }, testContext))
             {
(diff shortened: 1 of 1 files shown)


test/Microsoft.AspNetCore.Server.KestrelTests/EngineTests.cs
```diff
@@ -592,7 +593,7 @@ public void ConnectionCanReadAndWrite(TestServiceContext testContext)
                         "GET / HTTP/1.1",
                         "Connection: close",
                         "",
-                        "a");
+                        "");
```
halter73: RecieveForcedEnd

halter73: ReceiveForcedEnd

cesarbs:
Updated.

test/Microsoft.AspNetCore.Server.KestrelTests/EngineTests.cs
```diff
@@ -608,7 +609,7 @@ public void ConnectionCanReadAndWrite(TestServiceContext testContext)
                         "GET / HTTP/1.0",
                         "",
                         "a");
```
halter73: I assume we don't want to send the extra "a" here too.

cesarbs:
Updated & rebased.

benaadams:
This mean we can have Loopback Fast Path back? ;-)

cesarbs:
@benaadams Unrelated, to the best of my knowledge.

halter73:
:shipit:

PR #1021, opened 2016-07-28
Improve the performance of connection filters

halter73:
Fixes #768

Issue #768, opened 2016-04-25
StreamSocketOutput.WriteAsync is Sync on Full framework (SSL) 


benaadams:
From #569; PR #588 resolves coreclr, however full framework still has the issue.

Issue on full framework if made sync (it blocks)
![](http://aoa.blob.core.windows.net/aspnet/sslstream-async-issue.png)

Fullframework uses framework's SSL stream the corefx SSL steam only supports 

net46 (.NETFramework,Version=v4.6)
netstandard1.3 (.NETStandard,Version=v1.3)

So would currently need to version bump from v4.5.1 to use it or get the sslstream retargeted (if that would even work?)

benaadams:
Being looked at https://github.com/dotnet/corefx/pull/7800#issuecomment-214267754

benaadams:
See https://github.com/dotnet/corefx/issues/8358

muratg:
Let's try this in RTM.

PR #1187, opened 2016-10-31
Implement Begin/End Read/Write methods in LoggingStream

halter73:
- This allows the reads and writes from SslStream to be logged on desktop .NET

This is working around the issue in SslStream first described in https://github.com/aspnet/KestrelHttpServer/issues/768 and partially fixed by https://github.com/aspnet/KestrelHttpServer/pull/1021.

This issue is apparent when `.UseConnectionLogging()` is configured before `.UseHttps()` and then run on .NET 4.6.

### DIFF
--- a/test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/LoggingConnectionFilterTests.cs
+++ b/test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/LoggingConnectionFilterTests.cs
@@ -1,0 +1,47 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Threading.Tasks;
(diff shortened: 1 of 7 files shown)


test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/LoggingConnectionFilterTests.cs
```diff
@@ -0,0 +1,47 @@
...
+                });
+            })
+            .Build();
+
+            using (host)
+            {
+                host.Start();
+
+                var getTask = HttpClientSlim.GetStringAsync($"https://localhost:{host.GetPort()}/", validateCertificate: false);
+                var timeoutTask = await Task.WhenAny(getTask, Task.Delay(10000));
```
cesarbs: You could use the `TimeoutAfter()` extension method.

halter73:
@cesarbs updated.

PR #1190, opened 2016-11-02
Fix more test failures caused by connection resets

halter73:
#1167

### DIFF
--- a/test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/ResponseTests.cs
+++ b/test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/ResponseTests.cs
@@ -340,15 +340,12 @@ namespace Microsoft.AspNetCore.Server.Kestrel.FunctionalTests
         [Fact]
         public async Task SingleErrorResponseSentWhenAppSwallowsBadRequestException()
         {
+            BadHttpRequestException readException = null;
+
             using (var server = new TestServer(async httpContext =>
             {
-                try
-                {
-                    await httpContext.Request.Body.ReadAsync(new byte[1], 0, 1);
-                }
-                catch (BadHttpRequestException)
-                {
-                }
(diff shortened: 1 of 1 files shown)


cesarbs:
Failed https://ci.appveyor.com/project/aspnetci/kestrelhttpserver/build/1.0.3705

The fix is to change `Send` to `SendEnd`. What happens is that if we happen to close the connection before all data has been read, the connection will be reset and `TestConnection` will fail to read the response.

If you send a FIN then the RST is not sent when the socket is shutdown.

Here's a gist with some code to play with: https://gist.github.com/cesarbs/2f166d2ff21ee138d1ea8022a85e295b

Watch the network while running it and you'll see what I'm talking about:

* The way it is now, you'll only see FINs sent from the server on shutdown
* If you remove the `Shutdown` in the client thread, you'll see the server FIN followed by an RST
* If you change the `Receive` in the server thread to receive 2 bytes, you won't see RSTs

cesarbs:
I was playing with this and `SendEnd` isn't the right fix either. Switching to `SendAll` seems to do the trick i.e. avoid RSTs from the server so the response can be read. The problem with `Send`/`SendEnd` is that they send one character at a time - if it's still sending when the server sends the error response and closes the connection, `Receive` fails when trying to read the response.

Your change also doesn't work because the server just spins waiting for another char so it can begin parsing the chunked prefix (see `MessageBody.ForChunkedEncoding.ParseChunkedPrefix()`).

cesarbs:
One interesting addition to this test is to add a `bool` that's set to `true` in the `catch` block in the app. That way we can be sure we actually caught the `BadHttpRequestException`. Then `Assert.True()` it.

halter73:
I was able to fix this by simply adding one more "g" character. Our current implementation always attempts to read at least 2 bytes before starting to parse the chunk prefix. This could regress if the implementation changes, but no worse than it is today.

cesarbs:
Ok, I supposed that works too. Are you adding the `bool` I suggested?

Issue #1130, opened 2016-09-28
Reject requests that don't set Content-Length or Transfer-Encoding: chunked


cesarbs:
Requests containing a body (POST, PUT, CONNECT, PATCH, OPTIONS) should be rejected with a `411 Length Required` if they don't include a `Content-Length` or `Transfer-Encoding: chunked` header.

cesarbs:
I was re-reading the spec and it says this:

```
The presence of a message body in a request is signaled by a
Content-Length or Transfer-Encoding header field.  Request message
framing is independent of method semantics, even if the method does
not define any use for a message body.
```

https://tools.ietf.org/html/rfc7230#section-3.3

Which means there's no "hard rule" for what methods can or cannot contain a body. We could be prescriptive here and determine that POST and some other methods must send `Transfer-Encoding` or `Content-Length`, but that seems unnecessary.

We already somewhat guard against messages that don't specify a content length by assuming 0. This means things could silently fail or behave strangely for some apps. Deciding on a set of methods and sending a 411 is more explicit. But then the spec doesn't help a lot in deciding which methods to treat like that.

On the other hand, the spec says this:

```
If a message is received with both a Transfer-Encoding and a
Content-Length header field, the Transfer-Encoding overrides the
Content-Length.  Such a message might indicate an attempt to
perform request smuggling (Section 9.5) or response splitting
(Section 9.4) and ought to be handled as an error.  A sender MUST
remove the received Content-Length field prior to forwarding such
a message downstream.
```

https://tools.ietf.org/html/rfc7230#section-3.3.3

Which to me makes the abandoned https://github.com/aspnet/KestrelHttpServer/issues/1164 more relevant. Granted, the specs seems to be referring to intermediate servers that just foward messages. But why shouldn't Kestrel do the same? We can either strip `Content-Length` in those cases and log a warning, or throw an exception to apps attempting to set both headers.

cesarbs:
I'm thinking of making this change for POST and PUT only, leaving OPTIONS out (I'm not familiar with how clients use OPTIONS).

cesarbs:
This seems to be what IIS does btw. I know lighttpd requires `Content-Length` on POST too (and I assume PUT). I'll check nginx.

cesarbs:
Same for nginx.

muratg:
If we can't get this in soon, let's move it to 1.2.0.

PR #1193, opened 2016-11-03
Reject POST and PUT requests with no Content-Length or Transfer-Encoding (#1130)

cesarbs:
#1130 

@halter73 @davidfowl @Tratcher

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/MessageBody.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/MessageBody.cs
@@ -6,6 +6,7 @@ using System.IO;
 using System.Numerics;
 using System.Threading;
 using System.Threading.Tasks;
+using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.Server.Kestrel.Internal.Infrastructure;
 using Microsoft.Extensions.Internal;
 using Microsoft.Extensions.Primitives;
@@ -266,17 +267,29 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Http
                 return new ForChunkedEncoding(keepAlive, headers, context);
             }
 
-            var unparsedContentLength = headers.HeaderContentLength.ToString();
-            if (unparsedContentLength.Length > 0)
+            var unparsedContentLength = headers.HeaderContentLength;
+            if (unparsedContentLength.Count > 0)
             {
-                long contentLength;
-                if (!long.TryParse(unparsedContentLength, out contentLength) || contentLength < 0)
+                try
+                {
+                    var contentLength = FrameHeaders.ParseContentLength(unparsedContentLength);
+                    return new ForContentLength(keepAlive, contentLength, context);
+                }
+                catch (InvalidOperationException)
                 {
                     context.RejectRequest(RequestRejectionReason.InvalidContentLength, unparsedContentLength);
                 }
-                else
+            }
+
+            // Avoid slowing down most common case
+            if (!object.ReferenceEquals(context.Method, HttpMethods.Get))
+            {
+                // If we got here, request contains no Content-Length or Transfer-Encoding header.
+                // Reject with 411 Length Required.
+                if (HttpMethods.IsPost(context.Method) || HttpMethods.IsPut(context.Method))
                 {
-                    return new ForContentLength(keepAlive, contentLength, context);
+                    var requestRejectionReason = httpVersion == HttpVersion.Http11 ? RequestRejectionReason.LengthRequired : RequestRejectionReason.LengthRequiredHttp10;
+                    context.RejectRequest(requestRejectionReason, context.Method);
                 }
             }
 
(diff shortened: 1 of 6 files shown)


src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/MessageBody.cs
```diff
@@ -280,6 +281,13 @@ protected virtual void OnConsumedBytes(int count)
                 }
             }
 
+            // If we got here, request contains no Content-Length or Transfer-Encoding header.
+            // Reject with 411 Length Required.
+            if (HttpMethods.IsPost(context.Method) || HttpMethods.IsPut(context.Method))
```
halter73: No user code has run at this point. Do a strict reference equals.

test/Microsoft.AspNetCore.Server.KestrelTests/MessageBodyTests.cs
```diff
@@ -196,13 +196,29 @@ public void ForThrowsWhenFinalTransferCodingIsNotChunked()
             using (var input = new TestInput())
             {
                 var ex = Assert.Throws<BadHttpRequestException>(() =>
-                    MessageBody.For(HttpVersion.Http10, new FrameRequestHeaders { HeaderTransferEncoding = "chunked, not-chunked" }, input.FrameContext));
+                    MessageBody.For(HttpVersion.Http11, new FrameRequestHeaders { HeaderTransferEncoding = "chunked, not-chunked" }, input.FrameContext));
```
halter73: Why did this change?

test/Microsoft.AspNetCore.Server.KestrelTests/MessageBodyTests.cs
```diff
@@ -196,13 +196,29 @@ public void ForThrowsWhenFinalTransferCodingIsNotChunked()
...
+        [Theory]
+        [InlineData("POST")]
+        [InlineData("PUT")]
+        public void ForThrowsWhenMethodRequiresLengthButNoContentLengthOrTransferEncodingIsSet(string method)
+        {
+            using (var input = new TestInput())
+            {
+                input.FrameContext.Method = method;
+                var ex = Assert.Throws<BadHttpRequestException>(() =>
+                    MessageBody.For(HttpVersion.Http11, new FrameRequestHeaders(), input.FrameContext));
```
halter73: Should we test this for HTTP/1.0 too?

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/MessageBody.cs
halter73: Were there any other methods you considered requiring a content length for?

test/Microsoft.AspNetCore.Server.KestrelTests/MessageBodyTests.cs
cesarbs:
```
If a request
contains an entity body and Content-Length is not specified, and the
server does not recognize or cannot calculate the length from other
fields, then the server should send a 400 (bad request) response.
```

https://tools.ietf.org/html/rfc1945#section-7.2.2

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/MessageBody.cs
cesarbs: Nope. IIS and nginx seem to require a length only for POST and PUT, so I'm doing the same here.

cesarbs:
@halter73 Updated with early check for != GET.

halter73:
Both travis builds are failing with "Block being garbage collected instead of returned to pool: IncomingStart, /Users/travis/build/aspnet/KestrelHttpServer/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/SocketInput.cs, 66"

https://travis-ci.org/aspnet/KestrelHttpServer/builds/173249429

test/Microsoft.AspNetCore.Server.KestrelTests/MessageBodyTests.cs
```diff
@@ -196,13 +196,45 @@ public void ForThrowsWhenFinalTransferCodingIsNotChunked()
...
+
+                Assert.Equal(411, ex.StatusCode);
+                Assert.Equal($"{method} request contains no Content-Length or Transfer-Encoding header", ex.Message);
+            }
+        }
+
+        [Theory]
+        [InlineData("POST")]
+        [InlineData("PUT")]
+        public void ForThrowsWhenMethodRequiresLengthButNoContentLengthSetHttp10(string method)
```
halter73: My guess is that this test is throwing causing the block to be leaked.

halter73: Never mind this issue repros in the dev branch. I found your issue https://github.com/aspnet/KestrelHttpServer/issues/1195.

cesarbs:
Opened https://github.com/dotnet/corefx/issues/13369 to track test failures caused by hang on 0-byte async write to `SslStream`.

cesarbs:
Rebased with b3aca0413c20226694502deb8649abf4e909ce0f.

Issue #1171, opened 2016-10-17
[LTS] Kestrel considers `Connection: keep-alive, upgrade` as non-upgrade request


muratg:
Original issue: https://github.com/aspnet/KestrelHttpServer/issues/1170.

We'll need to fix this in both 1.1.0 and 1.0.2 versions of Kestrel.

muratg:
Make sure to bump Hosting dependency to 1.0.1 once  https://github.com/aspnet/Hosting/issues/873  is in.

Eilon:
This patch is **approved**, please ensure it is merged into the correct branch and building as part of the patch train.

cesarbs:
Was there a particular reason why we decided to port this to 1.0.2? A request with `Connection: keep-alive, upgrade` to Kestrel 1.0.1 gets a `MessageBody` type of `ForRemainingData`, which is expected.

@anurse Did you repro the original issue on Kestrel 1.0.1? Do you know up to what version Firefox was sending `Connection: keep-alive, upgrade`? Latest version sends `Connection: Upgrade`.

anurse:
I did not repro the original issue on 1.0.1, but neither did I try to repro it :). @BrennanConroy had the original repro. If the behavior is already correct in 1.0.1, I don't see a need to port the fix.

BrennanConroy:
I can check tomorrow

Eilon:
Derp 😄 Let's see what happens here - it's easy to *not* check something in 😄

BrennanConroy:
So 1.0.1 worked because it didn't even check for `"upgrade"` and by default returns a `ForRemainingData`, you can see this change https://github.com/aspnet/KestrelHttpServer/commit/73656f6503688b6eb47ed7a5e93b3db76cbb2280 is when it "broke"

So 1.0.2 can probably stay as is.

cesarbs:
Alright I'm gonna close this then.

anurse:
Cool. Could we port the test over so we ensure that the behavior is never regressed?

muratg:
Agreed, @cesarbs could you port the test? Also the labels on this bug :)

halter73:
The tests added in #1178 were mostly unit tests that make ensure Kestrel's properly interprets the header in it's internal data structures.

We need a new test verifying that it's possible to read all the data sent by the client over the upgraded connection given a `Connection: keep-alive, upgrade`  request. This test could be merged into 1.0.* and dev.

[...]

PR #1194, opened 2016-11-03
Add regression tests for 'Connection: keep-alive, upgrade' request header (#1171)

cesarbs:
#1171 

@halter73 @anurse @BrennanConroy


test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/RequestTests.cs
```diff
@@ -184,6 +185,47 @@ public Task RemoteIPv6Address()
             }
         }
 
+        [Fact]
+        public void CanUpgradeRequestWithConnectionKeepAliveUpgradeHeader()
```
cesarbs: We don't have this one in `dev`, but I'll port it.

halter73:
:shipit:

Issue #1195, opened 2016-11-04
*** Block being garbage collected instead of returned to pool: IncomingStart


cesarbs:
https://travis-ci.org/aspnet/KestrelHttpServer/jobs/173112719

https://travis-ci.org/aspnet/KestrelHttpServer/jobs/173112720

cc @halter73

halter73:
This looks like it might be caused by https://github.com/dotnet/corefx/pull/13083 which fixes https://github.com/dotnet/corefx/issues/10858. The previous bug wasn't so bad because we simply logged and more-or-less ignored the exception when it was thrown in StreamSocketOutput.ProducingComplete(). Now it looks like the write isn't completing at all in the ClientCertificateValidationGetsCalledWithNotNullParameters test.

halter73:
dotnet/corefx#13369 should give use the ultimate fix.

halter73:
@muratg This was a quickly caught regression that never made a release, so I'm not labeling it as a "bug" even though it kinda is.

PR #1196, opened 2016-11-04
Avoid zero-length writes to flush connection filter StreamSocketOutput

halter73:
- This works around a zero-length write bug in SslStream.

#1195


cesarbs:
🚢 once it's all green.

PR #1197, opened 2016-11-04
Prevent block leak when socket is closed before connection filter is applied

cesarbs:
When a connection filter is in use, there's a race in `Connection` when a connection is started and immediately ended with an error (e.g. when a client opens a socket and immediately disposes it, causing an RST to be sent to the server). The error prompts a call to `Connection.Abort()`, which ultimately leads to a call to `Connection.OnSocketClose()`. At the same time, `Connection.Start()` starts a task calling the filter's `OnConnectionAsync()` method, whose continuation will eventually call `Connection.ApplyConnectionFilter()`, which will instantiate a `FilteredStreamAdapter` and start it's read input task by calling `ReadInputAsync()`. This last task leases a block from the memory pool.

If `ApplyConnectionFilter()` has not been called by the time `OnSocketClose()` is called, `OnSocketClose()` will fail to dispose the `FilteredStreamAdapter` and the read input task will not be awaited. The block leased by `FilteredStreamAdapter` will thus never be returned to the pool and will be garbage collected instead.

The fix is to make sure `FilteredStreamAdapter` has been created when `OnSocketClosed()` is called. Since we also have to account for the case when there's no connection filter, a practical way to do this is to wait for a task that is completed once the frame has been started (which happens as soon as the connection is started when there's no filter, and as soon as the filter is applied (i.e. the `FilteredStreamAdapter` is created and it's read input task started) when there is one. `OnSocketClosed` can then wait until that task is completed and continue it with the code currently in place there.

cc @halter73

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Connection.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Connection.cs
@@ -154,15 +154,20 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Http
         // Called on Libuv thread
         public virtual void OnSocketClosed()
         {
-            if (_filteredStreamAdapter != null)
+            _frame.FrameStartedTask.ContinueWith((task, state) =>
             {
-                Task.WhenAll(_readInputTask, _frame.StopAsync()).ContinueWith((task, state) =>
+                var connection = (Connection)state;
+
+                if (_filteredStreamAdapter != null)
                 {
-                    var connection = (Connection)state;
-                    connection._filterContext.Connection.Dispose();
-                    connection._filteredStreamAdapter.Dispose();
-                }, this);
-            }
+                    Task.WhenAll(_readInputTask, _frame.StopAsync()).ContinueWith((task2, state2) =>
+                    {
+                        var connection2 = (Connection)state2;
+                        connection2._filterContext.Connection.Dispose();
+                        connection2._filteredStreamAdapter.Dispose();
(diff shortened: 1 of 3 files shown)


cesarbs:
It's a race, hard to add a consistent test.

halter73:
Can you post sample code I could run in a loop to see it myself?

I totally believe it happens, but I'd like to try it out.

cesarbs:
```csharp
[Fact]
public void BlockLeak()
{
    var loggerFactory = new HandshakeErrorLoggerFactory();

        .UseKestrel(options =>
        {
            options.UseHttps(@"TestResources/testCert.pfx", "testPassword");
        })
        .UseUrls("https://127.0.0.1:0/")
        .UseLoggerFactory(loggerFactory)
        .Configure(app => { });

    using (var host = hostBuilder.Build())
    {
        host.Start();

        while (true)
        {
            {
                socket.Connect(new IPEndPoint(IPAddress.Loopback, host.GetPort()));

                // Close socket immediately
                socket.LingerState = new LingerOption(true, 0);
            }
        }
    }
}
```

cesarbs:
I'm actually going to add that test and put a comment explaining it's a tentative test, and that if it ever causes the test run to blow up, it's a regression on this change.

cesarbs:
The good thing is that it doesn't even need a loop to repro without the fix - the failure is pretty consistent.

cesarbs:
Edited PR's description to explain the bug.

halter73:
👍 for adding the test. :shipit:

Issue #1176, opened 2016-10-21
Don't log ECONNRESET's between requests and during shutdown


halter73:
Connection resets between requests are fairly common and benign, so Kestrel shouldn't log this even at the informational level. Simply closing the browser (in this case IE 10) after a normal browsing session can lead to scary looking logs like the following as the browser closes its idle connections:

```
info: Microsoft.AspNetCore.Server.Kestrel[14]
      Connection id "0HKVPEE3563UP" communication error
Microsoft.AspNetCore.Server.Kestrel.Internal.Networking.UvException: Error -4077 ECONNRESET connection reset by peer
warn: Microsoft.AspNetCore.Server.Kestrel[0]
      Connection processing ended abnormally
System.IO.IOException: Error -4077 ECONNRESET connection reset by peer ---> Microsoft.AspNetCore.Server.Kestrel.Internal
.Networking.UvException: Error -4077 ECONNRESET connection reset by peer
   --- End of inner exception stack trace ---
   at Microsoft.AspNetCore.Server.Kestrel.Internal.Http.SocketInput.CheckConnectionError()
   at Microsoft.AspNetCore.Server.Kestrel.Internal.Http.Frame`1.<RequestProcessingAsync>d__2.MoveNext()
info: Microsoft.AspNetCore.Server.Kestrel[14]
      Connection id "0HKVPEE3563UQ" communication error
Microsoft.AspNetCore.Server.Kestrel.Internal.Networking.UvException: Error -4077 ECONNRESET connection reset by peer
warn: Microsoft.AspNetCore.Server.Kestrel[0]
      Connection processing ended abnormally
System.IO.IOException: Error -4077 ECONNRESET connection reset by peer ---> Microsoft.AspNetCore.Server.Kestrel.Internal
.Networking.UvException: Error -4077 ECONNRESET connection reset by peer
   --- End of inner exception stack trace ---
   at Microsoft.AspNetCore.Server.Kestrel.Internal.Http.SocketInput.CheckConnectionError()
   at Microsoft.AspNetCore.Server.Kestrel.Internal.Http.Frame`1.<RequestProcessingAsync>d__2.MoveNext()
```

In the case where Kestrel already logs "Some connections failed to close gracefully during server shutdown," we could also avoid logging the ECONNRESET communication error for each connection that was forcibly closed. This I'm less sure about since the ECONNRESET logs include the ConnectionIds of the relevant connection which could be useful.

[...]

PR #1199, opened 2016-11-05
Make connection reset logs less scary

halter73:
- Particularly between requests, connection resets aren't that abnormal.

#1176

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/FrameOfT.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/FrameOfT.cs
@@ -2,10 +2,12 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.IO;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Hosting.Server;
-using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Server.Kestrel.Internal.Infrastructure;
+using Microsoft.AspNetCore.Server.Kestrel.Internal.Networking;
 using Microsoft.Extensions.Logging;
 
 namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Http
@@ -29,21 +31,30 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Http
         /// </summary>
         public override async Task RequestProcessingAsync()
         {
+            var requestLineStatus = RequestLineStatus.Empty;
+
             try
             {
                 while (!_requestProcessingStopping)
                 {
                     ConnectionControl.SetTimeout(_keepAliveMilliseconds, TimeoutAction.CloseConnection);
 
-                    while (!_requestProcessingStopping && TakeStartLine(SocketInput) != RequestLineStatus.Done)
+                    while (!_requestProcessingStopping)
                     {
+                        requestLineStatus = TakeStartLine(SocketInput);
+
+                        if (requestLineStatus == RequestLineStatus.Done)
+                        {
+                            break;
+                        }
+
                         if (SocketInput.CheckFinOrThrow())
                         {
                             // We need to attempt to consume start lines and headers even after
                             // SocketInput.RemoteIntakeFin is set to true to ensure we don't close a
                             // connection without giving the application a chance to respond to a request
                             // sent immediately before the a FIN from the client.
-                            var requestLineStatus = TakeStartLine(SocketInput);
+                            requestLineStatus = TakeStartLine(SocketInput);
 
                             if (requestLineStatus == RequestLineStatus.Empty)
                             {
@@ -187,6 +198,15 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Http
                 // Handle BadHttpRequestException thrown during request line or header parsing.
                 // SetBadRequestState logs the error.
                 SetBadRequestState(ex);
+            }
+            catch (IOException ex) when (ex.InnerException is UvException)
+            {
+                // Don't log ECONNRESET errors made between requests. Browsers like IE will reset connections regularly.
+                if (requestLineStatus != RequestLineStatus.Empty ||
+                    ((UvException)ex.InnerException).StatusCode != Constants.ECONNRESET)
+                {
+                    Log.RequestProcessingError(ConnectionId, ex);
+                }
             }
             catch (Exception ex)
             {
--- a/src/Microsoft.AspNetCore.Server.Kestrel/KestrelServer.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/KestrelServer.cs
@@ -100,6 +100,11 @@ namespace Microsoft.AspNetCore.Server.Kestrel
                         "ThreadCount must be positive.");
                 }
 
+                if (!Constants.ECONNRESET.HasValue)
+                {
+                    _logger.LogWarning("Unable to determine ECONNRESET value on this platform.");
+                }
+
                 if (!Constants.EADDRINUSE.HasValue)
                 {
                     _logger.LogWarning("Unable to determine EADDRINUSE value on this platform.");
--- a/test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/RequestTests.cs
+++ b/test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/RequestTests.cs
@@ -17,6 +17,7 @@ using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.Http.Features;
 using Microsoft.AspNetCore.Server.Kestrel.Internal.Networking;
 using Microsoft.AspNetCore.Testing.xunit;
+using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Logging.Testing;
 using Newtonsoft.Json;
 using Newtonsoft.Json.Linq;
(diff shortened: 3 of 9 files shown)


halter73:
@cesarbs

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/FrameOfT.cs
```diff
@@ -188,6 +209,10 @@ public class Frame<TContext> : Frame
                 // SetBadRequestState logs the error.
                 SetBadRequestState(ex);
             }
+            catch (IOException ex) when (ex.InnerException is UvException)
```
cesarbs: Why don't you do all checks here and avoid the new `try`/`catch` block around the code that reads the request line?

src/Microsoft.AspNetCore.Server.Kestrel/KestrelServer.cs
```diff
@@ -100,6 +100,10 @@ public void Start<TContext>(IHttpApplication<TContext> application)
                         "ThreadCount must be positive.");
                 }
 
+                if (!Constants.ECONNRESET.HasValue)
+                {
+                    _logger.LogWarning("Unable to determine ECONNRESET value on this platform.");
+                }
```
cesarbs: nit: add new line after this block

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/RequestTests.cs
```diff
@@ -215,16 +216,119 @@ public Task RemoteIPv6Address()
...
+
+                    // Force a reset
+                    socket.LingerState = new LingerOption(true, 0);
+                }
+
+                // Wait until connection error is logged
+                Assert.True(await connectionResetLogged.WaitAsync(_semaphoreWaitTimeout));
+
+                // Check the no log level higher than Debug was used for a reset before a request
+                Assert.True(testSink.MaxLogLevel <= LogLevel.Debug);
```
cesarbs: Assert equality here.

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/RequestTests.cs
```diff
@@ -191,12 +192,12 @@ public Task RemoteIPv6Address()
         }
 
         [Fact]
-        public async Task ConnectionResetAbortsRequest()
+        public async Task ConnectionResetPriorToRequestIsLoggedQuietly()
```
cesarbs: `...LoggedAsDebug` is more precise

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/RequestTests.cs
```diff
@@ -215,16 +216,119 @@ public Task RemoteIPv6Address()
...
+
+                // Check the no log level higher than Debug was used for a reset between a requests
+                Assert.True(testSink.MaxLogLevel <= LogLevel.Debug);
+
+                // Check for expected message
+                Assert.NotNull(testSink.ConnectionResetMessage);
+                Assert.Contains("reset", testSink.ConnectionResetMessage);
+            }
+        }
+
```
cesarbs: nit: remove extra blank line

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/RequestTests.cs
```diff
@@ -215,16 +216,119 @@ public Task RemoteIPv6Address()
...
+                    socket.LingerState = new LingerOption(true, 0);
+                }
+
+                // Wait until connection error and request processingError is logged
+                var waitAsyncResults = await Task.WhenAll(connectionResetLogged.WaitAsync(_semaphoreWaitTimeout), requestProcessingErrorLogged.WaitAsync(_semaphoreWaitTimeout));
+
+                Assert.All(waitAsyncResults, Assert.True);
+
+                // Check the no log level lower than Information was used for a reset mid-request
+                Assert.True(testSink.MaxLogLevel >= LogLevel.Information);
```
cesarbs: Assert equality.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/FrameOfT.cs
```diff
@@ -37,31 +40,49 @@ public class Frame<TContext> : Frame
...
+                                }
+
+                                break;
+                            }
+                            await SocketInput;
+                        }
+                        catch (IOException ex)
+                        {
+                            // Silence ECONNRESET errors made between requests
+                            if (_requestLineStatus == RequestLineStatus.Empty &&
```
cesarbs: Place `requestLineStatus` in an outer scope - better than making it a `Frame` member.

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/RequestTests.cs
```diff
@@ -215,16 +216,119 @@ public Task RemoteIPv6Address()
...
+
+                // Check for expected message
+                Assert.NotNull(testSink.ConnectionResetMessage);
+                Assert.Contains("reset", testSink.ConnectionResetMessage);
+            }
+        }
+
+
+        [Fact]
+        public async Task ConnectionResetMidRequestIsLoggedLessQuietly()
```
cesarbs: `...LoggedAsInformation`

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/FrameOfT.cs
halter73: It means that we need to move TakeStartLine out of the while condition, or do assignment in the condition. Both seemd icky to me.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/FrameOfT.cs
halter73: This isn't for that. The try/catch above handles that and completely swallows it if it's incomplete. This replaces the "Connection processing ended abnormally" with a better one which has an eventId, connectionId, etc... Since we know a bad connection can cause SocketInput to throw and cause request processing to end abnormally, it's better to handle that explicitly. This catch will handle any connection error, not just resets.

`Log.LogWarning(0,...` is generally reserved for things we don't expect to happen under normalish circumstances.

src/Microsoft.AspNetCore.Server.Kestrel/KestrelServer.cs
halter73: I copied the formatting directly from 1.0.0. You wrote that, right? 😝 I think having them together is fine since they're so closely related.

halter73:
Addressed pr feedback.

src/Microsoft.AspNetCore.Server.Kestrel/KestrelServer.cs
cesarbs: lol

when I see two `if` blocks like that I sometimes mentally insert a non-existing `else` before the second `if` when reading the code.

But this is not really critical so you can leave it as is.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/FrameOfT.cs
cesarbs: Doesn't C# allow `while (... && (requestLineStatus = TakeStartLine(...)) != ...)` ?

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/FrameOfT.cs
cesarbs: Yeah, what I meant was, let the code above throw and catch it here, then check those conditions up there (`_requestLineStatus` and `StatusCodes`) here and make the appropriate trace call.

I might be missing something, but I don't see the need for the `try`/`catch` above and going through the whole thing with `_requestProcessingStopping` being false. You're ending the loop anyways, just let it throw.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/FrameOfT.cs
halter73: It does, but I don't like it.

cesarbs: You dislike it more than a `break` in a `catch` block? :trollface:

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/FrameOfT.cs
halter73: 👍 👍 👍 👍

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/FrameOfT.cs
```diff
@@ -188,6 +199,15 @@ public class Frame<TContext> : Frame
                 // SetBadRequestState logs the error.
                 SetBadRequestState(ex);
             }
+            catch (IOException ex) when (ex.InnerException is UvException)
+            {
+                // Don't log ECONNRESET errors made between requests. Browsers like IE will reset connections regularly.
+                if (requestLineStatus != RequestLineStatus.Empty ||
+                    ((UvException)ex.InnerException).StatusCode != Constants.ECONNRESET.Value)
```
cesarbs: Need to check `HasValue` first

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/FrameOfT.cs
```diff
@@ -188,6 +199,15 @@ public class Frame<TContext> : Frame
...
+                {
+                    Log.RequestProcessingError(ConnectionId, ex);
+                }
```
cesarbs: `else Log.ConnectionReset`?

halter73: We already log that in `Connection.OnRead`. Part of this was getting rid of that double logging.

Issue #1202, opened 2016-11-07
Attempting to write to HEAD response hangs the request


cesarbs:
Run the sample app without any connection filters and send a HEAD request and leave the connection open (this part is important). The client will hang waiting for a response.

This issue was introduced in https://github.com/aspnet/KestrelHttpServer/commit/e7e6b896ba5e5c201b61f70fa4ea352027bd6bd8 when we started ignoring writes to HEAD responses. The problem is that writing a response will make `Frame.HasResponseStarted` true, skipping the call to `ProduceEndAwaited` from `ProduceEnd` and therefore skipping the flush call in that method.

I caught this while working on https://github.com/aspnet/KestrelHttpServer/issues/1139. When I changed all tests to call `Send` instead of `SendEnd`, `ResponseTests.HeadResponseCanContainContentLengthHeaderButBodyNotWritten` started timing out on `TestConnection.Receive`.

muratg:
Approved by @Eilon, so triaging this one in.

PR #1203, opened 2016-11-07
Always flush responses (#1202)

cesarbs:
#1202

cc @halter73

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
@@ -770,20 +770,20 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Http
         {
             ProduceStart(appCompleted: true);
 
-            // Force flush
-            await SocketOutput.FlushAsync();
-
             await WriteSuffix();
         }
 
-        private Task WriteSuffix()
+        private async Task WriteSuffix()
         {
             // _autoChunk should be checked after we are sure ProduceStart() has been called
             // since ProduceStart() may set _autoChunk to true.
             if (_autoChunk)
             {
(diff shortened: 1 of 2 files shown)


test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/ResponseTests.cs
```diff
@@ -782,7 +782,7 @@ public Task ResponseStatusCodeSetBeforeHttpContextDisposedRequestMalformedReadIg
         }
 
         [Fact]
-        public async Task HeadResponseCanContainContentLengthHeaderButBodyNotWritten()
+        public async Task HeadResponseBodyNotWritten()
```
halter73: This test passes in `dev`. Why wouldn't this work?

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
```diff
@@ -770,21 +770,21 @@ protected Task ProduceEnd()
         {
             ProduceStart(appCompleted: true);
 
-            // Force flush
-            await SocketOutput.FlushAsync();
```
halter73: You can remove `ProduceEndAwaited` now. Change 

             if (!HasResponseStarted)
             {
                 return ProduceEndAwaited();
             }

to

             if (!HasResponseStarted)
             {
                 ProduceStart(appCompleted: true);
             }

in ProduceEnd().

halter73:
I actually think I prefer https://github.com/aspnet/KestrelHttpServer/commit/27049598854b8ed14ea3c68ded56404b3ae7f966. It makes HEAD responses flush the header at the same point in the app it would for GET responses. It also avoids creating new async methods.

cesarbs:
@halter73 Ok, I'll close this in favor of your change.

PR #1205, opened 2016-11-09
[WIP]  State machine start line and header parsing

pakrym:


pakrym:
This PR eliminates SIMD seeks and tries to parse everything in one pass through memory. 
During first tests showed ~2-3% improvement on big iron.

cesarbs:
We discussed this some time ago and concluded the added complexity wasn't worth the effort. What changed?

halter73:
If it's faster, functionally equivalent and it doesn't make the code harder to follow, I certainly won't complain 😄

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Infrastructure/MemoryPoolIterator.cs
```diff
@@ -77,6 +79,12 @@ public int Take()
                 return block.Array[index];
             }
 
+            return TakeSlow(wasLastBlock, block);
+        }
+
+        private int TakeSlow(bool wasLastBlock, MemoryPoolBlock block)
```
benaadams: Move `wasLastBlock` into method, isn't used in fast path

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
```diff
@@ -928,129 +942,271 @@ public RequestLineStatus TakeStartLine(SocketInput input)
...
-                        RejectRequest(RequestRejectionReason.InvalidRequestLine,
-                            Log.IsEnabled(LogLevel.Information) ? start.GetAsciiStringEscaped(end, MaxInvalidRequestLineChars) : string.Empty);
-                    }
+                MemoryPoolIterator verbEnd = start;
+                MemoryPoolIterator pathStart = start;
+                MemoryPoolIterator pathEnd = start;
+                MemoryPoolIterator queryStart = start;
+                MemoryPoolIterator versionStart = start;
+                MemoryPoolIterator queryEnd = start;
+                MemoryPoolIterator versionEnd = start;
```
cesarbs: `var`

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
```diff
@@ -928,129 +942,271 @@ public RequestLineStatus TakeStartLine(SocketInput input)
...
+                                }
+                                else if (ch != -1)
+                                {
+                                    queryEnd = scan;
+                                    queryRead = true;
+                                }
+                            } while (!scan.IsEnd && limit > 0);
+                            break;
+                        case StartLineParsingState.VersionFast:
+                        {
```
cesarbs: Why does only this `case` have a block? Why is the break outside the block?

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
```diff
@@ -928,129 +942,271 @@ public RequestLineStatus TakeStartLine(SocketInput input)
...
+                            int bytes;
 
-                    if (httpVersion == string.Empty)
+                            line.Seek(ref _vectorLFs, out bytes, ServerOptions.Limits.MaxRequestLineSize);
+                            line.Skip(1);
+                            RejectRequest(RequestRejectionReason.InvalidRequestLine,
+                                start.GetAsciiStringEscaped(line, MaxInvalidRequestLineChars));
+                            break;
+                        default:
+                            throw new ArgumentOutOfRangeException();
```
cesarbs: Is this the recommended exception type for this situation?

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
```diff
@@ -1174,159 +1330,214 @@ private bool RequestUrlStartsWithPathBase(string requestUrl, out bool caseMatche
...
+                                        consumed = scan;
+                                        return true;
+                                    }
+
+                                    // Headers don't end in CRLF line.
+                                    RejectRequest(RequestRejectionReason.HeadersCorruptedInvalidHeaderSequence);
+                                }
+                                break;
+                            case HeaderParsingState.Name:
+                                bool nameRead = false;
```
cesarbs: `var`

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
```diff
@@ -1174,159 +1330,214 @@ private bool RequestUrlStartsWithPathBase(string requestUrl, out bool caseMatche
...
+                                    limit--;
+                                    if (ch == '\r')
+                                    {
+                                        state = HeaderParsingState.Newline;
+                                        break;
+                                    }
+                                    else if (ch != -1)
+                                    {
+                                        //valueRead = true;
+                                        if (ch != ' ' && ch != '\t')
```
cesarbs: `&&` these in the `else if` above

mikeharder:
Can we write a micro benchmark on sample input using Benchmark.NET or similar?

Issue #1182, opened 2016-10-26
UvException (Error -4047 EPIPE broken pipe) timing out HTTP requests


davenewza:
The following exception is being logged by Kestrel irregularly when performing HTTP requests to our application hosted in an Azure Web App. The request ultimately times out with a 502 at the client with the response:

> The specified CGI application encountered an error and the server terminated the process.

**It appears that the server goes through periods of working perfectly fine, but then periods where this issue will repeatedly occur.** It is not consistent.
## Exception

**Logger name:** Microsoft.AspNetCore.Server.Kestrel
**Log level:** Error
**State:** TcpListenerPrimary.OnConnection
**Exception:**

```
Microsoft.AspNetCore.Server.Kestrel.Internal.Networking.UvException: Error -4047 EPIPE broken pipe
    at Microsoft.AspNetCore.Server.Kestrel.Internal.Networking.Libuv.Check(Int32 statusCode)
    at Microsoft.AspNetCore.Server.Kestrel.Internal.Networking.UvWriteReq.Write2(UvStreamHandle handle, ArraySegment`1 bufs, UvStreamHandle sendHandle, Action`4 callback, Object state)
    at Microsoft.AspNetCore.Server.Kestrel.Internal.Http.ListenerPrimary.DispatchConnection(UvStreamHandle socket)
    at Microsoft.AspNetCore.Server.Kestrel.Internal.Http.TcpListenerPrimary.OnConnection(UvStreamHandle listenSocket, Int32 status)
```
## project.json

```
{
  "webroot": "wwwroot",
  "dependencies": {
    "Microsoft.AspNetCore.Mvc": "1.0.1",
    "Microsoft.AspNetCore.Server.Kestrel": "1.0.1",
    "Microsoft.AspNetCore.Diagnostics": "1.0.0",
    "Microsoft.AspNetCore.StaticFiles": "1.0.0",
    "Microsoft.AspNetCore.Mvc.Cors": "1.0.1",
    "Microsoft.AspNetCore.Mvc.Core": "1.0.1",
    "Microsoft.AspNetCore.Mvc.Abstractions": "1.0.1",
    "Microsoft.AspNetCore.Http": "1.0.0",
    "Microsoft.AspNetCore.Http.Abstractions": "1.0.0",
    "Microsoft.Extensions.Configuration": "1.0.0",
    "Microsoft.Extensions.Configuration.Json": "1.0.0",
[...]

PR #1206, opened 2016-11-09
Ensure clients connecting to Kestrel's dispatch pipe are listeners

halter73:
#1182

@cesarbs @pakrym @davidfowl

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/ListenerPrimary.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/ListenerPrimary.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.IO;
 using System.Runtime.InteropServices;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Server.Kestrel.Internal.Infrastructure;
@@ -75,15 +76,19 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Http
             {
                 dispatchPipe.Init(Thread.Loop, Thread.QueueCloseHandle, true);
                 pipe.Accept(dispatchPipe);
+
+                // Ensure client sends "Kestrel" before adding pipe to _dispatchPipes.
+                var readContext = new PipeReadContext(this);
+                dispatchPipe.ReadStart(
+                    (handle, status2, state) => ((PipeReadContext)state).AllocCallback(handle, status2),
+                    (handle, status2, state) => ((PipeReadContext)state).ReadCallback(handle, status2),
+                    readContext);
             }
             catch (UvException ex)
             {
                 dispatchPipe.Dispose();
                 Log.LogError(0, ex, "ListenerPrimary.OnListenPipe");
-                return;
             }
-
-            _dispatchPipes.Add(dispatchPipe);
         }
 
         protected override void DispatchConnection(UvStreamHandle socket)
@@ -177,6 +182,56 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Http
                         dispatchPipe.Dispose();
                     }
                 }, this).ConfigureAwait(false);
+            }
+        }
+
+        private class PipeReadContext
+        {
+            private readonly ListenerPrimary _listener;
+            private readonly IntPtr _bufPtr;
+            private GCHandle _bufHandle;
+            private int _bytesRead;
+
+            public PipeReadContext(ListenerPrimary listener)
+            {
+                _listener = listener;
+                _bufHandle = GCHandle.Alloc(new byte[8], GCHandleType.Pinned);
+                _bufPtr = _bufHandle.AddrOfPinnedObject();
+            }
+
+            public Libuv.uv_buf_t AllocCallback(UvStreamHandle dispatchPipe, int suggestedSize)
+            {
+                return dispatchPipe.Libuv.buf_init(_bufPtr + _bytesRead, 8 - _bytesRead);
+            }
+
+            public unsafe void ReadCallback(UvStreamHandle dispatchPipe, int status)
+            {
+                try
+                {
+                    dispatchPipe.Libuv.ThrowIfErrored(status);
+
+                    _bytesRead += status;
+
+                    if (_bytesRead == 8)
+                    {
+                        if (*(ulong*)_bufPtr == Constants.PipeMessage)
+                        {
+                            _listener._dispatchPipes.Add((UvPipeHandle) dispatchPipe);
+                            dispatchPipe.ReadStop();
+                            _bufHandle.Free();
+                        }
+                        else
+                        {
+                            throw new IOException("Bad data sent over Kestrel pipe.");
+                        }
+                    }
+                }
+                catch (Exception ex)
+                {
+                    dispatchPipe.Dispose();
+                    _bufHandle.Free();
+                    _listener.Log.LogError(0, ex, "ListenerPrimary.ReadCallback");
+                }
             }
         }
     }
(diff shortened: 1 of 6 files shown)


halter73:
I plan to make a version of this PR for rel/1.0.2, but there are a bunch of restore errors like ` Unable to resolve 'Microsoft.AspNetCore.Hosting (>= 1.0.1)' for '.NETCoreApp,Version=v1.0'.`. It seems like these packages are not in the aspnetcore-rel-1-0-3 myget used by that branch.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Networking/UvWriteReq.cs
```diff
@@ -85,18 +85,30 @@ public void Init(UvLoopHandle loop)
                 _callback = null;
                 _state = null;
                 Unpin(this);
-
-                var block = start.Block;
```
davidfowl: Why was this code here to begin with

halter73: We used to "unpin" the blocks (though that never really worked properly). When we removed block.Unpin(), I guess this vestigial code was missed.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/ListenerSecondary.cs
```diff
@@ -89,17 +92,37 @@ private void ConnectedCallback(UvConnectRequest connect, int status, Exception e
                 return;
             }
 
+            var writeReq = new UvWriteReq(Log);
```
davidfowl: Feels like we should use the WriteReq pool here.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Networking/UvWriteReq.cs
```diff
@@ -85,18 +85,30 @@ public void Init(UvLoopHandle loop)
...
         }
 
-        public unsafe void Write2(
+        public void Write(
+            UvStreamHandle handle,
+            ArraySegment<ArraySegment<byte>> bufs,
+            Action<UvWriteReq, int, Exception, object> callback,
+            object state)
+        {
+            WriteArraySegmentInternal(handle, bufs, null, callback, state);
```
davidfowl: nit: sendHandle:null

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/ListenerPrimary.cs
```diff
@@ -179,5 +184,53 @@ private struct FILE_COMPLETION_INFORMATION
...
+        private class PipeReadContext
+        {
+            private readonly ListenerPrimary _listener;
+            private readonly IntPtr _bufPtr;
+            private int _bytesRead;
+
+            public PipeReadContext(ListenerPrimary listener)
+            {
+                _listener = listener;
+                _bufPtr = Marshal.AllocHGlobal(8);
```
davidfowl: I still think you should just allocate a managed buffer and pin it.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/ListenerSecondary.cs
halter73: Why? This happens a fixed number of times at the start of the process.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/ListenerPrimary.cs
halter73: Why? If we never unpin it, isn't that even worse? It would also be inconsistent within the same class. I don't think we should change `_fileCompletionInfoPtr` right now. And I think if we do change to pinning, we should change both together.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Networking/UvWriteReq.cs
davidfowl: boo hoo

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/ListenerSecondary.cs
davidfowl: That's right, connection churn doesn't affect this...

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/ListenerPrimary.cs
davidfowl: Allocating memory for a native struct VS allocating a buffer for reading feel like very different things. Sure they are different allocation patterns in the same class but for this it's not a big deal. Anyways, no real strong argument but using GC tracked memory is much nicer in general.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/ListenerPrimary.cs
```diff
@@ -179,5 +184,53 @@ private struct FILE_COMPLETION_INFORMATION
...
+            {
+                try
+                {
+                    dispatchPipe.Libuv.ThrowIfErrored(status);
+
+                    _bytesRead += status;
+
+                    if (_bytesRead == 8)
+                    {
+                        if (Marshal.PtrToStructure<ulong>(_bufPtr) == Constants.PipeMessage)
```
davidfowl: Any specific reason you used PtrToStructure? Instead of a cast? Am I missing something?

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/ListenerPrimary.cs
```diff
@@ -179,5 +184,53 @@ private struct FILE_COMPLETION_INFORMATION
...
+                return dispatchPipe.Libuv.buf_init(_bufPtr + _bytesRead, 8 - _bytesRead);
+            }
+
+            public void ReadCallback(UvStreamHandle dispatchPipe, int status)
+            {
+                try
+                {
+                    dispatchPipe.Libuv.ThrowIfErrored(status);
+
+                    _bytesRead += status;
```
halter73: YAGNI. Never experienced a prolonged pipe connection that wasn't our own. We're fixing a pretty specific issue.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/ListenerPrimary.cs
halter73: Now the code totally isn't `unsafe` 😝

halter73:
@davidfowl updated.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/ListenerPrimary.cs
```diff
@@ -179,5 +184,55 @@ private struct FILE_COMPLETION_INFORMATION
...
+            public PipeReadContext(ListenerPrimary listener)
+            {
+                _listener = listener;
+                _bufHandle = GCHandle.Alloc(new byte[8], GCHandleType.Pinned);
+                _bufPtr = _bufHandle.AddrOfPinnedObject();
+            }
+
+            public Libuv.uv_buf_t AllocCallback(UvStreamHandle dispatchPipe, int suggestedSize)
+            {
+                return dispatchPipe.Libuv.buf_init(_bufPtr + _bytesRead, 8 - _bytesRead);
```
davidfowl: Can we guard against (8 - _bytesRead) going negative?

halter73: >  I guess the read callback will fail first.

Exactly. We will either call uv_close or uv_read_stop once 8 bytes are read preventing the alloc callback from being called again. Even if we didn't, we would simply return a zero length buffer which would cause the subsequent read callback to fail with the [UV_ENOBUFS status](http://docs.libuv.org/en/v1.x/handle.html#c.uv_alloc_cb).

pakrym:
:shipit:

halter73:
This was pushed to rel/1.1.0 first and [merged into dev](https://github.com/aspnet/KestrelHttpServer/commit/4a1db745687098cc4e8273f03b2c8a2754587f1a).

PR #1225, opened 2016-11-22
[WIP] Validate Host header

### LINKED ISSUE #839: Reject requests that don't have a Host header
RFC 7230 section 5.4 - https://tools.ietf.org/html/rfc7230#section-5.4

> A server MUST respond with a 400 (Bad Request) status code to any HTTP/1.1 request message that lacks a Host header field and to any request message that contains more than one Host header field or a Host header field with an invalid field-value.

khellang:
Closes #839

This only validates the header for HTTP/1.1 requests, per the RFC. Does it make sense to reject HTTP/1.0 requests as well?

> a Host header field with an invalid field-value.

I'm assuming this would be taken care of by general header field validation?

Still have to take care of the functional tests. Quite tedious work :cry:

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Kestrel/BadHttpRequestException.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/BadHttpRequestException.cs
@@ -78,6 +78,12 @@ namespace Microsoft.AspNetCore.Server.Kestrel
                 case RequestRejectionReason.RequestTimeout:
                     ex = new BadHttpRequestException("Request timed out.", 408);
                     break;
+                case RequestRejectionReason.MissingHostHeader:
+                    ex = new BadHttpRequestException("Missing Host header.", 400);
+                    break;
+                case RequestRejectionReason.MultipleHostHeaders:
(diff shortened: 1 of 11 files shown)


Tratcher:
This is certainly easier than validating the value, but I have to question how much value it adds.

khellang:
> This is certainly easier than validating the value, but I have to question how much value it adds.

What do you mean? You don't think validating the header count adds any value?

Tratcher:
It adds some value, I'm just not sure how much. The presence of the header is a very low bar.

halter73:
We were considering validating this in configurable middleware. All aspects of the validation (ensuring there is exactly 1 Host header, ensuring that the specified host is the correct one, and returning a 400 if not) should be possible in middleware.

khellang:
Why would you put this in configurable middleware? What is there to configure? The RFC seems pretty clear on what's accepted and what's not.

I guess I'missed failing to see when you'd put something in MW and when you'd let the server handle it. It's to enforce the protocol, right? Is that optional?

halter73:
The thing to configure is what hostnames are allowed.

davidfowl:
Closing because pipelines change makes this unmergable

PR #1229, opened 2016-11-22
Refactor FrameTests and rename SocketInput SocketOutput properties

pakrym:
Renaming so we would have less merge conflicts during pipeline migration.

@halter73 @cesarbs @natemcmaster


davidfowl:
👏

PR #1236, opened 2016-11-26
[Experiment] Coalesce small writes

benaadams:
/cc @halter73 @pakrym @cesarbs Could you benchmark this on your environment? It might be interesting... (Am seeing a +50% improvement)

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Kestrel/Filter/Internal/LibuvStream.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/Filter/Internal/LibuvStream.cs
@@ -117,13 +117,12 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Filter.Internal
 
         public override void Flush()
         {
-            // No-op since writes are immediate.
+            _output.Flush();
         }
 
         public override Task FlushAsync(CancellationToken cancellationToken)
         {
-            // No-op since writes are immediate.
-            return TaskCache.CompletedTask;
+            return _output.FlushAsync(cancellationToken);
         }
 
         private ValueTask<int> ReadAsync(ArraySegment<byte> buffer)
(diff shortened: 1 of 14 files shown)


benaadams:
Test are unhappy, but

Pre
```
$ wrk -s pipeline.lua http://localhost:5001 -- 16
Pipeline depth: 16
Pipeline depth: 16
Running 10s test @ http://localhost:5001
  2 threads and 10 connections
    Latency   686.96us  539.68us  12.45ms   80.85%
    Req/Sec    62.71k     6.11k   70.79k    93.50%
  1248438 requests in 10.02s, 157.16MB read
Requests/sec: 124550.58
Transfer/sec:     15.68MB
```
Post
```
$ wrk -s pipeline.lua http://localhost:5001 -- 16
Pipeline depth: 16
Pipeline depth: 16
Running 10s test @ http://localhost:5001
  2 threads and 10 connections
    Latency   390.62us  684.60us  22.03ms   77.36%
    Req/Sec   107.14k     9.52k  113.60k    95.00%
  2132208 requests in 10.01s, 268.41MB read
Requests/sec: 213093.82
Transfer/sec:     26.83MB
```

pakrym:
How would we flush these bytes if app doesn't write anything else?

benaadams:
> How would we flush these bytes if app doesn't write anything else?

Logic should be improved in `FrameOf<T>`; are still some holes; but mainly want to know if they are worth solving by testing its effects on a real network.

halter73:
Did you do any measurements without pipelining?

cesarbs:
Not seeing much of a difference in our benchmarks:

dev                    | pr
----------------- | -------------
1239875.88 RPS | 1219477.48

Averages of 5 runs with pipeline depth of 16.

@benaadams Can you test with separate machines? localhost is not the best way to measure this.

benaadams:
@cesarbs interesting; this is the localhost rawcap I get



Will dig a bit more...

davidfowl:
Closing because pipelines change makes this unmergable

Issue #478, opened 2015-12-13
Add MaxUploadBytes parameter


benaadams:
Fail fast on `Content-Length` header value, count for body in chunked when limit passed.

Failure state: current `413 Payload Too Large` old `413 Request Entity Too Large`

_"The request is larger than the server is willing or able to process."_ [RFC7231](https://tools.ietf.org/html/rfc7231#section-6.5.11)

**.NET 4 - 4.6** [default](https://msdn.microsoft.com/en-us/library/system.web.configuration.httpruntimesection.maxrequestlength%28v=vs.100%29.aspx) `HttpRuntimeSection.MaxRequestLength` 4MB
**IIS** [default](https://www.iis.net/configreference/system.webserver/asp/limits) `maxRequestEntityAllowed` 200kB
or filtered [default](https://www.iis.net/configreference/system.webserver/security/requestfiltering/requestlimits) 28.6MB _(returns `404.13 Content Length Too Large`)_
**ngnix** [default](http://nginx.org/en/docs/http/ngx_http_core_module.html#client_max_body_size) `client_max_body_size` 1MB
**apache.tomcat** [default](http://tomcat.apache.org/tomcat-9.0-doc/config/http.html) `maxPostSize` 2MB
**apache.httpd** [default](http://httpd.apache.org/docs/2.4/mod/core.html#LimitRequestBody) `LimitRequestBody` unlimited

Will currently be enforced by reverse proxy config (e.g. IIS/nginx) when deployed in recommended configuration - but not in pure Kestrel deployment.

halter73:
I think we should be careful when implementing this not to limit upgraded/websocket requests.

muratg:
https://github.com/aspnet/KestrelHttpServer/issues/475

muratg:
cc @blowdart

DamianEdwards:
post-RC2

natemcmaster:
I've forgotten what we were planning to do here. Should this be reopened since we reverted the issue that originally closed it? 

Closed #1244
Reverted: https://github.com/aspnet/KestrelHttpServer/pull/1253

benaadams:
Solved in Kestrel by the back pressure https://github.com/aspnet/KestrelHttpServer/pull/1244#issuecomment-265546975 maybe be an issue further up the pipeline if it can't handle the data; but isn't a server issue per se

Tratcher:
[...]

PR #1244, opened 2016-11-30
Add MaxRequestBodySize parameter (#478)

cesarbs:
#478

@halter73 @pakrym @natemcmaster

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Kestrel/KestrelServerLimits.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/KestrelServerLimits.cs
@@ -23,6 +23,10 @@ namespace Microsoft.AspNetCore.Server.Kestrel
         // Matches the default LimitRequestFields in Apache httpd.
         private int _maxRequestHeaderCount = 100;
 
+        // Matches default HttpRuntimeSection.MaxRequestLength.
+        // https://msdn.microsoft.com/en-us/library/system.web.configuration.httpruntimesection.maxrequestlength%28v=vs.100%29.aspx
+        private long _maxRequestBodySize = 4 * 1024 * 1024;
+
         // Matches the default http.sys connectionTimeout.
         private TimeSpan _keepAliveTimeout = TimeSpan.FromMinutes(2);
 
@@ -141,6 +145,28 @@ namespace Microsoft.AspNetCore.Server.Kestrel
                     throw new ArgumentOutOfRangeException(nameof(value), "Value must a positive integer.");
                 }
                 _maxRequestHeaderCount = value;
+            }
+        }
+
+        /// <summary>
+        /// Gets or sets the maximum request body size, in bytes.
+        /// </summary>
+        /// <remarks>
+        /// Defaults to 4MB.
+        /// </remarks>
+        public long MaxRequestBodySize
+        {
+            get
+            {
+                return _maxRequestBodySize;
+            }
+            set
+            {
+                if (value <= 0)
+                {
+                    throw new ArgumentOutOfRangeException(nameof(value), "Value must a positive integer.");
+                }
+                _maxRequestBodySize = value;
             }
         }
 
--- a/src/Microsoft.AspNetCore.Server.Kestrel/BadHttpRequestException.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/BadHttpRequestException.cs
@@ -78,6 +78,9 @@ namespace Microsoft.AspNetCore.Server.Kestrel
                 case RequestRejectionReason.RequestTimeout:
                     ex = new BadHttpRequestException("Request timed out.", 408);
                     break;
+                case RequestRejectionReason.PayloadTooLarge:
+                    ex = new BadHttpRequestException("Payload too large.", 413);
+                    break;
                 default:
                     ex = new BadHttpRequestException("Bad request.", 400);
                     break;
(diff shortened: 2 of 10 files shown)


benaadams:
doit! 💃

cesarbs:
Ready for review.

benaadams:
That's surprisingly clean 😄

src/Microsoft.AspNetCore.Server.Kestrel/KestrelServerLimits.cs
```diff
@@ -23,6 +23,9 @@ public class KestrelServerLimits
         // Matches the default LimitRequestFields in Apache httpd.
         private int _maxRequestHeaderCount = 100;
 
+        // Matches the default maxRequestEntityAllowed in IIS.
+        private long _maxRequestBodySize = 200 * 1024;
```
natemcmaster: Any particular reason for matching IIS instead of nginx or tomcats's default size?

cesarbs: Nope, I just picked one. Needs design I guess. cc @shirhatti

benaadams: Picking one at random; .NET 4-4.6?


**.NET 4 - 4.6** [default](https://msdn.microsoft.com/en-us/library/system.web.configuration.httpruntimesection.maxrequestlength%28v=vs.100%29.aspx) `HttpRuntimeSection.MaxRequestLength` 4MB
**IIS** [default](https://www.iis.net/configreference/system.webserver/asp/limits) `maxRequestEntityAllowed` 200kB
or filtered [default](https://www.iis.net/configreference/system.webserver/security/requestfiltering/requestlimits) 28.6MB _(returns `404.13 Content Length Too Large`)_
**ngnix** [default](http://nginx.org/en/docs/http/ngx_http_core_module.html#client_max_body_size) `client_max_body_size` 1MB
**apache.tomcat** [default](http://tomcat.apache.org/tomcat-9.0-doc/config/http.html) `maxPostSize` 2MB
**apache.httpd** [default](http://httpd.apache.org/docs/2.4/mod/core.html#LimitRequestBody) `LimitRequestBody` unlimited

halter73:
Any tests verifying that the MaxRequestBodySize configuration doesn't affect Upgrade/WebSocket/ForRemainingData requests?

src/Microsoft.AspNetCore.Server.Kestrel/KestrelServerLimits.cs
blowdart: As an aside the comment needs to say it's in bytes or kilobytes or megabytes or whatever. Or the variable name needs to state it.

cesarbs:
@halter73 `MaxRequestBodySizeNotEnforcedOnUpgradedConnection`

halter73:
Perfect.

halter73:
:shipit:

src/Microsoft.AspNetCore.Server.Kestrel/KestrelServerLimits.cs
cesarbs: @blowdart The comment for the public property states that it's in bytes.

natemcmaster: +1 for 4 MB as default. 200kB seems too small. One common scenario that usually exceeds 200 kB: image uploads.

cesarbs:
Changed default to 4MB.

natemcmaster:
:shipit:

src/Microsoft.AspNetCore.Server.Kestrel/KestrelServerLimits.cs
```diff
@@ -23,6 +23,10 @@ public class KestrelServerLimits
         // Matches the default LimitRequestFields in Apache httpd.
         private int _maxRequestHeaderCount = 100;
 
+        // Matches default HttpRuntimeSection.MaxRequestLength.
+        // https://msdn.microsoft.com/en-us/library/system.web.configuration.httpruntimesection.maxrequestlength%28v=vs.100%29.aspx
+        private long _maxRequestBodySize = 4 * 1024 * 1024;
```
Tratcher: IIS7 raised this to 28.6MB
https://www.iis.net/configreference/system.webserver/security/requestfiltering/requestlimits

Tratcher:
@cesarbs @muratg this needs an Announcement.

halter73:
@davidfowl Would you remove this change entirely? Or make the default unlimited?

If we want to have endpoint specific configuration, that should probably be middleware. Unfortunately, middleware is almost always opt-in. Even if we update templates, people upgrading existing applications to new packages would still not have request body limits.

cesarbs:
@benaadams Can you tell us why this feature is so important to you (or so it seems 😄)?

benaadams:
@cesarbs I could tell you; but @blowdart wouldn't approve 😉 

I would suggest it is poor practice to have a server that by default allows someone external to it flood its memory infinitely; even if its only via localhost...

blowdart:
So kestrel, as a server, doesn't read the whole request, that's further down the stack, so memory flooding doesn't come into it until it reachs, say, a forms processor. Or so @davidfowl assures me.

benaadams:
Ahh... because `IBufferSizeControl.MaxRequestBufferSize` will throttle it?

blowdart:
I hope that's his thinking :)

benaadams:
May not need this then I suppose...

davidfowl:
Yes please, lets remove it. Sorry @cesarbs 😄

cesarbs:
Bummer.

PR #1246, opened 2016-12-05
Handle requests using absolute-form request URIs

### LINKED ISSUE #666: 500 when abs_path present in http_URL - 3.2.2 http URL

natemcmaster:
An absolute-form request URI has a start line in form: "GET http://host/path HTTP/1.1".
RFC 7230 section 5.3.2 stipulates that servers should allow absolute-form request URIs.

This change will handles requests using absolute-form. The scheme and authority
section of the absolute URI are ignored, but will still appear in `IHttpRequestFeature.RawTarget`.

Resolves #666 

cc @cesarbs @Tratcher

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
@@ -29,10 +29,13 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Http
         private const byte ByteCR = (byte)'\r';
         private const byte ByteLF = (byte)'\n';
         private const byte ByteColon = (byte)':';
+        private const byte ByteForwardSlash = (byte)'/';
         private const byte ByteSpace = (byte)' ';
         private const byte ByteTab = (byte)'\t';
         private const byte ByteQuestionMark = (byte)'?';
         private const byte BytePercentage = (byte)'%';
+
+        private const string EmptyPath = "/";
 
         private static readonly ArraySegment<byte> _endChunkedResponseBytes = CreateAsciiByteArraySegment("0\r\n\r\n");
         private static readonly ArraySegment<byte> _continueBytes = CreateAsciiByteArraySegment("HTTP/1.1 100 Continue\r\n\r\n");
@@ -976,7 +979,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Http
 
         public RequestLineStatus TakeStartLine(SocketInput input)
         {
-            const int MaxInvalidRequestLineChars = 32;
+            // expected start line format: https://tools.ietf.org/html/rfc7230#section-3.1.1
 
             var scan = input.ConsumingStart();
             var start = scan;
@@ -1014,22 +1017,21 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Http
                 }
                 end.Take();
 
+                // begin consuming method
                 string method;
                 var begin = scan;
                 if (!begin.GetKnownMethod(out method))
                 {
                     if (scan.Seek(ByteSpace, ref end) == -1)
                     {
-                        RejectRequest(RequestRejectionReason.InvalidRequestLine,
-                            Log.IsEnabled(LogLevel.Information) ? start.GetAsciiStringEscaped(end, MaxInvalidRequestLineChars) : string.Empty);
+                        RejectInvalidStartLine(start, end);
                     }
 
                     method = begin.GetAsciiString(ref scan);
 
                     if (method == null)
                     {
-                        RejectRequest(RequestRejectionReason.InvalidRequestLine,
-                            Log.IsEnabled(LogLevel.Information) ? start.GetAsciiStringEscaped(end, MaxInvalidRequestLineChars) : string.Empty);
+                        RejectInvalidStartLine(start, end);
                     }
 
                     // Note: We're not in the fast path any more (GetKnownMethod should have handled any HTTP Method we're aware of)
@@ -1038,8 +1040,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Http
                     {
                         if (!IsValidTokenChar(method[i]))
                         {
-                            RejectRequest(RequestRejectionReason.InvalidRequestLine,
-                                Log.IsEnabled(LogLevel.Information) ? start.GetAsciiStringEscaped(end, MaxInvalidRequestLineChars) : string.Empty);
+                            RejectInvalidStartLine(start, end);
                         }
                     }
                 }
@@ -1048,14 +1049,38 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Http
                     scan.Skip(method.Length);
                 }
 
+                // consume space
                 scan.Take();
+
+                // begin consuming request-target
                 begin = scan;
+
+                var targetBegin = scan;
+
+                if (targetBegin.Peek() != ByteForwardSlash)
+                {
+                    string requestUriScheme;
+                    if (scan.GetKnownHttpSchema(out requestUriScheme))
+                    {
+                        // Start-line can contain uri in absolute form. e.g. 'GET http://contoso.com/favicon.ico HTTP/1.1'
+                        // Clients should only send this to proxies, but the spec requires we handle it anyways.
+                        // cref https://tools.ietf.org/html/rfc7230#section-5.3
+                        scan.Skip(requestUriScheme.Length);
+
+                        if (scan.Seek(ByteForwardSlash, ByteSpace, ByteQuestionMark, ref end) == -1)
+                        {
+                            RejectInvalidStartLine(start, end);
+                        }
+
+                        begin = scan;
+                    }
+                }
+
                 var needDecode = false;
                 var chFound = scan.Seek(ByteSpace, ByteQuestionMark, BytePercentage, ref end);
                 if (chFound == -1)
                 {
-                    RejectRequest(RequestRejectionReason.InvalidRequestLine,
-                        Log.IsEnabled(LogLevel.Information) ? start.GetAsciiStringEscaped(end, MaxInvalidRequestLineChars) : string.Empty);
+                    RejectInvalidStartLine(start, end);
                 }
                 else if (chFound == BytePercentage)
                 {
@@ -1063,40 +1088,39 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Http
                     chFound = scan.Seek(ByteSpace, ByteQuestionMark, ref end);
                     if (chFound == -1)
                     {
-                        RejectRequest(RequestRejectionReason.InvalidRequestLine,
-                            Log.IsEnabled(LogLevel.Information) ? start.GetAsciiStringEscaped(end, MaxInvalidRequestLineChars) : string.Empty);
+                        RejectInvalidStartLine(start, end);
                     }
                 }
 
                 var pathBegin = begin;
                 var pathEnd = scan;
 
-                var queryString = "";
+                var queryString = string.Empty;
                 if (chFound == ByteQuestionMark)
                 {
                     begin = scan;
                     if (scan.Seek(ByteSpace, ref end) == -1)
                     {
-                        RejectRequest(RequestRejectionReason.InvalidRequestLine,
-                            Log.IsEnabled(LogLevel.Information) ? start.GetAsciiStringEscaped(end, MaxInvalidRequestLineChars) : string.Empty);
+                        RejectInvalidStartLine(start, end);
                     }
                     queryString = begin.GetAsciiString(ref scan);
                 }
 
                 var queryEnd = scan;
 
-                if (pathBegin.Peek() == ByteSpace)
+                if (pathBegin.Peek() == ByteSpace && targetBegin.Index == pathBegin.Index)
                 {
-                    RejectRequest(RequestRejectionReason.InvalidRequestLine,
-                        Log.IsEnabled(LogLevel.Information) ? start.GetAsciiStringEscaped(end, MaxInvalidRequestLineChars) : string.Empty);
+                    RejectInvalidStartLine(start, end);
                 }
 
+                // consume space
                 scan.Take();
+
+                // begin consuming HTTP-version
                 begin = scan;
                 if (scan.Seek(ByteCR, ref end) == -1)
                 {
-                    RejectRequest(RequestRejectionReason.InvalidRequestLine,
-                        Log.IsEnabled(LogLevel.Information) ? start.GetAsciiStringEscaped(end, MaxInvalidRequestLineChars) : string.Empty);
+                    RejectInvalidStartLine(start, end);
                 }
 
                 string httpVersion;
@@ -1106,8 +1130,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Http
 
                     if (httpVersion == string.Empty)
                     {
-                        RejectRequest(RequestRejectionReason.InvalidRequestLine,
-                            Log.IsEnabled(LogLevel.Information) ? start.GetAsciiStringEscaped(end, MaxInvalidRequestLineChars) : string.Empty);
+                        RejectInvalidStartLine(start, end);
                     }
                     else
                     {
@@ -1115,11 +1138,11 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Http
                     }
                 }
 
-                scan.Take(); // consume CR
+                // consume CR
+                scan.Take();
                 if (scan.Take() != ByteLF)
                 {
-                    RejectRequest(RequestRejectionReason.InvalidRequestLine,
-                        Log.IsEnabled(LogLevel.Information) ? start.GetAsciiStringEscaped(end, MaxInvalidRequestLineChars) : string.Empty);
+                    RejectInvalidStartLine(start, end);
                 }
 
                 // URIs are always encoded/escaped to ASCII https://tools.ietf.org/html/rfc3986#page-11
@@ -1130,7 +1153,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Http
                 if (needDecode)
                 {
                     // Read raw target before mutating memory.
-                    rawTarget = pathBegin.GetAsciiString(ref queryEnd);
+                    rawTarget = targetBegin.GetAsciiString(ref queryEnd);
 
                     // URI was encoded, unescape and then parse as utf8
                     pathEnd = UrlPathDecoder.Unescape(pathBegin, pathEnd);
@@ -1141,19 +1164,34 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Http
                     // URI wasn't encoded, parse as ASCII
                     requestUrlPath = pathBegin.GetAsciiString(ref pathEnd);
 
-                    if (queryString.Length == 0)
+                    if (queryString.Length == 0 && targetBegin.Index == pathBegin.Index)
                     {
                         // No need to allocate an extra string if the path didn't need
-                        // decoding and there's no query string following it.
+                        // decoding and there's no query string following it, and the
+                        // request-target isn't absolute-form
                         rawTarget = requestUrlPath;
                     }
                     else
                     {
-                        rawTarget = pathBegin.GetAsciiString(ref queryEnd);
+                        rawTarget = targetBegin.GetAsciiString(ref queryEnd);
                     }
                 }
 
-                var normalizedTarget = PathNormalizer.RemoveDotSegments(requestUrlPath);
+                if (targetBegin.Index < pathBegin.Index)
+                {
+                    // validation of absolute-form URI may be slow, but clients
+                    // should not be sending this form anyways, so perf optimization 
+                    // not high priority
+                    Uri _;
+                    if (!Uri.TryCreate(rawTarget, UriKind.Absolute, out _))
+                    {
+                        RejectInvalidStartLine(start, end);
+                    }
+                }
+
+                var normalizedUrlPath = requestUrlPath == null
+                    ? EmptyPath
+                    : PathNormalizer.RemoveDotSegments(requestUrlPath);
 
                 consumed = scan;
                 Method = method;
@@ -1162,17 +1200,24 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Http
                 HttpVersion = httpVersion;
 
                 bool caseMatches;
-                if (RequestUrlStartsWithPathBase(normalizedTarget, out caseMatches))
+                if (RequestUrlStartsWithPathBase(normalizedUrlPath, out caseMatches))
                 {
-                    PathBase = caseMatches ? _pathBase : normalizedTarget.Substring(0, _pathBase.Length);
-                    Path = normalizedTarget.Substring(_pathBase.Length);
+                    // request-target is in origin-form or absolute-form 
+                    // and path should be adjusted for matching the server base path
+                    PathBase = caseMatches ? _pathBase : normalizedUrlPath.Substring(0, _pathBase.Length);
+                    Path = normalizedUrlPath.Substring(_pathBase.Length);
                 }
-                else if (rawTarget[0] == '/') // check rawTarget since normalizedTarget can be "" or "/" after dot segment removal
+                else if ((requestUrlPath?.Length > 0 && requestUrlPath[0] == '/')
+                    || (rawTarget.Length > 0 && rawTarget[0] == '/')
+                    || ReferenceEquals(normalizedUrlPath, EmptyPath))
                 {
-                    Path = normalizedTarget;
+                    // request-target is in origin-form or absolute-form
+                    Path = normalizedUrlPath;
                 }
                 else
                 {
+                    // request-target is in asterisk-form and authority-form
+                    // also the catch-all for other malformed request-targets
                     Path = string.Empty;
                     PathBase = string.Empty;
                     QueryString = string.Empty;
@@ -1478,6 +1523,16 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Http
             throw new ObjectDisposedException(
                     "The response has been aborted due to an unhandled application exception.",
                     _applicationException);
+        }
+
+        private void RejectInvalidStartLine(MemoryPoolIterator start, MemoryPoolIterator end)
+        {
+            const int MaxInvalidRequestLineChars = 32;
+
+            RejectRequest(RequestRejectionReason.InvalidRequestLine,
+                Log.IsEnabled(LogLevel.Information)
+                    ? start.GetAsciiStringEscaped(end, MaxInvalidRequestLineChars)
+                    : string.Empty);
         }
 
         public void RejectRequest(RequestRejectionReason reason)
--- a/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Infrastructure/MemoryPoolIteratorExtensions.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Infrastructure/MemoryPoolIteratorExtensions.cs
@@ -14,6 +14,9 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Infrastructure
         public const string Http10Version = "HTTP/1.0";
         public const string Http11Version = "HTTP/1.1";
 
+        public const string HttpScheme = "http://";
+        public const string HttpsScheme = "https://";
+
         // readonly primitive statics can be Jit'd to consts https://github.com/dotnet/coreclr/issues/1079
         private readonly static ulong _httpConnectMethodLong = GetAsciiStringAsLong("CONNECT ");
         private readonly static ulong _httpDeleteMethodLong = GetAsciiStringAsLong("DELETE \0");
@@ -27,6 +30,9 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Infrastructure
 
         private readonly static ulong _http10VersionLong = GetAsciiStringAsLong("HTTP/1.0");
         private readonly static ulong _http11VersionLong = GetAsciiStringAsLong("HTTP/1.1");
+
+        private readonly static ulong _httpSchemeLong = GetAsciiStringAsLong("http://\0");
+        private readonly static ulong _httpsSchemeLong = GetAsciiStringAsLong("https://");
 
         private readonly static ulong _mask8Chars = GetMaskAsLong(new byte[] { 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff });
         private readonly static ulong _mask7Chars = GetMaskAsLong(new byte[] { 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0x00 });
@@ -152,6 +158,37 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Infrastructure
                     knownMethod = x.Item3;
                     return true;
                 }
+            }
+
+            return false;
+        }
+
+        /// <summary>
+        /// Checks 8 bytes from <paramref name="begin"/> that correspond to 'http://' or 'https://'
+        /// </summary>
+        /// <param name="begin">The iterator</param>
+        /// <param name="knownScheme">A reference to the known scheme, if the input matches any</param>
+        /// <returns>True when memory starts with known http or https schema</returns>
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public static bool GetKnownHttpSchema(this MemoryPoolIterator begin, out string knownScheme)
+        {
+            knownScheme = null;
+            ulong value;
+            if (!begin.TryPeekLong(out value))
+            {
+                return false;
+            }
+
+            if ((value & _mask7Chars) == _httpSchemeLong)
+            {
+                knownScheme = HttpScheme;
+                return true;
+            }
+
+            if (value == _httpsSchemeLong)
+            {
+                knownScheme = HttpsScheme;
+                return true;
             }
 
             return false;
--- a/test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/RequestTests.cs
+++ b/test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/RequestTests.cs
@@ -453,6 +453,72 @@ namespace Microsoft.AspNetCore.Server.Kestrel.FunctionalTests
             }
         }
 
+        [Theory]
+        [InlineData("http://localhost/abs/path", "/abs/path", null)]
+        [InlineData("https://localhost/abs/path", "/abs/path", null)] // handles mismatch scheme
+        [InlineData("https://localhost:22/abs/path", "/abs/path", null)] // handles mismatched ports
+        [InlineData("https://differenthost/abs/path", "/abs/path", null)] // handles mismatched hostname
+        [InlineData("http://localhost/", "/", null)]
+        [InlineData("http://root@contoso.com/path", "/path", null)]
+        [InlineData("http://root:password@contoso.com/path", "/path", null)]
+        [InlineData("https://localhost/", "/", null)]
+        [InlineData("http://localhost", "/", null)]
+        [InlineData("http://127.0.0.1/", "/", null)]
+        [InlineData("http://[::1]/", "/", null)]
+        [InlineData("http://[::1]:8080/", "/", null)]
+        [InlineData("http://localhost?q=123&w=xyz", "/", "123")]
+        [InlineData("http://localhost/?q=123&w=xyz", "/", "123")]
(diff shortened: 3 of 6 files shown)


src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
```diff
@@ -1118,31 +1154,33 @@ public RequestLineStatus TakeStartLine(SocketInput input)
...
-                var normalizedTarget = PathNormalizer.RemoveDotSegments(requestUrlPath);
+                var normalizedUrlPath = requestUrlPath == null
+                    ? string.Empty
+                    : PathNormalizer.RemoveDotSegments(requestUrlPath);
 
                 consumed = scan;
                 Method = method;
                 QueryString = queryString;
-                RawTarget = rawTarget;
+                RawTarget = requestUriScheme + requestAuthority + rawUrlPath;
```
cesarbs: I think you'll see a perf hit from these concatenations. Ideally you should get the raw target in one go and set it here, instead of building it from it's parts.

cesarbs: Also, `requestAuthority` might be null, no?

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/RequestTests.cs
```diff
@@ -420,6 +420,72 @@ public void CanUpgradeRequestWithConnectionKeepAliveUpgradeHeader()
...
+                    app.Run(async context =>
+                    {
+                        pathTcs.TrySetResult(context.Request.Path);
+                        hostTcs.TrySetResult(context.Request.Host);
+                        rawTargetTcs.TrySetResult(context.Features.Get<IHttpRequestFeature>().RawTarget);
+                        await context.Response.WriteAsync("Done");
+                    });
+                });
+
+            using (var host = builder.Build())
```
cesarbs: Use `TestServer`

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/RequestTests.cs
```diff
@@ -420,6 +420,72 @@ public void CanUpgradeRequestWithConnectionKeepAliveUpgradeHeader()
...
+            {
+                host.Start();
+
+                using (var socket = new Socket(AddressFamily.InterNetwork, SocketType.Stream, ProtocolType.Tcp))
```
cesarbs: Use `TestConnection`

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/Utilities/TaskExtensions.cs
```diff
@@ -0,0 +1,22 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Runtime.CompilerServices;
+using System.Threading.Tasks;
+
+namespace Microsoft.AspNetCore.Server.Kestrel.FunctionalTests
+{
+    public static class TaskExtensions
```
cesarbs: We already have this in https://github.com/aspnet/Testing/blob/dev/src/Microsoft.AspNetCore.Testing/TaskExtensions.cs

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/RequestTests.cs
```diff
@@ -420,6 +420,72 @@ public void CanUpgradeRequestWithConnectionKeepAliveUpgradeHeader()
...
+        [InlineData("http://localhost/abs/path", true, "/abs/path")]
+        [InlineData("https://localhost/abs/path", true, "/abs/path")] // handles mismatch scheme
+        [InlineData("https://localhost:22/abs/path", true, "/abs/path")] // handles mismatched ports
+        [InlineData("https://differenthost/abs/path", true, "/abs/path")] // handles mismatched hostname
+        [InlineData("http://localhost/", true, "/")]
+        [InlineData("https://localhost/", true, "/")]
+        [InlineData("http://localhost", true, "")]
+        [InlineData("http://", false, null)]
+        [InlineData("https://", false, null)]
+        public async Task CanHandleRequestsWithUrlInAbsoluteForm(string requestUrl, bool valid, string expectedPath)
```
cesarbs: Add test cases for bad requests with malformed absolute URLs.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Infrastructure/MemoryPoolIteratorExtensions.cs
```diff
@@ -158,6 +164,42 @@ public static bool GetKnownMethod(this MemoryPoolIterator begin, out string know
...
+        /// <remarks>
+        /// Currently recognizes these schemes:
+        /// - 'http://'
+        /// - 'https://'
+        /// </remarks>
+        /// <param name="begin">The iterator</param>
+        /// <param name="knownScheme">A reference to the known scheme, if the input matches any</param>
+        /// <returns></returns>
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public static bool GetKnownUriScheme(this MemoryPoolIterator begin, out string knownScheme)
```
cesarbs: Add unit tests (look at existing tests for the other `GetKnown`* methods).

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
natemcmaster: requestUriScheme  and requestAuthority will be null for most requests. Isn't concat with null a noop? Since the perf lab is down right now, I haven't been able to test if this incurs a significant perf hit or not.

cesarbs: Whoa, TIL. I thought you'd get a `NullReferenceException` when trying to concat a `null`.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
```diff
@@ -1021,8 +1025,38 @@ public RequestLineStatus TakeStartLine(SocketInput input)
...
+                scan.Take(); // consume space
+
+                // begin consuming request-target
                 begin = scan;
+
+                string requestUriScheme;
+                string requestAuthority = null;
+                if (scan.GetKnownUriScheme(out requestUriScheme))
+                {
+                    // Request URIs can be in absolute form
```
Tratcher: example

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
```diff
@@ -1021,8 +1025,38 @@ public RequestLineStatus TakeStartLine(SocketInput input)
...
+                    // see https://tools.ietf.org/html/rfc3986#section-4.3
+
+                    var pathIndex = scan.Seek(ByteForwardSlash, ByteSpace, ref end);
+                    if (pathIndex == -1)
+                    {
+                        RejectRequest(RequestRejectionReason.InvalidRequestLine,
+                            Log.IsEnabled(LogLevel.Information) ? start.GetAsciiStringEscaped(end, MaxInvalidRequestLineChars) : string.Empty);
+                    }
+
+                    // TODO consider handling unicode host names
```
Tratcher: UTF-8

natemcmaster:
🆙 :date:

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
```diff
@@ -1030,29 +1030,40 @@ public RequestLineStatus TakeStartLine(SocketInput input)
...
+                    targetIsAbsolute = true;
                     scan.Skip(requestUriScheme.Length);
                     begin = scan;
 
-                    // an absolute URI is not required to end in a slash but must not be empty
-                    // see https://tools.ietf.org/html/rfc3986#section-4.3
+                    var chNext = scan.Peek();
+                    if (chNext == ByteForwardSlash || chNext == ByteSpace)
+                    {
+                        // an absolute URI is not required to end in a slash but host must not be empty
```
cesarbs: Move this comment outside the `if`

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/RequestTests.cs
```diff
@@ -428,59 +429,44 @@ public void CanUpgradeRequestWithConnectionKeepAliveUpgradeHeader()
         [InlineData("http://localhost/", true, "/")]
         [InlineData("https://localhost/", true, "/")]
         [InlineData("http://localhost", true, "")]
+        // bad requests
```
cesarbs: Write a separate test method for bad requests. Also, you can add those to the `BadHttpRequestTests` class.

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/RequestTests.cs
```diff
@@ -428,59 +429,44 @@ public void CanUpgradeRequestWithConnectionKeepAliveUpgradeHeader()
...
+                using (var connection = server.CreateConnection())
                 {
-                    socket.Connect(new IPEndPoint(IPAddress.Loopback, host.GetPort()));
-                    var raw = $"GET {requestUrl} HTTP/1.1\r\n" +
-                              "Content-Length: 0\r\n" +
-                              "Host: localhost\r\n" +
-                              "Connection: close\r\n" +
-                              "\r\n";
+                    await connection.SendAll(
+                       $"GET {requestUrl} HTTP/1.1",
```
cesarbs: nit: indent

natemcmaster:
This implementation ignores the authority section altogether, which means some malformed URI's could still be accepted. Is this okay? e.g. `GET http://:80/abc HTTP/1.1` would be accepted.

 Or do should we do a safe but slow System.Uri check to make sure the full URI is valid?

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
```diff
@@ -1021,8 +1025,49 @@ public RequestLineStatus TakeStartLine(SocketInput input)
...
                 }
 
-                scan.Take();
+                scan.Take(); // consume space
+
+                // begin consuming request-target
                 begin = scan;
+
+                var targetIsAbsolute = false;
+                string requestUriScheme;
```
Tratcher: still not sure why you're creating these strings only to rebuild the rawTarget later. why not capture the full rawTarget all at once?

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Infrastructure/MemoryPoolIteratorExtensions.cs
```diff
@@ -158,6 +164,37 @@ public static bool GetKnownMethod(this MemoryPoolIterator begin, out string know
...
+        public static bool GetKnownHttpSchema(this MemoryPoolIterator begin, out string knownScheme)
+        {
+            knownScheme = null;
+            ulong value;
+            if (!begin.TryPeekLong(out value))
+            {
+                return false;
+            }
+
+            if ((value & _mask7Chars) == _httpSchemeLong)
```
Tratcher: does the spec allow for ws:// or wss:// (WebSockets)?

Tratcher:
So long as we're not acting on the unvalidated values it should be fine.

test/Microsoft.AspNetCore.Server.KestrelTests/BadHttpRequestTests.cs
```diff
@@ -52,6 +52,12 @@ public class BadHttpRequestTests
         [InlineData("} / HTTP/1.0\r\n")]
         [InlineData("get@ / HTTP/1.0\r\n")]
         [InlineData("post= / HTTP/1.0\r\n")]
+        // malformed request target
+        [InlineData("GET http:// HTTP/1.1\r\n")]
+        [InlineData("GET https:// HTTP/1.1\r\n")]
+        [InlineData("GET http:/// HTTP/1.1\r\n")]
+        [InlineData("GET http://// HTTP/1.1\r\n")]
```
cesarbs: Add:

- Not starting with `http://` or `https://`
- `http`
- `https`
- `http:`
- `https:/`
- `http:/`
- `https:/`
- `http:// ` (note the extra space)
- `https:// `
- `http:// /`
- `https:// /`

test/Microsoft.AspNetCore.Server.KestrelTests/MemoryPoolIteratorTests.cs
```diff
@@ -606,6 +606,68 @@ public void GetsKnownMethod(string input, bool expectedResult, string expectedKn
         }
 
         [Theory]
+        [InlineData("http://abc", true, MemoryPoolIteratorExtensions.HttpScheme)]
+        [InlineData("http:// ", true, MemoryPoolIteratorExtensions.HttpScheme)]
+        [InlineData("http:/  ", false, null)]
+        [InlineData("HtTp:// ", false, null)]
+        [InlineData("https://", true, MemoryPoolIteratorExtensions.HttpsScheme)]
+        [InlineData("https://abc", true, MemoryPoolIteratorExtensions.HttpsScheme)]
+        public void GetKnownHttpSchema(string input, bool expectedResult, string expectedString)
```
cesarbs: nit: `Gets`

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/RequestTests.cs
```diff
@@ -420,6 +421,45 @@ public void CanUpgradeRequestWithConnectionKeepAliveUpgradeHeader()
...
         }
 
+        [Theory]
+        [InlineData("http://localhost/abs/path", "/abs/path")]
+        [InlineData("https://localhost/abs/path", "/abs/path")] // handles mismatch scheme
+        [InlineData("https://localhost:22/abs/path", "/abs/path")] // handles mismatched ports
+        [InlineData("https://differenthost/abs/path", "/abs/path")] // handles mismatched hostname
+        [InlineData("http://localhost/", "/")]
+        [InlineData("https://localhost/", "/")]
+        [InlineData("http://localhost", "")]
```
cesarbs: If we decide things like `http://:80` are ok, you should add it here.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
natemcmaster: Will update

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Infrastructure/MemoryPoolIteratorExtensions.cs
natemcmaster: [RFC 6455 section 4.1](https://tools.ietf.org/html/rfc6455#section-4.1) states that websockets clients are required to send the initial HTTP request using 'origin-form' (normal request format), not 'absolute-form'. I don't think we have a valid client scenario for a request like `GET ws://host/path HTTP/1.1` so we're probably okay rejecting this format.

test/Microsoft.AspNetCore.Server.KestrelTests/BadHttpRequestTests.cs
```diff
@@ -52,12 +52,26 @@ public class BadHttpRequestTests
         [InlineData("} / HTTP/1.0\r\n")]
```
Tratcher: Is there a test somewhere for `OPTIONS * HTTP/1.1`? Path should be empty and RawTarget should be *.

cesarbs: Yes: `KestrelTests.RequestTargetProcessingTests.NonPathRequestTargetSetInRawTarget`.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
```diff
@@ -1021,8 +1025,34 @@ public RequestLineStatus TakeStartLine(SocketInput input)
...
+
+                // begin consuming request-target
                 begin = scan;
+
+                var targetBegin = scan;
+
+                if (targetBegin.Peek() != ByteForwardSlash)
+                {
+                    string requestUriScheme;
+                    if (scan.GetKnownHttpSchema(out requestUriScheme))
```
cesarbs: Given what we discussed, I don't think you need this method anymore - we're not really looking for any particular schemas, since we don't do anything about them. We do want to make sure it's a valid URI, otherwise we could just accept any kind of `<garbage>/a/b/c`, and that wouldn't make sense.

So here you could check if there isn't a `/`, like you're doing, and if there isn't this is the slow path and you grab the string from `targetBegin` to the slash and check if that's a valid URI (like you did below). Otherwise you're in the hot path and you should do things as before.

natemcmaster: Yes, as Ben said, one goal was to minimize allocations. 

Also, AFAIK there are no **valid** scenarios using absolute-form and schemes other than `http://` and`https://`. Kestrel should reject requests like `GET otherscheme://localhost/ HTTP/1.1` as invalid.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Infrastructure/MemoryPoolIteratorExtensions.cs
halter73: Technically it's not a valid client scenario to send absolute-form targets to the end-server, right? It's only meant to be sent to proxies AFAIK. I recall someone mentioning the spec requires that proxies change the target to origin-form, but end-servers still support absolute-form in case the proxy isn't compliant with the spec.

If we already going through this work to support buggy proxies, I think we should also allow ws/wss.

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/RequestTests.cs
```diff
@@ -420,6 +421,50 @@ public void CanUpgradeRequestWithConnectionKeepAliveUpgradeHeader()
...
+        [Theory]
+        [InlineData("http://localhost/abs/path", "/abs/path")]
+        [InlineData("https://localhost/abs/path", "/abs/path")] // handles mismatch scheme
+        [InlineData("https://localhost:22/abs/path", "/abs/path")] // handles mismatched ports
+        [InlineData("https://differenthost/abs/path", "/abs/path")] // handles mismatched hostname
+        [InlineData("http://localhost/", "/")]
+        [InlineData("http://root@contoso.com/path", "/path")]
+        [InlineData("http://root:password@contoso.com/path", "/path")]
+        [InlineData("https://localhost/", "/")]
+        [InlineData("http://localhost", "")]
```
halter73: Nice that you added a test for this. Don't like the result 😛

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
halter73: The scheme string wouldn't need to be allocated even if we supported unknown schema. The out parameter could be the length of the schema string since the schema from the target isn't stored anywhere.

Better yet, this could be like `GetKnownVersion` and advance the iterator itself. I see that `GetKnownMethod` doesn't advance the iterator. I would change that too for consistency.

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/RequestTests.cs
Tratcher: Agreed, path should not be empty. reject or set it to /

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
```diff
@@ -1103,7 +1135,7 @@ public RequestLineStatus TakeStartLine(SocketInput input)
                 if (needDecode)
                 {
                     // Read raw target before mutating memory.
-                    rawTarget = pathBegin.GetAsciiString(ref queryEnd);
+                    rawTarget = targetBegin.GetAsciiString(ref queryEnd);
```
natemcmaster: IIUC there RawTarget has always been the ASCII representation. Since we ignore the authority portion of the absolute-URI, we never bother to look for a `%`

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/RequestTests.cs
natemcmaster: I'm favor of keeping it and setting to "/". The HTTP spec says is `absolute-form = absolute-URI`, and absolute URI's don't require the trailing slash.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
Tratcher: % doesn't really apply to hosts anyways, they prefer punycode.

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/RequestTests.cs
cesarbs: Agreed.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
```diff
@@ -1021,8 +1025,34 @@ public RequestLineStatus TakeStartLine(SocketInput input)
...
+                        scan.Skip(requestUriScheme.Length);
+
+                        if (scan.Seek(ByteForwardSlash, ByteSpace, ref end) == -1)
+                        {
+                            RejectRequest(RequestRejectionReason.InvalidRequestLine,
+                               Log.IsEnabled(LogLevel.Information) ? start.GetAsciiStringEscaped(end, MaxInvalidRequestLineChars) : string.Empty);
+                        }
+                    
+                        begin = scan;
+                    }
```
cesarbs: `else` reject.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Infrastructure/MemoryPoolIteratorExtensions.cs
natemcmaster: To make a distinction here, we support accepting absolute-form b/c the spec requires it and some **http clients** may send it. I'm in favor of rejecting anything with 'ws://' or 'wss:// because that is a buggy **WebSockets client**.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
natemcmaster: Actually, we don't want to reject yet. e.g. 'asterisk-form'. is still okay.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
```diff
@@ -1045,55 +1046,79 @@ public RequestLineStatus TakeStartLine(SocketInput input)
...
+
+                if (targetBegin.Peek() != ByteForwardSlash)
+                {
+                    string requestUriScheme;
+                    if (scan.GetKnownHttpSchema(out requestUriScheme))
+                    {
+                        // Start-line can contain uri in absolute form. e.g. 'GET http://contoso.com/favicon.ico HTTP/1.1'
+                        // Clients should only send this to proxies, but the spec requires we handle it anyways.
+                        // cref https://tools.ietf.org/html/rfc7230#section-5.3
+
```
cesarbs: nit: remove this blank line

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
```diff
@@ -1138,19 +1162,35 @@ public RequestLineStatus TakeStartLine(SocketInput input)
...
+                        rawTarget = targetBegin.GetAsciiString(ref queryEnd);
+                    }
+                }
+
+                if (targetBegin.Index < pathBegin.Index)
+                {
+                    // validation of absolute-form URI may be slow, but clients
+                    // should not be sending this form anyways, so perf optimization 
+                    // not high priority
+
```
cesarbs: nit: remove this blank line

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/RequestTests.cs
```diff
@@ -453,6 +453,50 @@ public void CanUpgradeRequestWithConnectionKeepAliveUpgradeHeader()
...
+        [InlineData("https://localhost:22/abs/path", "/abs/path")] // handles mismatched ports
+        [InlineData("https://differenthost/abs/path", "/abs/path")] // handles mismatched hostname
+        [InlineData("http://localhost/", "/")]
+        [InlineData("http://root@contoso.com/path", "/path")]
+        [InlineData("http://root:password@contoso.com/path", "/path")]
+        [InlineData("https://localhost/", "/")]
+        [InlineData("http://localhost", "/")]
+        [InlineData("http://127.0.0.1/", "/")]
+        [InlineData("http://[::1]/", "/")]
+        [InlineData("http://[::1]:8080/", "/")]
```
cesarbs: Add cases with:

- Query string
- Percent encoding
- Both

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/RequestTests.cs
```diff
@@ -453,6 +453,50 @@ public void CanUpgradeRequestWithConnectionKeepAliveUpgradeHeader()
...
+                 {
+                     pathTcs.TrySetResult(context.Request.Path);
+                     hostTcs.TrySetResult(context.Request.Host);
+                     rawTargetTcs.TrySetResult(context.Features.Get<IHttpRequestFeature>().RawTarget);
+                     await context.Response.WriteAsync("Done");
+                 }))
+            {
+                using (var connection = server.CreateConnection())
+                {
+                    await connection.SendAll(
```
cesarbs: use `Send`

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/RequestTests.cs
```diff
@@ -453,6 +453,50 @@ public void CanUpgradeRequestWithConnectionKeepAliveUpgradeHeader()
...
+                        $"GET {requestUrl} HTTP/1.1",
+                        "Content-Length: 0",
+                        "Host: localhost",
+                        "",
+                        "");
+
```
cesarbs: `connection.Receive` expected response

test/Microsoft.AspNetCore.Server.KestrelTests/BadHttpRequestTests.cs
```diff
@@ -82,6 +107,11 @@ public class BadHttpRequestTests
         [InlineData("GET / HTTP/1.\r\n")]
         [InlineData("GET / hello\r\n")]
         [InlineData("GET / 8charact\r\n")]
+        [InlineData("GET /  HTTP/1.1\r\n")]
+        [InlineData("GET / / HTTP/1.1\r\n")]
+        [InlineData("GET http://  HTTP/1.1\r\n")]
+        [InlineData("GET https://  HTTP/1.1\r\n")]
+        [InlineData("GET https:// / HTTP/1.1\r\n")]
```
cesarbs: Why are these here? The version is valid in the new test cases.

test/Microsoft.AspNetCore.Server.KestrelTests/FrameTests.cs
```diff
@@ -474,6 +474,35 @@ public void TakeStartLineCallsConsumingCompleteWithFurthestExamined()
...
+        [InlineData("http://host", "/")]
+        [InlineData("http://user@host/", "/")]
+        [InlineData("http://127.0.0.1/", "/")]
+        [InlineData("http://user@127.0.0.1/", "/")]
+        [InlineData("http://user@127.0.0.1:8080/", "/")]
+        [InlineData("http://127.0.0.1:8080/", "/")]
+        [InlineData("http://[::1]", "/")]
+        [InlineData("http://[::1]/path", "/path")]
+        [InlineData("http://[::1]:8080/", "/")]
+        [InlineData("http://user@[::1]:8080/", "/")]
```
cesarbs: Add cases with query string and percent encoding here too.

test/Microsoft.AspNetCore.Server.KestrelTests/MemoryPoolIteratorTests.cs
```diff
@@ -606,6 +606,68 @@ public void GetsKnownMethod(string input, bool expectedResult, string expectedKn
         }
 
         [Theory]
+        [InlineData("http://abc", true, MemoryPoolIteratorExtensions.HttpScheme)]
+        [InlineData("http:// ", true, MemoryPoolIteratorExtensions.HttpScheme)]
+        [InlineData("http:/  ", false, null)]
+        [InlineData("HtTp:// ", false, null)]
```
cesarbs: Add wrong casing tests for https too.

test/Microsoft.AspNetCore.Server.KestrelTests/MemoryPoolIteratorTests.cs
```diff
@@ -606,6 +606,68 @@ public void GetsKnownMethod(string input, bool expectedResult, string expectedKn
...
+        [InlineData("https://", true, MemoryPoolIteratorExtensions.HttpsScheme)]
+        [InlineData("https://abc", true, MemoryPoolIteratorExtensions.HttpsScheme)]
```
cesarbs: Test same conditions on http and https. You're not testing https:/ nor followed by space.

For both: `http:` and `https:`

test/Microsoft.AspNetCore.Server.KestrelTests/BadHttpRequestTests.cs
natemcmaster: These are here because there is a space or a slash where the HTTP version should begin.

natemcmaster:
🆙 📅

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Infrastructure/MemoryPoolIteratorExtensions.cs
```diff
@@ -158,6 +164,39 @@ public static bool GetKnownMethod(this MemoryPoolIterator begin, out string know
...
+            {
+                knownScheme = HttpScheme;
+                begin.Skip(knownScheme.Length);
```
halter73: Is this doing anything? Isn't `begin` a copied struct that isn't used at all afterwards?

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Infrastructure/MemoryPoolIteratorExtensions.cs
```diff
@@ -158,6 +164,39 @@ public static bool GetKnownMethod(this MemoryPoolIterator begin, out string know
         }
 
         /// <summary>
+        /// Checks 8 bytes from <paramref name="begin"/> that correspond to 'http://' or 'https://'
+        /// </summary>
+        /// <param name="begin">The iterator</param>
+        /// <param name="knownScheme">A reference to the known scheme, if the input matches any</param>
+        /// <returns>True when memory starts with known http or https schema</returns>
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public static bool GetKnownHttpSchema(this MemoryPoolIterator begin, out string knownScheme)
```
halter73: Is the out parameter ever used? Would HasKnownHttpSchema be better?

test/Microsoft.AspNetCore.Server.KestrelTests/BadHttpRequestTests.cs
```diff
@@ -52,6 +52,29 @@ public class BadHttpRequestTests
...
+        [InlineData("GET http://:80 HTTP/1.1\r\n")]
+        [InlineData("GET http://:80/abc HTTP/1.1\r\n")]
+        [InlineData("GET http://user@ HTTP/1.1\r\n")]
+        [InlineData("GET http://user@/abc HTTP/1.1\r\n")]
+        [InlineData("GET http://abc%20xyz/abc HTTP/1.1\r\n")]
+        [InlineData("GET http://%20/abc?query=%0A HTTP/1.1\r\n")]
+        // TODO invalidate requests that don't meet the HTTP spec. See https://github.com/aspnet/KestrelHttpServer/issues/1279
+        //[InlineData("GET otherscheme://host/ HTTP/1.1\r\n")]
+        //[InlineData("GET ws://host/ HTTP/1.1\r\n")]
+        //[InlineData("GET wss://host/ HTTP/1.1\r\n")]
```
halter73: If there is an absolute-form request with a non HTTP(S) scheme, does Path become string.Empty since it doesn't start with '/' while not being rejected?

natemcmaster:
Currently, yes. That is the behavior, and was I opened https://github.com/aspnet/KestrelHttpServer/issues/1279 to address.

natemcmaster:
Yeah this can be removed. This method can be HasKnownHttpMethod instead, but we still need an out param to get the length that should be skipped. Will update the PR

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Infrastructure/MemoryPoolIteratorExtensions.cs
natemcmaster: Yes the out param is used.

test/Microsoft.AspNetCore.Server.KestrelTests/FrameTests.cs
cesarbs: Did you miss this one?

cesarbs:
@natemcmaster Can't you skip inside the method itself?

natemcmaster:
cc @pakrym this probably overlaps with work you are doing. Ok if I put this in before? Or would it be better to hold off?

natemcmaster:
In person discussion. Closing this for now as we expect large merge conflicts with work @pakrym is doing, and fixing the bug that this PR addresses is lower priority. We'll come back to this before 2.0.0, after @pakrym's refactorings.

PR #1222, opened 2016-11-22
Add Multithread perf test

benaadams:
Results on 16 core (32 HT) Server

> BenchmarkDotNet=v0.10.0
OS=Windows
Processor=?, ProcessorCount=32
Frequency=2343747 Hz, Resolution=426.6672 ns, Timer=TSC
Host Runtime=.NET Core 4.6.24628.01, Arch=64-bit  [RyuJIT]
GC=Concurrent Server
dotnet cli version=1.0.0-preview2-1-003177
Job Runtime(s):
        .NET Core 4.6.24410.01, Arch=64-bit  [RyuJIT]

> RemoveOutliers=False  Runtime=Core  Server=True
LaunchCount=3  RunStrategy=Throughput  TargetCount=10
WarmupCount=5

Single threaded
```
                   Method |          Mean |      StdDev | Scaled |           RPS |
------------------------- |-------------- |------------ |------- |-------------- |
           ParsePlaintext |   993.3004 ns |  23.2932 ns |   1.00 |  1,006,744.82 |
  ParsePipelinedPlaintext |   752.0448 ns |  15.0722 ns |   0.76 |  1,329,708.02 |
          ParseLiveAspNet | 4,347.4022 ns |  81.1657 ns |   4.38 |    230,022.43 |
 ParsePipelinedLiveAspNet | 4,026.5624 ns |  60.3856 ns |   4.06 |    248,350.80 |
             ParseUnicode | 6,859.2000 ns | 147.5681 ns |   6.91 |    145,789.60 |
    ParseUnicodePipelined | 6,679.8011 ns | 210.6329 ns |   6.73 |    149,705.06 |
```
Multi threaded
```
                   Method |          Mean |      StdDev | Scaled |           RPS |
------------------------- |-------------- |------------ |------- |-------------- |
           ParsePlaintext |   188.0574 ns |   1.5506 ns |   1.00 |  5,317,526.25 |
  ParsePipelinedPlaintext |    67.4741 ns |   1.9620 ns |   0.36 | 14,820,495.27 |
          ParseLiveAspNet |   576.7441 ns |   3.2424 ns |   3.07 |  1,733,871.11 |
 ParsePipelinedLiveAspNet |   309.2693 ns |  11.8032 ns |   1.64 |  3,233,428.06 |
             ParseUnicode |   727.9797 ns |  23.8009 ns |   3.87 |  1,373,664.74 |
[...]

PR #1218, opened 2016-11-18
Handle client FIN as request abort

cesarbs:
#1139

[This](https://github.com/aspnet/KestrelHttpServer/pull/1218/files#diff-4837ca204b80ef23b080588614cea7f8R206) part of the change is mostly to avoid test failures in `HttpsTests.EmptyRequestLoggedAsInformation`. After the change to handle FIN as abort, this test started failing consistently because the `OnConnectionAsync` continuation would run after the abort had taken place, the connection had ended and the server had been disposed - at that point `ReadInputAsync` would try to lease a block from an already disposed memory pool, triggering the assert and failing the entire test run.

@halter73 @Tratcher @pakrym

PR #1252, opened 2016-12-08
Add benchmarks for Frame writes

cesarbs:
@halter73 @pakrym


test/Microsoft.AspNetCore.Server.Kestrel.Performance/project.json
```diff
@@ -2,7 +2,8 @@
   "version": "1.0.0-*",
   "dependencies": {
     "BenchmarkDotNet": "0.10.0",
```
benaadams: Bump to 10.1 and netcoreapp1.1? as per https://github.com/aspnet/KestrelHttpServer/pull/1222/commits/8301721b80d57828fb106699685502c1d0ff5f58

Gives inbuilt accurate memory profiling

test/Microsoft.AspNetCore.Server.Kestrel.Performance/Writing.cs
```diff
@@ -0,0 +1,73 @@
...
+                Output = new MockSocketOutput(),
+                ConnectionControl = Mock.Of<IConnectionControl>()
+            };
+
+            _frame = new TestFrame<object>(application: null, context: connectionContext);
+            _frame.Reset();
+            _frame.InitializeHeaders();
+        }
+
+        [Benchmark]
```
halter73: I think we should have one variation for a set Content-Length and another for chunked responses.

test/Microsoft.AspNetCore.Server.Kestrel.Performance/Writing.cs
```diff
@@ -0,0 +1,125 @@
...
+            {
+                frame.Reset();
+                frame.RequestHeaders.Add("Content-Length", (InnerLoopCount * _writeData.Length).ToString(CultureInfo.InvariantCulture));
+            }
+        }
+
+        [Benchmark(OperationsPerInvoke = InnerLoopCount)]
+        public void Write()
+        {
+            _frameWrite.Write(new ArraySegment<byte>(_writeData));
```
benaadams: Should these functions have a for loop in them?

e.g.
```csharp
for (var i = 0; i < InnerLoopCount; i++)
{
    _frameWrite.Write(new ArraySegment<byte>(_writeData));
}
```
The attribute `OperationsPerInvoke` says how much to divide the total time by

cesarbs: I guess I misunderstood how the attribute works. Will change.

benaadams: It tells it how many times you loop; so you can measure with less function call overhead by adding a loop

cesarbs:
Made things simpler - doing 1-byte writes and setting `Content-Length` to a large enough value that it doesn't throw because of too many bytes written.

I'm having an issue though: the first benchmark is always bombed:

```
                   Method |        Mean |    StdErr |    StdDev |  Gen 0 | Allocated |
------------------------- |------------ |---------- |---------- |------- |---------- |
                    Write |          NA |        NA |        NA |    N/A |       N/A |
```

I'm seeing this error when it gets run:

```
// Validating benchmarks:
// **************************
// Benchmark: Writing.Write: DefaultJob
// *** Generate ***
// Result = Success
// BinariesDirectoryPath = D:\src\aspnet\KestrelHttpServer\test\Microsoft.AspNetCore.Server.Kestrel.Performance\BDN.Generated\binaries

// *** Build ***
C:\Program Files\dotnet\dotnet.exe compile-csc @D:\src\aspnet\KestrelHttpServer\test\Microsoft.AspNetCore.Server.Kestrel.Performance\obj\Release\netcoreapp1.1\dotnet-compile.rsp returned Exit Code 1
D:\src\aspnet\KestrelHttpServer\test\Microsoft.AspNetCore.Server.Kestrel.Performance\error CS2012: Cannot open 'D:\src\aspnet\KestrelHttpServer\test\Microsoft.AspNetCore.Server.Kestrel.Performance\bin\Release\netcoreapp1.1\Microsoft.AspNetCore.Server.Kestrel.Performance.dll' for writing -- 'The process cannot access the file 'D:\src\aspnet\KestrelHttpServer\test\Microsoft.AspNetCore.Server.Kestrel.Performance\bin\Release\netcoreapp1.1\Microsoft.AspNetCore.Server.Kestrel.Performance.dll' because it is being used by another process.'
// Result = Failure
```

Any ideias? I've even reset my machine to make sure no process was hanging on to that file, but it didn't help. I don't have VS open so that's not it either.

halter73:
Did you compare #1218 to dev using these benchmarks?

benaadams:
@cesarbs had that issue after killing it halfway through and re-running (maybe VS?)

Deleted the bin/Release directory and ran again at cmd and got these results

```
                   Method |        Mean |    StdDev |  Gen 0 | Allocated |
------------------------- |------------ |---------- |------- |---------- |
                    Write |  51.5520 ns | 0.7567 ns |      - |       0 B |
             WriteChunked |  52.2010 ns | 0.8147 ns |      - |       0 B |
               WriteAsync |  91.7909 ns | 0.4175 ns | 0.0142 |      63 B |
        WriteAsyncChunked |  89.3543 ns | 0.5601 ns | 0.0142 |      63 B |
        WriteAsyncAwaited | 134.9429 ns | 1.1241 ns | 0.0138 |      63 B |
 WriteAsyncAwaitedChunked | 133.7213 ns | 0.7650 ns | 0.0139 |      63 B |
               ProduceEnd |  47.2052 ns | 0.4014 ns | 0.0170 |      63 B |
        ProduceEndChunked |  47.2801 ns | 0.1889 ns | 0.0172 |      63 B |
```

cesarbs:
@benaadams No luck. What fixes it for me is adding a no-op benchmark that gets run first - that one takes the hit and the rest runs fine.

I'll try to repro in a separate app, might be a bug in BenchmarkDotNet.

cesarbs:
Initially I thought this was due to the switch to `netcoreapp1.1`, but it actually reproes even in `dev` with the existing `RequestParsing` benchmarks. And this is after doing a lot of cleanup:

- Deleted NuGet cache
- Cleaned git repo
- Restarted the machine
- Rebuilt in `dev`
- Ran the tests

This is crazy.

cesarbs:
Here's the issue:

`dotnet run -c Release` -> OK
`dotnet run -c release` -> Fails

cesarbs:
https://github.com/dotnet/BenchmarkDotNet/issues/322

cesarbs:
Posted measurements that prompted this change here: https://github.com/aspnet/KestrelHttpServer/pull/1218#issuecomment-266158979

benaadams:
> Here's the issue:

How oddly random 😀

PR #1257, opened 2016-12-13
Fix races in some EngineTests.

cesarbs:
Uncaught test failure while I was working on #1218.

In `OnStartingCallbacksAreCalledInLastInFirstOutOrder` and `OnCompletedCallbacksAreCalledInLastInFirstOutOrder`, we need to wait externally for all callbacks to be called.

Previously this worked fine because by the time we received the FIN from Kestrel, we could be sure those callbacks had been called, since the connection shut down gracefully. Now, however, the client FIN aborts the request, and depending on the timing and the order that things run, the test might see the connection being closed by the server before `OnStarting`/`OnCompleted` callbacks were called (because the abort runs concurrent to request processing).

@halter73 @natemcmaster

### DIFF
--- a/test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/ResponseTests.cs
+++ b/test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/ResponseTests.cs
@@ -590,14 +590,22 @@ namespace Microsoft.AspNetCore.Server.Kestrel.FunctionalTests
         [Fact]
         public async Task WhenAppWritesLessThanContentLengthErrorLogged()
         {
-            var testLogger = new TestApplicationErrorLogger();
-            var serviceContext = new TestServiceContext { Log = new TestKestrelTrace(testLogger) };
+            string errorMessage = null;
+            var logTcs = new TaskCompletionSource<object>();
+            var mockTrace = new Mock<IKestrelTrace>();
+            mockTrace
+                .Setup(trace => trace.ApplicationError(It.IsAny<string>(), It.IsAny<InvalidOperationException>()))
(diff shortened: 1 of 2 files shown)


cesarbs:
Fixed a similar race in `ResponseTests.WhenAppWritesLessThanContentLengthErrorLogged`.

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/ResponseTests.cs
```diff
@@ -590,14 +590,22 @@ public Task ResponseStatusCodeSetBeforeHttpContextDisposedRequestMalformedReadIg
         [Fact]
         public async Task WhenAppWritesLessThanContentLengthErrorLogged()
         {
-            var testLogger = new TestApplicationErrorLogger();
-            var serviceContext = new TestServiceContext { Log = new TestKestrelTrace(testLogger) };
+            string errorMessage = null;
```
muratg: var :D

muratg:
:shipit:

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/ResponseTests.cs
```diff
@@ -909,7 +919,7 @@ public Task ResponseStatusCodeSetBeforeHttpContextDisposedRequestMalformedReadIg
                         "POST / HTTP/1.1",
                         "Transfer-Encoding: chunked",
                         "",
-                        "wrong");
+                        "gg");
```
cesarbs: Unrelated but piggybacking this here. Test can fail if `Send` is still sending bytes but the server has already closed the connection because of the bad chunk length.

Issue #1112, opened 2016-09-17
'WriteReqPool'.) (Cannot access a disposed object.


benaadams:
During shutdown I received these errors (version `1.1.0-alpha1-22157`)

```
Application is shutting down...

Unhandled Exception: System.AggregateException: One or more errors occurred. (Cannot access a disposed object.
Object name: 'WriteReqPool'.) (Cannot access a disposed object.
Object name: 'WriteReqPool'.) (Cannot access a disposed object.
Object name: 'WriteReqPool'.) (Cannot access a disposed object.
Object name: 'WriteReqPool'.) (Cannot access a disposed object.
Object name: 'WriteReqPool'.) (Cannot access a disposed object.
Object name: 'WriteReqPool'.) (Cannot access a disposed object.
Object name: 'WriteReqPool'.) (Cannot access a disposed object.
Object name: 'WriteReqPool'.) (Cannot access a disposed object.
Object name: 'WriteReqPool'.) (Cannot access a disposed object.
Object name: 'WriteReqPool'.) (Cannot access a disposed object.
Object name: 'WriteReqPool'.) (Cannot access a disposed object.
Object name: 'WriteReqPool'.) (Cannot access a disposed object.
Object name: 'WriteReqPool'.) (Cannot access a disposed object.
Object name: 'WriteReqPool'.) (Cannot access a disposed object.
Object name: 'WriteReqPool'.) ---> System.ObjectDisposedException: Cannot access a disposed object.
Object name: 'WriteReqPool'.
   at Microsoft.AspNetCore.Server.Kestrel.Internal.Infrastructure.WriteReqPool.Allocate()
   at Microsoft.AspNetCore.Server.Kestrel.Internal.Http.SocketOutput.WriteContext.DoWriteIfNeeded()
   at Microsoft.AspNetCore.Server.Kestrel.Internal.Http.SocketOutput.WriteAllPending()
   at Microsoft.AspNetCore.Server.Kestrel.Internal.KestrelThread.DoPostWork()
   at Microsoft.AspNetCore.Server.Kestrel.Internal.KestrelThread.OnPost()
   at Microsoft.AspNetCore.Server.Kestrel.Internal.Networking.Libuv.NativeMethods.uv_run(UvLoopHandle handle, Int32 mode)
   at Microsoft.AspNetCore.Server.Kestrel.Internal.Networking.Libuv.run(UvLoopHandle handle, Int32 mode)
[...]

PR #1258, opened 2016-12-15
Abort connections not closed during shutdown

cesarbs:
Connections that don't stop during shutdown continue doing work and could be allocating memory blocks and write reqs, leading to exceptions being thrown during shutdown after pools are disposed.

With this change, Kestrel aborts any connections that were not gracefully stopped and waits for their sockets to be closed. This prevents the errors above.

Should take care of #1112.

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Infrastructure/KestrelThread.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Infrastructure/KestrelThread.cs
@@ -165,6 +165,11 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal
                 if (!await ConnectionManager.WalkConnectionsAndCloseAsync(_shutdownTimeout).ConfigureAwait(false))
                 {
                     _log.NotAllConnectionsClosedGracefully();
+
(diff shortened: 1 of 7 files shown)


cesarbs:
Apparently marking people as reviewers doesn't work the same as @ mentioning them.

@halter73 @pakrym @davidfowl @natemcmaster

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Infrastructure/KestrelThread.cs
```diff
@@ -160,6 +160,11 @@ public Task StartAsync()
                 if (!await ConnectionManager.WalkConnectionsAndCloseAsync(_shutdownTimeout).ConfigureAwait(false))
                 {
                     _log.NotAllConnectionsClosedGracefully();
+
+                    if (!await ConnectionManager.WalkConnectionsAndAbortAsync(_shutdownTimeout).ConfigureAwait(false))
```
halter73: I don't think I like repeating the `_shutdownTimeout` here. That could be configured to a very large value. This probably doesn't need to be configurable, since this isn't really a timeout meant for running application code.

I would probably make it 1 second.

cesarbs:
Updated.

halter73:
:shipit:

Issue #1260, opened 2016-12-15
Memory Leak in 1.1.0


meichtf:
Since updating to 1.1.0 (on Win10 64bit) my server keeps using memory at about ~0.5 Mb per minute. Is there a new setting that must be set or is there a memory leak in 1.1.0?

It's very easy to reproduce: 
1) Create a new "ASP.NET Core Web Application" in VS2015 (The wizard will create a 1.0.1 application). Start the application, the "Diagnostics Tools" in VS2015 will show a usage of about ~245.2Mb and will stay there.
2) Now update the app to netcoreapp1.1 (as a further test, leave Microsoft.AspNetCore.Server.Kestrel at version 1.0.0 and it will still not use more memory)
3) Use the new 1.1.0 of Microsoft.AspNetCore.Server.Kestrel and start the "Diagnostic Tool" again: It will start at ~245.2Mb again, but then the "Process Memory" keeps increasing

pakrym:
I wasn't able to reproduce it's locally, can you provide your `project.json` please?

pakrym:
Okay, I think I can see it, much slower on my machine and it looks like it leaks native memory because managed heap do not change in between memory snapshots. 

@cesarbs could it be something related to timers?

/cc @halter73

pakrym:
Yep, there is a leak and it is related to timers, changing `public const long HeartbeatMilliseconds = 1000;` to `10` makes it really obvious:


![image](https://cloud.githubusercontent.com/assets/1697911/21234675/773cf06c-c2a8-11e6-90a3-66c6ddd75e4f.png)

cesarbs:
@pakrym I'll look into it when I'm in the office.

agoretsky:
After upgrading all our app dependencies to 1.1.0 we also got memory leak problem on Linux (the process growing after start), but I was not able to reproduce it on Windows and unfortunately there is no tools to profile memory like dotMemory on Linux, so we rolled back to 1.0.1. I hope it's related and will be fixed

benaadams:
@agoretsky is this from cert use? (Releated bug and repo you raised in coreclr https://github.com/dotnet/coreclr/issues/8660); or unreleated issue?

agoretsky:
@benaadams, yes

meichtf:
I build the 1.1 server with the code changes from PR 1261 and the leak is gone.

pakrym:
@meichtf thank you for testing.

[...]

PR #1261, opened 2016-12-15
Prevent closure allocations in OnHeartbeat

pakrym:
Should prevent memory growing in https://github.com/aspnet/KestrelHttpServer/issues/1260

@halter73 @cesarbs

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Infrastructure/KestrelThread.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Infrastructure/KestrelThread.cs
@@ -23,7 +23,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal
 
         private static readonly Action<object, object> _postCallbackAdapter = (callback, state) => ((Action<object>)callback).Invoke(state);
         private static readonly Action<object, object> _postAsyncCallbackAdapter = (callback, state) => ((Action<object>)callback).Invoke(state);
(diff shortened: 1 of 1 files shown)


src/Microsoft.AspNetCore.Server.Kestrel/Internal/Infrastructure/KestrelThread.cs
```diff
@@ -23,6 +23,11 @@ public class KestrelThread
 
         private static readonly Action<object, object> _postCallbackAdapter = (callback, state) => ((Action<object>)callback).Invoke(state);
         private static readonly Action<object, object> _postAsyncCallbackAdapter = (callback, state) => ((Action<object>)callback).Invoke(state);
+        private static readonly Libuv.uv_walk_cb _heartbeatWalkCallback = (ptr, arg) =>
```
cesarbs: nit: call it `_heartbeatWalkCallbackAdapter` to be consistent with the other ones.

halter73:
`nit: call it _heartbeatWalkCallbackAdapter to be consistent with the other ones.` I think it show go back to `_heartbeatWalkCallback`. Unlike the other `Adapter`s, it's not calling a callback passed in as an argument.

cesarbs:
:shipit:

cesarbs:
@halter73 You're right, I overlooked that. @pakrym Sorry, revert the last commit :grin:

cesarbs:
Squash and :rocket:

PR #1262, opened 2016-12-15
Enable Server GC for Samples

mikeharder:
@halter73, @cesarbs


cesarbs:
What's the reason for the change?

mikeharder:
Ideally, ASP.NET Core would either default to Server GC, or warn customers if it's not enabled.  We currently have neither of these, so it's important we enable Server GC in every sample we publish to help customers know it needs to be enabled.

PR #1271, opened 2016-12-22
Implement Stream Flush+FlushAsync fully

benaadams:
Streams should pass through the flush and not assume the underlying Stream's Flush behaviour

SslStream's FlushAsync has been fixed for 1.2.0 (2.0?) https://github.com/dotnet/corefx/pull/14035


cesarbs:
Will merge this after we have https://github.com/dotnet/corefx/pull/14852 in our corefx deps. Just to be safe (don't want to introduce a lot of simultaneous changes around streams given recent issues).

benaadams:
Rebased

PR #1272, opened 2016-12-27
Fix NuGet downgrade warning

pranavkm:
Works around warning where a P2P reference is explicitly listed in the project. For e.g.

warn :  SampleApp (>= 1.1.0) -> Microsoft.AspNetCore.Server.Kestrel.Https (>= 1.2.0) -> Microsoft.AspNetCore.Server.Kestrel
warn :  SampleApp (>= 1.1.0) -> Microsoft.AspNetCore.Server.Kestrel (>= 1.2.0)

### DIFF
--- a/samples/SampleApp/project.json
+++ b/samples/SampleApp/project.json
@@ -1,7 +1,6 @@
 {
   "version": "1.1.0-*",
   "dependencies": {
-    "Microsoft.AspNetCore.Server.Kestrel": "1.2.0-*",
     "Microsoft.AspNetCore.Server.Kestrel.Https": "1.2.0-*",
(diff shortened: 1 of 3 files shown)


pranavkm:
> do we have a bug filed on Nuget for this?

Not sure. Would you like to follow up? 😄

emgarten:
> why does nuget see this as a downgrade

There was an issue in NuGet where it didn't handle ranges without a minimum correctly during the downgrade check. It comes from this project (which is correct): https://github.com/aspnet/KestrelHttpServer/blob/dev/src/Microsoft.AspNetCore.Server.Kestrel.Https/project.json#L23-L25

I believe it has been fixed in NuGet, the NuGet.Client repo use to have the same incorrect warnings.

PR #1275, opened 2016-12-30
[WIP][Tentative] Prevent hang on HTTPS async writes

cesarbs:
Possible workaround for `SslStream` issue after update to CoreFX 4.4 packages: https://github.com/dotnet/corefx/issues/14698#issuecomment-269725058

I think this is a good way to fix HTTPS, at least temporarily. Low overhead.

Reason why this works here: https://github.com/dotnet/corefx/issues/14698#issuecomment-269730422

cc @davidfowl @pakrym @benaadams


cesarbs:
Not needed after https://github.com/dotnet/corefx/commit/6031c507a0ea912ede85ca9fc3b8afc9a0fe70f0.

Issue #996, opened 2016-07-19
Create a direct way to configure endpoints on Kestrel


Tratcher:
Right now endpoints can only be configured via UseUrls. This limits the opportunity for validation and causes confusion about what's allowed/supported by Kestrel.

e.g. options.AddEndpoint(https, host, port); where scheme is "http" or "https", host is "localhost", "_", or an IPAddress, and port is valid. This avoids confusion around why registering "http://_.foo.com/" does not filter by host.

@DamianEdwards @davidfowl

muratg:
@davidfowl to schedule an api design meeting

DamianEdwards:
@shirhatti let's get this designed for 1.1.0 ASAP pls

halter73:
Are we willing to change the design/behavior of `UseUrls`? Will there be an analog for WebListener?

Tratcher:
WebListener already has `WebListenerOptions.ListenerSettings.UrlPrefixes.Add(UrlPrefix)`

shirhatti:
Based on our ( @davidfowl @Tratcher @halter73 @cesarbs )  design meeting yesterday, we made a couple of decisions. Some of these might be obvious but it's still worth explicitly calling out.

 - No change in the behavior of the `WebHostBuilder` extension method `.UseUrls()`. If you are using Kestrel as your server, we will not filter based on host name. The argument to `.UseUrls()` only determines if we bind to `IPAddress.Any` vs `IPAddress.Loopback`

- [ ] Add an extension method to `KestrelServerOptions` for binding to a TCP socket. There is no overload on `.Listen()` that allows you to configure an SSL cert without using an options lambda.
````
var host = new WebHostBuilder()
            .UseKestrel(options =>
            {
                // Easy mode (http only)
                options.Listen(IPAddress.Any, 80);

                // Verbose
                options.Listen(IPAddress.Any, 443, listenOptions => 
                {
                    // Enable https
                    listenOptions.UseHttps("server.pfx");
                });
            })
            .UseStartup<Startup>()
            .Build();

host.Run();
````

- [ ] Add an extension method to `KestrelServerOptions` for binding to a unix socket. There is no overload on `.ListenUnixSocket()` that allows you to configure an SSL cert without using an options lambda.
````
var host = new WebHostBuilder()
            .UseKestrel(options =>
            {
                // Easy mode
                options.ListenUnixSocket("/tmp/kestrel-test.sock");

                // Verbose
[...]

PR #1280, opened 2017-01-04
Create a direct way to configure endpoints on KestrelEngine

halter73:
- Replace endpoint configuration via .UseUrls() or --server.urls with Listen*
  methods on KestrelSerrverOptions.
- Replace IConnectionFilter with IConnectionAdapter which no longer exposes
  ServerAddress via a context.
- Simplify libuv Listener classes
- Support systemd socket activation
- Add docker-based test for systemd socket activation to be run on Travis

#996


src/Microsoft.AspNetCore.Server.Kestrel/KestrelServer.cs
```diff
@@ -113,84 +114,82 @@ public void Start<TContext>(IHttpApplication<TContext> application)
...
-                        {
-                            if ((ex.InnerException as UvException)?.StatusCode == Constants.EADDRINUSE)
-                            {
-                                throw new IOException($"Failed to bind to address {parsedAddress}: address already in use.", ex);
-                            }
-
-                            throw;
-                        }
+                        var joined = string.Join(", ", addresses);
+                        _logger.LogWarning($"Address(es) '{joined}' were overridden. Binding to endpoints defined in UseKestrel instead.");
```
davidfowl: Why would we do this instead of throwing? The only reason I can think of is if some component like IISIntegration set it up and then the user tried to override it by specifying the bindings directly?

src/Microsoft.AspNetCore.Server.Kestrel/KestrelServer.cs
```diff
@@ -113,84 +114,82 @@ public void Start<TContext>(IHttpApplication<TContext> application)
...
+                foreach (var endPoint in listenOptions)
+                {
+                    atLeastOneListener = true;
 
-                        try
+                    try
+                    {
+                        _disposables.Push(engine.CreateServer(endPoint));
+                    }
+                    catch (AggregateException ex)
```
davidfowl: Why did this change from the exception filter?

src/Microsoft.AspNetCore.Server.Kestrel/KestrelServer.cs
```diff
@@ -113,84 +114,82 @@ public void Start<TContext>(IHttpApplication<TContext> application)
...
-                            throw new InvalidOperationException("Dynamic port binding is not supported when binding to localhost. You must either bind to 127.0.0.1:0 or [::1]:0, or both.");
-                        }
 
-                        var ipv4Address = parsedAddress.WithHost("127.0.0.1");
-                        var exceptions = new List<Exception>();
+                    _serverAddresses.Addresses.Clear();
+                }
+                else
+                {
+                    // If no endpoints are configured directly using KestrelServerOptions, use those configured via --server.urls.
```
davidfowl: Just say via ServerAddresses, --server.urls doesn't *really* exist beyond hosting.

src/Microsoft.AspNetCore.Server.Kestrel/KestrelServerOptions.cs
```diff
@@ -70,6 +71,7 @@ public class KestrelServerOptions
         /// <remarks>
         /// Defaults to true.
         /// </remarks>
+        [Obsolete("This property is obsolete and will be removed in a future version. Use ListenOptions.NoDelay instead.")]
         public bool NoDelay { get; set; } = true;
```
davidfowl: You're already making a bunch of breaking changes, why obsolete this instead of removing it?

src/Microsoft.AspNetCore.Server.Kestrel/KestrelServerOptions.cs
```diff
@@ -115,5 +117,115 @@ private static int ProcessorThreadCount
...
+            ListenHandle(handle, _ => { });
+        }
+
+        /// <summary>
+        /// Open a socket file descriptor.
+        /// The callback configures endpoint-specific settings.
+        /// </summary>
+        public void ListenHandle(long handle, Action<ListenOptions> configure)
+        {
+            if (handle < 0)
```
davidfowl: ulong? :trollface:

src/Microsoft.AspNetCore.Server.Kestrel/ListenOptions.cs
```diff
@@ -0,0 +1,102 @@
...
+        /// </summary>
+        public KestrelServerOptions KestrelServerOptions { get; internal set; }
+
+        /// <summary>
+        /// Set to false to enable Nagle's algorithm for all connections.
+        /// </summary>
+        /// <remarks>
+        /// Defaults to true.
+        /// </remarks>
+        public bool NoDelay
```
davidfowl: We should consider splitting the configuration for accepted sockets vs the listen socket.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Listener.cs
```diff
@@ -56,11 +56,61 @@ protected Listener(ServiceContext serviceContext)
...
+                    }
+                    catch
+                    {
+                        pipe.Dispose();
+                        throw;
+                    }
+
+                    return pipe;
+                default:
+                    throw new InvalidOperationException();
```
davidfowl: NotSupportedException seems more appropriate

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Connection.cs
```diff
@@ -187,31 +157,65 @@ public void Tick(long timestamp)
...
+                    // Don't attempt to read input if connection has already closed.
+                    // This can happen if a client opens a connection and immediately closes it.
+                    _readInputTask = _socketClosedTcs.Task.Status == TaskStatus.WaitingForActivation
+                        ? _filteredStreamAdapter.ReadInputAsync()
+                        : TaskCache.CompletedTask;
+                }
 
-            _frame.PrepareRequest = _filterContext.PrepareRequest;
-            _frame.Start();
+                _frame.PrepareRequest = features =>
```
davidfowl: You could avoid the allocation by setting the list of adaptedConnections on the frame itself and having it do this logic in Frame. It's actually a bit cleaner than assigning a delegate in Connection.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/PipeListener.cs
```diff
@@ -1,61 +0,0 @@
-// Copyright (c) .NET Foundation. All rights reserved.
```
davidfowl: These deleted files make me so happy 👍

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Networking/SockAddr.cs
```diff
@@ -92,7 +92,7 @@ public unsafe IPEndPoint GetIPEndPoint()
                     *((long*)(b + 8)) = _field2;
                 }
 
-                return new IPEndPoint(new IPAddress(bytes), port);
+                return new IPEndPoint(new IPAddress(bytes, scopeid: _field3 & 0xFFFFFFFF), port);
```
davidfowl: This is comment worthy.

src/Microsoft.AspNetCore.Server.Kestrel/ListenOptions.cs
```diff
@@ -0,0 +1,102 @@
...
+        /// </summary>
+        public ListenType Type { get; }
+
+        // IPEndPoint is mutable so port 0 can be updated to the bound port.
+        public IPEndPoint IPEndPoint { get; internal set; }
+        public string SocketPath { get; }
+        public long FileDescriptor { get; }
+
+        /// <summary>
+        /// Enables and <see cref="IConnectionAdapter"/> to resolve and use services registered by the application during startup.
```
cesarbs: This doc comment doesn't look right.

src/Microsoft.AspNetCore.Server.Kestrel/ListenOptions.cs
```diff
@@ -0,0 +1,102 @@
...
+        /// <summary>
+        /// Gets the <see cref="List{IConnectionAdapter}"/> that allows each connection <see cref="System.IO.Stream"/>
+        /// to be intercepted and transformed.
+        /// Configured by the <c>UseHttps()</c> and <see cref="Hosting.ListenOptionsConnectionLoggingExtensions.UseConnectionLogging(ListenOptions)"/>
+        /// extension methods.
+        /// </summary>
+        /// <remarks>
+        /// Defaults to empty.
+        /// </remarks>
+        public List<IConnectionAdapter> ConnectionAdapters { get; } = new List<IConnectionAdapter>();
```
cesarbs: Different adapters per endpoint?

src/Microsoft.AspNetCore.Server.Kestrel/Systemd/KesterlServerOptionsSystemdExtensions.cs
```diff
@@ -0,0 +1,42 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Diagnostics;
+using System.Globalization;
+using Microsoft.AspNetCore.Server.Kestrel;
+
+namespace Microsoft.AspNetCore.Hosting
+{
+    public static class KesterlServerOptionsSystemdExtensions
```
cesarbs: We should offer this in a separate package.

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/SystemdActivation/docker.sh
```diff
@@ -0,0 +1,19 @@
+#!/usr/bin/env bash
```
cesarbs: Add `set -e` at the top to fail the script if any command fails.

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/SystemdActivation/docker.sh
```diff
@@ -0,0 +1,19 @@
...
+
+# Ensure that dotnet is added to the PATH
+scriptDir=$(dirname "${BASH_SOURCE[0]}")
+repoDir=$(cd $scriptDir/../../.. && pwd)
```
cesarbs: You can replace `cd $scriptDir/../../.. && pwd` with `readlink -e $scriptDir/../../..`

test/Microsoft.AspNetCore.Server.KestrelTests/KestrelServerOptionsTests.cs
```diff
@@ -37,6 +38,43 @@ public void MaxRequestBufferSizeSetsLimitsProperty()
...
+        {
+            Assert.NotNull(typeof(KestrelServerOptions)
+                .GetProperty(nameof(KestrelServerOptions.NoDelay))
+                .GetCustomAttributes(false)
+                .OfType<ObsoleteAttribute>()
+                .SingleOrDefault());
+        }
+
+        [Fact]
+        public void NoDelayPropogatesToListenOptions()
```
cesarbs: nit: `Propagates`

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/SystemdActivation/docker.sh
halter73: I actually want to change directories.

src/Microsoft.AspNetCore.Server.Kestrel.Https/HttpsConnectionAdapter.cs
```diff
@@ -0,0 +1,160 @@
...
+            {
+                throw new ArgumentNullException(nameof(options));
+            }
+            if (options.ServerCertificate == null)
+            {
+                throw new ArgumentException("The server certificate parameter is required.");
+            }
+
+            _options = options;
+            _logger = loggerFactory?.CreateLogger(nameof(HttpsConnectionAdapter));
```
halter73: When the HttpsConnectionAdapter is initialized directly using the other ctor.

src/Microsoft.AspNetCore.Server.Kestrel/Adapter/ListenOptionsConnectionLoggingExtensions.cs
```diff
@@ -0,0 +1,38 @@
...
+{
+    public static class ListenOptionsConnectionLoggingExtensions
+    {
+        /// <summary>
+        /// Emits verbose logs for bytes read from and written to the connection.
+        /// </summary>
+        /// <returns>
+        /// The <see cref="ListenOptions"/>.
+        /// </returns>
+        public static ListenOptions UseConnectionLogging(this ListenOptions listenOptions)
```
halter73: ConnectionAdapters are configured per endpoint. This is important for HTTPS.

If you want to configure all endpoints in the same way (e.g. to enable connection logging for everything), I would recommend creating a static method with a ListenOptions parameter to share the configuration code.

src/Microsoft.AspNetCore.Server.Kestrel/ListenOptions.cs
halter73: Even though we call uv_tcp_nodelay on the listen socket, I don't think it has any effect there (it *might* get inherited but that's pretty moot). I think the only important thing is configuring nodelay for the accepted sockets.

I certainly wouldn't split the configuration. We could remove the call uv_tcp_nodelay for the listen socket, but I would do that as part of another change and make absolutely sure that it doesn't affect perf.

src/Microsoft.AspNetCore.Server.Kestrel/ListenOptions.cs
halter73: @cesarbs is right. Different adapters per endpoint is nice for HTTPS.

src/Microsoft.AspNetCore.Server.Kestrel/Systemd/KesterlServerOptionsSystemdExtensions.cs
halter73: It seems pretty pointless to move a single class without any real extra dependencies to another assembly.

src/Microsoft.AspNetCore.Server.Kestrel/KestrelServer.cs
halter73: Not sure what you mean. This is the [same logic](https://github.com/aspnet/KestrelHttpServer/pull/1280/files#diff-2a47c9b99ecab7d9ea77789228f7e771L127) we had before for when anything other than "localhost" was specified.

We don't swallow any exceptions like we did when "localhost" was specified since there is no desire to gracefully handle binding only to the IPv4 loopback adapter and not the IPv6 one or vice-versa.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Listener.cs
```diff
@@ -77,7 +127,22 @@ private static void ConnectionCallback(UvStreamHandle stream, int status, Except
...
+
+            try
+            {
+                acceptSocket = CreateAcceptSocket();
+                listenSocket.Accept(acceptSocket);
+                DispatchConnection(acceptSocket);
+            }
+            catch (UvException ex)
+            {
+                Log.LogError(0, ex, "Listener.OnConnection");
```
halter73: Wouldn't this need to be the following then?

    Log.LogError(0, ex, $"{nameof(Listener)}.{nameof(Listener.OnConnection)}");

Not sure it's worth it since we never really expect the above to throw. If you search the Kestrel code base for "Log.LogError" you'll see a lot of places that need to change.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Networking/UvTcpHandle.cs
```diff
@@ -24,20 +25,23 @@ public void Init(UvLoopHandle loop, Action<Action<IntPtr>, IntPtr> queueCloseHan
             _uv.tcp_init(loop, this);
         }
 
-        public void Bind(ServerAddress address)
+        public void Open(long fileDescriptor)
```
halter73: long/ulong is specced to be exactly 64-bit. I don't think the would be a problem down casting that to a 32-bit IntPtr on 32-bit platforms. Longs are generally a little easier to work with in terms of literals, parsing from strings, etc...

If someone uses a value that was out of range on a 32-bit platform, it wasn't a valid handle to begin with.

src/Microsoft.AspNetCore.Server.Kestrel/ListenOptions.cs
halter73: Changed "and" to "an". Was there something else wrong?

samples/SampleApp/Startup.cs
```diff
@@ -33,24 +34,33 @@ public void Configure(IApplicationBuilder app, ILoggerFactory loggerFactory)
 
         public static void Main(string[] args)
         {
-            var host = new WebHostBuilder()
-                .UseKestrel(options =>
+            var hostBuilder = new WebHostBuilder().UseKestrel(options =>
+            {
+                options.Listen(IPAddress.Loopback, 5000, listenOptions =>
```
Tratcher: For a realistic sample the IP and port need to come from command line / env / config. We need to make that possible in one or two lines. Even if it means having a Listen overload that takes a string for `IP:port` that understands IPv4, IPv6, and *.
@shirhatti 
`options.Listen(config["IPAndPort"], listenOptions => ...`

Note IPEndpoint does not implement Parse so you can't just punt this over there.

Tratcher: Note that the VS 2017 self-host profile now sets the ASPNETCORE_URLS environment variable by default so apps are unlikely to start from VS on localhost:5000 anymore. /cc: @BillHiebert

src/Microsoft.AspNetCore.Server.Kestrel/KestrelServer.cs
```diff
@@ -113,84 +114,82 @@ public void Start<TContext>(IHttpApplication<TContext> application)
...
-                {
-                    var parsedAddress = ServerAddress.FromUrl(address);
-                    atLeastOneListener = true;
+                var listenOptions = Options.ListenOptions;
 
-                    if (!parsedAddress.Host.Equals("localhost", StringComparison.OrdinalIgnoreCase))
+                if (listenOptions.Any())
+                {
+                    var addresses = _serverAddresses.Addresses;
+                    if (addresses.SingleOrDefault() != "http://localhost:5000")
```
Tratcher: weird guesswork that makes assumptions about the intent of the caller. I'd rather reconsider the Hosting default logic. WebListener is going to have a similar issue when it's configured to attach to an existing queue. @davidfowl

src/Microsoft.AspNetCore.Server.Kestrel/Systemd/KesterlServerOptionsSystemdExtensions.cs
```diff
@@ -0,0 +1,42 @@
...
+        /// </summary>
+        /// <returns>
+        /// The <see cref="KestrelServerOptions"/>.
+        /// </returns>
+        public static KestrelServerOptions UseSystemd(this KestrelServerOptions options, Action<ListenOptions> configure)
+        {
+            if (string.Equals(Process.GetCurrentProcess().Id.ToString(CultureInfo.InvariantCulture), Environment.GetEnvironmentVariable("LISTEN_PID"), StringComparison.Ordinal))
+            {
+                // SD_LISTEN_FDS_START = 3
+                options.ListenHandle(3, configure);
```
Tratcher: define constants

src/Microsoft.AspNetCore.Server.Kestrel/Systemd/KesterlServerOptionsSystemdExtensions.cs
glennc: @davidfowl You mean we should talk about how we package and ship the systemd feature?

davidfowl: Yes, we should decide if we want it baked into the Kestrel package. Even https is a separate package.

glennc: Clearly since I wanted the feature it should be baked into Kestrel... ;). I will find you both to talk about it today.

Tratcher: It doesn't add any dependencies or much implementation, correct?

src/Microsoft.AspNetCore.Server.Kestrel/KestrelServer.cs
halter73: AFIAK, the idea is to remove the IServerAddresses feature entirely prior to 2.0, so this heuristic is only temporary.

src/Microsoft.AspNetCore.Server.Kestrel/Systemd/KesterlServerOptionsSystemdExtensions.cs
halter73: Technically the "System.Diagnostics.Process" dependency is added for netstandard, but that's in the shared runtime. Otherwise it's just one small static class (KesterlServerOptionsSystemdExtensions.cs) with two methods.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Networking/SockAddr.cs
halter73: I added to the big comment above.

src/Microsoft.AspNetCore.Server.Kestrel/KestrelServer.cs
Tratcher: Remove IServerAddresses? What's the rest of that plan?

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/SystemdActivation/docker.sh
halter73: Done.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Listener.cs
cesarbs: I tend to prefer `nameof` for things like this because it's easy to miss that if there's a rename.

PR #1285, opened 2017-01-05
Use status codes defined in HttpAbstractions

JunTaoLuo:
Reaction to https://github.com/aspnet/HttpAbstractions/pull/748

cc @cesarbs

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
@@ -141,7 +141,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Http
             set
             {
                 // GetKnownVersion returns versions which ReferenceEquals interned string
-                // As most common path, check for this only in fast-path and inline 
+                // As most common path, check for this only in fast-path and inline
                 if (ReferenceEquals(value, "HTTP/1.1"))
                 {
                     _httpVersion = Http.HttpVersion.Http11;
@@ -347,7 +347,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Http
             Path = null;
             QueryString = null;
             _httpVersion = Http.HttpVersion.Unset;
-            StatusCode = 200;
+            StatusCode = StatusCodes.Status200OK;
             ReasonPhrase = null;
 
             RemoteIpAddress = RemoteEndPoint?.Address;
@@ -816,7 +816,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Http
                 else
                 {
                     // 500 Internal Server Error
-                    SetErrorResponseHeaders(statusCode: 500);
+                    SetErrorResponseHeaders(statusCode: StatusCodes.Status500InternalServerError);
                 }
             }
 
@@ -909,7 +909,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Http
             {
                 if (!hasTransferEncoding && !responseHeaders.HasContentLength)
                 {
-                    if (appCompleted && StatusCode != 101)
+                    if (appCompleted && StatusCode != StatusCodes.Status101SwitchingProtocols)
                     {
                         // Since the app has completed and we are only now generating
                         // the headers we can safely set the Content-Length to 0.
@@ -924,7 +924,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Http
                         //
                         // A server MUST NOT send a response containing Transfer-Encoding unless the corresponding
                         // request indicates HTTP/1.1 (or later).
-                        if (_httpVersion == Http.HttpVersion.Http11 && StatusCode != 101)
+                        if (_httpVersion == Http.HttpVersion.Http11 && StatusCode != StatusCodes.Status101SwitchingProtocols)
                         {
                             _autoChunk = true;
                             responseHeaders.SetRawTransferEncoding("chunked", _bytesTransferEncodingChunked);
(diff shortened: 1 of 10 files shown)


cesarbs:
It's harmless but we use hardcoded status codes all over... Either let's leave it as it is, or change other places to use those consts.

Given numeric status codes are nothing cryptic (i.e. people are familiar with them), and not bound to change, I don't really see the need to do this other than to avoid typos (but tests should catch those).

We do have dupe code for reason phrases - https://github.com/aspnet/KestrelHttpServer/blob/49ff98f8cbcf511c343a7076ecdaece2b5306462/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/ReasonPhrases.cs. Consolidating that would be a more interesting change.

cesarbs:
@Tratcher We could pass in the strings from HttpAbstractions to the code that pre-encodes the reason phrases, so we don't have duplicate strings in the two repos. Would have to make the dict in HttpAbstractions public though.

Eilon:
I do like this change. I also don't think it needs to be all-or-none. We can gradually make our codebase better every day! 😄

JunTaoLuo:
@cesarbs @Tratcher Before I make the suggested changes, let's work out what to do regarding some of the status codes.

1. WebDAV status codes. @Tratcher suggested that we ignore these status codes but we are inconsistent here. For example, we don't have 102 in HttpAbstractions but we have 423. We should keep the status codes consistent between Kestrel and HttpAbstractions so maybe we need to add it?
2. 226. This isn't present in HttpAbstractions but is present in Kestrel. It's a status code unrelated to WebDAV and is used to send responses that involve instance-manipulation (delta encoding of responses). 
3. RFC2616 vs RFC7231 reason phrases. For example 413 used to be "Request Entity Too Large" but it is now "Payload Too Large" I'm leaning towards using the new ones since that's what we plan to have in HttpAbstractions.
4. Unofficial status code 419. This isn't defined in the RFC and is missing in Kestrel reason phrases. It's present in HttpAbstractions but I suppose if we need to use it in Kestrel, just specify the status code and reason phrases explicitly. This isn't a problem but we should probably be aware that we have unofficial status codes.
5. Other status codes. 421, 428, 429, 511. We should probably add these to Kestrel as well.

Tratcher:
A) Avoid removing any existing definitions where possible. Limit compiler breaking changes.
B) Adding everything in the IANA list should be fine. It'll be faster than debating the merits of individual codes.
C) Favor the latest RFC
D) 419 has an official RFC, it just got dropped from the IANA list. we can keep it.

JunTaoLuo:
:up::date:

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
```diff
@@ -1407,9 +1407,9 @@ public bool TakeMessageHeaders(SocketInput input, FrameRequestHeaders requestHea
         public bool StatusCanHaveBody(int statusCode)
         {
             // List of status codes taken from Microsoft.Net.Http.Server.Response
-            return statusCode != 204 &&
-                   statusCode != 205 &&
-                   statusCode != 304;
+            return statusCode != StatusCodes.Status204NoContent &&
+                   statusCode != StatusCodes.Status205ResetContent &&
+                   statusCode != StatusCodes.Status304NotModified;
```
Tratcher: I like this part.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/ReasonPhrases.cs
```diff
@@ -3,183 +3,217 @@
...
+        private static readonly byte[] _bytesStatus504 = CreateStatusBytes(StatusCodes.Status504GatewayTimeout);
+        private static readonly byte[] _bytesStatus505 = CreateStatusBytes(StatusCodes.Status505HttpVersionNotsupported);
+        private static readonly byte[] _bytesStatus506 = CreateStatusBytes(StatusCodes.Status506VariantAlsoNegotiates);
+        private static readonly byte[] _bytesStatus507 = CreateStatusBytes(StatusCodes.Status507InsufficientStorage);
+        private static readonly byte[] _bytesStatus508 = CreateStatusBytes(StatusCodes.Status508LoopDetected);
+        private static readonly byte[] _bytesStatus510 = CreateStatusBytes(StatusCodes.Status510NotExtended);
+        private static readonly byte[] _bytesStatus511 = CreateStatusBytes(StatusCodes.Status511NetworkAuthenticationRequired);
+
+        private static byte[] CreateStatusBytes(int statusCode) =>
+            Encoding.ASCII.GetBytes(statusCode.ToString(CultureInfo.InvariantCulture) + " " + WebUtilities.ReasonPhrases.GetReasonPhrase(statusCode));
```
Tratcher: You'd want at least an assert here that GetReasonPhrase returned a non-empty value in case they get out of sync. Or tests for each of these.

JunTaoLuo: Yea, I'm also considering other ways to keep them in sync since we might forget to add a status code here. Potentially we'll use reflection to scan for all the status codes defined and generate code.

halter73: This might be a good candidate for Microsoft.AspNetCore.Server.Kestrel.GeneratedCode.

JunTaoLuo: That still requires us to maintain a list of all the status codes somewhere since the codes are defined as consts. It would be easier if StatusCodes were an enum instead and we can use Enum.GetValues(). That'd be a rather dramatic change though.

Issue #1286, opened 2017-01-06
String.Equals(..., OrdinalIgnoreCase) is slow


benaadams:
Have been unsuccessful at making it fast so far 😢 

See: https://github.com/dotnet/coreclr/pull/8579

benaadams:
Removed in https://github.com/aspnet/KestrelHttpServer/pull/1291

Yves57:
I had an idea that seems to improve performances by limiting calls to `string.Equals()`.

I have changed `FrameHeaders.SetValueFast()` code like that:

Old: `if ("Pragma".Equals(key, StringComparison.OrdinalIgnoreCase))`
New: `if ((key[0] == 'P' || key[0] == 'p') && "Pragma".Equals(key, StringComparison.OrdinalIgnoreCase))`

And of course I choose an non-ambiguous char index to minimize the calls to `string.Equals()`.

On my old computer, a test like that

```
var chrono = new Stopwatch();
chrono.Start();

for (int i = 0; i < 1000000; i++)
{
	IHeaderDictionary frameHeaders = new FrameRequestHeaders();
	frameHeaders["Accept"] = new StringValues("text/html, application/xhtml+xml, image/jxr, */*");
	frameHeaders["Accept-Language"] = new StringValues("fr-FR,fr;q=0.5");
	frameHeaders["User-Agent"] = new StringValues("Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/51.0.2704.79 Safari/537.36 Edge/14.14393");
	frameHeaders["Accept-Encoding"] = new StringValues("gzip, deflate");
	frameHeaders["DNT"] = new StringValues("1");
	frameHeaders["Host"] = new StringValues("www.google.fr");
	frameHeaders["Connection"] = new StringValues("Keep-Alive");
}

chrono.Stop();
Console.WriteLine($"Total = {chrono.ElapsedMilliseconds}");
```

goes from ~560ms down to ~470ms.

Any thoughts?

benaadams:
Will never pass the `object.Equality` test that String.Equals does as the second string is always constructed, so has no advantage to doing the test; so the first char test is probably as close as can get replacement.

You could could bitwise or and change to single lower case compare `key[0] | 0x20 == 'p' ` rather than a double test?

PR #1287, opened 2017-01-06
Avoid String.Equals(..., OrdinalIgnoreCase)

benaadams:
Use alternative path than `String.Equals(..., OrdinalIgnoreCase)` see https://github.com/dotnet/coreclr/pull/8579

Resolves https://github.com/aspnet/KestrelHttpServer/issues/1286


benaadams:
Can do better...

PR #1291, opened 2017-01-07
[RFC] Simplify & Consolidate Headers

benaadams:
Changes the long hard coded numeric const functions to array lookups and simpler loops; drops long `string.Equals(.., IgnoreCase)` function, changing to data lookup and matching in loops (for both bytes in and chars out)

Then some jiggling around to get the perf back from loosing the unrolling and inline data.

Also tried to make the generated file nicely formatted and more legible; other than the spew of array initalization at the start... 

`Microsoft.AspNetCore.Server.Kestrel.dll` down by 26Kb

Request for comments /cc @halter73 @cesarbs @davidfowl


benaadams:
Looks like is a fraction faster for plaintext; and then better for more complex input and output headers

benaadams:
Closing, thought of something better, but will revisit post pipelines

Issue #1295, opened 2017-01-09
Not all request rejections are logged


cesarbs:
`BadHttpRequestException`s caught in https://github.com/aspnet/KestrelHttpServer/blob/e55c62444bdd34274173b011720c03f23c4db2f5/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/FrameOfT.cs#L196 are not being logged except when logged before being thrown (such as when calling `Frame.RejectRequest`), even though the comment in the catch block indicates that should be the case.

This is the case for requests containing invalid characters in headers (see https://github.com/aspnet/Mvc/issues/5662), [where the `BadHttpRequestException` is thrown from `MemoryPoolIterator.GetAsciiString`](https://github.com/aspnet/KestrelHttpServer/blob/2eba4017c1aa2f969e5a081f928c9c6a9df95327/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Infrastructure/MemoryPoolIterator.cs#L1085). This should be refactored BTW.

cesarbs:
@muratg Please triage.

PR #1277, opened 2016-12-30
[WIP] Pipeline migration

pakrym:

PR #1297, opened 2017-01-09
Log all rejected requests (#1295)

cesarbs:
#1295 is caused by there being places in the code where `BadHttpRequestException`s are thrown but not logged.

This PR changes the places where `BadHttpRequestException` is thrown (other than `Frame.RejectRequest()`) to throw a new exception type (`DecodingException`). This fixes layering since those places should not be throwing `BadHttpRequestException` (gross to do so in a place like `MemoryPoolIterator.GetAsciiString()`, for example). The new exception type is caught where required and `Frame.RejectRequest()` is called, effectively converting the lower-level exception into a `BadHttpRequestException`.

Additionally, I've done significant refactoring in `BadHttpRequestTests`.

I'll send another PR shortly where I fixed the same issue but in a different way. Still think this is the right one though.

Need to consider switching the place where we log the bad request from `Frame.RejectRequest()` to wherever it's caught (two places in `Frame<T>`). That would avoid having new code that throws it not via `Frame.RejectRequest()` getting bad requests not logged. Appreciate thoughts on this.

@halter73 @natemcmaster @Tratcher @pakrym


src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
```diff
@@ -1180,6 +1180,13 @@ public RequestLineStatus TakeStartLine(SocketInput input)
 
                 return RequestLineStatus.Done;
             }
+            catch (DecodingException)
+            {
+                RejectRequest(RequestRejectionReason.NonAsciiOrNullCharactersInRequestLine,
+                    Log.IsEnabled(LogLevel.Information) ? start.GetAsciiStringEscaped(end, MaxInvalidRequestLineChars) : string.Empty);
+                // won't run but compiler complains
```
natemcmaster: Would it work if we changed the return type of 'RejectRequest*' methods to Exception and instead did `throw RejectRequest(...` ?

cesarbs: Need to be careful to `throw` every time. I did consider this a while back. I prefer throwing from inside the method.

natemcmaster: Fine either way, but for readability I actually prefer the explicit throw. The first time I read this code, it was not immediately obvious that 'RejectRequest' and its overloads terminate execution.

halter73:
Nice catch and nice change. I want to get this in soon after #1277 is merged.

cesarbs:
@davidfowl Keeping this open to merge after pipelines. Will probably require some re-work.

davidfowl:
@cesarbs are you going to update this change and apply it to both branches?

cesarbs:
Holding the change against `feature/dev-si` for now. This was done in `dev` in ac60f13312b0776cbffe333e72ce521211845b89 and 11c7eb5665b294395db4227dd6f163e7e0fcbed5.

PR #1298, opened 2017-01-10
[Alt] Log all rejected requests (#1295)

cesarbs:
Similar to https://github.com/aspnet/KestrelHttpServer/pull/1297 but changes some APIs to `Try`* and throws when those return `false`.

I think the other PR is better but wanted to put this out here just in case.

Addresses #1295.

@halter73 @natemcmaster @Tratcher @pakrym


natemcmaster:
> I think the other PR is better

+1

benaadams:
Benchmarks between the two? 😉

PR #1300, opened 2017-01-10
Reduce hashtag ugliness

benaadams:
use #if DEBUG on readonly static rather than in each function

Diff mostly whitespace [https://github.com/aspnet/KestrelHttpServer/pull/1300/files?w=1](https://github.com/aspnet/KestrelHttpServer/pull/1300/files?w=1)


cesarbs:
Nice 👍

Waiting for AppVeyor and Travis before merging.

cesarbs:
I think this won't be relevant anymore after #1277. cc @pakrym

Issue #1289, opened 2017-01-06
InvalidOperationException: Response Content-Length mismatch when enabling BrowserLink


Tratcher:
_From @PinpointTownes on January 5, 2017 19:7_

_Moved from https://github.com/aspnet-contrib/AspNet.Security.OpenIdConnect.Server/issues/365._

___________________

Hello

I have managed to get the Authorization Code Flow to work. Everything works fine if the user 'Authorizes'. However, if the user denies the authorize the request (i.e. Forbid(OpenIdConnectServerDefaults.AuthenticationScheme) executes in the controller) I get the following exception:

```
System.InvalidOperationException: Response Content-Length mismatch: too many bytes written (1024 of 806).
   at Microsoft.AspNetCore.Server.Kestrel.Internal.Http.Frame.VerifyAndUpdateWrite(Int32 count)
   at Microsoft.AspNetCore.Server.Kestrel.Internal.Http.Frame.<WriteAsyncAwaited>d__183.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at Microsoft.VisualStudio.Web.BrowserLink.Runtime.ScriptInjectionFilterStream.<>c__DisplayClass40_0.<<CreateResponseHandler>b__0>d.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at Microsoft.VisualStudio.Web.BrowserLink.Runtime.SocketReader.<ReadBytesIntoResponseHandler>d__9.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at Microsoft.VisualStudio.Web.BrowserLink.Runtime.HttpSocketAdapter.ResponseReader.<ReadBytesIntoResponse>d__22.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
[...]

PR #1302, opened 2017-01-10
Call OnStarting before verifying response length (#1289)

cesarbs:
#1289

@Tratcher @halter73 @natemcmaster @pakrym

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
@@ -695,7 +695,13 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Http
                 responseHeaders.HeaderContentLengthValue.HasValue &&
                 _responseBytesWritten < responseHeaders.HeaderContentLengthValue.Value)
             {
-                _keepAlive = false;
+                // We need to close the connection if any bytes were written since the client
+                // cannot be certain of how many bytes it will receive.
+                if (_responseBytesWritten > 0)
+                {
+                    _keepAlive = false;
+                }
+
                 ReportApplicationError(new InvalidOperationException(
                     $"Response Content-Length mismatch: too few bytes written ({_responseBytesWritten} of {responseHeaders.HeaderContentLengthValue.Value})."));
             }
--- a/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
@@ -533,23 +533,29 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Http
 
         public void Flush()
         {
-            ProduceStartAndFireOnStarting().GetAwaiter().GetResult();
+            InitializeResponse(0).GetAwaiter().GetResult();
             Output.Flush();
         }
 
         public async Task FlushAsync(CancellationToken cancellationToken)
         {
-            await ProduceStartAndFireOnStarting();
+            await InitializeResponse(0);
             await Output.FlushAsync(cancellationToken);
         }
 
         public void Write(ArraySegment<byte> data)
         {
-            // For the first write, ensure headers are flushed if Write(Chunked)isn't called.
+            // For the first write, ensure headers are flushed if Write(Chunked) isn't called.
             var firstWrite = !HasResponseStarted;
 
-            VerifyAndUpdateWrite(data.Count);
-            ProduceStartAndFireOnStarting().GetAwaiter().GetResult();
+            if (firstWrite)
+            {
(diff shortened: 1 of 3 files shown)


test/Microsoft.AspNetCore.Server.KestrelTests/EngineTests.cs
```diff
@@ -878,11 +878,7 @@ public void ConnectionCanReadAndWrite(ListenOptions listenOptions)
                     throw onStartingException;
                 }, null);
 
-                response.Headers["Content-Length"] = new[] { "11" };
-
-                var writeException = await Assert.ThrowsAsync<ObjectDisposedException>(async () =>
-                    await response.Body.WriteAsync(Encoding.ASCII.GetBytes("Hello World"), 0, 11));
-
+                var writeException = await Assert.ThrowsAsync<ObjectDisposedException>(async () => await response.Body.FlushAsync());
```
cesarbs: Had to change so the check at https://github.com/aspnet/KestrelHttpServer/blob/e55c62444bdd34274173b011720c03f23c4db2f5/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/FrameOfT.cs#L110 wouldn't close the connection (because now the write doesn't go through before `OnStarting` as it did before, that method sees 0 bytes written and that's a mismatch against the old test which wrote 11 bytes).

cesarbs:
🔔

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
```diff
@@ -533,23 +533,29 @@ public void OnCompleted(Func<object, Task> callback, object state)
...
 
-            VerifyAndUpdateWrite(data.Count);
-            ProduceStartAndFireOnStarting().GetAwaiter().GetResult();
+            if (firstWrite)
+            {
+                InitializeResponse(data.Count).GetAwaiter().GetResult();
+            }
+            else
+            {
+                VerifyAndUpdateWrite(data.Count);
```
halter73: Can we change InitializeResponse to not call VerifyAndUpdateWrite, and just call VerifyAndUpdateWrite afterwards?

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/ResponseTests.cs
```diff
@@ -554,15 +554,52 @@ public Task ResponseStatusCodeSetBeforeHttpContextDisposedRequestMalformedReadIg
...
             var testLogger = new TestApplicationErrorLogger();
             var serviceContext = new TestServiceContext { Log = new TestKestrelTrace(testLogger) };
 
-            using (var server = new TestServer(async httpContext =>
+            using (var server = new TestServer(httpContext =>
             {
+                var response = Encoding.ASCII.GetBytes("hello, world");
                 httpContext.Response.ContentLength = 5;
-                await httpContext.Response.WriteAsync("hello, world");
+                httpContext.Response.Body.Write(response, 0, response.Length);
```
halter73: Why no longer async?

test/Microsoft.AspNetCore.Server.KestrelTests/EngineTests.cs
halter73: This seems bad. So you don't get a 500 anymore if you set a Content-Length.

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/ResponseTests.cs
cesarbs: The `WriteAsync` version of this test is just below. Need to verify both calls since they don't share code.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
cesarbs: We need the check in `InitializeResponse` so we can return a 500 if the first write exceeds Content-Length. Without the check there a 200 response still started, despite the code failing later.

test/Microsoft.AspNetCore.Server.KestrelTests/EngineTests.cs
cesarbs: You do. The point here is to not only test that you get a 500 when an `OnStarting` callback throws, but also that the connection is not closed in that situation. Because the test code captures the exception from `WriteAsync` and does not rethrow it, we end up verifying the amount of bytes written after the app runs (see link above), which also leads to a 500 response but this one also closes the connection (because too few bytes were written and that will confuse the client).

Switching this to `FlushAsync()` makes the test simpler and avoids that situation. An alternative would be to keep the `WriteAsync` call and rethrow the exception, but this will result in more errors being logged and will require changing the assertion for that.

cesarbs:
Updated.

halter73:
:shipit:

PR #1307, opened 2017-01-13
Don't overwhelm server with data in KeepAliveTimeoutTests.ConnectionNotTimedOutWhileRequestBeingSent()

cesarbs:
Terrible on slower machines.

@halter73


muratg:
What's the impact on the total duration of the test run?

cesarbs:
@muratg Varies a lot but on my crappy Mac I saw it take up to 3 times longer. This only affects slow machines like mine and what they seem to use on Travis. Our CI machines should be fine even without this.

halter73:
:shipit:

PR #1311, opened 2017-01-17
Better handle OnHeartbeat in 32-bit environments

halter73:
- Prevent potential overflow when casting a long to an IntPtr


cesarbs:
Commit hash?

halter73:
c4762cdfd5190de091d2a8b4624181a677af448e

PR #1312, opened 2017-01-18
Warn instead of throw when ignoring IServerAddressesFeature

halter73:
- Throwing could be too much when IServerAddressesFeature URLs come from VS

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Kestrel/KestrelServer.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/KestrelServer.cs
@@ -112,22 +112,24 @@ namespace Microsoft.AspNetCore.Server.Kestrel
                 }
 
                 engine.Start(threadCount);
-                var atLeastOneListener = false;
 
                 var listenOptions = Options.ListenOptions;
+                var hasListenOptions = listenOptions.Any();
+                var hasServerAddresses = _serverAddresses.Addresses.Any();
 
-                if (listenOptions.Any())
+                if (hasListenOptions && hasServerAddresses)
                 {
-                    var addresses = _serverAddresses.Addresses;
-                    if (addresses.SingleOrDefault() != "http://localhost:5000")
-                    {
-                        var joined = string.Join(", ", addresses);
-                        throw new NotSupportedException($"Specifying address(es) '{joined}' is incompatible with also configuring endpoint(s) in UseKestrel.");
-                    }
+                    var joined = string.Join(", ", _serverAddresses.Addresses);
+                    _logger.LogWarning($"Overriding address(es) '{joined}'. Binding to endpoints defined in {nameof(WebHostBuilderKestrelExtensions.UseKestrel)}() instead.");
 
                     _serverAddresses.Addresses.Clear();
                 }
-                else
+                else if (!hasListenOptions && !hasServerAddresses)
+                {
+                    _logger.LogDebug($"No listening endpoints were configured. Binding to {Constants.DefaultIPEndPoint} by default.");
+                    listenOptions.Add(new ListenOptions(Constants.DefaultIPEndPoint));
+                }
+                else if (!hasListenOptions)
                 {
                     // If no endpoints are configured directly using KestrelServerOptions, use those configured via the IServerAddressesFeature.
                     var copiedAddresses = _serverAddresses.Addresses.ToArray();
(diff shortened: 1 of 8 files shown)


src/Microsoft.AspNetCore.Server.Kestrel/KestrelServer.cs
```diff
@@ -122,7 +122,7 @@ public void Start<TContext>(IHttpApplication<TContext> application)
                     if (addresses.SingleOrDefault() != "http://localhost:5000")
                     {
                         var joined = string.Join(", ", addresses);
-                        throw new NotSupportedException($"Specifying address(es) '{joined}' is incompatible with also configuring endpoint(s) in UseKestrel.");
+                        _logger.LogWarning($"Address(es) '{joined}' were overridden. Binding to endpoints defined in UseKestrel instead.");
```
cesarbs: nit: `nameof(UseKestrel)`
nit 2: add `()` after `UseKestrel` to make it clearer it's referring to the method call

test/Microsoft.AspNetCore.Server.KestrelTests/KestrelServerTests.cs
```diff
@@ -55,6 +56,23 @@ public void StartWithEmptyAddressesThrows()
...
+            var options = new KestrelServerOptions();
+
+            // Directly configuring an endpoint using Listen, causes the IServerAddressesFeature to be ignored.
+            options.Listen(IPAddress.Loopback, 0);
+
+            using (var server = CreateServer(options, testLogger))
+            {
+                server.Features.Get<IServerAddressesFeature>().Addresses.Add("http://localhost:5001");
+                StartDummyApplication(server);
+                Assert.Equal(1, testLogger.TotalWarningsLogged);
```
cesarbs: Check actual message. Also add more addresses to the feature (or better, make it a theory and have cases for single and multiple addresses).

src/Microsoft.AspNetCore.Server.Kestrel/KestrelServer.cs
cesarbs: I think the wording needs to be a little different. It may be confusing to someone specifying multiple addresses in the feature but only overriding a single one of those in `UseKestrel()`.

Maybe something like "Addresses in `UseUrls()`/`IServerAddressesFeature` [pick one] were [are? since it's something that will always happen when using both] overriden by the call to `UseKestrel()`. Binding to endpoints defined in `UseKestrel()` instead."

halter73: Referencing UseUrls() or IServerAddressesFeature is the tricky part. UseUrls() might not be the thing that configured the addresses, and I doubt most ASP.NET Core devs know or care what IServerAddressesFeature is. That's why I just show the addresses that were configured.

cesarbs: Makes sense.

src/Microsoft.AspNetCore.Server.Kestrel/KestrelServer.cs
```diff
@@ -122,7 +122,7 @@ public void Start<TContext>(IHttpApplication<TContext> application)
                     if (addresses.SingleOrDefault() != "http://localhost:5000")
```
Tratcher: Since you don't throw anymore, remove this default check.

halter73:
⬆️  📅

src/Microsoft.AspNetCore.Server.Kestrel/KestrelServer.cs
```diff
@@ -115,19 +115,21 @@ public void Start<TContext>(IHttpApplication<TContext> application)
...
-                    }
+                    var joined = string.Join(", ", _serverAddresses.Addresses);
+                    _logger.LogWarning($"Address(es) '{joined}' were overridden. Binding to endpoints defined in {nameof(WebHostBuilderKestrelExtensions.UseKestrel)}() instead.");
 
                     _serverAddresses.Addresses.Clear();
                 }
-                else
+                else if (!hasListenOptions && !hasServerAddresses)
+                {
+                    listenOptions.Add(Constants.DefaultListenOptions);
```
Tratcher: Log?

PR #1313, opened 2017-01-20
Use ContentLength as primary data source

benaadams:
React to IHeaderDictionary ContentLength change https://github.com/aspnet/HttpAbstractions/pull/757

/cc @Tratcher


benaadams:
Run through a debug and it does correctly call via the interface to `FrameHeaders.ContentLength { get; set;}`

Tratcher:
Hows the perf?

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/FrameHeaders.Generated.cs
```diff
@@ -191,538 +191,537 @@ public StringValues HeaderAllow
...
                 _headers._AccessControlRequestHeaders = value; 
             }
         }
+        public StringValues HeaderContentLength
+        {
+            get
+            {
+                if (ContentLength.HasValue)
+                {
+                    return ContentLength.ToString();
```
Tratcher: HeaderUtilities.FormatInt64.
Would it be worth storing the string?

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/MessageBody.cs
```diff
@@ -266,18 +268,15 @@ protected virtual void OnConsumedBytes(int count)
                 return new ForChunkedEncoding(keepAlive, headers, context);
             }
 
-            var unparsedContentLength = headers.HeaderContentLength;
-            if (unparsedContentLength.Count > 0)
+            if (headers.ContentLength.HasValue)
```
Tratcher: is there still a 400 produced for an invalid content-length header?

benaadams: Is now (was throwing wrong exception, had to split request/response); is thrown during header parsing rather than message body - have added test.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/FrameHeaders.Generated.cs
benaadams: Changed all occurrences.

Don't think its worth storing the string; is another item to keep in sync; check if if exists/generate, clear. Is there a (non-test) scenario where its read multiple times as a string; when the numeric is available?

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/MessageBody.cs
benaadams: Not certain its logged in same way though... looking into it

benaadams:
The api change is having the desired effect on the Content-Length

Prior
```
         Method |                 Type |          Mean |    StdDev |           RPS | Allocated |
--------------- |--------------------- |-------------- |---------- |-------------- |---------- |
 BenchmarkAsync | ContentLengthNumeric |    92.4213 ns | 0.3669 ns | 10,820,017.35 |      32 B |
 BenchmarkAsync |  ContentLengthString |    65.1715 ns | 0.6884 ns | 15,344,128.34 |       0 B |
 BenchmarkAsync |            Plaintext |   128.8769 ns | 1.2060 ns |  7,759,339.56 |       0 B |
 BenchmarkAsync |              Primary |   239.4012 ns | 2.5521 ns |  4,177,087.82 |       0 B |
 BenchmarkAsync |              Unknown |   568.3201 ns | 4.6791 ns |  1,759,571.90 |       0 B |
```
This PR
```
         Method |                 Type |          Mean |     StdDev |           RPS | Allocated |
--------------- |--------------------- |-------------- |----------- |-------------- |---------- |
 BenchmarkAsync | ContentLengthNumeric |    87.9329 ns |  0.3409 ns | 11,372,308.13 |      32 B |
 BenchmarkAsync |  ContentLengthString |    66.5197 ns |  1.1896 ns | 15,033,135.95 |       0 B |
 BenchmarkAsync |            Plaintext |   136.7069 ns |  1.6866 ns |  7,314,920.75 |       0 B |
 BenchmarkAsync |              Primary |   256.5135 ns |  2.8525 ns |  3,898,430.80 |       0 B |
 BenchmarkAsync |              Unknown |   591.6517 ns | 12.8782 ns |  1,690,183.69 |       0 B |
```
This PR with HttpAbstractions
``` 
         Method |                 Type |          Mean |     StdDev |           RPS |
--------------- |--------------------- |-------------- |----------- |-------------- |
 BenchmarkAsync | ContentLengthNumeric |    21.8180 ns |  0.2589 ns | 45,833,772.51 |
 BenchmarkAsync |  ContentLengthString |    66.1873 ns |  0.7702 ns | 15,108,635.21 |
 BenchmarkAsync |            Plaintext |   137.8112 ns |  0.8571 ns |  7,256,306.67 |
 BenchmarkAsync |              Primary |   256.3452 ns |  2.2609 ns |  3,900,989.87 |
 BenchmarkAsync |              Unknown |   592.9667 ns |  7.4288 ns |  1,686,435.44 |
```

However, am getting a perf decrease overall when more headers are added which am looking into

benaadams:
The test is just for adding headers not for response output 😒

test/Microsoft.AspNetCore.Server.KestrelTests/FrameResponseHeadersTests.cs
```diff
@@ -175,16 +175,6 @@ public void ThrowsWhenSettingContentLengthToNonNumericValue(string contentLength
...
-        {
-            var headers = new FrameResponseHeaders();
-
-            var exception = Assert.Throws<InvalidOperationException>(() => headers.SetRawContentLength(contentLength, Encoding.ASCII.GetBytes(contentLength)));
-            Assert.Equal($"Invalid Content-Length: \"{contentLength}\". Value must be a positive integral number.", exception.Message);
-        }
-
-        [Theory]
-        [MemberData(nameof(BadContentLengths))]
         public void ThrowsWhenAssigningHeaderContentLengthToNonNumericValue(string contentLength)
```
halter73: It doesn't look like BadContentLengths contains a negative value. I think we should include that. It would be nice if BadRequestIfContentLengthInvalid also included a test case for negative values.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Infrastructure/MemoryPoolIterator.cs
```diff
@@ -1082,6 +1086,30 @@ public unsafe void CopyFromAscii(string data)
...
 
+        private static byte[] NumericBytesScratch => _numericBytesScratch ?? CreateNumericBytesScratch();
+
+        [MethodImpl(MethodImplOptions.NoInlining)]
+        private static byte[] CreateNumericBytesScratch()
+        {
+            return (_numericBytesScratch = new byte[_maxPositiveLongByteLength]);
+        }
+
+        public void CopyFromNumeric(long value)
```
cesarbs: Where is this being used?

tools/Microsoft.AspNetCore.Server.Kestrel.GeneratedCode/KnownHeaders.cs
```diff
@@ -31,19 +31,27 @@ static string If(bool condition, Func<string> formatter)
                     case {byLength.Key}:
                         {{{Each(byLength, header => $@"
                             if ({header.EqualIgnoreCaseBytes()})
-                            {{
+                            {{{(header.Identifier == "ContentLength" ? $@"
+                                if (ContentLength.HasValue)
+                                {{
+                                    ThrowInvalid{(className == "FrameResponseHeaders" ? "Response" : "Request")}ContentLengthException(AppendValue(HeaderUtilities.FormatInt64(ContentLength.Value), value).ToString());
```
halter73: Is this meant to guard against more than one Content-Length header? If so, an error message like "Invalid Content-Length: \\"{value}\\". Value must be a positive integral number." seems misleading.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Infrastructure/MemoryPoolIterator.cs
```diff
@@ -1082,6 +1086,30 @@ public unsafe void CopyFromAscii(string data)
             _index = blockIndex;
         }
 
+        private static byte[] NumericBytesScratch => _numericBytesScratch ?? CreateNumericBytesScratch();
+
+        [MethodImpl(MethodImplOptions.NoInlining)]
+        private static byte[] CreateNumericBytesScratch()
+        {
+            return (_numericBytesScratch = new byte[_maxPositiveLongByteLength]);
```
halter73: Don't love assignments inside expressions.

test/Microsoft.AspNetCore.Server.KestrelTests/FrameResponseHeadersTests.cs
halter73: I would also add a test specifically for setting a negative Content-Length using the new IHeaderDictionary property.

tools/Microsoft.AspNetCore.Server.Kestrel.GeneratedCode/KnownHeaders.cs
```diff
@@ -547,14 +639,16 @@ public bool MoveNext()
             {{
                 switch (_state)
                 {{
-                    {Each(loop.Headers, header => $@"
-                        case {header.Index}:
-                            goto state{header.Index};
+                    {Each(loop.Headers.Where(header => header.Index >= 0), header => $@"
```
halter73: Using `Index = -1` to distinguish the Content-Length header seems a little weird. Couldn't you just filter on `Name`?

tools/Microsoft.AspNetCore.Server.Kestrel.GeneratedCode/KnownHeaders.cs
```diff
@@ -442,14 +513,14 @@ protected override bool RemoveFast(string key)
         protected override void ClearFast()
         {{            
             MaybeUnknown?.Clear();
-            {(loop.ClassName == "FrameResponseHeaders" ? "_contentLength = null;" : "")}
-            if(FrameHeaders.BitCount(_bits) > 12)
+            ContentLength = null;
+            if(FrameHeaders.BitCount(_bits) > 11)
```
halter73: Wasn't this just a perf optimization for the case a lot of headers (specifically in `_headers` which now wouldn't include Content-Length) were set? Why would require less headers to be set to take this path now?

tools/Microsoft.AspNetCore.Server.Kestrel.GeneratedCode/KnownHeaders.cs
```diff
@@ -192,19 +200,28 @@ public static string GeneratedFile()
                 "TE",
                 "Translate",
                 "User-Agent",
-            }).Concat(corsRequestHeaders).Select((header, index) => new KnownHeader
+            })
+            .Concat(corsRequestHeaders)
+            .Where((header) => header != "Content-Length")
```
halter73: Would it be cleaner just to remove "Content-Length" from commonHeaders?

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/FrameHeaders.cs
```diff
@@ -15,9 +15,10 @@ public abstract class FrameHeaders : IHeaderDictionary
     {
         protected bool _isReadOnly;
         protected Dictionary<string, StringValues> MaybeUnknown;
-
         protected Dictionary<string, StringValues> Unknown => MaybeUnknown ?? (MaybeUnknown = new Dictionary<string, StringValues>(StringComparer.OrdinalIgnoreCase));
 
+        public long? ContentLength { get; set; }
```
halter73: We should have a custom setter, so we can throw immediately if this is set to a negative value or _isReadOnly is true.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
```diff
@@ -920,13 +919,13 @@ private Task WriteSuffix()
...
             {
-                if (!hasTransferEncoding && !responseHeaders.HasContentLength)
+                if (!hasTransferEncoding && !responseHeaders.ContentLength.HasValue)
                 {
                     if (appCompleted && StatusCode != StatusCodes.Status101SwitchingProtocols)
                     {
                         // Since the app has completed and we are only now generating
                         // the headers we can safely set the Content-Length to 0.
-                        responseHeaders.SetRawContentLength("0", _bytesContentLengthZero);
+                        responseHeaders.ContentLength = 0;
```
benaadams: Is an extra 8 byte field in every header set to make the occasional 1 byte write of `0` slightly faster (as most responses will have some content) - can always special case 0 in the copyfrom..?

test/Microsoft.AspNetCore.Server.KestrelTests/FrameResponseHeadersTests.cs
benaadams: Yep

tools/Microsoft.AspNetCore.Server.Kestrel.GeneratedCode/KnownHeaders.cs
benaadams: Will add extra exception

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
halter73: Your justification makes sense.

tools/Microsoft.AspNetCore.Server.Kestrel.GeneratedCode/KnownHeaders.cs
benaadams: Does throwing from header parsing get logged ok, or do I have to call logger?

halter73: Throwing a BadHttpRequestException should be sufficient, since it's supposed to be caught and logged [in RequestProcessingAsync](https://github.com/aspnet/KestrelHttpServer/blob/10fe5e6fa25e95ab9acd2b3bd733e78be4315a37/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/FrameOfT.cs#L199).

It looks like `SetBadRequestState` no longer logs the error and our tests didn't catch it though... 😢

benaadams: Is sometimes https://github.com/aspnet/KestrelHttpServer/blob/dev/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs#L1506-L1510

halter73: The logging is broken when throwing a BadHttpRequestException without using RejectRequest() for cases when the exception can't easily be thrown from inside the Frame class as is the case here.

A situation very similar to the one you're running into now can be found [elsewhere in this class](https://github.com/aspnet/KestrelHttpServer/blob/10fe5e6fa25e95ab9acd2b3bd733e78be4315a37/tools/Microsoft.AspNetCore.Server.Kestrel.GeneratedCode/KnownHeaders.cs#L58).

[This commit](https://github.com/aspnet/KestrelHttpServer/commit/78584799a412ef24327dca5b3c62315d708bb44c) regressed the logging, but I can't find the PR that changed it. The commit seems to have been pushed sometime between #1163 and #1166.

@cesarbs Can you take a look? We can create an issue if it requires a big change.

benaadams: This PR? https://github.com/aspnet/KestrelHttpServer/pull/993

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/FrameHeaders.cs
benaadams: With throw an InvalidOp for both Request and Response as will need to be manually set rather than input Request set (so server error 500)

tools/Microsoft.AspNetCore.Server.Kestrel.GeneratedCode/KnownHeaders.cs
halter73: Good find. I wasn't looking through @mikeharder's PRs.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/FrameHeaders.cs
halter73: Perfect.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Infrastructure/MemoryPoolIterator.cs
benaadams: Changing to `ulong` as it doesn't handle negatives; either that or make internal

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/FrameHeaders.cs
benaadams: ArgumentOutOfRangeException instead

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Infrastructure/MemoryPoolIterator.cs
benaadams: Changed

test/Microsoft.AspNetCore.Server.KestrelTests/FrameResponseHeadersTests.cs
benaadams: Added

benaadams:
Added bunch more tests; resolved feedback - other than GeneratedHeaders stuff, want to resolve the perf regression

tools/Microsoft.AspNetCore.Server.Kestrel.GeneratedCode/KnownHeaders.cs
benaadams: Throwing multiple content lengths instead

benaadams:
Is a double memory barrier in the dasm https://github.com/dotnet/coreclr/issues/9039 which seems weird

benaadams:
Ok... went down this "lost performance" rabbit hole as was setting the Content-Length as string in some of the tests 🤒 

Anyway, improves found along the way...

benaadams:
Looking better

Pre
```
                 Type |          Mean |           RPS | Allocated |
--------------------- |-------------- |-------------- |---------- |
               Common | 1,059.8157 ns |    943,560.26 |      31 B |
 ContentLengthNumeric |    84.4719 ns | 11,838,254.08 |      31 B |
  ContentLengthString |    58.5621 ns | 17,075,894.64 |       0 B |
            Plaintext |   156.3607 ns |  6,395,468.47 |      31 B |
              Unknown |   697.8674 ns |  1,432,937.05 |      31 B |
```
Post
```
                 Type |          Mean |           RPS | Allocated |
--------------------- |-------------- |-------------- |---------- |
               Common | 1,083.2790 ns |    923,123.25 |      31 B |
 ContentLengthNumeric |    97.4643 ns | 10,260,171.78 |      31 B |
  ContentLengthString |    76.1497 ns | 13,132,024.10 |       0 B |
            Plaintext |   146.6288 ns |  6,819,943.26 |      31 B |
              Unknown |   734.3215 ns |  1,361,801.27 |      31 B |
```
Post + HttpAbstractions
```
                 Type |          Mean |           RPS |
--------------------- | ------------- |-------------- |
               Common |   988.2274 ns |  1,011,912.81 |
 ContentLengthNumeric |    41.3301 ns | 24,195,469.03 |
  ContentLengthString |    76.1173 ns | 13,137,625.51 |
            Plaintext |    82.6533 ns | 12,098,724.17 |
              Unknown |   567.9350 ns |  1,760,764.73 |
```

Still need to clean-up KnownHeaders

benaadams:
Not happy with Output perf; working on it 😢

benaadams:
Fixed the Header output

Before
```
        Method |                 Type |          Mean |     StdDev |           RPS |
-------------- |--------------------- |-------------- |----------- |-------------- |
 OutputHeaders | ContentLengthNumeric |    80.0507 ns |  1.9779 ns | 12,492,079.57 |
 OutputHeaders |            Plaintext |   146.3276 ns |  1.6375 ns |  6,833,978.82 |
 OutputHeaders |               Common |   831.6703 ns | 12.6436 ns |  1,202,399.61 |
 OutputHeaders |              Unknown | 1,213.7093 ns | 10.6639 ns |    823,920.50 |
```

After
```
        Method |                 Type |          Mean |     StdDev |           RPS |
-------------- |--------------------- |-------------- |----------- |-------------- |
 OutputHeaders | ContentLengthNumeric |    68.0594 ns |  0.8850 ns | 14,693,048.41 |
 OutputHeaders |            Plaintext |   138.5749 ns |  2.3576 ns |  7,216,314.93 |
 OutputHeaders |               Common |   826.9283 ns | 13.1998 ns |  1,209,294.71 |
 OutputHeaders |              Unknown | 1,245.5972 ns | 10.3146 ns |    802,827.72 |
```

tools/Microsoft.AspNetCore.Server.Kestrel.GeneratedCode/KnownHeaders.cs
benaadams: Fixed

tools/Microsoft.AspNetCore.Server.Kestrel.GeneratedCode/KnownHeaders.cs
benaadams: Fixed

tools/Microsoft.AspNetCore.Server.Kestrel.GeneratedCode/KnownHeaders.cs
benaadams: Fixed

tools/Microsoft.AspNetCore.Server.Kestrel.GeneratedCode/KnownHeaders.cs
benaadams: Fixed

src/Microsoft.AspNetCore.Server.Kestrel/BadHttpRequestException.cs
```diff
@@ -43,6 +43,9 @@ internal static BadHttpRequestException GetException(RequestRejectionReason reas
                 case RequestRejectionReason.MalformedRequestInvalidHeaders:
                     ex = new BadHttpRequestException("Malformed request: invalid headers.", StatusCodes.Status400BadRequest);
                     break;
+                case RequestRejectionReason.MultipleContentLengths:
+                    ex = new BadHttpRequestException("Multiple content length headers.", StatusCodes.Status400BadRequest);
```
cesarbs: nit: use header name `Content-Length`

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/MessageBody.cs
```diff
@@ -309,6 +308,28 @@ protected override ValueTask<ArraySegment<byte>> PeekAsync(CancellationToken can
...
+            protected override ValueTask<ArraySegment<byte>> PeekAsync(CancellationToken cancellationToken)
+            {
+                return new ValueTask<ArraySegment<byte>>();
+            }
+
+            protected override void OnConsumedBytes(int count)
+            {
+                if (count > 0)
+                {
+                    throw new InvalidDataException("Consuming non-existant data");
```
cesarbs: non-exist**e**nt

test/Microsoft.AspNetCore.Server.KestrelTests/BadHttpRequestTests.cs
```diff
@@ -202,6 +202,19 @@ public class BadHttpRequestTests
             }
         }
 
+        [Fact]
+        public async Task BadRequestIfContentLengthInvalid()
```
cesarbs: Test with negative value too.

test/Microsoft.AspNetCore.Server.KestrelTests/FrameResponseHeadersTests.cs
```diff
@@ -151,6 +153,86 @@ public void ThrowsWhenClearingHeadersAfterReadOnlyIsSet()
             Assert.Throws<InvalidOperationException>(() => dictionary.Clear());
         }
 
+        [Fact]
+        public void CorrectContentLengthsOutput()
```
cesarbs: This should be in `MemoryPoolIteratorTests`.

src/Microsoft.AspNetCore.Server.Kestrel/BadHttpRequestException.cs
benaadams: Fixed

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/MessageBody.cs
benaadams: Fixed

test/Microsoft.AspNetCore.Server.KestrelTests/BadHttpRequestTests.cs
benaadams: Fixed

test/Microsoft.AspNetCore.Server.KestrelTests/FrameResponseHeadersTests.cs
benaadams: Moved

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/FrameHeaders.cs
```diff
@@ -5,19 +5,31 @@
...
 {
     public abstract class FrameHeaders : IHeaderDictionary
     {
+        protected long? _contentLength;
         protected bool _isReadOnly;
         protected Dictionary<string, StringValues> MaybeUnknown;
-
         protected Dictionary<string, StringValues> Unknown => MaybeUnknown ?? (MaybeUnknown = new Dictionary<string, StringValues>(StringComparer.OrdinalIgnoreCase));
 
+        public long? ContentLength {
```
Tratcher: nit: style, bracket on new line

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/FrameHeaders.cs
```diff
@@ -412,11 +423,26 @@ public static unsafe TransferCoding GetFinalTransferCoding(StringValues transfer
             return transferEncodingOptions;
         }
 
-        private static void ThrowInvalidContentLengthException(string value)
+        protected static void ThrowInvalidResponseContentLengthException(long value)
```
benaadams: These exceptions can move to respective classes (Request/Response)

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/FrameHeaders.cs
```diff
@@ -412,11 +423,26 @@ public static unsafe TransferCoding GetFinalTransferCoding(StringValues transfer
...
+        {
+            throw new ArgumentOutOfRangeException($"Invalid Content-Length: \"{value}\". Value must be a positive integral number.");
+        }
+
+        protected static void ThrowInvalidResponseContentLengthException(string value)
```
benaadams: As above

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/FrameHeaders.cs
```diff
@@ -412,11 +423,26 @@ public static unsafe TransferCoding GetFinalTransferCoding(StringValues transfer
...
         {
             throw new InvalidOperationException($"Invalid Content-Length: \"{value}\". Value must be a positive integral number.");
         }
 
+        protected static void ThrowInvalidRequestContentLengthException(string value)
```
benaadams: As above

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/FrameHeaders.cs
benaadams: Resolved

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/FrameHeaders.cs
benaadams: Resolved

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/FrameHeaders.cs
benaadams: Resolved

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/FrameHeaders.cs
benaadams: Resolved

tools/Microsoft.AspNetCore.Server.Kestrel.GeneratedCode/KnownHeaders.cs
```diff
@@ -507,35 +533,46 @@ protected void CopyToFast(ref MemoryPoolIterator output)
                         }}
                     }}
 
+                    if({header.TestNotTempBit()})
```
halter73: Do we know this is actually faster than doing the `tempBits &= ~{1L << header.Index}L;` before the if condition and then comparing `tempBits` to zero. If so, why are you doing it the old way just for the Content-Length header?

benaadams: Yes is faster, measured before and after, significant but not vast; missed content-length accidentally because its now a snowflake 😢

benaadams: Looks like I completely missed a commit 😱

tools/Microsoft.AspNetCore.Server.Kestrel.GeneratedCode/KnownHeaders.cs
```diff
@@ -228,13 +231,23 @@ public static string GeneratedFile()
...
-            }).ToArray();
+            })
+            .Concat(new[] { new KnownHeader
+            {
+                Name = "Content-Length",
+                Index = -1,
+                EnhancedSetter = enhancedHeaders.Contains("Content-Length"),
+                PrimaryHeader = responsePrimaryHeaders.Contains("Content-Length")
+            }})
+            .ToArray();
```
halter73: Nit: We should have done this before, but can we assert that `responseHeaders` and `requestHeaders` have no more than 64 items and/or the max `Index` is 62.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Infrastructure/MemoryPoolIterator.cs
```diff
@@ -1082,6 +1086,103 @@ public unsafe void CopyFromAscii(string data)
...
+
+            var blockIndex = _index;
+            var bytesLeftInBlock = block.Data.Offset + block.Data.Count - blockIndex;
+            var start = block.DataFixedPtr + blockIndex;
+
+            if (value < 10)
+            {
+                if (bytesLeftInBlock < 1)
+                {
+                    goto overflow;
```
halter73: Is this that much better than just calling `CopyFromNumericOverflow(value);` and returning? I prefer avoiding goto if we can.

tools/Microsoft.AspNetCore.Server.Kestrel.GeneratedCode/KnownHeaders.cs
benaadams: 63 for reponseHeaders as it steals one for Content-Length in `CopyTo(ref MemoryPoolIterator output)`

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Infrastructure/MemoryPoolIterator.cs
benaadams: Not in a way I could probably sensibility justify 😄

halter73:
:shipit: pending last few comments

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Infrastructure/MemoryPoolIterator.cs
benaadams: Not in a way I could probably sensibility justify... It generates the code I'd like the Jit to generate 😆

benaadams:
Git for windows didn't include all the changes in the final commit; fixed; also addressed feedback

benaadams:
Most of the changes were in; though hadn't picked up the `KnownHeaders` file; though picked up the Generated one.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Infrastructure/MemoryPoolIterator.cs
benaadams: Resolved

tools/Microsoft.AspNetCore.Server.Kestrel.GeneratedCode/KnownHeaders.cs
benaadams: Resolved

tools/Microsoft.AspNetCore.Server.Kestrel.GeneratedCode/KnownHeaders.cs
benaadams: Resolved

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/FrameRequestHeaders.cs
```diff
@@ -3,12 +3,59 @@
...
+            Unknown.TryGetValue(key, out existing);
+            Unknown[key] = AppendValue(existing, value);
+        }
+
+        private static void ThrowInvalidContentLengthException(string value)
+        {
+            throw BadHttpRequestException.GetException(RequestRejectionReason.InvalidContentLength, value);
+        }
+
+        private static void ThrowMultipleContentLengths()
```
cesarbs: nit: rename to `ThrowMultipleContentLengthsException`

benaadams: Resolved

PR #1323, opened 2017-01-24
[Don't merge] Update to version 1.0.3.

cesarbs:
Just want some eyes on it. Did I miss anything?

@pranavkm @Eilon @halter73 @pakrym


pranavkm:
Also add a SDK version to the global.json. I think older branches are missing those. Looks :shipit: otherwise.

Issue #1264, opened 2016-12-19
Memory Leak in 1.1.0


cesarbs:
Original issue: https://github.com/aspnet/KestrelHttpServer/issues/1260

Should include this in 1.1.1.

cesarbs:
cc @muratg

muratg:
Putting this in 1.1.1 for consideration.

Eilon:
This patch bug is **approved**. Please use the normal code review process w/ a PR and make sure the fix is in the correct branch, then close the bug and mark it as done.

Eilon:
@muratg please assign.

halter73:
We should make sure that #1311 also gets backported so the fix doesn't break timeouts on 32-bit platforms.

PR #1329, opened 2017-01-25
[1.1.1] Fix memory leak caused by closure allocations in KestrelThread (#1264)

cesarbs:
#1264 

@halter73 @pakrym


src/Microsoft.AspNetCore.Server.Kestrel/Internal/Infrastructure/KestrelThread.cs
```diff
@@ -19,10 +20,17 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal
     /// </summary>
     public class KestrelThread
     {
-        public const long HeartbeatMilliseconds = 1000;
+        public const long HeartbeatMilliseconds = 10;
```
halter73: ?

cesarbs:
🔔

halter73:
:shipit:

Issue #1303, opened 2017-01-10
Call OnStarting before verifying response length


cesarbs:
Original issue: https://github.com/aspnet/KestrelHttpServer/issues/1289

Eilon:
This patch bug is **approved**. Please use the normal code review process w/ a PR and make sure the fix is in the correct branch, then close the bug and mark it as done.

PR #1330, opened 2017-01-25
[1.1.1] Call OnStarting before verifying response length (#1303)

cesarbs:
#1303

Also don't close connection when Content-Length set but no bytes written.

@halter73 @pakrym


cesarbs:
🔔

halter73:
:shipit:

PR #1343, opened 2017-02-07
Generate code as part of the MSBuild build process

cesarbs:
@natemcmaster @halter73 @pakrym @davidfowl


davidfowl:
How often does this happen? Is it incremental? Whatever you do make sure it doesn't happen for every f5, that will destroy the build performance (unless you make it incremental).

cesarbs:
@davidfowl Should be incremental since I'm using the `Inputs` and `Outputs` properties in the target, but it runs every time nonetheless :disappointed: I guess I can add a project reference to the code gen csproj and skip the explicit restore, but I was trying to avoid that.

src/Microsoft.AspNetCore.Server.Kestrel/Microsoft.AspNetCore.Server.Kestrel.csproj
```diff
@@ -26,4 +26,21 @@
     <PackageReference Include="System.Threading.ThreadPool" Version="4.4.0-*" />
   </ItemGroup>
 
+  <ItemGroup>
+    <CodeGenInput Include="..\..\tools\Microsoft.AspNetCore.Server.Kestrel.GeneratedCode\FrameFeatureCollection.cs" />
+    <CodeGenInput Include="..\..\tools\Microsoft.AspNetCore.Server.Kestrel.GeneratedCode\KnownHeaders.cs" />
+    <CodeGenOutput Include="Internal\Http\FrameHeaders.Generated.cs" />
+    <CodeGenOutput Include="Internal\Http\Frame.Generated.cs" />
```
natemcmaster: The Generated.cs files are committed to source, right?

src/Microsoft.AspNetCore.Server.Kestrel/Microsoft.AspNetCore.Server.Kestrel.csproj
```diff
@@ -26,4 +26,21 @@
...
+  </ItemGroup>
+
+  <Target Name="GenerateCode"
+          Inputs="@(CodeGenInput)"
+          Outputs="@(CodeGenOutput)"
+          BeforeTargets="CoreCompile">
+    <Exec Command="dotnet restore"
```
natemcmaster: Is this not picked up automatically by the initial restore?

src/Microsoft.AspNetCore.Server.Kestrel/Microsoft.AspNetCore.Server.Kestrel.csproj
cesarbs: Yes.

src/Microsoft.AspNetCore.Server.Kestrel/Microsoft.AspNetCore.Server.Kestrel.csproj
cesarbs: It is if I run a full build. This would support building just `Microsoft.AspNetCore.Server.Kestrel.csproj`. We might not need that though - I did it so I could quickly iterate from empty repo to getting this working :smile:

davidfowl: Just add it as a project reference so restore always runs. Can you mark project references as build time only? (PrivateAssets=all?)

natemcmaster: We shouldn't need it. It should be picked up as a part of the normal restore operations.

natemcmaster:
MSBuild uses file timestamps, not file contents, to determine incremental build. I'm guessing this runs every time because the tool always writes contents to disk even if nothing changed.

cesarbs:
@natemcmaster Not really: https://github.com/aspnet/KestrelHttpServer/blob/dev/tools/Microsoft.AspNetCore.Server.Kestrel.GeneratedCode/Program.cs#L16-L19

src/Microsoft.AspNetCore.Server.Kestrel/Microsoft.AspNetCore.Server.Kestrel.csproj
davidfowl: If you do a solution restore sure. If you restore kestrel by itself, this won't restore without a p2p.

cesarbs:
I've made the restore conditional on `obj\project.assets.json` being there.

natemcmaster:
What is the main problem you are trying to solve? I should have asked this first. It seems like we don't touch the generated code part of this very often, so running one every build seems like overkill.

cesarbs:
I want to code gen as part of the build, but conditional on the files having changed. This would make the code gen step part of the VS build (right now you can only do it from the command line). Would also get rid of our makefile.

natemcmaster:
Did you look into using StartArguments? e.g. you could set the following in the project. Everytime you "dotnet run" from command line or "F5" from VS, it would launch the app with the right arguments.
```xml
<PropertyGroup>
    <StartWorkingDirectory>$(MSBuildThisFileDirectory)..\..\src\Microsoft.AspNetCore.Server.Kestrel\Internal\Http</StartWorkingDirectory>
    <StartArguments>FrameHeaders.Generated.cs Frame.Generated.cs</StartArguments>
</PropertyGroup>
```

My only concern here is that it shell executes "dotnet.exe", which is not necessarily the right thing to do inside a VS build.

src/Microsoft.AspNetCore.Server.Kestrel/Microsoft.AspNetCore.Server.Kestrel.csproj
```diff
@@ -26,4 +26,22 @@
...
+    <CodeGenInput Include="..\..\tools\Microsoft.AspNetCore.Server.Kestrel.GeneratedCode\FrameFeatureCollection.cs" />
+    <CodeGenOutput Include="Internal\Http\FrameHeaders.Generated.cs" />
+    <CodeGenOutput Include="Internal\Http\Frame.Generated.cs" />
+  </ItemGroup>
+
+  <Target Name="GenerateCode"
+          Inputs="@(CodeGenInput)"
+          Outputs="@(CodeGenOutput)"
+          BeforeTargets="CoreCompile">
+    <Exec Condition="!Exists('..\..\tools\Microsoft.AspNetCore.Server.Kestrel.GeneratedCode\obj\project.assets.json')"
```
halter73: Is this basically the equivalent for checking if the project.lock.json exists in the old build system? Does this mean I'll have to clean the repo if the dependencies get out of date/sync?

halter73:
@natemcmaster I would probably accept a PR that used .tt assuming it works well on all platforms if someone wants to put in the time. It is nice to be able to have the flexibility of C# though.

src/Microsoft.AspNetCore.Server.Kestrel/Microsoft.AspNetCore.Server.Kestrel.csproj
natemcmaster: This is the equivalent of the lock file. You would just need to re-run restore to update dependencies. Or git clean.

natemcmaster:
I'm 100% okay with the C# stuff. It's more straightforward than .tt (which is a VS-only thing anyways.)

Running it for every compilation of Kestrel seems like a bit of overkill. A more natural place for this is a post-build step into the generated code project itself. You can use the sln file to ensure MSBuild compiles the generated code project before the src projects without using a projectreference (which would have side-effects on the runtime project's dependencies).

cesarbs:
@natemcmaster I like that, will move it to the code gen csproj.

cesarbs:
Updated.

natemcmaster:
One more thing, can you update the SLN so the generated code project runs before the kestrel runtime build?

cesarbs:
@natemcmaster That wouldn't work in our command-line build, right? Why not a `ProjectReference` with `ReferenceOutputAssembly="false"`?

natemcmaster:
At the moment, no. But KoreBuild will update soon to build the sln file.

Also, you can't reference netcoreapp1.0 from a netstandard1.3 project.

cesarbs:
@natemcmaster Solution updated as requested.

natemcmaster:
:shipit:

halter73:
Let's add a functional test to verify the generated code is what it should be.

cesarbs:
Will close this an open a new PR for the code gen diff test.

Issue #1249, opened 2016-12-06
Unobserved task exception ECONNRESET 


jahmai:
I'm getting the following exception triggering the TaskScheduler.UnobservedTaskException:

System.AggregateException: A Task's exception(s) were not observed either by Waiting on the Task or accessing its Exception property. As a result, the unobserved exception was rethrown by the finalizer thread. ---> System.IO.IOException: Error -4077 ECONNRESET connection reset by peer ---> Microsoft.AspNetCore.Server.Kestrel.Internal.Networking.UvException: Error -4077 ECONNRESET connection reset by peer
   --- End of inner exception stack trace ---
   --- End of inner exception stack trace ---
   --- End of inner exception stack trace ---<---

The stack is next to useless but that's not my fault :-)

Seems Kestrel isn't catching or inspecting Task.Exception somewhere.

halter73:
@jahmai Do you have any associated stack trace?

jahmai:
@halter73 Sadly, you're looking at it. I've also stopped in the debugger and it actually looks like the exception wasn't actually "thrown" and so doesn't have a stack trace. Kind of like it was allocated and placed into a TaskCompletionSource and never await'd on.

cesarbs:
@jahmai Are you using HTTPS?

jahmai:
@cesarbs Not in Kestral. We terminate HTTPS at our load balancers and talk to Kestral via HTTP.

davidfowl:
@jahmai do you manually read the request body anywhere?

muratg:
@jahmai *ping*.  Do you read the request body in your code?

jahmai:
@muratg @davidfowl Sorry went on holidays and forgot about work for a while. We found one place where we use Request.Form.Files[0].OpenReadStream(), but don't dispose the Stream. Could this be the cause?

halter73:
@jahmai I doubt it. It's more likely that somewhere there is a call to an async API like Stream.ReadAsync(...) without an await or .Wait().

jahmai:
@halter73 Hmm well we don't use Request.Body at all. There a 3 usages of Request.Form.Files[0] as mentioned previously but the Stream content is consumed immediately. Nothing else directly accessing the Request other than that.

[...]

PR #1345, opened 2017-02-07
Don't throw unobserved exception from AdaptedPipeline.ReadInputAsync

halter73:
When I was investigating #1249 using Kestrel's SampleApp I saw the following unobserved exceptions.

```
   {"The decryption operation failed, see inner exception."}
   {"The certificate chain was issued by an authority that is not trusted"}
   at System.Net.Security._SslStream.EndRead(IAsyncResult asyncResult)
   at System.Threading.Tasks.TaskFactory`1.FromAsyncTrimPromise`1.Complete(TInstance thisRef, Func`3 endMethod, IAsyncResult asyncResult, Boolean requiresSynchronization)
   at Microsoft.AspNetCore.Server.Kestrel.Adapter.Internal.LoggingStream.<ReadAsync>d__17.MoveNext()
   at Microsoft.AspNetCore.Server.Kestrel.Adapter.Internal.AdaptedPipeline.<ReadInputAsync>d__9.MoveNext()
   
TaskCanceledException "The request was aborted"
      at Microsoft.AspNetCore.Server.Kestrel.Internal.Http.SocketInput.CheckConnectionError()
   at Microsoft.AspNetCore.Server.Kestrel.Internal.Http.SocketInput.GetResult()
   at Microsoft.AspNetCore.Server.Kestrel.Internal.Http.SocketInputExtensions.<ReadAsyncAwaited>d__1.MoveNext()
   at Microsoft.AspNetCore.Server.Kestrel.Adapter.Internal.LoggingStream.<ReadAsync>d__17.MoveNext()
   at Microsoft.AspNetCore.Server.Kestrel.Adapter.Internal.AdaptedPipeline.<ReadInputAsync>d__9.MoveNext()
```

When I removed the connection filters, I still saw "The request was aborted" TaskCanceledExceptions. Those aren't fixed by this PR. I think it is better to wait until #1277 until fixing the non ConnectionAdapter related unobserved exceptions.

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
@@ -404,10 +404,8 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Http
         /// </summary>
         public Task StopAsync()
         {
-            if (!_requestProcessingStopping)
-            {
-                _requestProcessingStopping = true;
-            }
+            _requestProcessingStopping = true;
+            Input.CompleteAwaiting();
 
             return _requestProcessingTask ?? TaskCache.CompletedTask;
         }
--- a/test/Microsoft.AspNetCore.Server.KestrelTests/ConnectionAdapterTests.cs
+++ b/test/Microsoft.AspNetCore.Server.KestrelTests/ConnectionAdapterTests.cs
@@ -109,7 +109,7 @@ namespace Microsoft.AspNetCore.Server.KestrelTests
         }
 
         [Fact]
-        public async Task ThrowingOnReadDoesNotUnobservedException()
+        public async Task ThrowingOnReadDoesNotCauseUnobservedException()
         {
             var unobservedExceptionThrown = false;
 
@@ -133,8 +133,16 @@ namespace Microsoft.AspNetCore.Server.KestrelTests
                 {
                     using (var connection = server.CreateConnection())
                     {
-                        await connection.Send("GET / HTTP/1.1\r\n\r\n");
-                        await connection.ReceiveEnd();
+                        try
+                        {
+                            await connection.Send("GET / HTTP/1.1\r\n\r\n");
+                            await connection.ReceiveEnd();
+                        }
+                        catch (IOException)
+                        {
+                            // Since Stream.ReadAsync throws, the send may fail to complete
+                            // if the server aborts the connection quickly enough.
+                        }
                     }
                 }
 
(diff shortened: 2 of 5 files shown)


test/Microsoft.AspNetCore.Server.KestrelTests/ConnectionAdapterTests.cs
```diff
@@ -109,6 +108,46 @@ public class ConnectionAdapterTests
             }
         }
 
+        [Fact]
+        public async Task ThrowingOnReadDoesNotUnobservedException()
```
cesarbs: nit: DoesNot**Throw**

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
```diff
@@ -404,10 +404,8 @@ public void Start()
         /// </summary>
         public Task StopAsync()
         {
-            if (!_requestProcessingStopping)
-            {
-                _requestProcessingStopping = true;
-            }
+            _requestProcessingStopping = true;
```
davidfowl: Why this change?

test/Microsoft.AspNetCore.Server.KestrelTests/ConnectionAdapterTests.cs
```diff
@@ -109,6 +108,54 @@ public class ConnectionAdapterTests
...
+            var unobservedExceptionThrown = false;
+
+            EventHandler<UnobservedTaskExceptionEventArgs> unobservedExceptionHandler = (sender, args) =>
+            {
+                unobservedExceptionThrown = true;
+            };
+
+            try
+            {
+                TaskScheduler.UnobservedTaskException += unobservedExceptionHandler;
```
davidfowl: This just looks like trouble waiting to happen. You may end up observing unobserved tasks in tests since they run in parallel.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
cesarbs: The check is completely redundant.

test/Microsoft.AspNetCore.Server.KestrelTests/ConnectionAdapterTests.cs
halter73: True, but after the PR there shouldn't be any 😄

Obviously there's a risk that other test code could do this though. Would you recommend reporting exception details so we could track down any unobserved tasks this happens to catch? Only fail if it's the specific exception instance we threw? Or just remove the test?

davidfowl: Remove the test and instead just have the handler in the sample. It can be used as a quick sanity check if we need to verify behavior. We could also potentially move it to an end to end test instead (part of the server tests).

cesarbs: Create a separate test project just for this 💩 :trollface:

Issue #1335, opened 2017-02-01
HTTPS connections are very slow in Kestrel 1.0.X on .NET Framework


mikeharder:
# Repro Steps
1. Create an application targeting Kestrel 1.0.X and .NET Framework which listens on both HTTP and HTTPS endpoints.  Sample at https://github.com/mikeharder/KestrelHttpsPerf.
2. Ensure the matching .NET Framework Developer Pack is installed (http://getdotnet.azurewebsites.net/target-dotnet-platforms.html).
3. Build and run.
4. From a Linux machine, run "wrk -c 256 -t 32 -d 10 http://server:port/path".  RPS should be normal.
5. Run the same wrk command against the HTTPS endpoint.  No responses will be sent.  Server CPU is 0%.

# Notes
* Only repros with Kestrel 1.0.X on .NET Framework.
  * Doesn't repro with Kestrel 1.0.X on on .NET Core.
  * Doesn't repro with Kestrel 1.1.0 on .NET Framework.

Eilon:
This patch bug is **approved**. Please use the normal code review process w/ a PR and make sure the fix is in the correct branch, then close the bug and mark it as done.

PR #1350, opened 2017-02-09
Implement APM methods in streams (#1335)

cesarbs:
#1335 

Basically just replicated the changes from https://github.com/aspnet/KestrelHttpServer/commit/4337d2a4a751ee7ccb0e0c460670c0abd044d040 and https://github.com/aspnet/KestrelHttpServer/commit/5aee5247685b3e3f9157e229dfc2c7ba9248dcb3.

@halter73 @mikeharder @Eilon

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Kestrel/Filter/Internal/LibuvStream.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/Filter/Internal/LibuvStream.cs
@@ -132,5 +132,83 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Filter.Internal
         {
             return _input.ReadAsync(buffer.Array, buffer.Offset, buffer.Count);
         }
+
+#if NET451
+        public override IAsyncResult BeginRead(byte[] buffer, int offset, int count, AsyncCallback callback, object state)
+        {
+            var task = ReadAsync(buffer, offset, count, default(CancellationToken), state);
+            if (callback != null)
+            {
+                task.ContinueWith(t => callback.Invoke(t));
+            }
+            return task;
+        }
+
+        public override int EndRead(IAsyncResult asyncResult)
+        {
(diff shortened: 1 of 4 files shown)


davidfowl:
Where's the CI run?

davidfowl:
Here's a much more efficient helper to go from Task to APM:

https://github.com/dotnet/corefxlab/blob/master/src/System.IO.Pipelines/TaskToApm.cs

halter73:
@davidfowl  Cool. Using a custom IAsyncResult or the original Task if it has the correct AsyncState seems a lot more efficient than always using ContinueWith() and a TCS. Special casing completed tasks to avoid closure allocations is also smart.

Unfortunately, this change is a 1.0.3 backport, so we should probably use the well tested code that still immensely improves the perf of the HttpsConnectionFilter on desktop .NET.

We should look into using TaskToApm for 2.0. Is TaskToApm available publicly in any packages we produce? If not, are there any plans to do so?

cesarbs:
Was about to reply that although that's more efficient, we should stick to the well tested code we have in 1.1 for this backport.

Drawaes:
I would be interested if it is public. Internal classes are making copy/paste boring

davidfowl:
> Unfortunately, this change is a 1.0.3 backport, so we should probably use the well tested code that still immensely improves the perf of the HttpsConnectionFilter on desktop .NET.

This code is well tested, it came from corefx and is used there. We're already using it indirectly (via SSL Stream).

> We should look into using TaskToApm for 2.0. Is TaskToApm available publicly in any packages we produce? If not, are there any plans to do so?

Maybe we can add something in System.Threading.Task.Extensions?
/cc @stephentoub

halter73:
Fair enough. The integration is well tested as well. TaskToApm seems simple enough to use correctly, but I don't think the perf of this backport is huge concern as long as it isn't so bad that it's completely broken like before due to the synchronous calls that were made.

We should also update all of our Stream implementations at once.

Issue #1347, opened 2017-02-07
[1.0.3] Fix Crash under high load (#1327)


halter73:
#1327

@Eilon

Eilon:
This patch bug is **approved**. Please use the normal code review process w/ a PR and make sure the fix is in the correct branch, then close the bug and mark it as done.

PR #1355, opened 2017-02-10
Prevent ODE when ReadStart/Stop() is queued after disposal

halter73:
#1347 

@cesarbs


halter73:
Replaced by #1356. This PR had the wrong branch name 😲

PR #1359, opened 2017-02-11
ReadingInput tweak

benaadams:
As raised by @stephentoub in https://github.com/aspnet/KestrelHttpServer/commit/3bb7f4e2c407be65323150fe769613ae2edcbf9e#commitcomment-20845728

/cc @pakrym not sure what the pipelines version does?

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/SocketInput.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/SocketInput.cs
@@ -43,7 +43,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Http
 
         public bool IsCompleted => ReferenceEquals(_awaitableState, _awaitableIsCompleted);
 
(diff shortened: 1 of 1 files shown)


davidfowl:
The pipelines version doesn't use a tcs anymore, it just checks the exception status. https://github.com/dotnet/corefxlab/blob/173835187f21124511ba0abad28b89b40f7295c1/src/System.IO.Pipelines/PipeCompletion.cs#L17

davidfowl:
I'm not sure if we're taking any more PRs until we unblock the pipelines change.

/cc @pakrym @halter73 @cesarbs

benaadams:
Strange CI error?

>   Failed   Microsoft.AspNetCore.Server.Kestrel.FunctionalTests.AddressRegistrationTests.RegisterAddresses_IPv6ScopeId_Success
EXEC : error Message:  [C:\projects\kestrelhttpserver\.build\makefile.proj]
   System.InvalidOperationException : The test method expected 2 parameter values, but 0 parameter values were provided.

pakrym:
I don't think we should take PR's for socketinput anymore. Most of the code is going away.

halter73:
Agreed.

PR #1378, opened 2017-02-21
WIP DO NOT MERGE: Better parsing

davidfowl:
The goal of this PR is to review a more efficient pattern for using the `ReadableBuffer` and friends to parse an http request. I haven't done extensive testing yet so I'm sure there are off by one errors somewhere. My goal was to reduce the number of calls per request. I tested a single request going through POSTMAN (https://www.getpostman.com/):

```C#
GET  HTTP/1.1
Host: localhost:5000
Cache-Control: no-cache
Postman-Token: f8c5de11-64f3-53e1-faa2-2b6c2d487c5b
```

Before:
image

After:

image

There are also changes to corefxlab (https://github.com/dotnet/corefxlab/compare/davidfowl/optimizations?expand=1) that are related to this change. I haven't done any throughput testing as yet since I need to make sure the tests still pass first but that's the next step.

/cc @cesarbs @halter73 @pakrym  @mikeharder @benaadams


src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
```diff
@@ -984,6 +984,253 @@ private Task WriteSuffix()
...
+            consumed = buffer.Start;
+
+            if (_requestProcessingStatus == RequestProcessingStatus.RequestPending)
+            {
+                ConnectionControl.ResetTimeout(_requestHeadersTimeoutMilliseconds, TimeoutAction.SendTimeoutResponse);
+            }
+
+            _requestProcessingStatus = RequestProcessingStatus.RequestStarted;
+
+            var limitedBuffer = buffer;
```
cesarbs: Just declare it here but put this assignment in an `else` block, avoiding the potentially wasted assignment.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
```diff
@@ -984,6 +984,253 @@ private Task WriteSuffix()
...
+            var versionEnd = versionToEndSpan.IndexOf(ByteCR);
+
+            if (versionEnd == -1)
+            {
+                RejectRequestLine(start, end);
+            }
+
+            if (!versionToEndSpan.GetKnownVersion(out string httpVersion))
+            {
+                httpVersion = new Utf8String(versionToEndSpan.Slice(0, versionEnd)).ToString();
```
cesarbs: Why UTF-8?

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
```diff
@@ -984,6 +984,253 @@ private Task WriteSuffix()
...
+
+            // URIs are always encoded/escaped to ASCII https://tools.ietf.org/html/rfc3986#page-11
+            // Multibyte Internationalized Resource Identifiers (IRIs) are first converted to utf8;
+            // then encoded/escaped to ASCII  https://www.ietf.org/rfc/rfc3987.txt "Mapping of IRIs to URIs"
+            string requestUrlPath;
+            string rawTarget;
+            if (needDecode)
+            {
+                // Read raw target before mutating memory.
+                rawTarget = targetBuffer.GetAsciiString() ?? string.Empty;
```
cesarbs: This should never be null. The target must either be in absolute form, or start with a `/`. cc @natemcmaster who looked into this part of the spec not long ago.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
```diff
@@ -984,6 +984,253 @@ private Task WriteSuffix()
...
+
+                // URI was encoded, unescape and then parse as utf8
+                var pathSpan = pathBuffer;
+                int pathLength = UrlEncoder.Decode(pathSpan, pathSpan);
+                requestUrlPath = new Utf8String(pathSpan.Slice(0, pathLength)).ToString();
+            }
+            else
+            {
+                // URI wasn't encoded, parse as ASCII
+                requestUrlPath = pathBuffer.GetAsciiString() ?? string.Empty;
```
cesarbs: Same here.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
```diff
@@ -1241,6 +1488,191 @@ private bool RequestUrlStartsWithPathBase(string requestUrl, out bool caseMatche
...
+                {
+                    RejectRequest(RequestRejectionReason.TooManyHeaders);
+                }
+
+                // Reset the reader since we're not at the end of headers
+                reader = start;
+
+                // Now parse a single header
+                ReadCursor lineEnd;
+                var limitedBuffer = buffer;
```
cesarbs: Same here about the potentially wasted assignment.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
```diff
@@ -1241,6 +1488,191 @@ private bool RequestUrlStartsWithPathBase(string requestUrl, out bool caseMatche
...
+                        byte* stackBuffer = stackalloc byte[stackAllocLimit];
+                        span = new Span<byte>(stackBuffer, headerBuffer.Length);
+                        headerBuffer.CopyTo(span);
+                    }
+                }
+                else
+                {
+                    // We're not a single span here but we can use pooled arrays to avoid allocations in the rare case
+                    span = new Span<byte>(new byte[headerBuffer.Length]);
+                    headerBuffer.CopyTo(span);
```
cesarbs: Can't you get the value string from multiple buffers? This copying seems wasteful. You could defer it till later and only copy the name into contiguous memory if you need to.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
davidfowl: This isn't new code.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
davidfowl: I'm not confident that doesn't overly complicate the code. The copy is wasteful but I'd be willing to measure how often this happens before doing it. Another option is to use the array pool here (if that indeed makes things better).

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
```diff
@@ -984,6 +984,253 @@ private Task WriteSuffix()
...
+                pathEnd = MinNonZero(spaceIndex, questionMarkIndex);
+                if (questionMarkIndex == -1 && spaceIndex == -1)
+                {
+                    RejectRequestLine(start, end);
+                }
+            }
+
+            var queryString = "";
+            var queryEnd = pathEnd;
+            if (questionMarkIndex != -1)
```
davidfowl: I need to run the tests to see if this even works. I'm pretty sure I botched the logic here 😄

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
cesarbs: Ok, the proper check is still in place: https://github.com/aspnet/KestrelHttpServer/pull/1378/files/391393b79db3c14d8a9f877a3bb719f4cae49fcc#diff-4a391b137e1879654de7fd4febc3db67R1120

Don't do `?? string.Empty`. Assert that `rawTarget != null`.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
```diff
@@ -984,6 +984,253 @@ private Task WriteSuffix()
...
+                {
+                    RejectRequest(RequestRejectionReason.RequestLineTooLong);
+                }
+                else
+                {
+                    return false;
+                }
+            }
+
+            const int stackAllocLimit = 512;
```
cesarbs: Magic numbers... 😝

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
```diff
@@ -1241,6 +1488,191 @@ private bool RequestUrlStartsWithPathBase(string requestUrl, out bool caseMatche
...
+                    {
+                        RejectRequest(RequestRejectionReason.HeadersExceedMaxTotalSize);
+                    }
+                    else
+                    {
+                        return false;
+                    }
+                }
+
+                const int stackAllocLimit = 512;
```
cesarbs: Why 512? I'm thinking of cookies.

davidfowl: What number do you like? 😄

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
```diff
@@ -1241,6 +1488,191 @@ private bool RequestUrlStartsWithPathBase(string requestUrl, out bool caseMatche
...
+                {
+                    // No copies, directly use the one and only span
+                    span = headerBuffer.ToSpan();
+                }
+                else if (headerBuffer.Length < stackAllocLimit)
+                {
+                    unsafe
+                    {
+                        // Multiple buffers and < stackAllocLimit, copy into a stack buffer
+                        byte* stackBuffer = stackalloc byte[stackAllocLimit];
```
benaadams: No need to go full on?

`byte* stackBuffer = stackalloc byte[headerBuffer.Length];`?

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
benaadams: Cookies are fattening... 😢 Can alloc if you are using heavy ones?

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
davidfowl: I believe you but the intent of this PR isn't to fix those things.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
```diff
@@ -984,6 +984,253 @@ private Task WriteSuffix()
...
+
+            const int stackAllocLimit = 512;
+
+            // Move 1 byte past the \r
+            end = limitedBuffer.Move(end, 1);
+            var startLineBuffer = limitedBuffer.Slice(0, end);
+
+            Span<byte> span;
+
+            if (startLineBuffer.IsSingleSpan)
```
pakrym: I don't like this. To common scenario to have ifs everywhere.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
```diff
@@ -984,6 +984,253 @@ private Task WriteSuffix()
...
+            }
+
+            var needDecode = false;
+            var pathBegin = methodEnd + 1;
+            var pathToEndSpan = span.Slice(pathBegin, span.Length - pathBegin);
+            // var pbs = Encoding.UTF8.GetString(pathToEndSpan.ToArray());
+            pathBegin = 0;
+
+            // TODO: IndexOfAny
+            var spaceIndex = pathToEndSpan.IndexOf(ByteSpace);
```
pakrym: I would refactor current IndexOf to have both span and RB overloads

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
davidfowl: Unfortunately it can't be refactored because of the stackalloc.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
davidfowl: ?

pakrym: Split current `IndexOf` methods in two parts, one that can search byte (or any of 3 bytes ) in span, and another that calls it on all spans inside `ReadableBuffer` and returns `ReadCursor`

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
davidfowl: Not sure what you mean. All of the code under here uses limitedBuffer so it needs to be assigned regardless.

davidfowl:
Reopening against dev

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
benaadams: Isn't necessarily `== buffer` though

halter73:
@davidfowl You can change an existing PR's base branch now btw.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
cesarbs: You assign it here then immediately check `buffer.Length >= ServerOptions.Limits.MaxRequestLineSize` and if that's true you reassign it. So every time that condition is true you waste this assignment.

davidfowl:
Shiettttt

pakrym:
@davidfowl ?

PR #1379, opened 2017-02-21
Remove custom BenchmarkDotNet toolchain

cesarbs:
Latest package in the AppVeyor feed contains the new csproj stuff: https://github.com/dotnet/BenchmarkDotNet/issues/300#issuecomment-281411381

Once they release the package to NuGet we can stop using the AppVeyor feed.

cc @adamsitnik

PR #1381, opened 2017-02-21
Fix 'No data found for...' errors in CI test runs

cesarbs:
xUnit complains about empty data even when we have skip conditions. This is because the data is evaluated before we evaluate our skip conditions, an an error happens if the data is empty:

https://github.com/xunit/xunit/blob/1b69df1378cb9d6b589d55ce23ab544458169485/src/xunit.execution/Sdk/Frameworks/TheoryDiscoverer.cs#L242

We could fix this in our Testing package (https://github.com/aspnet/Testing/issues/254), but the fix is ugly. We'd have to override `Discover` in `ConditionalTheoryDiscoverer`, call the base, check that the results contain any `ExecutionErrorTestCase` instances with messages starting with `No data found for` and remove those from the collection.


cesarbs:
@natemcmaster It won't because of https://github.com/aspnet/KestrelHttpServer/pull/1379. And that one won't pass because of this one.

natemcmaster:
YOLO push both.

PR #1384, opened 2017-02-21
Temporary add delay to test to make it past

pakrym:
@halter73


halter73:
Gross :shipit:

cesarbs:
Fix the typos in the commit the message then 🚢 this gross thing 😝

PR #1389, opened 2017-02-23
Avoid unobserved exceptions (dev)

halter73:
- Don't throw from AdaptedPipeline.ReadInputAsync
- Watch for unobserved exceptions in SampleApp


halter73:
@cesarbs @pakrym

PR #1391, opened 2017-02-23
Use pass through connection adapter instead of ssl for max buffer size tests

pakrym:
@halter73


cesarbs:
cc @mikeharder who authored that test.

cesarbs:
What is the reason for the change?

pakrym:
It hangs :) I think it's because we stopped posting `Pause` and `Resume` so when buffer size threshold is hit connection is paused immediately vs having it posted and allowing another read to pass through.

pakrym:
:up::date:

halter73:
:shipit:

PR #1392, opened 2017-02-23
Don't treat canceled reads as end of input

halter73:
@pakrym @cesarbs


halter73:
This fixes the failing RequestsCanBeAbortedMidRead in dev.

cesarbs:
Is the test that broke in the CI a proper test for this, or does this warrant a new test?

cesarbs:
How hard is it to mock `_input` and write a small unit test to verify this behavior?

pakrym:
Looks good, but having tests would be nice.

davidfowl:
Why is appveyor failing?

pakrym:
@davidfowl I think it's race that was fixed by pipelines PR.

halter73:
The test that this fixes is about as good as a functional test is going to get. I will look into adding unit tests for the two affected classes though.

PR #1394, opened 2017-02-23
Use Spans to parse the start line and headers

davidfowl:
- Use `Span<byte>` to parse the start line and headers
- This relies on 2 new APIs in corefxlab so the build will pass until we merge https://github.com/dotnet/corefxlab/pull/1240 (span.GetAsciiString and ReadableBufferReader.Skip)
- The tests pass and the performance results are here https://gist.github.com/davidfowl/c487ad4b49f013b6e9d4fc1ccc84bbf2

/cc @cesarbs @halter73 @pakrym 

The tests all pass (locally at least) with the corefxlab changes.


src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
```diff
@@ -982,6 +982,257 @@ private Task WriteSuffix()
...
+
+            const int stackAllocLimit = 512;
+
+            // Move 1 byte past the \r
+            end = limitedBuffer.Move(end, 1);
+            var startLineBuffer = limitedBuffer.Slice(0, end);
+
+            Span<byte> span;
+
+            if (startLineBuffer.IsSingleSpan)
```
pakrym: Still the same issue with this pattern.. Add a buffer to Frame and reuse it for all parsers.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
```diff
@@ -982,6 +982,257 @@ private Task WriteSuffix()
...
+
+        private int MinNonZero(int v1, int v2, int v3)
+        {
+            v1 = v1 == -1 ? int.MaxValue : v1;
+            v2 = v2 == -1 ? int.MaxValue : v2;
+            v3 = v3 == -1 ? int.MaxValue : v3;
+            return Math.Min(Math.Min(v1, v2), v3);
+        }
+
         public bool TakeStartLine(ReadableBuffer buffer, out ReadCursor consumed, out ReadCursor examined)
```
pakrym: Remove old method

pakrym:
If I remember correctly benchmark we use runs on top of single segment readable buffer, I'd like to see results using multiple buffers (~2k size) where request crosses boundary so it will hit slow paths.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
davidfowl: I'm not making that change as part of this PR. I attempted this earlier and realized that the buffer size would be an issue. We need to have smart logic to use the cached buffer if it was <= the size of the buffers crossing the spans or new up one from scratch if > than the buffer size. I think we should look at that after this change goes in.

davidfowl:
> If I remember correctly benchmark we use runs on top of single segment readable buffer, I'd like to see results using multiple buffers (~2k size) where request crosses boundary so it will hit slow paths.

IMO that's not a blocker for this PR. The goal is to incrementally make changes that are better than the previous ones. I'll gladly keep making more changes and add more benchmarks after but I really want to focus on the bigger ideas that improve performance. Will add more benchmarks once this goes in. We should add those to dev-si as well.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Infrastructure/MemoryPoolIteratorExtensions.cs
```diff
@@ -90,6 +90,19 @@ public static string GetAsciiStringEscaped(this ReadCursor start, ReadCursor end
             return sb.ToString();
         }
 
+        public static string GetAsciiStringEscaped(this Span<byte> span)
```
halter73: @davidfowl GetAsciiStringEscaped needs a maxChars parameter like before. Otherwise the error for an HTTPS connection to and HTTP endpoint that you suggested becomes over eight thousand characters.

PR #1399, opened 2017-02-25
Optimized parsing GET verb and version

KrzysztofCwalina:
### Before
```
                   Method |          Mean |     StdDev | Scaled | Scaled-StdDev |          RPS | Allocated |
------------------------- |-------------- |----------- |------- |-------------- |------------- |---------- |
           ParsePlaintext | 1,079.6836 ns |  8.1507 ns |   1.00 |          0.00 |   926,197.25 |     168 B |
  ParsePipelinedPlaintext |   877.8646 ns |  7.3689 ns |   0.81 |          0.01 | 1,139,127.77 |     168 B |
          ParseLiveAspNet | 3,826.8063 ns | 15.9470 ns |   3.54 |          0.03 |   261,314.51 |   2.02 kB |
 ParsePipelinedLiveAspNet | 3,828.8495 ns | 29.7095 ns |   3.55 |          0.04 |   261,175.06 |   2.02 kB |
             ParseUnicode | 6,870.2153 ns | 33.0606 ns |   6.36 |          0.06 |   145,555.85 |   3.41 kB |
    ParseUnicodePipelined | 6,944.8703 ns | 24.7683 ns |   6.43 |          0.05 |   143,991.17 |   3.41 kB |
```

### After
```
                   Method |          Mean |     StdDev | Scaled | Scaled-StdDev |          RPS | Allocated |
------------------------- |-------------- |----------- |------- |-------------- |------------- |---------- |
           ParsePlaintext | 1,061.8108 ns |  8.9208 ns |   1.00 |          0.00 |   941,787.34 |     168 B |
  ParsePipelinedPlaintext |   865.3721 ns |  6.0690 ns |   0.82 |          0.01 | 1,155,572.27 |     168 B |
          ParseLiveAspNet | 3,882.7252 ns | 85.4559 ns |   3.66 |          0.08 |   257,551.06 |   2.02 kB |
 ParsePipelinedLiveAspNet | 3,866.2318 ns | 21.6853 ns |   3.64 |          0.04 |   258,649.78 |   2.02 kB |
             ParseUnicode | 6,857.4925 ns | 67.3531 ns |   6.46 |          0.08 |   145,825.90 |   3.41 kB |
    ParseUnicodePipelined | 7,009.4501 ns | 39.9254 ns |   6.60 |          0.07 |   142,664.54 |   3.41 kB |
```

Wow, my home machine is faster than the one I have at work! 900K rps vs ~500K

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Infrastructure/MemoryPoolIteratorExtensions.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Infrastructure/MemoryPoolIteratorExtensions.cs
@@ -19,6 +19,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Infrastructure
         private readonly static ulong _httpConnectMethodLong = GetAsciiStringAsLong("CONNECT ");
         private readonly static ulong _httpDeleteMethodLong = GetAsciiStringAsLong("DELETE \0");
         private readonly static ulong _httpGetMethodLong = GetAsciiStringAsLong("GET \0\0\0\0");
+        private readonly static uint _httpGetMethodInt = GetAsciiStringAsInt("GET ");
         private readonly static ulong _httpHeadMethodLong = GetAsciiStringAsLong("HEAD \0\0\0");
         private readonly static ulong _httpPatchMethodLong = GetAsciiStringAsLong("PATCH \0\0");
         private readonly static ulong _httpPostMethodLong = GetAsciiStringAsLong("POST \0\0\0");
@@ -60,6 +61,19 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Infrastructure
                 return *(ulong*)ptr;
             }
         }
+
+        private unsafe static uint GetAsciiStringAsInt(string str)
+        {
+            Debug.Assert(str.Length == 4, "String must be exactly 8 (ASCII) characters long.");
+
+            var bytes = Encoding.ASCII.GetBytes(str);
+
+            fixed (byte* ptr = &bytes[0])
+            {
+                return *(uint*)ptr;
+            }
+        }
+
         private unsafe static ulong GetMaskAsLong(byte[] bytes)
         {
             Debug.Assert(bytes.Length == 8, "Mask must be exactly 8 bytes long.");
@@ -173,27 +187,28 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Infrastructure
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
         public static bool GetKnownMethod(this Span<byte> span, out string knownMethod)
         {
-            knownMethod = null;
-            if (span.Length < sizeof(ulong))
+            if (span.TryRead<uint>(out var possiblyGet))
             {
-                return false;
-            }
-
-            ulong value = span.Read<ulong>();
-            if ((value & _mask4Chars) == _httpGetMethodLong)
-            {
-                knownMethod = HttpMethods.Get;
-                return true;
-            }
-            foreach (var x in _knownMethods)
-            {
-                if ((value & x.Item1) == x.Item2)
+                if (possiblyGet == _httpGetMethodInt)
                 {
-                    knownMethod = x.Item3;
+                    knownMethod = HttpMethods.Get;
                     return true;
                 }
             }
 
+            if (span.TryRead<ulong>(out var value))
+            {
+                foreach (var x in _knownMethods)
+                {
+                    if ((value & x.Item1) == x.Item2)
+                    {
+                        knownMethod = x.Item3;
+                        return true;
+                    }
+                }
+            }
+            
+            knownMethod = null;
             return false;
         }
 
--- a/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Infrastructure/MemoryPoolIteratorExtensions.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Infrastructure/MemoryPoolIteratorExtensions.cs
@@ -1,4 +1,4 @@
-// Copyright (c) .NET Foundation. All rights reserved.
+﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
@@ -19,7 +19,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Infrastructure
         private readonly static ulong _httpConnectMethodLong = GetAsciiStringAsLong("CONNECT ");
         private readonly static ulong _httpDeleteMethodLong = GetAsciiStringAsLong("DELETE \0");
         private readonly static ulong _httpGetMethodLong = GetAsciiStringAsLong("GET \0\0\0\0");
-        private readonly static uint _httpGetMethodInt = GetAsciiStringAsInt("GET ");
+        private const uint _httpGetMethodInt = 542393671; // retun of GetAsciiStringAsInt("GET "); const results in better codegen
         private readonly static ulong _httpHeadMethodLong = GetAsciiStringAsLong("HEAD \0\0\0");
         private readonly static ulong _httpPatchMethodLong = GetAsciiStringAsLong("PATCH \0\0");
         private readonly static ulong _httpPostMethodLong = GetAsciiStringAsLong("POST \0\0\0");
@@ -64,7 +64,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Infrastructure
 
         private unsafe static uint GetAsciiStringAsInt(string str)
         {
-            Debug.Assert(str.Length == 4, "String must be exactly 8 (ASCII) characters long.");
+            Debug.Assert(str.Length == 4, "String must be exactly 4 (ASCII) characters long.");
 
             var bytes = Encoding.ASCII.GetBytes(str);
 
@@ -259,32 +259,30 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Infrastructure
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
         public static bool GetKnownVersion(this Span<byte> span, out string knownVersion)
         {
-            knownVersion = null;
-
-            if (span.Length < sizeof(ulong))
+            if (span.TryRead<ulong>(out var version))
             {
-                return false;
-            }
-
-            var value = span.Read<ulong>();
-            if (value == _http11VersionLong)
-            {
-                knownVersion = Http11Version;
-            }
-            else if (value == _http10VersionLong)
-            {
-                knownVersion = Http10Version;
-            }
-
-            if (knownVersion != null)
-            {
-                if (span[sizeof(ulong)] != (byte)'\r')
+                if (version == _http11VersionLong)
+                {
+                    knownVersion = Http11Version;
+                }
+                else if (version == _http10VersionLong)
+                {
+                    knownVersion = Http10Version;
+                }
+                else
                 {
                     knownVersion = null;
+                    return false;
+                }
+
+                if (span[sizeof(ulong)] == (byte)'\r')
+                {
+                    return true;
                 }
             }
 
-            return knownVersion != null;
+            knownVersion = null;
+            return false;
         }
     }
 }
--- a/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Infrastructure/MemoryPoolIteratorExtensions.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Infrastructure/MemoryPoolIteratorExtensions.cs
@@ -27,8 +27,8 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Infrastructure
         private readonly static ulong _httpOptionsMethodLong = GetAsciiStringAsLong("OPTIONS ");
         private readonly static ulong _httpTraceMethodLong = GetAsciiStringAsLong("TRACE \0\0");
 
-        private readonly static ulong _http10VersionLong = GetAsciiStringAsLong("HTTP/1.0");
-        private readonly static ulong _http11VersionLong = GetAsciiStringAsLong("HTTP/1.1");
+        private const ulong _http10VersionLong = 3471766442030158920; // GetAsciiStringAsLong("HTTP/1.0"); const results in better codegen
+        private const ulong _http11VersionLong = 3543824036068086856; // GetAsciiStringAsLong("HTTP/1.1"); const results in better codegen
 
         private readonly static ulong _mask8Chars = GetMaskAsLong(new byte[] { 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff });
         private readonly static ulong _mask7Chars = GetMaskAsLong(new byte[] { 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0x00 });
--- a/test/Microsoft.AspNetCore.Server.Kestrel.Performance/Program.cs
+++ b/test/Microsoft.AspNetCore.Server.Kestrel.Performance/Program.cs
@@ -40,6 +40,10 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Performance
             {
                 BenchmarkRunner.Run<PipeThroughput>();
             }
+            if (type.HasFlag(BenchmarkType.KnownStrings)) 
+            {
+                BenchmarkRunner.Run<KnownStrings>();
+            }
         }
     }
 
@@ -49,6 +53,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Performance
         RequestParsing = 1,
         Writing = 2,
         Throughput = 4,
+        KnownStrings = 8,
         // add new ones in powers of two - e.g. 2,4,8,16...
 
         All = uint.MaxValue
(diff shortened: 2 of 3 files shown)


KrzysztofCwalina:
cc: @davidfowl, @pakrym

KrzysztofCwalina:
it's strange but the build system is not picking up the newest System.Slices from myget-corfxlab

davidfowl:
@KrzysztofCwalina it takes a while to propagate, we mirror your feeds.

/cc @pranavkm

KrzysztofCwalina:
Please do not merge. I need to dig a bit more into this. Initially I saw a nice perf improvement, but I think it might be because I optimized Read in corfxlab. These Kestrel changes might not be improving anything or could be actually detrimental. 

I don't have time to do it today. I will finish it Monday.

KrzysztofCwalina:
@davidfowl, I convinced myself there is a small improvement here. I think this should be merged.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Infrastructure/MemoryPoolIteratorExtensions.cs
```diff
@@ -60,6 +61,19 @@ private unsafe static ulong GetAsciiStringAsLong(string str)
                 return *(ulong*)ptr;
             }
         }
+
+        private unsafe static uint GetAsciiStringAsInt(string str)
+        {
+            Debug.Assert(str.Length == 4, "String must be exactly 8 (ASCII) characters long.");
```
cesarbs: nit: 4

pakrym:
@KrzysztofCwalina you can write a benchmark just for these methods

halter73:
@KrzysztofCwalina What convinces you?

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Infrastructure/MemoryPoolIteratorExtensions.cs
```diff
@@ -19,6 +19,7 @@ public static class MemoryPoolIteratorExtensions
         private readonly static ulong _httpConnectMethodLong = GetAsciiStringAsLong("CONNECT ");
         private readonly static ulong _httpDeleteMethodLong = GetAsciiStringAsLong("DELETE \0");
         private readonly static ulong _httpGetMethodLong = GetAsciiStringAsLong("GET \0\0\0\0");
```
cesarbs: We can remove the `long`, right?

KrzysztofCwalina:
I did two things:
1. I ran the PlainText benchmark in more controlled eniroment, e.g. not browsing the web while running benchmarks, and the results were more consistent and in favod of the change.
2. I looked at the assembly. The after changes assembly is cleaner. The whole method is quite long, but here is the front of the method (the part that parses succesful GET):

###Before:
```assembly
          knownMethod = null;
00007FFCFC3A3482  push        rdi  
00007FFCFC3A3483  push        rsi  
00007FFCFC3A3484  push        rbp  
00007FFCFC3A3485  push        rbx  
00007FFCFC3A3486  sub         rsp,20h  
00007FFCFC3A348A  mov         rsi,rdx  
00007FFCFC3A348D  xor         eax,eax  
00007FFCFC3A348F  mov         qword ptr [rsi],rax  
00007FFCFC3A3492  cmp         dword ptr [rcx+10h],8  
00007FFCFC3A3496  jge         00007FFCFC3A34A5  
            {
                return false;
00007FFCFC3A3498  xor         eax,eax  
00007FFCFC3A349A  add         rsp,20h  
00007FFCFC3A349E  pop         rbx  
00007FFCFC3A349F  pop         rbp  
00007FFCFC3A34A0  pop         rsi  
00007FFCFC3A34A1  pop         rdi  
00007FFCFC3A34A2  pop         r14  
00007FFCFC3A34A4  ret  
            }

            ulong value = span.Read<ulong>();
00007FFCFC3A34A5  mov         rdi,qword ptr [rcx]  
00007FFCFC3A34A8  mov         rbx,qword ptr [rcx+8]  
00007FFCFC3A34AC  mov         eax,dword ptr [rcx+10h]  
00007FFCFC3A34AF  cmp         eax,8  
00007FFCFC3A34B2  jae         00007FFCFC3A34C1  
00007FFCFC3A34B4  call        00007FFCFC3A3188  
00007FFCFC3A34B9  mov         rcx,rax  
00007FFCFC3A34BC  call        00007FFD5BE7DC40  
00007FFCFC3A34C1  test        rdi,rdi  
00007FFCFC3A34C4  jne         00007FFCFC3A34C8  
00007FFCFC3A34C6  jmp         00007FFCFC3A34CF  
00007FFCFC3A34C8  lea         rcx,[rdi+8]  
00007FFCFC3A34CC  add         rbx,rcx  
00007FFCFC3A34CF  mov         rdi,qword ptr [rbx]  
            if ((value & _mask4Chars) == _httpGetMethodLong)
00007FFCFC3A34D2  mov         rcx,7FFCFC245640h  
00007FFCFC3A34DC  mov         edx,1  
00007FFCFC3A34E1  call        00007FFD5C0B8020  
00007FFCFC3A34E6  mov         rdx,rdi  
00007FFCFC3A34E9  and         rdx,qword ptr [7FFCFC245680h]  
00007FFCFC3A34F0  cmp         rdx,qword ptr [7FFCFC245678h]  
00007FFCFC3A34F7  jne         00007FFCFC3A351E  
            {
                knownMethod = Get;
00007FFCFC3A34F9  mov         rdx,27F313827A8h  
00007FFCFC3A3503  mov         rdx,qword ptr [rdx]  
00007FFCFC3A3506  mov         rcx,rsi  
00007FFCFC3A3509  call        00007FFD5C0B5AD0  
                return true;
00007FFCFC3A350E  mov         eax,1  
00007FFCFC3A3513  add         rsp,20h  
00007FFCFC3A3517  pop         rbx  
00007FFCFC3A3518  pop         rbp  
00007FFCFC3A3519  pop         rsi  
00007FFCFC3A351A  pop         rdi  
00007FFCFC3A351B  pop         r14  
00007FFCFC3A351D  ret  
```
###After:
```assembly
         if (span.TryRead<uint>(out var possiblyGet))
00007FFCFC3A35A2  push        rdi  
00007FFCFC3A35A3  push        rsi  
00007FFCFC3A35A4  push        rbp  
00007FFCFC3A35A5  push        rbx  
00007FFCFC3A35A6  sub         rsp,20h  
00007FFCFC3A35AA  mov         rsi,rcx  
00007FFCFC3A35AD  mov         rdi,rdx  
00007FFCFC3A35B0  mov         rcx,qword ptr [rsi]  
00007FFCFC3A35B3  mov         rdx,qword ptr [rsi+8]  
00007FFCFC3A35B7  mov         eax,dword ptr [rsi+10h]  
00007FFCFC3A35BA  and         eax,0FFFFFFFFh  
00007FFCFC3A35BF  cmp         rax,4  
00007FFCFC3A35C3  jl          00007FFCFC3A3616  
00007FFCFC3A35C5  test        rcx,rcx  
00007FFCFC3A35C8  jne         00007FFCFC3A35CC  
00007FFCFC3A35CA  jmp         00007FFCFC3A35D3  
00007FFCFC3A35CC  add         rcx,8  
00007FFCFC3A35D0  add         rdx,rcx  
00007FFCFC3A35D3  mov         ebx,dword ptr [rdx]  
            {
                if (possiblyGet == _httpGetMethodInt)
00007FFCFC3A35D5  mov         rcx,7FFCFC245640h  
00007FFCFC3A35DF  mov         edx,1  
00007FFCFC3A35E4  call        00007FFD5C0B8020  
00007FFCFC3A35E9  cmp         ebx,dword ptr [7FFCFC245688h]  
00007FFCFC3A35EF  jne         00007FFCFC3A3616  
                {
                    knownMethod = Get;
00007FFCFC3A35F1  mov         rdx,27F313827A8h  
00007FFCFC3A35FB  mov         rdx,qword ptr [rdx]  
00007FFCFC3A35FE  mov         rcx,rdi  
00007FFCFC3A3601  call        00007FFD5C0B5AD0  
                    return true;
00007FFCFC3A3606  mov         eax,1  
00007FFCFC3A360B  add         rsp,20h  
00007FFCFC3A360F  pop         rbx  
00007FFCFC3A3610  pop         rbp  
00007FFCFC3A3611  pop         rsi  
00007FFCFC3A3612  pop         rdi  
00007FFCFC3A3613  pop         r14  
00007FFCFC3A3615  ret  
```

KrzysztofCwalina:
If you still want me to write benchmarks, I can, but to me the codegen difference is pretty convincing. Let me know.

benaadams:
Looking at the asm, the current would work better if `_maskXChars` were consts rather than readonly static conversions from byte arrays (so memory load on use)

cesarbs:
@KrzysztofCwalina Please add the benchmark as requested by @pakrym. While the shorter code is good evidence, we need an actual measurement. E2E benchmarking doesn't catch this change (too much noise).

KrzysztofCwalina:
@benaadams, I changed _httpGetMethodInt to a const and the "if (possiblyGet == _httpGetMethodInt)" block codegened to:
```assembly
00007FFCFCF53473  mov         edx,dword ptr [rax]  
            {
                if (possiblyGet == _httpGetMethodInt)
00007FFCFCF53475  cmp         edx,20544547h  
00007FFCFCF5347B  jne         00007FFCFCF534B6  
                {
                    knownMethod = Get;
```

I update the PR to use const

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Infrastructure/MemoryPoolIteratorExtensions.cs
```diff
@@ -19,6 +19,7 @@ public static class MemoryPoolIteratorExtensions
...
+        const uint _httpGetMethodInt = ‭20544547‬; // returned from GetAsciiStringAsInt("GET "); const results in slightly better codegen
```
benaadams: `20544547‬u`?

benaadams:
Could change the `ulong`s to consts also ;-)

KrzysztofCwalina:
Here are the microbenchmark results:

``` ini

BenchmarkDotNet=v0.10.1, OS=Microsoft Windows NT 6.2.9200.0
Processor=Intel(R) Core(TM) i7-6700K CPU 4.00GHz, ProcessorCount=8
Frequency=3914061 Hz, Resolution=255.4891 ns, Timer=TSC
  [Host]     : Clr 4.0.30319.42000, 64bit RyuJIT-v4.6.1586.0
  DefaultJob : Clr 4.0.30319.42000, 64bit RyuJIT-v4.6.1586.0

Allocated=0 B  

```
 Method |       Mean |    StdDev |
------- |----------- |---------- |
  After | 28.5595 us | 0.0070 us |
 Before | 36.1777 us | 0.1141 us |
```

**Comment by KrzysztofCwalina** [comment]
@benaadams, I will leave it up to the Kestrel guys to decided if they want consts for all the ulongs. They are not on super critical path, and I am not sure if we want to tradeoff maintainability for a couple of instructions.

File: src/Microsoft.AspNetCore.Server.Kestrel/Internal/Infrastructure/MemoryPoolIteratorExtensions.cs (hunk 2 above)
Comment (KrzysztofCwalina) [inline src/Microsoft.AspNetCore.Server.Kestrel/Internal/Infrastructure/MemoryPoolIteratorExtensions.cs]: it's used in GetKnownMethod(this ReadableBuffer begin, out string knownMethod)

**Comment by arespr** [comment]
I have found a function that creates key for lookup into array _knownMethods with extended length to 17.
It's twice faster than foreach (var x in _knownMethods) in microbenchmarks:

https://github.com/arespr/ConsoleApp1/blob/09ac0c1635914eb55b546c1c8b8fcba75f02e4ec/ConsoleApp1/KnownMethods.cs

**Comment by cesarbs** [comment]
@arespr Interesting. We used to have something similar: https://github.com/aspnet/KestrelHttpServer/blob/349af5097778f9df82c8e0efbe62a9877cfbcd68/src/Microsoft.AspNet.Server.Kestrel/Infrastructure/MemoryPoolIterator2Extensions.cs#L47 But the `for` loop turned out to be faster. Can you share your benchmark results?

**Comment by KrzysztofCwalina** [comment]
I think _knownMethods optimizations should be considered/done as a separate issue/PR. The loop is not on the hot path in mainline scenarios.

**Comment by KrzysztofCwalina** [comment]
@davidfowl/@cesarbs, anything else you would like me to do?

**Comment by davidfowl** [comment]
@KrzysztofCwalina where did you put the code for the micro benchmark?

**Comment by KrzysztofCwalina** [comment]
I did not add a microbenchmark to this change. I just wrote it an ran it locally. I can add it if you want, but as I said I think the routine is too small to benchmark with timer-based harness. The results vary from run to run. But your call; just let me know where to add it.

**Comment by cesarbs** [comment]
I'd like to have the benchmarks checked in. Please add one for `GetKnownVersion` too.

**Comment by davidfowl** [comment]
@KrzysztofCwalina add a GetKnownVersion benchmark here https://github.com/aspnet/KestrelHttpServer/tree/dev/test/Microsoft.AspNetCore.Server.Kestrel.Performance

**Comment by KrzysztofCwalina** [comment]
I added microbenchmarks for GetKnownMethod and GetKnownVersion

File: test/Microsoft.AspNetCore.Server.Kestrel.Performance/Microbenchmarks.cs [hunk 4]
```diff
@@ -0,0 +1,38 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using BenchmarkDotNet.Attributes;
+using Microsoft.AspNetCore.Server.Kestrel.Internal.Infrastructure;
+using System;
+using System.Text;
+
+namespace Microsoft.AspNetCore.Server.Kestrel.Performance
+{
+    public class Microbenchmarks
```
cesarbs: Rename it to KnownStrings.

test/Microsoft.AspNetCore.Server.Kestrel.Performance/Microbenchmarks.cs
```diff
@@ -0,0 +1,38 @@
...
+    {
+        static byte[] _method = Encoding.UTF8.GetBytes("GET ");
+        static byte[] _version = Encoding.UTF8.GetBytes("HTTP/1.1\r\n");
+
+        [Benchmark]
+        public string GetKnownMethod()
```
cesarbs: `GetKnownMethod_GET`, or something of the sort, would be a better name.

test/Microsoft.AspNetCore.Server.Kestrel.Performance/Program.cs
```diff
@@ -49,6 +53,7 @@ public enum BenchmarkType : uint
         RequestParsing = 1,
         Writing = 2,
         Throughput = 4,
+        Microbenchmarks = 8,
```
cesarbs: Don't forget to rename it here too.

KrzysztofCwalina:
I fixed the issues in the benchmarks. 
I also unrolled the benchmark loop. @vancem suggested it as a way to minimize the noise stemming from the fact that these routines are very short.

test/Microsoft.AspNetCore.Server.Kestrel.Performance/Microbenchmarks.cs
```diff
@@ -0,0 +1,78 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using BenchmarkDotNet.Attributes;
+using Microsoft.AspNetCore.Server.Kestrel.Internal.Infrastructure;
+using System;
+using System.Text;
```
cesarbs: nit: `System` directives first

test/Microsoft.AspNetCore.Server.Kestrel.Performance/Microbenchmarks.cs
```diff
@@ -0,0 +1,78 @@
...
+                data.GetKnownMethod(out method);
+                len += method.Length;
+                data.GetKnownMethod(out method);
+                len += method.Length;
+                data.GetKnownMethod(out method);
+                len += method.Length;
+                data.GetKnownMethod(out method);
+                len += method.Length;
+            }
+            return len;
```
cesarbs: Where is the return value used?

test/Microsoft.AspNetCore.Server.Kestrel.Performance/Microbenchmarks.cs
```diff
@@ -0,0 +1,78 @@
...
+using System.Text;
+
+namespace Microsoft.AspNetCore.Server.Kestrel.Performance
+{
+    public class KnownStrings
+    {
+        static byte[] _method = Encoding.UTF8.GetBytes("GET ");
+        static byte[] _version = Encoding.UTF8.GetBytes("HTTP/1.1\r\n");
+
+        [Benchmark]
```
benaadams: Since you are actually doing 10k parses per function you want this attribute to be
`[Benchmark(OperationsPerInvoke = 10000)]`
so it knows to divide down the time.

KrzysztofCwalina: thanks. will do.

test/Microsoft.AspNetCore.Server.Kestrel.Performance/Microbenchmarks.cs
KrzysztofCwalina: I thought Benchmark.NET will use it to avoid JIT optimizing it out.

test/Microsoft.AspNetCore.Server.Kestrel.Performance/Microbenchmarks.cs
KrzysztofCwalina: Oh, you guys don't sort it alphabetically. OK, I will update

test/Microsoft.AspNetCore.Server.Kestrel.Performance/Microbenchmarks.cs
cesarbs: Oh, makes sense.

KrzysztofCwalina:
@cesarbs, anything else?

PR #1400, opened 2017-02-25
Fewer passes over the buffer

davidfowl:
- Single pass to find all start line delimiters instead
of calling IndexOf multiple times.
- Use a state machine for header parsing. It's likely that it can be simplified
but this passes all of the existing parsing tests (down to the specific errors)

PS: Still measuring performance impact of this change.

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
@@ -1082,15 +1082,45 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Http
 
             var needDecode = false;
             var pathBegin = methodEnd + 1;
-            var pathToEndSpan = span.Slice(pathBegin, span.Length - pathBegin);
-            pathBegin = 0;
+            var spaceIndex = -1;
+            var questionMarkIndex = -1;
+            var percentageIndex = -1;
+            var pathEnd = -1;
 
-            // TODO: IndexOfAny
-            var spaceIndex = pathToEndSpan.IndexOf(ByteSpace);
-            var questionMarkIndex = pathToEndSpan.IndexOf(ByteQuestionMark);
-            var percentageIndex = pathToEndSpan.IndexOf(BytePercentage);
+            // TODO: State machineify
+            for (int i = pathBegin; i < span.Length; i++)
+            {
+                var ch = span[i];
+                if (spaceIndex == -1 && ch == ByteSpace)
+                {
+                    if (pathEnd == -1)
+                    {
+                        pathEnd = i;
+                    }
 
-            var pathEnd = MinNonZero(spaceIndex, questionMarkIndex, percentageIndex);
+                    spaceIndex = i;
+                }
+
+                if (questionMarkIndex == -1 && ch == ByteQuestionMark)
+                {
+                    if (pathEnd == -1)
+                    {
+                        pathEnd = i;
+                    }
+
+                    questionMarkIndex = i;
+                }
+
+                if (percentageIndex == -1 && ch == BytePercentage)
+                {
+                    if (pathEnd == -1)
+                    {
+                        pathEnd = i;
+                    }
+
+                    percentageIndex = i;
+                }
+            }
 
             if (spaceIndex == -1 && questionMarkIndex == -1 && percentageIndex == -1)
             {
@@ -1099,8 +1129,8 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Http
             else if (percentageIndex != -1)
             {
                 needDecode = true;
-
                 pathEnd = MinNonZero(spaceIndex, questionMarkIndex);
+
                 if (questionMarkIndex == -1 && spaceIndex == -1)
                 {
                     RejectRequestLine(start, end);
@@ -1117,7 +1147,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Http
                     RejectRequestLine(start, end);
                 }
 
-                queryString = pathToEndSpan.Slice(pathEnd, queryEnd - pathEnd).GetAsciiString();
+                queryString = span.Slice(pathEnd, queryEnd - pathEnd).GetAsciiString();
             }
 
             if (pathBegin == pathEnd)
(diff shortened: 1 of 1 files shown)


src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
```diff
@@ -1082,15 +1082,45 @@ public bool TakeStartLine(ReadableBuffer buffer, out ReadCursor consumed, out Re
...
+                    }
+
+                    questionMarkIndex = i;
+                }
+
+                if (percentageIndex == -1 && ch == BytePercentage)
+                {
+                    if (pathEnd == -1)
+                    {
+                        pathEnd = i;
```
cesarbs: This doesn't look right. For `/a%20b` you'd end up with `Path` = `/a`.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
```diff
@@ -1082,15 +1082,45 @@ public bool TakeStartLine(ReadableBuffer buffer, out ReadCursor consumed, out Re
...
+            var questionMarkIndex = -1;
+            var percentageIndex = -1;
+            var pathEnd = -1;
 
-            // TODO: IndexOfAny
-            var spaceIndex = pathToEndSpan.IndexOf(ByteSpace);
-            var questionMarkIndex = pathToEndSpan.IndexOf(ByteQuestionMark);
-            var percentageIndex = pathToEndSpan.IndexOf(BytePercentage);
+            // TODO: State machineify
+            for (int i = pathBegin; i < span.Length; i++)
```
cesarbs: Shouldn't it be `i < pathBegin + span.Length`?

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
```diff
@@ -1126,18 +1156,25 @@ public bool TakeStartLine(ReadableBuffer buffer, out ReadCursor consumed, out Re
             }
 
             var versionBegin = queryEnd + 1;
-            var versionToEndSpan = pathToEndSpan.Slice(versionBegin, pathToEndSpan.Length - versionBegin);
-            versionBegin = 0;
-            var versionEnd = versionToEndSpan.IndexOf(ByteCR);
+            var versionEnd = -1;
+            for (int i = versionBegin; i < span.Length; i++)
```
cesarbs: nit: `var`

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
```diff
@@ -1411,57 +1447,126 @@ public bool TakeMessageHeaders(ReadableBuffer buffer, FrameRequestHeaders reques
...
-                if (endNameIndex == -1)
+                var state = HeaderState.ExpectName;
+                var endNameIndex = -1;
+                var startNameIndex = 0;
+                var endValueIndex = -1;
+                var startValueIndex = -1;
+                var sawWhitespaceInName = false;
+                var previouslyWhitespace = false;
+
+                for (int i = 0; i < span.Length; i++)
```
cesarbs: nit: `var`

davidfowl:
Superseded by https://github.com/aspnet/KestrelHttpServer/pull/1401

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
```diff
@@ -1411,57 +1447,126 @@ public bool TakeMessageHeaders(ReadableBuffer buffer, FrameRequestHeaders reques
...
+                            {
+                                // Mark the first non whitespace char as the start of the
+                                // header value and change the state to expect to the header value
+                                startValueIndex = i;
+                                state = HeaderState.ExpectValue;
+                            }
+                            // If we see a CR then jump to the next state directly
+                            else if (ch == ByteCR)
+                            {
+                                goto case HeaderState.ExpectValue;
```
cesarbs: :hankey:

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
```diff
@@ -1411,57 +1447,126 @@ public bool TakeMessageHeaders(ReadableBuffer buffer, FrameRequestHeaders reques
...
+                                endNameIndex = i;
+                            }
+
+                            if (ch == ByteSpace || ch == ByteTab)
+                            {
+                                sawWhitespaceInName = true;
+                            }
+                            break;
+                        case HeaderState.ExpectWhitespace:
+                            var sws = ch == ByteTab || ch == ByteSpace || ch == ByteCR;
```
cesarbs: nit: rename to something better

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
```diff
@@ -1411,57 +1447,126 @@ public bool TakeMessageHeaders(ReadableBuffer buffer, FrameRequestHeaders reques
...
+                            }
+                            // If we see a CR then jump to the next state directly
+                            else if (ch == ByteCR)
+                            {
+                                goto case HeaderState.ExpectValue;
+                            }
+
+                            break;
+                        case HeaderState.ExpectValue:
+                            var ews = ch == ByteTab || ch == ByteSpace;
```
cesarbs: nit: rename to something better

cesarbs:
@davidfowl lol had just reviewed it

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
davidfowl: if this doesn't look right and all of the tests pass, we have a *big* problem 😄

davidfowl:
@cesarbs I can make some of the changes in the other PR

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
davidfowl: This is irrelevant in the other PR but not that's not right. pathBegin is an index between 0 and span.Length.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
davidfowl: goto!

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
davidfowl: I'm looping over valid indexes up to `span.Length`. `pathBegin` can't be > span.Length since span is 0 based. Unlike ArraySegment, we don't need to add the offset.

PR #1401, opened 2017-02-26
Convert TakeStartLine and TakeMessageHeaders to be state machines

davidfowl:
- Convert TakeStartLine and TakeMessageHeaders to be state machines
- Use pointers to parse byte by byte instead of span indexer to access bytes

Results:

https://gist.github.com/davidfowl/c487ad4b49f013b6e9d4fc1ccc84bbf2#file-results-md

Seems to be the best implementation so far. Still a way to go to get back to 1.2M for pipelined but it looks like progress. 

/cc @benaadams

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
@@ -1029,7 +1029,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Http
             if (startLineBuffer.IsSingleSpan)
             {
                 // No copies, directly use the one and only span
-                span = startLineBuffer.ToSpan();
+                span = startLineBuffer.First.Span;
             }
             else if (startLineBuffer.Length < stackAllocLimit)
             {
@@ -1327,8 +1327,18 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Http
             examined = buffer.End;
 
             var bufferEnd = buffer.End;
-            var bufferLength = buffer.Length;
             var reader = new ReadableBufferReader(buffer);
+            
+            // Make sure the buffer is limited
+            var overLength = false;
+            if (buffer.Length >= _remainingRequestHeadersBytesAllowed)
+            {
+                bufferEnd = buffer.Move(consumed, _remainingRequestHeadersBytesAllowed);
+
+                // If we sliced it means the current buffer bigger than what we're 
+                // allowed to look at
+                overLength = true;
+            }
 
             while (true)
             {
@@ -1374,18 +1384,6 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Http
                 // Reset the reader since we're not at the end of headers
                 reader = start;
 
-                // Now parse a single header
-                var overLength = false;
-
-                if (bufferLength >= _remainingRequestHeadersBytesAllowed)
-                {
-                    bufferEnd = buffer.Move(consumed, _remainingRequestHeadersBytesAllowed);
-
-                    // If we sliced it means the current buffer bigger than what we're 
-                    // allowed to look at
-                    overLength = true;
-                }
-
                 if (ReadCursorOperations.Seek(consumed, bufferEnd, out var lineEnd, ByteLF) == -1)
                 {
                     // We didn't find a \n in the current buffer and we had to slice it so it's an issue
@@ -1412,7 +1410,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Http
                 if (headerBuffer.IsSingleSpan)
                 {
                     // No copies, directly use the one and only span
-                    span = headerBuffer.ToSpan();
+                    span = headerBuffer.First.Span;
                 }
                 else if (headerBuffer.Length < stackAllocLimit)
                 {
@@ -1429,17 +1427,17 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Http
                 }
 
                 var state = HeaderState.ExpectName;
-                var endNameIndex = -1;
-                var startNameIndex = 0;
-                var endValueIndex = -1;
-                var startValueIndex = -1;
-                var sawWhitespaceInName = false;
+                var nameStart = 0;
+                var nameEnd = -1;
+                var valueStart = -1;
+                var valueEnd = -1;
+                var nameHasWhitespace = false;
                 var previouslyWhitespace = false;
+                var headerLineLength = span.Length;
 
                 fixed (byte* data = &span.DangerousGetPinnableReference())
                 {
-                    var length = span.Length;
-                    for (var i = 0; i < length; i++)
+                    for (var i = 0; i < headerLineLength; i++)
                     {
                         var ch = data[i];
 
@@ -1448,18 +1446,18 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Http
                             case HeaderState.ExpectName:
                                 if (ch == ByteColon)
                                 {
-                                    if (sawWhitespaceInName)
+                                    if (nameHasWhitespace)
                                     {
                                         RejectRequest(RequestRejectionReason.WhitespaceIsNotAllowedInHeaderName);
                                     }
 
                                     state = HeaderState.ExpectWhitespace;
-                                    endNameIndex = i;
+                                    nameEnd = i;
                                 }
 
                                 if (ch == ByteSpace || ch == ByteTab)
                                 {
-                                    sawWhitespaceInName = true;
+                                    nameHasWhitespace = true;
                                 }
                                 break;
                             case HeaderState.ExpectWhitespace:
@@ -1470,12 +1468,13 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Http
                                     {
                                         // Mark the first non whitespace char as the start of the
                                         // header value and change the state to expect to the header value
-                                        startValueIndex = i;
+                                        valueStart = i;
                                         state = HeaderState.ExpectValue;
                                     }
                                     // If we see a CR then jump to the next state directly
                                     else if (ch == ByteCR)
                                     {
+                                        state = HeaderState.ExpectValue;
                                         goto case HeaderState.ExpectValue;
                                     }
                                 }
@@ -1490,22 +1489,22 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Http
                                         {
                                             // If we see a whitespace char then maybe it's end of the
                                             // header value
-                                            endValueIndex = i;
+                                            valueEnd = i;
                                         }
                                     }
                                     else if (ch == ByteCR)
                                     {
                                         // If we see a CR and we haven't ever seen whitespace then
                                         // this is the end of the header value
-                                        if (endValueIndex == -1)
+                                        if (valueEnd == -1)
                                         {
-                                            endValueIndex = i;
+                                            valueEnd = i;
                                         }
 
                                         // We never saw a non whitespace character before the CR
-                                        if (startValueIndex == -1)
+                                        if (valueStart == -1)
                                         {
-                                            startValueIndex = endValueIndex;
+                                            valueStart = valueEnd;
                                         }
 
                                         state = HeaderState.ExpectNewLine;
@@ -1513,7 +1512,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Http
                                     else
                                     {
                                         // If we find a non whitespace char that isn't CR then reset the end index
-                                        endValueIndex = -1;
+                                        valueEnd = -1;
                                     }
 
                                     previouslyWhitespace = whitespace;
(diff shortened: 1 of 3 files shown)


src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
```diff
@@ -1082,137 +1079,179 @@ public bool TakeStartLine(ReadableBuffer buffer, out ReadCursor consumed, out Re
...
+                    if (percentageIndex == -1 && ch == BytePercentage)
+                    {
+                        if (pathEnd == -1)
+                        {
+                            pathEnd = i;
+                        }
+
+                        percentageIndex = i;
+                    }
                 }
```
arespr: Try replace loop between lines 1093-1125 for code below.
I tested this piece of code in console application and it's seems to be 40% faster.
`for (int i = pathBegin; i < length; i++)
{
    var ch = data[i];

    switch (ch)
    {
        case ByteSpace:
            if (spaceIndex == -1)
            {
                if (pathEnd == -1)
                {
                    pathEnd = i;
                }

                spaceIndex = i;

                if (questionMarkIndex != -1 && percentageIndex != -1)
                {
                    goto loopExit;
                }
            }
            break;
        case ByteQuestionMark:
            if (questionMarkIndex == -1)
            {
                if (pathEnd == -1)
                {
                    pathEnd = i;
                }

                questionMarkIndex = i;

                if (spaceIndex != -1 && percentageIndex != -1)
                {
                    goto loopExit;
                }
            }
            break;
        case BytePercentage:
            if (percentageIndex == -1)
            {
                if (pathEnd == -1)
                {
                    pathEnd = i;
                }

                percentageIndex = i;

                if (spaceIndex != -1 && questionMarkIndex != -1)
                {
                    goto loopExit;
                }
            }
            break;
    }
}

loopExit:`

davidfowl: I'll try it, what input did you test? How did you come up with the 40% number.

arespr:
I made a mistake in test input data and new algorithm ended earlier when browsing the array. 
This is one of it's advantages. After i changed test data it's seems to be faster by only 20%
in worst case.

davidfowl:
Can you share your benchmark?

arespr:
https://github.com/arespr/ConsoleApp1/blob/master/ConsoleApp1/Program.cs

davidfowl:
@arespr I just ran the microbenchmark for request parsing here:https://github.com/aspnet/KestrelHttpServer/blob/5692f51bf72c82d480a13f9f4282a3e84739e954/test/Microsoft.AspNetCore.Server.Kestrel.Performance/RequestParsing.cs 

The algorithm you posted was worse on my machine for plaintext and plaintext pipelined.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
```diff
@@ -1448,18 +1446,18 @@ public unsafe bool TakeMessageHeaders(ReadableBuffer buffer, FrameRequestHeaders
...
 
                                     state = HeaderState.ExpectWhitespace;
-                                    endNameIndex = i;
+                                    nameEnd = i;
                                 }
 
                                 if (ch == ByteSpace || ch == ByteTab)
                                 {
-                                    sawWhitespaceInName = true;
+                                    nameHasWhitespace = true;
```
pakrym: Why not reject request right here?

davidfowl: Because I didn't want to change the tests 😄

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
```diff
@@ -1049,113 +1045,156 @@ public bool TakeStartLine(ReadableBuffer buffer, out ReadCursor consumed, out Re
...
+                                goto case StartLineState.ExpectPath;
+                            }
+
+                            state = StartLineState.ExpectUnknownMethod;
+                            goto case StartLineState.ExpectUnknownMethod;
+
+                        case StartLineState.ExpectUnknownMethod:
+                            if (ch == ByteSpace)
+                            {
+                                method = span.Slice(0, i).GetAsciiString();
```
pakrym: Move all materialization to the end.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
```diff
@@ -1049,113 +1045,156 @@ public bool TakeStartLine(ReadableBuffer buffer, out ReadCursor consumed, out Re
...
+                                state = StartLineState.ExpectPath;
+                            }
+                            else if (!IsValidTokenChar((char)ch))
+                            {
+                                RejectRequestLine(start, end);
+                            }
+
+                            break;
+                        case StartLineState.ExpectPath:
+                            if (ch == ByteSpace)
```
pakrym: switch?

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
```diff
@@ -1638,5 +1736,26 @@ private enum RequestProcessingStatus
             RequestStarted,
             ResponseStarted
         }
+
+        private enum StartLineState
+        {
+            ExpectKnownMethod,
```
pakrym: Remove expects

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
```diff
@@ -1049,113 +1045,156 @@ public bool TakeStartLine(ReadableBuffer buffer, out ReadCursor consumed, out Re
...
+                                pathEnd = i;
+                                queryStart = i;
+                                state = StartLineState.ExpectQueryString;
+                            }
+                            else if (ch == BytePercentage)
+                            {
+                                needDecode = true;
+                            }
+
+                            if (pathStart == -1)
```
pakrym: Did you think about having another state to capture start and then switch to another to parse the rest?

pakrym: Or set it in known method parsers

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
```diff
@@ -1049,113 +1045,156 @@ public bool TakeStartLine(ReadableBuffer buffer, out ReadCursor consumed, out Re
...
+                            }
+                            break;
+                        case StartLineState.ExpectKnownVersion:
+                            // REVIEW: We don't *need* to slice here but it makes the API
+                            // nicer, slicing should be free :)
+                            if (span.Slice(i).GetKnownVersion(out httpVersion))
+                            {
+                                // Update the index, current char, state and jump directly
+                                // to the next state
+                                i += httpVersion.Length + 1;
```
pakrym: I don't like math here, would much prefer out parameter in `GetKnownVersion`

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
```diff
@@ -1049,113 +1045,156 @@ public bool TakeStartLine(ReadableBuffer buffer, out ReadCursor consumed, out Re
...
+                                ch = data[i];
+                                state = StartLineState.ExpectNewLine;
+
+                                goto case StartLineState.ExpectNewLine;
+                            }
+
+                            versionStart = i;
+                            state = StartLineState.ExpectUnknownVersion;
```
pakrym: ugh.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
davidfowl: I thought about that but it messes with the error handling a bit. I guess I can throw directly here if i == 0.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
davidfowl: switch in a switch!?

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
davidfowl: Also, it's ugly because we have a fast path that does materialization inline so I'm not sure it makes sense here. There would need to be a check for if method == null.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
```diff
@@ -1411,59 +1426,139 @@ public bool TakeMessageHeaders(ReadableBuffer buffer, FrameRequestHeaders reques
...
+                                }
+
+                                if (ch == ByteSpace || ch == ByteTab)
+                                {
+                                    nameHasWhitespace = true;
+                                }
+                                break;
+                            case HeaderState.ExpectWhitespace:
+                                {
+                                    var whitespace = ch == ByteTab || ch == ByteSpace || ch == ByteCR;
```
pakrym: Inline

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
```diff
@@ -1411,59 +1426,139 @@ public bool TakeMessageHeaders(ReadableBuffer buffer, FrameRequestHeaders reques
...
+                                        // Mark the first non whitespace char as the start of the
+                                        // header value and change the state to expect to the header value
+                                        valueStart = i;
+                                        state = HeaderState.ExpectValue;
+                                    }
+                                    // If we see a CR then jump to the next state directly
+                                    else if (ch == ByteCR)
+                                    {
+                                        state = HeaderState.ExpectValue;
+                                        goto case HeaderState.ExpectValue;
```
pakrym: Does it speed things up? If not lets remove it.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
```diff
@@ -1411,59 +1426,139 @@ public bool TakeMessageHeaders(ReadableBuffer buffer, FrameRequestHeaders reques
...
+                                    }
+                                }
+                                break;
+                            case HeaderState.ExpectValue:
+                                {
+                                    var whitespace = ch == ByteTab || ch == ByteSpace;
```
pakrym: Inline

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
davidfowl: It's not for speed, it's for correctness (and to remove duplication).

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
```diff
@@ -1049,113 +1045,156 @@ public bool TakeStartLine(ReadableBuffer buffer, out ReadCursor consumed, out Re
...
+                    var ch = data[i];
+
+                    switch (state)
+                    {
+                        case StartLineState.ExpectKnownMethod:
+                            if (span.GetKnownMethod(out method))
+                            {
+                                // Update the index, current char, state and jump directly
+                                // to the next state
+                                i += method.Length + 1;
```
pakrym: out argument on GetKnownMethod

cesarbs:
Looking good on plaintext E2E:

dev                    | this PR
----------------- | --------
849035.906 RPS | 893540.628

Averages of 10 runs on each.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
```diff
@@ -1049,113 +1045,156 @@ public bool TakeStartLine(ReadableBuffer buffer, out ReadCursor consumed, out Re
...
-            {
-                httpVersion = versionToEndSpan.Slice(0, versionEnd).GetAsciiStringEscaped();
-
-                if (httpVersion == string.Empty)
-                {
-                    RejectRequestLine(start, end);
-                }
-                else
+                var length = span.Length;
+                for (var i = 0; i < length; i++)
```
halter73: Does this ever actually loop? Or did you set this up just so you break? If the latter, the loop is a bit misleading.

halter73: Nvm. Now I see the breaks are for the switch. Still weird to automatically move forward one byte, and otherwise goto.

davidfowl: yes, it's actually parsing byte by byte right now, there just happens to be cases where it jumps forward (fast method and version).

Issue #1390, opened 2017-02-23
Add an option to Kestrel to disable threadpool dispatching


halter73:

tmds:
@halter73 @davidfowl This is now an InternalKestrelServerOptions, is the plan to make it a KestrelServerOptions?

halter73:
I don't think there are currently plans to put it on KestrelServerOptions proper. We are using the internal options in some of our benchmarks, and the improvement seen by disabling threadpool dispatching is negligible in netcoreapp2.0.

tmds:
I did a benchmark using two 8 core Azure machines (D4 I think) a la Techempower plaintext it was using 1.1 runtime, don't know if that makes a big difference.
With threadpool, giving Kestrel 2 threads had best performance.
Without threadpool, giving Kestrel 8 threads had best performance.
The increase was about 7%. Did you see a similar increase? Perhaps my benchmarking was not ok.
I think enabling this depends on the application and for some Transports the benefit may be more than for others.

tmds:
Form the looks of aspnet/benchmarks it seems no dispatching is ran with kestrelthreads equal to 1/2.
For plaintext, this compares "dispatching using all cores" to "no dispatching using 2 cores".
Shouldn't no dispatching be set to the number of cores of the machine to make it a fair comparison?

benaadams:
Threadpool has been improved in 2.0. No-dispatching means the application code blocks I/O. Plaintext reposnse stream is just a memory copy of a cached set of 13 bytes so no real application work (just server work). With threadpool the total I/O thread count being less that the physical core count makes sense as you still need CPU for the threadpool to do the application work.

Not dispatching introduces [Head-of-Line blocking](https://en.wikipedia.org/wiki/Head-of-line_blocking); (which was one of the "failures" of http1.1 pipelining); where a slow request on one connection will stop the processing of all other connections on the same thread; regardless of CPU being free.

That's not to mention if the user application code makes a sync(blocking) Read/Write Task.Wait() or Task.Result, sync SQL Connect/Execute call when you've knocked out an entire Kestrel thread of I/O processing.

Perhaps just dispatching the application code to the thread pool; but keeping the Kestrel server code on the same thread might seem like a good comprise between the two? 

[...]

PR #1405, opened 2017-02-27
Add an option to Kestrel to disable threadpool dispatching

halter73:
Up for review. I don't like awaiting KestrelThread in FrameOfT. Will create PR for feature/dev-si once this is approved.

#1390 @pakrym @cesarbs @davidfowl


src/Microsoft.AspNetCore.Server.Kestrel/Internal/Infrastructure/InlineLoggingThreadPool.cs
```diff
@@ -0,0 +1,77 @@
...
+
+        public void Cancel(TaskCompletionSource<object> tcs)
+        {
+            try
+            {
+                tcs.TrySetCanceled();
+            }
+            catch (Exception e)
+            {
+                _log.LogError(0, e, "LoggingThreadPool.Cancel");
```
pakrym: InlineLoggingThreadPool

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Infrastructure/InlineLoggingThreadPool.cs
```diff
@@ -0,0 +1,77 @@
...
+
+        public void Error(TaskCompletionSource<object> tcs, Exception ex)
+        {
+            try
+            {
+                tcs.TrySetException(ex);
+            }
+            catch (Exception e)
+            {
+                _log.LogError(0, e, "LoggingThreadPool.Error");
```
pakrym: InlineLoggingThreadPool

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Infrastructure/InlineLoggingThreadPool.cs
cesarbs: `nameof` 😄

src/Microsoft.AspNetCore.Server.Kestrel/Internal/InternalKestrelServerOptions.cs
```diff
@@ -0,0 +1,12 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.AspNetCore.Server.Kestrel.Internal
+{
+    public class InternalKestrelServerOptions
```
cesarbs: As small as it is, it should still have its unit test class. Add a unit test for the default value.

halter73: This class isn't intended to make it to 2.0

Issue #1279, opened 2017-01-04
Enforce restrictions on request-target formats


natemcmaster:
We currently accept some malformed requests that are not adherent to the HTTP/1.1 spec.

Example:

 - relative paths `GET ../../ HTTP/1.1` is not a valid format at all. Should be rejected.
 - authority-form `GET www.contoso.com HTTP/1.1` is invalid, but Kestrel will accept anyways.
 - asterisk-form `GET * HTTP/1.1` is invalid, but Kestrel will accept anyways.

Restrictions on request-target that are not currently enforced:

- authority-form are only valid for CONNECT requests https://tools.ietf.org/html/rfc7230#section-5.3.3
- asterisk-form should only by used with OPTIONS requests https://tools.ietf.org/html/rfc7230#section-5.3.4

PR #1406, opened 2017-02-28
Added test for query string without path

davidfowl:
/cc @halter73 @cesarbs


davidfowl:
@cesarbs Turns out this explodes on the current kestrel 😄

natemcmaster:
This is one of the issues I'll address in https://github.com/aspnet/KestrelHttpServer/issues/1279

halter73:
@davidfowl Do you want someone to update the davidfowl/added-bad-target-test branch so this test passes and then merge?

davidfowl:
@halter73 I think it depends on how complicate the fix it. Goal was to reduce number of changes to this branch unless necessary. I found this while writing the state machine for dev. @natemcmaster claims that it'll be fixed with other scheduled work so I'm not sure if there's a short term thing we should do here.

davidfowl:
Closing for now

PR #1408, opened 2017-02-28
Add an option to Kestrel to disable threadpool dispatching

halter73:


halter73:
Same as #1405 but fixes feature/dev-si

#1390 @pakrym @cesarbs @davidfowl

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Infrastructure/InlineLoggingThreadPool.cs
```diff
@@ -0,0 +1,78 @@
...
+                _log.LogError(0, e, "InlineLoggingThreadPool.Error");
+            }
+        }
+
+        public void UnsafeRun(WaitCallback action, object state)
+        {
+            action(state);
+        }
+
+        public void Schedule(Action action)
```
davidfowl: Remove

Issue #1393, opened 2017-02-23
Add more HttpParsing unit tests


davidfowl:
Trying to rewrite the http parser is annoying because we don't have any dedicated tests that catch all of the error cases. Right now I'm in a state where the unit tests pass but functional tests fail "sometimes" and the benchmarks fail sometimes.

davidfowl:
I was wrong...

cesarbs:
Will work on those this week.

PR #1409, opened 2017-02-28
Improve HTTP parsing tests (#1393)

cesarbs:
#1393

- Add several more test cases
- Share data between functional and unit tests


src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
```diff
@@ -1129,6 +1129,12 @@ public unsafe bool TakeStartLine(ReadableBuffer buffer, out ReadCursor consumed,
                             }
                             else if (ch == BytePercentage)
                             {
+                                if (pathStart == -1)
```
davidfowl: This is the `GET % HTTP/1.1`  scenario right?

cesarbs: Yeah

PR #1414, opened 2017-02-28
Add IHttpParser interface

pakrym:


src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
```diff
@@ -378,7 +374,8 @@ public void Reset()
             _manuallySetRequestAbortToken = null;
             _abortedCts = null;
 
-            _remainingRequestHeadersBytesAllowed = ServerOptions.Limits.MaxRequestHeadersTotalSize;
+            // Alow to bytes for \r\n after headers
```
cesarbs: nit: allow two

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/IHttpParser.cs
```diff
@@ -0,0 +1,11 @@
+using System.IO.Pipelines;
+
+namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Http
+{
+    public interface IHttpParser
+    {
+        bool ParseStartLine<T>(T handler, ReadableBuffer buffer, out ReadCursor consumed, out ReadCursor examined) where T : IHttpStartLineHandler;
```
cesarbs: nit: the correct name is "request line", so let's use the right one here :)

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/IHttpStartLineHandler.cs
```diff
@@ -0,0 +1,9 @@
+using System;
+
+namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Http
+{
+    public interface IHttpStartLineHandler
+    {
+        void OnStartLine(HttpMethod method, HttpVersion version, Span<byte> target, Span<byte> path, Span<byte> query, Span<byte> customMethod);
```
cesarbs: "Request line" here too.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
```diff
@@ -1001,305 +993,17 @@ public unsafe bool TakeStartLine(ReadableBuffer buffer, out ReadCursor consumed,
             var overLength = false;
             if (buffer.Length >= ServerOptions.Limits.MaxRequestLineSize)
             {
-                bufferEnd = buffer.Move(start, ServerOptions.Limits.MaxRequestLineSize);
-
+                buffer = buffer.Slice(start, ServerOptions.Limits.MaxRequestLineSize);
```
cesarbs: Have you considered handling limits in the parser implementation itself? Our current parser does this upfront, but what about a parser that keeps track of the count as it sees new bytes?

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/KestrelHttpParser.cs
```diff
@@ -0,0 +1,541 @@
...
+                    var ch = data[i];
+
+                    switch (state)
+                    {
+                        case StartLineState.KnownMethod:
+                            if (span.GetKnownMethod(out method, out var methodLenght))
+                            {
+                                // Update the index, current char, state and jump directly
+                                // to the next state
+                                i += methodLenght + 1;
```
cesarbs: nit: Length

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/IHttpStartLineHandler.cs
cesarbs: What is the advantage of doing this string -> enum conversions for methods? There should be no cost for known methods since we're using the same string refs everywhere. If you keep it a string as before, the API looks much cleaner. Having `method` and `customMethod` in the call (and so far from each other 😝) is kinda crummy.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
davidfowl: Why?

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
```diff
@@ -1334,282 +1038,31 @@ private bool RequestUrlStartsWithPathBase(string requestUrl, out bool caseMatche
...
-            var bufferEnd = buffer.End;
-            var reader = new ReadableBufferReader(buffer);
-            
             // Make sure the buffer is limited
-            var overLength = false;
+            bool overLength = false;
             if (buffer.Length >= _remainingRequestHeadersBytesAllowed)
             {
-                bufferEnd = buffer.Move(consumed, _remainingRequestHeadersBytesAllowed);
+                buffer = buffer.Slice(0, _remainingRequestHeadersBytesAllowed);
```
davidfowl: buffer.Slice(buffer.Start, _remainingRequestHeadersBytesAllowed);

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
```diff
@@ -1750,25 +1203,81 @@ private enum RequestProcessingStatus
...
+                Method = MemoryPoolIteratorExtensions.MethodToString(method) ?? String.Empty;
+            }
+            else
+            {
+                Method = customMethod.GetAsciiString() ?? string.Empty;
+            }
+
+            QueryString = query.GetAsciiString() ?? string.Empty;
+            RawTarget = rawTarget;
+            HttpVersion = MemoryPoolIteratorExtensions.VersionToString(version);
```
davidfowl: Move this method to another type. `HttpUtilities`

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/KestrelHttpParser.cs
```diff
@@ -0,0 +1,546 @@
+using System;
+using System.IO.Pipelines;
+using Microsoft.AspNetCore.Server.Kestrel.Internal.Infrastructure;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Http
+{
+    public class KestrelHttpParser: IHttpParser
```
davidfowl: nit: formatting, space between `r` and `:`

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Infrastructure/MemoryPoolIteratorExtensions.cs
```diff
@@ -281,8 +228,50 @@ public static bool GetKnownVersion(this Span<byte> span, out string knownVersion
                 }
             }
 
-            knownVersion = null;
+            knownVersion = HttpVersion.Unset;
```
davidfowl: Unknown

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/IHttpParser.cs
```diff
@@ -0,0 +1,11 @@
+using System.IO.Pipelines;
+
+namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Http
+{
+    public interface IHttpParser
```
davidfowl: Since http2 is coming do we want to call this something more specific?

davidfowl: I think the parser needs to Reset

pakrym: In current design parser is singleton, let's not complicate it for now

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Infrastructure/MemoryPoolIteratorExtensions.cs
```diff
@@ -281,8 +228,50 @@ public static bool GetKnownVersion(this Span<byte> span, out string knownVersion
...
+                    return Http10Version;
+                    break;
+                case HttpVersion.Http11:
+                    return Http11Version;
+                    break;
+                default:
+                    throw new ArgumentOutOfRangeException(nameof(httpVersion), httpVersion, null);
+            }
+        }
+        public static string MethodToString(HttpMethod method)
```
KrzysztofCwalina: If this is called frequently, you might want to create an array of strings, cast the enum value to an int, and then use the int to index into the array.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/IHttpHeadersHandler.cs
```diff
@@ -0,0 +1,9 @@
+using System;
+
+namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Http
+{
+    public interface IHttpHeadersHandler
```
KrzysztofCwalina: An interface call per header might be more expensive than we want to pay. Do we really need this to be abstracted? I think a parser abstraction makes sense, but it seems to me like a separate abstraction for handling headers (one by one) might be an overkill.

benaadams: `IHttpHeadersHandler` and `IHttpRequestLineHandler` are used as a generic constraint for `IHttpParser` should just be direct call in use?

KrzysztofCwalina: Ah, ok. thanks.

benaadams: As aside shame you can't mark interfaces as constraint only; would save accidentally boxing worries with structs

pakrym: You can ship an analyzer along with attribute.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/KestrelHttpParser.cs
```diff
@@ -0,0 +1,549 @@
...
+        // byte types don't have a data type annotation so we pre-cast them; to avoid in-place casts
+        private const byte ByteCR = (byte)'\r';
+        private const byte ByteLF = (byte)'\n';
+        private const byte ByteColon = (byte)':';
+        private const byte ByteSpace = (byte)' ';
+        private const byte ByteTab = (byte)'\t';
+        private const byte ByteQuestionMark = (byte)'?';
+        private const byte BytePercentage = (byte)'%';
+
+        private enum HeaderState
```
davidfowl: Move these to the bottom of the file.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/KestrelHttpParser.cs
```diff
@@ -0,0 +1,549 @@
...
+        }
+
+        private void RejectRequestLine(Span<byte> span)
+        {
+            const int MaxRequestLineError = 32;
+            RejectRequest(RequestRejectionReason.InvalidRequestLine,
+                Log.IsEnabled(LogLevel.Information) ? span.GetAsciiStringEscaped(MaxRequestLineError) : string.Empty);
+        }
+
+        IKestrelTrace Log { get; }
```
davidfowl: Move to the top and add private

PR #1415, opened 2017-03-01
Make the IHttpParser per frame and add a reset

davidfowl:
- Made the IHttpParser a per frame object so state can be stored
across method calls and parses.
- Added HttpParserFactory to ServiceContext

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
@@ -381,7 +381,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Http
             _responseBytesWritten = 0;
 
             // When testing parser can be null
-            _parser?.Reset();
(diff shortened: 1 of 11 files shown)


src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
```diff
@@ -379,6 +379,9 @@ public void Reset()
             _requestHeadersParsed = 0;
 
             _responseBytesWritten = 0;
+
+            // When testing parser can be null
+            _parser?.Reset();
```
halter73: Why not automatically reset when done with headers? Dispose I get, but this? YAGI.

halter73: Alright just keep it @davidfowl

halter73: But remove `?.`

PR #1417, opened 2017-03-02
Loop over bytes inside states of parser state machine

pakrym:
@davidfowl @halter73


KrzysztofCwalina:
This is a very nice change, but it looks like we now don't need the switch/case statement. The code is either sequential or doing goto's.

PR #1421, opened 2017-03-02
Single span fast paths

davidfowl:
This change adds single span fast paths for header and start line parsing. The downside is that the header parsing's outer loop is duplicated with the multi buffer path (we should look at fixing that). Parsing individual headers is shared between both routines.

/cc @benaadams

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/KestrelHttpParser.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/KestrelHttpParser.cs
@@ -2,6 +2,7 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.Buffers;
 using System.IO.Pipelines;
 using Microsoft.AspNetCore.Server.Kestrel.Internal.Infrastructure;
 using Microsoft.Extensions.Logging;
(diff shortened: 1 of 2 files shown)


pakrym:
Perf?

davidfowl:
https://gist.githubusercontent.com/davidfowl/c487ad4b49f013b6e9d4fc1ccc84bbf2/raw/9013b7b5573a7a014b73bffc63e732a2476efa37/Azurevm.md

pakrym:
@davidfowl where does improvement come from?

davidfowl:
So I made a bunch of commits in this PR:

- Added a fast path for single span in the start line parsing
- Added a fast path for single span header parsing
- Changed the out header loop to be pointer based (instead of slicing)

Each of those changes gave a bump in throughput. Headers was the biggest win though.

pakrym:
I had this idea where we can parse headers without Seeking and Slicing at all, in a single state machine, so I'm a bit reluctant to separate methods.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/KestrelHttpParser.cs
```diff
@@ -473,13 +384,244 @@ public KestrelHttpParser(IKestrelTrace log)
...
+            fixed (byte* data = &headersSpan.DangerousGetPinnableReference())
+            {
+                while (true)
+                {
+                    if (remaining < 2)
+                    {
+                        return false;
+                    }
+
+                    var ch1 = data[index];
```
pakrym: Why do you have this, and RBR checking the same thing?

davidfowl:
> I had this idea where we can parse headers without Seeking and Slicing at all, in a single state machine, so I'm a bit reluctant to separate methods.

You can still always rewrite the thing (like you did the original state machine). So far this gives the biggest win since the initial change.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/KestrelHttpParser.cs
davidfowl: I don't understand, this method is a fork. The outer loop is duplicated and written against spans here.

pakrym: Oh, now I get it. Makes me sad. I feel this over complicates the main parser making it hard to work on it.

PR #1423, opened 2017-03-02
Add alocations column (dev-si)

benaadams:
Was wondering why it wasn't output allocations as strings are definitely allocated.

/cc @davidfowl


adamsitnik:
@benaadams MemoryDiagnoser is now [disabled by default](https://github.com/dotnet/BenchmarkDotNet/issues/369). This is, unfortunately, a breaking change. Sorry for that!

benaadams:
@adamsitnik for a time I was impressed it wasn't allocating; before deciding that couldn't be true 😄

PR #1426, opened 2017-03-02
Add more microbenchmarks

cesarbs:
- Separate `KestrelHttpParser` benchmarks for request line and headers.
- Frame overhead benchmarks with "no-op" parser

### DIFF
--- a/test/Microsoft.AspNetCore.Server.Kestrel.Performance/FrameParsingOverhead.cs
+++ b/test/Microsoft.AspNetCore.Server.Kestrel.Performance/FrameParsingOverhead.cs
@@ -1,0 +1,135 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.IO.Pipelines;
+using System.Text;
+using BenchmarkDotNet.Attributes;
+using Microsoft.AspNetCore.Server.Kestrel.Internal.Http;
+using Microsoft.AspNetCore.Testing;
+
+namespace Microsoft.AspNetCore.Server.Kestrel.Performance
+{
+    [Config(typeof(CoreConfig))]
+    public class FrameParsingOverhead
+    {
+        private const int InnerLoopCount = 512;
+
+        public ReadableBuffer _buffer;
+        public Frame<object> _frame;
+
+        [Setup]
+        public void Setup()
+        {
+            var connectionContext = new MockConnection(new KestrelServerOptions());
+            connectionContext.ListenerContext.ServiceContext.HttpParserFactory = frame => NullParser.Instance;
+
+            _frame = new Frame<object>(application: null, context: connectionContext);
+        }
+
+        [Benchmark(Baseline = true, OperationsPerInvoke = InnerLoopCount)]
+        public void FrameOverheadTotal()
+        {
+            for (var i = 0; i < InnerLoopCount; i++)
+            {
+                ParseRequest();
+            }
+        }
+
+        [Benchmark(OperationsPerInvoke = InnerLoopCount)]
+        public void FrameOverheadRequestLine()
+        {
+            for (var i = 0; i < InnerLoopCount; i++)
+            {
+                ParseRequestLine();
+            }
+        }
+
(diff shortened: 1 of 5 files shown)


test/Microsoft.AspNetCore.Server.Kestrel.Performance/KestrelHttpParserRequestHeaders.cs
```diff
@@ -0,0 +1,153 @@
...
+        private void ParseData(ReadableBuffer buffer)
+        {
+            _parser.ParseHeaders(this, buffer, out var consumed, out var examined, out var consumedBytes);
+        }
+
+        public void OnHeader(Span<byte> name, Span<byte> value)
+        {
+        }
+
+        private class NullLog : IKestrelTrace
```
pakrym: It should be fine to pass null if you are not expecting reject request to be called.

test/Microsoft.AspNetCore.Server.Kestrel.Performance/RequestHeadersParsing.cs
```diff
@@ -0,0 +1,108 @@
...
+using System.Linq;
+using System.Text;
+using BenchmarkDotNet.Attributes;
+using Microsoft.AspNetCore.Server.Kestrel.Internal.Http;
+using Microsoft.AspNetCore.Testing;
+
+namespace Microsoft.AspNetCore.Server.Kestrel.Performance
+{
+    [Config(typeof(CoreConfig))]
+    public class RequestHeadersParsing
```
pakrym: I'm not sure if we need these benchmarks. I would prefer having one that only shows frame overhead without a parser. /cc @halter73

cesarbs: I want to be able to measure things separately.

I'll add benchmarks for frame overhead.

cesarbs:
I've undone some of the changes. Left split benchmarks for the parser only.

test/Microsoft.AspNetCore.Server.Kestrel.Performance/FrameOverhead.cs
```diff
@@ -0,0 +1,132 @@
...
+using System.IO.Pipelines;
+using System.Text;
+using BenchmarkDotNet.Attributes;
+using Microsoft.AspNetCore.Server.Kestrel.Internal.Http;
+using Microsoft.AspNetCore.Testing;
+
+namespace Microsoft.AspNetCore.Server.Kestrel.Performance
+{
+    [Config(typeof(CoreConfig))]
+    public class FrameOverhead
```
halter73: FrameParsingOverhead.

test/Microsoft.AspNetCore.Server.Kestrel.Performance/KestrelHttpParserRequestHeaders.cs
```diff
@@ -0,0 +1,84 @@
...
+using System.Linq;
+using System.Text;
+using BenchmarkDotNet.Attributes;
+using Microsoft.AspNetCore.Server.Kestrel.Internal.Http;
+
+namespace Microsoft.AspNetCore.Server.Kestrel.Performance
+{
+    [Config(typeof(CoreConfig))]
+
+    public class KestrelHttpParserRequestHeaders : IHttpHeadersHandler
```
halter73: Remove this class

test/Microsoft.AspNetCore.Server.Kestrel.Performance/KestrelHttpParserRequestLine.cs
```diff
@@ -0,0 +1,99 @@
...
+using System;
+using System.IO.Pipelines;
+using System.Text;
+using BenchmarkDotNet.Attributes;
+using Microsoft.AspNetCore.Server.Kestrel.Internal.Http;
+
+namespace Microsoft.AspNetCore.Server.Kestrel.Performance
+{
+    [Config(typeof(CoreConfig))]
+    public class KestrelHttpParserRequestLine : IHttpRequestLineHandler
```
halter73: Remove this class

Issue #1434, opened 2017-03-03
Kestrel binds to ipv4 by default if no addresses specified


davidfowl:
It should probably bind to both "localhost" by default.

PR #1436, opened 2017-03-03
Enable default server address test

JunTaoLuo:
https://github.com/aspnet/Hosting/issues/917 has now been merged. So enabling this test. Will merge to feature/dev-si once approved.

### DIFF
--- a/test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/AddressRegistrationTests.cs
+++ b/test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/AddressRegistrationTests.cs
@@ -14,6 +14,7 @@ using Microsoft.AspNetCore.Hosting;
 using Microsoft.AspNetCore.Hosting.Server.Features;
 using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.Http.Extensions;
+using Microsoft.AspNetCore.Server.Kestrel.Internal.Infrastructure;
 using Microsoft.AspNetCore.Testing;
 using Microsoft.AspNetCore.Testing.xunit;
 using Microsoft.Extensions.DependencyInjection;
@@ -129,7 +130,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.FunctionalTests
             }
         }
 
-        [ConditionalFact(Skip = "Waiting on https://github.com/aspnet/Hosting/issues/917")]
+        [ConditionalFact]
         [PortSupportedCondition(5000)]
         public async Task DefaultsToPort5000()
         {
@@ -147,13 +148,14 @@ namespace Microsoft.AspNetCore.Server.Kestrel.FunctionalTests
             {
                 host.Start();
 
-                var debugLog = testLogger.Messages.Single(log => log.LogLevel == LogLevel.Debug);
-                Assert.True(debugLog.Message.Contains("default"));
+                Assert.Equal(5000, host.GetPort());
+                Assert.Single(testLogger.Messages, log => log.LogLevel == LogLevel.Debug &&
+                    string.Equals($"No listening endpoints were configured. Binding to {Constants.DefaultIPEndPoint} by default.",
+                    log.Message, StringComparison.Ordinal));
 
-                foreach (var testUrl in new[] { "http://127.0.0.1:5000", "http://localhost:5000" })
+                foreach (var testUrl in new[] { "http://127.0.0.1:5000", /* "http://[::1]:5000" */})
                 {
-                    var response = await HttpClientSlim.GetStringAsync(testUrl);
-                    Assert.Equal(new Uri(testUrl).ToString(), response);
+                    Assert.Equal(new Uri(testUrl).ToString(), await HttpClientSlim.GetStringAsync(testUrl));
                 }
             }
         }


test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/AddressRegistrationTests.cs
```diff
@@ -147,7 +147,8 @@ public class AddressRegistrationTests
             {
                 host.Start();
 
-                var debugLog = testLogger.Messages.Single(log => log.LogLevel == LogLevel.Debug);
+                var debugLog = testLogger.Messages.Single(log => log.LogLevel == LogLevel.Debug &&
+                    !log.Message.StartsWith("hosting", StringComparison.OrdinalIgnoreCase));
```
JunTaoLuo: Need to filter out two hosting logs "Hosting starting" and "Hosting started".

cesarbs: Do a case-sensitive comparison. You want to be as specific as possible here.

cesarbs:
Some thoughts:

- If we're verifying log messages, we should actually assert the exact string we're looking for.
- We have an extension method for `WebHostBuilder` called `GetPort()`. You could assert that in the test too.
- Having the test hit `localhost` makes the test imprecise. What is it resolving to? If the default address is `127.0.0.1:5000`, hit that. If `[::1]:5000` is also a default, hit that too. Hitting `localhost` is not testing anything relevant.

I know those things were already there, but since you're at it... 😄

halter73:
I agree with @cesarbs about testing both 127.0.0.1:5000 and [::1]:5000 though.

JunTaoLuo:
The changes in https://github.com/aspnet/KestrelHttpServer/pull/1312/files seems to only add the ipv4 by default when no address is configured. I'll need to address https://github.com/aspnet/KestrelHttpServer/issues/1434 and bind to both by default. I'll adapt this PR and actually change the behaviour.

halter73:
Right, I forgot that the change to hosting was going to break that.

It's slightly trickier to fix than it looks since we need to special case not throwing in KestrelServer.Start if binding to only one of 127.0.0.1:5000 and [::1]:5000 fails (like on Azure Web Apps). We also need to make sure the logs aren't too scary when the binding fails.

@JunTaoLuo Merge this as is.

JunTaoLuo:
I see okay then I'll just merge the current changes.

JunTaoLuo:
:up::date: @cesarbs

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/AddressRegistrationTests.cs
```diff
@@ -147,10 +148,13 @@ public class AddressRegistrationTests
             {
                 host.Start();
 
-                var debugLog = testLogger.Messages.Single(log => log.LogLevel == LogLevel.Debug);
-                Assert.True(debugLog.Message.Contains("default"));
+                Assert.Equal(5000, host.GetPort());
+
+                var debugLog = testLogger.Messages.Single(log => log.LogLevel == LogLevel.Debug &&
+                    !log.Message.StartsWith("Hosting start", StringComparison.Ordinal)); // Filter out "Hosting starting" and "Hosting started"
+                Assert.True(string.Equals(debugLog.Message, $"No listening endpoints were configured. Binding to {Constants.DefaultIPEndPoint} by default.", StringComparison.Ordinal));
```
cesarbs: Can make this simpler by doing

```csharp
Assert.Equal(
    $"No listening endpoints were configured. Binding to {Constants.DefaultIPEndPoint} by default.",
    testLogger.Messages.SingleOrDefault(log => log.LogLevel == LogLevel.Debug && !log.Message.StartsWith("Hosting start", StringComparison.Ordinal)));
```

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/AddressRegistrationTests.cs
```diff
@@ -149,15 +149,13 @@ public class AddressRegistrationTests
                 host.Start();
 
                 Assert.Equal(5000, host.GetPort());
-
-                var debugLog = testLogger.Messages.Single(log => log.LogLevel == LogLevel.Debug &&
-                    !log.Message.StartsWith("Hosting start", StringComparison.Ordinal)); // Filter out "Hosting starting" and "Hosting started"
-                Assert.True(string.Equals(debugLog.Message, $"No listening endpoints were configured. Binding to {Constants.DefaultIPEndPoint} by default.", StringComparison.Ordinal));
+                Assert.Equal($"No listening endpoints were configured. Binding to {Constants.DefaultIPEndPoint} by default.",
```
cesarbs: Even simpler (sorry, should have realized this earlier):

```csharp
Assert.NotNull(testLogger.Messages.SingleOrDefault(log =>
    log.LogLevel == LogLevel.Debug &&
    log.Message.Equals($"No listening endpoints were configured. Binding to {Constants.DefaultIPEndPoint} by default.", StringComparison.Ordinal)));
```

You get your `StringComparison.Orginal` back :stuck_out_tongue_closed_eyes:

JunTaoLuo: That depends on the original intention of the test. I read it as, there is only one message from Kestrel and it's the binding to default address test. With this change, it becomes, there's is only one binding to default address message but I don't care if there are other messages from Kestrel. Functionally, there's no difference since right now there is only one message on startup.

cesarbs: All we care about is checking that the message is logged. Doesn't matter if there are other messages in there.

BTW, xUnit has `Assert.Single` which makes this even shorter 😄

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/AddressRegistrationTests.cs
```diff
@@ -147,13 +148,14 @@ public class AddressRegistrationTests
...
 
-                var debugLog = testLogger.Messages.Single(log => log.LogLevel == LogLevel.Debug);
-                Assert.True(debugLog.Message.Contains("default"));
+                Assert.Equal(5000, host.GetPort());
+                Assert.Single(testLogger.Messages.Where(log => log.LogLevel == LogLevel.Debug &&
+                    string.Equals($"No listening endpoints were configured. Binding to {Constants.DefaultIPEndPoint} by default.",
+                    log.Message, StringComparison.Ordinal)));
 
-                foreach (var testUrl in new[] { "http://127.0.0.1:5000", "http://localhost:5000" })
+                foreach (var testUrl in new[] { "http://127.0.0.1:5000", /* "http://[::1]:5000" */})
```
cesarbs: Don't leave IPv6 commented out if it's currently not part of the default binding. We can add it if/when it becomes the default.

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/AddressRegistrationTests.cs
```diff
@@ -147,13 +148,14 @@ public class AddressRegistrationTests
             {
                 host.Start();
 
-                var debugLog = testLogger.Messages.Single(log => log.LogLevel == LogLevel.Debug);
-                Assert.True(debugLog.Message.Contains("default"));
+                Assert.Equal(5000, host.GetPort());
+                Assert.Single(testLogger.Messages.Where(log => log.LogLevel == LogLevel.Debug &&
```
cesarbs: Use the overload that takes a collection and a predicate, then you don't need the `Where`.

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/AddressRegistrationTests.cs
JunTaoLuo: We need to add it back soon. Similar to https://github.com/aspnet/KestrelHttpServer/commit/4533383612fccc324337528f50ce324d7fc10d27.

cesarbs: When is soon?

JunTaoLuo: I'm working on it now

PR #1437, opened 2017-03-03
Turn generic into concrete template

adamjez:


adamjez:
I accidentally hit enter when viewing that pull request

PR #1445, opened 2017-03-04
Use ascii decoding routine that disallows null chars

davidfowl:
- GetAsciiString() in newer corefxlab builds allows 0 (which is a valid
ascii char). To avoid future regressions, GetAsciiStringNonNullCharacters()
was added and used in place of GetAsciiString() when interpreting the
request line and headers

/cc @benaadams

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
@@ -1228,7 +1228,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Http
             if (needDecode)
             {
                 // Read raw target before mutating memory.
-                rawTarget = target.GetAsciiString() ?? string.Empty;
+                rawTarget = target.GetAsciiStringNonNullCharacters() ?? string.Empty;
 
                 // URI was encoded, unescape and then parse as utf8
                 int pathLength = UrlEncoder.Decode(path, path);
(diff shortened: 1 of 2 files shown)


src/Microsoft.AspNetCore.Server.Kestrel/Internal/Infrastructure/HttpUtilities.cs
```diff
@@ -94,6 +94,28 @@ private unsafe static ulong GetMaskAsLong(byte[] bytes)
             }
         }
 
+        public unsafe static string GetAsciiStringNonNullCharacters(this Span<byte> span)
+        {
+            if (span.IsEmpty)
+            {
+                return null;
```
benaadams: return `string.Empty` here as always doing `?? string.Empty` in the caller?

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
```diff
@@ -1228,7 +1228,7 @@ public void OnStartLine(HttpMethod method, HttpVersion version, Span<byte> targe
             if (needDecode)
             {
                 // Read raw target before mutating memory.
-                rawTarget = target.GetAsciiString() ?? string.Empty;
+                rawTarget = target.GetAsciiStringNonNullCharacters() ?? string.Empty;
```
benaadams: Return `string.Empty` from the call then drop the null check?

davidfowl:
Merging to help unblock mirror

PR #1446, opened 2017-03-05
SocketOutput WriteReq

benaadams:
* Extract SocketOutput WriteReq from WriteReq and simplify
* Reduce parameter passing (less on stack passing)
* List indexer for clearing rather than heavier List enumerator
* Resolve this pointer in Action to simplify functions (drop `req.`)
* Move some functions that were all `Self.` into SocketOutput


davidfowl:
We're moving SocketOutput to pipes, closing this out

PR #1451, opened 2017-03-06
Add ipv6 loopback address by default

JunTaoLuo:
#1434

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Infrastructure/Constants.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Infrastructure/Constants.cs
@@ -17,7 +17,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Infrastructure
         /// <summary>
         /// The IPEndPoint Kestrel will bind to if nothing else is specified.
         /// </summary>
-        public static readonly IPEndPoint DefaultIPEndPoint = new IPEndPoint(IPAddress.Loopback, 5000);
+        public static readonly string DefaultServerAddress = "http://localhost:5000";
 
         /// <summary>
         /// Prefix of host name used to specify Unix sockets in the configuration.
--- a/src/Microsoft.AspNetCore.Server.Kestrel/KestrelServer.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/KestrelServer.cs
@@ -142,8 +142,16 @@ namespace Microsoft.AspNetCore.Server.Kestrel
                 }
                 else if (!hasListenOptions && !hasServerAddresses)
                 {
-                    _logger.LogDebug($"No listening endpoints were configured. Binding to {Constants.DefaultIPEndPoint} by default.");
-                    listenOptions.Add(new ListenOptions(Constants.DefaultIPEndPoint));
+                    _logger.LogDebug($"No listening endpoints were configured. Binding to {Constants.DefaultServerAddress} by default.");
+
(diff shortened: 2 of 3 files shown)


src/Microsoft.AspNetCore.Server.Kestrel/Internal/Infrastructure/Constants.cs
```diff
@@ -17,7 +17,7 @@ internal static class Constants
         /// <summary>
         /// The IPEndPoint Kestrel will bind to if nothing else is specified.
         /// </summary>
-        public static readonly IPEndPoint DefaultIPEndPoint = new IPEndPoint(IPAddress.Loopback, 5000);
+        public static readonly string DefaultServerAddress = "http://localhost:5000";
```
JunTaoLuo: Might want to pre-parse this and store as `ServerAddress`

src/Microsoft.AspNetCore.Server.Kestrel/KestrelServer.cs
```diff
@@ -142,8 +142,16 @@ public void Start<TContext>(IHttpApplication<TContext> application)
...
+                    _logger.LogDebug($"No listening endpoints were configured. Binding to {Constants.DefaultServerAddress} by default.");
+
+                    // "localhost" for both IPv4 and IPv6 can't be represented as an IPEndPoint.
+                    StartLocalhost(engine, ServerAddress.FromUrl(Constants.DefaultServerAddress));
+
+                    // If StartLocalhost doesn't throw, there is at least one listener.
+                    // The port cannot change for "localhost".
+                    _serverAddresses.Addresses.Add(Constants.DefaultServerAddress);
+
+                    return;
```
JunTaoLuo: Don't really need this since the rest of the method will no-op but it's also safe to return early.

cesarbs:
nit: reference issue # in commit message.

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/AddressRegistrationTests.cs
```diff
@@ -132,16 +132,29 @@ public class AddressRegistrationTests
...
-        public async Task DefaultsToPort5000()
+        public Task DefaultsServerAddress_BindsToIPv4()
+        {
+            return RegisterDefaultServerAddresses_Success(new[] { "http://127.0.0.1:5000" });
+        }
+
+        [ConditionalFact]
+        [IPv6SupportedCondition]
+        [PortSupportedCondition(5000)]
+        public Task DefaultsServerAddress_BindsToIPv6()
```
cesarbs: What happens if one of the interfaces is already bound to port 5000?

JunTaoLuo:
Added more tests to address a portion of https://github.com/aspnet/KestrelHttpServer/issues/1452. I will update the commit message when I squash. Will also update feature/dev-si when done

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/AddressRegistrationTests.cs
JunTaoLuo: https://github.com/aspnet/KestrelHttpServer/issues/1452 was filed for this.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Infrastructure/Constants.cs
JunTaoLuo: Spoke offline. The string here is sufficient.

Issue #1416, opened 2017-03-01
Move http parser tests to be against the abstraction instead of Frame


davidfowl:
/cc @pakrym @cesarbs

davidfowl:
We should do the multi buffer tests as part of this change.

PR #1459, opened 2017-03-07
Convert HTTP parsing FrameTests to IHttpParser tests (#1416)

cesarbs:
Also fixes an issue in `KestrelHttpParser` where `"Header: \r\n"` is parsed with a value of `" "` instead of `""`.

Pending (will be addressed in a subsequent changes):

- Use mock parser in remaining `Frame` tests
- Parameterize parser type so tests can be run on multiple parsers
- Split test data into multiple spans

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/KestrelHttpParser.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/KestrelHttpParser.cs
@@ -459,7 +459,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Http
             }
 
             // Ignore end whitespace
-            for (; valueEnd > valueStart; valueEnd--)
+            for (; valueEnd >= valueStart; valueEnd--)
             {
                 var ch = headerLine[valueEnd];
                 if (ch != ByteTab && ch != ByteSpace)
--- a/test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/BadHttpRequestTests.cs
+++ b/test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/BadHttpRequestTests.cs
@@ -1,9 +1,7 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
-using System;
 using System.Collections.Generic;
-using System.Linq;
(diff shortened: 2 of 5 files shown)


src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/KestrelHttpParser.cs
```diff
@@ -459,7 +459,7 @@ private static unsafe int FindEndOfName(byte* headerLine, int length)
             }
 
             // Ignore end whitespace
-            for (; valueEnd > valueStart; valueEnd--)
+            for (; valueEnd >= valueStart; valueEnd--)
```
cesarbs: This is the empty value fix. cc @benaadams

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/BadHttpRequestTests.cs
```diff
@@ -127,25 +125,34 @@ public Task BadRequestIfContentLengthInvalid(string contentLength)
...
+            get
             {
-                requestLine,
-                "Invalid request line."
-            }));
+                var data = new TheoryData<string, string>();
+
+                foreach (var requestLine in HttpParsingData.RequestLineInvalidData)
+                {
+
```
davidfowl: nit: space

test/Microsoft.AspNetCore.Server.KestrelTests/HttpParserTests.cs
```diff
@@ -0,0 +1,383 @@
...
+                    It.IsAny<HttpVersion>(),
+                    It.IsAny<Span<byte>>(),
+                    It.IsAny<Span<byte>>(),
+                    It.IsAny<Span<byte>>(),
+                    It.IsAny<Span<byte>>()))
+                .Callback<HttpMethod, HttpVersion, Span<byte>, Span<byte>, Span<byte>, Span<byte>>((method, version, target, path, query, customMethod) =>
+                {
+                    parsedMethod = method != HttpMethod.Custom ? HttpUtilities.MethodToString(method) : customMethod.DecodeAscii();
+                    parsedVersion = HttpUtilities.VersionToString(version);
+                    parsedRawTarget = target.DecodeAscii();
```
davidfowl: I would switch to our internal `GetAsciiStringNonNullChars` helper

Issue #1460, opened 2017-03-07
Parameterize parser type in HttpParserTests


cesarbs:

cesarbs:
@davidfowl Do we still want this?

davidfowl:
Nope

PR #1461, opened 2017-03-08
Parameterize parser type in HttpParserTests (#1460)

cesarbs:


test/Microsoft.AspNetCore.Server.KestrelTests/HttpParserTests.cs
```diff
@@ -17,11 +17,14 @@
...
 namespace Microsoft.AspNetCore.Server.KestrelTests
 {
-    public class HttpParserTests
+    public abstract class HttpParserTests<TParser>
+        where TParser : IHttpParser
     {
         // Returns true when all headers parsed
         // Return false otherwise
 
+        protected abstract IHttpParser CreateParser(object state);
```
khellang: `TParser`? Why `object state`?

cesarbs:
Not doing this now.

PR #1462, opened 2017-03-08
Fix CopyTo implementation in benchmark.

davidfowl:
See https://github.com/dotnet/corefx/issues/16840. Once we switch over, I'll change it back.


pakrym:
`#if NETCOREAPP1.1` ?

davidfowl:
The person that makes the benchmarks run on .NET Core 2.0 can add that 😄

PR #1466, opened 2017-03-08
Remove dependency on Utf8String

davidfowl:
- Also remove System.IO.Pipelines.Text.Primitives

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
@@ -6,7 +6,6 @@ using System.Collections.Generic;
 using System.Diagnostics;
 using System.IO;
 using System.IO.Pipelines;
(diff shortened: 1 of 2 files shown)


cesarbs:
Did you run benchmarks to see how this affects the unicode code path? I'm not opposed to the change but I'd like to know what's the impact for percent-encoded paths.

davidfowl:
> Did you run benchmarks to see how this affects the unicode code path? I'm not opposed to the change but I'd like to know what's the impact for percent-encoded paths.

`Utf8String` isn't baked. So we need to remove all dependencies on it, that's the higher order bit.

davidfowl:
It should be on par with what we did before:

https://github.com/aspnet/KestrelHttpServer/blob/6c8fa020992b9f4e77bf74bff8422cf46a8a2ad5/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Infrastructure/MemoryPoolIterator.cs#L1251

PR #1469, opened 2017-03-08
Log rejected request targets

cesarbs:
Instead of logging only `Invalid request line`.

A similar change is not necessary for bad chars in custom methods, since this already handled. Same thing for the query string, since we throw on bad chars when getting the raw target string. But I still added frame tests to verify methods and query strings, since the order in which we do things there could change.

~~Note: currently based on `cesarbs/single-bad-header-log`, will rebase to `dev` once that one is merged.~~

PR #1463, opened 2017-03-08
Speed up ParseRequestLine

benaadams:
dev
```
               Method |          Mean |     StdDev | Scaled |          RPS | Allocated |
--------------------- |-------------- |----------- |------- |------------- |---------- |
 PlaintextTechEmpower |   531.1228 ns |  5.7545 ns |   1.00 | 1,882,803.68 |     112 B |
           LiveAspNet |   917.0677 ns |  9.0826 ns |   1.73 | 1,090,432.07 |     112 B |
              Unicode | 1,300.7374 ns | 18.2239 ns |   2.45 |   768,794.70 |     112 B |
```
pr
```
               Method |          Mean |     StdDev | Scaled |          RPS | Allocated |
--------------------- |-------------- |----------- |------- |------------- |---------- |
 PlaintextTechEmpower |   513.2275 ns |  6.5310 ns |   1.00 | 1,948,453.72 |     112 B |
           LiveAspNet |   909.4809 ns | 10.0160 ns |   1.77 | 1,099,528.33 |     112 B |
              Unicode | 1,283.9817 ns | 13.0742 ns |   2.50 |   778,827.32 |     112 B |
```

PR #1470, opened 2017-03-09
Handle absolute, asterisk, and authority-form request targets

### LINKED ISSUE #666: 500 when abs_path present in http_URL - 3.2.2 http URL

### LINKED ISSUE #1279: Enforce restrictions on request-target formats

natemcmaster:
Improves compliance with RFC 7230 on the expected handling of requests that have URI or asterisk in the request target.

This means rejecting asterisk requests that are not OPTIONS and rejecting authority-form requests that are not CONNECT. "http 405 method not allowed" response requires you set "Allow: " with acceptable methods

This also means the server will handle the path and query on targets with absolute URIs as request-targets.

Resolves #666 
Resolves #1279

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Kestrel/BadHttpRequestException.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/BadHttpRequestException.cs
@@ -4,18 +4,31 @@
 using System.IO;
 using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.Server.Kestrel.Internal.Http;
+using Microsoft.AspNetCore.Server.Kestrel.Internal.Infrastructure;
+using Microsoft.Extensions.Primitives;
 
 namespace Microsoft.AspNetCore.Server.Kestrel
 {
     public sealed class BadHttpRequestException : IOException
     {
         private BadHttpRequestException(string message, int statusCode)
+            : this(message, statusCode, null)
+        { }
+
+        private BadHttpRequestException(string message, int statusCode, HttpMethod? requiredMethod)
             : base(message)
         {
             StatusCode = statusCode;
+
+            if (requiredMethod.HasValue)
+            {
+                AllowedHeader = HttpUtilities.MethodToString(requiredMethod.Value);
+            }
         }
 
         internal int StatusCode { get; }
+
+        internal StringValues AllowedHeader { get; }
 
         internal static BadHttpRequestException GetException(RequestRejectionReason reason)
         {
@@ -60,6 +73,12 @@ namespace Microsoft.AspNetCore.Server.Kestrel
                     break;
                 case RequestRejectionReason.RequestTimeout:
                     ex = new BadHttpRequestException("Request timed out.", StatusCodes.Status408RequestTimeout);
+                    break;
+                case RequestRejectionReason.OptionsMethodRequired:
+                    ex = new BadHttpRequestException("Method not allowed.", StatusCodes.Status405MethodNotAllowed, HttpMethod.Options);
+                    break;
+                case RequestRejectionReason.ConnectMethodRequired:
+                    ex = new BadHttpRequestException("Method not allowed.", StatusCodes.Status405MethodNotAllowed, HttpMethod.Connect);
                     break;
                 default:
                     ex = new BadHttpRequestException("Bad request.", StatusCodes.Status400BadRequest);
--- a/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
@@ -11,7 +11,6 @@ using System.Net;
 using System.Runtime.CompilerServices;
 using System.Text;
 using System.Text.Encodings.Web.Utf8;
-using System.Text.Utf8;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Http;
@@ -27,6 +26,8 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Http
 {
     public abstract partial class Frame : IFrameControl, IHttpRequestLineHandler, IHttpHeadersHandler
     {
+        private const byte ByteAsterisk = (byte)'*';
+        private const byte ByteForwardSlash = (byte)'/';
         private const byte BytePercentage = (byte)'%';
 
         private static readonly ArraySegment<byte> _endChunkedResponseBytes = CreateAsciiByteArraySegment("0\r\n\r\n");
@@ -38,6 +39,9 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Http
         private static readonly byte[] _bytesHttpVersion11 = Encoding.ASCII.GetBytes("HTTP/1.1 ");
         private static readonly byte[] _bytesEndHeaders = Encoding.ASCII.GetBytes("\r\n\r\n");
         private static readonly byte[] _bytesServer = Encoding.ASCII.GetBytes("\r\nServer: Kestrel");
+
+        private const string EmptyPath = "/";
+        private const string Asterisk = "*";
 
         private readonly object _onStartingSync = new Object();
         private readonly object _onCompletedSync = new Object();
@@ -818,7 +822,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Http
                 // that should take precedence.
                 if (_requestRejectedException != null)
                 {
-                    SetErrorResponseHeaders(statusCode: _requestRejectedException.StatusCode);
+                    SetErrorResponseException(_requestRejectedException);
                 }
                 else
                 {
@@ -1077,7 +1081,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Http
             {
                 buffer = buffer.Slice(buffer.Start, _remainingRequestHeadersBytesAllowed);
 
-                // If we sliced it means the current buffer bigger than what we're 
+                // If we sliced it means the current buffer bigger than what we're
                 // allowed to look at
                 overLength = true;
             }
@@ -1124,6 +1128,16 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Http
 
                 // 500 Internal Server Error
                 SetErrorResponseHeaders(statusCode: StatusCodes.Status500InternalServerError);
+            }
+        }
+
+        private void SetErrorResponseException(BadHttpRequestException ex)
+        {
+            SetErrorResponseHeaders(ex.StatusCode);
+
+            if (!StringValues.IsNullOrEmpty(ex.AllowedHeader))
+            {
+                FrameResponseHeaders.HeaderAllow = ex.AllowedHeader;
             }
         }
 
@@ -1179,6 +1193,15 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Http
             throw BadHttpRequestException.GetException(reason, value);
         }
 
+        private void RejectRequestLine(Span<byte> requestLine)
+        {
+            Debug.Assert(Log.IsEnabled(LogLevel.Information) == true, "Use RejectRequest instead to improve inlining when log is disabled");
+
+            const int MaxRequestLineError = 32;
+            var line = requestLine.GetAsciiStringEscaped(MaxRequestLineError);
+            throw BadHttpRequestException.GetException(RequestRejectionReason.InvalidRequestLine, line);
+        }
+
         public void SetBadRequestState(RequestRejectionReason reason)
         {
             SetBadRequestState(BadHttpRequestException.GetException(reason));
@@ -1190,7 +1213,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Http
 
             if (!HasResponseStarted)
             {
-                SetErrorResponseHeaders(ex.StatusCode);
+                SetErrorResponseException(ex);
             }
 
             _keepAlive = false;
@@ -1216,8 +1239,51 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Http
             Log.ApplicationError(ConnectionId, ex);
         }
 
-        public void OnStartLine(HttpMethod method, HttpVersion version, Span<byte> target, Span<byte> path, Span<byte> query, Span<byte> customMethod, bool pathEncoded)
+        public void OnStartLine(HttpMethod method, HttpVersion version, Span<byte> target, Span<byte> path, Span<byte> query, Span<byte> customMethod, Span<byte> line, bool pathEncoded)
         {
+            Debug.Assert(target.Length != 0, "Request target must be non-zero length");
+
+            var ch = target[0];
+            if (ch == ByteForwardSlash)
+            {
+                // origin-form.
+                // The most common form of request-target.
+                // https://tools.ietf.org/html/rfc7230#section-5.3.1
+                OnOriginFormTarget(method, version, target, path, query, customMethod, pathEncoded);
+            }
+            else if (ch == ByteAsterisk && target.Length == 1)
+            {
+                OnAsteriskFormTarget(method);
+            }
+            else if (target.GetKnownHttpScheme(out var scheme))
+            {
+                OnAbsoluteFormTarget(target, query, line);
+            }
+            else
+            {
+                // Assume anything else is considered authority form.
+                // FYI: this should be an edge case. This should only happen when
+                // a client mistakenly things this server is a proxy server.
+
+                OnAuthorityFormTarget(method, target, line);
+            }
+
+            Method = method != HttpMethod.Custom
+                ? HttpUtilities.MethodToString(method) ?? string.Empty
+                : customMethod.GetAsciiStringNonNullCharacters();
+            HttpVersion = HttpUtilities.VersionToString(version);
+
+            Debug.Assert(RawTarget != null, "RawTarget was not set");
+            Debug.Assert(Method != null, "Method was not set");
+            Debug.Assert(Path != null, "Path was not set");
+            Debug.Assert(QueryString != "QueryString was not set");
+            Debug.Assert(HttpVersion != "HttpVersion was not set");
+        }
+
+        private void OnOriginFormTarget(HttpMethod method, HttpVersion version, Span<byte> target, Span<byte> path, Span<byte> query, Span<byte> customMethod, bool pathEncoded)
+        {
+            Debug.Assert(target[0] == ByteForwardSlash, "Should only be called when path starts with /");
+
             // URIs are always encoded/escaped to ASCII https://tools.ietf.org/html/rfc3986#page-11
             // Multibyte Internationalized Resource Identifiers (IRIs) are first converted to utf8;
             // then encoded/escaped to ASCII  https://www.ietf.org/rfc/rfc3987.txt "Mapping of IRIs to URIs"
@@ -1249,34 +1315,111 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Http
                 }
             }
 
-            var normalizedTarget = PathNormalizer.RemoveDotSegments(requestUrlPath);
-            if (method != HttpMethod.Custom)
-            {
-                Method = HttpUtilities.MethodToString(method) ?? string.Empty;
-            }
-            else
-            {
-                Method = customMethod.GetAsciiStringNonNullCharacters();
-            }
-
             QueryString = query.GetAsciiStringNonNullCharacters();
             RawTarget = rawTarget;
-            HttpVersion = HttpUtilities.VersionToString(version);
+            SetNormalizedPath(requestUrlPath);
+        }
 
+        private void OnAuthorityFormTarget(HttpMethod method, Span<byte> target, Span<byte> line)
+        {
+            // TODO Validate that target is a correct host[:port] string.
+            // Reject as 400 if not. This is just a quick scan for invalid characters
+            // but doesn't check that the target fully matches the URI spec.
+            for (var i = 0; i < target.Length; i++)
+            {
+                var ch = target[i];
+                if (!UriUtilities.IsValidAuthorityCharacter(ch))
+                {
+                    if (Log.IsEnabled(LogLevel.Information))
+                    {
+                        RejectRequestLine(line);
+                    }
+
+                    throw BadHttpRequestException.GetException(RequestRejectionReason.InvalidRequestLine);
+                }
+            }
+
+            // The authority-form of request-target is only used for CONNECT
+            // requests (https://tools.ietf.org/html/rfc7231#section-4.3.6).
+            if (method != HttpMethod.Connect)
+            {
+                RejectRequest(RequestRejectionReason.ConnectMethodRequired);
+            }
+
+            // When making a CONNECT request to establish a tunnel through one or
+            // more proxies, a client MUST send only the target URI's authority
+            // component(excluding any userinfo and its "@" delimiter) as the
+            // request - target.For example,
+            //
+            //  CONNECT www.example.com:80 HTTP/1.1
+            //
+            // Allowed characters in the 'host + port' section of authority.
+            // See https://tools.ietf.org/html/rfc3986#section-3.2
+
+            RawTarget = target.GetAsciiStringNonNullCharacters();
+            Path = string.Empty;
+            PathBase = string.Empty;
+            QueryString = string.Empty;
+        }
+
+        private void OnAsteriskFormTarget(HttpMethod method)
+        {
+            // The asterisk-form of request-target is only used for a server-wide
+            // OPTIONS request (https://tools.ietf.org/html/rfc7231#section-4.3.7).
+            if (method != HttpMethod.Options)
+            {
+                RejectRequest(RequestRejectionReason.OptionsMethodRequired);
+            }
+
+            RawTarget = Asterisk;
+            Path = string.Empty;
+            PathBase = string.Empty;
+            QueryString = string.Empty;
+        }
+
+        private void OnAbsoluteFormTarget(Span<byte> target, Span<byte> query, Span<byte> line)
+        {
+            // absolute-form
+            // https://tools.ietf.org/html/rfc7230#section-5.3.2
+
+            // This code should be the edge-case.
+
+            // From the spec:
+            //    a server MUST accept the absolute-form in requests, even though
+            //    HTTP/1.1 clients will only send them in requests to proxies.
+
+            RawTarget = target.GetAsciiStringNonNullCharacters();
+
+            // Validation of absolute URIs is slow, but clients
+            // should not be sending this form anyways, so perf optimization
+            // not high priority
+
+            if (!Uri.TryCreate(RawTarget, UriKind.Absolute, out var uri))
+            {
+                if (Log.IsEnabled(LogLevel.Information))
+                {
+                    RejectRequestLine(line);
+                }
+
+                throw BadHttpRequestException.GetException(RequestRejectionReason.InvalidRequestLine);
+            }
+
+            SetNormalizedPath(uri.LocalPath);
+            // don't use uri.Query because we need the unescaped version
+            QueryString = query.GetAsciiStringNonNullCharacters();
+        }
+
+        private void SetNormalizedPath(string requestPath)
+        {
+            var normalizedTarget = PathNormalizer.RemoveDotSegments(requestPath);
             if (RequestUrlStartsWithPathBase(normalizedTarget, out bool caseMatches))
             {
                 PathBase = caseMatches ? _pathBase : normalizedTarget.Substring(0, _pathBase.Length);
                 Path = normalizedTarget.Substring(_pathBase.Length);
             }
-            else if (rawTarget[0] == '/') // check rawTarget since normalizedTarget can be "" or "/" after dot segment removal
-            {
-                Path = normalizedTarget;
-            }
             else
             {
-                Path = string.Empty;
-                PathBase = string.Empty;
-                QueryString = string.Empty;
+                Path = normalizedTarget;
             }
         }
 
--- a/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/HttpScheme.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/HttpScheme.cs
@@ -1,0 +1,12 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Http
+{
+    public enum HttpScheme
+    {
+        Unknown = -1,
+        Http = 0,
+        Https = 1
+    }
+}
--- a/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/KestrelHttpParser.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/KestrelHttpParser.cs
@@ -157,7 +157,9 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Http
                 RejectRequestLine(data, length);
             }
 
-            handler.OnStartLine(method, httpVersion, targetBuffer, pathBuffer, query, customMethod, pathEncoded);
+            var line = new Span<byte>(data, length);
+
+            handler.OnStartLine(method, httpVersion, targetBuffer, pathBuffer, query, customMethod, line, pathEncoded);
         }
 
         public unsafe bool ParseHeaders<T>(T handler, ReadableBuffer buffer, out ReadCursor consumed, out ReadCursor examined, out int consumedBytes) where T : IHttpHeadersHandler
@@ -341,7 +343,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Http
             // Skip colon from value start
             var valueStart = nameEnd + 1;
             // Ignore start whitespace
-            for(; valueStart < valueEnd; valueStart++)
+            for (; valueStart < valueEnd; valueStart++)
             {
                 var ch = headerLine[valueStart];
                 if (ch != ByteTab && ch != ByteSpace && ch != ByteCR)
@@ -409,7 +411,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Http
                 }
             }
             return false;
-        found:
+            found:
             return true;
         }
 
--- a/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Infrastructure/HttpUtilities.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Infrastructure/HttpUtilities.cs
@@ -15,7 +15,13 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Infrastructure
         public const string Http10Version = "HTTP/1.0";
         public const string Http11Version = "HTTP/1.1";
 
+        public const string HttpUriScheme = "http://";
+        public const string HttpsUriScheme = "https://";
+
         // readonly primitive statics can be Jit'd to consts https://github.com/dotnet/coreclr/issues/1079
+
+        private readonly static ulong _httpSchemeLong = GetAsciiStringAsLong(HttpUriScheme + "\0");
+        private readonly static ulong _httpsSchemeLong = GetAsciiStringAsLong(HttpsUriScheme);
         private readonly static ulong _httpConnectMethodLong = GetAsciiStringAsLong("CONNECT ");
         private readonly static ulong _httpDeleteMethodLong = GetAsciiStringAsLong("DELETE \0");
         private const uint _httpGetMethodInt = 542393671; // retun of GetAsciiStringAsInt("GET "); const results in better codegen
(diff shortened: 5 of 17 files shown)


natemcmaster:
Will follow up with comparison benchmarks when they finish running.

davidfowl:
PERFFFF

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
```diff
@@ -1250,34 +1317,103 @@ public void OnStartLine(HttpMethod method, HttpVersion version, Span<byte> targe
...
+            SetNormalizedPath(requestUrlPath);
+        }
+
+
+        private void OnAuthorityFormTarget(HttpMethod method, Span<byte> target, Span<byte> line)
+        {
+            // TODO Validate that target is a correct host[:port] string.
+            // Reject as 400 if not. This is just a quick scan for invalid characters
+            // but doesn't check that the target fully matches the URI spec.
+            for (var i = 0; i < target.Length; i++)
```
davidfowl: I fear repeated scanning of strings can be bad for performance numbers. I think we'll end up moving more validation into the parser.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
```diff
@@ -1250,34 +1317,103 @@ public void OnStartLine(HttpMethod method, HttpVersion version, Span<byte> targe
...
+            {
+                var ch = (char)target[i];
+                if (!char.IsLetterOrDigit(ch) && ch != ':' && ch != '.')
```
davidfowl: Instead of using IsLetterOrDigit, can you use a subset of this:

https://github.com/aspnet/KestrelHttpServer/blob/dev/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/KestrelHttpParser.cs#L505

We should move this a HttpUtilities

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
natemcmaster: Last time we talked about this, we agreed we we're okay with sub-optimal performance on edge cases. This code shouldn't be hit unless we have a bad client or malformed requests.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
```diff
@@ -1250,34 +1317,103 @@ public void OnStartLine(HttpMethod method, HttpVersion version, Span<byte> targe
...
+                {
+                    if (Log.IsEnabled(LogLevel.Information))
```
davidfowl: Why are we checking IsEnabled all over the code before exceptions now?

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
natemcmaster: No. The "IsValidTokenChar" method includes characters not allowed in the request target for authority-form, like "*".

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/KestrelHttpParser.cs
```diff
@@ -162,6 +185,10 @@ public KestrelHttpParser(IKestrelTrace log)
                                     RejectRequestLine(span);
                                 }
                             }
+                            else if (ch == 0)
+                            {
+                                RejectRequestLine(span);
```
davidfowl: Why are we checking this here?

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
natemcmaster: https://github.com/aspnet/KestrelHttpServer/pull/1469#issuecomment-285161010

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/KestrelHttpParser.cs
natemcmaster: Fail early when possible. \0 is never acceptable on the request line.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
davidfowl: Yea, I'm not saying we should use it.  I said a subset into HttpUtilities.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
davidfowl: > Last time we talked about this

How long ago was that?

I'll wait for the benchmarks.

natemcmaster:
**PR**
```
                        Method |                                                          ParserType |      Mean |    StdErr |    StdDev | Scaled | Scaled-StdDev |        RPS | Allocated |
------------------------------ |-------------------------------------------------------------------- |---------- |---------- |---------- |------- |-------------- |----------- |---------- |
          PlaintextTechEmpower | Microsoft.AspNetCore.Server.Kestrel.Internal.Http.KestrelHttpParser | 1.3479 us | 0.0118 us | 0.0646 us |   1.00 |          0.00 | 741,905.06 |     353 B |
 PipelinedPlaintextTechEmpower | Microsoft.AspNetCore.Server.Kestrel.Internal.Http.KestrelHttpParser | 1.0828 us | 0.0046 us | 0.0251 us |   0.80 |          0.04 | 923,522.40 |     351 B |
                    LiveAspNet | Microsoft.AspNetCore.Server.Kestrel.Internal.Http.KestrelHttpParser | 3.4407 us | 0.0347 us | 0.1902 us |   2.56 |          0.18 | 290,634.41 |   1.41 kB |
           PipelinedLiveAspNet | Microsoft.AspNetCore.Server.Kestrel.Internal.Http.KestrelHttpParser | 3.1495 us | 0.0243 us | 0.1332 us |   2.34 |          0.14 | 317,512.69 |   1.39 kB |
                       Unicode | Microsoft.AspNetCore.Server.Kestrel.Internal.Http.KestrelHttpParser | 7.0485 us | 0.0673 us | 0.3684 us |   5.24 |          0.35 | 141,873.35 |   3.91 kB |
              UnicodePipelined | Microsoft.AspNetCore.Server.Kestrel.Internal.Http.KestrelHttpParser | 7.0970 us | 0.0602 us | 0.3298 us |   5.28 |          0.33 | 140,904.71 |   3.89 kB |
```

**dev**
```
                        Method |                                                          ParserType |      Mean |    StdErr |    StdDev | Scaled | Scaled-StdDev |        RPS | Allocated |
------------------------------ |-------------------------------------------------------------------- |---------- |---------- |---------- |------- |-------------- |----------- |---------- |
          PlaintextTechEmpower | Microsoft.AspNetCore.Server.Kestrel.Internal.Http.KestrelHttpParser | 1.2991 us | 0.0187 us | 0.1026 us |   1.00 |          0.00 | 769,784.52 |     353 B |
 PipelinedPlaintextTechEmpower | Microsoft.AspNetCore.Server.Kestrel.Internal.Http.KestrelHttpParser | 1.0524 us | 0.0051 us | 0.0282 us |   0.81 |          0.06 | 950,246.05 |     351 B |
                    LiveAspNet | Microsoft.AspNetCore.Server.Kestrel.Internal.Http.KestrelHttpParser | 2.6336 us | 0.0256 us | 0.1401 us |   2.04 |          0.17 | 379,709.69 |   1.12 kB |
           PipelinedLiveAspNet | Microsoft.AspNetCore.Server.Kestrel.Internal.Http.KestrelHttpParser | 2.3801 us | 0.0135 us | 0.0739 us |   1.84 |          0.13 | 420,148.35 |   1.11 kB |
                       Unicode | Microsoft.AspNetCore.Server.Kestrel.Internal.Http.KestrelHttpParser | 4.1385 us | 0.0238 us | 0.1304 us |   3.20 |          0.23 | 241,633.89 |   1.99 kB |
              UnicodePipelined | Microsoft.AspNetCore.Server.Kestrel.Internal.Http.KestrelHttpParser | 4.2248 us | 0.0318 us | 0.1744 us |   3.27 |          0.25 | 236,696.38 |   1.98 kB |
```

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
natemcmaster: We talked about this the first time I implemented this https://github.com/aspnet/KestrelHttpServer/pull/1246. We didn't end up merging that one because we want to prioritize @pakrym's work on pipelines.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
natemcmaster: ah, thx. yes, this can move to the utils class.

davidfowl:
That doesn't look good (esp the unicode and unicode pipelined). Profile and see where the time is spent

natemcmaster:
Looking into it. Correctness comes before perf, but I'll profile more to see what I can gain back

natemcmaster:
Uri.Create.

https://github.com/aspnet/KestrelHttpServer/pull/1471

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
```diff
@@ -1250,34 +1317,103 @@ public void OnStartLine(HttpMethod method, HttpVersion version, Span<byte> targe
...
+        }
+
+        private void OnAbsoluteFormTarget(Span<byte> target, Span<byte> query, Span<byte> line)
+        {
+            // absolute-form
+            // https://tools.ietf.org/html/rfc7230#section-5.3.2
+
+            RawTarget = target.GetAsciiStringNonNullCharacters();
+
+            if (!Uri.TryCreate(RawTarget, UriKind.Absolute, out var uri))
```
davidfowl: Do we really need to use `Uri.Create`? Maybe we should roll our own.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
benaadams: That's the extra work to avoid 😉  push it to right of throw and combine the two
```csharp
private BadHttpRequestException GetRequestLineRejection(Span<byte> line)
{
    const int MaxRequestLineError = 32;
    if (Log.IsEnabled(LogLevel.Information))
    {
         var line = span.GetAsciiStringEscaped(MaxRequestLineError);
         return BadHttpRequestException.GetException(RequestRejectionReason.InvalidRequestLine, line);
    }
    else
    {
         return BadHttpRequestException.GetException(RequestRejectionReason.InvalidRequestLine);
    }
}

private void ThrowRequestLineRejection(Span<byte> line)
{
    throw GetRequestLineRejection(line)
}
```
Then
```csharp
if (!char.IsLetterOrDigit(ch) && ch != ':' && ch != '.')
{
    ThrowRequestLineRejection(line);
}
```

benaadams: That's the extra work to avoid 😉  push it to right of throw and combine the two
```csharp
private BadHttpRequestException GetRequestLineRejection(Span<byte> line)
{
    const int MaxRequestLineError = 32;
    if (Log.IsEnabled(LogLevel.Information))
    {
         var line = line.GetAsciiStringEscaped(MaxRequestLineError);
         return BadHttpRequestException.GetException(RequestRejectionReason.InvalidRequestLine, line);
    }
    else
    {
         return BadHttpRequestException.GetException(RequestRejectionReason.InvalidRequestLine);
    }
}

private void ThrowRequestLineRejection(Span<byte> line)
{
    throw GetRequestLineRejection(line)
}
```
Then
```csharp
if (!char.IsLetterOrDigit(ch) && ch != ':' && ch != '.')
{
    ThrowRequestLineRejection(line);
}
```
Keep the main body tight with only code that's mostly always executed; push everything else out

benaadams: Unless you have to pass a bucket of params to do it 😝

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/KestrelHttpParser.cs
```diff
@@ -66,6 +68,27 @@ public KestrelHttpParser(IKestrelTrace log)
...
+            // |    |       ||
+            // /path?query=1 HTTP/1.1
+
+            // pathStart
+            // |               pathEnd and queryStart
+            // |               |
+            // |               |       queryEnd
+            // |               |       |versionStart
+            // |               |       ||
+            // http://host/path?query=1 HTTP/1.1
```
cesarbs: `pathStart` should be at the first `/` in the actual path. For the special cases `http://host` and `http://host?query`, make a span with `/` and pass that to `OnStartLine`.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
cesarbs: We should rename `IsValidTokenChar` to `IsValidMethodChar`.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/KestrelHttpParser.cs
davidfowl: The problem is we need to do that in a bunch of places and this probably doesn't handle all of those cases.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
natemcmaster: This will change when @cesarbs puts https://github.com/aspnet/KestrelHttpServer/pull/1469 in.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
davidfowl: and move to HttpUtlities

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
```diff
@@ -1216,8 +1232,59 @@ protected void ReportApplicationError(Exception ex)
             Log.ApplicationError(ConnectionId, ex);
         }
 
-        public void OnStartLine(HttpMethod method, HttpVersion version, Span<byte> target, Span<byte> path, Span<byte> query, Span<byte> customMethod)
+        public void OnStartLine(HttpMethod method, HttpVersion version, Span<byte> target, Span<byte> path, Span<byte> query, Span<byte> customMethod, Span<byte> line)
         {
+            if (target.Length == 0)
```
davidfowl: Target can't be empty here can it?

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/KestrelHttpParser.cs
```diff
@@ -66,6 +68,27 @@ public KestrelHttpParser(IKestrelTrace log)
                 span = startLineBuffer.ToSpan();
             }
 
+            // pathStart
+            // |    pathEnd and queryStart
```
natemcmaster: do we need pathEnd _and_ queryStart? I couldn't think of a scenario when they are different.

benaadams: Only if path had been decoded in place and query not moved (e.g. path could get shorter) - though this doesn't currently happen

benaadams: have removed `pathEnd` in https://github.com/aspnet/KestrelHttpServer/pull/1463

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
natemcmaster: It should. This could be a debug.assert instead.

natemcmaster:
Here are the new comparisons with the https://github.com/aspnet/KestrelHttpServer/pull/1471 fix in benchmarks.

**dev**
```
                        Method |      Mean |    StdErr |    StdDev |    Median | Scaled | Scaled-StdDev |        RPS | Allocated |
------------------------------ |---------- |---------- |---------- |---------- |------- |-------------- |----------- |---------- |
          PlaintextTechEmpower | 1.4258 us | 0.0287 us | 0.1572 us | 1.3617 us |   1.00 |          0.00 | 701,360.54 |     353 B |
          PlaintextAbsoluteUri | 1.4000 us | 0.0103 us | 0.0566 us | 1.3791 us |   0.99 |          0.11 | 714,310.21 |     387 B |
 PipelinedPlaintextTechEmpower | 1.0874 us | 0.0106 us | 0.0578 us | 1.0690 us |   0.77 |          0.09 | 919,605.26 |     351 B |
                    LiveAspNet | 2.5348 us | 0.0224 us | 0.1227 us | 2.5159 us |   1.80 |          0.20 | 394,501.35 |   1.08 kB |
           PipelinedLiveAspNet | 2.5066 us | 0.0393 us | 0.2150 us | 2.3992 us |   1.78 |          0.23 | 398,949.37 |   1.07 kB |
                       Unicode | 3.9702 us | 0.0447 us | 0.2448 us | 3.8809 us |   2.81 |          0.33 | 251,878.79 |   1.89 kB |
              UnicodePipelined | 3.8373 us | 0.0128 us | 0.0699 us | 3.8163 us |   2.72 |          0.27 | 260,600.24 |   1.86 kB |
```

**PR**
```
                        Method |       Mean |    StdErr |    StdDev | Scaled | Scaled-StdDev |        RPS | Allocated |
------------------------------ | ---------- |---------- |---------- |------- |-------------- |----------- |---------- |
          PlaintextTechEmpower |  1.3305 us | 0.0053 us | 0.0292 us |   1.00 |          0.00 | 751,581.09 |     353 B |
          PlaintextAbsoluteUri |  2.1944 us | 0.0153 us | 0.0837 us |   1.65 |          0.07 | 455,696.01 |     698 B |
 PipelinedPlaintextTechEmpower |  1.0833 us | 0.0040 us | 0.0220 us |   0.81 |          0.02 | 923,131.88 |     351 B |
                    LiveAspNet |  2.5116 us | 0.0147 us | 0.0805 us |   1.89 |          0.07 | 398,159.47 |   1.08 kB |
           PipelinedLiveAspNet |  2.3130 us | 0.0068 us | 0.0372 us |   1.74 |          0.05 | 432,341.14 |   1.07 kB |
                       Unicode |  4.2242 us | 0.0505 us | 0.2767 us |   3.18 |          0.22 | 236,730.41 |   1.89 kB |
              UnicodePipelined |  4.0434 us | 0.0663 us | 0.3632 us |   3.04 |          0.28 | 247,318.74 |   1.86 kB |
```

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
JamesNK: `Uri` is a garbage hole for performance.

natemcmaster: This code is about correctness, not perf. For context, we talked about this in the [first iteration](https://github.com/aspnet/KestrelHttpServer/pull/1246) of absolute-URI support. We're aware this is slow, but it's not on the hot path. In fact, it's not really on any commonly used code path. The HTTP spec requires we handle requests like `GET http://localhost/path HTTP/1.1`, but in practice, clients never do. Instead they use origin-form, `GET /path HTTP/1.1`. We debated whether to bother at all, but since our goal is edge-server capabilities, we will comply with the HTTP spec, including its edge-cases. Fast URI parsing is not a high priority.

natemcmaster:
🆙 :date: had a bunch of merge conflicts with https://github.com/aspnet/KestrelHttpServer/pull/1463. Re-running benchmarks.

(Will squash before merging)

natemcmaster:
dev
```
                        Method |      Mean |    StdErr |    StdDev |    Median | Scaled | Scaled-StdDev |        RPS | Allocated |
------------------------------ |---------- |---------- |---------- |---------- |------- |-------------- |----------- |---------- |
          PlaintextTechEmpower | 1.2376 us | 0.0152 us | 0.0834 us | 1.2097 us |   1.00 |          0.00 | 808,015.22 |     349 B |
          PlaintextAbsoluteUri | 1.2648 us | 0.0062 us | 0.0339 us | 1.2635 us |   1.03 |          0.07 | 790,612.89 |     381 B |
 PipelinedPlaintextTechEmpower | 1.0391 us | 0.0255 us | 0.1395 us | 0.9648 us |   0.84 |          0.12 | 962,407.67 |     348 B |
                    LiveAspNet | 2.4872 us | 0.0665 us | 0.3644 us | 2.4205 us |   2.02 |          0.32 | 402,063.77 |   1.07 kB |
           PipelinedLiveAspNet | 2.1457 us | 0.0075 us | 0.0410 us | 2.1392 us |   1.74 |          0.11 | 466,051.66 |   1.07 kB |
                       Unicode | 3.9496 us | 0.0641 us | 0.3508 us | 3.7639 us |   3.20 |          0.35 | 253,191.49 |   1.85 kB |
              UnicodePipelined | 3.8724 us | 0.0461 us | 0.2525 us | 3.7603 us |   3.14 |          0.28 | 258,234.69 |   1.86 kB |
```

PR
```
                        Method |          Mean |      StdDev |        Median | Scaled | Scaled-StdDev |          RPS | Allocated |
------------------------------ |-------------- |------------ |-------------- |------- |-------------- |------------- |---------- |
          PlaintextTechEmpower | 1,197.5448 ns |  22.3411 ns | 1,195.7931 ns |   1.00 |          0.00 |   835,041.84 |     349 B |
          PlaintextAbsoluteUri | 2,035.7134 ns |  41.9057 ns | 2,031.9959 ns |   1.70 |          0.05 |   491,228.28 |     692 B |
 PipelinedPlaintextTechEmpower |   990.8188 ns |  28.8461 ns |   984.0139 ns |   0.83 |          0.03 | 1,009,266.29 |     348 B |
                    LiveAspNet | 2,318.5341 ns |  42.4130 ns | 2,303.5764 ns |   1.94 |          0.05 |   431,307.01 |   1.07 kB |
           PipelinedLiveAspNet | 2,197.7497 ns |  55.2662 ns | 2,184.0570 ns |   1.84 |          0.06 |   455,010.87 |   1.07 kB |
                       Unicode | 3,762.2063 ns | 179.1014 ns | 3,661.2269 ns |   3.14 |          0.16 |   265,801.48 |   1.85 kB |
              UnicodePipelined | 3,645.7075 ns |  74.3930 ns | 3,624.3606 ns |   3.05 |          0.08 |   274,295.18 |   1.86 kB |
```

davidfowl:
Wait, you made it faster?

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/KestrelHttpParser.cs
```diff
@@ -157,7 +162,7 @@ public KestrelHttpParser(IKestrelTrace log)
                 RejectRequestLine(data, length);
             }
 
-            handler.OnStartLine(method, httpVersion, targetBuffer, pathBuffer, query, customMethod, pathEncoded);
+            handler.OnStartLine(method, httpVersion, targetBuffer, pathBuffer, query, customMethod, new Span<byte>(data, length), pathEncoded);
```
davidfowl: Declare a local

natemcmaster:
All but the PlaintextAbsoluteUri benchmark which is almost cut in half. But that's going from 800k RPS of incorrect handling to 480k RPS of the request.

src/Microsoft.AspNetCore.Server.Kestrel/BadHttpRequestException.cs
```diff
@@ -1,14 +1,20 @@
...
 using Microsoft.AspNetCore.Server.Kestrel.Internal.Http;
+using Microsoft.AspNetCore.Server.Kestrel.Internal.Infrastructure;
+using Microsoft.Extensions.Primitives;
 
 namespace Microsoft.AspNetCore.Server.Kestrel
 {
     public sealed class BadHttpRequestException : IOException
     {
+        // used in rare cases
+        private HttpMethod? _requiredMethod;
```
cesarbs: `readonly` and add second ctor.

cesarbs: `readonly` and add second ctor.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
```diff
@@ -1179,6 +1185,15 @@ public void RejectRequest(RequestRejectionReason reason, string value)
             throw BadHttpRequestException.GetException(reason, value);
         }
 
+        private void RejectRequestLine(Span<byte> span)
```
cesarbs: This will change with #1469.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
```diff
@@ -1216,8 +1232,51 @@ protected void ReportApplicationError(Exception ex)
             Log.ApplicationError(ConnectionId, ex);
         }
 
-        public void OnStartLine(HttpMethod method, HttpVersion version, Span<byte> target, Span<byte> path, Span<byte> query, Span<byte> customMethod, bool pathEncoded)
+        public void OnStartLine(HttpMethod method, HttpVersion version, Span<byte> target, Span<byte> path, Span<byte> query, Span<byte> customMethod, Span<byte> line, bool pathEncoded)
```
cesarbs: #1469 introduces rejection of request targets, so you won't have to pass the entire line here anymore.

src/Microsoft.AspNetCore.Server.Kestrel/BadHttpRequestException.cs
```diff
@@ -17,6 +23,14 @@ private BadHttpRequestException(string message, int statusCode)
 
         internal int StatusCode { get; }
 
+        internal void SetAdditionalHeaders(FrameResponseHeaders headers)
```
pakrym: It's a bit strange having logic in exception and calling methods on it.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
```diff
@@ -1249,34 +1308,113 @@ public void OnStartLine(HttpMethod method, HttpVersion version, Span<byte> targe
                 }
             }
 
-            var normalizedTarget = PathNormalizer.RemoveDotSegments(requestUrlPath);
-            if (method != HttpMethod.Custom)
+            QueryString = query.GetAsciiStringNonNullCharacters();
+            RawTarget = rawTarget;
+            SetNormalizedPath(requestUrlPath);
+        }
+
```
cesarbs: nit: remove extra blank line

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
```diff
@@ -1249,34 +1308,113 @@ public void OnStartLine(HttpMethod method, HttpVersion version, Span<byte> targe
...
+                        RejectRequestLine(line);
+                    }
+
+                    throw BadHttpRequestException.GetException(RequestRejectionReason.InvalidRequestLine);
+                }
+            }
+
+            // The authority-form of request-target is only used for CONNECT
+            // requests (https://tools.ietf.org/html/rfc7231#section-4.3.6).
+
```
cesarbs: nit: blank line

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
```diff
@@ -1249,34 +1308,113 @@ public void OnStartLine(HttpMethod method, HttpVersion version, Span<byte> targe
...
+            // When making a CONNECT request to establish a tunnel through one or
+            // more proxies, a client MUST send only the target URI's authority
+            // component(excluding any userinfo and its "@" delimiter) as the
+            // request - target.For example,
+            //
+            //  CONNECT www.example.com:80 HTTP/1.1
+            //
+            // Allowed characters in the 'host + port' section of authority.
+            // See https://tools.ietf.org/html/rfc3986#section-3.2
+
```
cesarbs: nit: blank line

src/Microsoft.AspNetCore.Server.Kestrel/BadHttpRequestException.cs
pakrym: I would prefer subclass of `BadHttpRequestException` with `RequiredMethod` property and downcast check in `ProduceEnd`

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/HttpScheme.cs
```diff
@@ -0,0 +1,12 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Http
+{
+    public enum HttpScheme
+    {
+        Unknown = -1,
+        Http = 0,
+        HttpSecure = 1
```
cesarbs: `Https`

src/Microsoft.AspNetCore.Server.Kestrel/BadHttpRequestException.cs
natemcmaster: BadHttpRequestException is sealed :-/

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/KestrelHttpParser.cs
```diff
@@ -109,6 +109,11 @@ public KestrelHttpParser(IKestrelTrace log)
 
                     pathEncoded = true;
                 }
+                else if (ch == 0)
```
cesarbs: Why is this being checked here now? Should blow up when trying to get ASCII string.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Infrastructure/HttpUtilities.cs
```diff
@@ -15,7 +15,13 @@ public static class HttpUtilities
         public const string Http10Version = "HTTP/1.0";
         public const string Http11Version = "HTTP/1.1";
 
+        public const string HttpUriScheme = "http://";
+        public const string HttpSecureUriScheme = "https://";
```
cesarbs: `HttpsUriScheme`

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
davidfowl: rename span to requestLine

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Infrastructure/HttpUtilities.cs
natemcmaster: I was intentionally trying to emphasize the "S" of https by expanding it to Secure. It's easy to miss that extra character when reviewing/writing code.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/KestrelHttpParser.cs
natemcmaster: Fail fast. The null character is never valid in the start line.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/HttpScheme.cs
natemcmaster: How strong is your preference for this? I'm chose the name to make it super obvious that it's http + TLS, not plain-old http.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/KestrelHttpParser.cs
pakrym: We are validating these during materialization of values, fail fast is less important then less checks in hot paths

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/HttpScheme.cs
davidfowl: We use https everywhere else

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
```diff
@@ -1127,6 +1131,15 @@ private void RejectNonBodyTransferEncodingResponse(bool appCompleted)
             }
         }
 
+        private void SetErrorResponseException(BadHttpRequestException ex)
+        {
+            SetErrorResponseHeaders(ex.StatusCode);
+            if (ex.AllowedHeader != default(StringValues))
```
davidfowl: StringValues.IsNullOrEmpty

src/Microsoft.AspNetCore.Server.Kestrel/BadHttpRequestException.cs
```diff
@@ -12,25 +11,25 @@ namespace Microsoft.AspNetCore.Server.Kestrel
 {
     public sealed class BadHttpRequestException : IOException
     {
-        // used in rare cases
-        private HttpMethod? _requiredMethod;
-
         private BadHttpRequestException(string message, int statusCode)
+            :this(message, statusCode, null)
```
cesarbs: nit: `: this`

natemcmaster:
Squashed, formatted, benchmarked, etc. Awaiting sign off.

src/Microsoft.AspNetCore.Server.Kestrel/BadHttpRequestException.cs
natemcmaster: done

PR #1471, opened 2017-03-09
Make benchmarks closer to what a client should send

natemcmaster:
We shouldn't benchmark a "bad client". Absolute-form requests should only be sent to proxies, not end-servers.

### DIFF
--- a/test/Microsoft.AspNetCore.Server.Kestrel.Performance/RequestParsing.cs
+++ b/test/Microsoft.AspNetCore.Server.Kestrel.Performance/RequestParsing.cs
@@ -42,6 +42,16 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Performance
             }
         }
 
+        [Benchmark(Baseline = true, OperationsPerInvoke = RequestParsingData.InnerLoopCount)]
+        public void PlaintextAbsoluteUri()
+        {
+            for (var i = 0; i < RequestParsingData.InnerLoopCount; i++)
+            {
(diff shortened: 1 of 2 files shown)


davidfowl:
Sneaky, fix the benchmark 😄. I'd also run the end to end tests.

natemcmaster:
hehe, path of least resistance :)

natemcmaster:
To clarify:
```

   To allow for transition to the absolute-form for all requests in some
   future version of HTTP, a server MUST accept the absolute-form in
   requests, even though HTTP/1.1 clients will only send them in
   requests to proxies.
```

https://tools.ietf.org/html/rfc7230#section-5.3.2

natemcmaster:
Essentially these benchmarks are testing a scenario that clients should never do.

natemcmaster:
:up: :date:

test/Microsoft.AspNetCore.Server.Kestrel.Performance/RequestParsing.cs
```diff
@@ -42,6 +42,16 @@ public void PlaintextTechEmpower()
             }
         }
 
+        [Benchmark(Baseline = true, OperationsPerInvoke = RequestParsingData.InnerLoopCount)]
```
benaadams: Can't have two benchmarks with `Baseline = true`

natemcmaster: oops. sorry, i'm the new here. will fix it.

PR #1483, opened 2017-03-12
Remove explicit static constructors

justinvp:
Explicit static cctors cause the C# compiler to not mark types as beforefieldinit, which means the JIT will add checks to each static method and instance constructors of the type to make sure that the static constructor was previously called.

See https://msdn.microsoft.com/en-us/library/ms182275.aspx


halter73:
Interesting MSDN article. I had no idea that the JIT optimizes inline static initialization better than explicit static ctors.

Drawaes:
After reading this I am off deleting them from all my code!

benaadams:
> I had no idea that the JIT optimizes inline static initialization better than explicit static ctors.

It puts the initalization check in the first place at runtime that accesses a static variable of the class; can be a good place can be a bad place; but is only one place. Whereas the static .cctor is check put everywhere that accesses the class.

Some people prefer the static .cctor as its more deterministic (e.g. if it also does something like writing to a log); though I've never quite understood it.

As an aside the first function that causes the instantiation of the static variables will not have branch elimination performed by the jit for readonly statics; and the readonly statics wil not be treated as consts in that function, (As the function asm generation output for that function has started and they are not consts "yet").

Corert will still have the checks everywhere; as its jit time magic 😢

PR #1487, opened 2017-03-12
Clean up benchmarks

davidfowl:
- Benchmarks are suffixed with *Benchmarks
- Removed Custom RpsColumn.cs and used Ops/Second column
- Removed params from RequestParsingBenchmark

What Ops/S looks like:

```
                      Method |        Mean |     StdErr |      StdDev |       Op/s | Scaled | Scaled-StdDev | Allocated |
---------------------------- |------------ |----------- |------------ |----------- |------- |-------------- |---------- |
          FrameOverheadTotal | 680.3866 ns | 28.4564 ns | 155.8620 ns | 1469752.57 |   1.00 |          0.00 |     359 B |
    FrameOverheadRequestLine | 234.0376 ns |  1.5485 ns |   8.4815 ns | 4272817.88 |   0.36 |          0.07 |      48 B |
 FrameOverheadRequestHeaders | 395.2060 ns |  5.1925 ns |  28.4404 ns | 2530326.07 |   0.61 |          0.12 |     311 B |
```


benaadams:
`InnerLoopCount` and the for loops can mostly disappear the tasks are slow enough and when they allocate it causes more GC per run; which introduces more noise in timing.

Still need `OperationsPerInvoke` for pipelining; but that should just become:

`[Benchmark(OperationsPerInvoke = RequestParsingDataBenchmark.Pipelining)]`

and the others

`[Benchmark]`

or

`[Benchmark(Baseline = true)]`

benaadams:
Though your `FrameOverheadRequestLine` examples might need the loop as that's probably pretty close to timer frequency/function call overhead 😄

benaadams:
Actually maybe just reduce the const a little from e.g. 512 to 16

test/Microsoft.AspNetCore.Server.Kestrel.Performance/FrameParsingOverheadBenchmark.cs
```diff
@@ -60,14 +60,14 @@ private void ParseRequest()
 
             if (!_frame.TakeStartLine(_buffer, out var consumed, out var examined))
             {
-                RequestParsing.ThrowInvalidRequestLine();
+                RequestParsingBenchmark.ThrowInvalidRequestLine();
```
cesarbs: While you're at it, let's move the `Throw`* methods into a separate class (`ErrorUtilities` or something).

test/Microsoft.AspNetCore.Server.Kestrel.Performance/RequestParsingDataBenchmark.cs
```diff
@@ -8,7 +8,7 @@
 namespace Microsoft.AspNetCore.Server.Kestrel.Performance
 {
     [Config(typeof(CoreConfig))]
-    public class RequestParsingData
+    public class RequestParsingDataBenchmark
```
cesarbs: This class shouldn't have the `Benchmark` suffix, since it's not a benchmark class.

You can remove the `Config` attribute from it too, it was a copy & paste mishap when I created the class (hadn't realized it till now).

davidfowl:
I'll can change the inner loop count. Not the orther stuff as part of this PR. That'll require a re-run to make sure each of the benchmarks still shows valid data

benaadams:
:shipit: 

I'll do a follow up with loop count; I have been questioning my choice of number after running it so many times

davidfowl:
@cesarbs wtf, merge conflicts..

PR #1488, opened 2017-03-13
InitializeHeaders only at start of parsing/Fix remaining

benaadams:
Should only InitalizeHeaders at start of read rather than in each loop

### DIFF
--- a/test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/RequestHeaderLimitsTests.cs
+++ b/test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/RequestHeaderLimitsTests.cs
@@ -1,11 +1,16 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System;
 using System.Linq;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.Testing;
+using Microsoft.AspNetCore.Server.Kestrel.Internal.Http;
 using Xunit;
+using Microsoft.Extensions.Primitives;
+using System.Collections;
+using System.Collections.Generic;
 
 namespace Microsoft.AspNetCore.Server.Kestrel.FunctionalTests
 {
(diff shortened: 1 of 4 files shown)


davidfowl:
We shouldn't fix this without a test. Seems like a regression.

benaadams:
I initially thought it was worse than is it is (that it also reset the headers); but is only a [speed bump](https://github.com/aspnet/KestrelHttpServer/blob/dev/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs#L281-L296) restoring the header collections if they were swapped out (and creating if they didn't originally exist).

`feature-si` only performed it once so is a perf regression; will look to test.

benaadams:
Found a bug adding tests; fixing

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/RequestHeaderLimitsTests.cs
```diff
@@ -122,11 +231,24 @@ public class RequestHeaderLimitsTests
             }
         }
 
-        private static string MakeHeaders(int count)
+        private static async Task WaitForCondition(Func<bool> condition)
+        {
+            const int MaxWaitLoop = 250;
```
cesarbs: Add a timeout param and make this a function of the timeout period.

davidfowl:
Can we not do functional tests for this.

benaadams:
> What bug did you find?

ServerAcceptsRequestAcrossSends gives up parsing headers for various values; didn't expect it to and haven't isolated the cause yet. (Its not across the sends...)

> Message: Assert.Equal() Failure
Expected: 700
Actual: 150

> Message: Assert.Equal() Failure
Expected: 600
Actual: 336

> Message: Assert.Equal() Failure
Expected: 2000
Actual: 1384

cesarbs:
😱

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/RequestHeaderLimitsTests.cs
```diff
@@ -76,6 +81,110 @@ public class RequestHeaderLimitsTests
...
         [Theory]
+        [InlineData(1, 1)]
+        [InlineData(5, 5)]
+        [InlineData(100, 100)]
+        [InlineData(600, 100)]
+        [InlineData(700, 1)]
+        [InlineData(1, 700)]
+        [InlineData(1000, 1000)]
+        [InlineData(4000, 1)]
+        public async Task ServerAcceptsRequestAcrossSends(int header0Count, int header1Count)
```
cesarbs: `AcceptsHeaders`?

benaadams:
remaining 33; length 22
span length 1024
337 headers parsed

benaadams:
Fixed the bug https://github.com/aspnet/KestrelHttpServer/pull/1488/commits/30abd32933348a58634fe44cf8b8a2c7052de9d4

Issue #1050, opened 2016-08-17
Remove splitting of path and path base in Kestrel


JunTaoLuo:
This is now implemented as the `UsePathBase` Middleware in HttpAbstractions. We can now remove this feature in Kestrel.

davidfowl:
👏

John0King:
Will this effect .net Core app running with IIS as a subapplication ?

brockallen:
Why add `UsePathBase` and not just use `Map`?

Tratcher:
Map branches, UsePathBase trims (if it matches) and continues.

brockallen:
So UsePathBase is the same as Map, if you had just taken everything in the Configure and moved it into the config inside the Map?

Tratcher:
The path logic is the same, but for Map the other branch would always be empty and 404.

For the urls `http://localhost/favicon.ico` and `http://localhost/PathBase/favicon.ico`, if you use UsePathBase then the two both result in the same `/favicon.ico` for the next middleware, but if you Map then you get two pipelines that may have different middleware. If you only configure the Map pipeline, then the non-Map pipeline always 404s.

brockallen:
I guess all I was getting at was that it seemed odd to have a new API when you can get the same effect with an existing API.

> For the urls http://localhost/favicon.ico and http://localhost/PathBase/favicon.ico, if you use UsePathBase then the two both result in the same /favicon.ico for the next middleware

Wait, so when you said "strip" this is what you mean I guess... meaning if the URL starts with the ~/BaseUrl then it's stripped and then proceeds to the next middleware... Is that what UseUrls used to do? Or am I still confused? Perhaps I guess I need to get the nightlys and test.

Tratcher:
Yes, that's the same as what UseUrls used to do (in Kestrel).

brockallen:
Ok, I setup a quick test sample on v1.1. In short UseUrls("http://localhost:123/foo") still accepts requests to "http://localhost:123/bar", and that's my confusion. This behavior (that I was previously unaware of) is pretty counter intuitive. Your explanation above now makes more sense, but I am left feeling like I'm missing some major context as to why the counter intuitive behavior existed in the first place. I guess it's a "kestrel thing" (or maybe libuv?), as opposed to how I'd expect http.sys reserved URLs to behave?

Tratcher:
[...]

PR #1489, opened 2017-03-13
Remove splitting of path and path base (#1050)

cesarbs:
#1050


Tratcher:
Post a breaking change announcement for this.

natemcmaster:
This should improve our perf a little, right? Curious what the delta is.

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/RequestTests.cs
```diff
@@ -520,6 +521,53 @@ public void CanUpgradeRequestWithConnectionKeepAliveUpgradeHeader()
             }
         }
 
+        [Theory]
```
davidfowl: Why a functional test?

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/RequestTests.cs
```diff
@@ -520,6 +521,53 @@ public void CanUpgradeRequestWithConnectionKeepAliveUpgradeHeader()
...
+                });
+
+            using (var host = builder.Build())
+            {
+                host.Start();
+
+                using (var connection = new TestConnection(host.GetPort()))
+                {
+                    await connection.Send($"GET {requestPath} HTTP/1.1\r\n\r\n");
+                    await done.WaitAsync(TimeSpan.FromSeconds(10));
```
davidfowl: :\

cesarbs:
@natemcmaster There seems to be a small improvement in the `FrameParsingOverhead` benchmark. Look at the second row in each table:

### dev

```
                      Method |        Mean |     StdDev |       Op/s | Scaled | Scaled-StdDev | Allocated |
---------------------------- |------------ |----------- |----------- |------- |-------------- |---------- |
          FrameOverheadTotal | 604.8350 ns | 24.2150 ns | 1653343.36 |   1.00 |          0.00 |     359 B |
    FrameOverheadRequestLine | 271.7461 ns | 11.2063 ns | 3679905.09 |   0.45 |          0.02 |      48 B |
 FrameOverheadRequestHeaders | 407.2028 ns |  6.0652 ns | 2455778.95 |   0.67 |          0.03 |     311 B |
```

### cesarbs/1050

``` 
                      Method |        Mean |     StdDev |       Op/s | Scaled | Scaled-StdDev | Allocated |
---------------------------- |------------ |----------- |----------- |------- |-------------- |---------- |
          FrameOverheadTotal | 585.6759 ns | 12.9140 ns | 1707428.91 |   1.00 |          0.00 |     359 B |
    FrameOverheadRequestLine | 259.4311 ns |  4.5198 ns | 3854587.27 |   0.44 |          0.01 |      48 B |
 FrameOverheadRequestHeaders | 403.2611 ns | 18.2439 ns | 2479783.25 |   0.69 |          0.03 |     311 B |
```

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/RequestTests.cs
cesarbs: Because it was already there (I moved it here from `RequestTargetProcessingTests` and adapted as accordingly) 😝 

Will write a unit test.

davidfowl: Yes, please nuke this.

cesarbs: > ~~Will write a unit test.~~

Since there's no `PathBase` left anywhere else but `Frame`, it's hard to write a unit test for this. A functional test is still the only place where I can specify a path base in `UseUrls` and verify Kestrel does nothing with it.

So either we leave no tests for this or we keep the functional test.

natemcmaster: > there's no PathBase left anywhere else but Frame

IIUC we would never set or get PathBase on Frame, right? Could we make the implementation of IHttpRequestFeature.PathBase in Frame.FeatureCollection.cs return string.Empty and throw on set?

cesarbs: It already returns `string.Empty` if not set. We can't throw on set because it's set by middleware (`UsePathBaseMiddleware` in HttpAbstractions).

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs
```diff
@@ -1356,7 +1321,6 @@ private void OnAuthorityFormTarget(HttpMethod method, Span<byte> target)
             // See https://tools.ietf.org/html/rfc3986#section-3.2
             RawTarget = target.GetAsciiStringNonNullCharacters();
             Path = string.Empty;
-            PathBase = string.Empty;
```
davidfowl: Doesn't it still need to be set to `string.Empty`?

cesarbs: Nope: https://github.com/aspnet/KestrelHttpServer/blob/57b368566d74b3cb7da60f412609938d2c96fdf6/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.FeatureCollection.cs#L120

src/Microsoft.AspNetCore.Server.Kestrel/ListenOptions.cs
```diff
@@ -81,9 +81,7 @@ internal ListenOptions(ulong fileHandle)
         /// </remarks>
         public List<IConnectionAdapter> ConnectionAdapters { get; } = new List<IConnectionAdapter>();
 
-        // PathBase and Scheme are hopefully only a temporary measure for back compat with IServerAddressesFeature.
-        // This allows a ListenOptions to describe all the information encoded in IWebHostBuilder.UseUrls.
```
JunTaoLuo: Why remove this line in the comments? It still applies no?

halter73:
@davidfowl We need decide if/how we are going change string based binding via IServerAddressesFeatures (e.g. UseUrls). After this change, it looks like we are just going to ignore any path base supplied this way without warning.

src/Microsoft.AspNetCore.Server.Kestrel/ListenOptions.cs
halter73: As the original writer of this comment, I think what's left there is enough.

Tratcher:
Since it's a breaking change you should at least log a warning.

halter73:
+1 for at least logging a warning for now. If we remove IServerAddressesFeatures by 2.0.0 (I'm not holding my breath on that), we can remove the warning then.

davidfowl:
There should be no warning. Let's just document the breaking change.

cesarbs:
Added warning...

PR #1493, opened 2017-03-13
Add EventSource events

### LINKED ISSUE #1211: Add EventSource events
Events:
- Connection Id (to establish correlation with the current request)

natemcmaster:
Adds an event source named 'Microsoft-AspNetCore-Server-Kestrel' with the following event and properties:

- ConnectionStart:
   - connectionId
   - scheme
   - localEndPoint
   - remoteEndPoint
- ConnectionStop:
   - connectionId

Resolves #1211 

@kichalla @shirhatti this should contain the events and details we discussed over email. Can you let me know if more are needed?

@halter73 are there more places we need to put logging for ConnectionStop? We need to ensure ConnectionStop is logged once and only once per connection opened.

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Connection.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Connection.cs
@@ -86,6 +86,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Http
         public void Start()
         {
             Log.ConnectionStart(ConnectionId);
+            KestrelEventSource.Log.ConnectionStart(this);
 
             // Start socket prior to applying the ConnectionAdapter
             _socket.ReadStart(_allocCallback, _readCallback, this);
@@ -137,6 +138,8 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Http
         // Called on Libuv thread
         public virtual void OnSocketClosed()
         {
+            KestrelEventSource.Log.ConnectionStop(this);
+
             _frame.FrameStartedTask.ContinueWith((task, state) =>
             {
                 var connection = (Connection)state;
--- a/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/SocketOutput.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/SocketOutput.cs
@@ -199,7 +199,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Http
                 ScheduleWrite();
             }
 
-            // Return TaskCompletionSource's Task if set, otherwise completed Task 
+            // Return TaskCompletionSource's Task if set, otherwise completed Task
             return tcs?.Task ?? TaskCache.CompletedTask;
         }
 
--- a/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Infrastructure/KestrelEventSource.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Infrastructure/KestrelEventSource.cs
@@ -1,0 +1,73 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Diagnostics.Tracing;
+using System.Runtime.CompilerServices;
+using Microsoft.AspNetCore.Server.Kestrel.Internal.Http;
+
+namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Infrastructure
+{
+    [EventSource(Name = "Microsoft-AspNetCore-Server-Kestrel")]
+    public sealed class KestrelEventSource : EventSource
+    {
+        public static readonly KestrelEventSource Log = new KestrelEventSource();
+
+        private KestrelEventSource()
+        {
+        }
+
+        // NOTE
+        // - The 'Start' and 'Stop' suffixes on the following event names have special meaning in EventSource. They
+        //   enable creating 'activities'.
+        //   For more information, take a look at the following blog post:
+        //   https://blogs.msdn.microsoft.com/vancem/2015/09/14/exploring-eventsource-activity-correlation-and-causation-features/
+        // - A stop event's event id must be next one after its start event.
+        // - Avoid renaming methods or parameters marked with EventAttribute. EventSource uses these to form the event object.
+
+        [NonEvent]
+        public void ConnectionStart(Connection connection)
+        {
+            // avoid allocating strings unless this event source is enabled
+            if (IsEnabled())
+            {
+                ConnectionStart(
+                    connection.ConnectionId,
+                    connection.ListenerContext.ListenOptions.Scheme,
+                    connection.LocalEndPoint.ToString(),
+                    connection.RemoteEndPoint.ToString());
+            }
+        }
+
+        [MethodImpl(MethodImplOptions.NoInlining)]
+        [Event(1, Level = EventLevel.Verbose)]
+        private void ConnectionStart(string connectionId,
+            string scheme,
+            string localEndPoint,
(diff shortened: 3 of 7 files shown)


src/Microsoft.AspNetCore.Server.Kestrel/Internal/Infrastructure/KestrelEventSource.cs
```diff
@@ -0,0 +1,62 @@
...
+        [NonEvent]
+        public void ConnectionStop(Connection connection)
+        {
+            if (IsEnabled())
+            {
+                ConnectionStop(connection.ConnectionId);
+            }
+        }
+
+        [Event(1, Level = EventLevel.Informational)]
```
kichalla: Not sure if we want to change the level here. Request start/stop (they are informational currently) are probably more common to be used than connection information. 'Verbose' maybe?

natemcmaster: Thoughts @shirhatti? I don't have enough context to know if `EventLevel.Verbose` is more appropriate than `EventLevel.Informational`. This entirely depends on the desired UX.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Connection.cs
```diff
@@ -86,6 +86,7 @@ internal Connection()
         public void Start()
         {
             Log.ConnectionStart(ConnectionId);
+            KestrelEventSource.Log.ConnectionStart(this);
```
cesarbs: Can we move this into `KestrelTrace`?

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/EventSourceTests.cs
```diff
@@ -0,0 +1,86 @@
...
+            Assert.Equal(listenOptions.IPEndPoint.ToString(), GetProperty(start, "localEndPoint"));
+            Assert.Equal(connectionId, GetProperty(start, "connectionId"));
+
+            var stop = Assert.Single(events, e => e.EventName == "ConnectionStop");
+            Assert.All(new[] { "connectionId" }, p => Assert.Contains(p, start.PayloadNames));
+            Assert.Equal("connectionId", Assert.Single(stop.PayloadNames));
+            Assert.Equal(connectionId, GetProperty(stop, "connectionId"));
+            Assert.Same(KestrelEventSource.Log, start.EventSource);
+        }
+        private string GetProperty(EventWrittenEventArgs data, string propName)
```
cesarbs: nit: newline

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/EventSourceTests.cs
```diff
@@ -0,0 +1,86 @@
...
+    {
+        [Fact]
+        public Task LogsConnection()
+        {
+            return TestEventSource(new ListenOptions(new IPEndPoint(IPAddress.Loopback, 0)));
+        }
+
+        [ConditionalFact]
+        [IPv6SupportedCondition]
+        public Task LogsIpv6Connection()
```
cesarbs: nit: `IP`

Why do you need to test both address families?

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Connection.cs
natemcmaster: Event source is conceptually different from logging. It didn't seem appropriate to be event sourcing inside the logger implementation.

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/EventSourceTests.cs
natemcmaster: When I was playing around with TestServer, I noticed that TestServer always assumes IPv4. I wrote this to validate TestServer used ipv6. This test isn't really necessary so I'll back it out, but I'd like to keep the TestServer changes.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Infrastructure/KestrelEventSource.cs
shirhatti: I agree with @kichalla, let's make this `EventLevel.Verbose`
It also seems more consistent with what we log to `ILogger` where Request Start/Stop is `LogLevel.Information` whereas Connection Start/Stop is `LogLevel.Debug`

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Infrastructure/KestrelEventSource.cs
```diff
@@ -0,0 +1,68 @@
...
+                connectionId,
+                scheme,
+                localEndPoint,
+                remoteEndPoint
+            );
+        }
+
+        [NonEvent]
+        public void ConnectionStop(Connection connection)
+            => ConnectionStop(connection.ConnectionId);
```
benaadams: `if (IsEnabled())`?

benaadams: `if (IsEnabled())`? Else you are allocating the `ConnectionId`

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Infrastructure/KestrelEventSource.cs
natemcmaster: Updated. I set it to `Verbose`

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Infrastructure/KestrelEventSource.cs
natemcmaster: updated

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/SocketOutput.cs
```diff
@@ -673,6 +673,7 @@ private void DoDisconnectIfNeeded()
                 Self._socket.Dispose();
                 Self._connection.OnSocketClosed();
                 Self._log.ConnectionStop(Self._connectionId);
+                KestrelEventSource.Log.ConnectionStop(Self._connection);
```
natemcmaster: 🔔 @halter73. Is this the right place to put this? We need to ensure this is _always_ be called for any connection successfully opened.

halter73: Normally I would put stuff like this in the Connection class. So Connection.OnSocketClosed() would probably be better.

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/EventSourceTests.cs
```diff
@@ -0,0 +1,75 @@
...
+            }
+
+            // capture list here as other tests executing in parallel may log events
+            Assert.NotNull(connectionId);
+            var events = listener.EventData.Where(e => e != null && GetProperty(e, "connectionId") == connectionId).ToList();
+
+            var start = Assert.Single(events, e => e.EventName == "ConnectionStart");
+            Assert.All(new[] { "connectionId", "scheme", "remoteEndPoint", "localEndPoint" }, p => Assert.Contains(p, start.PayloadNames));
+            Assert.Equal("http", GetProperty(start, "scheme"));
+            Assert.Equal(listenOptions.IPEndPoint.ToString(), GetProperty(start, "localEndPoint"));
```
halter73: Nit: Should we not verify that the remoteEndPoint is 127.0.0.1?

natemcmaster:
:up: :date:

Issue #1498, opened 2017-03-17
Check if request is aborted before verifying response bytes written


cesarbs:
See #1376 for details.

PR #1499, opened 2017-03-17
Check if request is aborted before verifying response bytes written (#1498)

cesarbs:
#1498

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/FrameOfT.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/FrameOfT.cs
@@ -94,7 +94,11 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Http
                             try
                             {
                                 await _application.ProcessRequestAsync(context).ConfigureAwait(false);
-                                VerifyResponseContentLength();
+
+                                if (Volatile.Read(ref _requestAborted) == 0)
+                                {
+                                    VerifyResponseContentLength();
+                                }
                             }
                             catch (Exception ex)
                             {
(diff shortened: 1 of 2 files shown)


src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/FrameOfT.cs
```diff
@@ -94,7 +94,11 @@ public Frame(IHttpApplication<TContext> application, ConnectionContext context)
                             try
                             {
                                 await _application.ProcessRequestAsync(context).ConfigureAwait(false);
-                                VerifyResponseContentLength();
+
+                                if (Volatile.Read(ref _requestAborted) == 0)
```
Tratcher: Is this set for both dead sockets and anything that's called Abort?

cesarbs: Yes.

cesarbs:
Checking Travis failure.

davidfowl:
@cesarbs file a bug for this failure. After all of this churn we need to make sure the tests are stable again.

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/ResponseTests.cs
```diff
@@ -659,12 +662,55 @@ public Task ResponseStatusCodeSetBeforeHttpContextDisposedRequestMalformedReadIg
...
+            {
+                httpContext.RequestAborted.Register(() =>
+                {
+                    abortTcs.SetResult(null);
+                });
+                httpContext.Response.ContentLength = 12;
+                await httpContext.Response.WriteAsync("hello,");
+
+                // Wait until request is aborted so we know Frame will skip the response content length check.
+                await abortTcs.Task.TimeoutAfter(TimeSpan.FromSeconds(10));
```
halter73: Would this test still pass of this timed out after 10s without being aborted? If so, make it so it fails if the request doesn't get aborted. I hat tests that time out and silently pass.

cesarbs: Good catch. Updated.

Issue #1027, opened 2016-08-03
Remove deprecated property KestrelServerOptions.MaxRequestBufferSize


cesarbs:
Will be deprecated once https://github.com/aspnet/KestrelHttpServer/pull/1022 is in.

cesarbs:
cc @muratg

PR #1500, opened 2017-03-17
Remove deprecated property KestrelServerOptions.MaxRequestBufferSize (#1027)

cesarbs:
#1027

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Kestrel.Core/KestrelServerOptions.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel.Core/KestrelServerOptions.cs
@@ -35,32 +35,6 @@ namespace Microsoft.AspNetCore.Server.Kestrel
         public IServiceProvider ApplicationServices { get; set; }
 
         /// <summary>
-        /// <para>
-        /// This property is obsolete and will be removed in a future version.
-        /// Use <c>Limits.MaxRequestBufferSize</c> instead.
-        /// </para>
-        /// <para>
-        /// Gets or sets the maximum size of the request buffer.
-        /// </para>
-        /// </summary>
-        /// <remarks>
-        /// When set to null, the size of the request buffer is unlimited.
-        /// Defaults to 1,048,576 bytes (1 MB).
-        /// </remarks>
-        [Obsolete("This property is obsolete and will be removed in a future version. Use Limits.MaxRequestBufferSize instead.")]
-        public long? MaxRequestBufferSize
-        {
-            get
-            {
-                return Limits.MaxRequestBufferSize;
-            }
-            set
-            {
-                Limits.MaxRequestBufferSize = value;
-            }
-        }
-
-        /// <summary>
         /// Provides access to request limit options.
         /// </summary>
         public KestrelServerLimits Limits { get; } = new KestrelServerLimits();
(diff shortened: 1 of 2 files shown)


halter73:
You can just look at the existing exceptions.net45.json and exceptions.netcore.json and copy the pattern for removing `KestrelServerOptions.ConnectionFilter`. I made this change when the API check was still being run.

I think it will look like this:

```
  {
    "OldTypeId": "public class Microsoft.AspNetCore.Server.Kestrel.KestrelServerOptions",
    "OldMemberId": "public System.Nullable<System.Int64> get_MaxRequestBufferSize()",
    "Kind": "Removal"
  },
  {
    "OldTypeId": "public class Microsoft.AspNetCore.Server.Kestrel.KestrelServerOptions",
    "OldMemberId": "public System.Void set_MaxRequestBufferSize(System.Nullable<System.Int64> value)",
    "Kind": "Removal"
  },
```

If it's wrong we can fix it once the API check tool gets run.

dougbu:
> how close are you to having API check ready for MSBuild?

Close but I've got another item to complete which is higher priority. I expect to get a PR for API Check next Monday or Tuesday.

Actually, the main thing blocking check-in is determining which repos or projects (when only one or two are broken in a repo) need API checks disabled initially. I'll probably run a Universe build over the weekend to get some idea of the current state. Worst case, could merge the feature but leave it off everywhere by default.

halter73:
@cesarbs Are you going to merge this soon?

cesarbs:
@halter73 I will once I figure out why ApiCheck gives me these error messages:

```
ERROR: The following exclusion is in the exclusion file, but is no longer necessary:
    Old type: public class Microsoft.AspNetCore.Server.Kestrel.KestrelServerOptions
    Old member: public System.Nullable<System.Int64> get_MaxRequestBufferSize()
    New type:
    New member:

ERROR: The following exclusion is in the exclusion file, but is no longer necessary:
    Old type: public class Microsoft.AspNetCore.Server.Kestrel.KestrelServerOptions
    Old member: public System.Void set_MaxRequestBufferSize(System.Nullable<System.Int64> value)
    New type:
    New member:
```

This is how I ran it:

```
dotnet run -f netcoreapp1.1 compare -b D:\src\aspnet\KestrelHttpServer\src\Microsoft.AspNetCo
re.Server.Kestrel\baseline.netcore.json -e D:\src\aspnet\KestrelHttpServer\src\Microsoft.AspNetCore.Server.Kestrel\exceptions.netcore.json -a D:\src\aspnet\Kest
relHttpServer\src\Microsoft.AspNetCore.Server.Kestrel\bin\Debug\netstandard1.3\Microsoft.AspNetCore.Server.Kestrel.dll -p D:\src\aspnet\KestrelHttpServer\src\Mi
crosoft.AspNetCore.Server.Kestrel\obj\project.assets.json -f netstandard1.3 -epi
```

cesarbs:
Updated. At this point, we don't care about API check anymore right?

dougbu:
> At this point, we don't care about API check anymore right?

We care in general. However, I believe __the powers that be:tm:__ 👑 decided to disable the checks for this repo. Surprised a commit to do that hasn't already gone in -- if I'm correct.

davidfowl:
@cesarbs merge it 😄

PR #1514, opened 2017-03-20
Add global.json to .gitignore

cesarbs:


halter73:
@natemcmaster Did you say there were any cons to doing this? Like having to run build.cmd before opening in VS or something?

natemcmaster:
> Like having to run build.cmd before opening in VS or something?

Unclear at the moment. VS may require this file, but not sure yet.

PR #1515, opened 2017-03-20
Fix benchmarks (broken)

benaadams:
Benchmarks are broken and error when run
ResponseHeadersWritingBenchmark isn't testing common path as it doesn't set httpVersion or keepAlive

PR #1516, opened 2017-03-20
Use correct config for response buffer limit

halter73:


cesarbs:
Check Travis/AppVeyor before merging.

benaadams:
This is causing the benchmarks to throw in the Frame.ctor with null ref because `ServerOptions` is null

https://github.com/aspnet/KestrelHttpServer/blob/dev/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs#L87-L93

benaadams:
Added fix to https://github.com/aspnet/KestrelHttpServer/pull/1515

PR #1525, opened 2017-03-21
Get statusBytes after final potential rejection

benaadams:
Status code can be set to 500 in `CreateResponseHeader` by
```csharp
else if (hasTransferEncoding)
{
    RejectNonBodyTransferEncodingResponse(appCompleted);
}
```

So statusBytes should be generated after this

/cc @cesarbs @halter73


davidfowl:
@benaadams I basically made this change here https://github.com/aspnet/KestrelHttpServer/pull/1526 😄

davidfowl:
Fixed in #1526

Issue #1296, opened 2017-01-09
Kestrel doesn't log now listening on correctly for https


JunTaoLuo:
I configured the basic middleware RewriteSample with
```
var host = new WebHostBuilder()
                .UseKestrel(options =>
                {
                    options.Listen(IPAddress.Loopback, 5000);
                    options.Listen(IPAddress.Loopback, 5001, listenOptions =>
                    {
                        listenOptions.UseHttps("testCert.pfx", "testPassword");
                    });
                })
```
When I run the sample I see the output:
```
Hosting environment: Production
Content root path: C:\gh\BasicMiddleware\samples\RewriteSample
Now listening on: http://127.0.0.1:5000
Now listening on: http://127.0.0.1:5001
Application started. Press Ctrl+C to shut down.
```

the https url should be `https://127.0.0.1:5001`

muratg:
@natemcmaster please ping @halter73 before you start looking into this.

natemcmaster:
Related work: 

https://github.com/aspnet/Hosting/issues/968
https://github.com/aspnet/KestrelHttpServer/issues/1519
https://github.com/aspnet/KestrelHttpServer/issues/1495

cesarbs:
Given #1519, I think you can close this.

natemcmaster:
I've already forgotten what we decided to do here. Was there going to be a new field on `IServerAddressesFeature` for scheme that Kestrel can set when appropriate? cc @JunTaoLuo @Tratcher

Tratcher:
None of the design changes we discussed affect this issue, nor does #1519.

```
    options.Listen(IPAddress.Loopback, 5001, listenOptions =>
    {
        listenOptions.UseHttps("testCert.pfx", "testPassword");
        listenOptions.UseConnectionLogging();
    });
```
Should result in IServerAddressesFeature values being replaced with "https://127.0.0.1:5001/".
(unless IServerAddressesFeature has the new PreferHostingConfig option enabled, in which case you ignore Listen).

natemcmaster:
Took a look at this again. Here are two options

**The simple one**
Make `ListenOptions.Scheme` public & settable. UseHttps overloads set `options.Scheme = "https"`.

**The make-an-abstraction one** 
Something like
```c#
public interface IServerAddressProvider
{
   // DisplayName = trying to communicate that the value is only used in console output or loggers
   string GetServerAddressDisplayName();
}
```

`KestrelServer.Start` would 

[...]

PR #1533, opened 2017-03-21
Throw if UseUrls specifies HTTPS or path base (#1519)

cesarbs:
#1519

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Kestrel.Core/KestrelServer.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel.Core/KestrelServer.cs
@@ -146,6 +146,16 @@ namespace Microsoft.AspNetCore.Server.Kestrel
                     {
                         var parsedAddress = ServerAddress.FromUrl(address);
 
+                        if (parsedAddress.Scheme.Equals("https", StringComparison.OrdinalIgnoreCase))
+                        {
+                            throw new InvalidOperationException($"HTTPS endpoints can only be configured using {nameof(KestrelServerOptions)}.{nameof(KestrelServerOptions.Listen)}().");
+                        }
+
+                        if (!string.IsNullOrEmpty(parsedAddress.PathBase))
+                        {
+                            throw new InvalidOperationException($"A path base can only be configured using {nameof(IApplicationBuilder)}.UsePathBase().");
+                        }
+
                         if (!string.IsNullOrEmpty(parsedAddress.PathBase))
                         {
                             _logger.LogWarning($"Path base in address {address} is not supported and will be ignored. To specify a path base, use {nameof(IApplicationBuilder)}.UsePathBase().");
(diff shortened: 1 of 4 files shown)


src/Microsoft.AspNetCore.Server.Kestrel/KestrelServer.cs
```diff
@@ -164,9 +164,14 @@ public void Start<TContext>(IHttpApplication<TContext> application)
                     {
                         var parsedAddress = ServerAddress.FromUrl(address);
 
+                        if (parsedAddress.Scheme.Equals("https", StringComparison.OrdinalIgnoreCase))
+                        {
+                            throw new InvalidOperationException($"HTTPS addresses are not supported. Use {nameof(KestrelServerOptions)}.{nameof(KestrelServerOptions.Listen)}() to configure an HTTPS endpoint.");
+                        }
+
                         if (!string.IsNullOrEmpty(parsedAddress.PathBase))
```
Tratcher: What happens if it's just '/'?
Where does it throw? This just logs.

cesarbs: Oops 😄

cesarbs: There's a `ServerAddress` test that checks that just '/' is not set as path base: https://github.com/aspnet/KestrelHttpServer/blob/632780dd16729298933f96796e6a969f3470d810/test/Microsoft.AspNetCore.Server.KestrelTests/ServerAddressTests.cs#L42

src/Microsoft.AspNetCore.Server.Kestrel/KestrelServer.cs
```diff
@@ -164,9 +164,14 @@ public void Start<TContext>(IHttpApplication<TContext> application)
                     {
                         var parsedAddress = ServerAddress.FromUrl(address);
 
+                        if (parsedAddress.Scheme.Equals("https", StringComparison.OrdinalIgnoreCase))
+                        {
+                            throw new InvalidOperationException($"HTTPS addresses are not supported. Use {nameof(KestrelServerOptions)}.{nameof(KestrelServerOptions.Listen)}() to configure an HTTPS endpoint.");
```
halter73: I don't like the phrasing of "HTTPS addresses are not supported..."

I would change this to "HTTPS endpoints may only be configured using {nameof(KestrelServerOptions)}.{nameof(KestrelServerOptions.Listen)}()".

Tratcher: may?

halter73: can, must, whatever.

cesarbs:
@halter73 I've changed the path base message too, for consistency.

natemcmaster:
Does this incidentally fix https://github.com/aspnet/KestrelHttpServer/issues/1296 too?

cesarbs:
@natemcmaster No. See my updated comment there.

cesarbs:
Updated.

Issue #1534, opened 2017-03-21
Test failure: macOS, RegisterAddresses_IPv6ScopeId_Success timeout 


natemcmaster:
https://travis-ci.org/aspnet/KestrelHttpServer/jobs/213529760#L1293

```
Failed   Microsoft.AspNetCore.Server.Kestrel.FunctionalTests.AddressRegistrationTests.RegisterAddresses_IPv6ScopeId_Success(addressInput: "http://[fe80::88fc:3f5a:f0fb:3da1%utun0]:0/", testUrls: Func`2 { Method = System.String[] GetTestUrls(Microsoft.AspNetCore.Hosting.Server.Features.IServerAddressesFeature), Target = null })
Error Message:
 System.Net.Sockets.SocketException : Operation timed out
Stack Trace:
   at Microsoft.AspNetCore.Testing.HttpClientSlim.<GetSocket>d__7.MoveNext() in /Users/travis/build/aspnet/KestrelHttpServer/test/shared/HttpClientSlim.cs:line 114
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
```

halter73:
This is now failing consistently on Travis and our internal macOS CI. The utun[0-9] scopeids appear to be years old, but I still think something must have changed in a recent OS update.

muratg:
Fixed

PR #1536, opened 2017-03-22
Verify scopeids are connectable using Socket before testing Kestrel

halter73:
#1534

@mikeharder @cesarbs @natemcmaster

### DIFF
--- a/test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/AddressRegistrationTests.cs
+++ b/test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/AddressRegistrationTests.cs
@@ -8,6 +8,7 @@ using System.Linq;
 using System.Net;
 using System.Net.NetworkInformation;
 using System.Net.Sockets;
+using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Builder;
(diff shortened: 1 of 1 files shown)


halter73:
I've verified that on our macOS agent, this will skip the "utun0" scopeid while still testing "en0" and "lo0".

This doesn't cause anything new to be skipped on Windows.

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/AddressRegistrationTests.cs
```diff
@@ -531,6 +534,37 @@ private static int GetNextPort()
...
+
+                    var mre = new ManualResetEventSlim();
+                    socketArgs.Completed += (s, e) =>
+                    {
+                        mre.Set();
+                        e.ConnectSocket.Dispose();
+                    };
+
+                    // Connect can take a couple minutes to time out.
+                    return Socket.ConnectAsync(SocketType.Stream, ProtocolType.Tcp, socketArgs) && mre.Wait(5000);
```
cesarbs: `ConnectAsync` will return false if the operation completes synchronously. In that case the callback is not called and you have to call it manually. You have to check the result in the callback to make sure the connect really succeeded.

Issue #1530, opened 2017-03-21
"Unexpected end of request content" preventing terminating chunk from being sent


stephentoub:
More details here:
https://github.com/dotnet/corefx/issues/17330#issuecomment-288078562

In summary, some clients (e.g. libcurl) stop sending request content if they concurrently receive an error status code in response from the server.  If an app on kestrel ignores the request data and immediately sends a response with an error status, the client may as such stop sending the request data, which in turn causes kestrel to fail with a BadHttpRequestException, but it doesn't complete the chunked encoding response by sending a terminating 0-length chunk, nor does it close the connection.  As a result, the client is left waiting indefinitely for the remainder of the response that never comes, and effectively deadlocks until timeout.

davidfowl:
Were you using 1.1?

stephentoub:
Yes:
```
  <PropertyGroup>
    <TargetFramework>netcoreapp1.1</TargetFramework>
  </PropertyGroup>

  <ItemGroup>
    <Folder Include="wwwroot\" />
  </ItemGroup>

  <ItemGroup>
    <PackageReference Include="Microsoft.AspNetCore" Version="1.1.1" />
  </ItemGroup>
```

cesarbs:
Reading through the other issue, will have an answer soon.

cesarbs:
Self-assigning for investigation. @stephentoub see https://github.com/dotnet/corefx/issues/17330#issuecomment-288168960

cesarbs:
I thought we'd shipped "FIN as connection reset" in Kestrel 1.1 but we didn't. This was the change: https://github.com/aspnet/KestrelHttpServer/pull/1218

The current behavior is bad though, in that the server at least do *something*. Given 1.1 doesn't close the connection after the client FIN, it should attempt to complete the response. This may be a candidate for a patch, but I need to assess how large the change would be. cc @muratg @Eilon

muratg:
Filed https://github.com/aspnet/KestrelHttpServer/issues/1535 to track this for 1.1.2 patch release.

cesarbs:
Diagnosis here: https://github.com/dotnet/corefx/issues/17330#issuecomment-288248663

lewislabs:
We're still seeing this error when using Kestrel 1.1.2. Getting a `BadHttpRequestException` with the message `Unexpected end of request content` being thrown with this StackTrace. 
```
[...]

PR #1537, opened 2017-03-22
Call ProduceEnd() before consuming request body if response has already started (#1530)

cesarbs:
#1530

Prevents hanging clients waiting for the chunk terminator while server consumes the rest of the request body.

See https://github.com/dotnet/corefx/issues/17330#issuecomment-288248663.

cc @stephentoub

### DIFF
--- a/test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/ResponseTests.cs
+++ b/test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/ResponseTests.cs
@@ -341,7 +341,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.FunctionalTests
             {
                 readException = await Assert.ThrowsAsync<BadHttpRequestException>(
                     async () => await httpContext.Request.Body.ReadAsync(new byte[1], 0, 1));
-            }, new TestServiceContext()))
+            }))
             {
                 using (var connection = server.CreateConnection())
                 {
@@ -370,7 +370,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.FunctionalTests
             {
                 await httpContext.Response.WriteAsync("hello, ");
                 await httpContext.Response.WriteAsync("world");
-            }, new TestServiceContext()))
+            }))
             {
                 using (var connection = server.CreateConnection())
                 {
@@ -404,7 +404,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.FunctionalTests
             {
                 httpContext.Response.StatusCode = statusCode;
                 return TaskCache.CompletedTask;
-            }, new TestServiceContext()))
+            }))
             {
                 using (var connection = server.CreateConnection())
                 {
@@ -427,7 +427,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.FunctionalTests
             using (var server = new TestServer(httpContext =>
             {
                 return TaskCache.CompletedTask;
-            }, new TestServiceContext()))
+            }))
             {
                 using (var connection = server.CreateConnection())
                 {
@@ -880,7 +880,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.FunctionalTests
             {
                 httpContext.Response.ContentLength = 42;
                 return TaskCache.CompletedTask;
-            }, new TestServiceContext()))
+            }))
             {
                 using (var connection = server.CreateConnection())
                 {
@@ -908,7 +908,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.FunctionalTests
                 httpContext.Response.ContentLength = 12;
                 await httpContext.Response.WriteAsync("hello, world");
                 await flushed.WaitAsync();
-            }, new TestServiceContext()))
+            }))
             {
                 using (var connection = server.CreateConnection())
                 {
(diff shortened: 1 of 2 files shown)


davidfowl:
@cesarbs can you rebase on top of @halter73 's change

```
Failed   Microsoft.AspNetCore.Server.Kestrel.FunctionalTests.AddressRegistrationTests.RegisterAddresses_IPv6ScopeId_Success(addressInput: "http://[fe80::1208:2b32:1916:e4b0%utun0]:0/", testUrls: Func`2 { Method = System.String[] GetTestUrls(Microsoft.AspNetCore.Hosting.Server.Features.IServerAddressesFeature), Target = null })
```

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/FrameOfT.cs
```diff
@@ -133,16 +133,28 @@ public Frame(IHttpApplication<TContext> application, ConnectionContext context)
                             {
                                 ResumeStreams();
 
+                                if (HasResponseStarted)
+                                {
+                                    // If the response has already started, call ProduceEnd() before
```
davidfowl: Point to the bug in here

cesarbs:
@davidfowl Rebased.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/FrameOfT.cs
cesarbs: Pointed to the explanation in the corefx bug.

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/ResponseTests.cs
```diff
@@ -1368,6 +1368,116 @@ public Task ResponseStatusCodeSetBeforeHttpContextDisposedRequestMalformedReadIg
...
 
+        [Fact]
+        public async Task WhenResponseAlreadyStartedResponseEndedBeforeDrainingRequestBody()
+        {
+            using (var server = new TestServer(async httpContext =>
+            {
+                await httpContext.Response.WriteAsync("hello, world");
+
+                // Force response start
+                await httpContext.Response.Body.FlushAsync();
```
davidfowl: You don't have to call FlushAsync to start the response, that happens on write

davidfowl:
Can you make sure we have a test for all of the relevant message body types? (Opaque, ContentLength, Chunked)

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/ResponseTests.cs
```diff
@@ -1368,6 +1368,113 @@ public Task ResponseStatusCodeSetBeforeHttpContextDisposedRequestMalformedReadIg
...
+                        $"Transfer-Encoding: chunked",
+                        "",
+                        "c",
+                        "hello, world",
+                        "");
+
+                    // If the expected behavior is regressed, this will hang because the
+                    // server will try to consume the request body before flushing the chunked
+                    // terminator.
+                    await connection.Receive(
```
halter73: Did you split this into two calls to connection.Receive() just to insert the above comment?

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/ResponseTests.cs
```diff
@@ -1368,6 +1368,113 @@ public Task ResponseStatusCodeSetBeforeHttpContextDisposedRequestMalformedReadIg
...
+                        "0",
+                        "",
+                        "");
+                }
+            }
+        }
+
+        [Fact]
+        public async Task WhenResponseNotStartedResponseEndedAfterDrainingRequestBody()
```
halter73: Good test 👍

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/ResponseTests.cs
```diff
@@ -1368,6 +1368,113 @@ public Task ResponseStatusCodeSetBeforeHttpContextDisposedRequestMalformedReadIg
...
+                        $"Date: {server.Context.DateHeaderValue}",
+                        "Content-Length: 0",
+                        "",
+                        "");
+                }
+            }
+        }
+
+        [Fact]
+        public async Task Sending100ContinueDoesNotStartResponse()
```
halter73: Can we test sending the chunk terminator pre-consume when we send a 100-continue and the app sends a response?

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/ResponseTests.cs
cesarbs: No. It's to make it clear the first part will be received fine, but if there's a regression the connection will time out waiting for the chunk terminator.

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/ResponseTests.cs
cesarbs: Added.

cesarbs:
Updated.

cesarbs:
@davidfowl This is unrelated to the message body type, so having tests per type won't add value. What we care about here is the order in which things happen.

cesarbs:
📢

PR #1538, opened 2017-03-22
Optimize calls into StreamSocketOuput to WriteFast further

davidfowl:
- Added overloads to array, offset, length to avoid implicit conversions to
ReadOnlySpan.
- Use similar optimizations for multi buffer writes for strings and ints
- Use ref locals in multi write instead of pointers and pinning

## dev

```
 Method |                       Type |      Mean |    StdDev |      Op/s | Allocated |
------- |--------------------------- |---------- |---------- |---------- |---------- |
 Output |                 LiveAspNet | 3.9868 us | 0.1728 us | 250824.84 |      69 B |
 Output |           PlaintextChunked | 2.3238 us | 0.0770 us | 430328.37 |      66 B |
 Output | PlaintextChunkedWithCookie | 2.9382 us | 0.1158 us | 340345.04 |      72 B |
 Output |        PlaintextWithCookie | 3.1480 us | 0.1054 us | 317657.94 |      72 B |
 Output |       TechEmpowerPlaintext | 2.4437 us | 0.0840 us | 409209.51 |      67 B |
```

## PR

```
 Method |                       Type |      Mean |    StdDev |      Op/s | Allocated |
------- |--------------------------- |---------- |---------- |---------- |---------- |
 Output |                 LiveAspNet | 3.7100 us | 0.1729 us | 269539.10 |      69 B |
 Output |           PlaintextChunked | 2.0887 us | 0.0640 us | 478769.50 |      66 B |
 Output | PlaintextChunkedWithCookie | 2.7416 us | 0.0796 us | 364744.34 |      71 B |
 Output |        PlaintextWithCookie | 2.7983 us | 0.0673 us | 357357.40 |      72 B |
 Output |       TechEmpowerPlaintext | 2.2297 us | 0.0660 us | 448499.22 |      67 B |
 ```

### CODE REVIEW

File: src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/PipelineExtensions.cs [hunk 1]
```diff
@@ -94,7 +94,17 @@ public static ArraySegment<byte> GetArray(this Buffer<byte> buffer)
         }
 
         // Temporary until the fast write implementation propagates from corefx
-        public unsafe static void WriteFast(this WritableBuffer buffer, ReadOnlySpan<byte> source)
+        public unsafe static void WriteFast(this WritableBuffer buffer, byte[] source)
```
benaadams: This version get called much? If so add direct implementation of it

```chsarp
public unsafe static void WriteFast(this WritableBuffer buffer, byte[] source)
{
    var dest = buffer.Buffer.Span;
    var destLength = dest.Length;

    if (destLength == 0)
    {
        buffer.Ensure();

        // Get the new span and length
        dest = buffer.Buffer.Span;
        destLength = dest.Length;
    }

    var sourceLength = source.length;
    if (sourceLength <= destLength)
    {
        ref byte pSource = ref source[0];
        ref byte pDest = ref dest.DangerousGetPinnableReference();
        Unsafe.CopyBlockUnaligned(ref pDest, ref pSource, (uint)sourceLength);
        buffer.Advance(sourceLength);
        return;
    }

    buffer.WriteMultiBuffer(source, offset, sourceLength);
}
```

davidfowl: Really? What about a shared method that takes ref byte?

benaadams: nvm; is 4 args which is under pass by reg limit

PR #1539, opened 2017-03-22
React to corefxlab changes

davidfowl:
- Removed use of TryRead and follow the pin and use
pattern in the other APIs

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Infrastructure/HttpUtilities.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Infrastructure/HttpUtilities.cs
@@ -270,23 +270,32 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Infrastructure
         /// <param name="knownScheme">A reference to the known scheme, if the input matches any</param>
         /// <returns>True when memory starts with known http or https schema</returns>
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
-        public static bool GetKnownHttpScheme(this Span<byte> span, out HttpScheme knownScheme)
+        public static unsafe bool GetKnownHttpScheme(this Span<byte> span, out HttpScheme knownScheme)
         {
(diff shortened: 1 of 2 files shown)


NuGet.config
```diff
@@ -1,6 +1,8 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <configuration>
   <packageSources>
+    <add key="CoreFxLabs" value="https://dotnet.myget.org/F/dotnet-corefxlab/api/v3/index.json" />
```
davidfowl: This won't be merged. This branch just needs to exist for the mirror.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Infrastructure/HttpUtilities.cs
```diff
@@ -270,23 +270,32 @@ internal unsafe static HttpVersion GetKnownVersion(byte* location, int length)
         /// <param name="knownScheme">A reference to the known scheme, if the input matches any</param>
         /// <returns>True when memory starts with known http or https schema</returns>
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
-        public static bool GetKnownHttpScheme(this Span<byte> span, out HttpScheme knownScheme)
+        public static unsafe bool GetKnownHttpScheme(this Span<byte> span, out HttpScheme knownScheme)
         {
-            if (span.TryRead<ulong>(out var value))
+            fixed (byte* data = &span.DangerousGetPinnableReference())
```
natemcmaster: Did they remove this API or is this a perf tweak?

davidfowl: It moved to another assembly and I didn't want to take a new dependency.

natemcmaster: ok, i don't know enough about what corefx is actually planning to release so this is fine for now.

Issue #1520, opened 2017-03-20
Race with AddressRegistrationTests


JunTaoLuo:
Seeing this failure on the CI http://aspnetci/viewLog.html?buildId=189191&tab=buildResultsDiv&buildTypeId=Lite_UniverseTest.

Not sure what which tests this race is between but it seems like some other app that is returning `hello, world` is being run in parallel with the AddressRegistrationTests and collided on the same port.
```
[05:12:36] :			 [Microsoft.AspNetCore.Server.Kestrel.FunctionalTests.AddressRegistrationTests.RegisterAddresses_IPv6_Success(addressInput: "http://*:8004/", testUrls: Func`2 { Method = System.String[] <get_AddressRegistrationDataIPv6>b__4(Microsoft.AspNetCore.Hosting.Server.Features.IServerAddressesFeature), Target = <>c__DisplayClass27_0 { port = 8004 } })] Assert.Equal() Failure
             (pos 1)
Expected: http://127.0.0.1:8004/
Actual:   hello, world
             (pos 1)
   at Microsoft.AspNetCore.Server.Kestrel.FunctionalTests.AddressRegistrationTests.<RegisterAddresses_Success>d__7.MoveNext() in C:\BuildAgent\work\33bdfc1cae7b2a38\.r\KestrelHttpServer\test\Microsoft.AspNetCore.Server.Kestrel.FunctionalTests\AddressRegistrationTests.cs:line 98
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at Microsoft.AspNetCore.Server.Kestrel.FunctionalTests.AddressRegistrationTests.<RegisterAddresses_IPv6_Success>d__4.MoveNext() in C:\BuildAgent\work\33bdfc1cae7b2a38\.r\KestrelHttpServer\test\Microsoft.AspNetCore.Server.Kestrel.FunctionalTests\AddressRegistrationTests.cs:line 61
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
--- End of stack trace from previous location where exception was thrown ---
[...]

PR #1552, opened 2017-03-24
Fix races on port acquisition in AddressRegistrationTests (#1520)

cesarbs:
#1520

cc @jbagga


Drawaes:
Much better...less code as well :)

cesarbs:
@JunTaoLuo It works a little differently in Linux in that it increments further than +1, but races due to port reuse are only really a problem on Windows (where it's on by default).

Drawaes:
On Linux it uses a mod and buckets, but yeah for this test it will have the same result you won't race as there is still an increasing hint.

Issue #1508, opened 2017-03-18
Write tests for methods added during improving socket output performance


pakrym:
In this PR: https://github.com/aspnet/KestrelHttpServer/pull/1504
/cc @davidfowl

muratg:
Two tests:
- Write numeric
- Write ASCII

natemcmaster:
I'm looking at the unicode issue. @davidfowl you seemed to thing the Unicode + WriteAscii method is okay. What was the justification?

Before #1504, we used Encoding.ASCII.GetBytes which does a range change on data and replaces out-of-range characters with '?'. The new implementation doesn't range check. 
Comparison:
```
Before: WriteAscii("ñ٢⛄⛵") => "????"
After:  WriteAscii("ñ٢⛄⛵") => "ñbõÄ"
```

The concern here is that it's possible this method will transform non-ascii into ascii, e.g. `٢` => `b`.
cc @benaadams

benaadams:
@natemcmaster as per current Kestrel. 

They are response headers and [ValidateHeaderCharacters](https://github.com/aspnet/KestrelHttpServer/blob/dev/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/FrameHeaders.cs#L237-L259) is run on them when they are added to the collections (so they throw an exception in user code when added if not ascii) so there is no need to check them again on output.

davidfowl:
@natemcmaster if you want to feel better about the method, you can rename it to `WriteAsciiNoValidation`

natemcmaster:
Thanks for clarifying.

PR #1551, opened 2017-03-24
Transport agnostic kestrel refactoring

halter73:
## What I plan to do as soon as I'm physically capable so this can be quickly merged to unblock people

- Get unit tests compiling and passing again
- Get performance project compiling and running again
- Run perf tests

# Possible issues/work items

## Quickish fixes

- Fix API check (exceptions.json)
- Rewrite RequestHeaderLimitsTests "functional" tests to be unit tests or not access Frame
- Fix namespaces/folders/class names (rename Microsoft.AspNetCore.Server.Kestrel.Libuv to Microsoft.AspNetCore.Server.Kestrel.Transport.Libuv)
- Create Microsoft.AspNetCore.Server.Kestrel.Transport.Abstractions project
- Add doc comments for new public classes and interfaces
- Don't use KestrelTrace class from Libuv project
- Re-enable KestrelEventSource

## Bigger work items

- Move shutdown logic to Core and make it tranport agnostic
- Move timeout logic to Core and make it tranport agnostic
- Remove the IConnectionInformation.ConnectionControl (helped by above items)
- Clean up IConnectionContext methods
- Share KestrelThreads (LibuvThreads) between endpoints (will be easier when shutdown logic is moved)
- Split transport-specific tests and general tests into distinct test projects
- Write managed socket transport to truly test the abstraction

## Notes:
- The Https project no longer depends on the meta project which contains UseKestrel(). This is a breaking change.

#828

PR #1553, opened 2017-03-24
Add tests for WritableBuffer extensions and rename WriteAscii => WriteAsciiNoValidation

natemcmaster:
Add tests for WritableBuffer.{WriteNumeric, WriteAscii}
Rename WriteAsciiNoValidation to make it clearer that it does not perform ascii validation.

Also, this changes the xunit display settings to make it easier to view tests in Test Explorer. (cref microsoft/vstest#623 )
image

Resolves https://github.com/aspnet/KestrelHttpServer/issues/1508

### DIFF
--- a/test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests.csproj
+++ b/test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests.csproj
@@ -13,6 +13,7 @@
   <ItemGroup>
     <Compile Include="..\shared\**\*.cs" />
     <Content Include="..\shared\TestResources\testCert.pfx" CopyToOutputDirectory="PreserveNewest" />
+    <None Update="xunit.runner.json" CopyToOutputDirectory="PreserveNewest" />
   </ItemGroup>
 
   <ItemGroup>
--- a/test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/xunit.runner.json
+++ b/test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/xunit.runner.json
@@ -1,6 +1,5 @@
 {
   "$schema": "http://json.schemastore.org/xunit.runner.schema",
-  "appDomain": "denied",
   "methodDisplay": "method",
   "longRunningTestSeconds": 60
 }
--- a/test/Microsoft.AspNetCore.Server.KestrelTests/PipelineExtensionTests.cs
+++ b/test/Microsoft.AspNetCore.Server.KestrelTests/PipelineExtensionTests.cs
@@ -1,0 +1,167 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.IO.Pipelines;
+using System.Text;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Server.Kestrel.Internal.Http;
+using Xunit;
+
+namespace Microsoft.AspNetCore.Server.KestrelTests
+{
+    public class PipelineExtensionTests : IDisposable
+    {
+        // ulong.MaxValue.ToString().Length
+        private const int _ulongMaxValueLength = 20;
+
+        private readonly IPipe _pipe;
+        private readonly PipeFactory _pipeFactory = new PipeFactory();
+
+        public PipelineExtensionTests()
+        {
+            _pipe = _pipeFactory.Create();
+        }
+
+        public void Dispose()
+        {
+            _pipeFactory.Dispose();
+        }
+
+        [Theory]
+        [InlineData(ulong.MinValue)]
+        [InlineData(ulong.MaxValue)]
+        [InlineData(4_8_15_16_23_42)]
+        public async Task WritesNumericToAscii(ulong number)
+        {
+            var writer = _pipe.Writer.Alloc();
+            writer.WriteNumeric(number);
+            await writer.FlushAsync();
+
+            var reader = await _pipe.Reader.ReadAsync();
+            var numAsStr = number.ToString();
+            var expected = Encoding.ASCII.GetBytes(numAsStr);
+            AssertExtensions.Equal(expected, reader.Buffer.Slice(0, numAsStr.Length).ToArray());
+        }
+
+        [Theory]
+        [InlineData(1)]
+        [InlineData(_ulongMaxValueLength / 2)]
+        [InlineData(_ulongMaxValueLength - 1)]
+        public void WritesNumericAcrossSpanBoundaries(int gapSize)
+        {
+            var writer = _pipe.Writer.Alloc(100);
+            // almost fill up the first block
+            var spacer = new Span<byte>(new byte[writer.Buffer.Length - gapSize]);
+            writer.Write(spacer);
+
+            var bufferLength = writer.Buffer.Length;
+            writer.WriteNumeric(ulong.MaxValue);
+            Assert.NotEqual(bufferLength, writer.Buffer.Length);
+
+            writer.FlushAsync().GetAwaiter().GetResult();
+
+            var reader = _pipe.Reader.ReadAsync().GetAwaiter().GetResult();
+            var numAsString = ulong.MaxValue.ToString();
+            var written = reader.Buffer.Slice(spacer.Length, numAsString.Length);
+            Assert.False(written.IsSingleSpan, "The buffer should cross spans");
+            AssertExtensions.Equal(Encoding.ASCII.GetBytes(numAsString), written.ToArray());
+        }
+
+        [Theory]
+        [InlineData("\0abcxyz", new byte[] { 0, 97, 98, 99, 120, 121, 122 })]
+        [InlineData("!#$%i", new byte[] { 33, 35, 36, 37, 105 })]
+        [InlineData("!#$%", new byte[] { 33, 35, 36, 37 })]
+        [InlineData("!#$", new byte[] { 33, 35, 36 })]
+        [InlineData("!#", new byte[] { 33, 35 })]
+        [InlineData("!", new byte[] { 33 })]
+        // null or empty
+        [InlineData("", new byte[0])]
+        [InlineData(null, new byte[0])]
+        public async Task EncodesAsAscii(string input, byte[] expected)
+        {
+            var writer = _pipe.Writer.Alloc();
+            writer.WriteAsciiNoValidation(input);
+            await writer.FlushAsync();
+            var reader = await _pipe.Reader.ReadAsync();
+
+            if (expected.Length > 0)
+            {
+                AssertExtensions.Equal(
+                    expected,
+                    reader.Buffer.ToArray());
+            }
+            else
+            {
+                Assert.Equal(0, reader.Buffer.Length);
+            }
+        }
+
+        [Theory]
+        // non-ascii characters stored in 32 bits
+        [InlineData("𤭢𐐝")]
+        // non-ascii characters stored in 16 bits
+        [InlineData("ñ٢⛄⛵")]
+        public async Task WriteAsciiNoValidationWritesOnlyOneBytePerChar(string input)
+        {
+            // WriteAscii doesn't validate if characters are in the ASCII range
+            // but it shouldn't produce more than one byte per character
+            var writer = _pipe.Writer.Alloc();
+            writer.WriteAsciiNoValidation(input);
+            await writer.FlushAsync();
+            var reader = await _pipe.Reader.ReadAsync();
+
+            Assert.Equal(input.Length, reader.Buffer.Length);
+        }
+
+        [Fact]
+        public async Task WriteAsciiNoValidation()
+        {
+            const byte maxAscii = 0x7f;
+            var writer = _pipe.Writer.Alloc();
+            for (var i = 0; i < maxAscii; i++)
+            {
+                writer.WriteAsciiNoValidation(new string((char)i, 1));
+            }
+            await writer.FlushAsync();
+
+            var reader = await _pipe.Reader.ReadAsync();
+            var data = reader.Buffer.Slice(0, maxAscii).ToArray();
+            for (var i = 0; i < maxAscii; i++)
+            {
+                Assert.Equal(i, data[i]);
+            }
+        }
+
+        [Theory]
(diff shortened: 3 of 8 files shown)


benaadams:
> Also, this changes the xunit display settings to make it easier to view the write test in Test Explorer

TFFT 👍

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests.csproj
```diff
@@ -37,6 +37,12 @@
     <Service Include="{82a7f48d-3b50-4b1e-b82e-3ada8210c358}" />
   </ItemGroup>
 
+  <ItemGroup>
+    <None Update="xunit.runner.json">
+      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
```
cesarbs: Attribute.

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/xunit.runner.json
```diff
@@ -0,0 +1,5 @@
+{
+  "$schema": "http://json.schemastore.org/xunit.runner.schema",
+  "methodDisplay": "method",
```
cesarbs: Is there a way to display as `Class.Method`?

test/Microsoft.AspNetCore.Server.KestrelTests/PipelineExtensionTests.cs
```diff
@@ -0,0 +1,149 @@
...
+using System.Text;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Server.Kestrel.Internal.Http;
+using Xunit;
+
+namespace Microsoft.AspNetCore.Server.KestrelTests
+{
+    public class PipelineExtensionTests : IDisposable
+    {
+        // ulong.MaxValue.ToString().Length
```
cesarbs: Make it `readonly` and call that here.

test/Microsoft.AspNetCore.Server.KestrelTests/PipelineExtensionTests.cs
```diff
@@ -0,0 +1,149 @@
...
+        private const int _ulongMaxValueLength = 20; 
+
+        private readonly IPipe _pipe;
```
cesarbs: Why shared state in tests? 😢

test/Microsoft.AspNetCore.Server.KestrelTests/PipelineExtensionTests.cs
```diff
@@ -0,0 +1,149 @@
...
+
+        public PipelineExtensionTests()
+        {
+            _pipe = _pipeFactory.Create();
+        }
+
+        [Theory]
+        [InlineData(ulong.MinValue)]
+        [InlineData(ulong.MaxValue)]
+        [InlineData(4_8_15_16_23_42)]
```
cesarbs: ?

test/Microsoft.AspNetCore.Server.KestrelTests/TestHelpers/AssertEx.cs
```diff
@@ -0,0 +1,27 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using Xunit.Sdk;
+
+namespace Xunit
+{
+    public static class AssertEx
```
cesarbs: Call it `AssertExtensions`.

test/Microsoft.AspNetCore.Server.KestrelTests/PipelineExtensionTests.cs
```diff
@@ -0,0 +1,149 @@
...
+        [InlineData(4, 2)]
+        [InlineData(5, 3)]
+        [InlineData(7, 4)]
+        [InlineData(8, 3)]
+        [InlineData(8, 4)]
+        [InlineData(8, 5)]
+        [InlineData(100, 48)]
+        public void WritesAsciiAcrossBlockBoundaries(int stringLength, int gapSize)
+        {
+            var testString = new String(' ', stringLength);
```
cesarbs: `string`

test/Microsoft.AspNetCore.Server.KestrelTests/PipelineExtensionTests.cs
```diff
@@ -0,0 +1,149 @@
...
+        [InlineData("ñ٢⛄⛵")]
+        public async Task WriteAsciiWritesOnlyOneBytePerChar(string input)
+        {
+            // WriteAscii doesn't validate if characters are in the ASCII range
+            // but it shouldn't produce more than one byte per character
+            var reader = await WriteAscii(input);
+            Assert.Equal(input.Length, reader.Buffer.Length);
+        }
+
+        private async Task<ReadResult> WriteAscii(string input)
```
cesarbs: Call it `WriteAsciiNoValidation` too, otherwise it's a little confusing to read.

test/Microsoft.AspNetCore.Server.KestrelTests/PipelineExtensionTests.cs
```diff
@@ -0,0 +1,149 @@
...
+                Assert.Equal(0, reader.Buffer.Length);
+            }
+        }
+
+        [Theory]
+        // non-ascii characters stored in 32 bits
+        [InlineData("𤭢𐐝")]
+        // non-ascii characters stored in 16 bits
+        [InlineData("ñ٢⛄⛵")]
+        public async Task WriteAsciiWritesOnlyOneBytePerChar(string input)
```
cesarbs: `WriteAsciiNoValidationWritesOnlyOneBytePerChar`

test/Microsoft.AspNetCore.Server.KestrelTests/PipelineExtensionTests.cs
```diff
@@ -0,0 +1,149 @@
...
+        private readonly IPipe _pipe;
+        private readonly PipeFactory _pipeFactory = new PipeFactory();
+
+        public PipelineExtensionTests()
+        {
+            _pipe = _pipeFactory.Create();
+        }
+
+        [Theory]
+        [InlineData(ulong.MinValue)]
```
cesarbs: Add case for single digit.

test/Microsoft.AspNetCore.Server.KestrelTests/PipelineExtensionTests.cs
```diff
@@ -0,0 +1,149 @@
...
+using System;
+using System.IO.Pipelines;
+using System.Text;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Server.Kestrel.Internal.Http;
+using Xunit;
+
+namespace Microsoft.AspNetCore.Server.KestrelTests
+{
+    public class PipelineExtensionTests : IDisposable
```
cesarbs: Add a `WriteAsciiNoValidation` test encoding and verifying a single char, with data from 0-255. This guaranteed to character value is special cased in that method.

test/Microsoft.AspNetCore.Server.KestrelTests/PipelineExtensionTests.cs
```diff
@@ -0,0 +1,149 @@
...
+        [Theory]
+        [InlineData("\0abcxyz", new byte[] { 0, 97, 98, 99, 120, 121, 122 })]
+        [InlineData("!#$%i", new byte[] { 33, 35, 36, 37, 105 })]
+        [InlineData("!#$%", new byte[] { 33, 35, 36, 37 })]
+        [InlineData("!#$", new byte[] { 33, 35, 36 })]
+        [InlineData("!#", new byte[] { 33, 35 })]
+        [InlineData("!", new byte[] { 33 })]
+        // null or empty
+        [InlineData("", new byte[0])]
+        [InlineData(default(string), new byte[0])]
```
cesarbs: Explicit `null`.

test/Microsoft.AspNetCore.Server.KestrelTests/PipelineExtensionTests.cs
natemcmaster: I wanted to test a random number somewhere between 0 and ulong.MaxValue. http://lost.wikia.com/wiki/The_Numbers

test/Microsoft.AspNetCore.Server.KestrelTests/PipelineExtensionTests.cs
natemcmaster: xUnit.net creates a new instance of the test class for every test that is run

test/Microsoft.AspNetCore.Server.KestrelTests/PipelineExtensionTests.cs
cesarbs: Ah, I'm not familiar with Lost references.

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/xunit.runner.json
natemcmaster: I wish. There are only two options. `"classAndMethod"` and `"method"`. The former is the default and is what results in the massively long test names in Test Explorer.

test/Microsoft.AspNetCore.Server.KestrelTests/PipelineExtensionTests.cs
natemcmaster: Like `0`? :trollface:

cesarbs: It's `ulong` not `long` that my brain parsed so ignore this.

test/Microsoft.AspNetCore.Server.KestrelTests/PipelineExtensionTests.cs
cesarbs: Ok. I'm still sad but I can accept this. I think.

I like tests to be self-contained. They can call helper methods and such, but generally I avoid having class state like the plague. Ideally every test should be a self-contained mini spec of an expected behavior of the system.

test/Microsoft.AspNetCore.Server.KestrelTests/PipelineExtensionTests.cs
davidfowl: LOST

test/Microsoft.AspNetCore.Server.KestrelTests/PipelineExtensionTests.cs
natemcmaster: Because it is used in an attribute it has to be a const.

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/xunit.runner.json
cesarbs: 😞

test/Microsoft.AspNetCore.Server.KestrelTests/PipelineExtensionTests.cs
natemcmaster: I added a test for everything in the ASCII range, 0x00 - 0x7F. As mentioned here https://github.com/aspnet/KestrelHttpServer/issues/1508#issuecomment-289083405, characters outside the ASCII range should never reach this point...in theory.

test/Microsoft.AspNetCore.Server.KestrelTests/PipelineExtensionTests.cs
```diff
@@ -0,0 +1,166 @@
...
+
+            Assert.Equal(input.Length, reader.Buffer.Length);
+        }
+
+        [Fact]
+        public async Task WriteAsciiNoValidation()
+        {
+            const byte maxAscii = 0x7f;
+            var writer = _pipe.Writer.Alloc();
+            for (var i = 0; i < maxAscii; i++)
```
cesarbs: Use `MemberData` instead of looping. Easier to figure out what's wrong if this ever breaks.

test/Microsoft.AspNetCore.Server.KestrelTests/PipelineExtensionTests.cs
```diff
@@ -0,0 +1,166 @@
...
+
+            writer.FlushAsync().GetAwaiter().GetResult();
+
+            var reader = _pipe.Reader.ReadAsync().GetAwaiter().GetResult();
+            var written = reader.Buffer.Slice(spacer.Length, stringLength);
+            Assert.False(written.IsSingleSpan, "The buffer should cross spans");
+            AssertExtensions.Equal(Encoding.ASCII.GetBytes(testString), written.ToArray());
+        }
+
+        public void Dispose()
```
cesarbs: nit: move this to the top

test/Microsoft.AspNetCore.Server.KestrelTests/TestHelpers/AssertExtensions.cs
```diff
@@ -0,0 +1,27 @@
...
+            if (expected.Length != actual.Length)
+            {
+                throw new XunitException($"Expected length to be {expected.Length} but was {actual.Length}");
+            }
+
+            for (var i = 0; i < expected.Length; i++)
+            {
+                if (expected[i] != actual[i])
+                {
+                    throw new XunitException($@"Expected byte at index {i} to be '\x{expected[i]}' but was '\x{actual[i]}'");
```
cesarbs: Bytes will be printed in decimal. Seems like you wanted hex though.

test/Microsoft.AspNetCore.Server.KestrelTests/PipelineExtensionTests.cs
pakrym: Would it kill vstest performance even more?

natemcmaster: Seems unnecessary to make it a theory. The test only has one assert. So if it goes wrong, the assert error would be 
```
Assert.Equal() Failure
Expected: (x)
Actual: (y)
```

cesarbs: Sounds good.

natemcmaster:
@halter73 ok if I merge, or should I wait for your transport refactor #1551?

natemcmaster:
Rebased. :watch: CI

PR #1555, opened 2017-03-24
Short folder names

natemcmaster:
As long as we are rearranging out entire codebase...

https://github.com/aspnet/KestrelHttpServer/tree/namc/folders

image

image


pakrym:
I think if we should do it only if it's going to be done for entire asp.net core. Also not before @halter73 change gets merged.

natemcmaster:
To be clear, there are no plans to do it for all of aspnet core, but I don't think that should stop us from using a cleaner layout. There is at least [one other repo](https://github.com/aspnet/EntityFramework/tree/dev/src) that has already done this.

Btw, this is on top off @halter73's change.

cesarbs:
Farewell GitHub history 😝

cesarbs:
Not entirely opposed to the change but let's first get the new packages' names right (`*.Transport.*`)

natemcmaster:
So, this?
```
- src/
  + Kestrel
  + Kestrel.Core
  + Kestrel.Transport.Libuv
  + Kestrel.Https
```

In the future we can add Kestrel.Transport.(Whatever).

> Farewell GitHub history 

Yeah :-/. We're gonna lose it no matter what in https://github.com/aspnet/KestrelHttpServer/pull/1551. If only there was [some way](https://git-scm.com/docs/git-log#git-log---follow) to follow git renames. :trollface:

davidfowl:
I'm going to say no to this since it's inconsistent I also prefer the long names. Get the github expander plugin.

davidfowl:
I also adds no value and history ends up being messed up for no good reason (and yes, GitHub should fix their CSS 🤣)

davidfowl:
I also agree with Pavel, if we're not doing this for all of ASP.NET Core it's pointless. Consistency is good across repos and EF for tend to be a bit different with some things done in the repo.

natemcmaster:
@davidfowl has had a change of heart.

pakrym:
Are we doing it for all repos?

Tratcher:
This kind of structure should be consistent across repos. All or none...

I can only assume this breaks some tooling regexes..

natemcmaster:
FYI all - details begin discussed internally, but the primary motivation is continual problems with filenames that grow beyond max path issues. Not sure yet if all repos will get this treatment or not, but regardless, I want this for Kestrel as this project has particularly long filenames, eg `src/Microsoft.AspNetCore.Server.Kestrel.Transport.Abstractions/Microsoft.AspNetCore.Server.Kestrel.Transport.Abstractions.csproj`

natemcmaster:
:watch: waiting until the work on rel/2.0.0 is done. This would make it painful to merge changes in rel/2.0.0 back to dev.

PR #1556, opened 2017-03-25
Add VSCode settings

natemcmaster:
Two new categories of editor setting:

1. ~~Editor config. It's awesome, IMO should be used in all aspnet repos.~~ See https://github.com/aspnet/Universe/issues/500
2. VS Code settings.

~~VS 2017 already supports .editorconfig files. Future updates of Roslyn will actually enforce coding rules in the editorconfig file.~~

Got tired of re-writing my VSCode settings every time I forget to filter the .vscode folder when running git clean -xfd. Can we add here

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -1,0 +1,75 @@
+{
+  "version": "0.2.0",
+  "configurations": [
+    {
+      "name": "Attach: .NET Core",
+      "type": "coreclr",
+      "request": "attach",
+      "processId": "${command:pickProcess}"
+    },
+    {
+      "name": "Debug: SampleApp",
+      "type": "coreclr",
+      "request": "launch",
+      "preLaunchTask": "Compile: solution",
+      "program": "${workspaceRoot}/samples/SampleApp/bin/Debug/netcoreapp1.1/SampleApp.dll",
+      "args": [],
+      "cwd": "${workspaceRoot}/samples/SampleApp",
+      "console": "internalConsole",
+      "stopAtEntry": false,
+      "internalConsoleOptions": "openOnSessionStart",
+      "launchBrowser": {
+        "enabled": true,
+        "args": "${auto-detect-url}",
+        "windows": {
+          "command": "cmd.exe",
+          "args": "/C start ${auto-detect-url}"
+        },
+        "osx": {
+          "command": "open"
+        },
+        "linux": {
+          "command": "xdg-open"
+        }
+      }
+    },
+    {
+      "name": "Debug: LargeResponseApp",
+      "type": "coreclr",
+      "request": "launch",
+      "preLaunchTask": "Compile: solution",
+      "program": "${workspaceRoot}/samples/LargeResponseApp/bin/Debug/netcoreapp1.1/LargeResponseApp.dll",
+      "args": [],
+      "cwd": "${workspaceRoot}/samples/LargeResponseApp",
+      "console": "internalConsole",
+      "stopAtEntry": false,
+      "internalConsoleOptions": "openOnSessionStart",
+      "launchBrowser": {
+        "enabled": true,
+        "args": "${auto-detect-url}",
+        "windows": {
+          "command": "cmd.exe",
+          "args": "/C start ${auto-detect-url}"
+        },
+        "osx": {
+          "command": "open"
+        },
+        "linux": {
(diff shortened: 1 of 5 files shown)


.editorconfig
```diff
@@ -0,0 +1,23 @@
+# EditorConfig is awesome:http://EditorConfig.org
```
davidfowl: Did you talk to @pranavkm and @anurse to make sure these settings are the same as elsewhere?

natemcmaster: I didn't. Hey @pranavkm @anurse, any other settings we should add?

davidfowl: I'd start with this. https://github.com/aspnet/Razor/blob/dev/.editorconfig

natemcmaster: The Razor editorconfig has a bunch of settings that are already the VS default. IIUC the only non-default we use is sorting System.* usings first.

davidfowl: No reason to not copy it. Or at least change it to match the kestrel one. When I make new repos I usually copy from an arbitrary repository with the assumption that they are all the same. If this one is better, then you can change the razor one, but they shouldn't differ.

benaadams: > already the VS default

Someone is going to change their settings from the deafult :)

natemcmaster: > Someone is going to change their settings from the deafult

That's a good point. The logical conclusion to that is that we should specify _all_ editorconfig settings, not just some. See https://github.com/dotnet/roslyn/blob/master/.editorconfig

anurse: This looks like a good start. I tinkered with a much more in-depth file here: https://github.com/aspnet/SignalR/commit/76174580fd79c2f06289aff177203c97ae292ed0 , but was disappointed to learn that VS doesn't enforce many of those options it seems, and neither does the compiler.

.vscode/tasks.json
```diff
@@ -0,0 +1,87 @@
+{
+  "version": "2.0.0",
+  "options": {
+    "env": {
+      "DOTNET_SKIP_FIRST_TIME_EXPERIENCE": "true"
+    }
+  },
+  "tasks": [
+    {
+      "taskName": "Restore",
```
cesarbs: Isn't this already built-in in VS Code?

.vscode/settings.json
```diff
@@ -0,0 +1,11 @@
+{
+  "[json]": {
+    "editor.tabSize": 2
```
anurse: Does VSCode not honor editorconfig here?

.vscode/launch.json
```diff
@@ -0,0 +1,75 @@
...
+        "osx": {
+          "command": "open"
+        },
+        "linux": {
+          "command": "xdg-open"
+        }
+      }
+    },
+    {
+      "name": "Debug CodeGenerator",
```
cesarbs: nit: for consistency's sake let's call this `Debug: code generator`

.vscode/launch.json
```diff
@@ -0,0 +1,75 @@
+{
+  "version": "0.2.0",
+  "configurations": [
+    {
+      "name": ".NET Core Attach",
+      "type": "coreclr",
+      "request": "attach",
+      "processId": "${command:pickProcess}"
```
cesarbs: Does this work cross-plat?

.vscode/settings.json
natemcmaster: You have to install the editorconfig extension first.

.vscode/launch.json
natemcmaster: Yes

.vscode/tasks.json
natemcmaster: It will run dotnet restore individual csproj files, but not on solutions.

.editorconfig
natemcmaster: Enforcing code style is a WIP. See https://github.com/dotnet/roslyn/projects/4

natemcmaster: I don't have enough bandwidth to address everyones concerns about editorconfig, and frankly, don't care. I'm going to back out this file-- leaving only VS code settings. Use https://github.com/aspnet/Universe/issues/500 for more discussion on this.

natemcmaster:
:up:  :date: Applied consistent naming to tasks and debug configs
image

image

natemcmaster:
:bell:

PR #1572, opened 2017-03-30
More decoupling

davidfowl:
- Remove Libuv and Hosting dependency from Kestrel.Core
- Kestrel.Core depends on Hosting.Abstractions

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/ServerAddressesFeature.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/ServerAddressesFeature.cs
@@ -1,0 +1,12 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+using Microsoft.AspNetCore.Hosting.Server.Features;
+
+namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal
+{
+    internal class ServerAddressesFeature : IServerAddressesFeature
+    {
+        public ICollection<string> Addresses { get; } = new List<string>();
+    }
+}
--- a/src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/ServerAddressesFeature.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/ServerAddressesFeature.cs
@@ -1,4 +1,7 @@
-﻿using System;
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
 using System.Collections.Generic;
 using System.Text;
 using Microsoft.AspNetCore.Hosting.Server.Features;
@@ -8,5 +11,6 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal
     internal class ServerAddressesFeature : IServerAddressesFeature
     {
         public ICollection<string> Addresses { get; } = new List<string>();
+        public bool PreferHostingUrls { get; set; }
     }
 }
--- a/src/Microsoft.AspNetCore.Server.Kestrel.Core/Microsoft.AspNetCore.Server.Kestrel.Core.csproj
+++ b/src/Microsoft.AspNetCore.Server.Kestrel.Core/Microsoft.AspNetCore.Server.Kestrel.Core.csproj
@@ -13,9 +13,11 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Libuv" Version="$(LibUvVersion)" />
-    <PackageReference Include="Microsoft.AspNetCore.Hosting" Version="$(AspNetCoreVersion)" />
+    <PackageReference Include="Microsoft.AspNetCore.Hosting.Abstractions" Version="$(AspNetCoreVersion)" />
+    <PackageReference Include="Microsoft.AspNetCore.WebUtilities" Version="$(AspNetCoreVersion)" />
+    <PackageReference Include="Microsoft.Net.Http.Headers" Version="$(AspNetCoreVersion)" />
     <PackageReference Include="Microsoft.Extensions.Logging.Abstractions" Version="$(AspNetCoreVersion)" />
+    <PackageReference Include="Microsoft.Extensions.Options" Version="$(AspNetCoreVersion)" />
     <PackageReference Include="Microsoft.Extensions.TaskCache.Sources" Version="$(AspNetCoreVersion)" PrivateAssets="All" />
     <PackageReference Include="System.Numerics.Vectors" Version="$(CoreFxVersion)" />
     <PackageReference Include="System.Threading.Tasks.Extensions" Version="$(CoreFxVersion)" />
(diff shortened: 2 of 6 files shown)


cesarbs:
Cool, except for `KestrelServer : IServerAddressesFeature`.

src/Microsoft.AspNetCore.Server.Kestrel.Core/Microsoft.AspNetCore.Server.Kestrel.Core.csproj
```diff
@@ -13,9 +13,11 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Libuv" Version="$(LibUvVersion)" />
-    <PackageReference Include="Microsoft.AspNetCore.Hosting" Version="$(AspNetCoreVersion)" />
+    <PackageReference Include="Microsoft.AspNetCore.Hosting.Abstractions" Version="$(AspNetCoreVersion)" />
+    <PackageReference Include="Microsoft.AspNetCore.WebUtilities" Version="$(AspNetCoreVersion)" />
+    <PackageReference Include="Microsoft.Net.Http.Headers" Version="$(AspNetCoreVersion)" />
```
davidfowl: I blame @JunTaoLuo for this

src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/ServerAddressesFeature.cs
```diff
@@ -0,0 +1,12 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+using Microsoft.AspNetCore.Hosting.Server.Features;
+
+namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal
+{
+    internal class ServerAddressesFeature : IServerAddressesFeature
```
Tratcher: heads up @JunTaoLuo

src/Microsoft.AspNetCore.Server.Kestrel.Https/Microsoft.AspNetCore.Server.Kestrel.Https.csproj
```diff
@@ -16,6 +16,7 @@
   </ItemGroup>
 
   <ItemGroup>
+    <PackageReference Include="Microsoft.AspNetCore.Http" Version="$(AspNetCoreVersion)" />
```
davidfowl: This could be Hosting.Abstractions

src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/ServerAddressesFeature.cs
davidfowl: 😄

src/Microsoft.AspNetCore.Server.Kestrel.Https/Microsoft.AspNetCore.Server.Kestrel.Https.csproj
halter73: What type(s) was this reference added for? Was it being brought in transitively through Core's reference to Hosting before?

src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/ServerAddressesFeature.cs
halter73: Could this class just be moved to Hosting.Abstractions?

davidfowl: Sure after this change.

src/Microsoft.AspNetCore.Server.Kestrel.Https/Microsoft.AspNetCore.Server.Kestrel.Https.csproj
davidfowl: https://github.com/aspnet/HttpAbstractions/blob/a0bb843dd581c1521ad8bcbe21923959158cc1f2/src/Microsoft.AspNetCore.Http/Features/TlsConnectionFeature.cs

src/Microsoft.AspNetCore.Server.Kestrel/Microsoft.AspNetCore.Server.Kestrel.csproj
```diff
@@ -12,6 +12,10 @@
   </PropertyGroup>
 
   <ItemGroup>
+    <PackageReference Include="Microsoft.AspNetCore.Hosting" Version="$(AspNetCoreVersion)" />
```
halter73: This is needed, right? The only class in this project, WebHostBuilderKestrelExtension, only seems to reference IWebHostBuilder, IServer and some config stuff. That should come from Hosting.Abstractions, Hosting.Server.Abstractions and Options respectively.

All of this should come in transitively. If that isn't the case, why is the Libuv project building which also has IWebHostBuilder extensions?

Tratcher: Kestrel pulls it in so the app doesn't have to.

davidfowl: This is to not break upstream components that might have been relying on Kestrel depending on Hosting. I don't want to change that as part of this. The goal here was to clean up some of the dependencies after the initial change.

src/Microsoft.AspNetCore.Server.Kestrel.Https/Microsoft.AspNetCore.Server.Kestrel.Https.csproj
halter73: So if you copied TlsConnectionFeature.cs like you did for ServerAddressesFeature.cs, you could get rid of this package reference decouple even more.

src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/ServerAddressesFeature.cs
```diff
@@ -0,0 +1,12 @@
+﻿using System;
```
cesarbs: Copyright header.

src/Microsoft.AspNetCore.Server.Kestrel/Microsoft.AspNetCore.Server.Kestrel.csproj
halter73: What if the app wants to use a non-libuv transport? Maybe we should keep the reference in Core then.

src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/ServerAddressesFeature.cs
Tratcher: I'd rather you clone it than move it to Abstractions

src/Microsoft.AspNetCore.Server.Kestrel.Https/Microsoft.AspNetCore.Server.Kestrel.Https.csproj
davidfowl: Sure, I'm fine with that.

src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/ServerAddressesFeature.cs
JunTaoLuo: @davidfowl you'll need to update this after https://github.com/aspnet/Hosting/pull/992 😄

davidfowl:
Spoke to @halter73 offline. I've added a private `TlsConnectionFeature` in the https connection adapter and added copyrights. This is the first stage to further dependency decoupling.

Issue #1367, opened 2017-02-16
Log TraceIdentifier for ApplicationErrors


halter73:
If we implemented IHttpRequestIdentifierFeature in Frame.FeatureCollection.cs, we would have access to the TraceIdentifier (Request ID) instead of just the ConnectionId. I think Kestrel should log both identifiers when possible. I think something (maybe Hosting) should log which TraceIdentifiers correspond with which ConnectionIds for every request.

Hosting should just pick up our implementation of IHttpRequestIdentifierFeature instead of adding the one from HttpAbstractions to the IFeatureCollection.

Below are the current logs from an unhandled exception fro a request with TraceIdentifier: 0HL2MNKLICAB4 and ConnectionId: 0HL2MNKLGDKRK.

```
~\dev\Universe\KestrelHttpServer\samples\SampleApp [pakrym/pipelines-halter73/sync +0 ~1 -0 !]> dotnet run -c Release -f netcoreapp1.1
Startup.cs(27,17): warning CS0162: Unreachable code detected [C:\Users\shalter\dev\Universe\KestrelHttpServer\samples\SampleApp\SampleApp.csproj]
Hosting environment: Production
Content root path: C:\Users\shalter\dev\Universe\KestrelHttpServer\samples\SampleApp
Now listening on: http://127.0.0.1:5000
Now listening on: http://127.0.0.1:5001
Application started. Press Ctrl+C to shut down.
fail: Microsoft.AspNetCore.Server.Kestrel[13]
      Connection id "0HL2MNKLGDKRK": An unhandled exception was thrown by the application.
System.Exception: test
   at SampleApp.Startup.<>c__DisplayClass0_0.<<Configure>b__0>d.MoveNext() in C:\Users\shalter\dev\Universe\KestrelHttpServer\samples\SampleApp\Startup.cs:line 25
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at Microsoft.AspNetCore.Hosting.Internal.RequestServicesContainerMiddleware.<Invoke>d__3.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
[...]

PR #1578, opened 2017-03-31
Implement IHttpRequestIdentifierFeature on Frame

natemcmaster:
This feature generates a unique ID per request. This unique ID can be used in event source and logging.

Also, this change improves KestrelEventSource by moving it back into the Kestrel.Core assembly and de-coupling from the Libuv transport. This adds two new events, RequestStart and RequestStop, which can be used to identify the correlation between connection ID and request trace
identifier.

Resolves https://github.com/aspnet/KestrelHttpServer/issues/1367

### Benchmark

-n Plaintext --kestrelThreadCount 2
**dev**
RPS: 1344136.86
RPS: 1374674.53

**log-request-id**
RPS: 1341543.62
RPS: 1360439.85

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/Http/Frame.FeatureCollection.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/Http/Frame.FeatureCollection.cs
@@ -19,7 +19,8 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Http
                                  IHttpResponseFeature,
                                  IHttpUpgradeFeature,
                                  IHttpConnectionFeature,
-                                 IHttpRequestLifetimeFeature
+                                 IHttpRequestLifetimeFeature,
+                                 IHttpRequestIdentifierFeature
     {
         // NOTE: When feature interfaces are added to or removed from this Frame class implementation,
         // then the list of `implementedFeatures` in the generated code project MUST also be updated.
@@ -75,188 +76,89 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Http
 
         string IHttpRequestFeature.Protocol
         {
-            get
-            {
-                return HttpVersion;
-            }
-
-            set
-            {
-                HttpVersion = value;
-            }
+            get => HttpVersion;
+            set => HttpVersion = value;
         }
 
         string IHttpRequestFeature.Scheme
         {
-            get
-            {
-                return Scheme ?? "http";
-            }
-
-            set
-            {
-                Scheme = value;
-            }
+            get => Scheme ?? "http";
+            set => Scheme = value;
         }
 
         string IHttpRequestFeature.Method
         {
-            get
-            {
-                return Method;
-            }
-
-            set
-            {
-                Method = value;
-            }
+            get => Method;
+            set => Method = value;
         }
 
         string IHttpRequestFeature.PathBase
         {
-            get
-            {
-                return PathBase ?? "";
-            }
-
-            set
-            {
-                PathBase = value;
-            }
+            get => PathBase ?? "";
+            set => PathBase = value;
         }
 
         string IHttpRequestFeature.Path
         {
-            get
-            {
-                return Path;
-            }
-
-            set
-            {
-                Path = value;
-            }
+            get => Path;
+            set => Path = value;
         }
 
         string IHttpRequestFeature.QueryString
         {
-            get
-            {
-                return QueryString;
-            }
-
-            set
-            {
-                QueryString = value;
-            }
+            get => QueryString;
+            set => QueryString = value;
         }
 
         string IHttpRequestFeature.RawTarget
         {
-            get
-            {
-                return RawTarget;
-            }
-            set
-            {
-                RawTarget = value;
-            }
+            get => RawTarget;
+            set => RawTarget = value;
         }
 
         IHeaderDictionary IHttpRequestFeature.Headers
         {
-            get
-            {
-                return RequestHeaders;
-            }
-
-            set
-            {
-                RequestHeaders = value;
-            }
+            get => RequestHeaders;
+            set => RequestHeaders = value;
         }
 
         Stream IHttpRequestFeature.Body
         {
-            get
-            {
-                return RequestBody;
-            }
-
-            set
-            {
-                RequestBody = value;
-            }
+            get => RequestBody;
+            set => RequestBody = value;
         }
 
         int IHttpResponseFeature.StatusCode
         {
-            get
-            {
-                return StatusCode;
-            }
-
-            set
-            {
-                StatusCode = value;
-            }
+            get => StatusCode;
+            set => StatusCode = value;
         }
 
         string IHttpResponseFeature.ReasonPhrase
         {
-            get
-            {
-                return ReasonPhrase;
-            }
-
-            set
-            {
-                ReasonPhrase = value;
-            }
+            get => ReasonPhrase;
+            set => ReasonPhrase = value;
         }
 
         IHeaderDictionary IHttpResponseFeature.Headers
         {
-            get
-            {
-                return ResponseHeaders;
-            }
-
-            set
-            {
-                ResponseHeaders = value;
-            }
+            get => ResponseHeaders;
+            set => ResponseHeaders = value;
         }
 
         Stream IHttpResponseFeature.Body
         {
-            get
-            {
-                return ResponseBody;
-            }
-
-            set
-            {
-                ResponseBody = value;
-            }
+            get => ResponseBody;
+            set => ResponseBody = value;
         }
 
         CancellationToken IHttpRequestLifetimeFeature.RequestAborted
         {
-            get
-            {
-                return RequestAborted;
-            }
-            set
-            {
-                RequestAborted = value;
-            }
+            get => RequestAborted;
+            set => RequestAborted = value;
         }
 
-        bool IHttpResponseFeature.HasStarted
-        {
-            get { return HasResponseStarted; }
-        }
+        bool IHttpResponseFeature.HasStarted => HasResponseStarted;
 
         bool IHttpUpgradeFeature.IsUpgradableRequest => _upgrade;
 
@@ -266,38 +168,44 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Http
 
         IPAddress IHttpConnectionFeature.RemoteIpAddress
         {
-            get { return RemoteIpAddress; }
-            set { RemoteIpAddress = value; }
+            get => RemoteIpAddress;
+            set => RemoteIpAddress = value;
         }
 
         IPAddress IHttpConnectionFeature.LocalIpAddress
         {
-            get { return LocalIpAddress; }
-            set { LocalIpAddress = value; }
+            get => LocalIpAddress;
+            set => LocalIpAddress = value;
         }
 
         int IHttpConnectionFeature.RemotePort
         {
-            get { return RemotePort; }
-            set { RemotePort = value; }
+            get => RemotePort;
+            set => RemotePort = value;
         }
 
         int IHttpConnectionFeature.LocalPort
         {
-            get { return LocalPort; }
-            set { LocalPort = value; }
+            get => LocalPort;
+            set => LocalPort = value;
         }
 
         string IHttpConnectionFeature.ConnectionId
         {
-            get { return ConnectionIdFeature; }
-            set { ConnectionIdFeature = value; }
+            get => ConnectionIdFeature;
+            set => ConnectionIdFeature = value;
+        }
+
+        string IHttpRequestIdentifierFeature.TraceIdentifier
+        {
+            get => TraceIdentifier;
+            set => TraceIdentifier = value;
         }
 
         object IFeatureCollection.this[Type key]
         {
-            get { return FastFeatureGet(key); }
-            set { FastFeatureSet(key, value); }
+            get => FastFeatureGet(key);
+            set => FastFeatureSet(key, value);
         }
 
         TFeature IFeatureCollection.Get<TFeature>()
--- a/src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/Http/FrameOfT.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/Http/FrameOfT.cs
@@ -6,6 +6,7 @@ using System.IO;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Hosting.Server;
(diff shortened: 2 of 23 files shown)


src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/Infrastructure/CorrelationIdGenerator.cs
```diff
@@ -0,0 +1,48 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Threading;
+
+namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Infrastructure
+{
+    internal static class CorrelationIdGenerator
```
natemcmaster: Depending on the outcome of https://github.com/aspnet/Common/pull/197, this may move to Microsoft.Extensions.Primitives

natemcmaster:
Sample log when application request has an error:
```
fail: Microsoft.AspNetCore.Server.Kestrel[13]
      Connection id "0HL3OB48ARG6C", Request id "0HL3OB48ARG6H": An unhandled exception was thrown by the application.
```

src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/Infrastructure/CorrelationIdGenerator.cs
Tratcher: Why not leave it in HttpAbstractions? Kestrel and Hosting would have access to it there. Or does something else need it?

natemcmaster: I'll refer you to https://github.com/aspnet/Common/pull/197 on that one. @muratg is looking into it.

src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/Http/Frame.FeatureCollection.cs
```diff
@@ -75,187 +76,91 @@ private void ExtraFeatureSet(Type key, object value)
...
-                RequestAborted = value;
-            }
+            get => RequestAborted;
+            set => RequestAborted = value;
         }
 
         bool IHttpResponseFeature.HasStarted
         {
-            get { return HasResponseStarted; }
+            get => HasResponseStarted;
```
Tratcher: These can be reduced to `bool IHttpResponseFeature.HasStarted => HasResponseStarted;` right?

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/EventSourceTests.cs
```diff
@@ -47,16 +49,33 @@ public EventSourceTests()
...
-            var start = Assert.Single(events, e => e.EventName == "ConnectionStart");
-            Assert.All(new[] { "connectionId", "scheme", "remoteEndPoint", "localEndPoint" }, p => Assert.Contains(p, start.PayloadNames));
-            Assert.Equal("http", GetProperty(start, "scheme"));
-            Assert.Equal($"127.0.0.1:{port}", GetProperty(start, "localEndPoint"));
+            var events = _listener.EventData.Where(e => e != null && GetProperty(e, "connectionId") == connectionId).ToList();
 
-            var stop = Assert.Single(events, e => e.EventName == "ConnectionStop");
-            Assert.All(new[] { "connectionId" }, p => Assert.Contains(p, stop.PayloadNames));
-            Assert.Same(KestrelEventSource.Log, stop.EventSource);
+            {
```
cesarbs: Why the scopes?

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/RequestTests.cs
```diff
@@ -566,6 +567,83 @@ public void CanUpgradeRequestWithConnectionKeepAliveUpgradeHeader()
...
+        {
+            const int IdentifierLength = 13;
+            const int iterations = 100;
+
+            using (var server = new TestServer(context =>
+            {
+                var id = Encoding.ASCII.GetBytes(context.TraceIdentifier);
+                Assert.Equal(IdentifierLength, id.Length);
+                context.Response.ContentLength = id.Length;
+                context.Response.Body.Write(id, 0, id.Length);
```
cesarbs: `WriteAsync`

src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/Http/Frame.FeatureCollection.cs
natemcmaster: Yes. Will update.

src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/Http/FrameOfT.cs
```diff
@@ -95,6 +96,8 @@ public Frame(IHttpApplication<TContext> application, FrameContext frameContext)
                         {
                             try
                             {
+                                KestrelEventSource.Log.RequestStart(this);
```
davidfowl: I wonder if we need these events. Hosting does the exact same thing what value do we add here?

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/RequestTests.cs
```diff
@@ -566,6 +567,83 @@ public void CanUpgradeRequestWithConnectionKeepAliveUpgradeHeader()
...
+                return Task.CompletedTask;
+            }))
+            {
+                var usedIds = new HashSet<string>();
+                var usedIdsLock = new object();
```
cesarbs: Why not `ConcurrentBag` instead of set + lock?

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/RequestTests.cs
```diff
@@ -566,6 +567,83 @@ public void CanUpgradeRequestWithConnectionKeepAliveUpgradeHeader()
...
+                    var id = await HttpClientSlim.GetStringAsync(uri);
+                    lock (usedIdsLock)
+                    {
+                        Assert.DoesNotContain(id, usedIds);
+                        usedIds.Add(id);
+                    }
+                });
+
+                // requests on same connection
+                using (var conn = server.CreateConnection())
```
cesarbs: nit: call it `connection` since that's we call it in all other tests

src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/Http/FrameOfT.cs
natemcmaster: Hosting only [logs method and path](https://github.com/aspnet/Hosting/blob/dev/src/Microsoft.AspNetCore.Hosting/Internal/HostingEventSource.cs#L36-L39). This logs connectionId and requestId to make it possible to correlate connections and requests.

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/RequestTests.cs
```diff
@@ -566,6 +567,83 @@ public void CanUpgradeRequestWithConnectionKeepAliveUpgradeHeader()
...
+                {
+                    var buff = new char[IdentifierLength];
```
cesarbs: Buff is a funny name, call it either `buf` or `buffer` 😄

test/Microsoft.AspNetCore.Server.KestrelTests/FrameTests.cs
```diff
@@ -124,6 +124,33 @@ public void ResetResetsScheme()
...
+        [Fact]
+        public void TraceIdentifierGeneratesWhenNull()
+        {
+            _frame.TraceIdentifier = null;
+            var id = _frame.TraceIdentifier;
+            Assert.NotNull(id);
+            Assert.Equal(id, _frame.TraceIdentifier);
+
+            _frame.TraceIdentifier = null;
+            Assert.NotEqual(id, _frame.TraceIdentifier);
```
cesarbs: `Assert.NotNull` again?

src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/Infrastructure/IKestrelTrace.cs
```diff
@@ -45,6 +45,6 @@ public interface IKestrelTrace : ILogger
 
         void NotAllConnectionsAborted();
 
-        void ApplicationError(string connectionId, Exception ex);
+        void ApplicationError(string connectionId, string traceIdentifier, Exception ex);
```
halter73: Nice!

halter73:
Would it make sense to give connection and request id's unique prefixes to make them easier to distinguish in logs?

src/Microsoft.AspNetCore.Server.Kestrel.Transport.Libuv/Internal/Infrastructure/EventSourceExtensions.cs
```diff
@@ -0,0 +1,31 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.AspNetCore.Server.Kestrel.Internal.Http;
+
+namespace Microsoft.AspNetCore.Server.Kestrel.Internal.Infrastructure
+{
+    public static class EventSourceExtensions
+    {
+        public static void ConnectionStart(this KestrelEventSource eventSource, Connection connection)
```
halter73: Since you are moving KestrelEventSource back to the Core project, can this change also move the connection start/stop logging to core? It would be nice if this tracing could be shared between transports.

natemcmaster:
> Would it make sense to give connection and request id's unique prefixes to make them easier to distinguish in logs?

@shirhatti thoughts?

Another idea I played with making each request ID contain the connection id so correlation is even easier.
```c#
// pseudocode
var count = _connectionRequests++;
var requestId = $"{_connectionId}:{count}";
```

src/Microsoft.AspNetCore.Server.Kestrel.Transport.Libuv/Internal/Infrastructure/EventSourceExtensions.cs
natemcmaster: This change was required to decouple `Connection` which is in Transport.Libuv and `KestrelEventSource` which is now in Core. ConnectionStop/Start are still on KestrelEventSource which is in Kestrel.Core, but the public methods now take string arguments directly instead of `Connection`.

test/Microsoft.AspNetCore.Server.KestrelTests/FrameTests.cs
natemcmaster: It's implicit as we already asserted id != null.

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/RequestTests.cs
natemcmaster: ᕙ༼ຈل͜ຈ༽ᕗ

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/EventSourceTests.cs
natemcmaster: avoiding copy+paste errors. Because the asserts for each event type looks so similar.

src/Microsoft.AspNetCore.Server.Kestrel.Transport.Libuv/Internal/Infrastructure/EventSourceExtensions.cs
halter73: I still think this is worth doing now. Rather than partially moving KestrelEventSource to Core, let's move it completely. I think this might require adding some sort of OnConnectionClosed() method to IConnectionContext, but it should still be a pretty quick/simple change.

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/RequestTests.cs
natemcmaster: Yeah, i'll use that. will update

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/RequestTests.cs
natemcmaster: Returns chunked encoding, which made it more difficult to pull the response body via StreamReader.

src/Microsoft.AspNetCore.Server.Kestrel.Transport.Libuv/Internal/Infrastructure/EventSourceExtensions.cs
natemcmaster: Updated.

natemcmaster:
:up: :date:

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/RequestTests.cs
```diff
@@ -566,6 +567,76 @@ public void CanUpgradeRequestWithConnectionKeepAliveUpgradeHeader()
...
+
+                // requests on same connection
+                using (var connection = server.CreateConnection())
+                {
+                    var buffer = new char[IdentifierLength];
+                    for (var i = 0; i < iterations; i++)
+                    {
+                        await connection.Send("GET / HTTP/1.1",
+                            "Content-Length: 0",
+                            "Host: localhost",
```
natemcmaster: Updated.

shirhatti:
> Another idea I played with making each request ID contain the connection id so correlation is even easier.

YES please!

@glennc Thoughts?

natemcmaster:
@shirhatti I will follow up via email. Not everyone checks PR comments.

src/Microsoft.AspNetCore.Server.Kestrel.Core/Transport/IConnectionContext.cs
```diff
@@ -12,6 +12,7 @@ public interface IConnectionContext
         string ConnectionId { get; }
         IPipeWriter Input { get; }
         IPipeReader Output { get; }
+        void OnConnectionClosed();
```
halter73: Nit: Move below "// TODO: Remove these (Use Pipes instead?)"

There is a task that will be added to Pipes that should let us track when SocketOutputConsumer complete's it's pipe reader which will be right after disposing the socket.

Issue #1582, opened 2017-03-31
Fix namespaces/folders/class names


halter73:
- [x] Create Microsoft.AspNetCore.Server.Kestrel.Transport.Abstractions project
- [x] Fix namespaces/folders/class names (rename Microsoft.AspNetCore.Server.Kestrel.Libuv to Microsoft.AspNetCore.Server.Kestrel.Transport.Libuv)

PR #1594, opened 2017-04-02
Add Microsoft.AspNetCore.Server.Kestrel.Transport.Abstractions project (#1582)

cesarbs:
Partially addresses #1582. I'll rename/fix namespaces in `Transport.Libuv` in a separate PR.

The way connection adapters are flowing is a bit of a wart. Already discussed with @davidfowl and for now we've decided to cast `IEndPointInformation` to `ListenOptions` in Core. Alternatives would be:

- Move connection adapter interfaces to the new project and keep adapter-related properties in `IEndPointInformation` (in which case I'd just call it `IListenOptions`). But IMO this smells, transports shouldn't "see" adapter stuff.
- Have a new `Adapter.Abstractions` package to hold connection adapter interfaces, then `Transport.Abstractions` could reference that.
- Create an endpoint -> connection adapter(s) mapping in Core, which is queried by `ConnectionHandler` when the transport calls it with a new connection.


src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/ConnectionHandler.cs
```diff
@@ -44,7 +44,7 @@ public IConnectionContext OnConnection(IConnectionInformation connectionInfo)
                 ConnectionId = connectionId,
                 ServiceContext = _serviceContext,
                 PipeFactory = connectionInfo.PipeFactory,
-                ConnectionAdapters = connectionInfo.ListenOptions.ConnectionAdapters,
+                ConnectionAdapters = ((ListenOptions)connectionInfo.EndPointInformation).ConnectionAdapters,
```
davidfowl: `// TODO: This is a hack, we need to associate kestrel specific things with endpoint information somehow`

src/Microsoft.AspNetCore.Server.Kestrel.Transport.Abstractions/Microsoft.AspNetCore.Server.Kestrel.Transport.Abstractions.csproj
```diff
@@ -0,0 +1,19 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <Import Project="..\..\build\common.props" />
+
+  <PropertyGroup>
+    <Description>Transport abstractions for the ASP.NET Core Kestrel cross-platform web server.</Description>
+    <TargetFrameworks>netstandard1.3;net46</TargetFrameworks>
```
davidfowl: Remove net46

cesarbs:
Updated with the change to have one `ConnectionHandler` per endpoint.

PR #1595, opened 2017-04-02
Remove Kestrel.Core dependency from Transport.Libuv (#1584)

cesarbs:
#1584

Adds `ILibuvTrace` and `LibuvTrace` so the libuv transport can have its own trace and not depend on `IKestrelTrace`.

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Kestrel.Transport.Libuv/Microsoft.AspNetCore.Server.Kestrel.Transport.Libuv.csproj
+++ b/src/Microsoft.AspNetCore.Server.Kestrel.Transport.Libuv/Microsoft.AspNetCore.Server.Kestrel.Transport.Libuv.csproj
@@ -4,7 +4,7 @@
 
   <PropertyGroup>
     <Description>Libuv transport for the ASP.NET Core Kestrel cross-platform web server.</Description>
-    <TargetFrameworks>netstandard1.3;net46</TargetFrameworks>
+    <TargetFramework>netstandard1.3</TargetFramework>
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
     <PackageTags>aspnetcore;kestrel</PackageTags>
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
@@ -13,6 +13,8 @@
   </PropertyGroup>
 
   <ItemGroup>
+    <PackageReference Include="System.Threading.Thread" Version="$(CoreFxVersion)" />
+    <PackageReference Include="System.Threading.ThreadPool" Version="$(CoreFxVersion)" />
     <PackageReference Include="Libuv" Version="$(LibUvVersion)" />
     <PackageReference Include="Microsoft.AspNetCore.Hosting.Abstractions" Version="$(AspNetCoreVersion)" />
     <PackageReference Include="Microsoft.Extensions.Logging.Abstractions" Version="$(AspNetCoreVersion)" />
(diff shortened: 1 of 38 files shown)


halter73:
Other than removing the comments from IKestrelTrace.cs and removing the unused trace methods from KestrelTrace.cs, what else needs to be done?

cesarbs:
Updated as discussed offline.

For now I've kept the same method names and event IDs.

test/Microsoft.AspNetCore.Server.KestrelTests/MultipleLoopTests.cs
```diff
@@ -7,24 +7,19 @@
...
     {
-        private readonly LibuvFunctions _uv;
-        private readonly IKestrelTrace _logger;
-        public MultipleLoopTests()
-        {
-            _uv = new LibuvFunctions();
-            _logger = new TestKestrelTrace();
-        }
+        private readonly LibuvFunctions _uv = new LibuvFunctions();
+        private readonly ILibuvTrace _logger = Mock.Of<ILibuvTrace>();
```
halter73: The mock `ILibuvTrace` instances should log to the TestApplicationErrorLogger. Then we can do the `#if true` to `#if false` trick, look at the Messages property, etc...

src/Microsoft.AspNetCore.Server.Kestrel.Transport.Libuv/Microsoft.AspNetCore.Server.Kestrel.Transport.Libuv.csproj
```diff
@@ -14,11 +14,19 @@
...
+    <PackageReference Include="Microsoft.Extensions.Options" Version="$(AspNetCoreVersion)" />
     <PackageReference Include="Microsoft.Extensions.TaskCache.Sources" Version="$(AspNetCoreVersion)" PrivateAssets="All" />
   </ItemGroup>
 
   <ItemGroup>
-    <ProjectReference Include="..\Microsoft.AspNetCore.Server.Kestrel.Core\Microsoft.AspNetCore.Server.Kestrel.Core.csproj" />
+    <ProjectReference Include="..\Microsoft.AspNetCore.Server.Kestrel.Transport.Abstractions\Microsoft.AspNetCore.Server.Kestrel.Transport.Abstractions.csproj" />
+  </ItemGroup>
+
+  <ItemGroup Condition=" '$(TargetFramework)' == 'netstandard1.3' ">
```
davidfowl: remove net46

cesarbs:
Updated.

test/Microsoft.AspNetCore.Server.KestrelTests/UvTimerHandleTests.cs
```diff
@@ -2,22 +2,23 @@
...
 using Microsoft.AspNetCore.Server.Kestrel.Internal.Networking;
+using Microsoft.AspNetCore.Server.Kestrel.Transport.Libuv.Internal.Infrastructure;
 using Microsoft.AspNetCore.Testing;
 using Xunit;
 
 namespace Microsoft.AspNetCore.Server.KestrelTests
 {
     public class UvTimerHandleTests
     {
+        private readonly ILibuvTrace _trace = new LibuvTrace(new TestApplicationErrorLogger());
```
davidfowl: Any specific reason this is class level? Are you embracing xunit?

cesarbs: Being consistent with other test classes. I don't like it 😝 But we don't really check any logs in these tests so whatever.

PR #1596, opened 2017-04-03
[WIP][ALT] Remove Kestrel.Core dependency from Transport.Libuv (#1584)

cesarbs:
Alternative to #1595.

Implements `ITransportTrace` in `KestrelTrace` and passes it to the transport via `ITransportFactory.Create()`.

Requires creating a short-lived logged in `LibuvTransportFactory` to log warnings about `ECONNRESET` and   `EADDRINUSE`.

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Kestrel.Transport.Abstractions/ITransportTrace.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel.Transport.Abstractions/ITransportTrace.cs
@@ -1,0 +1,29 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.AspNetCore.Server.Kestrel.Transport.Abstractions
+{
+    public interface ITransportTrace : ILogger
+    {
+        void ConnectionRead(string connectionId, int count);
+
+        void ConnectionReadFin(string connectionId);
+
+        void ConnectionWriteFin(string connectionId);
+
+        void ConnectionWroteFin(string connectionId, int status);
+
+        void ConnectionWriteCallback(string connectionId, int status);
+
+        void ConnectionError(string connectionId, Exception ex);
+
+        void ConnectionReset(string connectionId);
+
+        void NotAllConnectionsClosedGracefully();
+
+        void NotAllConnectionsAborted();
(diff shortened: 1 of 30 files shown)


src/Microsoft.AspNetCore.Server.Kestrel.Transport.Abstractions/ITransportTrace.cs
```diff
@@ -0,0 +1,29 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.AspNetCore.Server.Kestrel.Transport.Abstractions
+{
+    public interface ITransportTrace : ILogger
```
davidfowl: This should probably be an abstract base class so that we can add logs

src/Microsoft.AspNetCore.Server.Kestrel.Transport.Libuv/Microsoft.AspNetCore.Server.Kestrel.Transport.Libuv.csproj
```diff
@@ -14,11 +14,18 @@
...
+    <PackageReference Include="Microsoft.Extensions.Options" Version="$(AspNetCoreVersion)" />
     <PackageReference Include="Microsoft.Extensions.TaskCache.Sources" Version="$(AspNetCoreVersion)" PrivateAssets="All" />
   </ItemGroup>
 
   <ItemGroup>
-    <ProjectReference Include="..\Microsoft.AspNetCore.Server.Kestrel.Core\Microsoft.AspNetCore.Server.Kestrel.Core.csproj" />
+    <ProjectReference Include="..\Microsoft.AspNetCore.Server.Kestrel.Transport.Abstractions\Microsoft.AspNetCore.Server.Kestrel.Transport.Abstractions.csproj" />
+  </ItemGroup>
+
+  <ItemGroup Condition=" '$(TargetFramework)' == 'netstandard1.3' ">
```
davidfowl: You could probably remove net46 from here too.

src/Microsoft.AspNetCore.Server.Kestrel.Transport.Abstractions/ITransportTrace.cs
pakrym: If this is transport specific, do we need an interface?

PR #1602, opened 2017-04-04
React to corefxlab changes

pakrym:
https://github.com/dotnet/corefxlab/commit/f436de4b13afe13d52a75f146f026ecbf9b4fe15
@davidfowl 

Also unblocks coreclr gate


davidfowl:
Shouldn't this be to dev-gate?

pakrym:
Works in `dev` so doesn't matter.

PR #1605, opened 2017-04-04
Re-write UvStreamHandleTests.ReadStopIsIdempotent without Moq

cesarbs:
@mikeharder You were the original author of this test. Do you remember why you had to use Moq? The test works the same way even if I don't use it.


cesarbs:
Looking at the AppVeyor failure. It's unrelated to this change.

mikeharder:
I don't remember anything beyond the code comment, if the new code works it seems fine.

PR #1610, opened 2017-04-05
[WIP]: Read next request if the request body is empty

davidfowl:
Trying some crazy stuff.


benaadams:
Even need to read? Use what is still in the buffer (If there is something)?

davidfowl:
True, can just slice the buffer

PR #1613, opened 2017-04-06
Use 4.3.0 version of System.Threading.Tasks.Extensions

pranavkm:


pranavkm:
Got a assert fail when I built locally. Trying it here to figure out if it's a local build thing.

cc @mikeharder  \ @davidfowl

Issue #1588, opened 2017-03-31
Split transport-specific tests and general tests into distinct test projects


halter73:

cesarbs:
As part of this, `LibuvTransportTests` (former `EngineTests`) should be revisited and all the tests unrelated to the transport should be moved out of that class.

PR #1614, opened 2017-04-06
Split transport-specific tests and general tests into distinct test projects (#1588)

cesarbs:
Here's today's Huge PR Of The Day™, addressing #1588.

I suggest reading through the commit history instead of looking at everything at once.


cesarbs:
@mikeharder I'd like your input on the `MaxRequestBufferSizeTests` changes.

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/MaxRequestBufferSizeTests.cs
```diff
@@ -245,11 +245,6 @@ public static IEnumerable<object[]> LargeUploadData
         private static Socket CreateSocket(int port)
         {
             var socket = new Socket(AddressFamily.InterNetwork, SocketType.Stream, ProtocolType.Tcp);
-
-            // Timeouts large enough to prevent false positives, but small enough to fail quickly.
-            socket.SendTimeout = 10 * 1000;
```
mikeharder: Why was this changed?  I think there might be small advantages to set the timeouts on the socket, otherwise I think the manage task might timeout but the native socket will stay around longer.

cesarbs: We thought this was causing problems but it wasn't really the case - we're seeing the same failures on dev. It's most likely slowness caused by having the pipelines code embedded into Kestrel and compiled for debug. This change makes those failures more pronounced because it moves a lot of tests into the functional tests project, so there's more stuff running at the same time and that may be slowing things down even more and causing timeouts.

@davidfowl and I discussed yesterday about re-writing this test in a different way. We already have several tests that watch the logs for certain events, and we could do the same thing here and wait until we see "connection paused" and "connection resumed" logs, instead of being dependent on time as this test is right now.

mikeharder: OK, so i should wait for the new design to review again?

cesarbs: Will probably be a separate change since it's not specific to the changes introduced here, but yeah I'll want your feedback on it.

cesarbs:
@davidfowl told me to :shipit: in person 😛

PR #1615, opened 2017-04-06
[WIP]: Drain the read buffer before going back to the pipe

davidfowl:
- Avoid calling Input.ReadAsync if the existing buffer hasn't
been drained as yet.

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/Http/FrameOfT.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/Http/FrameOfT.cs
@@ -110,10 +110,13 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
                             }
                         }
 
-                        // Incomplete message, we need a new buffer from the pipe
-                        Input.Advance(consumed, examined);
+                        if (!needBuffer)
+                        {
+                            // Incomplete message, we need a new buffer from the pipe
+                            Input.Advance(consumed, examined);
 
-                        needBuffer = true;
+                            needBuffer = true;
+                        }
                     }
 
                     if (!_requestProcessingStopping)
--- a/src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/Http/FrameOfT.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/Http/FrameOfT.cs
@@ -60,9 +60,9 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
                         {
                             ParseRequest(ref buffer, out consumed, out examined);
 
-                            if (buffer.Length == 0)
+                            if (buffer.End == examined)
                             {
-                                // We've consumed the entire buffer so we need to advance the pipe
+                                // We've observed or consumed the entire buffer so we need to advance the pipe
                                 needAdvance = true;
                             }
                         }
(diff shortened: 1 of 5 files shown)


src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/Http/FrameOfT.cs
```diff
@@ -40,26 +47,47 @@ public Frame(IHttpApplication<TContext> application, FrameContext frameContext)
...
+                        }
+
+                        var needAdvance = false;
 
                         try
                         {
-                            ParseRequest(result.Buffer, out consumed, out examined);
+                            ParseRequest(ref buffer, out consumed, out examined);
+
+                            if (buffer.Length == 0)
```
cesarbs: Back when we still had `SocketInput`, we had a bug where we weren't advancing to the furthest examined position. This caused the server to spin a lot while waiting for the full request, because the `await` always completed immediately.

Wouldn't the same happen here?

davidfowl: Why don't we have a test to cover that?

benaadams: Would be something like `buffer.Length == 0 || examined.IsEnd` cover it?

benaadams: Or maybe just `examined.IsEnd`?

davidfowl: Hmmmm, after thinking about this for a bit. It could happen. We could check if `examined == buffer.End` and set needBuffer to true so it doesn't spin. Will add a unit test for this.

davidfowl: It's basically checking if progress was made after a parse.

cesarbs: My bad. We had it until I converted a bunch of tests to parser tests: https://github.com/aspnet/KestrelHttpServer/commit/5743d740b454874a3d056d2f480297be51abcb71#diff-78adb33626b9b8aa43b7d4529f4f0bec

I'm bringing it back.

cesarbs: Ah, I remember why I removed it.

`Frame` does not move the cursors anymore, it's the parser that does that. In the parser tests we have checks for incomplete request lines and headers and we assert the examined cursor there [[1](https://github.com/aspnet/KestrelHttpServer/blob/dd02ba577fdb17ff540863df87286062a33579b7/test/Microsoft.AspNetCore.Server.KestrelTests/HttpParserTests.cs#L62)] [[2](https://github.com/aspnet/KestrelHttpServer/blob/dd02ba577fdb17ff540863df87286062a33579b7/test/Microsoft.AspNetCore.Server.KestrelTests/HttpParserTests.cs#L200)]. So that's covered.

There's no issue with the parser either. It will always:

1. Either use the entire buffer and set `examined` to `buffer.End`; or
1. Slice the buffer and
   1. Either successfully parse the request line and set `examined` to the end of the slice, which is fine because it means that whatever awaits next will just complete immediately and get the remaining data in the buffer; or
   1. Reject the request because the entire request line or headers didn't fit in the slice, meaning they exceed their limits. At this point we don't care about advancing the reader properly anymore, since we're closing the connection.

benaadams:
No good?

davidfowl:
End to end didn't improve

PR #1622, opened 2017-04-07
Don't reassign _adaptedPipelineTask in FrameConnection

cesarbs:
Prevents a race where the server is disposed while an adapted connection is still being setup. If `StopAsync` is called before the task is re-assigned, everything is disposed and then `ReadInputAsync` throws a `Debug.Assert` because it tries to alloc from a disposed memory pool.

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/FrameConnection.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/FrameConnection.cs
@@ -25,7 +25,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal
 
         private AdaptedPipeline _adaptedPipeline;
         private Stream _filteredStream;
-        private Task _adaptedPipelineTask = TaskCache.CompletedTask;
+        private TaskCompletionSource<object> _adaptedPipelineTcs = new TaskCompletionSource<object>();
 
         public FrameConnection(FrameConnectionContext context)
         {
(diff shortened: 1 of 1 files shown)


src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/FrameConnection.cs
```diff
@@ -123,7 +124,7 @@ public void Timeout()
                     _frame.Input = _adaptedPipeline.Input.Reader;
                     _frame.Output = _adaptedPipeline.Output;
 
-                    _adaptedPipelineTask = RunAdaptedPipeline();
+                    var ignore = RunAdaptedPipeline();
```
davidfowl: Is it possible for this code path to throw and `_adaptedPipelineTcs` never get set?

src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/FrameConnection.cs
```diff
@@ -25,7 +24,7 @@ public class FrameConnection : IConnectionContext
 
         private AdaptedPipeline _adaptedPipeline;
         private Stream _filteredStream;
-        private Task _adaptedPipelineTask = TaskCache.CompletedTask;
+        private TaskCompletionSource<object> _adaptedPipelineTcs = new TaskCompletionSource<object>();
```
halter73: This is unnecessary since we have _frameStartedTcs

Issue #1616, opened 2017-04-06
Flaky Test: Kestrel.FunctionalTests sometimes fails with "unable to communicate with test host process" or assert


mikeharder:
Kestrel.FunctionalTests sometimes fails with "unable to communicate with test host process" (on netcoreapp2.0) or an assert (net46).  A workaround is to use release instead of debug (`build.cmd /p:Configuration=Release`).

# Assert
```
Block being leased from disposed pool!
   at Microsoft.AspNetCore.Server.Kestrel.Internal.System.IO.Pipelines.MemoryPool.Lease()
   at Microsoft.AspNetCore.Server.Kestrel.Internal.System.IO.Pipelines.MemoryPool.Rent(Int32 size)
   at Microsoft.AspNetCore.Server.Kestrel.Internal.System.IO.Pipelines.Pipe.AllocateWriteHeadUnsynchronized(Int32 count)
   at Microsoft.AspNetCore.Server.Kestrel.Internal.System.IO.Pipelines.Pipe.Microsoft.AspNetCore.Server.Kestrel.Internal.System.IO.Pipelines.IPipeWriter.Alloc(Int32 minimumSize)
   at Microsoft.AspNetCore.Server.Kestrel.Core.Adapter.Internal.AdaptedPipeline.<ReadInputAsync>d__10.MoveNext()
   at System.Runtime.CompilerServices.AsyncTaskMethodBuilder.Start[TStateMachine](TStateMachine& stateMachine)
   at Microsoft.AspNetCore.Server.Kestrel.Core.Adapter.Internal.AdaptedPipeline.ReadInputAsync()
   at Microsoft.AspNetCore.Server.Kestrel.Core.Adapter.Internal.AdaptedPipeline.<RunAsync>d__9.MoveNext()
   at System.Runtime.CompilerServices.AsyncTaskMethodBuilder.Start[TStateMachine](TStateMachine& stateMachine)
   at Microsoft.AspNetCore.Server.Kestrel.Core.Adapter.Internal.AdaptedPipeline.RunAsync()
   at Microsoft.AspNetCore.Server.Kestrel.Core.Internal.FrameConnection.<RunAdaptedPipeline>d__26.MoveNext()
   at System.Runtime.CompilerServices.AsyncTaskMethodBuilder.Start[TStateMachine](TStateMachine& stateMachine)
   at Microsoft.AspNetCore.Server.Kestrel.Core.Internal.FrameConnection.RunAdaptedPipeline()
[...]

PR #1623, opened 2017-04-07
Don't return data from RawStream after pipe is completed

cesarbs:
Makes adapters fail faster after the connection closes, preventing ungraceful shutdowns that lead to #1616.

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Kestrel.Core/Adapter/Internal/RawStream.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel.Core/Adapter/Internal/RawStream.cs
@@ -115,16 +115,17 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Adapter.Internal
                 var readableBuffer = result.Buffer;
                 try
                 {
+                    if (result.IsCompleted)
+                    {
+                        return 0;
+                    }
+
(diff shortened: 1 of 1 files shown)


cesarbs:
- [ ]TODO: add tests

src/Microsoft.AspNetCore.Server.Kestrel.Core/Adapter/Internal/RawStream.cs
```diff
@@ -115,17 +115,18 @@ public override Task FlushAsync(CancellationToken cancellationToken)
                 var readableBuffer = result.Buffer;
                 try
                 {
+                    if (result.IsCompleted)
+                    {
+                        return 0;
```
halter73: If we want to make this similar to the request and response streams both read and write should throw.

halter73:
@cesarbs I don't expect you to change this PR on vacation. I'm just starting a discussion on what behavior we want. I don't want someone else just merging this unless we're sure about this.

cesarbs:
@halter73 Actually I'd first like to understand what exactly is happening there. I guess we're good wrt tests because increasing the socket timeout seems to mitigate it. So this could wait a while.

cesarbs:
This has gone stale. Closing.

PR #1626, opened 2017-04-07
Dispatch only if not already on TP

benaadams:
Don't double dispatch. 

Input pipe can dispatch (flush); then output pipe can also dispatch for write (flush)

Detect if already on TP (as either behaviour could change) and don't dispatch a second time if so.

/cc @davidfowl


benaadams:
nvm :)

PR #1637, opened 2017-04-08
Use ref locals for the HttpParser instead of unsafe native pointers

davidfowl:


src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/Http/HttpParser.cs
```diff
@@ -340,69 +333,81 @@ private unsafe int FindEndOfName(byte* headerLine, int length)
...
 
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
-        private static unsafe bool Contains(byte* searchSpace, int length, byte value)
+        private static unsafe Span<byte> MakeSpan(ref byte data, int length)
+        {
+            // We can use this fast path on .NET Core 2.0 to make a span from ref byte
+            // return Span<byte>.DangerousCreate(null, ref data, length);
+            // For slow span, there's no way to do this because we don't have the backing object so we
+            // need to fix the data
+            fixed (byte* pointer = &data)
```
davidfowl: I believe this causes a GC hole in the general case. Kestrel is fine because we use pinned memory but in general,  the GC can only track interior pointers for fast span, slow span requires the backing object otherwise it might get collected and this span would be pointing to no where...

PR #1638, opened 2017-04-09
Use Span.IndexOf rather than Contains

benaadams:
It will only search for the byte if its found; so shouldn't be slower in fast path

/cc @davidfowl

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/Http/HttpParser.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/Http/HttpParser.cs
@@ -404,7 +404,9 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
                 valueBuffer = new Span<byte>(headerLine + valueStart, valueEnd - valueStart + 1);
             }
 
-            handler.OnHeader(new Span<byte>(headerLine, nameEnd), valueBuffer);
+            var nameBuffer = new Span<byte>(headerLine, nameEnd);
+
+            handler.OnHeader(nameBuffer, valueBuffer);
         }
 
         [MethodImpl(MethodImplOptions.NoInlining)]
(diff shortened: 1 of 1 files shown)


src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/Http/HttpParser.cs
```diff
@@ -393,50 +399,16 @@ private unsafe void TakeSingleHeader(byte* headerLine, int length, IHttpHeadersH
                 {
                     break;
                 }
+
+                valueLength--;
```
davidfowl: Put this in the for loop.

src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/Http/HttpParser.cs
```diff
@@ -69,6 +69,8 @@ public unsafe bool ParseRequestLine(IHttpRequestLineHandler handler, ReadableBuf
 
         private unsafe void ParseRequestLine(IHttpRequestLineHandler handler, byte* data, int length)
         {
+            Debug.Assert(length >= 0);
```
davidfowl: Remove the asserts

src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/Http/HttpParser.cs
benaadams: Changed to bool; and resuse span if not changed

src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/Http/HttpParser.cs
benaadams: Removed

src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/Http/HttpParser.cs
```diff
@@ -379,62 +378,35 @@ private unsafe void TakeSingleHeader(byte* headerLine, int length, IHttpHeadersH
             }
 
             // Check for CR in value
-            var i = valueStart + 1;
-            if (Contains(headerLine + i, valueEnd - i, ByteCR))
+            var valueLength = valueEnd - valueStart + 1;
```
davidfowl: Get rid of this local, it isn't kept up to date.

Issue #1575, opened 2017-03-30
Respect PreferHostingUrls


natemcmaster:
We should update Kestrel to respect new options available on Hosting. See https://github.com/aspnet/Hosting/issues/968

PR #1639, opened 2017-04-09
Honor PreferHostingUrls

JunTaoLuo:
Addresses #1575.

We want to override the endpoints configured in the listen options when PreferHostingUrls is set to true with the addresses in the IServerAddressesFeature if it's not empty.

Will squash with a meaningful commit message once ready to merge.

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Kestrel.Core/KestrelServer.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel.Core/KestrelServer.cs
@@ -111,14 +111,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core
                 var hasListenOptions = listenOptions.Any();
                 var hasServerAddresses = _serverAddresses.Addresses.Any();
 
-                if (hasListenOptions && hasServerAddresses)
-                {
-                    var joined = string.Join(", ", _serverAddresses.Addresses);
-                    _logger.LogWarning($"Overriding address(es) '{joined}'. Binding to endpoints defined in UseKestrel() instead.");
-
-                    _serverAddresses.Addresses.Clear();
-                }
-                else if (!hasListenOptions && !hasServerAddresses)
+                if (!hasListenOptions && !hasServerAddresses)
                 {
                     _logger.LogDebug($"No listening endpoints were configured. Binding to {Constants.DefaultServerAddress} by default.");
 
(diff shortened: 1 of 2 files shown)


davidfowl:
nit: English spelling of Honour 😉

src/Microsoft.AspNetCore.Server.Kestrel.Core/KestrelServer.cs
```diff
@@ -131,8 +131,15 @@ public void Start<TContext>(IHttpApplication<TContext> application)
 
                     return;
                 }
-                else if (!hasListenOptions)
+                else if (!hasListenOptions || (_serverAddresses.PreferHostingUrls && hasServerAddresses))
```
JunTaoLuo: Not a big fan of how these if else are structured since it's somewhat convoluted. As far as I can tell, it's to reduce code duplication for translating server addresses to listen options and binding to the listen options.

I have an alternative approach added to the branch at https://github.com/aspnet/KestrelHttpServer/compare/johluo/use-hosting-urls?diff=split&expand=1&name=johluo%2Fuse-hosting-urls but there's a bit of moving things around so I'm not sure if it's worth the effort.

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/AddressRegistrationTests.cs
```diff
@@ -153,10 +153,7 @@ public Task DefaultsServerAddress_BindsToIPv6()
 
             var hostBuilder = new WebHostBuilder()
                .UseKestrel()
-               .ConfigureServices(services =>
-               {
-                   services.AddSingleton<ILoggerFactory>(new KestrelTestLoggerFactory(testLogger));
-               })
+               .UseLoggerFactory(_ => new KestrelTestLoggerFactory(testLogger))
```
davidfowl: Why not use the overload that takes an instance?

JunTaoLuo: We were thinking of obsoleting that overload. https://github.com/aspnet/Hosting/issues/1007. We want to use the overloads that also take in the WebHostBuilderContext.

src/Microsoft.AspNetCore.Server.Kestrel.Core/KestrelServer.cs
natemcmaster: I like the alternative approach; it's easier to follow the logic in that version.

JunTaoLuo:
closing in favour of https://github.com/aspnet/KestrelHttpServer/pull/1655

PR #1650, opened 2017-04-11
Don't fail tests if port 5000 is in use

halter73:
@pakrym


mikeharder:
I'm not sure about this change.  Does it mean these tests will never be run if port 5000 is always in use on a machine?  This seems dangerous.  We've seen issues in the past where a process will fail to shutdown and be using port 5000, and other tests failing helps detect this and drive fixes into the test which isn't cleaning up correctly.

mikeharder:
That said, if we want this change for other reasons, we can find a better way to ensure processes and ports are cleaned up before/after CI runs.  Relying on other tests to do this isn't really correct either.

pakrym:
@mikeharder we do it for port 80 and 443 already

mikeharder:
I think 80 and 443 are different, since they're standard ports which may be used by IIS, nginx, etc.  What's the reason for this change?  So unit tests will pass on a dev machine while port 5000 is in use?  I wouldn't make this change to help CI builds, I actually think this change is worse for the health of CI builds.

Issue #1585, opened 2017-03-31
Move shutdown logic to Core and make it transport agnostic


halter73:

halter73:
Fixed by #1707

PR #1653, opened 2017-04-11
Make IServer Start and Stop async

Tratcher:
React to https://github.com/aspnet/Hosting/pull/995
Will follow a file up bug for honoring the cancellation token. It will require a more significant refactor than we want to get into here.

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Kestrel.Core/KestrelServer.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel.Core/KestrelServer.cs
@@ -6,6 +6,7 @@ using System.Collections.Generic;
 using System.IO;
 using System.Linq;
 using System.Net;
+using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Builder;
 using Microsoft.AspNetCore.Hosting.Server;
@@ -29,6 +30,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core
         private readonly ITransportFactory _transportFactory;
 
         private bool _isRunning;
+        private int _stopped;
         private Heartbeat _heartbeat;
 
         public KestrelServer(
@@ -67,7 +69,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core
 
         private InternalKestrelServerOptions InternalOptions { get; }
 
-        public void Start<TContext>(IHttpApplication<TContext> application)
+        public async Task StartAsync<TContext>(IHttpApplication<TContext> application, CancellationToken cancellationToken)
         {
             try
             {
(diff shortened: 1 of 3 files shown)


src/Microsoft.AspNetCore.Server.Kestrel.Core/KestrelServer.cs
```diff
@@ -213,28 +215,41 @@ public void Start<TContext>(IHttpApplication<TContext> application)
             }
         }
 
-        public void Dispose()
+        // Graceful shutdown if possible
+        public async Task StopAsync(CancellationToken cancellationToken)
         {
+            _isStopping = true;
```
halter73: If you're going to have an `_isStopping` field, you might as well guard against parallel stops. Might want something like `if (Interlocked.Exchange(ref hasStartedStopping, 1) == 0) ...`.

src/Microsoft.AspNetCore.Server.Kestrel.Core/KestrelServer.cs
```diff
@@ -213,28 +215,41 @@ public void Start<TContext>(IHttpApplication<TContext> application)
...
+                await Task.WhenAll(tasks);
             }
 
             _dateHeaderValueManager?.Dispose();
         }
 
+        // Ungraceful shutdown
+        public void Dispose()
+        {
+            if (!_isStopping)
```
halter73: I would have StopAsync guard itself being called twice instead of checking it here. Just call StopAsync every time.

halter73:
The follow up work for honoring the passed in cancellationtoken will probably be part of the work for #1585.

Tratcher:
Updated to fix the exception handling.

src/Microsoft.AspNetCore.Server.Kestrel.Core/KestrelServer.cs
```diff
@@ -29,6 +30,7 @@ public class KestrelServer : IServer
         private readonly ITransportFactory _transportFactory;
 
         private bool _isRunning;
+        private int _isStopping;
```
halter73: Nit: Since this remains set after StopAsync completes, I prefer calling this _stopped instead of _isStopping.

src/Microsoft.AspNetCore.Server.Kestrel.Core/KestrelServer.cs
```diff
@@ -213,28 +215,41 @@ public void Start<TContext>(IHttpApplication<TContext> application)
...
         }
 
-        public void Dispose()
+        // Graceful shutdown if possible
+        public async Task StopAsync(CancellationToken cancellationToken)
         {
+            if (Interlocked.Exchange(ref _isStopping, 1) == 1)
+            {
+                return;
+            }
```
halter73: Nit: add newline.

PR #1658, opened 2017-04-11
Switching to the new toolset compiler.

VSadov:
The new version is
https://dotnet.myget.org/feed/roslyn/package/nuget/Microsoft.Net.Compilers/2.3.0-rdonly-ref-61607-04

The new version enforces the following rules:
- `Span<T>` and `ReadonlySpan<T>` are stack-only types. (no boxing, invalid use in async, etc...)
- `Span<T>` and `ReadonlySpan<T>` cannot be passed via a writeable parameter.

There were, however, no violations of the rules. (I had to introduce and then remove some violatins, just to se that detection works)

NOTE: we are not yet validating the usage of span-containing structs. There could be more indrect violations of stack safety through that.

### DIFF
--- a/NuGet.config
+++ b/NuGet.config
@@ -4,5 +4,6 @@
     <add key="AspNetCore" value="https://dotnet.myget.org/F/aspnetcore-ci-dev/api/v3/index.json" />
     <add key="AspNetCoreTools" value="https://dotnet.myget.org/F/aspnetcore-tools/api/v3/index.json" />
     <add key="NuGet" value="https://api.nuget.org/v3/index.json" />
+    <add key="roslyn" value="https://dotnet.myget.org/F/roslyn/" />
   </packageSources>
 </configuration>
--- a/build/common.props
+++ b/build/common.props
@@ -20,4 +20,7 @@
     <PackageReference Include="NETStandard.Library" Version="$(NetStandardImplicitPackageVersion)" />
   </ItemGroup>
 
+  <ItemGroup>
+    <PackageReference Include="Microsoft.Net.Compilers" Version="$(RoslynVersion)" PrivateAssets="All" />
+  </ItemGroup>
 </Project>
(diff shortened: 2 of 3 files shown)


VSadov:
CC: @davidfowl

davidfowl:
@pakrym @pranavkm Should we mirror this to our feed so we don't have to add the roslyn feed? It's a tool time reference only so it doesn't have the same impact as a runtime dependency.

davidfowl:
@VSadov for some reason this is broken on non-windows, any ideas why?

/cc @jaredpar

pakrym:
@davidfowl I'm afraid that it will cause same issues with building from command line when VS is running as is happening in corefxlab.

VSadov:
/home/travis/.nuget/packages/microsoft.net.compilers/2.3.0-rdonly-ref-61607-04/tools/Microsoft.CSharp.Core.targets(71,5): error MSB3883: Unexpected exception:  [/home/travis/build/aspnet/KestrelHttpServer/src/Microsoft.AspNetCore.Server.Kestrel.Transport.Abstractions/Microsoft.AspNetCore.Server.Kestrel.Transport.Abstractions.csproj]
/home/travis/.nuget/packages/microsoft.net.compilers/2.3.0-rdonly-ref-61607-04/tools/Microsoft.CSharp.Core.targets(71,5): error : System.ComponentModel.Win32Exception (0x80004005): Permission denied

VSadov:
Can this be some nuget issue with file IO permissions to restored packages?

VSadov:
@davidfowl - perhaps it is just something with test machine? Is there a way to rerun those tests?

VSadov:
test this please

agocke:
This is a 4.6 only package -- we currently do not have an easily consumable toolset compiler that runs on CoreCLR.

davidfowl:
@VSadov unfortunately that makes this a non starter...

davidfowl:
Closing until theres an xplat compiler

PR #1659, opened 2017-04-12
add Socket transport for Kestrel

geoffkizer:
This is failing 6 tests in Kestrel.FunctionalTests currently; the failures seem to be related to:
(1) Timeouts -- will be fixed in a subsequent PR to refactor timeout handling
(2) Logging -- not sure what needs to be done here; the logging stuff seems to be tied to libuv

@davidfowl, @halter73


src/Microsoft.AspNetCore.Server.Kestrel.Transport.Sockets/Microsoft.AspNetCore.Server.Kestrel.Transport.Sockets.csproj
```diff
@@ -0,0 +1,19 @@
...
+    <TargetFrameworks>netstandard1.3;net46</TargetFrameworks>
+    <GenerateDocumentationFile>true</GenerateDocumentationFile>
+    <PackageTags>aspnetcore;kestrel</PackageTags>
+    <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
+    <NoWarn>CS1591;$(NoWarn)</NoWarn>
+    <EnableApiCheck>false</EnableApiCheck>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="..\Microsoft.AspNetCore.Server.Kestrel.Core\Microsoft.AspNetCore.Server.Kestrel.Core.csproj" />
```
davidfowl: This shouldn't depend on any of the types in Core and should instead rely on Transport.Abstractions

src/Microsoft.AspNetCore.Server.Kestrel.Transport.Sockets/SocketConnection.cs
```diff
@@ -0,0 +1,213 @@
...
+            // We should always clear the list after the send
+            Debug.Assert(_sendBufferList.Count == 0);
+
+            foreach (var b in buffer)
+            {
+                _sendBufferList.Add(GetArraySegment(b));
+            }
+        }
+
+        private async Task DoSend()
```
davidfowl: This needs similar logic to the libuv version
- If the read is cancelled you should sent the FIN to the client.
- After that, you should close complete the reader.
- It also looks like you're not completing the reader which will leak memory. 
- There should an outer finally and calls Complete without an exception.
- You should call `context.OnConnectionClosed()` after this loop ends.

src/Microsoft.AspNetCore.Server.Kestrel.Transport.Sockets/SocketConnection.cs
```diff
@@ -0,0 +1,213 @@
...
+                    WritableBuffer buffer = _input.Alloc(MinAllocBufferSize);
+
+                    try
+                    {
+                        int bytesReceived = await _socket.ReceiveAsync(GetArraySegment(buffer.Buffer), SocketFlags.None);
+
+                        if (bytesReceived == 0)
+                        {
+                            // EOF
+                            _input.Complete();
```
davidfowl: Move this to an outer finally, it's easier to follow that way.

src/Microsoft.AspNetCore.Server.Kestrel.Transport.Sockets/SocketConnection.cs
```diff
@@ -0,0 +1,213 @@
...
+                            done = true;
+                            break;
+                        }
+
+                        // record what data we filled into the buffer and push to pipe
+                        buffer.Advance(bytesReceived);
+                    }
+                    finally
+                    {
+                        var result = await buffer.FlushAsync();
```
davidfowl: The interaction here is a bit more complicated that it probably should be but you need to call `context.Abort`. The fact that all of the tests pass worry me a bit because I'm sure there are some busted things.

src/Microsoft.AspNetCore.Server.Kestrel.Transport.Sockets/SocketTransport.cs
```diff
@@ -0,0 +1,127 @@
...
+
+            _listenSocket = null;
+            _listenTask = null;
+        }
+
+        public Task BindAsync()
+        {
+            if (_listenSocket != null)
+            {
+                throw new InvalidOperationException();
```
davidfowl: Add some exception text?

src/Microsoft.AspNetCore.Server.Kestrel.Transport.Sockets/SocketTransport.cs
```diff
@@ -0,0 +1,127 @@
...
+        {
+            if (_listenSocket != null)
+            {
+                var listenSocket = _listenSocket;
+                _listenSocket = null;
+
+                listenSocket.Dispose();
+
+                Debug.Assert(_listenTask != null);
+                await _listenTask;
```
davidfowl: This can throw right? @halter73 do we handle unbind throwing properly?

src/Microsoft.AspNetCore.Server.Kestrel.Transport.Sockets/SocketTransportFactory.cs
```diff
@@ -0,0 +1,41 @@
...
+
+using Microsoft.AspNetCore.Server.Kestrel.Internal.System.IO.Pipelines;
+using Microsoft.AspNetCore.Server.Kestrel.Transport.Abstractions;
+using System;
+
+namespace Microsoft.AspNetCore.Server.Kestrel.Transport.Sockets
+{
+    public sealed class SocketTransportFactory : ITransportFactory
+    {
+        private readonly PipeFactory _pipeFactory;
```
davidfowl: New it up inline instead of in the ctor

src/Microsoft.AspNetCore.Server.Kestrel.Transport.Sockets/SocketTransportFactory.cs
```diff
@@ -0,0 +1,41 @@
...
+        public ITransport Create(IEndPointInformation endPointInformation, IConnectionHandler handler)
+        {
+            if (endPointInformation == null)
+            {
+                throw new ArgumentNullException(nameof(endPointInformation));
+            }
+
+            if (endPointInformation.Type != ListenType.IPEndPoint)
+            {
+                throw new ArgumentException(nameof(endPointInformation));
```
davidfowl: Add some text in this exception saying what's wrong.

test/Microsoft.AspNetCore.Server.Kestrel.Core.Tests/FrameTests.cs
```diff
@@ -815,8 +815,10 @@ private class MockConnectionInformation : IConnectionInformation
             public IPEndPoint RemoteEndPoint { get; }
             public IPEndPoint LocalEndPoint { get; }
             public PipeFactory PipeFactory { get; }
+            public IScheduler InputReaderScheduler { get; }
```
davidfowl: Undo per comment above.

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/ChunkedResponseTests.cs
```diff
@@ -349,7 +349,7 @@ public class ChunkedResponseTests
                 await response.Body.WriteAsync(Encoding.ASCII.GetBytes("Hello "), 0, 6);
 
                 // Don't complete response until client has received the first chunk.
-                flushWh.Wait();
+                await Task.Run(() => flushWh.Wait());
```
davidfowl: Explain?

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/MaxRequestBufferSizeTests.cs
```diff
@@ -275,7 +275,7 @@ public static IEnumerable<object[]> LargeUploadData
                 .UseContentRoot(Directory.GetCurrentDirectory())
                 .Configure(app => app.Run(async context =>
                 {
-                    startReadingRequestBody.WaitOne();
+                    await Task.Run(() => startReadingRequestBody.WaitOne());
```
davidfowl: Explain?

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/MaxRequestBufferSizeTests.cs
```diff
@@ -284,7 +284,7 @@ public static IEnumerable<object[]> LargeUploadData
                         bytesRead += await context.Request.Body.ReadAsync(buffer, bytesRead, buffer.Length - bytesRead);
                     }
 
-                    clientFinishedSendingRequestBody.WaitOne();
+                    await Task.Run(() => clientFinishedSendingRequestBody.WaitOne());
```
davidfowl: Explain?

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/ResponseTests.cs
```diff
@@ -947,12 +947,11 @@ public Task ResponseStatusCodeSetBeforeHttpContextDisposedRequestMalformedReadIg
...
             var flushed = new SemaphoreSlim(0, 1);
 
-            using (var server = new TestServer(httpContext =>
+            using (var server = new TestServer(async httpContext =>
             {
                 httpContext.Response.ContentLength = 12;
                 httpContext.Response.Body.Write(Encoding.ASCII.GetBytes("hello, world"), 0, 12);
-                flushed.Wait();
-                return TaskCache.CompletedTask;
+                await Task.Run(() => flushed.Wait());
```
davidfowl: Explain?

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/TestServer.cs
```diff
@@ -46,37 +48,52 @@ public TestServer(RequestDelegate app, TestServiceContext context, ListenOptions
...
             catch
             {
                 _transport.UnbindAsync().Wait();
                 _transport.StopAsync().Wait();
+                _transport = null;
                 throw;
             }
         }
 
+        // Switch this to test on socket transport
```
davidfowl: Ah, this is why the tests all work 😄

test/Microsoft.AspNetCore.Server.Kestrel.Performance/Mocks/MockConnectionInformation.cs
```diff
@@ -13,8 +13,10 @@ public class MockConnectionInformation : IConnectionInformation
         public IPEndPoint LocalEndPoint { get; }
 
         public PipeFactory PipeFactory { get; }
+        public IScheduler InputReaderScheduler { get; }
```
davidfowl: Undo per comment above

src/Microsoft.AspNetCore.Server.Kestrel.Transport.Sockets/SocketConnection.cs
davidfowl: See 
https://github.com/geoffkizer/KestrelHttpServer/blob/5d59b631fa61cddff364632e410bca57e6ed95a7/src/Microsoft.AspNetCore.Server.Kestrel.Transport.Libuv/Internal/LibuvConnection.cs#L220-L223 for more details. I'm not sure how FIN is surfaced via the socket API.

src/Microsoft.AspNetCore.Server.Kestrel.Transport.Abstractions/IConnectionInformation.cs
```diff
@@ -12,9 +12,10 @@ public interface IConnectionInformation
         IPEndPoint LocalEndPoint { get; }
 
         PipeFactory PipeFactory { get; }
+        IScheduler InputReaderScheduler { get; }
```
geoffkizer: The motivation for this is to avoid unnecessary dispatch.  Socket completion calls are already running on the thread pool, so there is no reason to dispatch again to the thread pool.

I'm happy to do this some other way though; what do you suggest?

davidfowl: Undo these changes and leave the default dispatching to the thread pool. That dispatching to the application can be controlled via another flag directly in kestrel itself.

geoffkizer: What flag?  Or are you saying it should be a new flag?

benaadams: Is `ThreadPoolDispatching` in `InternalKestrelServerOptions`

Also avoiding double dispatch could be baked into the threadpool instead e.g. https://github.com/aspnet/KestrelHttpServer/pull/1626/files

davidfowl: Right now it's pretty hidden on purpose https://github.com/aspnet/KestrelHttpServer/blob/dev/src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/InternalKestrelServerOptions.cs#L10.

tmds: Should the user manually change the flag based on the Transport?

davidfowl: I just discussed this with @geoffkizer, I think we'll end up with a `RequiresDispatch` bool on `IConnectionContext` that will determine the default for a transport's default (stay on the same thread or dispatch to the thread pool). The option on KestrelOptions will override this of course.

tmds: An alternative would be to give the Transport control over the scheduler (here I go again). If meaningful for the Transport, it can expose dispatch control via a Transport specific option.

src/Microsoft.AspNetCore.Server.Kestrel.Transport.Sockets/Microsoft.AspNetCore.Server.Kestrel.Transport.Sockets.csproj
```diff
@@ -0,0 +1,19 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <Import Project="..\..\build\common.props" />
+
+  <PropertyGroup>
+    <Description>Managed socket transport for the ASP.NET Core Kestrel cross-platform web server.</Description>
+    <TargetFrameworks>netstandard1.3;net46</TargetFrameworks>
```
markrendle: That should be on posters all over campus.

src/Microsoft.AspNetCore.Server.Kestrel.Transport.Sockets/SocketTransport.cs
geoffkizer: Yes, it can throw.

Actually, in general, I am not sure what should be done with errors from Accept.  There doesn't seem to be a way to report them today; am I missing something?  These should be rare, but they can still happen.

davidfowl: It should be fine I think it propagates properly.

src/Microsoft.AspNetCore.Server.Kestrel.Transport.Sockets/SocketConnection.cs
geoffkizer: I added OnConnectionClosed, but I'm still seeing the failure in EmitsConnectionStartAndStop.  Any thoughts?  I'll add the rest and then dig into it further.

davidfowl: How does it fail? Are you sure it's getting called?

geoffkizer:
```
Failed   EmitsConnectionStartAndStop
Error Message:
 The collection contained 0 matching element(s) instead of 1.
Stack Trace:
   at Microsoft.AspNetCore.Server.Kestrel.FunctionalTests.EventSourceTests.<EmitsConnectionStartAndStop>d__2.MoveNext()
```

I haven't debugged, will after I've made the other changes here.

src/Microsoft.AspNetCore.Server.Kestrel.Transport.Sockets/SocketConnection.cs
geoffkizer: What should happen if FlushResult.IsCancelled is true?  It looks like libuv just ignores this.

geoffkizer: btw, socket.Shutdown will send the FIN.

davidfowl: Nobody cancelled this today.

davidfowl: BTW, we're making it so you don't have to call Abort, completing the pipe with an exception should be enough @halter73 should be making that change tomorrow.

geoffkizer: ok.  will Abort be removed?  I'm wondering if I should just wait for the change.

src/Microsoft.AspNetCore.Server.Kestrel.Transport.Sockets/SocketConnection.cs
geoffkizer: I can't do this because I don't want to call Complete in the case where the FlushResult returns IsCompleted.  (When I tried doing this, it caused unhandled exceptions.)

davidfowl: I see, I guess it's fine.

src/Microsoft.AspNetCore.Server.Kestrel.Transport.Sockets/SocketConnection.cs
davidfowl: Well it shouldn't block this PR is all.

src/Microsoft.AspNetCore.Server.Kestrel.Transport.Sockets/SocketConnection.cs
geoffkizer: nevermind, adding the Abort logic seems to have made this work.

src/Microsoft.AspNetCore.Server.Kestrel.Transport.Sockets/SocketTransport.cs
halter73: If Unbind throws we wont call Stop for any of the ITransports. This could be changed though.

src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/ConnectionHandler.cs
```diff
@@ -70,18 +70,18 @@ public IConnectionContext OnConnection(IConnectionInformation connectionInfo)
         }
 
         // Internal for testing
-        internal PipeOptions GetInputPipeOptions(IScheduler writerScheduler) => new PipeOptions
+        internal PipeOptions GetInputPipeOptions(bool requiresDispatch, IScheduler writerScheduler) => new PipeOptions
         {
-            ReaderScheduler = _serviceContext.ThreadPool,
+            ReaderScheduler = (requiresDispatch ? (IScheduler)_serviceContext.ThreadPool : (IScheduler)InlineScheduler.Default),
```
davidfowl: This doesn't require a cast

src/Microsoft.AspNetCore.Server.Kestrel.Transport.Sockets/SocketTransport.cs
davidfowl: @halter73 Let's file an issue for that.

src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/ConnectionHandler.cs
geoffkizer: it requires a cast for one of them -- i could remove one

davidfowl: Ah I see, leave it

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/MaxRequestBufferSizeTests.cs
```diff
@@ -172,78 +172,6 @@ public static IEnumerable<object[]> LargeUploadData
             }
         }
 
-        [Fact]
-        public async Task ServerShutsDownGracefullyWhenMaxRequestBufferSizeExceeded()
```
davidfowl: Mistake?

geoffkizer: yeah, will fix

PR #1661, opened 2017-04-12
Reject requests that don't have a Host header

natemcmaster:
**Reviewers**: I split into two commits to make it easier to see actual change. 
Commit 1 is adding "Host:" to almost every functional test. 
Commit 2 is the good stuff you want to read. ====> https://github.com/aspnet/KestrelHttpServer/pull/1661/commits/e42525bc23f7eb5a624b58ffb0b84a5c41937c06

----------

Resolves https://github.com/aspnet/KestrelHttpServer/issues/839. - HTTP/1.1 requires requests to have a Host: header. This updates Kestrel to reject with HTTP 400 Bad Request when request:

 - has 0 or many Host headers
 - has a host header that does not match the request target (only authority and absolute-form requests)

**Valid**
```
GET / HTTP/1.0

GET / HTTP/1.1
Host:

CONNECT www.contoso.com HTTP/1.1
Host: www.contoso.com

GET http://localhost.com/ HTTP/1.1
Host: localhost.com
```

**Invalid**
```
GET / HTTP/1.1

GET / HTTP/1.1
Host: one
Host: two

GET http://localhost/ HTTP/1.1
Host: localhost:9999

GET http://localhost/ HTTP/1.1
Host: example.org
```


natemcmaster:
## Benchmarks
**dev**
```
                                   Method |          Mean |     StdErr |      StdDev |        Median |       Op/s | Scaled | Allocated |
----------------------------------------- |-------------- |----------- |------------ |-------------- |----------- |------- |---------- |
                     PlaintextTechEmpower |   977.8449 ns |  8.4893 ns |  46.4977 ns |   967.9578 ns | 1022657.12 |   1.00 |     349 B |
                     PlaintextAbsoluteUri | 1,544.1159 ns | 19.2523 ns | 105.4491 ns | 1,523.5828 ns |  647619.78 |   1.58 |     573 B |
            PipelinedPlaintextTechEmpower |   871.8040 ns |  8.0088 ns |  43.8662 ns |   852.4480 ns | 1147046.79 |   0.89 |     348 B |
 PipelinedPlaintextTechEmpowerDrainBuffer |   744.9199 ns |  6.9866 ns |  38.2671 ns |   721.7044 ns | 1342426.24 |   0.76 |     348 B |
                               LiveAspNet | 1,919.2460 ns | 22.1606 ns | 121.3787 ns | 1,893.5222 ns |  521037.95 |   1.97 |   1.07 kB |
                      PipelinedLiveAspNet | 1,954.7165 ns | 17.4261 ns |  95.4467 ns | 2,010.9247 ns |  511583.13 |   2.00 |   1.07 kB |
                                  Unicode | 3,320.8422 ns | 42.7175 ns | 233.9736 ns | 3,249.0104 ns |  301128.43 |   3.40 |   1.85 kB |
                         UnicodePipelined | 3,468.9448 ns | 25.6115 ns | 140.2800 ns | 3,427.7552 ns |   288272.1 |   3.56 |   1.86 kB |
```

**namc/rfc-host-header**
```
                                   Method |          Mean |     StdErr |      StdDev |        Median |       Op/s | Scaled | Allocated |
----------------------------------------- |-------------- |----------- |------------ |-------------- |----------- |------- |---------- |
                     PlaintextTechEmpower |   998.9642 ns |  8.4538 ns |  46.3034 ns |   984.8956 ns | 1001036.84 |   1.00 |     349 B |
                     PlaintextAbsoluteUri | 1,558.9192 ns | 18.3909 ns | 100.7313 ns | 1,530.6435 ns |  641470.08 |   1.56 |     573 B |
            PipelinedPlaintextTechEmpower |   869.4526 ns | 10.8814 ns |  59.5999 ns |   847.1008 ns | 1150148.92 |   0.87 |     348 B |
 PipelinedPlaintextTechEmpowerDrainBuffer |   769.1206 ns |  9.9964 ns |  54.7523 ns |   745.1765 ns | 1300186.15 |   0.77 |     348 B |
                               LiveAspNet | 1,932.3517 ns | 19.1757 ns | 105.0296 ns | 1,895.5229 ns |  517504.13 |   1.94 |   1.07 kB |
                      PipelinedLiveAspNet | 1,941.2317 ns | 18.4719 ns | 101.1745 ns | 1,891.1152 ns |  515136.85 |   1.95 |   1.07 kB |
                                  Unicode | 3,317.8114 ns | 30.3530 ns | 166.2502 ns | 3,239.5499 ns |  301403.51 |   3.33 |   1.85 kB |
                         UnicodePipelined | 3,430.3654 ns | 31.7224 ns | 173.7509 ns | 3,442.8987 ns |  291514.13 |   3.44 |   1.86 kB |
```

src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/Http/Frame.cs
```diff
@@ -1421,5 +1447,57 @@ public void OnHeader(Span<byte> name, Span<byte> value)
...
         }
+
+        public void OnRequestParsingComplete()
+        {
+            ValidateHostHeader();
+        }
+
+        private void ValidateHostHeader()
+        {
+            if (_httpVersion == Http.HttpVersion.Http10)
```
Tratcher: <=?

natemcmaster: I don't think kestrel supports them. HttpVersion is an enum.
```
    public enum HttpVersion
    {
        Unknown = -1,
        Http10 = 0,
        Http11 = 1
    }
```

src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/Http/Frame.cs
```diff
@@ -1325,6 +1344,8 @@ private void OnOriginFormTarget(HttpMethod method, HttpVersion version, Span<byt
 
         private void OnAuthorityFormTarget(HttpMethod method, Span<byte> target)
         {
+            _requestTargetForm = HttpRequestTarget.AuthorityForm;
+
             // TODO Validate that target is a correct host[:port] string.
             // Reject as 400 if not. This is just a quick scan for invalid characters
             // but doesn't check that the target fully matches the URI spec.
```
halter73: Remove TODO

src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/Http/Frame.cs
```diff
@@ -1420,5 +1446,57 @@ public void OnHeader(Span<byte> name, Span<byte> value)
...
+            {
+                RejectRequest(RequestRejectionReason.MissingHostHeader);
+            }
+            else if (host.Count > 1)
+            {
+                RejectRequest(RequestRejectionReason.MultipleHostHeaders);
+            }
+            else if (_requestTargetForm == HttpRequestTarget.AuthorityForm)
+            {
+                if (!host.Equals(RawTarget))
```
halter73: Does this treat "www.example.com:80" and "www.example.com" as equivalent? Should it?

src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/Http/FrameOfT.cs
```diff
@@ -85,6 +85,8 @@ public Frame(IHttpApplication<TContext> application, FrameContext frameContext)
 
                     if (!_requestProcessingStopping)
                     {
+                        OnRequestParsingComplete();
```
halter73: I think it's clearer to call ValidateHostHeader() directly.

src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/Http/Frame.cs
```diff
@@ -1262,10 +1269,22 @@ public void OnStartLine(HttpMethod method, HttpVersion version, Span<byte> targe
             Debug.Assert(HttpVersion != null, "HttpVersion was not set");
         }
 
+        private enum HttpRequestTarget
```
halter73: Nit: put private classes/structs/enums at the bottom of the parent class.

src/Microsoft.AspNetCore.Server.Kestrel.Transport.Libuv/Internal/LibuvThread.cs
```diff
@@ -38,9 +38,17 @@ public class LibuvThread : IScheduler
         private ExceptionDispatchInfo _closeError;
         private readonly ILibuvTrace _log;
         private readonly TimeSpan _shutdownTimeout;
+#if DEBUG
+        private readonly string _stackTrace;
```
halter73: While useful for debugging, I don't like adding this since Environment.StackTrace can really slow tests down. Even though it only affects DEBUG builds, I don't want to slow down development.

This is super trivial to add manually if you are trying to find out which test is crashing. To be useful, you would want to print the stack trace somewhere anyway.

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/TestServer.cs
```diff
@@ -65,8 +65,11 @@ public TestServer(RequestDelegate app, TestServiceContext context, ListenOptions
             }
             catch
             {
-                _transport.UnbindAsync().Wait();
-                _transport.StopAsync().Wait();
+                if (_transport != null)
```
halter73: Is the libuvtransport ctor throwing? Why?

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/ResponseTests.cs
```diff
@@ -1073,10 +1095,11 @@ public Task ResponseStatusCodeSetBeforeHttpContextDisposedRequestMalformedReadIg
                 {
                     await connection.Send(
                         "POST / HTTP/1.1",
+                        "Host:",
                         "Transfer-Encoding: chunked",
                         "",
                         "gg");
-                    await responseWritten.WaitAsync();
+                    await responseWritten.WaitAsync().TimeoutAfter(TimeSpan.FromSeconds(10));
```
halter73: Any particular reason for a 10s timeout here when you used 30s elsewhere?

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests.csproj
```diff
@@ -8,6 +8,7 @@
     <PlatformTarget Condition="'$(TargetFramework)' == 'net46'">x64</PlatformTarget>
     <AutoGenerateBindingRedirects>true</AutoGenerateBindingRedirects>
     <GenerateBindingRedirectsOutputType>true</GenerateBindingRedirectsOutputType>
+    <ServerGarbageCollection>true</ServerGarbageCollection>
```
halter73: Does this speed up the tests noticeably?

halter73:
So we've decided not to validate the host header vs addresses passed in IServerAddressesFeature? If so, I'm absolutely fine with that.

Drawaes:
Is there any possible security implications of a client requesting a host name of X during the TLS handshake (when that is possible) but then a different host header in the HTTP request? I am trying to think of potential security problems with it, and at the moment I haven't come up with any, but maybe others can?

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests.csproj
natemcmaster: Now that tests are fixed, I'll try and measure it. I set this so that our functional tests use the same GC setting Kestrel would normally run under.

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/ResponseTests.cs
natemcmaster: other than my impatience, no. I don't think we're consistent in which timeout we use.

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/TestServer.cs
natemcmaster: Yes, it was throwing. The NRE thrown from the catch was hiding the underlying issue so I didn't figure out what was going on.

src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/Http/Frame.cs
natemcmaster: The spec says
 
> If the target URI includes an authority component, then a
   client MUST send a field-value for Host that is identical to that
   authority component, excluding any userinfo subcomponent and its "@"
   delimiter (Section 2.7.1).

I interpret identical to mean "include port in both places". So no, "www.example.com:80" and "www.example.com" as not equivalent

natemcmaster:
@Drawaes the validation I'm adding here is about validating form, not content. I think you're implying a good question, though: should Kestrel validate the _content_ of Host header? My current thinking: it's an application-level concern as Kestrel routes all requests into the app, regardless of the Host header. You could write middleware that does "`if(allowedHosts.DoesNotContains(HostHeader)) { Reject() }`". 

That said, it's a good question, worthy of opening a new issue.

natemcmaster:
> So we've decided not to validate the host header vs addresses passed in IServerAddressesFeature? 

We haven't decided that yet. This is only about validating that the header exists. We should discuss this more.

natemcmaster:
===> this https://github.com/aspnet/KestrelHttpServer/issues/1688#issuecomment-294215623

natemcmaster:
:up: :date: pr feedback

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/ResponseTests.cs
halter73: I'd use 30s then. It only require more patience if it fails, and potentially makes the test less flaky.

Issue #1495, opened 2017-03-15
Move IConnectionFilter to internal namespace


davidfowl:
This API is going to change a bunch so we should internalize it.

Drawaes:
So you can't write connection filters anymore? I have them in prod so there will be a version of released kestrel that can't be extended?

davidfowl:
> So you can't write connection filters anymore? I have them in prod so there will be a version of released kestrel that can't be extended?

It's already broken in 2.0. We *could* keep them public and break them again in the next version but I'd rather not do that. The .Internal namespace lets you write filters but tells you this isn't the final API.

Drawaes:
Ahh so .internal but not internal class modifier... cool have at it

muratg:
@DamianEdwards Are you OK with this?  This will prevent breaking users (again) when we expose pipelines in the future.

PR #1664, opened 2017-04-12
Remove ListenOptions.Scheme and move IConnectionAdapter to .Internal namespace

natemcmaster:
We know for sure we don't want ListenOptions.Scheme as public API when we release 2.0, so let's get rid of it now.

This fixes https://github.com/aspnet/KestrelHttpServer/issues/1296.

Resolves https://github.com/aspnet/KestrelHttpServer/issues/1495

This is obviously a temporary fix as the design of connection adapters and configuration will churn before release. I will open a follow-up issue to create a better design.

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Kestrel.Core/KestrelServer.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel.Core/KestrelServer.cs
@@ -153,6 +153,10 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core
                         {
                             throw new InvalidOperationException($"HTTPS endpoints can only be configured using {nameof(KestrelServerOptions)}.{nameof(KestrelServerOptions.Listen)}().");
                         }
+                        else if (!parsedAddress.Scheme.Equals("http", StringComparison.OrdinalIgnoreCase))
+                        {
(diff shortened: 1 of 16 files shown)


natemcmaster:
image

natemcmaster:
:up: :date: - moved IConnectionAdapter to .Internal and add `bool IsHttps`

src/Microsoft.AspNetCore.Server.Kestrel.Core/KestrelServer.cs
```diff
@@ -145,6 +145,10 @@ public void Start<TContext>(IHttpApplication<TContext> application)
                         {
                             throw new InvalidOperationException($"HTTPS endpoints can only be configured using {nameof(KestrelServerOptions)}.{nameof(KestrelServerOptions.Listen)}().");
                         }
+                        else if (!parsedAddress.Scheme.Equals("http", StringComparison.OrdinalIgnoreCase))
+                        {
+                            throw new InvalidOperationException($"Unrecognized scheme in server address '{address}'. Only 'http://' is supported.");
```
davidfowl: Is there a test for this case?

natemcmaster: There is now :)

PR #1670, opened 2017-04-13
Add ConfigureAwaits to prevent test deadlocks

Tratcher:
Adding ConfigureAwait to Hosting was not adequate, they needed to be in Kestrel. Hitting some weird asserts that I can't tell if they're related.

Still running another test pass without the hosting ones to see if they're redundant.


Tratcher:
The hosting ones are also required.

Merging to unblock builds.

PR #1674, opened 2017-04-13
[1.0.4] Call ProduceEnd() before consuming request body if response has already started

natemcmaster:
This fix was authored by @cesarbs in #1537. I've verified #1530 repros in 1.0.3 and that this change fixed it.

Close https://github.com/aspnet/KestrelHttpServer/issues/1544

### DIFF
--- a/global.json
+++ b/global.json
@@ -1,6 +1,9 @@
 {
-    "projects": ["src"],
-    "sdk": {
-      "version": "1.0.0-preview2-003154"
-    }
+  "projects": [
+    "src",
+    "test"
+  ],
+  "sdk": {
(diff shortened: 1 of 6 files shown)


global.json
```diff
@@ -1,6 +1,9 @@
 {
-    "projects": ["src"],
-    "sdk": {
-      "version": "1.0.0-preview2-003154"
-    }
+  "projects": [
+    "src",
+    "test"
```
halter73: Why does test need to be added?

natemcmaster: In our 1.0 code, TestServer is in KestrelTests, not FunctionalTests. Had to add a p2p reference from FunctionalTests to KestrelTests.

PR #1238, opened 2016-11-28
Don't log stack traces for localhost bind failures.

Tratcher:
#1001 
We're getting almost daily hits on this issue. A non-fatal error is being displayed like a fatal error and it distracts users from their real problems. Azure may or may not ever change their implementation.
@muratg @halter73 

I also recommend we backport this to 1.0.x and 1.1.x.

Before:
```
dbug: Microsoft.AspNetCore.Hosting.Internal.WebHost[3]
      Hosting starting
warn: Microsoft.AspNetCore.Server.Kestrel[0]
      Unable to bind to http://localhost:19786 on the IPv6 loopback interface.
System.AggregateException: One or more errors occurred. ---> Microsoft.AspNetCore.Server.Kestrel.Internal.Networking.UvException: Error -4089 EAFNOSUPPORT address family not supported
   at Microsoft.AspNetCore.Server.Kestrel.Internal.Networking.Libuv.tcp_bind(UvTcpHandle handle, SockAddr& addr, Int32 flags) in C:\git\Universe\KestrelHttpServer\src\Microsoft.AspNetCore.Server.Kestrel\Internal\Networking\Libuv.cs:line 202
   at Microsoft.AspNetCore.Server.Kestrel.Internal.Networking.UvTcpHandle.Bind(ServerAddress address) in C:\git\Universe\KestrelHttpServer\src\Microsoft.AspNetCore.Server.Kestrel\Internal\Networking\UvTcpHandle.cs:line 48
   at Microsoft.AspNetCore.Server.Kestrel.Internal.Http.TcpListenerPrimary.CreateListenSocket() in C:\git\Universe\KestrelHttpServer\src\Microsoft.AspNetCore.Server.Kestrel\Internal\Http\TcpListenerPrimary.cs:line 39
   at Microsoft.AspNetCore.Server.Kestrel.Internal.Http.Listener.<StartAsync>b__8_0(Object state) in C:\git\Universe\KestrelHttpServer\src\Microsoft.AspNetCore.Server.Kestrel\Internal\Http\Listener.cs:line 39
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
[...]

Issue #1474, opened 2017-03-10
Microsoft.AspNetCore.Server.Kestrel.Internal.Networking.UvException: Error -97 EAFNOSUPPORT address family not supported


MarkusAmshove:
Hello,

I'm using the ASP.NET Core project template from Visual Studio 2017 and built it with docker, pushed it to the registry and then pulled it on my QNAP NAS.

When trying to start the container I get the following stacktrace:

```
crit: Microsoft.AspNetCore.Server.Kestrel[0]                                                                                                                                                                                                                                                                                                                                             
      Unable to start Kestrel.                                                                                                                                                                                                                                                                                                                                                           
System.AggregateException: One or more errors occurred. (Error -97 EAFNOSUPPORT address family not supported) ---> Microsoft.AspNetCore.Server.Kestrel.Internal.Networking.UvException: Error -97 EAFNOSUPPORT address family not supported                                                                                                                                              
   at Microsoft.AspNetCore.Server.Kestrel.Internal.Networking.Libuv.tcp_bind(UvTcpHandle handle, SockAddr& addr, Int32 flags)                                                                                                                                                                                                                                                            
   at Microsoft.AspNetCore.Server.Kestrel.Internal.Networking.UvTcpHandle.Bind(ServerAddress address)                                                                                                                                                                                                                                                                                    
   at Microsoft.AspNetCore.Server.Kestrel.Internal.Http.TcpListenerPrimary.CreateListenSocket()                                                                                                                                                                                                                                                                                          
   at Microsoft.AspNetCore.Server.Kestrel.Internal.Http.Listener.<StartAsync>b__8_0(Object state)                                                                                                                                                                                                                                                                                        
--- End of stack trace from previous location where exception was thrown ---                                                                                                                                                                                                                                                                                                             
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()                                                                                                                                                                                                                                                                                                                     
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)                                                                                                                                                                                                                                                                                     
   at System.Runtime.CompilerServices.ConfiguredTaskAwaitable.ConfiguredTaskAwaiter.GetResult()                                                                                                                                                                                                                                                                                          
   at Microsoft.AspNetCore.Server.Kestrel.Internal.Http.ListenerPrimary.<StartAsync>d__12.MoveNext()                                                                                                                                                                                                                                                                                     
[...]

PR #1677, opened 2017-04-13
[1.1.2] Don't log stack traces for localhost bind failures

### LINKED ISSUE #1629: [1.1.2] Don't log stack traces for localhost bind failures (port fix to 1.1.x)
Port this fix to 1.1.x: https://github.com/aspnet/KestrelHttpServer/pull/1238

natemcmaster:
#1238, backported to 1.1.x.

Resolves #1629


natemcmaster:
Still trying to find a machine that will throw EAFNOSUPPORT for ipv6. Originally had tried to make Linux do it, but it seems even w/ ipv6 disabled, libuv can still bind to ipv6 loopback. The original issue mentioned that this error also happens in Azure. I'll follow up on this in https://github.com/aspnet/KestrelHttpServer/issues/1474

PR #1695, opened 2017-04-16
Expose a UseTransportThread property on KestrelServerOptions

davidfowl:
- This property will force Kestrel to use whatever scheduler the transport
used when write and read callbacks are fired. The default value is false so
all calls to user code including connection adapters, and the application function,
and cancellation token callbacks.
- Transports may expose configuration that changes what the transport thread is.
- Removed InternalKestrelServerOptions.cs
- Added a configurable UseSockets overload (even though there are no options yet)
- Remove RequiresDispatch from the IConnectionInformation
- Made writes dispatch to the thread pool for writes in the Socket transport


samples/SampleApp/Startup.cs
```diff
@@ -65,14 +69,12 @@ public static void Main(string[] args)
                 .UseLibuv(options =>
                 {
                     // Uncomment the following line to change the default number of libuv threads for all endpoints.
-                    options.ThreadCount = 4;
+                    // options.ThreadCount = 4;
```
halter73: Why are you putting unrelated changes in this PR? JK.

src/Microsoft.AspNetCore.Server.Kestrel.Transport.Sockets/SocketConnection.cs
```diff
@@ -215,10 +215,8 @@ private static ArraySegment<byte> GetArraySegment(Buffer<byte> buffer)
 
         public PipeFactory PipeFactory => _transport.TransportFactory.PipeFactory;
 
-        public bool RequiresDispatch => _transport.TransportFactory.ForceDispatch;
-
         public IScheduler InputWriterScheduler => InlineScheduler.Default;
 
-        public IScheduler OutputReaderScheduler => InlineScheduler.Default;
```
halter73: More unrelated changes...

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/TestServer.cs
```diff
@@ -49,7 +49,11 @@ public TestServer(RequestDelegate app, TestServiceContext context, ListenOptions
 
             Context = context;
 
-            _transport = s_transportFactory.Create(listenOptions, new ConnectionHandler<HttpContext>(listenOptions, context, new DummyApplication(app, httpContextFactory)));
+            // Switch this to test on socket transport
+            var transportFactory = CreateLibuvTransportFactory(context);
+            // var transportFactory = CreateSocketTransportFactory(context);
```
halter73: Are you adding a base class or something to test both transports? If not, remove comment.

src/Microsoft.AspNetCore.Server.Kestrel.Core/KestrelServerOptions.cs
```diff
@@ -29,6 +29,11 @@ public class KestrelServerOptions
         public bool AddServerHeader { get; set; } = true;
 
         /// <summary>
+        /// Gets or sets a value that determines if Kestrel should use the transport thread thread when executing user code.
```
halter73: Now this isn't pubternal, can we try scaring people away from using this in the doc comments?

PR #1698, opened 2017-04-16
Make all UvRequest objects normal GC handles

davidfowl:
- This avoids the cost of using GCHandle.Alloc per operation.
- It *does* mean that we need to explicitly dispose UvRequest objects
after using them (which we did before anyways). This change does
add a few try catch statements to make sure we always dispose the UvRequest
if there are synchronous exceptions.
- This is ~1.5% of the overhead in the benchmarks today

/cc @vancem

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Kestrel.Transport.Libuv/Internal/LibuvOutputConsumer.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel.Transport.Libuv/Internal/LibuvOutputConsumer.cs
@@ -47,14 +47,22 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Transport.Libuv.Internal
                     if (!buffer.IsEmpty)
                     {
                         var writeReq = _writeReqPool.Allocate();
-                        var writeResult = await writeReq.WriteAsync(_socket, buffer);
-                        _writeReqPool.Return(writeReq);
 
-                        LogWriteInfo(writeResult.Status, writeResult.Error);
-
-                        if (writeResult.Error != null)
+                        try
                         {
-                            throw writeResult.Error;
+                            var writeResult = await writeReq.WriteAsync(_socket, buffer);
+
+                            LogWriteInfo(writeResult.Status, writeResult.Error);
+
+                            if (writeResult.Error != null)
+                            {
+                                throw writeResult.Error;
+                            }
+                        }
+                        finally
+                        {
+                            // Make sure we return the writeReq to the pool
+                            _writeReqPool.Return(writeReq);
                         }
                     }
 
@@ -104,15 +112,23 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Transport.Libuv.Internal
             _log.ConnectionWriteFin(_connectionId);
 
             var shutdownReq = new UvShutdownReq(_log);
-            shutdownReq.Init(_thread.Loop);
-            shutdownReq.Shutdown(_socket, (req, status, state) =>
+            try
             {
-                req.Dispose();
-                _log.ConnectionWroteFin(_connectionId, status);
+                shutdownReq.Init(_thread.Loop);
+                shutdownReq.Shutdown(_socket, (req, status, state) =>
+                {
+                    req.Dispose();
+                    _log.ConnectionWroteFin(_connectionId, status);
 
-                tcs.TrySetResult(null);
-            },
-            this);
+                    tcs.TrySetResult(null);
+                },
+                this);
+            }
+            catch (Exception)
+            {
+                shutdownReq.Dispose();
+                throw;
+            }
 
             return tcs.Task;
         }
--- a/src/Microsoft.AspNetCore.Server.Kestrel.Transport.Libuv/Internal/LibuvOutputConsumer.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel.Transport.Libuv/Internal/LibuvOutputConsumer.cs
@@ -114,7 +114,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Transport.Libuv.Internal
             var shutdownReq = new UvShutdownReq(_log);
             try
             {
-                shutdownReq.Init(_thread.Loop);
+                shutdownReq.Init(_thread);
                 shutdownReq.Shutdown(_socket, (req, status, state) =>
                 {
                     req.Dispose();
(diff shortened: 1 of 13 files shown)


halter73:
Do you have before and after numbers for the techempower benchmarks

How are we going to prevent memory leaks from not disposing requests? Are we 100% sure that we are disposing requests everywhere we should be currently (or once this PR is merged). As you mentioned, we should have been disposing requests before anyway, but now the consequences are more sever if we don't.

davidfowl:
> Do you have before and after numbers for the techempower benchmarks

I'll get these tomorrow.

> How are we going to prevent memory leaks from not disposing requests?

We can add some logic that makes sure things aren't getting finalized. I'll look into this.

> Are we 100% sure that we are disposing requests everywhere we should be currently (or once this PR is merged). As you mentioned, we should have been disposing requests before anyway, but now the consequences are more sever if we don't.

I'll do another look but there were some places we didn't dispose before and this change fixes that. The only time `UvRequest` objects won't get collected is if for some reason a callback fails to be called (and I'm not sure if that's possible).

vancem:
Because UvRequests are often (mostly) used with Asyc calls where they are being 'used' by native code, I LIKE the approach that they live forever until explicitly disposed (which is what happens if they contain a normal GC handle that points at themselves).   This allow you to 'just us them' for async operations without worrying about how they are kept alive (they are alive until disposed), which avoids overhead, which is great.

I like the idea of a DEBUG logic to insure that we are not leaking them (I don't think you can do this with finalizers, since they are only called when things DIE not when the leak), however you can have DEBUG logic that keeps a list of WeakHandles when UVRequests are constructed.   Then you can either

1) Shut down all users/pools of UVRequests and see if any weak handles are still alive OR
2) Simply assert from time to time that the number (e.g. in constructor) of live UVRequests is under a reasonable bound.   If they are leaking it will fire.  This is not as good as (1) but easier to implement, and is much better than nothing (so it could be step one).

davidfowl:
Added leak detection (and it did find a leak 😄 )

vancem:
LGTM

src/Microsoft.AspNetCore.Server.Kestrel.Transport.Libuv/Internal/LibuvOutputConsumer.cs
```diff
@@ -47,14 +47,22 @@ public class LibuvOutputConsumer
...
+                            if (writeResult.Error != null)
+                            {
+                                throw writeResult.Error;
+                            }
+                        }
+                        finally
                         {
-                            throw writeResult.Error;
+                            // Make sure we return the writeReq to the pool
+                            _writeReqPool.Return(writeReq);
```
halter73: Why add a nested try/finally? Can WriteAsync throw? This code makes it look like it could. What was wrong with the old way of returning this?

src/Microsoft.AspNetCore.Server.Kestrel.Transport.Libuv/Internal/LibuvOutputConsumer.cs
```diff
@@ -104,15 +112,23 @@ private Task ShutdownAsync()
...
-                tcs.TrySetResult(null);
-            },
-            this);
+                    tcs.TrySetResult(null);
+                },
+                this);
+            }
+            catch (Exception)
+            {
+                shutdownReq.Dispose();
```
halter73: When does this actually throw?

src/Microsoft.AspNetCore.Server.Kestrel.Transport.Libuv/Internal/LibuvThread.cs
```diff
@@ -141,6 +146,21 @@ public Task StartAsync()
             }
         }
 
+#if DEBUG
+        private void CheckUvReqLeaks()
```
halter73: Love it.

src/Microsoft.AspNetCore.Server.Kestrel.Transport.Libuv/Internal/Networking/UvWriteReq.cs
```diff
@@ -31,6 +31,13 @@ public UvWriteReq(ILibuvTrace logger) : base(logger)
...
         }
 
+        public override void Init(LibuvThread thread)
+        {
+            Init(thread.Loop);
+
+            base.Init(thread);
+        }
+
         public void Init(UvLoopHandle loop)
```
halter73: We should make the `Init(UvLoopHandle loop)` method private, so it doesn't accidentally get called causing a Req object to not be tracked. That would be a really easy mistake to make.

src/Microsoft.AspNetCore.Server.Kestrel.Transport.Libuv/Internal/ListenerSecondary.cs
```diff
@@ -117,10 +118,13 @@ private static void ConnectCallback(UvConnectRequest connect, int status, Except
             }
             catch (Exception ex)
             {
-                writeReq.Dispose();
                 DispatchPipe.Dispose();
                 tcs.SetException(ex);
             }
+            finally
+            {
+                writeReq.Dispose();
```
halter73: Nice catch.

src/Microsoft.AspNetCore.Server.Kestrel.Transport.Libuv/Internal/Networking/UvWriteReq.cs
davidfowl: Yea thats how I started it but I don't want to refactor those tests as part of this change. It would need to be internal and it's still accessible from all of the code in there.

src/Microsoft.AspNetCore.Server.Kestrel.Transport.Libuv/Internal/LibuvOutputConsumer.cs
davidfowl: It should never throw, but if it does... 😄

src/Microsoft.AspNetCore.Server.Kestrel.Transport.Libuv/Internal/LibuvOutputConsumer.cs
davidfowl: It can throw if there's an issue with the `WriteReq`. I've seen it throw in the past when code was moved around.

src/Microsoft.AspNetCore.Server.Kestrel.Transport.Libuv/Internal/Networking/UvWriteReq.cs
halter73: Can you use LibuvThread/MockLibuv and use the other init ctor.

PR #1710, opened 2017-04-18
Add option to disable dispatching writes to thread pool

davidfowl:
- The socket transport gets free output buffering when writes are dispatched to the thread pool.
We want to allow experimentation where callbacks all run on the same thread.

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Kestrel.Transport.Sockets/SocketConnection.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel.Transport.Sockets/SocketConnection.cs
@@ -36,6 +36,9 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Transport.Sockets
 
             _localEndPoint = (IPEndPoint)_socket.LocalEndPoint;
             _remoteEndPoint = (IPEndPoint)_socket.RemoteEndPoint;
+
+            OutputReaderScheduler = _transport.TransportFactory.Options.DispatchWritesToThreadPool ?
+                                    (IScheduler)TaskRunScheduler.Default : InlineScheduler.Default;
         }
 
         public async void Start(IConnectionHandler connectionHandler)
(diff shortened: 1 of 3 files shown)


halter73:
Maybe we should rename UseTransportThread to DispatchApplicationToThreadPool and invert the logic for consistency.

cesarbs:
I'd like to suggest `UseThreadPoolDispatch`.

halter73:
@cesarbs That is very close to my original name `ThreadPoolDispatching` which @davidfowl apparently didn't like considering he changed it to `UseTransportThread`.

tmds:
Shorter and less complex sounding names:
`DispatchWritesToThreadPool` -> `DeferWrites`
`UseTransportThread` -> `DispatchRequests`

tmds:
I find the `UseTransportThread` improperly placed on the `KestrelServerOptions`.
For some transports, requests will be dispatched already, so this causes double dispatch.
For other transports, not dispatching may be inappropriate and cause issues.

davidfowl:
@tmds yes, we but kestrel dispatches more than just the transport callbacks.

> @cesarbs That is very close to my original name ThreadPoolDispatching which @davidfowl apparently didn't like considering he changed it to UseTransportThread.

The other reason I picked `UseTransportThread` is because defaulting bools to false is just better 😄 . I'll change the name of the bool (again) though.

davidfowl:
Here's the design we're going to use:

```C#
public interface IConnectionInformation
{
    IScheduler ApplicationScheduler { get; }
}

public class KestrelServerOptions
{
    public SchedulingMode ApplicationSchedulingMode { get; set; }
}

public enum SchedulingMode
{
    Default,
    ThreadPool,
    Inline
}
```

tmds:
I don't understand the design. Why does the transport pass the applicationscheduler to the application?

davidfowl:
@tmds Kestrel.Core has the mode exposed on its options. If the mode is set to default then the transport gets to pick the scheduler. This lets the transport pick where application callbacks run. By default, our transports will dispatch application code to the thread pool when appropriate. For example, on linux, the sockets transports already dispatches so there's no need to do it again. On windows however, callbacks can run on the IO thread, so it would dispatch by default.

If the user always wants to dispatch despite what the transports pick for the application scheduler, then it can override by setting the SchedulingMode to TheadPool. If the user wants to run on whatever thread they happen to be on basically disabling all dispatching then they can choose Inline.

tmds:
Ok, I get it

Issue #1684, opened 2017-04-14
Flaky Test: TestKeepAliveTimeout


davidfowl:
# Latest failure message

```
[TestKeepAliveTimeout] System.IO.IOException : Unable to transfer data on the transport connection: An established connection was aborted by the software in your host machine.
---- System.Net.Sockets.SocketException : An established connection was aborted by the software in your host machine
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at System.IO.StreamWriter.<FlushAsyncInternal>d__65.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at Microsoft.AspNetCore.Testing.TestConnection.<Send>d__11.MoveNext() in C:\BuildAgent\work\33bdfc1cae7b2a38\.r\KestrelHttpServer\test\shared\TestConnection.cs:line 68
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at Microsoft.AspNetCore.Server.Kestrel.FunctionalTests.KeepAliveTimeoutTests.<ConnectionKeptAliveBetweenRequests>d__5.MoveNext() in C:\BuildAgent\work\33bdfc1cae7b2a38\.r\KestrelHttpServer\test\Microsoft.AspNetCore.Server.Kestrel.FunctionalTests\KeepAliveTimeoutTests.cs:line 73
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
[...]

PR #1714, opened 2017-04-18
Disabling flaky TestKeepAliveTimeout test

Tratcher:
#1684 Disabling to unblock builds.


halter73:
This is testing a very high churn area of the code base and is the only test that really validates that keep alive timeouts work. @cesarbs Can we modify the timeouts to make this more reliable?

cesarbs:
@halter73 We could try, but the timeouts are already fairly high. I would like to see some evidence that these are really failing due to test slowness before increasing the timeouts.

halter73:
I think this test is far more stable post a053ca475822f4b4eb5d52a40b880afb59cb0244 and e4ba1d01ce81d836641246ddf4f68c56d8fe3f41. We can reopen if we see more failures.

PR #1715, opened 2017-04-18
Skip ConnectionsGetRoundRobinedToSecondaryListeners (#1691)

pakrym:


halter73:
I don't think skipping this test is terrible, but I think we're going to make a test fix in dev that *could* be backported.

pakrym:
Merging for now, after @natemcmaster fixes it we are going to backport.

pranavkm:
Thanks @halter73 - we're trying to get a build out of patch and this one's blocking it. I'd be happy if we can fix the actual test, but the first step is to unblock the build while we investigate a solution.

PR #1717, opened 2017-04-19
Move address binding into separate class and handle EAFNOSUPPORT

natemcmaster:
- Simplify KestrelServer by moving address binding into a separate class
 - Use strategy pattern to implement address binding for different scenarios
 - Add fallback from binding 0.0.0.0 if binding to [::] fails (can happen if UvException with EAFNOSUPPORT is thrown)

Closes https://github.com/aspnet/KestrelHttpServer/issues/1474


natemcmaster:
TODO:

 - [ ] Manually verify this fixes errors with binding to `http://*:80` on Antares

test/Microsoft.AspNetCore.Server.Kestrel.Core.Tests/AddressBinderTests.cs
```diff
@@ -0,0 +1,91 @@
...
+                options,
+                NullLogger.Instance,
+                endpoint => throw new AddressInUseException("already in use")));
+        }
+
+        [Theory]
+        [InlineData("http://*:80")]
+        [InlineData("http://+:80")]
+        [InlineData("http://contoso.com:80")]
+        public async Task FallbackToIpV4Any(string address)
```
cesarbs: spelling nit: IPv4

test/Microsoft.AspNetCore.Server.Kestrel.Core.Tests/AddressBinderTests.cs
```diff
@@ -0,0 +1,91 @@
...
+        {
+            Assert.True(AddressBinder.TryCreateIPEndPoint(
+                ServerAddress.FromUrl($"http://{host}:5000/"), out var endpoint));
+            Assert.NotNull(endpoint);
+            Assert.Equal(IPAddress.Parse(expectedAddress), endpoint.Address);
+            Assert.Equal(5000, endpoint.Port);
+        }
+
+        [Theory]
+        [InlineData("*")]
```
cesarbs: So we're dropping `*`?

davidfowl:
Something is hanging on the CI

test/Microsoft.AspNetCore.Server.Kestrel.Core.Tests/AddressBinderTests.cs
natemcmaster: Still supported. This is not a valid IP address so we fallback to IPv6Any.

natemcmaster:
Passes locally in Release mode. :-/ If only the test framework could tell me which test hung.

src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/AddressBinder.cs
```diff
@@ -0,0 +1,307 @@
...
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Builder;
+using Microsoft.AspNetCore.Hosting.Server.Features;
+using Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Infrastructure;
+using Microsoft.AspNetCore.Server.Kestrel.Transport.Abstractions;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal
+{
+    class AddressBinder
```
Tratcher: visibility

src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/AddressBinder.cs
```diff
@@ -0,0 +1,307 @@
...
+                        options = new ListenOptions(new IPEndPoint(IPAddress.Any, parsedAddress.Port));
+                        await BindEndpointAsync(options, context);
+                    }
+                }
+
+                context.Addresses.Add(options.GetDisplayName());
+            }
+        }
+
+        private interface IBindStrategy
```
Tratcher: The name of your chosen pattern is leaking into your API names, viraly. IBinder, DefaultAddressBinder, etc..?

test/Microsoft.AspNetCore.Server.Kestrel.Core.Tests/AddressBinderTests.cs
cesarbs: The test name is confusing then. Let's name it `DefaultsToIPv6AnyOnInvalidIPAddress`, or something like that.

natemcmaster: Only problem with `DefaultsToIPv6AnyOnInvalidIPAddress` is that this test doesn't actually test if AddressBinder defaults to IPv6Any only on invalid IP addresses.

However, DefaultsToIPv6AnyOnInvalidIPAddress is a good test to have so I'll added it.

cesarbs: I see. How about renaming this test to `DoesNotCreateIPEndPointOnInvalidIPAddress`?

natemcmaster: Yeah, that's better. Will update

natemcmaster:
Verified that address binding falls back to IPv4Any when IPv6Any is not available:

image

image

src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/AddressBinder.cs
natemcmaster: updated

halter73:
@natemcmaster The image you show with the warning on antares says "Now listening on: http://*localhost*:6000" which doesn't look like IPv6Any.

The image showing "Now listening on http://0.0.0.0:6000" when "http://`*`:6000" was specified doesn't show any warning.

Did you verify that Antares failed to bind when attempting to bind to http://[::]/ prior to this change?

natemcmaster:
We are currently inconsistent in what we put into IServerAddressesFeature. When we detect a ServerAddress with a hostname, we hit this:

if hostname == localhost  => Now listening on http://localhost:port
else => Now listening on http://ipaddress:port

IIRC this behavior is consistent with kestrel before this change. However, this is a change from 1.1 in which we always show "Now listening on http://hostname:port"

> Did you verify that Antares failed to bind when attempting to bind to http://[::]/ prior to this change?

Yes. It is still trying IPAddress.IPv6Any before IPAddress.Any.

Issue #1331, opened 2017-01-27
Resuming Connection on closed socket hides connection reset


cesarbs:
https://github.com/aspnet/KestrelHttpServer/blob/448a927a0ffdf8f59bbb2770e24757773d85cd8c/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Connection.cs#L302

davidfowl:
Fixed in https://github.com/aspnet/KestrelHttpServer/commit/a749939be4fc7a2bec620c35cdf816a844233870

Issue #941, opened 2016-06-27
Kestrel faills periodically:Error -107 ENOTCONN


agoretsky:
I've upgraded my app to rc2 and now my app periodically fails after some hours of running.
I am using ubuntu 15.04, mono 4.2.3.4 + asp.net core rc2 (https://www.microsoft.com/net/core#ubuntu)

`Unhandled Exception:
Microsoft.AspNetCore.Server.Kestrel.Networking.UvException: Error -107 ENOTCONN socket is not connected
at Microsoft.AspNetCore.Server.Kestrel.Networking.Libuv.Check (Int32 statusCode) <0x40abee50 + 0x0004f> in <filename unknown>:0
at Microsoft.AspNetCore.Server.Kestrel.Networking.Libuv.tcp_getpeername (Microsoft.AspNetCore.Server.Kestrel.Networking.UvTcpHandle handle, Microsoft.AspNetCore.Server.Kestrel.Networking.SockAddr& addr, System.Int32& namelen
at Microsoft.AspNetCore.Server.Kestrel.Networking.UvTcpHandle.GetPeerIPEndPoint () <0x40ad1940 + 0x0004b> in <filename unknown>:0
at Microsoft.AspNetCore.Server.Kestrel.Http.Connection.Start () <0x40ad0e00 + 0x0010b> in <filename unknown>:0
at Microsoft.AspNetCore.Server.Kestrel.Http.ListenerSecondary.ReadStartCallback (Microsoft.AspNetCore.Server.Kestrel.Networking.UvStreamHandle handle, Int32 status) <0x40ace1d0 + 0x003a3> in <filename unknown>:0
at Microsoft.AspNetCore.Server.Kestrel.Http.ListenerSecondary+<>c.<ConnectedCallback>b__13_1 (Microsoft.AspNetCore.Server.Kestrel.Networking.UvStreamHandle handle, Int32 status2, System.Object state) <0x40acdde0 + 0x0005b> i
at Microsoft.AspNetCore.Server.Kestrel.Networking.UvStreamHandle.UvReadCb (IntPtr handle, Int32 status, Microsoft.AspNetCore.Server.Kestrel.Networking.uv_buf_t& buf) <0x40acdbf0 + 0x0006a> in <filename unknown>:0
--- End of stack trace from previous location where exception was thrown ---
at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw () <0x40f33640 + 0x0002b> in <filename unknown>:0
[...]

PR #1726, opened 2017-04-20
Clean up libuv connection

davidfowl:
- Cancel all pending flushes on the input writer before
disposing the stream handle.
- Complete the pipe before disposing the socket
- Added logging for connection pause/resume.
- Added test

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Kestrel.Transport.Libuv/Internal/LibuvConnection.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel.Transport.Libuv/Internal/LibuvConnection.cs
@@ -4,11 +4,10 @@
 using System;
 using System.Diagnostics;
 using System.IO;
-using System.Threading;
 using System.Threading.Tasks;
-using Microsoft.AspNetCore.Server.Kestrel.Transport.Libuv.Internal.Networking;
 using Microsoft.AspNetCore.Server.Kestrel.Internal.System.IO.Pipelines;
 using Microsoft.AspNetCore.Server.Kestrel.Transport.Abstractions;
+using Microsoft.AspNetCore.Server.Kestrel.Transport.Libuv.Internal.Networking;
 using Microsoft.Extensions.Logging;
 
 namespace Microsoft.AspNetCore.Server.Kestrel.Transport.Libuv.Internal
@@ -207,7 +206,17 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Transport.Libuv.Internal
 
         private void StartReading()
         {
-            _socket.ReadStart(_allocCallback, _readCallback, this);
+            try
+            {
+                _socket.ReadStart(_allocCallback, _readCallback, this);
+            }
+            catch (UvException)
+            {
+                // ReadStart() can throw a UvException in some cases (e.g. socket is no longer connected).
+                // This should be treated the same as OnRead() seeing a "normalDone" condition.
+                Log.ConnectionReadFin(ConnectionId);
+                Input.Complete();
+            }
         }
     }
 }
--- a/src/Microsoft.AspNetCore.Server.Kestrel.Transport.Libuv/Internal/LibuvConnection.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel.Transport.Libuv/Internal/LibuvConnection.cs
@@ -53,7 +53,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Transport.Libuv.Internal
         private IConnectionHandler ConnectionHandler => ListenerContext.TransportContext.ConnectionHandler;
         private LibuvThread Thread => ListenerContext.Thread;
 
-        public async void Start()
+        public async Task Start()
         {
             try
             {
@@ -64,7 +64,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Transport.Libuv.Internal
                 Output = new LibuvOutputConsumer(_connectionContext.Output, Thread, _socket, ConnectionId, Log);
 
                 // Start socket prior to applying the ConnectionAdapter
-                _socket.ReadStart(_allocCallback, _readCallback, this);
+                StartReading();
 
                 try
                 {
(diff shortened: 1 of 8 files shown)


src/Microsoft.AspNetCore.Server.Kestrel.Transport.Libuv/Internal/LibuvConnection.cs
```diff
@@ -189,32 +200,14 @@ private static void ReadCallback(UvStreamHandle handle, int status, object state
...
-                }
-                catch (UvException)
-                {
-                    // ReadStart() can throw a UvException in some cases (e.g. socket is no longer connected).
-                    // This should be treated the same as OnRead() seeing a "normalDone" condition.
-                    Log.ConnectionReadFin(ConnectionId);
-                    Input.Complete();
-                }
-            }
+            _socket.ReadStart(_allocCallback, _readCallback, this);
```
davidfowl: I can add the try catch back but I'd like to understand if it's truly needed (and write a test for it maybe).

halter73: I suspect this could fail for the [same reason that uv_tcp_getpeername sometimes does on linux](https://github.com/aspnet/KestrelHttpServer/issues/941). This is normally caught and logged by the listener instead of crashing the server, but that wouldn't be the case for Resume().

Remember that before Resume() is called there is no read callback wired to tell us the underlying connection is no closed or invalid.

src/Microsoft.AspNetCore.Server.Kestrel.Transport.Libuv/Internal/LibuvConnection.cs
```diff
@@ -80,9 +80,17 @@ public LibuvConnection()
                 }
                 finally
                 {
-                    // Ensure the socket is disposed prior to completing in the input writer.
-                    _socket.Dispose();
+                    // Make sure it isn't possible for a paused read to resume reading after calling uv_close
+                    // on the stream handle
+                    Input.CancelPendingFlush();
```
halter73: Is there a Pipe test that verifies that the writer in fact observes cancellation when:

1. The writer continuation is scheduled after a call to IPipeReader.Advance.
2. Next CancelPendingFlush is called.
3. Lastly the _writerScheduler actually runs the continuation.

It looks like this is already the case, but it's important this never regresses.

test/Microsoft.AspNetCore.Server.Kestrel.Transport.Libuv.Tests/ConnectionTests.cs
```diff
@@ -16,38 +18,86 @@ public class ConnectionTests
...
+                Assert.False(readAwaitable.IsCompleted);
+            }
+            finally
+            {
+                await thread.StopAsync(TimeSpan.FromSeconds(1));
+            }
+        }
+
+        [Fact]
+        public async Task ConnectionDoesNotResumeAfterSocketCloseIfBackpressureIsApplied()
```
halter73: Does this test actually consume the "request" data in such a way that the old code would have even tried to resume? Looking at the test, it seems like it would pass in dev since (await flushTask).IsCompleted would already be true.

test/Microsoft.AspNetCore.Server.Kestrel.Transport.Libuv.Tests/ConnectionTests.cs
```diff
@@ -16,38 +18,86 @@ public class ConnectionTests
...
+        {
+            var mockConnectionHandler = new MockConnectionHandler();
+            mockConnectionHandler.InputOptions.MaximumSizeHigh = 3;
+            var mockLibuv = new MockLibuv();
+            var transportContext = new TestLibuvTransportContext() { ConnectionHandler = mockConnectionHandler };
+            var transport = new LibuvTransport(mockLibuv, transportContext, null);
+            var thread = new LibuvThread(transport);
+            // We don't set the output scheduler here since we want to run the callback inline
```
halter73: Nit: We don't set the output *writer* scheduler

src/Microsoft.AspNetCore.Server.Kestrel.Transport.Libuv/Internal/LibuvConnection.cs
pakrym: https://github.com/dotnet/corefxlab/pull/1489 ?

test/Microsoft.AspNetCore.Server.Kestrel.Transport.Libuv.Tests/ConnectionTests.cs
davidfowl: `(await flushTask).IsCompleted` is never true be cause we never complete the `Input.Reader`. Regardless, I'm attempting to add another test.

davidfowl:
🆙 📅

src/Microsoft.AspNetCore.Server.Kestrel.Transport.Libuv/Internal/LibuvConnection.cs
```diff
@@ -189,31 +199,23 @@ private static void ReadCallback(UvStreamHandle handle, int status, object state
...
-                    Input.Complete();
-                }
+                _socket.ReadStart(_allocCallback, _readCallback, this);
+            }
+            catch (UvException)
+            {
+                // ReadStart() can throw a UvException in some cases (e.g. socket is no longer connected).
+                // This should be treated the same as OnRead() seeing a "normalDone" condition.
+                Log.ConnectionReadFin(ConnectionId);
+                Input.Complete();
```
halter73: Ironically, this is the only place the input writer is completed without an error. You could fix #1331 right now by calling `Input.Complete(uvEx)` here.

src/Microsoft.AspNetCore.Server.Kestrel.Transport.Libuv/Internal/Listener.cs
```diff
@@ -132,7 +132,7 @@ private void OnConnection(UvStreamHandle listenSocket, int status)
         protected virtual void DispatchConnection(UvStreamHandle socket)
         {
             var connection = new LibuvConnection(this, socket);
-            connection.Start();
+            _ = connection.Start();
```
halter73: *Soo* much better than async void lol.

PR #1728, opened 2017-04-20
Attempt to make UnrootedConnectionsGetRemovedFromHeartbeat reliable

davidfowl:
- Attempt to make this test more reliable by using TryStartNoGCRegion and `MethodImplOptions.NoOptimization`


davidfowl:
kewl

Issue #1729, opened 2017-04-20
[Patch]: Flaky test on OSX Microsoft.AspNetCore.Server.Kestrel.FunctionalTests.AddressRegistrationTests.RegisterAddresses_IPv4_Success


pranavkm:
```C#
System.Net.Sockets.SocketException : Network is unreachable
   at Microsoft.AspNetCore.Testing.HttpClientSlim.<GetSocket>d__7.MoveNext() in /Users/asplab/TeamCity/work/d6337ac2abf63a4b/.repositories/KestrelHttpServer/test/shared/HttpClientSlim.cs:line 114
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at Microsoft.AspNetCore.Testing.HttpClientSlim.<GetStream>d__6.MoveNext() in /Users/asplab/TeamCity/work/d6337ac2abf63a4b/.repositories/KestrelHttpServer/test/shared/HttpClientSlim.cs:line 85
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at Microsoft.AspNetCore.Testing.HttpClientSlim.<GetStringAsync>d__1.MoveNext() in /Users/asplab/TeamCity/work/d6337ac2abf63a4b/.repositories/KestrelHttpServer/test/shared/HttpClientSlim.cs:line 24
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at Microsoft.AspNetCore.Server.Kestrel.FunctionalTests.AddressRegistrationTests.<RegisterAddresses_Success>d__7.MoveNext() in /Users/asplab/TeamCity/work/d6337ac2abf63a4b/.repositories/KestrelHttpServer/test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/AddressRegistrationTests.cs:line 93
--- End of stack trace from previous location where exception was thrown ---
[...]

PR #1731, opened 2017-04-20
Skip potentially flaky tests in address registration tests

natemcmaster:
Skip a potentially flaky test in the patch build. The test is still in dev branch. If it continues to be a problem in dev, then we'll look for a better solution.

Close https://github.com/aspnet/KestrelHttpServer/issues/1729

### DIFF
--- a/test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/AddressRegistrationTests.cs
+++ b/test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/AddressRegistrationTests.cs
@@ -226,7 +226,6 @@ namespace Microsoft.AspNetCore.Server.Kestrel.FunctionalTests
 
                 // Dynamic port and non-loopback addresses
                 dataset.Add("http://127.0.0.1:0/;https://127.0.0.1:0/", GetTestUrls);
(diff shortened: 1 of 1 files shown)


natemcmaster:
How would we detect that? We don't bind to hostnames. We rely on DNS settings to redirect requests to the machines IPv6Any endpoint

halter73:
Bind to IPv4Any using a managed socket, then try to connect to the given host name from the client.

natemcmaster:
> Bind to IPv4Any using a managed socket, then try to connect to the given host name from the client.

This assumes too much about how DNS is configured. What we're actually trying to test here is binding to IPv4Any and dynamic ports, not that hostname resolution is working.

halter73:
It doesn't assume anything. That's why we test that it could work using a managed socket.

natemcmaster:
Maybe we're talking about different things...but the socket we open in HttpClientSlim to test this still attempts DNS resolution. The DNS resolution is the flaky part, and what I'm trying to remove from our tests. See #1729 and #1734

natemcmaster:
I'll backport https://github.com/aspnet/KestrelHttpServer/pull/1751 instead of doing this.

PR #1736, opened 2017-04-21
Add cancellation support for write async back

pakrym:


test/Microsoft.AspNetCore.Server.Kestrel.Transport.Libuv.Tests/LibuvOutputConsumerTests.cs
```diff
@@ -382,16 +382,161 @@ public void Dispose()
                 Assert.False(task4Success.IsFaulted);
 
                 // Third task is now canceled
-                // TODO: Cancellation isn't supported right now
-                // await Assert.ThrowsAsync<TaskCanceledException>(() => task3Canceled);
-                // Assert.True(task3Canceled.IsCanceled);
```
halter73: You should also be able to remove `var task4Success = socketOutput.WriteAsync(fullBuffer, default(CancellationToken));` since you don't advance any number of bytes on a write failure any more.

src/Microsoft.AspNetCore.Server.Kestrel.Transport.Libuv/Internal/LibuvOutputConsumer.cs
```diff
@@ -81,7 +83,7 @@ public class LibuvOutputConsumer
                 }
                 finally
                 {
-                    _pipe.Advance(result.Buffer.End);
+                    _pipe.Advance(consumed);
```
davidfowl: Why?

src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/Http/OutputProducer.cs
```diff
@@ -112,8 +113,12 @@ private Task FlushAsyncAwaited(WritableBufferAwaitable awaitable)
                     });
                 }
             }
+            await _flushTcs.Task;
 
-            return _flushTcs.Task;
+            if (cancellationToken.IsCancellationRequested)
```
davidfowl: Why don't you just cancel the `_flushTcs` instead of making this method async?

pakrym: If you run WriteAsync without CT, and on with CT I didn't want to cancel first task

davidfowl: I see, this is how you fixed the problem of "everyone having the same task". The CT is captured in the state machine and you check it on completion. Clever 👍

src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/Http/OutputProducer.cs
```diff
@@ -112,8 +113,12 @@ private Task FlushAsyncAwaited(WritableBufferAwaitable awaitable)
...
+            {
+                throw new TaskCanceledException();
```
davidfowl: Just call `cancellationToken.ThrowIfCancellationRequested()` afterwards. Maybe this should be:

```C#
if (cancellationToken.CanBeCancelled)
{
    cancellationToken.ThrowIfCancellationRequested();
}
```

pakrym: `ThrowIfCancellationRequested` would return faulted task, `TaskCanceledException` would return cancelled one.

src/Microsoft.AspNetCore.Server.Kestrel.Transport.Libuv/Internal/LibuvOutputConsumer.cs
pakrym: When error happens during execution of a write request abort gets _posted_ and if `Advance` relieves back pressure there is a race happening between flush completion and cancellation token callback. So not releasing back pressure in case of an error prevents the race.

src/Microsoft.AspNetCore.Server.Kestrel.Transport.Libuv/Internal/LibuvOutputConsumer.cs
```diff
@@ -56,6 +57,7 @@ public class LibuvOutputConsumer
 
                             if (writeResult.Error != null)
                             {
+                                consumed = result.Buffer.Start;
```
davidfowl: buffer.Start

src/Microsoft.AspNetCore.Server.Kestrel.Transport.Libuv/Internal/LibuvOutputConsumer.cs
```diff
@@ -41,6 +41,7 @@ public class LibuvOutputConsumer
             {
                 var result = await _pipe.ReadAsync();
                 var buffer = result.Buffer;
+                var consumed = result.Buffer.End;
```
halter73: `buffer.End`

PR #1738, opened 2017-04-22
Remove unused package references

pakrym:


src/Microsoft.AspNetCore.Server.Kestrel/Microsoft.AspNetCore.Server.Kestrel.csproj
```diff
@@ -12,9 +12,9 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.AspNetCore.Hosting" Version="$(AspNetCoreVersion)" />
+    <PackageReference Include="Microsoft.AspNetCore.Hosting.Abstractions" Version="$(AspNetCoreVersion)" />
```
davidfowl: This is done for compatibility. Undo....

pakrym: Oh, ok. Compatibility with what?

davidfowl: https://www.nuget.org/packages/Microsoft.AspNetCore.Server.Kestrel/1.1.1. The idea was that if you include Kestrel, you get hosting for free. We also did this with WebListener. It *might* be ok to drop this now that we have the meta package but we left it in to reduce the chance of a breaking change (since this was intentional).

/cc @Tratcher @DamianEdwards

src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/AddressBinder.cs
```diff
@@ -174,7 +173,7 @@ private static Task BindEndpointAsync(IPEndPoint endpoint, AddressBindContext co
 
             if (!string.IsNullOrEmpty(parsedAddress.PathBase))
             {
-                throw new InvalidOperationException($"A path base can only be configured using {nameof(IApplicationBuilder)}.UsePathBase().");
+                throw new InvalidOperationException($"A path base can only be configured using IApplicationBuilder.UsePathBase().");
```
cesarbs: Why? You're still pulling in `Microsoft.AspNetCore.Http.Abstractions`.

pakrym:
Stale. Closing.

Issue #1621, opened 2017-04-07
Flaky Test: Functional tests on Linux "disposing listeners failed"


JunTaoLuo:
Seeing this as part of our partner acquisition linux builds but I suspect they are flaky on all platforms:

Microsoft.AspNetCore.Server.Kestrel.FunctionalTests.HttpsTests.EmptyRequestLoggedAsInformation
```
Disposing listeners failed
Expected: True
Actual:   False
   at Microsoft.AspNetCore.Server.Kestrel.FunctionalTests.HttpsTests.<EmptyRequestLoggedAsInformation>d__0.MoveNext() in /datadrive/169923259d6dcfec/Universe/.r/KestrelHttpServer/test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/HttpsTests.cs:line 56
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
```
Microsoft.AspNetCore.Server.Kestrel.FunctionalTests.HttpsTests.EmptyRequestLoggedAsInformation
```
Disposing listeners failed
Expected: True
Actual:   False
   at Microsoft.AspNetCore.Server.Kestrel.FunctionalTests.HttpsTests.<EmptyRequestLoggedAsInformation>d__0.MoveNext() in /datadrive/169923259d6dcfec/Universe/.r/KestrelHttpServer/test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/HttpsTests.cs:line 56
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
[...]

PR #1740, opened 2017-04-22
Attempt to make test more reliable

davidfowl:
- Sometimes this test fails with a "disposing listeners failed".
It can just take a longer time for the PostAsync to complete it seems.
As a result, we ignore that error in this test. Alternatively,
we could consider removing that log or consider not making it an error.
- I haven't been able to find a root cause

#1621


davidfowl:
The fact that this test and `LoggingConnectionAdapterCanBeAddedBeforeAndAfterHttpsAdapter` keep failing due to "flakyness" makes me wonder if there's some issue with connection adapters.

davidfowl:
Seems to be passing now

PR #1741, opened 2017-04-22
Schedule connection adapter reads on configured thread pool

davidfowl:
- This should speed up connection adapters as we don't block
new reads. It *might* also help some of the test flakiness. This is more
similar to what we did before with `SocketInput` (https://github.com/aspnet/KestrelHttpServer/blob/rel/1.1.1/src/Microsoft.AspNetCore.Server.Kestrel/Filter/Internal/FilteredStreamAdapter.cs#L58) (Socket Input always used the thread pool)

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/FrameConnection.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/FrameConnection.cs
@@ -49,7 +49,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal
         // Internal for testing
         internal PipeOptions AdaptedInputPipeOptions => new PipeOptions
         {
-            ReaderScheduler = InlineScheduler.Default,
+            ReaderScheduler = _context.ServiceContext.ThreadPool,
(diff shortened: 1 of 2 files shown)


src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/FrameConnection.cs
```diff
@@ -49,7 +49,7 @@ public FrameConnection(FrameConnectionContext context)
         // Internal for testing
         internal PipeOptions AdaptedInputPipeOptions => new PipeOptions
         {
-            ReaderScheduler = InlineScheduler.Default,
+            ReaderScheduler = _context.ServiceContext.ThreadPool,
```
halter73: Does this makes sense for the HTTPS adapter? Wouldn't that put unnecessary computation on the IO thread? The call to uv_write would be delayed/posted anyway.

davidfowl: Wrong scheduler. This is the reader scheduler for input.

PR #1743, opened 2017-04-23
More clean up of LibuvConnection

davidfowl:
- Use C# 7
- Use Buffer<T>.Pin to get access to the underlying
pointer instead of using TryGetPointer.

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Kestrel.Transport.Libuv/Internal/LibuvConnection.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel.Transport.Libuv/Internal/LibuvConnection.cs
@@ -5,6 +5,7 @@ using System;
 using System.Diagnostics;
 using System.IO;
 using System.Threading.Tasks;
+using Microsoft.AspNetCore.Server.Kestrel.Internal.System.Buffers;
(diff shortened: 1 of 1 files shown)


src/Microsoft.AspNetCore.Server.Kestrel.Transport.Libuv/Internal/LibuvConnection.cs
```diff
@@ -107,16 +108,12 @@ private static LibuvFunctions.uv_buf_t AllocCallback(UvStreamHandle handle, int
 
         private unsafe LibuvFunctions.uv_buf_t OnAlloc(UvStreamHandle handle, int suggestedSize)
         {
-            Debug.Assert(_currentWritableBuffer == null);
```
cesarbs: Why did you remove the `Assert`?

davidfowl: Didn't seem necessary. @pakrym originally added it when we ported.

davidfowl: Added it back. It can't hurt.

PR #1744, opened 2017-04-23
Clean up LibuvOutputConsumer

davidfowl:
- Added UvShutdownReq.ShutdownAsync
- Added Debug.Assert in LibuvAwaitable since it should never race.

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Kestrel.Transport.Libuv/Internal/LibuvAwaitable.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel.Transport.Libuv/Internal/LibuvAwaitable.cs
@@ -2,9 +2,9 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using System;
+using System.Diagnostics;
 using System.Runtime.CompilerServices;
 using System.Threading;
-using System.Threading.Tasks;
 using Microsoft.AspNetCore.Server.Kestrel.Transport.Libuv.Internal.Networking;
 
 namespace Microsoft.AspNetCore.Server.Kestrel.Transport.Libuv.Internal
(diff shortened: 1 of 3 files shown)


src/Microsoft.AspNetCore.Server.Kestrel.Transport.Libuv/Internal/LibuvAwaitable.cs
```diff
@@ -49,10 +49,16 @@ public UvWriteResult GetResult()
...
         public void OnCompleted(Action continuation)
         {
+            // There should never be a race between IsCompleted and OnCompleted since both operations
+            // should always be on the libuv thread
+
             if (_callback == _callbackCompleted ||
                 Interlocked.CompareExchange(ref _callback, continuation, null) == _callbackCompleted)
             {
-                Task.Run(continuation);
+                Debug.Fail($"{typeof(LibuvAwaitable<TRequest>)}.OnComplated raced with IsCompleted, running callback inline.");
```
cesarbs: typo: OnCompl**a**ted

src/Microsoft.AspNetCore.Server.Kestrel.Transport.Libuv/Internal/LibuvAwaitable.cs
```diff
@@ -49,10 +49,16 @@ public UvWriteResult GetResult()
 
         public void OnCompleted(Action continuation)
         {
+            // There should never be a race between IsCompleted and OnCompleted since both operations
+            // should always be on the libuv thread
```
pakrym: Add ability to assert libuv thread and use it extensively ?

davidfowl: I had a commit that passed the thread in but that complicated the code. This should be good enough for most cases. I also considered getting rid of the Interlocked.* operations since everything is thread safe.

davidfowl: That plus we already assert before every operation that we're on the libuv thread. https://github.com/aspnet/KestrelHttpServer/blob/42d82a507da9c7522ae64c1a4c5cdd6556035162/src/Microsoft.AspNetCore.Server.Kestrel.Transport.Libuv/Internal/Networking/UvMemory.cs#L84

PR #1746, opened 2017-04-24
Remove async from OnRead

davidfowl:
- Remove async from fast path.
- Added ApplyBackpressureAsync that pauses and resumes reading.


src/Microsoft.AspNetCore.Server.Kestrel.Transport.Libuv/Internal/LibuvConnection.cs
```diff
@@ -175,27 +175,34 @@ private static void ReadCallback(UvStreamHandle handle, int status, object state
...
             }
+            else if (flushTask?.IsCompleted == false)
+            {
+                // We wrote too many bytes too the reader so pause reading and resume when
+                // we hit the low water mark
+                _ = ApplyBackpressureAsync(flushTask.Value);
+            }
+        }
+
+        private async Task ApplyBackpressureAsync(WritableBufferAwaitable flushTask)
```
halter73: nit: since this also releases backpressure maybe name this RegulateBackpressureAsync

davidfowl:
> Do you expect this to improve perf?

Unlikely.

PR #1747, opened 2017-04-24
Capture the write error for logging purposes

davidfowl:
Store the write error and log it for writes that happen after failure.


davidfowl:
> What do you think about logging only the first disconnected write, and logging any write errors as they occur?

I think we shouldn't bother logging this at all really.

src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/Http/OutputProducer.cs
```diff
@@ -124,6 +121,7 @@ private void AbortIfNeeded(WritableBufferAwaitable awaitable)
             }
             catch (Exception ex)
             {
+                // Store the write exception for logging purposes
```
halter73: Remove comment

PR #1748, opened 2017-04-24
Switching to the new toolset compiler.

VSadov:
The new version is
https://dotnet.myget.org/feed/roslyn/package/nuget/Microsoft.Net.Compilers/2.3.0-rdonly-ref-61625-05

The new version enforces the following rules:
- `Span<T>` and `ReadonlySpan<T>` are stack-only types. (no boxing, invalid use in async, etc...)
- `Span<T>` and `ReadonlySpan<T>` cannot be passed via a writeable parameter.

There were, however, no violations of the rules. (I had to introduce and then remove some violatins, just to see that detection works)

NOTE: we are not yet validating the usage of span-containing structs. There could be more indrect violations of stack safety through that.

### DIFF
--- a/NuGet.config
+++ b/NuGet.config
@@ -4,5 +4,6 @@
     <add key="AspNetCore" value="https://dotnet.myget.org/F/aspnetcore-ci-dev/api/v3/index.json" />
     <add key="AspNetCoreTools" value="https://dotnet.myget.org/F/aspnetcore-tools/api/v3/index.json" />
     <add key="NuGet" value="https://api.nuget.org/v3/index.json" />
+    <add key="roslyn" value="https://dotnet.myget.org/F/roslyn/api/v3/index.json" />
   </packageSources>
 </configuration>
--- a/build/common.props
+++ b/build/common.props
@@ -20,4 +20,7 @@
     <PackageReference Include="NETStandard.Library" Version="$(BundledNETStandardPackageVersion)" />
   </ItemGroup>
 
+  <ItemGroup>
+    <PackageReference Include="Microsoft.NETCore.Compilers" Version="$(RoslynVersion)" PrivateAssets="All" />
+  </ItemGroup>
 </Project>
--- a/build/dependencies.props
+++ b/build/dependencies.props
@@ -5,6 +5,7 @@
     <CoreFxLabsVersion>0.1.0-*</CoreFxLabsVersion>
     <CoreFxVersion>4.3.0</CoreFxVersion>
     <InternalAspNetCoreSdkVersion>2.1.0-*</InternalAspNetCoreSdkVersion>
+    <RoslynVersion>2.3.0-rdonly-ref-61625-05</RoslynVersion>
     <LibUvVersion>1.10.0-*</LibUvVersion>
     <JsonNetVersion>9.0.1</JsonNetVersion>
     <MoqVersion>4.7.1</MoqVersion>
--- a/test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/SystemdActivation/docker.sh
+++ b/test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/SystemdActivation/docker.sh
@@ -1,5 +1,8 @@
 #!/usr/bin/env bash
 
+# adding dotnet to the path. it is needed to run toolset csc.
+export PATH=$HOME/.dotnet:$PATH
+
 set -e
 
 scriptDir=$(dirname "${BASH_SOURCE[0]}")


davidfowl:
Still broken.

agocke:
Looks like the permissions still aren't properly set. Maybe InitialTargets is an insufficient dependency? I'll also add a BeforeTargets requirement.

VSadov:
@davidfowl somehow the toolset compiler does not want to run on Linux. The theory is that it the +x bit is not set on the binary.

agocke:
Alright, I've tried this locally a half dozen times and it seems like a BeforeTargets constraint works. We'll see if it works on the build machine too.

VSadov:
Win and Mac passed, but not Linux - no idea if that is the issue with the toolset or something else

VSadov:
/home/travis/.nuget/packages/microsoft.netcore.compilers/2.3.0-rdonly-ref-61625-05/tools/Microsoft.CSharp.Core.targets(71,5): error MSB6006: "RunCsc" exited with code 127. 

It means that some binary was not found - it is not there and is not on the PATH.
(found, but not executable would be 126, i think)

VSadov:
the command does:

`dotnet $THISDIR/csc.exe "$@"`

Is `dotnet` on the path?

agocke:
@davidfowl Is there any way to get the full msbuild log out of Travis? I tried this locally and it seems to work fine on my machine.

natemcmaster:
@agocke The main build.sh script passed just fine. The failure came from the systemd tests that run in Docker. This test is launched with `test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/SystemdActivation/docker.sh`. 
 cc @halter73 - ideas on how to debug this test?

halter73:
Someone will have to run docker.sh with a user that has permissions to create an run containers.

@natemcmaster Do you have any idea what the following output from the last Ubuntu travis run means?

```
/home/travis/.nuget/packages/microsoft.netcore.compilers/2.3.0-rdonly-ref-61625-05/tools/Microsoft.CSharp.Core.targets(71,5): error MSB6006: "RunCsc" exited with code 127. [/home/travis/build/aspnet/KestrelHttpServer/src/Microsoft.AspNetCore.Server.Kestrel.Transport.Abstractions/Microsoft.AspNetCore.Server.Kestrel.Transport.Abstractions.csproj]
/home/travis/.dotnet/sdk/2.0.0-preview1-005825/Sdks/Microsoft.NET.Sdk/build/Microsoft.NET.Sdk.targets(93,5): error : Cannot find project info for '/home/travis/build/aspnet/KestrelHttpServer/src/Microsoft.AspNetCore.Server.Kestrel.Transport.Abstractions/Microsoft.AspNetCore.Server.Kestrel.Transport.Abstractions.csproj'. This can indicate a missing project reference. [/home/travis/build/aspnet/KestrelHttpServer/src/Microsoft.AspNetCore.Server.Kestrel.Core/Microsoft.AspNetCore.Server.Kestrel.Core.csproj]
/home/travis/.dotnet/sdk/2.0.0-preview1-005825/Sdks/Microsoft.NET.Sdk/build/Microsoft.NET.Sdk.targets(93,5): error : Cannot find project info for '/home/travis/build/aspnet/KestrelHttpServer/src/Microsoft.AspNetCore.Server.Kestrel.Transport.Abstractions/Microsoft.AspNetCore.Server.Kestrel.Transport.Abstractions.csproj'. This can indicate a missing project reference. [/home/travis/build/aspnet/KestrelHttpServer/src/Microsoft.AspNetCore.Server.Kestrel.Transport.Sockets/Microsoft.AspNetCore.Server.Kestrel.Transport.Sockets.csproj]
/home/travis/.dotnet/sdk/2.0.0-preview1-005825/Sdks/Microsoft.NET.Sdk/build/Microsoft.NET.Sdk.targets(93,5): error : Cannot find project info for '/home/travis/build/aspnet/KestrelHttpServer/src/Microsoft.AspNetCore.Server.Kestrel.Core/Microsoft.AspNetCore.Server.Kestrel.Core.csproj'. This can indicate a missing project reference. [/home/travis/build/aspnet/KestrelHttpServer/src/Microsoft.AspNetCore.Server.Kestrel/Microsoft.AspNetCore.Server.Kestrel.csproj]
/home/travis/.dotnet/sdk/2.0.0-preview1-005825/Sdks/Microsoft.NET.Sdk/build/Microsoft.NET.Sdk.targets(93,5): error : Cannot find project info for '/home/travis/build/aspnet/KestrelHttpServer/src/Microsoft.AspNetCore.Server.Kestrel.Core/Microsoft.AspNetCore.Server.Kestrel.Core.csproj'. This can indicate a missing project reference. [/home/travis/build/aspnet/KestrelHttpServer/src/Microsoft.AspNetCore.Server.Kestrel.Https/Microsoft.AspNetCore.Server.Kestrel.Https.csproj]
/home/travis/.dotnet/sdk/2.0.0-preview1-005825/Sdks/Microsoft.NET.Sdk/build/Microsoft.NET.Sdk.targets(93,5): error : Cannot find project info for '/home/travis/build/aspnet/KestrelHttpServer/src/Microsoft.AspNetCore.Server.Kestrel/Microsoft.AspNetCore.Server.Kestrel.csproj'. This can indicate a missing project reference. [/home/travis/build/aspnet/KestrelHttpServer/samples/SampleApp/SampleApp.csproj]
```

natemcmaster:
The first error indicates the the compiler crashed trying to compile icrosoft.AspNetCore.Server.Kestrel.Transport.Abstractions. The 5 errors after are a downstream results of that failure.

halter73:
Do you think that build failure is from the "~/.dotnet/dotnet publish -f netcoreapp2.0 ./samples/SampleApp/" line in docker.sh? That would explain it. But why would the compiler crash?

natemcmaster:
Probably. Try adding `/v:detailed` to the dotnet-publish call in docker.sh. This will show which step in  Microsoft.Csharp.Core.targets failed.

VSadov:
Compiler should compile that project just fine. It does on WinNT and Mac. And as I understand everything is fine outside of the docker too.

Most likely there is a problem with environment - files are missing, do not have permissions, something like that.

VSadov:
RunCsc   is not the compier itself, it is a script that does:

```
#!/bin/sh

THISDIR=$(dirname $0)

dotnet $THISDIR/csc.exe "$@"
```

natemcmaster:
Might be a PATH issue? RunCsc assumes dotnet is on the path _and_ that the one on PATH is the right dotnet executable to use.

VSadov:
@natemcmaster    not having dotnet on the path could result in this failure. 
The question is then, how anything else works and how can we .run csc.

natemcmaster:
build.sh adds $HOME/.dotnet to PATH. docker.sh does not.

agocke:
I think @natemcmaster hit on the problem. We should add the same code from build.sh to docker.sh

VSadov:
@natemcmaster   can that be added?  We need to run a managed binary - "csc.exe", that needs a clr host - "dotnet" in this case.

natemcmaster:
Sure. Just add `export PATH=$HOME/.dotnet:$PATH` to the top of the docker.sh file in your branch.

VSadov:
@natemcmaster will do. Thanks!!

natemcmaster:
🎉 it passed on all CI runs :)

Btw, that might be a good thing to fix in RunCsc.{sh, cmd}. I didn't see RunCsc.cmd anywhere in dotnet/roslyn. Where does this file come from?

VSadov:
Woohoo!!

agocke:
@natemcmaster This is part of the tooslet NuGet package. Eventually I want to build a mechanism to plug in the proper host through the MSBuild task, but that's going to be in a later build.

VSadov:
RunCsc.sh   is a way to invoke custom csc  form an msbuild task. It is only needed on Unix.

The right solution is likely introduce and propagat "ClrHost" variable through msbuild. That would allow running on mono, if needed or on a particular "dotnet", not necessarily on the path. But that is a more involved solution.

VSadov:
Who can sign off for this PR?

Yes, i will squash the commits :)

natemcmaster:
> The right solution is likely introduce and propagat "ClrHost" variable through msbuild.

🤞 they fix https://github.com/Microsoft/msbuild/issues/1669 soon. We've had the similar issues with PATH and dotnet in some aspnet targets.

NuGet.config
```diff
@@ -4,5 +4,6 @@
     <add key="AspNetCore" value="https://dotnet.myget.org/F/aspnetcore-ci-dev/api/v3/index.json" />
     <add key="AspNetCoreTools" value="https://dotnet.myget.org/F/aspnetcore-tools/api/v3/index.json" />
     <add key="NuGet" value="https://api.nuget.org/v3/index.json" />
+    <add key="roslyn" value="https://dotnet.myget.org/F/roslyn/" />
```
natemcmaster: One minor nit: can you add api/v3/index.json to the end of the URL for consistency with our other feeds?

VSadov: done

natemcmaster:
Yes. @VSadov @agocke when do we expect the changes in 2.3.0-rdonly-ref-61625-05 to be available in a released version of roslyn?

VSadov:
@natemcmaster  The features are planned for C#7.2
can't be sure about actual ship dates. The actual mechanics around shipping/building Kestrel are being worked out, as I understand.

davidfowl:
Let me make all of the powers that be aware of this change before merging. Assigning to myself. We also need to test this in Visual Studio to make sure our regular workflow isn't disrupted.

@VSadov is there a public VSIX that matches the branch?

VSadov:
@davidfowl   note this change only brings stack-only rules and only for Span/ReadonlySpan, not yet for generalized span-like structs. 
The main goal here was to actually enable an alternative toolset compiler

We are making VSIXes. The internal pass is:
\\cpvsbuild\drops\Roslyn\Roslyn-Signed\readonly-ref\Release\20170425.5\Vsix\Roslyn.Deployment.Full.Next

We are likely dropping it to MyGet too, just need to figure where.

VSadov:
CC: @jaredpar

VSadov:
The  MyGet location for the VSIX is 
https://dotnet.myget.org/feed/roslyn/package/vsix/0b48e25b-9903-4d8b-ad39-d4cca196e3c7/2.3.0.6162505

VSadov:
close/reopen just to see it is still passing tests.

VSadov:
close/reopen just to see it is still passing tests.

VSadov:
close/reopen just to see it is still passing tests.

ahsonkhan:
Question: Do you install Microsoft.NETCore.App version 1.1.0 during all your CI runs?

I see the following in [Console logs](https://ci.appveyor.com/project/aspnetci/kestrelhttpserver/build/1.0.5307):
`dotnet-install: Installation finished
dotnet-install: Downloading https://dotnetcli.azureedge.net/dotnet/release/1.1.0/Binaries/1.1.1/dotnet-win-x64.1.1.1.zip
dotnet-install: Extracting zip from https://dotnetcli.azureedge.net/dotnet/release/1.1.0/Binaries/1.1.1/dotnet-win-x64.1.1.1.zip
dotnet-install: Adding to current process PATH: "C:\Users\appveyor\.dotnet\x64\". Note: This change will not be visible if PowerShell was run as a child process.`

We don't do this in corefxlab: https://github.com/dotnet/corefxlab/issues/1536

VSadov:
close/reopen just to see it is still passing tests.

natemcmaster:
We may not want to use it on production yet, but we can make this run automatically so you don't have to keep opening/closing the PR to test it.

Here's how I think we could do that:

```xml
<PackageReference Include="Microsoft.NETCore.Compilers" Condition="'$(Experimental)' == 'true'" />
```

```yml
# .travis.yml
matrix:
  fast_finish: true
env:
  matrix:
  - Experimental: false
  - Experimental: true
```

```yml
# appveyor.yml
matrix:
  fast_finish: true
environment:
  matrix:
  - Experimental: false
  - Experimental: true
```

natemcmaster:
> Do you install Microsoft.NETCore.App version 1.1.0 during all your CI runs?

@ahsonkhan yes

VSadov:
@natemcmaster - I think @ahsonkhan  now installs 1.1.0 as well. CoreFexLab has switched to this package and runs CI tests on Linux as well.

VSadov:
@natemcmaster - I hear that this PR will be merged "real soon", so I am not sure if the parallel run is really necessary.

build/common.props
```diff
@@ -20,4 +20,7 @@
     <PackageReference Include="NETStandard.Library" Version="$(BundledNETStandardPackageVersion)" />
   </ItemGroup>
 
+  <ItemGroup>
+    <PackageReference Include="Microsoft.NETCore.Compilers" Version="$(RoslynVersion)" PrivateAssets="All" />
```
davidfowl: We'd like to use the new compiler for single repo builds but not for coherence-signed builds.

/cc @pranavkm @natemcmaster

pranavkm: We build each repo as a separate process, so this should be fine.

VSadov: This should be doable for now, but become hard to do eventually. 

Right now the only difference is that the new compiler has extra checks. However at some point you will need to use 7.2 features such as `ref struct` . Then the same source will not be buildable by both compilers unless you use `#if`

VSadov:
This compiler is very old now.

Issue #1662, opened 2017-04-12
Allow the transport suggest an application scheduler


tmds:
Request to support a no dispatch mode with the libuv transport.

cfr https://github.com/aspnet/KestrelHttpServer/issues/1390#issuecomment-293384383

tmds:
https://github.com/aspnet/KestrelHttpServer/pull/1695 implements part of this.
The other part is to no longer make the ListenerPrimary handle requests but only do dispatching. This is to avoid accepting new connections gets blocked by request handling.

davidfowl:
Kestrel.Core has the mode exposed on its options. If the mode is set to default then the transport gets to pick the scheduler. This lets the transport pick where application callbacks run. By default, our transports will dispatch application code to the thread pool when appropriate. For example, on linux, the sockets transports already dispatches so there's no need to do it again. On windows however, callbacks can run on the IO thread, so it would dispatch by default.

If the user always wants to dispatch despite what the transports pick for the application scheduler, then it can override by setting the SchedulingMode to TheadPool. If the user wants to run on whatever thread they happen to be on basically disabling all dispatching then they can choose Inline.

```C#
public interface IConnectionInformation
{
    IScheduler ApplicationScheduler { get; }
}

public class KestrelServerOptions
{
    public SchedulingMode ApplicationSchedulingMode { get; set; }
}

public enum SchedulingMode
{
    // Use the transport provided ApplicationScheduler 
    Default,
    // Use the thread pool
    ThreadPool,
    // Run on the "current" thread
    Inline
}
```

One issue is that using the `IScheduler` type doesn't have an overload that accepts a state object. This means we'd need to either change that or we decide to move `IThreadPool` to Transport.Abstractions. I'd rather change the `IScheduler` interface.

/cc @pakrym

davidfowl:
Filed https://github.com/dotnet/corefxlab/issues/1495

tmds:
This is specifically for Libuv. When dispatching inline, it is safer to no longer handle requests on the ListenerPrimary. This is to avoid blocking the accept Thread. [`base.DispatchConnection`](https://github.com/aspnet/KestrelHttpServer/blob/42d82a507da9c7522ae64c1a4c5cdd6556035162/src/Microsoft.AspNetCore.Server.Kestrel.Transport.Libuv/Internal/ListenerPrimary.cs#L101) should not called.

muratg:
[...]

PR #1759, opened 2017-04-25
Added ApplicationSchedulingMode to KestrelServerOptions

### LINKED ISSUE #1750: Add an enum for scheduling mode and expose it in KestrelServerOptions
We'll look into https://github.com/aspnet/KestrelHttpServer/issues/1662 later.

davidfowl:
Will remove `UseTransportThread` in another PR after I update http://github.com/aspnet/benchmarks.

Fixes #1750

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Kestrel.Core/KestrelServerOptions.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel.Core/KestrelServerOptions.cs
@@ -37,8 +37,8 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core
         /// <summary>
         /// Gets or sets a value that determines how Kestrel should schedule user callbacks.
         /// </summary>
-        /// <remarks>The default mode is <see cref="SchedulingMode.ThreadPool"/></remarks>
-        public SchedulingMode ApplicationSchedulingMode { get; set; } = SchedulingMode.ThreadPool;
+        /// <remarks>The default mode is <see cref="SchedulingMode.Default"/></remarks>
+        public SchedulingMode ApplicationSchedulingMode { get; set; } = SchedulingMode.Default;
 
         /// <summary>
         /// Enables the Listen options callback to resolve and use services registered by the application during startup.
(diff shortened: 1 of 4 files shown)


src/Microsoft.AspNetCore.Server.Kestrel.Transport.Abstractions/SchedulingMode.cs
```diff
@@ -0,0 +1,9 @@
+﻿namespace Microsoft.AspNetCore.Server.Kestrel.Transport.Abstractions
```
muratg: nit: Copyright  (simulating Troy)

src/Microsoft.AspNetCore.Server.Kestrel.Transport.Abstractions/SchedulingMode.cs
```diff
@@ -0,0 +1,9 @@
...
+{
+    public enum SchedulingMode
+    {
+        Default,
```
muratg: q: Why do we need default? Can't we do with the other two? Seems like default is ambiguous.

davidfowl: See https://github.com/aspnet/KestrelHttpServer/issues/1662#issuecomment-296498429

src/Microsoft.AspNetCore.Server.Kestrel.Core/KestrelServerOptions.cs
```diff
@@ -34,6 +35,12 @@ public class KestrelServerOptions
         public bool UseTransportThread { get; set; }
 
         /// <summary>
+        /// Gets or sets a value that determines how Kestrel should schedule user callbacks.
+        /// </summary>
+        /// <remarks>The default mode is <see cref="SchedulingMode.ThreadPool"/></remarks>
+        public SchedulingMode ApplicationSchedulingMode { get; set; } = SchedulingMode.ThreadPool;
```
halter73: Shouldn't this be set to `SchedulingMode.Default`?

davidfowl: Yea I'll just do that since I made default and thread pool the same in this PR. Originally I was going to make Default throw

PR #1760, opened 2017-04-25
More shutdown tweaks

davidfowl:
- Added assert if loop has ended before starting the shutdown
sequence.

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Kestrel.Transport.Libuv/Internal/LibuvThread.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel.Transport.Libuv/Internal/LibuvThread.cs
@@ -33,6 +33,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Transport.Libuv.Internal
         private Queue<CloseHandle> _closeHandleRunning = new Queue<CloseHandle>(256);
         private readonly object _workSync = new object();
         private readonly object _startSync = new object();
+        private bool _stopping = false;
         private bool _stopImmediate = false;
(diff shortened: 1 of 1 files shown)


src/Microsoft.AspNetCore.Server.Kestrel.Transport.Libuv/Internal/LibuvThread.cs
```diff
@@ -302,7 +300,7 @@ private void OnPost()
                 wasWork = DoPostWork();
                 wasWork = DoPostCloseHandle() || wasWork;
                 loopsRemaining--;
-            } while (wasWork && loopsRemaining > 0);
+            } while (!_stopping && wasWork && loopsRemaining > 0);
```
halter73: I don't understand this part of the change.

halter73: What are you trying to achieve here? If you are afraid of this running during the second uv_run, you would want this as a while instead of a do/while condition.

davidfowl: Today we do at least 8 loops after work has been posted to check for more work (on the libuv thread). When shutting down, this flag just ensures that we don't waste any time looking for more work.

halter73: It's not like that takes much time. Also, the loop exits immediately after `_workAdding` is empty.

PR #1769, opened 2017-04-25
Add wait in ListenerPrimaryTests to avoid race condition with List.Add

### LINKED ISSUE #1681: Fix race in ListenerPrimaryTests.ConnectionsGetRoundRobinedToSecondaryListeners test
We saw it in 1.1.3 patch builds but same can happen in dev. 
```
Unhandled Exception: System.AggregateException: One or more errors occurred. (Assert.Equal() Failure
          ↓ (pos 0)
Expected: Secondary
Actual:   Primary
          ↑ (pos 0)) ---> Xunit.Sdk.EqualException: Assert.Equal() Failure
   at Xunit.Assert.Equal(String expected, String actual, Boolean ignoreCase, Boolean ignoreLineEndingDifferences, Boolean ignoreWhiteSpaceDifferences)
   at Microsoft.AspNetCore.Server.KestrelTests.ListenerPrimaryTests.<ConnectionsGetRoundRobinedToSecondaryListeners>d__0.MoveNext() in /home/pakrym/KestrelHttpServer/test/Microsoft.AspNetCore.Server.KestrelTests/ListenerPrimaryTests.cs:line 80
```

natemcmaster:
I couldn't actually reproduce the race condition in ListenerPrimaryTests even in a tight loop of 100,000 iterations. However, from discussion with @pakrym, we think this should fix the issue in the test.

Resolves #1681. We can re-open and try something else if the test still fails.

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Kestrel.Transport.Libuv/Internal/ListenerPrimary.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel.Transport.Libuv/Internal/ListenerPrimary.cs
@@ -33,6 +33,11 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Transport.Libuv.Internal
         {
         }
 
+        /// <summary>
+        /// For testing purposes.
+        /// </summary>
+        public int UvPipeCount => _dispatchPipes.Count;
+
         private UvPipeHandle ListenPipe { get; set; }
 
         public async Task StartAsync(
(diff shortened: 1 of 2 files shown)


src/Microsoft.AspNetCore.Server.Kestrel.Transport.Libuv/Internal/ListenerPrimary.cs
```diff
@@ -33,6 +33,11 @@ public ListenerPrimary(LibuvTransportContext transportContext) : base(transportC
         {
         }
 
+        /// <summary>
+        /// For testing purposes.
+        /// </summary>
+        public int UvPipeCount => _dispatchPipes.Count;
```
cesarbs: Make it `internal` if it's just for testing.

natemcmaster: That was my first approach...which led to more issues. Setting InternalsVisibleTo on the Transport.Libuv assembly caused conflicts in the test project because both Transport.LibUv and .Core use TaskCache.Sources. This cause compile to fail because Microsoft.Extensions.Internal.TaskCache is ambiguous between the assemblies.

cesarbs: 💩 I see.

PR #1775, opened 2017-04-26
React to Logging API changes

BrennanConroy:


samples/SampleApp/Startup.cs
```diff
@@ -42,6 +41,10 @@ public static void Main(string[] args)
             };
 
             var host = new WebHostBuilder()
+                .ConfigureLogging((_, factory) =>
+                {
+                    factory.AddConsole();
```
cesarbs: What happened to the log level?

BrennanConroy: Trace is the lowest level, and by default all logs are enabled now

PR #1776, opened 2017-04-26
Simplify connection lifetime control flow

### LINKED ISSUE #1693: Fail: Uncaught exception from the PrepareRequest method of an IAdaptedConnection
When trying out SSL in ASP.NET Core I see the following in the logging output:

```
fail: Microsoft.AspNetCore.Server.Kestrel[0]
      Uncaught exception from the PrepareRequest method of an IAdaptedConnection.
System.ObjectDisposedException: Cannot access a disposed object.
Object name: 'SslStream'.
   at System.Net.Security.SslState.CheckThrow(Boolean authSuccessCheck, Boolean shutdownCheck)
   at System.Net.Security.SslStream.get_RemoteCertificate()
   at Microsoft.AspNetCore.Server.Kestrel.Https.HttpsConnectionAdapter.HttpsAdaptedConnection.PrepareRequest(IFeatureCollection requestFeatures)
   at Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http.Frame.Reset()
```

Sample app here: https://github.com/danroth27/active-directory-dotnet-webapp-openidconnect-aspnetcore-b2c/tree/danroth27/aspnetcore2

halter73:
Fixes #1693

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/Http/FrameOfT.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/Http/FrameOfT.cs
@@ -28,15 +28,22 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
         /// The resulting Task from this loop is preserved in a field which is used when the server needs
         /// to drain and close all currently active connections.
         /// </summary>
-        public override async Task RequestProcessingAsync()
+        public override async Task ProcessRequestsAsync()
         {
+            var dataRecieved = false;
+
             try
             {
                 while (!_requestProcessingStopping)
                 {
                     TimeoutControl.SetTimeout(_keepAliveTicks, TimeoutAction.CloseConnection);
 
-                    InitializeHeaders();
+                    if (dataRecieved)
+                    {
+                        // Call Reset() before data arrives for the 2nd, 3rd, 4th... keep-alive requests,
+                        // so idle connections don't root objects unnecessarily.
+                        Reset();
+                    }
 
                     while (!_requestProcessingStopping)
                     {
@@ -46,6 +53,14 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
 
                         try
                         {
+                            if (!result.Buffer.IsEmpty && !dataRecieved)
+                            {
+                                // Wait to call Reset() until data arrives for the 1st request in order to give
+                                // FrameConnection.ApplyConnectionAdaptersAsync() the opportunity to set AdaptedConnections.
+                                Reset();
+                                dataRecieved = true;
+                            }
+
                             ParseRequest(result.Buffer, out consumed, out examined);
                         }
                         catch (InvalidOperationException)
@@ -77,8 +92,8 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
                                     throw BadHttpRequestException.GetException(
                                         RequestRejectionReason.InvalidRequestLine);
                                 case RequestProcessingStatus.ParsingHeaders:
-                                    throw BadHttpRequestException.GetException(RequestRejectionReason
-                                        .MalformedRequestInvalidHeaders);
+                                    throw BadHttpRequestException.GetException(
+                                        RequestRejectionReason.MalformedRequestInvalidHeaders);
                             }
                         }
                     }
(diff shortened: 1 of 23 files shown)


src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/Http/FrameOfT.cs
```diff
@@ -28,15 +28,22 @@ public Frame(IHttpApplication<TContext> application, FrameContext frameContext)
         /// The resulting Task from this loop is preserved in a field which is used when the server needs
         /// to drain and close all currently active connections.
         /// </summary>
-        public override async Task RequestProcessingAsync()
+        public override async Task ProcessRequestsAsync()
         {
+            var dataRecieved = false;
```
cesarbs: nit: typo

src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/Http/FrameOfT.cs
```diff
@@ -28,15 +28,22 @@ public Frame(IHttpApplication<TContext> application, FrameContext frameContext)
...
+
             try
             {
                 while (!_requestProcessingStopping)
                 {
                     TimeoutControl.SetTimeout(_keepAliveTicks, TimeoutAction.CloseConnection);
 
-                    InitializeHeaders();
+                    if (dataRecieved)
```
davidfowl: This is pretty gross. What's the deal?

src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/Http/FrameOfT.cs
```diff
@@ -46,6 +53,14 @@ public Frame(IHttpApplication<TContext> application, FrameContext frameContext)
 
                         try
                         {
+                            if (!result.Buffer.IsEmpty && !dataRecieved)
+                            {
+                                // Wait to call Reset() until data arrives for the 1st request in order to give
```
davidfowl: There must be a cleaner way to fix this.

src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/Http/FrameOfT.cs
halter73: @pakrym and I discussed this. The bool wasn't there before, but this seemed cleaner than the `while (true) ... break` I had earlier.

Is there some alternative you're proposing. Do you think we should go back to delaying the call to RequestProcessingAsync() and re-adding the frame started TCS?

One benefit of this design is that dictionary initilization is elided for empty connections and delayed until a request is actually sent.

benaadams: Don't need to add an extra `bool` to the state machine just check the headers? `if (FrameRequestHeaders != null)`

halter73: You definitely could do that. I liked being a little more explicit.

benaadams: Actually why not add this one to the end of the function, and called always rather than the start behind a flag?

halter73: If you look at the full diff for FrameOfT, you'll see that's exactly where Reset() used to be called (there and in Frame.Start()). It was still behind a `if (!_requestProcessing)` flag though with a big comment explaining how request information needs to be preserved in this case and AdaptedConnection.PrepareRequest() might throw.

The way Reset() was called was always gross. This just puts all the callers closer together in one file. It's also more efficient since it delays or elides the first call to Reset(). I understand disliking waiting for Input.ReadAsync() to complete in order to ensure the AdaptedConnections property is set. That's why I call it out so call it out in a comment, but it just makes sense to delay Reset until we actually start processing a request.

Another thing I'm considering to make this even cleaner is to move the calls to `new Frame(Request|Response)Headers()` before the first call Reset() in order to avoid the `if (FrameRequestHeaders == null)` check that's currently in the Reset() method of this PR.

benaadams: Yeah see, maybe just naming then, with `Reset()` being first thing rather than last thing. Something like `SetupRequest()` `PrepareRequest()`, `InitalizeRequest()` etc?

benaadams: Following on what you said second (first) could be
```csharp
if (firstTest)
{
   InitalizeFrame(); // new Frame(Request|Response)Headers()
   InitalizeRequest();
}
```

halter73:
⬆️ 📅

PR #1779, opened 2017-04-26
Move user-facing strings into resource files

### LINKED ISSUE #1722: Move user-facing strings to resx
Exception and log messages can be moved to the resx file.

natemcmaster:
Resolves #1722


src/Microsoft.AspNetCore.Server.Kestrel.Transport.Sockets/SocketsStrings.resx
```diff
@@ -0,0 +1,126 @@
...
+  <resheader name="version">
+    <value>2.0</value>
+  </resheader>
+  <resheader name="reader">
+    <value>System.Resources.ResXResourceReader, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
+  </resheader>
+  <resheader name="writer">
+    <value>System.Resources.ResXResourceWriter, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
+  </resheader>
+  <data name="OnlyIpEndPointsSupported" xml:space="preserve">
```
cesarbs: nit: I**P**

src/Microsoft.AspNetCore.Server.Kestrel.Core/CoreStrings.resx
```diff
@@ -117,10 +117,193 @@
...
+    <value>Bad chunk size data.</value>
+  </data>
+  <data name="BadRequest_BadChunkSuffix" xml:space="preserve">
+    <value>Bad chunk suffix.</value>
+  </data>
+  <data name="BadRequest_ChunkedRequestIncomplete" xml:space="preserve">
+    <value>Chunked request incomplete.</value>
+  </data>
+  <data name="BadRequest_FinalTransferCodingNotChunked" xml:space="preserve">
+    <value>Final transfer coding is not "chunked": "{detail}"</value>
```
cesarbs: This change made me realize we're inconsistent with the use of `''` and `""` in error messages. And sometimes we use neither.

natemcmaster: I updated this error message to use single quotes. Also, I didn't understand the error message when I first saw it, so I changed it to `The message body length cannot be determined because the final transfer encoding was set to '{detail}' instead of 'chunked'.`. Okay with you?

src/Microsoft.AspNetCore.Server.Kestrel.Core/CoreStrings.resx
```diff
@@ -130,7 +130,7 @@
     <value>Chunked request incomplete.</value>
   </data>
   <data name="BadRequest_FinalTransferCodingNotChunked" xml:space="preserve">
-    <value>Final transfer coding is not "chunked": "{detail}"</value>
+    <value>The message body length cannot be determined because the final transfer encoding was set to '{detail}' instead of 'chunked'.</value>
```
cesarbs: For whatever reason the spec always refers to this as "transfer coding", despite the header name being called Transfer-**En**coding. That's why it was spelled like that in the original message.

natemcmaster:
FYI team - if you need to make changes, you can regenerate resx files by running

1. `./build.ps1 /t:resx`
2. `cd src/projectDir/ && dotnet msbuild /t:resx`

src/Microsoft.AspNetCore.Server.Kestrel.Core/CoreStrings.resx
natemcmaster: > If a Transfer-Encoding header field is present in a response and
       the chunked transfer coding is not the final encoding, the
       message body length is determined by reading the connection until
       it is closed by the server.  If a Transfer-Encoding header field
       is present in a request and the chunked transfer coding is not
       the final encoding, the message body length cannot be determined
       reliably; the server MUST respond with the 400 (Bad Request)
       status code and then close the connection.

Hmm, yup you're right. Ok to leave as encoding though? It makes more sense to me.

cesarbs: Let's follow the RFC's lingo.

natemcmaster: ok

Issue #1686, opened 2017-04-14
Socket Transport 


davidfowl:
- [x] - The functional tests support running both on both transports but tests aren't running for sockets right now. We need to parameterize these. 
- [x] - ~~TimeoutServerTests hard code the libuv transport, this needs to support any transport.~~
- [x] - Get the socket transport on the benchmarks server (https://github.com/aspnet/benchmarks/issues/250)
- [x] - Add logging to the SocketsTransport, today it has nothing
- [x] - Add thread pool dispatching by default. We may not get time to do this properly for 2.0.0 https://github.com/aspnet/KestrelHttpServer/issues/1641, so until then, we should just abuse thread pool dispatch and offer knobs on the transport to turn it off.
- [x] - The systemd activation test assumes we can bind using a unix pipe and an open file descriptor. We need to only run that test when using the libuv transport as sockets only supports binding via and IPEndPoint.
- [x] - Implement graceful server shutdown
- [x] - Throw reasonable exceptions for unsupported ListenOptions
- [ ] - Add SocketTransportOptions
- [ ] - Smarter Output Buffering 
- [ ] - 0 byte reads to help the memory issues
- [x] - Use the new Memory<byte> APIs in .NET Core 2.1
- [ ] - Security review

None of the performance work is listed here specifically since that will be on going. We need to experiment with different ways of using `SocketAsyncEventArgs` to see if what we have is the most efficient.

One other note is that we aren't doing 0 byte reads right now (which libuv *does* so), this is usually done to avoid allocating until data is ready but we'll see how it affects both memory and throughput.

/cc @geoffkizer @stephentoub

vancem:
>  Add thread pool dispatching by default. We may not get time to do this properly for 2.0.0 #1641, so until then, we should just abuse thread pool dispatch and offer knobs on the transport to turn it off.

Why do we want thread pool dispatching by default?    For Sockets there is no need to hop threads, so we should not.   There is this issue that we would like some batching of outgoing writes, but that seems like a separate issue which we would solve differently (so that the fast path does NOT involve queuing threadpool work items).

benaadams:
libuv needs dispatching (otherwise application code will block the I/O) and that is currently the default transport

davidfowl:
@vancem It's a stop gap until we decide which layers do how much buffering.

davidfowl:
Some notes:
This test is libuv specific right now - ConnectionResetBetweenRequestsIsLoggedAsDebug
[...]

PR #1782, opened 2017-04-27
Close the input if the output ends but not vice versa

### LINKED ISSUE #1774: SocketTransport: FIN handling
If a client sends a request and then sends a FIN to indicate it won't send anymore data, this should
- not be [treated as an exception](https://github.com/aspnet/KestrelHttpServer/blob/e76737b8a63d0230feae1a493d81be8b676387bf/src/Microsoft.AspNetCore.Server.Kestrel.Transport.Sockets/SocketConnection.cs#L95)
- not cause the server to [stop sending a reply](https://github.com/aspnet/KestrelHttpServer/blob/e76737b8a63d0230feae1a493d81be8b676387bf/src/Microsoft.AspNetCore.Server.Kestrel.Transport.Sockets/SocketConnection.cs#L58).

davidfowl:
- This gives kestrel control over when the output closes

Fixes #1774

EDIT: I'll merge when all of the tests pass locally on my dev box. Right now I'm seeing a couple of failures:

```
ConnectionResetMidRequestIsLoggedAsDebug
ConnectionResetBetweenRequestsIsLoggedAsDebug
ConnectionResetPriorToRequestIsLoggedAsDebug
```

These all work via the specific logger name (the libuv logger) so they don't work right no with the socket transport. It shouldn't block this commit though. I'll add these to the issue here #1686

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Kestrel.Transport.Sockets/SocketConnection.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel.Transport.Sockets/SocketConnection.cs
@@ -4,8 +4,10 @@
 using System;
 using System.Collections.Generic;
 using System.Diagnostics;
+using System.IO;
 using System.Net;
 using System.Net.Sockets;
+using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Server.Kestrel.Internal.System.Buffers;
 using Microsoft.AspNetCore.Server.Kestrel.Internal.System.IO.Pipelines;
@@ -23,8 +25,8 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Transport.Sockets
         private IPipeWriter _input;
         private IPipeReader _output;
         private IList<ArraySegment<byte>> _sendBufferList;
-
-        private const int MinAllocBufferSize = 2048;        // from libuv transport
+        private const int MinAllocBufferSize = 2048;
+        private bool _aborted;
 
         internal SocketConnection(Socket socket, SocketTransport transport)
         {
@@ -56,6 +58,9 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Transport.Sockets
                 // will trigger the output closing once the input is complete.
                 if (await Task.WhenAny(receiveTask, sendTask) == sendTask)
                 {
+                    // Tell the reader it's being aborted
+                    Volatile.Write(ref _aborted, true);
+
                     // Shutdown the both sides
                     _socket.Shutdown(SocketShutdown.Both);
                 }
@@ -93,9 +98,14 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Transport.Sockets
 
                         if (bytesReceived == 0)
                         {
-                            // We receive a FIN so throw an exception so that we cancel the input
-                            // with an error
-                            throw new TaskCanceledException("The request was aborted");
+                            // Forced FIN
+                            if (Volatile.Read(ref _aborted))
+                            {
+                                throw new TaskCanceledException("The request was aborted");
+                            }
+
+                            // FIN
+                            break;
                         }
 
                         buffer.Advance(bytesReceived);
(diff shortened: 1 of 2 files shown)


src/Microsoft.AspNetCore.Server.Kestrel.Transport.Sockets/SocketConnection.cs
```diff
@@ -51,11 +53,17 @@ internal SocketConnection(Socket socket, SocketTransport transport)
...
 
-                // Wait for eiher of them to complete (note they won't throw exceptions)
-                await Task.WhenAny(receiveTask, sendTask);
+                // If the sending task completes then close the receive
+                // We don't need to do this in the other direction because the kestrel
+                // will trigger the output closing once the input is complete.
+                if (await Task.WhenAny(receiveTask, sendTask) == sendTask)
+                {
+                    // Tell the reader it's being aborted
+                    Volatile.Write(ref _aborted, true);
```
davidfowl: One alternative would be to dispose the socket here and catch that exception in the reader but that seems grosser.

davidfowl: Actually that looks cleaner...

src/Microsoft.AspNetCore.Server.Kestrel.Transport.Sockets/SocketConnection.cs
```diff
@@ -111,12 +114,37 @@ internal SocketConnection(Socket socket, SocketTransport transport)
...
+                if (ex is SocketException se)
+                {
+                    if (se.SocketErrorCode == SocketError.ConnectionReset)
+                    {
+                        // Connection reset
+                        error = new ConnectionResetException(ex.Message, ex);
+                    }
+                    else if (se.SocketErrorCode == SocketError.OperationAborted)
+                    {
+                        error = new TaskCanceledException("The request was aborted");
```
tmds: disposing the socket can also cause ObjectDisposedException

tmds:
@davidfowl unrelated, I see `_socket.Shutdown(SocketShutdown.Receive)` is called.

Shutting down receive doesn't do anything at the TCP level.
A use case for calling this on Linux is to unblock an async receive. That isn't the case here. Perhaps this is also it's purpose on Windows... unsure.
There doesn't seem to be a meaningful reason to call it here, so you can remove it.

src/Microsoft.AspNetCore.Server.Kestrel.Transport.Sockets/SocketConnection.cs
davidfowl: I was originally checking for that but couldn't reproduce it. I guess it would happen on the next ReadAsync call? I'll confirm.

tmds: Yes, I looked in the source code. If you call ReceiveAsync and the socket is already Disposed, it throws ObjectDisposedException.
If the ReceiveAsync is in-flight and the socket gets disposed, it throws SocketError.OperationAborted.
So it's very unlikely too occur, but not impossible.

davidfowl: Updated, I was able to reproduce it (by forcing it).

davidfowl:
> @davidfowl unrelated, I see _socket.Shutdown(SocketShutdown.Receive) is called.

I removed it.

src/Microsoft.AspNetCore.Server.Kestrel.Transport.Sockets/SocketConnection.cs
```diff
@@ -106,17 +109,45 @@ internal SocketConnection(Socket socket, SocketTransport transport)
...
+                    }
+                }
+
+                if (ex is ObjectDisposedException)
```
halter73: Nit: if this is changed to an `else if`, the last `else if (error == null)` could be changed to just an `else`.

PR #1784, opened 2017-04-27
Implement new request trace identifier format

### LINKED ISSUE #1591: Change request trace ID format
Follow up to https://github.com/aspnet/KestrelHttpServer/pull/1578. In this PR, we Kestrel started generating request IDs instead of letting aspnet/Hosting do this. At the moment, request IDs use the same code to generate connection ID and request ID.

We want to enhance this change further by making the requestID to the following format:

```c#
//pseudocode
var count = ++_connection.RequestCount;
HttpContext.TraceIdentifier = $"{_connection.ConnectionId}:{count.ToString("X").PadLeft(5, '0')}";
```

Result:
```
Connection 1 Id = 001BCA
  Request 1: Id = 001BCA:00000
  Request 2: Id = 001BCA:00001
...
  Request 11: Id= 001BCA:0000B

Connection 2 Id = 001BCB
  Request 1: Id = 001BCB:00000
  Request 2: Id = 001BCB:00001
```

In the unlikely event there are more than 0xFFFFF (1,048,575) requests per connection, the counter must not truncate to 5 characters, but use as many characters required to express request count in hex.

natemcmaster:
Resolves #1591.

The format:
The trace identifier begins with connection ID and ends with a number that increments with each request per connection.

Example:
Connection ID = xyz
Request 1 = "xyz:00000001"
Request 2 = "xyz:00000002"
...
Request 15 = "xyz:0000000F"
Request 16 = "xyz:00000010"

**Benchmark**

I implemented the unsafe 'ConcatAsHexSuffix' method because it allocates one fewer string and is almost 3x faster in a tight loop of 0.5 million iterations.

```
            Method |       Mean |    StdDev |  Op/s | Scaled | Scaled-StdDev | Allocated |
------------------ |----------- |---------- |------ |------- |-------------- |---------- |
      UintToString | 50.0737 ms | 1.4121 ms | 19.97 |   1.00 |          0.00 |     60 MB |
 ConcatAsHexSuffix | 18.3235 ms | 0.9086 ms | 54.57 |   0.37 |          0.02 |     36 MB |
```

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/Infrastructure/StringUtilities.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/Infrastructure/StringUtilities.cs
@@ -3,7 +3,7 @@
 
 namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Infrastructure
 {
-    internal class AsciiUtilities
+    internal class StringUtilities
     {
         public static unsafe bool TryGetAsciiString(byte* input, char* output, int count)
         {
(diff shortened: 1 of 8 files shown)


src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/Infrastructure/StringUtilities.cs
```diff
@@ -75,5 +75,49 @@ public static unsafe bool TryGetAsciiString(byte* input, char* output, int count
...
+            var i = 0;
+            if (str != null)
+            {
+                for (i = 0; i < str.Length; i++)
+                {
+                    charBuffer[i] = str[i];
+                }
+            }
+
+            charBuffer[i++] = separator;
```
benaadams: Don't take a read+write dependency on `i`?
```
charBuffer[ i ] =
charBuffer[i+1] =
charBuffer[i+2] =
```

natemcmaster: Nice! That bumps perf even more.

```
            Method |       Mean |    StdDev |  Op/s | Scaled | Scaled-StdDev | Allocated |
------------------ |----------- |---------- |------ |------- |-------------- |---------- |
      UintToString | 50.0750 ms | 1.6295 ms | 19.97 |   1.00 |          0.00 |     60 MB |
 ConcatAsHexSuffix | 17.6745 ms | 0.7687 ms | 56.58 |   0.35 |          0.02 |     36 MB |
```

test/Microsoft.AspNetCore.Server.Kestrel.Performance/StringUtilitiesBenchmark.cs
```diff
@@ -0,0 +1,34 @@
...
+using Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Infrastructure;
+
+namespace Microsoft.AspNetCore.Server.Kestrel.Performance
+{
+    [Config(typeof(CoreConfig))]
+    public class StringUtilitiesBenchmark
+    {
+        private const int Iterations = 500_000;
+
+        [Benchmark(Baseline = true)]
```
pakrym: Set iteration count attribute

natemcmaster:
Fixed the math on the benchmark:
```
            Method |        Mean |    StdErr |    StdDev |        Op/s | Scaled | Scaled-StdDev | Allocated |
------------------ |------------ |---------- |---------- |------------ |------- |-------------- |---------- |
      UintToString | 100.7636 ns | 1.6000 ns | 8.7635 ns |  9924216.61 |   1.00 |          0.00 |     120 B |
 ConcatAsHexSuffix |  36.5012 ns | 0.3564 ns | 1.9523 ns | 27396353.45 |   0.36 |          0.03 |      72 B |
```

halter73:
It would take 16^8 requests over a single connection to get a collision. I don't think this is a big deal, but a collision is certainly possible. If 1,000,000 requests could be made over a single connection per second (which is more than we've ever measured), it would take ~72 minutes of that sustained throughput to get a collision.

Is there anything up the stack that would fail badly given a duplicate TraceIdentifier?

natemcmaster:
Logging is the only usage of TraceIdentifer that I'm aware of. I doubt a collision would be disastrous.

But, to be safe we could jump to `ulong`. 16^16 requests on a single connection at 1 million RPS would take ~584,000 years to cause a collision. Of course, we're right back in the same boat as we improve perf. At a mere 1 trillion requests per seconds, we'll hit a trace identifier collision in about 5100 hours.

Personally I think `uint` is sufficient, though [you never know](https://techcrunch.com/2014/12/03/gangnam-style-has-been-viewed-so-many-times-it-broke-youtubes-code/).

Issue #1721, opened 2017-04-19
Test failure: RegisterAddresses_IPv4_Success


cesarbs:
```
System.IO.IOException : Failed to bind to address http://[::]:52353: address already in use.
---- Microsoft.AspNetCore.Server.Kestrel.Transport.Abstractions.AddressInUseException : Error -98 EADDRINUSE address already in use
-------- Microsoft.AspNetCore.Server.Kestrel.Transport.Libuv.Internal.Networking.UvException : Error -98 EADDRINUSE address already in use
   at Microsoft.AspNetCore.Server.Kestrel.Core.KestrelServer.<BindToEndpoints>d__16`1.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at Microsoft.AspNetCore.Server.Kestrel.Core.KestrelServer.<BindToServerAddresses>d__15`1.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at Microsoft.AspNetCore.Server.Kestrel.Core.KestrelServer.<StartAsync>d__14`1.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at Microsoft.AspNetCore.Hosting.Internal.WebHost.<StartAsync>d__25.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at Microsoft.AspNetCore.Hosting.WebHostExtensions.Start(IWebHost host)
   at Microsoft.AspNetCore.Server.Kestrel.FunctionalTests.AddressRegistrationTests.<RegisterAddresses_Success>d__9.MoveNext()
[...]

PR #1787, opened 2017-04-28
Avoid port conflicts in AddressRegistrationTests (#1721)

cesarbs:
#1721

There's still some room for conflicts in the \*`StaticPort`\* tests after 10 retries, but we have to test binding to a static port.

### DIFF
--- a/test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/AddressRegistrationTests.cs
+++ b/test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/AddressRegistrationTests.cs
@@ -24,101 +24,145 @@ using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Options;
 using Xunit;
 using Xunit.Abstractions;
+using Xunit.Sdk;
 
 namespace Microsoft.AspNetCore.Server.Kestrel.FunctionalTests
 {
     public class AddressRegistrationTests
     {
-        private readonly ILoggerFactory _logger;
+        private const int MaxRetries = 10;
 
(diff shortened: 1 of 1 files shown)


natemcmaster:
LGTM. I assume it passes? It looks like we need to add rel/* to our travis/appveyor config so tests run on these branches.

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/AddressRegistrationTests.cs
```diff
@@ -137,15 +181,51 @@ string FixUrl(string url)
...
-        private async Task RegisterIPEndPoint_Success(IPEndPoint endPoint, Func<IPEndPoint, string> testUrl)
+        private Task RegisterAddresses_Success(string addressInput, string testUrl, bool appendPortToTestUrl)
+            => RegisterAddresses_Success(addressInput, new[] { testUrl }, appendPortToTestUrl);
+
+        private async Task RegisterAddresses_StaticPort_Success(string addressInput, string[] testUrls)
+        {
+            var retryCount = 0;
+            var errors = new List<Exception>();
+
+            while (retryCount < MaxRetries)
```
davidfowl: How about less retries and a Delay?

cesarbs: I thought of a delay too, but @halter73 pointed out that we're getting a new port each time, so why delay?

davidfowl: I see

PR #1791, opened 2017-04-28
Prepare for OnReader/WriterCallbacks changes

### LINKED ISSUE #1771: Socket close on Timeout
I am looking in the code to see how the Timeout is implemented.
I think something is missing that unblocks the socket read or I missed it.

Pseudo code of socket reading:
```
do {
buffer = socket.read();                   // (1)
completed = pipewriter.flush(buffer);
} while (!completed);
pipewriter.complete();
socket.shutdown(readend);                 // (2)
```

On timeout we want to reach (2) this requires getting unblocked at (1).

davidfowl:
- This change does a few things:

1. It adds the events we will replace with
pipe events to IConnectionContext and IConnectionInformation to get out of
band notifications about pipe completions.

2. It also implements those callbacks
and exposing slight changes we'll need to make once we have them. The idea is
that we can delete/replace these methods once we have the new pipe API and things
will keep working

3. ~It fixes an issue where sockets potentially wouldn't close because of
output back pressure from the OS. If there was no pending pipe.ReadAsync call the
signal to cancel the pending read could be missed.~

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/FrameConnection.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/FrameConnection.cs
@@ -87,8 +87,11 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal
             }
         }
 
-        public async void OnConnectionClosed()
+        public async void OnConnectionClosed(Exception ex)
         {
+            // Abort the connection (if it isn't already aborted)
+            _frame.Abort(ex);
+
             Log.ConnectionStop(ConnectionId);
             KestrelEventSource.Log.ConnectionStop(this);
             _socketClosedTcs.SetResult(null);
--- a/src/Microsoft.AspNetCore.Server.Kestrel.Transport.Libuv/Internal/LibuvConnection.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel.Transport.Libuv/Internal/LibuvConnection.cs
@@ -66,17 +66,18 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Transport.Libuv.Internal
                 // Start socket prior to applying the ConnectionAdapter
                 StartReading();
 
+                Exception error = null;
+
                 try
                 {
                     // This *must* happen after socket.ReadStart
                     // The socket output consumer is the only thing that can close the connection. If the
                     // output pipe is already closed by the time we start then it's fine since, it'll close gracefully afterwards.
                     await Output.WriteOutputAsync();
-                    _connectionContext.Output.Complete();
                 }
                 catch (UvException ex)
                 {
-                    _connectionContext.Output.Complete(ex);
+                    error = new IOException(ex.Message, ex);
                 }
                 finally
                 {
@@ -91,13 +92,24 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Transport.Libuv.Internal
                     _socket.Dispose();
 
                     // Tell the kestrel we're done with this connection
-                    _connectionContext.OnConnectionClosed();
+                    _connectionContext.OnConnectionClosed(error);
+                    _connectionContext.Output.Complete(error);
                 }
             }
             catch (Exception e)
             {
                 Log.LogCritical(0, e, $"{nameof(LibuvConnection)}.{nameof(Start)}() {ConnectionId}");
             }
+        }
+
+        public override void OnApplicationComplete()
+        {
+            Thread.Post(connection =>
+            {
+                // Call uv_close which will fire
+                connection._socket.Dispose();
+            },
+            this);
         }
 
         // Called on Libuv thread
@@ -221,7 +233,10 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Transport.Libuv.Internal
                 // ReadStart() can throw a UvException in some cases (e.g. socket is no longer connected).
                 // This should be treated the same as OnRead() seeing a "normalDone" condition.
                 Log.ConnectionReadFin(ConnectionId);
-                Input.Complete(new IOException(ex.Message, ex));
+                var error = new IOException(ex.Message, ex);
+
+                _connectionContext.Abort(error);
+                Input.Complete(error);
             }
         }
     }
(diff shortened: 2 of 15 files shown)


src/Microsoft.AspNetCore.Server.Kestrel.Transport.Sockets/SocketConnection.cs
```diff
@@ -220,13 +215,27 @@ private void SetupSendBuffers(ReadableBuffer buffer)
...
                     }
                 }
-
-                // We're done reading
-                _output.Complete();
             }
             catch (Exception ex)
             {
-                _output.Complete(ex);
+                if (ex is SocketException se && se.SocketErrorCode == SocketError.OperationAborted || ex is ObjectDisposedException)
```
cesarbs:
```csharp
catch (SocketException se) when (se.SocketErrorCode == SocketError.OperationAborted || ex is ObjectDisposedException)
````

File: src/Microsoft.AspNetCore.Server.Kestrel.Transport.Sockets/SocketConnection.cs [hunk 2]
```diff
@@ -220,13 +215,27 @@ private void SetupSendBuffers(ReadableBuffer buffer)
...
+                {
+                    // We had to abort the write
+                    error = null;
+                }
+                else if (ex is IOException)
```
cesarbs:
```csharp
catch (IOException ex)
```

src/Microsoft.AspNetCore.Server.Kestrel.Transport.Libuv/Internal/LibuvConnection.cs
```diff
@@ -100,6 +102,16 @@ public LibuvConnection()
             }
         }
 
+        public override void OnApplicationComplete()
+        {
+            Thread.Post(connection =>
+            {
+                // Call uv_close which will fire any pending uv_write callbacks
+                connection._socket.Dispose();
```
halter73: > EDIT: This needs a bunch of work. It's too easy to cut valid responses off.

This here is why. Since WriteAsync is a lie, the app can complete well before the socket can be closed.

src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/FrameConnection.cs
```diff
@@ -87,8 +87,11 @@ public void StartRequestProcessing()
             }
         }
 
-        public async void OnConnectionClosed()
+        public async void OnConnectionClosed(Exception ex)
         {
+            // Abort the connection (if it isn't already aborted)
+            _frame.Abort(ex);
```
halter73: Why?

test/Microsoft.AspNetCore.Server.Kestrel.Transport.Libuv.Tests/LibuvOutputConsumerTests.cs
```diff
@@ -282,10 +284,10 @@ public void Dispose()
...
 
                     // Add more bytes to the write-behind buffer to prevent the next write from
-                    ((ISocketOutput) socketOutput).Write((writableBuffer, state) =>
-                        {
-                            writableBuffer.Write(state);
-                        },
+                    ((ISocketOutput)socketOutput).Write((writableBuffer, state) =>
+                       {
+                           writableBuffer.Write(state);
+                       },
```
halter73: Nit: these curly braces are only indented 3 spaces.

src/Microsoft.AspNetCore.Server.Kestrel.Transport.Sockets/SocketConnection.cs
```diff
@@ -220,13 +215,27 @@ private void SetupSendBuffers(ReadableBuffer buffer)
...
+                    error = ex;
+                }
+                else
+                {
+                    error = new IOException(ex.Message, ex);
+                }
+            }
+            finally
+            {
+                _connectionContext.OnConnectionClosed(error);
```
halter73: Semantically it makes sense to call this after the socket is disposed. If it socket is always closed after the call to OnApplicationComplete(), OnConnectionClosed() wouldn't need to exist.

src/Microsoft.AspNetCore.Server.Kestrel.Transport.Sockets/SocketConnection.cs
halter73: Also change DoReceive() to follow the same pattern @cesarbs suggests.

src/Microsoft.AspNetCore.Server.Kestrel.Transport.Libuv/Internal/LibuvOutputConsumer.cs
```diff
@@ -65,6 +65,11 @@ public class LibuvOutputConsumer
                         }
                     }
 
+                    if (_socket.IsClosed)
```
halter73: Gross. The OS should timeout writes for completely unresponsive clients to begin with. Long term, the better solution is to enforce a minimum minimum data rate for responses.

This might require a way to cancel LibuvAwaitables, but it definitely not OK to immediately kill the socket and any ongoing writes just because the request processing loop completed.

You could consider doing that after some kind of timeout, but at that point you might as well just enforce a minimum rate.

davidfowl: There still needs to be an explicit way to signal the transport.

src/Microsoft.AspNetCore.Server.Kestrel.Transport.Sockets/SocketConnection.cs
davidfowl: Will do

src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/FrameConnection.cs
davidfowl: It replaces https://github.com/aspnet/KestrelHttpServer/blob/3b2d0a52f39fa024431c5b553f31bc4894291c0b/src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/Http/OutputProducer.cs#L117

Issue #1761, opened 2017-04-25
EBUSY when running functional tests in a loop on linux


davidfowl:
I ran the functional tests on an azure machine in a tight loop and got an EBUSY. Unfortunately, I don't know what test caused it. Nor do I have any logs.

```
Microsoft.AspNetCore.Server.Kestrel.Transport.Libuv.Internal.Networking.UvException: Error -16 EBUSY resource busy or locked
   at Microsoft.AspNetCore.Server.Kestrel.Transport.Libuv.Internal.Networking.LibuvFunctions.ThrowError(Int32 statusCode)
   at Microsoft.AspNetCore.Server.Kestrel.Transport.Libuv.Internal.Networking.LibuvFunctions.ThrowIfErrored(Int32 statusCode)
   at Microsoft.AspNetCore.Server.Kestrel.Transport.Libuv.Internal.Networking.LibuvFunctions.loop_close(UvLoopHandle handle)
   at Microsoft.AspNetCore.Server.Kestrel.Transport.Libuv.Internal.Networking.UvLoopHandle.ReleaseHandle()
   at System.Runtime.InteropServices.SafeHandle.InternalDispose()
   at Microsoft.AspNetCore.Server.Kestrel.Transport.Libuv.Internal.LibuvThread.ThreadStart(Object parameter)
```

halter73:
Well, we *do* know how to get rid of the EBUSY 😉.

davidfowl:
Got more info with logging:

There seems to be a rogue pipe handle still open when we get the EBUSY:

`Microsoft.AspNetCore.Server.Kestrel.Transport.Libuv.Internal.Networking.UvPipeHandle`

davidfowl:
It gets better:

```
Starting thread 31
Setting up tcp listener
Creating dispatch pipe 1
Disposing the listen socket
Disposing listen pipe and 0 dispatch pipes
Microsoft.AspNetCore.Server.Kestrel.Transport.Libuv.Internal.Networking.UvPipeHandle - OnListenPipe.dispatchPipe
Exception:
================
Microsoft.AspNetCore.Server.Kestrel.Transport.Libuv.Internal.Networking.UvException: Error -16 EBUSY resource busy or locked
   at Microsoft.AspNetCore.Server.Kestrel.Transport.Libuv.Internal.Networking.LibuvFunctions.ThrowError(Int32 statusCode)
[...]

PR #1798, opened 2017-04-29
Fix EBUSY errors on uv_loop_close 

davidfowl:
We need this walk because we call ReadStop on 2 places:
 1. On the dispatch pipe UvPipeHandle after reading the correct message
 2. On on accepted connections when there's back pressure

Calling ReadStop makes the handle as in-active which means the loop can
end while there's still valid handles around. This makes loop.Dispose throw
with an EBUSY. To avoid that, we walk all of the handles and dispose them.

#1761

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Kestrel.Transport.Libuv/Internal/LibuvThread.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel.Transport.Libuv/Internal/LibuvThread.cs
@@ -264,6 +264,19 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Transport.Libuv.Internal
                 _post.Reference();
                 _post.Dispose();
 
+                // We need this walk because we call ReadStop on 2 places:
+                // 1. On the dispatch pipe UvPipeHandle after reading the correct message
+                // 2. On on accepted connections when there's back pressure
+                // Calling ReadStop makes the handle as in-active which means the loop can
+                // end while there's still valid handles around. This makes loop.Dispose throw
+                // with an EBUSY. To avoid that, we walk all of the handles and dispose them.
+                Walk(ptr =>
+                {
+                    var handle = UvMemory.FromIntPtr<UvHandle>(ptr);
+                    // handle can be null because UvMemory.FromIntPtr looks up a weak reference
+                    handle?.Dispose();
+                });
+
                 // Ensure the Dispose operations complete in the event loop.
                 _loop.Run();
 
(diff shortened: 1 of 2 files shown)


davidfowl:
This happened on OSX:

```
[xUnit.net 00:00:00.8682320]   Discovering: Microsoft.AspNetCore.Server.Kestrel.FunctionalTests
[xUnit.net 00:00:07.2262650]   Discovered:  Microsoft.AspNetCore.Server.Kestrel.FunctionalTests
[xUnit.net 00:00:07.5893480]   Starting:    Microsoft.AspNetCore.Server.Kestrel.FunctionalTests
The active test run was aborted. Reason: Unable to communicate with test host process.
```

src/Microsoft.AspNetCore.Server.Kestrel.Transport.Libuv/Internal/LibuvThread.cs
```diff
@@ -264,6 +264,19 @@ private void ThreadStart(object parameter)
                 _post.Reference();
                 _post.Dispose();
 
+                // We need this walk because we call ReadStop on 2 places:
+                // 1. On the dispatch pipe UvPipeHandle after reading the correct message
+                // 2. On on accepted connections when there's back pressure
+                // Calling ReadStop makes the handle as in-active which means the loop can
+                // end while there's still valid handles around. This makes loop.Dispose throw
+                // with an EBUSY. To avoid that, we walk all of the handles and dispose them.
+                Walk(ptr =>
```
halter73: You could call OnStopRude() directly here to do the same thing.

src/Microsoft.AspNetCore.Server.Kestrel.Transport.Libuv/Internal/ListenerPrimary.cs
```diff
@@ -18,7 +18,10 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Transport.Libuv.Internal
     /// </summary>
     public class ListenerPrimary : Listener
     {
+        // The list of pipes that can be dispatched to (where we've confirmed the _pipeMessage)
         private readonly List<UvPipeHandle> _dispatchPipes = new List<UvPipeHandle>();
+        // The list of pipes we've created but may not be part of _dispatchPipes
```
halter73: Good catch.

src/Microsoft.AspNetCore.Server.Kestrel.Transport.Libuv/Internal/LibuvThread.cs
```diff
@@ -264,6 +264,19 @@ private void ThreadStart(object parameter)
                 _post.Reference();
                 _post.Dispose();
 
+                // We need this walk because we call ReadStop on 2 places:
+                // 1. On the dispatch pipe UvPipeHandle after reading the correct message
```
halter73: Is the dispatch pipe now always disposed before this point now that you fixed ListenerPrimary?

halter73:
[So you've finally joined the dark side](https://github.com/aspnet/KestrelHttpServer/issues/1761#issuecomment-297113966). Did you, like I did a while back, see uv_tcp_ts getting walked even after uv_run completed? You should have let me change it to walk before the second uv_run back then 😛 Although we probably wouldn't have caught the dispatch pipe issue then.

PR #1804, opened 2017-05-02
Use WebHostBuilder as an implementation detail of TestServer

davidfowl:
- This will make tests run more similarly so we can swap out the
transport more easily.

### DIFF
--- a/test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/TestServer.cs
+++ b/test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/TestServer.cs
@@ -4,13 +4,16 @@
 using System;
 using System.Net;
 using System.Net.Sockets;
-using System.Threading;
+using System.Reflection;
+using Microsoft.AspNetCore.Builder;
+using Microsoft.AspNetCore.Hosting;
+using Microsoft.AspNetCore.Hosting.Server;
 using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.Server.Kestrel.Core;
 using Microsoft.AspNetCore.Server.Kestrel.Transport.Abstractions;
-using Microsoft.AspNetCore.Server.Kestrel.Transport.Libuv;
-using Microsoft.AspNetCore.Server.Kestrel.Transport.Sockets;
 using Microsoft.AspNetCore.Testing;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Options;
 
 namespace Microsoft.AspNetCore.Server.Kestrel.FunctionalTests
@@ -18,12 +21,11 @@ namespace Microsoft.AspNetCore.Server.Kestrel.FunctionalTests
     /// <summary>
     /// Summary description for TestServer
     /// </summary>
-    public class TestServer : IDisposable
+    public class TestServer : IDisposable, IStartup
     {
-        private static TimeSpan _shutdownTimeout = TimeSpan.FromSeconds(5);
-
-        private KestrelServer _server;
+        private IWebHost _host;
         private ListenOptions _listenOptions;
+        private readonly RequestDelegate _app;
 
         public TestServer(RequestDelegate app)
             : this(app, new TestServiceContext())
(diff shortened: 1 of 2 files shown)


test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/TestServer.cs
```diff
@@ -47,63 +49,68 @@ public TestServer(RequestDelegate app, TestServiceContext context, ListenOptions
...
         public TestServiceContext Context { get; }
 
+        void IStartup.Configure(IApplicationBuilder app)
+        {
+            app.Run(_app);
+        }
+
+        IServiceProvider IStartup.ConfigureServices(IServiceCollection services)
+        {
+            // Unfortunately, this needs to be replaced in IStartup.ConfigureServices
```
halter73: Why?

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/TestServer.cs
```diff
@@ -47,63 +49,68 @@ public TestServer(RequestDelegate app, TestServiceContext context, ListenOptions
...
+                             // We're doing this so we can use the same instance that was passed in
+                             var configureOptions = sp.GetServices<IConfigureOptions<KestrelServerOptions>>();
+                             foreach (var c in configureOptions)
+                             {
+                                 c.Configure(context.ServerOptions);
+                             }
+                             return new KestrelServer(sp.GetRequiredService<ITransportFactory>(), context);
+                         });
+                     })
+                     .UseSetting(WebHostDefaults.ApplicationKey, typeof(TestServer).GetTypeInfo().Assembly.FullName)
```
halter73: What happens if this isn't here?

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/TestServer.cs
```diff
@@ -47,63 +49,68 @@ public TestServer(RequestDelegate app, TestServiceContext context, ListenOptions
...
-                new LifetimeNotImplemented(),
-                new KestrelTestLoggerFactory(new TestApplicationErrorLogger()));
-
-            return transportFactory;
+            _host = new WebHostBuilder()
+                     .UseKestrel(o =>
+                     {
+                         o.ListenOptions.Add(_listenOptions);
+                     })
+                     .ConfigureServices(services =>
```
halter73: Does this get called before or after IStartup.ConfigureServices? Why do we need both?

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/TestServer.cs
```diff
@@ -47,63 +49,68 @@ public TestServer(RequestDelegate app, TestServiceContext context, ListenOptions
...
+                         services.AddSingleton<IStartup>(this);
+                         services.AddSingleton<ILoggerFactory>(new KestrelTestLoggerFactory(context.ErrorLogger));
+                         services.AddSingleton<IServer>(sp =>
+                         {
+                             // Manually configure options on the TestServiceContext.
+                             // We're doing this so we can use the same instance that was passed in
+                             var configureOptions = sp.GetServices<IConfigureOptions<KestrelServerOptions>>();
+                             foreach (var c in configureOptions)
+                             {
+                                 c.Configure(context.ServerOptions);
```
halter73: `o.ListenOptions.Add(_listenOptions);` is the only thing configured, right? Why not just call `context.ServerOptions.ListenOptions.Add(_listenOptions);` here?

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/TestServer.cs
davidfowl: null ref trying to determine the app name. I need to file an issue in hosting for this.

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/TestServer.cs
davidfowl: Because in the future if we add more stuff that configures options, it'll just work.

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/TestServer.cs
davidfowl: Before, IStartup.ConfigureServices and yes you need both. You need to replace some HostingServices and you need to replace the `IApplicationLifetime` in IStartup.ConfigureServices because that's where it's added (we should have called TryAdd in hosting).

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/TestServer.cs
davidfowl: See the comment below.

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/TestServer.cs
davidfowl: /cc @anurse  (fallout from removal of PlatformAbstractions)

PR #1805, opened 2017-05-02
[WIP]: Make sockets the default

davidfowl:


samples/SystemdSample/Program.cs
```diff
@@ -0,0 +1,29 @@
+﻿using Microsoft.AspNetCore.Builder;
```
cesarbs: Copyright.

PR #1806, opened 2017-05-02
Added a ConnectionAbortedException to Transport.Abstractions

davidfowl:
- To avoid hard coding TaskCanceledException in each transport
- This PR tries to keep compatibility by converting the ConnectionAbortedException
to a TaskCanceledException on exceptions in FrameRequestStream. The downside is that
this conversion causes an async state machine to be created per call to ReadAsync.
CopyToAsync isn't that bad because it's a single long running task.

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/Http/FrameRequestStream.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/Http/FrameRequestStream.cs
@@ -104,16 +104,22 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
 #error target frameworks need to be updated
 #endif
 
-        public override async Task<int> ReadAsync(byte[] buffer, int offset, int count, CancellationToken cancellationToken)
+        public override Task<int> ReadAsync(byte[] buffer, int offset, int count, CancellationToken cancellationToken)
+        {
+            var task = ValidateState(cancellationToken);
+            if (task != null)
+            {
+                return task;
+            }
+
+            return ReadAsyncInternal(buffer, offset, count, cancellationToken);
+        }
+
+        private async Task<int> ReadAsyncInternal(byte[] buffer, int offset, int count, CancellationToken cancellationToken)
         {
             try
             {
-                var task = ValidateState(cancellationToken);
-                if (task == null)
-                {
-                    return await _body.ReadAsync(new ArraySegment<byte>(buffer, offset, count), cancellationToken);
-                }
-                return await task;
+                return await _body.ReadAsync(new ArraySegment<byte>(buffer, offset, count), cancellationToken);
             }
             catch (ConnectionAbortedException)
             {
@@ -123,7 +129,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
             }
         }
 
-        public override async Task CopyToAsync(Stream destination, int bufferSize, CancellationToken cancellationToken)
+        public override Task CopyToAsync(Stream destination, int bufferSize, CancellationToken cancellationToken)
         {
             if (destination == null)
             {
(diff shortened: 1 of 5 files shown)


src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/Http/FrameRequestStream.cs
```diff
@@ -105,11 +107,26 @@ private Task<int> ReadAsync(byte[] buffer, int offset, int count, CancellationTo
...
+
+        private async Task<int> ReadAsyncInternal(byte[] buffer, int offset, int count, CancellationToken cancellationToken)
+        {
+            try
+            {
+                return await _body.ReadAsync(new ArraySegment<byte>(buffer, offset, count), cancellationToken);
+            }
+            catch (ConnectionAbortedException)
+            {
+                ExceptionDispatchInfo.Capture(new TaskCanceledException("The request was aborted")).Throw();
```
halter73: `ExceptionDispatchInfo.Capture` makes sense for preserving the original stack trace, bus since there isn't one why not just `throw new TaskCanceledException("The request was aborted")`? This way you can remove the deceptive `return 0`.

src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/Http/FrameRequestStream.cs
```diff
@@ -105,11 +107,26 @@ private Task<int> ReadAsync(byte[] buffer, int offset, int count, CancellationTo
...
+            return ReadAsyncInternal(buffer, offset, count, cancellationToken);
+        }
+
+        private async Task<int> ReadAsyncInternal(byte[] buffer, int offset, int count, CancellationToken cancellationToken)
+        {
+            try
+            {
+                return await _body.ReadAsync(new ArraySegment<byte>(buffer, offset, count), cancellationToken);
+            }
+            catch (ConnectionAbortedException)
```
halter73: Can we just make ConnectionAbortedException inherit from TaskCanceledException? That way it will still also be an OperationCanceledException, and the catch/rethrow will be unnecessary.

halter73: If we do this, we technically wouldn't be taking advantage of ConnectionAbortedException being it's own exception type, but I think that's fine.

davidfowl:
> Is this fixing any functional issues btw?

No, it's just cleaning up the layering.

src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/Http/FrameRequestStream.cs
davidfowl: > Can we just make ConnectionAbortedException inherit from TaskCanceledException? That way it will still also be an OperationCanceledException, and the catch/rethrow will be unnecessary.
  
I almost did that but it's wrong. There's no task involved so it wouldn't be right to derive from that exception. Doing it at the point where we're using tasks makes more sense.

src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/Http/FrameRequestStream.cs
davidfowl: I think I meant to throw TaskCancelled with an inner exception, dunno why I ended up with this.

src/Microsoft.AspNetCore.Server.Kestrel.Transport.Abstractions/Exceptions/ConnectionAbortedException.cs
```diff
@@ -0,0 +1,21 @@
+﻿using System;
+
+namespace Microsoft.AspNetCore.Server.Kestrel.Transport.Abstractions
+{
+    public class ConnectionAbortedException : OperationCanceledException
+    {
+        public ConnectionAbortedException() :
+            this("The connection was aborted")
```
halter73: Why aren't you using CoreStrings?

davidfowl: Because it's in Core only and we don't cross reference resources from other assemblies. I could add a resource file to transport abstractions but I'm not going to because it's pubternal and not being localized 😝

Issue #1808, opened 2017-05-04
Investigate removing uv_shutdown call before closing a conneciton


halter73:
It appears that if a FIN might be sent with just a uv_close. At least on Windows this [seems to be the case](https://github.com/libuv/libuv/blob/cb0cc7346d1c2b5bdf2bd89e4006a50fb6f9c3dd/src/win/tcp.c#L1380).

davidfowl:
Fixed in https://github.com/aspnet/KestrelHttpServer/commit/cd1568d7f4c90a1079b8412a7db6cbbf2bc59de4

PR #1811, opened 2017-05-05
Remove uv_shutdown because FIN is already sent during uv_close

davidfowl:
I also ran a manual test with a linux VM and windows client with the following code:

Client

```C#
Socket socket = new Socket(SocketType.Stream, ProtocolType.Tcp);
socket.Connect(new IPEndPoint(IPAddress.Parse(<IP>), 5000));
var request = Encoding.ASCII.GetBytes("GET / HTTP/1.1\r\nHost:\r\n\r\n");
socket.Send(request);
var result = socket.Receive(new byte[32]);
Console.WriteLine(result);
```

Server

```C#
app.Run(context =>
{
    context.Abort();
    return Task.CompletedTask;
});
```

It prints out 0

Also confirmed it in Wireshark:

image

#1808


cesarbs:
Add a unit test.

src/Microsoft.AspNetCore.Server.Kestrel.Transport.Libuv/Internal/LibuvOutputConsumer.cs
```diff
@@ -67,18 +67,6 @@ public class LibuvOutputConsumer
 
                     if (result.IsCancelled)
                     {
-                        // Send a FIN
-                        _log.ConnectionWriteFin(_connectionId);
-
-                        using (var shutdownReq = new UvShutdownReq(_log))
```
cesarbs: I think this is the only place where we use `UvShutdownReq`. Can we remove the type altogether? Or do you prefer to keep it around?

davidfowl: I prefer to keep it around,

cesarbs:
:rage1:

davidfowl:
:rage4:

cesarbs:
🚢 when green.

Issue #1815, opened 2017-05-06
It's possible to lease blocks from the pool after the pool has been disposed


davidfowl:
When using connection adapters, it's possible to leak memory on shutdown if the adapted pipeline runs after the libuv thread is stopped. Generally though, this shouldn't happen if the shutdown is graceful.

davidfowl:
Here's what a typical trace looks like:

```
Hosting starting
Hosting started
Loaded hosting startup assembly Microsoft.AspNetCore.Server.Kestrel.FunctionalTests, Version=2.0.0.0, Culture=neutral, PublicKeyToken=adb9793829ddae60
Connection id "0HL4L3C6U9KGO" started.
Applying connection adapters
before Microsoft.AspNetCore.Server.Kestrel.FunctionalTests.ConnectionAdapterTests+AsyncConnectionAdapter.ApplyConnectionAdaptersAsync
Connection id "0HL4L3C6U9KGO" received FIN.
Connection id "0HL4L3C6U9KGO" disconnecting.
Connection id "0HL4L3C6U9KGO" stopped.
Connection id "0HL4L3C6U9KGO" sending FIN.
Hosting shutdown
Disposing everything (this runs after the libuv thread ends)
after Microsoft.AspNetCore.Server.Kestrel.FunctionalTests.ConnectionAdapterTests+AsyncConnectionAdapter.ApplyConnectionAdaptersAsync
Make the adapted pipeline
before adaptedPipeline.RunAsync()
```

davidfowl:
It's because it's possible to start the shutdown sequence before the frame is added to the list of connections.

PR #1816, opened 2017-05-07
Refactoring and of FrameConnection and Frame

davidfowl:
- Building on top of the last refactoring of FrameConnection, this change aims to clean up
the communication between the Frame and FrameConnection by removing some concepts and
being consistent about the communication between Frame and FrameConnection with or without
connection adapters. Changes include:
- Removing ConnectionLifetimeControl, ISocketOutput, StreamSocketOutput
- Moving more initialization of the frame to FrameConnection after the pipes
are setup
- OutputProducer communicates cancellation via the IPipeWriter instead of the output's IPipeReader.
- Frame always communicates via the pipes and that communications flows through the layers to the transport.
This means that each 1/2 of the adapted pipeline handles closing the right side of the transport at the
right time, propagating exceptions as necessary.
- This is how the flow looks now:
```
            ->                        ->
[transport]     [connection adapters]     [frame]
            <-                        <-
```
- Transports need to handle a ConnectionAbortedException on the output as a signal to stop
writing and end the connection. This will no longer try to drain the output but will just stop
writing and end the response immediately.

#1815


test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/RequestTests.cs
```diff
@@ -1051,24 +1051,28 @@ public void AbortingTheConnectionSendsFIN()
             {
                 using (var connection = server.CreateConnection())
                 {
-                    // Never send the body so CopyToAsync always fails.
+                    // Full request and response
```
davidfowl: I had to change this because we don't necessarily drain all queued writes before aborting for a FIN.

src/Microsoft.AspNetCore.Server.Kestrel.Core/Adapter/Internal/AdaptedPipeline.cs
```diff
@@ -7,73 +7,131 @@
...
+                        if (buffer.IsEmpty && readResult.IsCompleted)
+                        {
+                            break;
+                        }
 
-                _output.Dispose();
+                        if (buffer.IsEmpty)
+                        {
+                            await _stream.FlushAsync();
+                        }
```
NickCraver: Since this is a hot path, optimize the common double check?
```c#
if (buffer.IsEmpty)
{
    if (readResult.IsCompleted)
    {
        break;
    }
    await _stream.FlushAsync();
}
```

src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/FrameConnection.cs
```diff
@@ -69,114 +68,141 @@ public void StartRequestProcessing()
...
-                    try
+                    if (adaptedPipeline == null)
                     {
-                        var adaptedPipeline = await ApplyConnectionAdaptersAsync(rawStream);
-
-                        _frame.Input = adaptedPipeline.Input.Reader;
-                        _frame.Output = adaptedPipeline.Output;
+                        // We need to assign these anyways because frame.Abort calls via OutputProducer.Abort
+                        _frame.Input = input;
+                        _frame.Output = new OutputProducer(output, _frame, ConnectionId, Log);
```
cesarbs: You instantiate `OutputProducer` in the two places were you assign it - create it before the `if` and assign the same instance in both places.

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/ConnectionAdapterTests.cs
```diff
@@ -77,6 +78,70 @@ public class ConnectionAdapterTests
         }
 
         [Fact]
+        public async Task ImmediateFinAfterOnConnectionAsync()
```
cesarbs: What is the expected outcome of this test?

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/ConnectionAdapterTests.cs
```diff
@@ -77,6 +78,70 @@ public class ConnectionAdapterTests
...
+                {
+                    // FIN
+                    connection.Shutdown(SocketShutdown.Send);
+                    await connection.WaitForConnectionClose();
+                }
+            }
+        }
+
+        [Fact]
+        public async Task ImmediateShutdownAfterOnConnectionAsync()
```
cesarbs: What is the expected outcome of this test?

src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/Http/OutputProducer.cs
```diff
@@ -42,7 +43,83 @@ public OutputProducer(IPipeWriter pipe, Frame frame, string connectionId, IKestr
...
+        public void Write(ArraySegment<byte> buffer, bool chunk = false)
+        {
+            WriteAsync(buffer, default(CancellationToken), chunk).GetAwaiter().GetResult();
+        }
+
+        public Task WriteAsync(ArraySegment<byte> buffer, bool chunk = false, CancellationToken cancellationToken = default(CancellationToken))
+        {
+            if (cancellationToken.IsCancellationRequested)
+            {
+                _frame.Abort(error: null);
```
cesarbs: I understand this was here already, but it doesn't seem right to abort the entire thing because a write is cancelled. Since this is under user control, someone could be using a ct to decide not to write at a given point if the token is cancelled, but would still want to continue running the app and doing further writes. TBH I don't know if this is realistic, but it's easily doable.

cesarbs: Discussed offline and this is fine.

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/ConnectionAdapterTests.cs
davidfowl: That it does't throw or crash. Unfortunately that's hard to assert but I can change the test name.

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/ConnectionAdapterTests.cs
davidfowl: *DoesNotCrash

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/ConnectionAdapterTests.cs
cesarbs: Yeah, put it in the test name.

PR #1820, opened 2017-05-09
More FrameConnection refactoring

davidfowl:
- This change reverts the change to complete the writer with an
exception on abort because of the number of first chance exceptions
that get thrown.
- This change also moves connection logging into FrameConnection instead
of being split between the ConnectionHandler and FrameConnection.
- Fixed issues with LibuvOutputConsumerTests that leak WriteReq since
cancelled writes no longer end the connection.

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Kestrel.Core/Adapter/Internal/AdaptedPipeline.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel.Core/Adapter/Internal/AdaptedPipeline.cs
@@ -71,11 +71,6 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Adapter.Internal
 
                         if (buffer.IsEmpty)
                         {
-                            if (result.IsCompleted)
-                            {
-                                break;
-                            }
-
                             await stream.FlushAsync();
                         }
                         else if (buffer.IsSingleSpan)
@@ -90,6 +85,11 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Adapter.Internal
                                 var array = memory.GetArray();
                                 await stream.WriteAsync(array.Array, array.Offset, array.Count);
                             }
+                        }
+
+                        if (result.IsCompleted)
+                        {
+                            break;
                         }
                     }
                     finally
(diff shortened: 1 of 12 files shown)


src/Microsoft.AspNetCore.Server.Kestrel.Core/Adapter/Internal/AdaptedPipeline.cs
```diff
@@ -58,18 +57,25 @@ public class AdaptedPipeline
...
+                        if (result.IsCancelled)
                         {
+                            // Forward the cancellation to the transport pipe
+                            _transportOutputPipe.Reader.CancelPendingRead();
                             break;
                         }
 
                         if (buffer.IsEmpty)
                         {
+                            if (result.IsCompleted)
```
halter73: Any particular reason not to flush in this case?

src/Microsoft.AspNetCore.Server.Kestrel.Core/Adapter/Internal/AdaptedPipeline.cs
```diff
@@ -111,8 +117,8 @@ public class AdaptedPipeline
             {
                 if (stream == null)
                 {
-                    // If the stream is null then we're going to abort the connection
-                    throw new ConnectionAbortedException();
+                    // REVIEW: Do we need an exception here?
+                    return;
```
halter73: Instead of throwing here, it should be the responsibility of ReadInputAsync's caller to not call it with an empty stream. That means the caller would also need to call `Input.Writer.Complete(Exception)` and `_transportInputPipeReader.Complete();`.

src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/FrameConnection.cs
```diff
@@ -114,6 +117,9 @@ public void StartRequestProcessing()
             {
                 _context.ServiceContext.ConnectionManager.RemoveConnection(_context.FrameConnectionId);
                 DisposeAdaptedConnections();
+
+                Log.ConnectionStop(ConnectionId);
+                KestrelEventSource.Log.ConnectionStop(this);
```
halter73: I love putting the start/stop logging in one method. 🎉

src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/Http/OutputProducer.cs
```diff
@@ -112,7 +105,7 @@ public void Abort()
 
                 _log.ConnectionDisconnect(_connectionId);
                 _completed = true;
-                _pipe.Complete(new ConnectionAbortedException());
```
halter73: Ewwww

src/Microsoft.AspNetCore.Server.Kestrel.Transport.Sockets/SocketConnection.cs
```diff
@@ -207,6 +204,8 @@ private void SetupSendBuffers(ReadableBuffer buffer)
                         _output.Advance(buffer.End);
                     }
                 }
+
+                _socket.Shutdown(SocketShutdown.Send);
```
halter73: Do managed sockets send a FIN implicitly like libuv if you close the socket without calling Shutdown?

PR #1821, opened 2017-05-09
Add support to use System.Memory as a package instead of internilized source

pakrym:
https://github.com/aspnet/CoreCLR/issues/179

### DIFF
--- a/build/dependencies.props
+++ b/build/dependencies.props
@@ -1,6 +1,7 @@
 <Project>
   <PropertyGroup>
     <AspNetCoreVersion>2.0.0-*</AspNetCoreVersion>
+    <SystemMemoryVersion>4.4.0-*</SystemMemoryVersion>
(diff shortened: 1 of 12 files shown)


natemcmaster:
Looks like this needs a little more work.
```
error CS0104: 'Span<>' is an ambiguous reference between 'System.Span<T>' and 'Microsoft.AspNetCore.Server.Kestrel.Internal.System.Span<T>' 
```

pakrym:
https://github.com/aspnet/CoreCLR/pull/206 is not merged yet so the source package we have on the feed is not updated hence CI failures.

halter73:
Anyone who references System.Memory will now need to add `using System;` and `using Microsoft.AspNetCore.Server.Kestrel.Internal.System;` It's going to be nearly impossible to keep this compiling both with and without System.Memory as a package going forward.

AFAIK, `Ctrl+.`, `Alt+Enter`, "Remove and sort usings" all break. Is there a way we can enforce that the correct "usings" are used?

PR #1822, opened 2017-05-10
Invert the dependency between connection adapters and Frame

davidfowl:
- Removed PrepareRequest from IAdaptedConnection and instead added
a feature collection to the ConnectionAdapterContext. This allows features to be set
once by the adapter instead of per request. It's the Frame's job to copy features
from the connection level feature collection into the per request feature collection.
- Set the scheme to "https" based on the presence of ITlsConnectionFeature.
- Always set ITlsConnection feature if the HttpsAdaptedConnection doesn't throw during
the handshake

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Kestrel.Core/Adapter/Internal/ConnectionAdapterContext.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel.Core/Adapter/Internal/ConnectionAdapterContext.cs
@@ -2,6 +2,7 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System.IO;
+using Microsoft.AspNetCore.Http.Features;
 
 namespace Microsoft.AspNetCore.Server.Kestrel.Core.Adapter.Internal
 {
@@ -9,10 +10,13 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Adapter.Internal
     // we want to add more connection metadata later.
     public class ConnectionAdapterContext
     {
-        internal ConnectionAdapterContext(Stream connectionStream)
+        internal ConnectionAdapterContext(IFeatureCollection features, Stream connectionStream)
         {
+            Features = features;
             ConnectionStream = connectionStream;
         }
+
+        public IFeatureCollection Features { get; }
 
         public Stream ConnectionStream { get; }
     }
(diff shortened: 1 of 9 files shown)


src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/FrameConnection.cs
```diff
@@ -159,17 +159,19 @@ public void Timeout()
 
         private async Task<Stream> ApplyConnectionAdaptersAsync()
         {
+            var features = new FeatureCollection();
```
davidfowl: Eventually this will flow all the way from the transport.

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/HttpsConnectionAdapterTests.cs
```diff
@@ -94,7 +94,9 @@ public class HttpsConnectionAdapterTests
 
             using (var server = new TestServer(context =>
                 {
-                    Assert.Equal(context.Features.Get<ITlsConnectionFeature>(), null);
+                    var tlsFeature = context.Features.Get<ITlsConnectionFeature>();
+                    Assert.NotNull(tlsFeature);
+                    Assert.Equal(tlsFeature.ClientCertificate, null);
```
cesarbs: Flip args.

src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/FrameConnection.cs
```diff
@@ -159,17 +159,19 @@ public void Timeout()
...
+            var connectionAdapters = _context.ConnectionAdapters;
             var stream = new RawStream(_context.Input.Reader, _context.Output.Writer);
-            var adapterContext = new ConnectionAdapterContext(stream);
-            var adaptedConnections = new IAdaptedConnection[_connectionAdapters.Count];
+            var adapterContext = new ConnectionAdapterContext(features, stream);
+            _adaptedConnections = new List<IAdaptedConnection>(connectionAdapters.Count);
```
halter73: Why is this now a List?

src/Microsoft.AspNetCore.Server.Kestrel.Core/Adapter/Internal/ConnectionAdapterContext.cs
```diff
@@ -2,18 +2,22 @@
...
     public class ConnectionAdapterContext
     {
-        internal ConnectionAdapterContext(Stream connectionStream)
+        internal ConnectionAdapterContext(IFeatureCollection features, Stream connectionStream)
         {
+            Features = features;
             ConnectionStream = connectionStream;
         }
 
+        public IFeatureCollection Features { get; }
```
halter73: Do you plan to adding something like this to IConnectionAdapter and removing IsHttps?

davidfowl: Yes, that's the plan.

Issue #1824, opened 2017-05-10
LibuvConnection.OnRead throws Nullable object must have a value


tmds:
To make the wrk tool generate large network requests, I set the number of pipelined requests to 2048. This causes an exception in LibuvConnection.OnRead which terminates the webserver.

"Microsoft.AspNetCore.Server.Kestrel.Transport.Libuv": "2.0.0-preview2-24956"

To reproduce:

csproj
```xml
<Project Sdk="Microsoft.NET.Sdk.Web">
  <PropertyGroup>
    <TargetFramework>netcoreapp2.0</TargetFramework>
    <DebugType>portable</DebugType>
    <OutputType>Exe</OutputType>
  </PropertyGroup>
  <ItemGroup>
    <PackageReference Include="Microsoft.AspNetCore.Server.Kestrel" Version="2.0.0-*" />
    <PackageReference Include="Microsoft.AspNetCore.Server.Kestrel.Transport.Libuv" Version="2.0.0-*" />
  </ItemGroup>
</Project>
```

Startup.cs
```c#
using System;
using System.Net;
using Microsoft.AspNetCore.Builder;
using Microsoft.AspNetCore.Hosting;
using Microsoft.AspNetCore.Http;

namespace SampleApp
{
    public class Startup
    {
        public void Configure(IApplicationBuilder app)
        {
            app.Run(async context =>
            {
                var response = $"hello, world{Environment.NewLine}";
                context.Response.ContentLength = response.Length;
                context.Response.ContentType = "text/plain";
                await context.Response.WriteAsync(response);
            });
        }

        public static void Main(string[] args)
        {
            var hostBuilder = new WebHostBuilder()
                .UseKestrel(options =>
                {
     		    options.Listen(IPAddress.Parse("1.1.1.2"), 5000, listenOptions => {} );
                })
                .UseStartup<Startup>();
            hostBuilder = hostBuilder.UseLibuv(options => options.ThreadCount = 1);

            var host = hostBuilder.Build();
            host.Run();
        }
    }
}
```

NuGet.config
```xml
<?xml version="1.0" encoding="utf-8"?>
<configuration>
  <packageSources>
    <add key="dotnet-core" value="https://dotnet.myget.org/F/dotnet-core/api/v3/index.json"  />
    <add key="dotnet-corefxlab" value="https://dotnet.myget.org/F/dotnet-corefxlab/" />    
    <add key="nuget.org" value="https://api.nuget.org/v3/index.json"  />
[...]

PR #1828, opened 2017-05-11
Simplify LibuvConnection.OnRead()

halter73:
- Fix a null reference that sometimes occurs given an EOF status

#1824

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Kestrel.Transport.Libuv/Internal/LibuvConnection.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel.Transport.Libuv/Internal/LibuvConnection.cs
@@ -128,70 +128,57 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Transport.Libuv.Internal
 
         private void OnRead(UvStreamHandle handle, int status)
         {
-            var normalRead = status >= 0;
-            var normalDone = status == LibuvConstants.EOF;
-            var errorDone = !(normalDone || normalRead);
-            var readCount = normalRead ? status : 0;
-
-            if (normalRead)
+            if (status >= 0)
             {
-                Log.ConnectionRead(ConnectionId, readCount);
+                Log.ConnectionRead(ConnectionId, status);
+
+                Debug.Assert(_currentWritableBuffer != null);
+                var currentWritableBuffer = _currentWritableBuffer.Value;
+                currentWritableBuffer.Advance(status);
+                var flushTask = currentWritableBuffer.FlushAsync();
+
+                if (!flushTask.IsCompleted)
+                {
+                    // We wrote too many bytes too the reader so pause reading and resume when
+                    // we hit the low water mark
+                    _ = ApplyBackpressureAsync(flushTask);
+                }
             }
             else
             {
                 _socket.ReadStop();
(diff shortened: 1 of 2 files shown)


src/Microsoft.AspNetCore.Server.Kestrel.Transport.Libuv/Internal/LibuvConnection.cs
```diff
@@ -128,70 +128,57 @@ private static void ReadCallback(UvStreamHandle handle, int status, object state
 
         private void OnRead(UvStreamHandle handle, int status)
         {
-            var normalRead = status >= 0;
-            var normalDone = status == LibuvConstants.EOF;
-            var errorDone = !(normalDone || normalRead);
-            var readCount = normalRead ? status : 0;
-
-            if (normalRead)
+            if (status >= 0)
```
davidfowl: Should we have:

```
if (status == 0)
{
    _currentWritableBuffer?.Commit();
}
```

halter73: Should be functionally equivalent, no? Before this change, Advance(0) and FlushAsync() were already being called in this case.

davidfowl: It its but I think we should just avoid signaling the reader if nothing was written.

halter73: But who will think of the cyclomatic complexity 😉

src/Microsoft.AspNetCore.Server.Kestrel.Transport.Libuv/Internal/LibuvConnection.cs
```diff
@@ -128,70 +128,57 @@ private static void ReadCallback(UvStreamHandle handle, int status, object state
...
-                if (normalDone)
+                Debug.Assert(_currentWritableBuffer != null);
+                var currentWritableBuffer = _currentWritableBuffer.Value;
+                currentWritableBuffer.Advance(status);
+                var flushTask = currentWritableBuffer.FlushAsync();
+
+                if (!flushTask.IsCompleted)
                 {
-                    Log.ConnectionReadFin(ConnectionId);
+                    // We wrote too many bytes too the reader so pause reading and resume when
```
cesarbs: Since you're moving this, might as well fix the typo: the second "too" should be "to" 😄

src/Microsoft.AspNetCore.Server.Kestrel.Transport.Libuv/Internal/LibuvConnection.cs
davidfowl: Can OnRead with 0 be called without an OnAlloc call

halter73: No. The purpose for calling OnRead with 0 in the first place is to return the OnAlloc buffer.

http://docs.libuv.org/en/v1.x/stream.html#c.uv_read_cb

davidfowl:
We should add more tests here in general.

Issue #1827, opened 2017-05-11
Kestrel fails to shutdown when consuming request body with inline scheduler


cesarbs:
Repro:

- Run SampleApp
- Use wrk to send a bunch of request data (get [Benchmarks](https://github.com/aspnet/benchmarks) and run `wrk -c 256 -t 32 http://<address>:5000 -s scripts/post.lua -- 32`)
- Ctrl+C

Fails to shutdown:

```
> dotnet .\bin\Debug\netcoreapp2.0\SampleApp.dll
Hosting environment: Production
Content root path: D:\src\aspnet\KestrelHttpServer\samples\SampleApp
Now listening on: http://0.0.0.0:5000
Application started. Press Ctrl+C to shut down.
Application is shutting down...
fail: Microsoft.AspNetCore.Server.Kestrel.Transport.Libuv[0]
      Disposing listeners failed
crit: Microsoft.AspNetCore.Server.Kestrel.Transport.Libuv[0]
      LibuvThread.StopAsync failed to terminate libuv thread.
crit: Microsoft.AspNetCore.Server.Kestrel.Transport.Libuv[0]
      LibuvThread.StopAsync failed to terminate libuv thread.
crit: Microsoft.AspNetCore.Server.Kestrel.Transport.Libuv[0]
      LibuvThread.StopAsync failed to terminate libuv thread.
crit: Microsoft.AspNetCore.Server.Kestrel.Transport.Libuv[0]
      LibuvThread.StopAsync failed to terminate libuv thread.
crit: Microsoft.AspNetCore.Server.Kestrel.Transport.Libuv[0]
      LibuvThread.StopAsync failed to terminate libuv thread.
```

It just hangs there.

Stacks:

![image](https://cloud.githubusercontent.com/assets/924164/25975609/32b83246-3664-11e7-9762-82b97a0745f0.png)

Code is stuck in a loop here:

https://github.com/aspnet/KestrelHttpServer/blob/53b0eea2ec2518725b49d52e7a6f517023afe990/src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/Http/PipelineExtensions.cs#L53

`result` always looks like this:

![image](https://cloud.githubusercontent.com/assets/924164/25975643/729dfcd8-3664-11e7-8342-07e94fbb581c.png)

i.e. not completed, not cancelled and the buffer is empty:

[...]

PR #1829, opened 2017-05-12
Prevent infinite loop in PipelineExtensions.PeekAsyncAwaited() (#1827)

cesarbs:
#1827

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/Http/PipelineExtensions.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/Http/PipelineExtensions.cs
@@ -40,7 +40,9 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
                 }
                 finally
                 {
-                    pipelineReader.Advance(result.Buffer.Start, result.Buffer.Start);
+                    pipelineReader.Advance(result.Buffer.Start, result.Buffer.IsEmpty
+                        ? result.Buffer.End
+                        : result.Buffer.Start);
                 }
                 input = pipelineReader.ReadAsync();
             }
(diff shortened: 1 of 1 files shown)


davidfowl:
Make sure you file an issue on pipes in corefxlab

cesarbs:
@davidfowl https://github.com/dotnet/corefxlab/issues/1540

cesarbs:
@davidfowl Can you take another look? I realized I should have done the same change to the "non-awaited"  method too.

src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/Http/PipelineExtensions.cs
```diff
@@ -68,7 +70,9 @@ public static ValueTask<ArraySegment<byte>> PeekAsync(this IPipeReader pipelineR
                 }
                 finally
                 {
-                    pipelineReader.Advance(result.Buffer.Start, result.Buffer.Start);
+                    pipelineReader.Advance(result.Buffer.Start, result.Buffer.IsEmpty
+                        ? result.Buffer.End
+                        : result.Buffer.Start);
                 }
 
                 readingTask = pipelineReader.ReadAsync();
```
halter73: Wouldn't it be simpler to just put `pipelineReader.Advance(result.Buffer.End, result.Buffer.End)` right before this line? No need for any extra IsEmpty checks that way. If PeekAsync returns, ConsumedBytes should call Advance anyway, right?

cesarbs: Passing `End` to both `consumed` and `examined` would break the idea of "peek" though.

`ConsumedBytes` calls `Advance`, but if we consume here then that call would double consume bytes from the input.

cesarbs: I think I misunderstood you on my first read.

But `ConsumedBytes` calls `ReadAsync` as well, so this call has to call `Advance` before returning otherwise that `ReadAsync` will fail.

halter73: Nevermind. I missed that ConsumedBytes also calls ReadAsync.

PR #1830, opened 2017-05-12
Remove rogue ConfigureAwait

davidfowl:
- We don't do it anywhere else, I assume this is left over
from older legacy

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/Http/FrameOfT.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/Http/FrameOfT.cs
@@ -100,7 +100,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
                             {
                                 KestrelEventSource.Log.RequestStart(this);
 
-                                await _application.ProcessRequestAsync(context).ConfigureAwait(false);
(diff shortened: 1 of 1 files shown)


benaadams:
Is because it's calling user code and who knows what they may do?

davidfowl:
@benaadams so does https://github.com/aspnet/KestrelHttpServer/blob/557c04c93295d7c6965618eaa66ed9725343af0f/src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/Http/FrameOfT.cs#L129 and https://github.com/aspnet/KestrelHttpServer/blob/557c04c93295d7c6965618eaa66ed9725343af0f/src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/Http/FrameOfT.cs#L136

benaadams:
That's all I've got :)

benaadams:
Guessing it doesn't matter as its the outer await that calls user code; rather than an inner one user code calls. Bit early an not had coffee yet :)

davidfowl:
> Actually can you still specify the scheduler stuff gets run on? Might be more important now?
Yes, it'll be fun to see what people come up with (by shooting themselves in the foot 😄 ). We need to make sure everything runs properly when you run on IO threads if you don't do anything wrong. 

@cesarbs already found a problem https://github.com/aspnet/KestrelHttpServer/pull/1829

PR #1838, opened 2017-05-15
Remove synchronous Write APIs on IFrameControl

davidfowl:
Remove sync write APIs. There were already doing sync over async so there's no real behavior change here. I was lazy and didn't make ProduceContinue async because it would conflict with @cesarbs's body changes.

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/Http/OutputProducer.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/Http/OutputProducer.cs
@@ -39,11 +39,6 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
             _flushCompleted = OnFlushCompleted;
         }
 
-        public void Write(ArraySegment<byte> buffer, bool chunk = false)
-        {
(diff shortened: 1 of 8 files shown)


src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/Http/OutputProducer.cs
```diff
@@ -39,11 +39,6 @@ public OutputProducer(IPipe pipe, string connectionId, IKestrelTrace log)
             _flushCompleted = OnFlushCompleted;
         }
 
-        public void Write(ArraySegment<byte> buffer, bool chunk = false)
```
halter73: You missed `Write<T>(...)`! I'm kidding ofc.

test/Microsoft.AspNetCore.Server.Kestrel.Core.Tests/FrameTests.cs
```diff
@@ -493,14 +493,14 @@ public void WriteThrowsForNonBodyResponse()
         }
 
         [Fact]
-        public void WriteDoesNotThrowForHeadResponse()
+        public async Task WriteDoesNotThrowForHeadResponse()
```
cesarbs: nit: put `WriteAsync` in test name

PR #1773, opened 2017-04-26
Add request body reader

cesarbs:
Will be used in a subsequent change to add a request body timeout/minimum data rate.

Issue #1270, opened 2016-12-22
ListenerPrimaryTests.NonListenerPipeConnectionsAreLoggedAndIgnored test failure on OS X on Travis


cesarbs:
https://travis-ci.org/aspnet/KestrelHttpServer/jobs/185896490

halter73:
The link now points to a successful build. Any other details about the failure?

cesarbs:
Saw it again on #1840, but I don't think it's related.

```
[xUnit.net 00:00:02.5756350]     NonListenerPipeConnectionsAreLoggedAndIgnored [FAIL]
[xUnit.net 00:00:02.5770670]       System.IO.IOException : Unable to transfer data on the transport connection: Connection reset by peer.
[xUnit.net 00:00:02.5771710]       ---- System.Net.Sockets.SocketException : Connection reset by peer
[xUnit.net 00:00:02.5783520]       Stack Trace:
[xUnit.net 00:00:02.5802200]            at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
[xUnit.net 00:00:02.5803200]            at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
[xUnit.net 00:00:02.5804090]            at System.Runtime.CompilerServices.ConfiguredTaskAwaitable`1.ConfiguredTaskAwaiter.GetResult()
[xUnit.net 00:00:02.5804690]            at System.IO.StreamReader.<ReadBufferAsync>d__65.MoveNext()
[xUnit.net 00:00:02.5805180]         --- End of stack trace from previous location where exception was thrown ---
[xUnit.net 00:00:02.5805840]            at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
[xUnit.net 00:00:02.5806490]            at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
[xUnit.net 00:00:02.5807130]            at System.Runtime.CompilerServices.ConfiguredTaskAwaitable`1.ConfiguredTaskAwaiter.GetResult()
[xUnit.net 00:00:02.5807710]            at System.IO.StreamReader.<ReadToEndAsyncInternal>d__61.MoveNext()
[xUnit.net 00:00:02.5808310]         --- End of stack trace from previous location where exception was thrown ---
[...]

PR #1840, opened 2017-05-16
Read request body concurrent to app execution

cesarbs:
A better version of what I was doing in https://github.com/aspnet/KestrelHttpServer/pull/1773.

TODO:

- [x] await request body task

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/Http/Frame.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/Http/Frame.cs
@@ -98,7 +98,10 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
             _requestHeadersTimeoutTicks = ServerOptions.Limits.RequestHeadersTimeout.Ticks;
 
             Output = new OutputProducer(frameContext.Output, frameContext.ConnectionId, frameContext.ServiceContext.Log);
+            RequestBodyPipe = CreateRequestBodyPipe();
         }
+
+        public IPipe RequestBodyPipe { get; }
 
         public ServiceContext ServiceContext => _frameContext.ServiceContext;
         public IConnectionInformation ConnectionInformation => _frameContext.ConnectionInformation;
@@ -1364,6 +1367,15 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
                 }
             }
         }
+
+        private IPipe CreateRequestBodyPipe()
+            => ConnectionInformation.PipeFactory.Create(new PipeOptions
+            {
+                ReaderScheduler = ServiceContext.ThreadPool,
+                WriterScheduler = ConnectionInformation.InputWriterScheduler,
+                MaximumSizeHigh = ServiceContext.ServerOptions.Limits.MaxRequestBufferSize ?? 0,
+                MaximumSizeLow = ServiceContext.ServerOptions.Limits.MaxRequestBufferSize ?? 0
+            });
 
         private enum HttpRequestTarget
         {
--- a/src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/Http/Frame.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/Http/Frame.cs
@@ -377,11 +377,6 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
 
             _responseBytesWritten = 0;
             _requestCount++;
-
-            // TODO: probably not the right place to complete
-            RequestBodyPipe.Reader.Complete();
-            RequestBodyPipe.Writer.Complete();
-            RequestBodyPipe.Reset();
         }
 
         /// <summary>
--- a/src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/Http/Frame.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/Http/Frame.cs
@@ -1371,8 +1371,8 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
         private IPipe CreateRequestBodyPipe()
             => ConnectionInformation.PipeFactory.Create(new PipeOptions
             {
-                ReaderScheduler = ServiceContext.InlineScheduler,
-                WriterScheduler = ServiceContext.InlineScheduler,
+                ReaderScheduler = ServiceContext.ThreadPool,
+                WriterScheduler = ConnectionInformation.InputWriterScheduler,
                 MaximumSizeHigh = ServiceContext.ServerOptions.Limits.MaxRequestBufferSize ?? 0,
                 MaximumSizeLow = ServiceContext.ServerOptions.Limits.MaxRequestBufferSize ?? 0
             });
--- a/src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/Http/Frame.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/Http/Frame.cs
@@ -377,6 +377,11 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
 
             _responseBytesWritten = 0;
             _requestCount++;
+
+            // TODO: probably not the right place to complete
+            RequestBodyPipe.Reader.Complete();
+            RequestBodyPipe.Writer.Complete();
+            RequestBodyPipe.Reset();
         }
 
         /// <summary>
--- a/src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/Http/Frame.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/Http/Frame.cs
@@ -1371,8 +1371,8 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
         private IPipe CreateRequestBodyPipe()
             => ConnectionInformation.PipeFactory.Create(new PipeOptions
             {
-                ReaderScheduler = ServiceContext.ThreadPool,
-                WriterScheduler = ConnectionInformation.InputWriterScheduler,
+                ReaderScheduler = ServiceContext.InlineScheduler,
+                WriterScheduler = ServiceContext.InlineScheduler,
                 MaximumSizeHigh = ServiceContext.ServerOptions.Limits.MaxRequestBufferSize ?? 0,
                 MaximumSizeLow = ServiceContext.ServerOptions.Limits.MaxRequestBufferSize ?? 0
             });
--- a/src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/Http/FrameOfT.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/Http/FrameOfT.cs
@@ -158,12 +158,13 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
                                     await ProduceEnd();
                                 }
 
-                                if (_keepAlive)
+                                // Finish reading the request body in case the app did not.
+                                if (_upgrade)
                                 {
-                                    // Finish reading the request body in case the app did not.
-                                    await messageBody.ConsumeAsync();
-                                    await messageBodyTask;
+                                    RequestBodyPipe.Reader.CancelPendingRead();
                                 }
+
+                                await messageBody.ConsumeAsync();
 
                                 if (!HasResponseStarted)
                                 {
@@ -193,9 +194,17 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
                             StopStreams();
                         }
 
-                        // If not keep-alive, ConsumeAsync isn't called and the reader isn't completed
+                        // After consuming, the pipe reader is completed.
+                        // Await the message body read task so we know the writer is completed as well.
+                        if (_upgrade)
+                        {
+                            Input.CancelPendingRead();
+                        }
+
+                        await messageBodyTask;
+
                         // ForZeroContentLength does not complete the reader nor the writer
-                        if (_keepAlive && !messageBody.IsEmpty)
+                        if (!messageBody.IsEmpty)
                         {
                             RequestBodyPipe.Reset();
                         }
--- a/src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/Http/FrameOfT.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/Http/FrameOfT.cs
@@ -158,8 +158,13 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
                                     await ProduceEnd();
                                 }
 
+                                if (!_keepAlive)
+                                {
+                                    messageBody.Cancel();
+                                }
+
                                 // An upgraded request has no defined request body length.
-                                // Cancel any pending read so the request body pipe can be drained.
+                                // Cancel any pending read so the read loop ends.
                                 if (_upgrade)
                                 {
                                     Input.CancelPendingRead();
--- a/src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/Http/FrameOfT.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/Http/FrameOfT.cs
@@ -93,6 +93,8 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
 
                         InitializeStreams(messageBody);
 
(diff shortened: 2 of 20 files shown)


src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/Http/Frame.cs
```diff
@@ -374,6 +377,11 @@ public void Reset()
 
             _responseBytesWritten = 0;
             _requestCount++;
+
+            // TODO: probably not the right place to complete
+            RequestBodyPipe.Reader.Complete();
+            RequestBodyPipe.Writer.Complete();
+            RequestBodyPipe.Reset();
```
halter73: The call the Reset() should go here. Completing the reader/writer should be localized to where the reading/writing occurs as much as possible.

src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/Http/Frame.cs
```diff
@@ -1422,6 +1430,15 @@ protected void EnsureHostHeaderExists()
             }
         }
 
+        private IPipe CreateRequestBodyPipe()
+            => ConnectionInformation.PipeFactory.Create(new PipeOptions
+            {
+                ReaderScheduler = ServiceContext.ThreadPool,
+                WriterScheduler = ConnectionInformation.InputWriterScheduler,
```
halter73: Won't this cause both reads and writes to double dispatch?

src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/Http/MessageBody.cs
```diff
@@ -30,179 +32,165 @@ protected MessageBody(Frame context)
...
-                if (count == 0)
-                {
-                    // Completed Task, end of stream
-                    return;
-                }
-
-                ConsumedBytes(count);
-                currentTask = PeekAsync(cancellationToken);
+                writableBuffer.Commit();
+                _currentWritableBuffer = writableBuffer;
```
halter73: Isn't there some better way to do this than setting a nullable WritableBuffer in a OnData callback?

src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/Http/MessageBody.cs
```diff
@@ -30,179 +32,165 @@ protected MessageBody(Frame context)
...
-                    {
-                        // Completed Task, end of stream
-                        return TaskCache.CompletedTask;
-                    }
-
-                    ConsumedBytes(task.Result.Count);
-                }
+            try
+            {
+                writableBuffer.Append(readableBuffer);
```
halter73: Append!!!

src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/Http/Frame.cs
davidfowl: Agreed

src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/Http/MessageBody.cs
davidfowl: Let's not append as yet.....

src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/Http/Frame.cs
halter73: Inline would be equivalent to the way we previously implemented the request and response body streams. That makes the most sense to me.

src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/Http/Frame.cs
cesarbs: `Reset()` fails if the reader and writer are not completed:

```
System.InvalidOperationException : Both reader and writer need to be completed to be able to reset
Stack Trace:
     at Microsoft.AspNetCore.Server.Kestrel.Internal.System.IO.Pipelines.Pipe.Reset()
     at Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http.Frame.Reset()
     at Microsoft.AspNetCore.Server.Kestrel.Core.Tests.FrameTests..ctor()
```

and we call `Frame.Reset()` essentially before anything else: https://github.com/aspnet/KestrelHttpServer/blob/1a706113d9f75999d9642ccc05b41593f7c0aa50/src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/Http/FrameOfT.cs#L39

src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/Http/MessageBody.cs
cesarbs: I could use the same pattern from `LibuvConnection`: https://github.com/aspnet/KestrelHttpServer/blob/41f1922502f475802c6cf3ba2cbe3653ab5d00fd/src/Microsoft.AspNetCore.Server.Kestrel.Transport.Libuv/Internal/LibuvConnection.cs#L140-L147

And call `FlushAsync()` in `OnData()`. I'll have to care about backpressure anyways, when I add the timeout.

halter73: Storing the WritableBuffer in a field there there is also super ugly. But there we have an excuse: it's a callback-based native library that's forcing us into it.

src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/Http/Frame.cs
halter73: The call to Reset can be skipped prior to the first request. You could even move the call to `RequestBodyPipe.Reset()` out of `Frame.Reset()` and directly into the end of `ProcessRequestsAsync()` loop if that's cleaner.

src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/Http/MessageBody.cs
cesarbs: Yeah, what I meant is, if I don't away the flush then I don't have to store the buffer in a field, since I could call flush in `OnData`.

halter73: Couldn't you make OnData async and flush in OnData?

src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/Http/Frame.cs
cesarbs: https://github.com/aspnet/KestrelHttpServer/pull/1840/commits/4b928941b0ffe2c06a8e7beb90e22288f0970ddf

src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/Http/MessageBody.cs
cesarbs: Yep, has issues:

```
[xUnit.net 00:00:25.2336688]     LargeUpload(maxRequestBufferSize: 16384, connectionAdapter: False, expectPause: True) [FAIL]
[xUnit.net 00:00:25.2347046]       Assert.Contains() Failure
[xUnit.net 00:00:25.2347876]       Not found: bytesRead: 20971520
[xUnit.net 00:00:25.2348301]       In value:  HTTP/1.1 500 Internal Server Error
[xUnit.net 00:00:25.2348650]       Connection: close
[xUnit.net 00:00:25.2348993]       Date: Tue, 16 May 2017 21:00:06 GMT
[xUnit.net 00:00:25.2349313]       Server: Kestrel
[xUnit.net 00:00:25.2349635]
[xUnit.net 00:00:25.2349902]       Bytes received do not match expectedBody at position 9394515
[xUnit.net 00:00:25.2358474]       Stack Trace:
[xUnit.net 00:00:25.2368072]            at Microsoft.AspNetCore.Server.Kestrel.FunctionalTests.MaxRequestBufferSizeTests.<LargeUpload>d__4.MoveNext()
[xUnit.net 00:00:25.2368993]         --- End of stack trace from previous location where exception was thrown ---
[xUnit.net 00:00:25.2369515]            at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
[xUnit.net 00:00:25.2369958]            at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
[xUnit.net 00:00:25.2370383]         --- End of stack trace from previous location where exception was thrown ---
[xUnit.net 00:00:25.2370782]            at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
[xUnit.net 00:00:25.2371207]            at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
[xUnit.net 00:00:25.2371547]         --- End of stack trace from previous location where exception was thrown ---
[xUnit.net 00:00:25.2371937]            at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
[xUnit.net 00:00:25.2372342]            at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
Failed   LargeUpload(maxRequestBufferSize: 16384, connectionAdapter: False, expectPause: True)
Error Message:
 Assert.Contains() Failure
Not found: bytesRead: 20971520
In value:  HTTP/1.1 500 Internal Server Error
Connection: close
Date: Tue, 16 May 2017 21:00:06 GMT
Server: Kestrel

Bytes received do not match expectedBody at position 9394515
Stack Trace:
   at Microsoft.AspNetCore.Server.Kestrel.FunctionalTests.MaxRequestBufferSizeTests.<LargeUpload>d__4.MoveNext()
```

Don't see the same when switching to `Write()`.

cesarbs:
Removed `OnData()`: e2e5f5793c510ce7d01745d37f8987eaf4967d90

cesarbs:
Inline scheduler: 189662e545caa2d15f3debc09bb6145e6f017ca1

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/MaxRequestBufferSizeTests.cs
```diff
@@ -165,7 +165,7 @@ public static IEnumerable<object[]> LargeUploadData
 
                     using (var reader = new StreamReader(stream, Encoding.ASCII))
                     {
-                        var response = reader.ReadToEnd();
+                        var response = await reader.ReadToEndAsync();
```
cesarbs: This test times out when using non-async APIs.

halter73: Weird? Is it just because ReadToEnd() is slower, or is it actually getting blocked somehow?

halter73: Weird. Is it just because ReadToEnd() is slower, or is it actually getting blocked somehow?

cesarbs: It's getting blocked. I can show you the stack trace tomorrow in the office.

src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/Http/FrameOfT.cs
```diff
@@ -158,13 +158,14 @@ public Frame(IHttpApplication<TContext> application, FrameContext frameContext)
...
                                 }
 
-                                if (_keepAlive)
+                                // Finish reading the request body in case the app did not.
+                                if (_upgrade)
                                 {
-                                    // Finish reading the request body in case the app did not.
-                                    await messageBody.ConsumeAsync();
-                                    await messageBodyTask;
+                                    RequestBodyPipe.Reader.CancelPendingRead();
```
davidfowl: This deserves an uber comment.

src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/ServiceContext.cs
```diff
@@ -13,6 +14,8 @@ public class ServiceContext
 
         public IThreadPool ThreadPool { get; set; }
 
+        public IScheduler InlineScheduler { get; set; }
```
davidfowl: This isn't right.

cesarbs:
Not using the inline scheduler anymore. Causes deadlocks when mixing sync and async I/O, and @davidfowl and I decided that dispatching is better.

halter73:
How does the inline scheduler cause deadlocks in this code?

davidfowl:
> How does the inline scheduler cause deadlocks in this code?

Calling ReadAsync means that user code continues on the user writer's thread. If that blocks after consuming all written data, the writer can never make progress. For e.g.

```
messageBody.StartAsync()
    writer.FlushAsync()
       await reader.ReadAsync()
             Read()  <- stuck
```

The thread would need to yield control back to the writer to not block forever.

cesarbs:
@halter73 The synchronous read is run on the same thread as the writer's `FlushAsync`. The synchronous read is `ReadAsync().Result`, but that `ReadAsync` won't yield before the writer yields. But the writer can't yield because it's running the continuation that called the read.

halter73:
Makes sense. The first ReadAsync().Result would succeed, but the second one would block the loop. This doesn't explain why the WriterScheduler can't be inline though. The code calling FlushAsync() never blocks, right? That should be the case now that FlushAsync() itself doesn't block given the ReaderScheduler dispatching.

cesarbs:
I should point out that the deadlock only happens when mixing sync and async code. I could go back to the inline scheduler and change the hanging test again to be all async, and things would be good.

Then we could tell users not to mix sync and async, but I'm pretty sure we'll be breaking someone by doing that. Given that, I'd rather stick to dispatching.

benaadams:
Throw exceptions from the sync methods and have a config switch that doesn't throw the exceptions and changes to threadpool scheduler? Or you can skip the scheduler changes and just throw the exceptions if opt in to sync is not done? 😈

cesarbs:
I've reopened #1270 for the Travis failure. I don't have an exact repro but I can make it fail when running in a loop on macOS. The error I'm seeing though is a "too many files open", which could have happened on Travis as well but in a way such that it caused a connection reset instead of bubbling that exact exception all the way up. I don't think it's related to this change.

We can address the scheduler and sync/async stuff as a subsequent change. Unless there are any issues with the actual implementation of how the body is read, I'd like to get this merged so I can work on the bigger change that this is part of.

src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/Http/MessageBody.cs
```diff
@@ -30,179 +30,171 @@ protected MessageBody(Frame context)
...
-                    await destination.WriteAsync(segment.Array, segment.Offset, segment.Count, cancellationToken);
+                    if (!readableBuffer.IsEmpty)
+                    {
+                        foreach (var memory in readableBuffer)
+                        {
+                            var array = memory.GetArray();
+                            await destination.WriteAsync(array.Array, array.Offset, array.Count, cancellationToken);
+                        }
+                    }
+                    else if (result.IsCompleted || result.IsCancelled)
```
cesarbs: I think we should throw `TaskCanceledException` on `IsCancelled`. Any reason not to?

src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/Http/MessageBody.cs
```diff
@@ -30,179 +30,171 @@ protected MessageBody(Frame context)
...
 
-            if (consumed != 0)
+                            if (done)
+                            {
+                                break;
+                            }
+                        }
+                        else if (result.IsCancelled)
+                        {
+                            break;
```
halter73: Frame.StopAsync() calls Input.CancelPendingRead(), but we don't want that to kill the request body. We only want to do that given an abort.

cesarbs:
Added tests for request body cancelation and request consumption: 3c9a506dd3a23dc3312b51ea53cd1fabaf8f7076 4ed93f24955d8e31d472ac433221574f0ae4ab25

cesarbs:
@halter73 @davidfowl I believe this covers what we discussed about input cancelation: 914cc1ca3ae8ca80a4d4f4c7b0b09d8efbbfab9d

cesarbs:
@davidfowl Fixed `Copy()`: bda2eef212c716071f54938142838baf82fe80f0

cesarbs:
The test failure I mentioned in person is happening in dev too: https://github.com/aspnet/KestrelHttpServer/issues/1850

Merging.

PR #1843, opened 2017-05-17
Fix flaky ResponseTests.FailedWritesResultInAbortedRequest.

cesarbs:
Sometimes it throws `TaskCancelledException`, sometimes `OperationCancelledException`. Depends on how far the write that's cancelled goes.

This makes the test more deterministic so it always throws `TaskCancelledException`.

### DIFF
--- a/test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/ResponseTests.cs
+++ b/test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/ResponseTests.cs
@@ -2063,7 +2063,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.FunctionalTests
 
         [Theory]
         [MemberData(nameof(ConnectionAdapterData))]
-        public async Task FailedWritesResultInAbortedRequest(ListenOptions listenOptions)
+        public async Task ThrowsOnWriteAfterRequestIsAborted(ListenOptions listenOptions)
         {
             // This should match _maxBytesPreCompleted in SocketOutput
             var maxBytesPreCompleted = 65536;
@@ -2072,7 +2072,6 @@ namespace Microsoft.AspNetCore.Server.Kestrel.FunctionalTests
 
             var writeTcs = new TaskCompletionSource<object>();
             var requestAbortedWh = new ManualResetEventSlim();
-            var connectionCloseWh = new ManualResetEventSlim();
             var requestStartWh = new ManualResetEventSlim();
 
             using (var server = new TestServer(async httpContext =>
@@ -2113,8 +2112,6 @@ namespace Microsoft.AspNetCore.Server.Kestrel.FunctionalTests
 
                     Assert.True(requestStartWh.Wait(TimeSpan.FromSeconds(10)));
                 }
-
-                connectionCloseWh.Set();
 
                 // Write failed - can throw TaskCanceledException or OperationCanceledException,
                 // dependending on how far the canceled write goes.
--- a/test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/ResponseTests.cs
+++ b/test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/ResponseTests.cs
@@ -2065,37 +2065,34 @@ namespace Microsoft.AspNetCore.Server.Kestrel.FunctionalTests
         [MemberData(nameof(ConnectionAdapterData))]
         public async Task FailedWritesResultInAbortedRequest(ListenOptions listenOptions)
         {
-            var testContext = new TestServiceContext();
-
             // This should match _maxBytesPreCompleted in SocketOutput
             var maxBytesPreCompleted = 65536;
             // Ensure string is long enough to disable write-behind buffering
             var largeString = new string('a', maxBytesPreCompleted + 1);
 
             var writeTcs = new TaskCompletionSource<object>();
-            var registrationWh = new ManualResetEventSlim();
+            var requestAbortedWh = new ManualResetEventSlim();
             var connectionCloseWh = new ManualResetEventSlim();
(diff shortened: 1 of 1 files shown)


test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/ResponseTests.cs
```diff
@@ -2064,37 +2064,34 @@ public Task ResponseStatusCodeSetBeforeHttpContextDisposedRequestMalformedReadIg
...
                 var response = httpContext.Response;
                 var request = httpContext.Request;
                 var lifetime = httpContext.Features.Get<IHttpRequestLifetimeFeature>();
 
-                lifetime.RequestAborted.Register(() => registrationWh.Set());
+                lifetime.RequestAborted.Register(() => requestAbortedWh.Set());
 
-                await request.Body.CopyToAsync(Stream.Null);
-                Assert.True(connectionCloseWh.Wait(10_000));
+                requestAbortedWh.Wait(TimeSpan.FromSeconds(10));
```
pakrym: Add `Assert.True`

cesarbs:
@pakrym Switched to `ThrowsAnyAsync`.

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/ResponseTests.cs
```diff
@@ -2064,37 +2064,34 @@ public Task ResponseStatusCodeSetBeforeHttpContextDisposedRequestMalformedReadIg
         [MemberData(nameof(ConnectionAdapterData))]
         public async Task FailedWritesResultInAbortedRequest(ListenOptions listenOptions)
```
halter73: Change the name of the test.

This should no longer be called FailedWrites*Result*InAbortedRequest. Now the test is written so that the request is *already* aborted before it simply verifies that calling WriteAsync(..., RequestAborted) throws. It's no longer a write that actually trips RequestAborted token like in Kestrel 1.x. Now LibuvConnection.OnRead() trips the RequestAborted token anytime the client closes the connection, not just for RSTs.

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/ResponseTests.cs
```diff
@@ -2064,37 +2064,34 @@ public Task ResponseStatusCodeSetBeforeHttpContextDisposedRequestMalformedReadIg
...
-
             // This should match _maxBytesPreCompleted in SocketOutput
             var maxBytesPreCompleted = 65536;
             // Ensure string is long enough to disable write-behind buffering
             var largeString = new string('a', maxBytesPreCompleted + 1);
 
             var writeTcs = new TaskCompletionSource<object>();
-            var registrationWh = new ManualResetEventSlim();
+            var requestAbortedWh = new ManualResetEventSlim();
             var connectionCloseWh = new ManualResetEventSlim();
```
halter73: Is connectionCloseWh still being used?

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/ResponseTests.cs
```diff
@@ -2064,37 +2064,34 @@ public Task ResponseStatusCodeSetBeforeHttpContextDisposedRequestMalformedReadIg
...
                 requestStartWh.Set();
+
                 var response = httpContext.Response;
                 var request = httpContext.Request;
                 var lifetime = httpContext.Features.Get<IHttpRequestLifetimeFeature>();
 
-                lifetime.RequestAborted.Register(() => registrationWh.Set());
+                lifetime.RequestAborted.Register(() => requestAbortedWh.Set());
 
-                await request.Body.CopyToAsync(Stream.Null);
```
halter73: Removing this didn't fix anything, right? I'm assuming it was just unnecessary.

cesarbs:
🆙 📅

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/ResponseTests.cs
```diff
@@ -2063,39 +2063,34 @@ public Task ResponseStatusCodeSetBeforeHttpContextDisposedRequestMalformedReadIg
...
+                lifetime.RequestAborted.Register(() => requestAbortedWh.Set());
+                Assert.True(requestAbortedWh.Wait(TimeSpan.FromSeconds(10)));
 
                 try
                 {
                     // Ensure write is long enough to disable write-behind buffering
-                    for (int i = 0; i < 100; i++)
+                    for (var i = 0; i < 100; i++)
                     {
                         await response.WriteAsync(largeString, lifetime.RequestAborted);
```
halter73: Since you've already called `Assert.True(requestAbortedWh.Wait(TimeSpan.FromSeconds(10)));` above, there's no reason to call this in a loop. This should always throw an OperationCanceledException on the first call to WriteAsync.

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/ResponseTests.cs
```diff
@@ -2063,39 +2063,34 @@ public Task ResponseStatusCodeSetBeforeHttpContextDisposedRequestMalformedReadIg
 
         [Theory]
         [MemberData(nameof(ConnectionAdapterData))]
-        public async Task FailedWritesResultInAbortedRequest(ListenOptions listenOptions)
+        public async Task ThrowsOnWriteAfterRequestIsAborted(ListenOptions listenOptions)
```
halter73: Nit: Maybe ThrowsOnWriteWithRequestAbortedTokenAfterRequestIsAborted since normally we don't throw on write after a request is aborted.

PR #1844, opened 2017-05-17
Throw from UpgradeAsync when request is not upgradable

natemcmaster:
Add a check to IHttpUpgradeFeature.UpgradeAsync to ensure IHttpUpgradeFeature.IsUpgradableRequest is true before upgrading the request and sending HTTP 101.


cesarbs:
CI is upset.

natemcmaster:
Yeah, this broke a few functional tests that were calling UpgradeAsync without sending `Connection: Upgrade`. I've rebased on dev and updated.

PR #1845, opened 2017-05-18
Expose Heartbeat and enable adding and removing handlers

cesarbs:
This would allow hooking `MessageBody` directly to `Heartbeat` after #1840, to implement the request body timeout. Allows for a cleaner timeout implementation than passing `ITimeoutControl` to `MessageBody` and doing tricks like `SetTimeout(0)` to force a timeout when using a variable timeout based on the request rate.

### DIFF
--- a/test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/TestServer.cs
+++ b/test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/TestServer.cs
@@ -55,33 +55,33 @@ namespace Microsoft.AspNetCore.Server.Kestrel.FunctionalTests
             Context = context;
 
             _host = new WebHostBuilder()
-                     .UseKestrel(o =>
-                     {
-                         o.ListenOptions.Add(_listenOptions);
-                     })
-                     .ConfigureServices(services =>
-                     {
-                         if (httpContextFactory != null)
-                         {
-                             services.AddSingleton(httpContextFactory);
-                         }
+                .UseKestrel(o =>
+                {
+                    o.ListenOptions.Add(_listenOptions);
+                })
+                .ConfigureServices(services =>
+                {
+                    if (httpContextFactory != null)
+                    {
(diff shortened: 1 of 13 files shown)


test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/TestServer.cs
```diff
@@ -55,33 +55,33 @@ public TestServer(RequestDelegate app, TestServiceContext context, ListenOptions
...
-                             foreach (var c in configureOptions)
-                             {
-                                 c.Configure(context.ServerOptions);
-                             }
-                             return new KestrelServer(sp.GetRequiredService<ITransportFactory>(), context);
-                         });
-                     })
-                     .UseSetting(WebHostDefaults.ApplicationKey, typeof(TestServer).GetTypeInfo().Assembly.FullName)
-                     .Build();
+                .UseKestrel(o =>
```
davidfowl: Why did you mess up the formatting?

cesarbs: ?

halter73:
I would close this PR without merging.

FrameHeartbeatManager isn't doing anything anymore now that the WalkCallback is empty yet the it's still there. This is netting over 100 additional lines of code to do the exact same thing as before while making the Heartbeat class more complicated and harder to use safely.

As far as things like avoiding calling SetTimeout(0) in MessageBody, you could just call Timeout() directly instead.

Leave Heartbeat the way it is. If you need to walk a dynamic collection of objects every beat, make another class like FrameConnectionManager which allows you to simply walk the active objects on every beat like FrameHeartbeat manager. FrameConnectionManager and FrameHeartbeatManager are well less than 100 lines combined and allow for a much simpler Heartbeat class that just manages the timer and does some tracing.

cesarbs:
Thanks for the feedback. Opened https://github.com/aspnet/KestrelHttpServer/pull/1847.

PR #1846, opened 2017-05-18
React to pipeline scheduler changes

pakrym:


natemcmaster:
cc @Tratcher

src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/Infrastructure/InlineLoggingThreadPool.cs
```diff
@@ -33,9 +33,9 @@ public void UnsafeRun(WaitCallback action, object state)
             action(state);
         }
 
-        public void Schedule(Action action)
+        public void Schedule(Action<object> action, object state)
         {
-            Run(action);
+            action(state);
```
campersau: No more `try` `catch`? So this is the same as `UnsafeRun`.

PR #1017, opened 2016-07-25
Better logging for connection errors during HTTPS handshake

halter73:
Addresses #960

PR #1854, opened 2017-05-24
Implement max connection limits

natemcmaster:
- Added new options to allow configuring the maximum number of concurrent connections and upgraded connections
  - `KestrelServerLimits.MaxConcurrentConnections` defaults to `null` (unlimited)
  - `KestrelServerLimits.MaxConcurrentUpgradedConnections` defaults to  `null` (unlimited)
- Calls to IHttpUpgradeFeature.UpgradeAsync() will throw when the MaxConcurrentUpgradedConnections limit has been reached
~~- Kestrel will respond with HTTP 503 when MaxConcurrentConnections is reached~~ Updated: we're going to close the connection ASAP.

Resolves https://github.com/aspnet/Coherence-Signed/issues/500


natemcmaster:
PlainText

**dev**
```
RPS: 1278386.2
RPS: 1299188.05
```

**namc/connection-limits**
```
RPS: 1291492.32
RPS: 1307191.9
```

src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/Http/Frame.FeatureCollection.cs
```diff
@@ -235,6 +235,13 @@ void IHttpResponseFeature.OnCompleted(Func<object, Task> callback, object state)
                 throw new InvalidOperationException(CoreStrings.CannotUpgradeNonUpgradableRequest);
             }
 
+            if (!ServiceContext.Resources.UpgradedConnections.TryLockOne())
```
halter73: @DamianEdwards @davidfowl  You don't want upgraded connections applying to the total connection limit, right?

natemcmaster:
:up: :date:

src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/Infrastructure/ResourceCounter.cs
```diff
@@ -0,0 +1,68 @@
...
+                {
+                    throw new ArgumentOutOfRangeException(CoreStrings.NonNegativeNumberRequired);
+                }
+
+                _max = max;
+            }
+
+            public override bool TryLockOne()
+            {
+                var next = Interlocked.Increment(ref _count);
```
davidfowl: Feels like there should be a compare exchange way of doing this type of logic to make it more atomic but I haven't thought about it

src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/RejectionConnection.cs
```diff
@@ -0,0 +1,48 @@
...
+
+using System;
+using Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Infrastructure;
+using Microsoft.AspNetCore.Server.Kestrel.Internal.System.IO.Pipelines;
+using Microsoft.AspNetCore.Server.Kestrel.Transport.Abstractions;
+
+namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal
+{
+    public class RejectionConnection : IConnectionContext
+    {
```
davidfowl: I still hate this.

src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/RejectionConnection.cs
```diff
@@ -0,0 +1,48 @@
...
+        }
+
+        // TODO: Remove these (https://github.com/aspnet/KestrelHttpServer/issues/1772)
+        public void OnConnectionClosed(Exception ex)
+        {
+        }
+
+        public void Abort(Exception ex)
+        {
+            Input.CancelPendingFlush();
```
davidfowl: What's this for?

davidfowl: I think you just need to complete both pipes and call it a day

src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/RejectionConnection.cs
natemcmaster: We've gone in circles on this. At this point, this is the lowest cost implementation that avoids issues such as throttling TLS handshake, setting timeouts, etc.

I'll open an issue to add the ability to send a 503 response. I've kept my partial implementation of that so we can revisit this post 2.0.

src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/Infrastructure/ResourceCounter.cs
benaadams: Atomic increment with cap https://github.com/dotnet/coreclr/blob/master/src/mscorlib/src/System/Threading/ThreadPool.cs#L392-L410

natemcmaster:
Getting errors from LibuvOutputConsumer.WriteOutputAsync. Not sure what to do about this

```
Reading is not allowed after reader was completed
```

src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/Infrastructure/ResourceCounter.cs
benaadams: Atomic decrement with floor
https://github.com/dotnet/coreclr/blob/master/src/mscorlib/src/System/Threading/ThreadPool.cs#L412-L430

src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/RejectionConnection.cs
davidfowl: Ok for 2.0 aborting is fine

src/Microsoft.AspNetCore.Server.Kestrel.Transport.Libuv/Internal/LibuvOutputConsumer.cs
```diff
@@ -37,7 +37,16 @@ public class LibuvOutputConsumer
 
             while (true)
             {
-                var result = await _pipe.ReadAsync();
+                ReadResult result;
+                try
+                {
+                    result = await _pipe.ReadAsync();
```
davidfowl: Undo this as it isn't correct.

davidfowl: You just need to close the correct sides of the pipes

src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/RejectionConnection.cs
```diff
@@ -0,0 +1,45 @@
...
+        }
+
+        // TODO: Remove these (https://github.com/aspnet/KestrelHttpServer/issues/1772)
+        public void OnConnectionClosed(Exception ex)
+        {
+        }
+
+        public void Abort(Exception ex)
+        {
+            Input.Complete(ex);
```
davidfowl: You're closing the wrong sides of the pipe. You close the input reader and output writer. The transport owns the other side.

src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/Infrastructure/ResourceCounter.cs
natemcmaster: Thanks Ben.

src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/RejectionConnection.cs
natemcmaster: :embarrassed: oh duh. Updated.

src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/Infrastructure/KestrelTrace.cs
```diff
@@ -30,6 +30,9 @@ public class KestrelTrace : IKestrelTrace
         private static readonly Action<ILogger, string, Exception> _connectionDisconnect =
             LoggerMessage.Define<string>(LogLevel.Debug, 10, @"Connection id ""{ConnectionId}"" disconnecting.");
 
+        private static readonly Action<ILogger, string, Exception> _connectionRejected =
+            LoggerMessage.Define<string>(LogLevel.Warning, 11, @"Connection id ""{ConnectionId}"" rejected.");
```
natemcmaster: Soliciting feedback. Is `Warning` too high? Should this be debug or info instead?

natemcmaster:
:up: :date:

Fix the "Reading is not allowed" error.

Pivoted this to closing connections immediately instead of sending a 503 response. We were concerned about some the complexity of dealing with timeouts and connection filters (e.g. should we bother with the TLS handshake?) and felt it warranted more discussion.

src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/Infrastructure/KestrelTrace.cs
cesarbs: 11 is marked as reserved in `LibuvTrace`, so you'll have to pick another id.

src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/Infrastructure/ResourceManager.cs
```diff
@@ -0,0 +1,24 @@
...
+        public ResourceManager(ResourceCounter normalConnections, ResourceCounter upgradedConnections)
+        {
+            NormalConnections = normalConnections;
+            UpgradedConnections = upgradedConnections;
+        }
+
+        /// <summary>
+        /// TCP connections processed by Kestrel.
+        /// </summary>
+        public ResourceCounter NormalConnections { get; }
```
cesarbs: I think these properties should be suffixed with `Count`.

src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/RejectionConnection.cs
```diff
@@ -0,0 +1,46 @@
...
+        public void Reject()
+        {
+            KestrelEventSource.Log.ConnectionRejected(ConnectionId);
+            _log.ConnectionRejected(ConnectionId);
+            _input.Reader.Complete();
+            _output.Writer.Complete();
+        }
+
+        // TODO: Remove these (https://github.com/aspnet/KestrelHttpServer/issues/1772)
+        public void OnConnectionClosed(Exception ex)
```
halter73: You could implement `IConnectionContext.OnConnectionClosed/Abort` explicitly so it's guaranteed to be removed when the interface changes.

test/Microsoft.AspNetCore.Server.Kestrel.Core.Tests/KestrelServerLimitsTests.cs
```diff
@@ -187,5 +187,58 @@ public void RequestHeadersTimeoutValid(double seconds)
...
+                MaxConcurrentConnections = value
+            };
+
+            Assert.Equal(value, limits.MaxConcurrentConnections);
+        }
+
+        [Fact]
+        public void MaxConnectionsInvalid()
+        {
+            Assert.Throws<ArgumentOutOfRangeException>(() => new KestrelServerLimits().MaxConcurrentUpgradedConnections = 0);
```
cesarbs: This is checking the wrong property.

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/ConnectionLimitTests.cs
```diff
@@ -0,0 +1,152 @@
...
+namespace Microsoft.AspNetCore.Server.Kestrel.FunctionalTests
+{
+    public class ConnectionLimitTests
+    {
+        private static async Task KeepOpen(HttpContext context)
+        {
+            // begin the response
+            await context.Response.WriteAsync("Hello");
+
+            // keep it running until abored
```
cesarbs: nit: typo

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/MaxRequestBufferSizeTests.cs
```diff
@@ -55,7 +55,7 @@ public static IEnumerable<object[]> LargeUploadData
                     Tuple.Create((long?)_dataLength - 1, false),
 
                     // Buffer is exactly the same size as data.  Exposed race condition where
-                    // IConnectionControl.Resume() was called after socket was disconnected.
+                    // ITimeoutControl.Resume() was called after socket was disconnected.
```
cesarbs: ?

src/Microsoft.AspNetCore.Server.Kestrel.Core/KestrelServerLimits.cs
```diff
@@ -170,13 +152,51 @@ public TimeSpan KeepAliveTimeout
...
+        /// <summary>
+        /// Gets or sets the maximum number of upgraded connections that can be open. When set to null, the number of upgraded connections is unlimited.
+        /// An upgraded connection is one that has been switched from HTTP to another protocol, such as WebSockets.
+        /// </summary>
+        /// <remarks>
+        /// <para>
+        /// Defaults to <see cref="uint.MaxValue"/>.
+        /// </para>
+        /// <para>
+        /// This limit is independent of the value of <see cref="MaxConcurrentConnections" />.
```
halter73: A clearer way to phrase this might be `Upgraded connections do not count towards the <see cref="MaxConcurrentConnections" /> limit.`

src/Microsoft.AspNetCore.Server.Kestrel.Core/KestrelServerLimits.cs
```diff
@@ -170,13 +152,51 @@ public TimeSpan KeepAliveTimeout
...
+        /// </para>
+        /// </remarks>
+        public uint? MaxConcurrentUpgradedConnections
+        {
+            get => _maxConcurrentUpgradedConnections;
             set
             {
-                _requestHeadersTimeout = value;
+                if (value.HasValue && value <= 0)
```
halter73: I don't see why zero wouldn't be valid. A little weird, but not completely pointless.

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/MaxRequestBufferSizeTests.cs
natemcmaster: This happened because the first draft of this renamed `ITimeoutControl => IConnectionControl`. Reverting that change found this. Not sure what this is supposed to be. We don't have an IConnectionControl in dev or this PR.

src/Microsoft.AspNetCore.Server.Kestrel.Core/KestrelServer.cs
```diff
@@ -88,6 +88,14 @@ private static ServiceContext CreateServiceContext(IOptions<KestrelServerOptions
...
             }
 
+            var normalConnections = serverOptions.Limits.MaxConcurrentConnections.HasValue
+                    ? ResourceCounter.Quota(serverOptions.Limits.MaxConcurrentConnections.Value)
+                    : ResourceCounter.Unlimited;
+
+            var upgradedConnections = serverOptions.Limits.MaxConcurrentUpgradedConnections.HasValue
+                    ? ResourceCounter.Quota(serverOptions.Limits.MaxConcurrentUpgradedConnections.Value)
+                    : ResourceCounter.Unlimited;
+
```
halter73: Nit: move the above to the ResourceManager ctor.

src/Microsoft.AspNetCore.Server.Kestrel.Core/KestrelServer.cs
```diff
@@ -96,7 +104,8 @@ private static ServiceContext CreateServiceContext(IOptions<KestrelServerOptions
                 SystemClock = systemClock,
                 DateHeaderValueManager = dateHeaderValueManager,
                 ConnectionManager = connectionManager,
-                ServerOptions = serverOptions
+                Resources = new ResourceManager(normalConnections, upgradedConnections),
+                ServerOptions = serverOptions,
```
halter73: What do you think about moving ResourceManager into a filed of ConnectionManager?

I think you could have ConnectionManager.TryAddConnection which calls Resources.NormalConnections.TryLockOne(),  Log.ConnectionStart, EventSource.ConnectionStart etc... in addition to what it does already assuming TryLockOne succeeds. ConnectionManager.RemoveConnection would do the dual.

This would consolidate/encapsulate the connection tracking/logging logic which should make it a little easier to follow in addition to removing a top-level property on ServiceContext. FrameConnectionManager could just take KestrelServerLimits as a ctor param and initialize the ResourceCounters itself.

src/Microsoft.AspNetCore.Server.Kestrel.Core/KestrelServerLimits.cs
```diff
@@ -170,13 +152,51 @@ public TimeSpan KeepAliveTimeout
...
+            set => _requestHeadersTimeout = value;
+        }
+
+        /// <summary>
+        /// Gets or sets the maximum number of connections that will accept requests. When set to null, the number of connections is unlimited.
+        /// </summary>
+        /// <remarks>
+        /// Defaults to <see cref="uint.MaxValue"/>.
+        /// </remarks>
+        public uint? MaxConcurrentConnections
```
halter73: While I don't think we're still worried about  CLS compliance these days, let's avoid unsigned numeric types for consistency with the other Kestrel limits and other .NET APIs in general.

Can we just make it a long and default to null?

src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/Http/Frame.FeatureCollection.cs
```diff
@@ -235,6 +235,18 @@ void IHttpResponseFeature.OnCompleted(Func<object, Task> callback, object state)
...
+            {
+                throw new InvalidOperationException(CoreStrings.UpgradeCannotBeCalledMultipleTimes);
+            }
+
+            if (!ServiceContext.Resources.UpgradedConnections.TryLockOne())
+            {
+                throw new InvalidOperationException(CoreStrings.UpgradedConnectionLimitReached);
+            }
+
+            _wasUpgraded = true;
```
halter73: @natemcmaster Do you want to change this to call NormalConnections.ReleaseOne() here (and of course avoid calling it twice), or do you want to remove MaxConcurrentUpgradedConnections? I'll approve either way, but we shouldn't have a separate upgraded connection limit that still applies to the normal connection limit.

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/MaxRequestBufferSizeTests.cs
cesarbs: Oh yeah, that's old, didn't realize that at first. We can change that to "the connection was resumed after socket was disconnected."

src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/Infrastructure/KestrelTrace.cs
halter73: If we reserved an ID in LibuvTrace that was already used in KestrelTrace, I would have been fine with it, but since this ID was previously used for the same logger name, let's change this. To be safe I always choose an ID above all the existing ones defined in that project.

halter73: We decided when working on #1017 that if there's nothing the application developer could do to prevent a client from forcing the log (e.g. bad request), then the log level should not be higher than Information.

I think this is an exception to that rule though, because the client won't be getting a 503 (for now) and hitting the connection limit can cause connection refusals for clients that aren't responsible for the limit being hit.

src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/Http/Frame.FeatureCollection.cs
natemcmaster: This was our original design requirement, so I'll make this change.

src/Microsoft.AspNetCore.Server.Kestrel.Core/KestrelServerLimits.cs
natemcmaster: Good point. Will update.

src/Microsoft.AspNetCore.Server.Kestrel.Core/KestrelServer.cs
natemcmaster: Would be tough in the current implementation. This might be a good cleanup when/if we consolidate RejectionConnection back into FrameConnection.

src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/Http/Frame.FeatureCollection.cs
natemcmaster: Updated

natemcmaster:
FYI - I opened https://github.com/aspnet/KestrelHttpServer/issues/1864 to follow up in the future on some ideas discussed here such as sending a 503

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/ConnectionLimitTests.cs
```diff
@@ -0,0 +1,199 @@
...
+{
+    public class ConnectionLimitTests
+    {
+        private static async Task KeepOpen(HttpContext context)
+        {
+            // begin the response
+            await context.Response.WriteAsync("Hello");
+
+            // keep it running until aborted
+            while (!context.RequestAborted.IsCancellationRequested)
```
davidfowl: Instead of this why it just use a tcs that gets triggered on the cancellation callback?

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/ConnectionLimitTests.cs
```diff
@@ -0,0 +1,199 @@
...
+                disposables.Push(connection);
+
+                await connection.SendEmptyGetAsKeepAlive();
+                await connection.Receive("HTTP/1.1 200 OK");
+
+                using (var rejected = server.CreateConnection())
+                {
+                    try
+                    {
+                        // this may throw IOException, depending on how fast Kestrel closes the socket
```
davidfowl: Can we make it behave in a consistent way with another tcs? Or is that impossible?

natemcmaster: We probably could, but it doesn't seem super important as the real thing we're interested in here is asserting that the connection is closed without writing data back, which is asserted by `connection.WaitForConnectionClose()`

halter73: You don't actually have to send a request for `connection.WaitForConnectionClose()` to complete, do you?

natemcmaster: True, but I want this test to ensure our server closes without responding to requests.

src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/Infrastructure/ResourceManager.cs
natemcmaster: updated that too

natemcmaster:
rebased, squashed, and updated. Any final comments? We'd like to get this in today.

halter73:
LGTM!

Issue #1835, opened 2017-05-13
Flaky Test: ConnectionResetBetweenRequestsIsLoggedAsDebug


halter73:
```
Assert.True() Failure
Expected: True
Actual:   False
   at Microsoft.AspNetCore.Server.Kestrel.FunctionalTests.RequestTests.<ConnectionResetBetweenRequestsIsLoggedAsDebug>d__13.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
```

http://aspnetci/viewLog.html?buildId=219517&tab=buildResultsDiv&buildTypeId=CoreCLR_Darwin

kichalla:
01b8d5fad1753a1eda61124d1b08447a81a8f0c0

PR #1859, opened 2017-05-24
Fix flaky test and improve TestServer

halter73:
#1835


test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/RequestTests.cs
```diff
@@ -34,7 +33,6 @@ namespace Microsoft.AspNetCore.Server.Kestrel.FunctionalTests
     public class RequestTests
     {
         private const int _connectionStartedEventId = 1;
-        private const int _connectionKeepAliveEventId = 9;
```
halter73: On the flip side this makes it slightly more obvious when we make a breaking change to logging by changing the event ids for the events that are expected on a given test.

PR #1870, opened 2017-05-30
Temporary workaround for GenerateResource task error on VS

cesarbs:
cc @dsplaisted @rainersigwald

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Kestrel.Core/Microsoft.AspNetCore.Server.Kestrel.Core.csproj
+++ b/src/Microsoft.AspNetCore.Server.Kestrel.Core/Microsoft.AspNetCore.Server.Kestrel.Core.csproj
@@ -10,8 +10,8 @@
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
     <NoWarn>CS1591;$(NoWarn)</NoWarn>
     <EnableApiCheck>false</EnableApiCheck>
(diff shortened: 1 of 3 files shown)


src/Microsoft.AspNetCore.Server.Kestrel.Core/Microsoft.AspNetCore.Server.Kestrel.Core.csproj
```diff
@@ -10,6 +10,8 @@
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
     <NoWarn>CS1591;$(NoWarn)</NoWarn>
     <EnableApiCheck>false</EnableApiCheck>
+    <!-- TODO: remove when https://github.com/Microsoft/msbuild/issues/1293 is fixed -->
+    <GenerateResourceMSBuildRuntime Condition="'$(TargetFrameworkIdentifier)' != '.NETFramework'">CurrentRuntime</GenerateResourceMSBuildRuntime>
```
rainersigwald: No need for the condition, since these projects don't target (desktop) .NET 2.0/3.5.

cesarbs:
@rainersigwald Thanks for the link to the SDK issue. I'll track it and remove this workaround once we start consuming an SDK with the fix.

@natemcmaster As discussed offline, I'll keep this here since we should be getting a fix in the SDK soon.

Issue #1871, opened 2017-05-30
Test failure on macOS: ConnectionCountingReturnsToZero 


cesarbs:
```
[xUnit.net 00:02:59.1255330]     ConnectionCountingReturnsToZero [FAIL]
[xUnit.net 00:02:59.1297990]       System.TimeoutException : The operation at /Users/travis/build/aspnet/KestrelHttpServer/test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/ConnectionLimitTests.cs:182 timed out after reaching the limit of 60000ms.
[xUnit.net 00:02:59.1326380]       Stack Trace:
[xUnit.net 00:02:59.1363270]            at Microsoft.AspNetCore.Testing.TaskExtensions.<TimeoutAfter>d__0`1.MoveNext()
[xUnit.net 00:02:59.1375630]         --- End of stack trace from previous location where exception was thrown ---
[xUnit.net 00:02:59.1378600]            at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
[xUnit.net 00:02:59.1380610]            at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
[xUnit.net 00:02:59.1383590]         /Users/travis/build/aspnet/KestrelHttpServer/test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/ConnectionLimitTests.cs(182,0): at Microsoft.AspNetCore.Server.Kestrel.FunctionalTests.ConnectionLimitTests.<ConnectionCountingReturnsToZero>d__3.MoveNext()
[xUnit.net 00:02:59.1387530]         --- End of stack trace from previous location where exception was thrown ---
[xUnit.net 00:02:59.1388940]            at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
[xUnit.net 00:02:59.1390480]            at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
[xUnit.net 00:02:59.1391820]         --- End of stack trace from previous location where exception was thrown ---
[xUnit.net 00:02:59.1393440]            at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
[xUnit.net 00:02:59.1398360]            at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
[...]

PR #1872, opened 2017-05-30
Decrease connections used in ConnectionCountingReturnsToZero for slow test agents

natemcmaster:
Fixes https://github.com/aspnet/KestrelHttpServer/issues/1871


natemcmaster:
FYI- test runs on Travis/macOS average 9-10 minutes while their Linux equivalent on Travis runs in 4-5 min.

PR #1874, opened 2017-05-31
Add request body minimum data rate feature

cesarbs:
TODO:

- [x] Deal with backpressure
- [x] Add tests for backpressure
- [x] Better logs
- [x] Tests
- [x] Validate timeout values on server init
- [x] Add feature to make timeout configurable per request
- [x] Only start timing reads after sending 100 Continue.
- [x] Don't enforce timeout on upgrade requests


src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/FrameConnection.cs
```diff
@@ -231,12 +235,34 @@ private void DisposeAdaptedConnections()
...
-            Debug.Assert(_frame != null, $"nameof({_frame}) is null");
+            Debug.Assert(_frame != null, $"{nameof(_frame)} is null");
 
             var timestamp = now.Ticks;
 
-            // TODO: Use PlatformApis.VolatileRead equivalent again
-            if (timestamp > Interlocked.Read(ref _timeoutTimestamp))
+            if (_meteringReads)
+            {
+                var elapsed = TimeSpan.FromTicks(timestamp - _meteringStartTicks);
```
davidfowl: Why is this logic in FrameConnection and not frame?

src/Microsoft.AspNetCore.Server.Kestrel.Core/RequestBodyTimeout.cs
```diff
@@ -0,0 +1,14 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+
+namespace Microsoft.AspNetCore.Server.Kestrel.Core
+{
+    public class RequestBodyTimeout
```
natemcmaster: Needs xml docs. The API names alone don't provide enough explanation about what these mean and how to set them.

Also, we should provide some validation of inputs, such as MaxTime must be > MinTime, MinimumRate must be >= 0, etc. Consider making the properties readonly and adding a constructor that does these validations.

src/Microsoft.AspNetCore.Server.Kestrel.Core/KestrelServerLimits.cs
```diff
@@ -207,5 +209,7 @@ public TimeSpan RequestHeadersTimeout
                 _maxConcurrentUpgradedConnections = value;
             }
         }
+
+        public RequestBodyTimeout DefaultRequestBodyTimeout => _defaultRequestBodyTimeout;
```
natemcmaster: xml docs here too

src/Microsoft.AspNetCore.Server.Kestrel.Core/RequestBodyTimeout.cs
cesarbs: Yep, I'll validate that in `KestrelServer` like we do with other limits.

cesarbs:
@Tratcher The extended timeout and minimum rate must be set at the same time. And I didn't want to allow for inconsistensies (e.g. extended timeout shorter than the "base" timeout).

cesarbs:
@Tratcher This is implementing the same as https://httpd.apache.org/docs/2.4/mod/mod_reqtimeout.html#requestreadtimeout.

cesarbs:
@Tratcher Commit 17ca98882a3e6cdc19171ab6b6ad22ef4929d064 disables the timeout on upgrade requests.

cesarbs:
Latest commit (5e90345b6e8cb0e406212854432a499e3a13d2ab) changes the behavior so far to send a 408 instead of aborting the connection.

Tratcher:
How does that help? Will you then try to drain the timed out request body at the end of the request processing?

Tratcher:
You still have to close the connection.

cesarbs:
@Tratcher It will either throw in the app (on a read) or when attempting to consume the rest of the request body (if the app catches the exception). Either way the connection will be closed.

src/Microsoft.AspNetCore.Server.Kestrel.Core/KestrelServer.cs
```diff
@@ -208,6 +208,19 @@ private void ValidateOptions()
...
             }
+
+            if (Options.Limits.DefaultRequestBodyExtendedTimeout.HasValue &&
+                Options.Limits.DefaultRequestBodyExtendedTimeout <= Options.Limits.DefaultRequestBodyTimeout)
+            {
+                throw new InvalidOperationException(
+                    CoreStrings.FormatRequestBodyExtendedTimeoutSmallerThanTimeout(Options.Limits.DefaultRequestBodyExtendedTimeout, Options.Limits.DefaultRequestBodyTimeout));
+            }
+
+            if (Options.Limits.DefaultRequestBodyExtendedTimeout.HasValue ^ Options.Limits.DefaultRequestBodyMinimumDataRate.HasValue)
```
natemcmaster: Since they are both required to be null or set, can we make this enforceable as a compile-time check? Makes the API slightly easier to use

Ideas:
```c#
// add a type
public class KestrelServerLimits
{
    public DataRateLimit DefaultRequestBodyRate { get; set; }
}

public class DataRateLimit
{
    public TimeSpan Timeout { get;set;}
    public double Rate { get; set; }
}

```
```c#
// add a tuple literal
public class KestrelServerLimits
{
    public (TimeSpan timespan, double rate) DefaultRequestBodyRate { get; set; }
}
```
```c#
// add a tuple type
public class KestrelServerLimits
{
    public Tuple<TimeSpan, double> DefaultRequestBodyRate { get; set; }
}
```

src/Microsoft.AspNetCore.Server.Kestrel.Core/Features/IHttpRequestBodyTimeoutFeature.cs
```diff
@@ -0,0 +1,46 @@
...
+        /// <summary>
+        /// The minimum amount of time allowed for the request body to be entirely read.
+        /// </summary>
+        TimeSpan Timeout { get; }
+
+        /// <summary>
+        /// The maximum amount of time allowed for the request body to be entirely read.
+        /// Only effective when <see cref="MinimumDataRate"/> is also set.
+        /// </summary>
+        TimeSpan? ExtendedTimeout { get; }
```
Tratcher: Still haven't resolved the redundancy between ExtendedTimeout and MinimumDataRate. MinimumDataRate scales with the size of the body, ExtendedTimeout doesn't. ExtendedTimeout will be a pain for any customer that tries to use it.

cesarbs:
Made changes to the API as discussed in person. Now we have:

- Request body timeout: max. amount of time allowed for the request body to be received
- Request body minimum data rate: self-explanatory
- Request body minimum data rate grace period: the initial period of time during which the minimum rate is not enforced

There was no need for a new data structure. I added the `SetDefaultRequestBodyMinimumDataRate(double, TimeSpan)` method to `KestrelServerLimits` to configure the default minimum rate, and changed `IHttpRequestBodyTimeoutFeature` to have the analogous `SetMinimumDataRate(double, TimeSpan)` method. The overall request body timeout is not a settable property in the feature.

src/Microsoft.AspNetCore.Server.Kestrel.Core/Features/IHttpRequestBodyTimeoutFeature.cs
```diff
@@ -0,0 +1,35 @@
...
+    /// </summary>
+    public interface IHttpRequestBodyTimeoutFeature
+    {
+        /// <summary>
+        /// The maximum amount of time in which the request body should be fully received.
+        /// </summary>
+        TimeSpan RequestBodyTimeout { get; set; }
+
+        /// <summary>
+        /// The minimum data rate at which the request body should be received.
```
natemcmaster: Would be good to specify the units. bytes/second? MB/min? etc.

src/Microsoft.AspNetCore.Server.Kestrel.Core/Features/IHttpRequestBodyTimeoutFeature.cs
```diff
@@ -0,0 +1,35 @@
...
+        /// </summary>
+        double MinimumDataRate { get; }
+
+        /// <summary>
+        /// The period of time during which the minimum data rate will not be enforced.
```
natemcmaster: Suggested change of wording: `The amount of time to delay enforcement of <see cref="MinimumDataRate" />. When set to <see cref="TimeSpan.Zero"/>, enforcement begins when the server starts reading the request body.`

src/Microsoft.AspNetCore.Server.Kestrel.Core/KestrelServerLimits.cs
```diff
@@ -207,5 +212,50 @@ public TimeSpan RequestHeadersTimeout
...
+
+        /// <summary>
+        /// Gets the default grace period after which the request body minimum data rate is enforced.
+        /// </summary>
+        public TimeSpan? DefaultRequestBodyMinimumDataRateGracePeriod => _defaultRequestBodyMinimumDataRateGracePeriod;
+
+        /// <summary>
+        /// Sets the default request body minimum data rate. No minimum data rate is enforced by default.
+        /// </summary>
+        public void SetDefaultRequestBodyMinimumDataRate(double minimumDataRate, TimeSpan gracePeriod)
```
natemcmaster: Add `<param>` elements to the docs

src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/Http/Frame.cs
```diff
@@ -298,6 +298,12 @@ private CancellationTokenSource RequestAbortedSource
 
         protected FrameResponseHeaders FrameResponseHeaders { get; } = new FrameResponseHeaders();
 
+        public TimeSpan RequestBodyTimeout { get; set; }
+
+        public double RequestBodyMinimumDataRate { get; set; }
+
+        public TimeSpan RequestBodyMinimumDataRateGracePeriod { get; set; }
+
         public void InitializeStreams(MessageBody messageBody)
```
davidfowl: nit: Probably should renamed to `InitializeBody`

src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/Http/MessageBody.cs
```diff
@@ -215,6 +245,26 @@ private void TryProduceContinue()
...
+            {
+                Log.RequestBodyStart(_context.ConnectionIdFeature, _context.TraceIdentifier);
+                _context.TimeoutControl.StartTimingReads();
+                _timingReads = true;
+            }
+        }
+
+        private void TryStopTimingReads()
+        {
+            if (!RequestUpgrade)
```
cesarbs: That bool is not necessary and I'm removing it.

cesarbs:
### Benchmarks

#### Plaintext

rel/2.0.0-preview2: 1201645.23
PR: 1194168.72

#### CopyToAsync

rel/2.0.0-preview2: 10706.16
PR: 10669.96

src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/FrameConnection.cs
```diff
@@ -231,21 +255,49 @@ private void DisposeAdaptedConnections()
...
+                    {
+                        Log.RequestBodyMininumDataRateNotSatisfied(_context.ConnectionId, _frame.TraceIdentifier, _frame.RequestBodyMinimumDataRate);
+                        Timeout();
+                    }
+                }
 
-                _frame.Stop();
+                // PauseTimingReads() cannot just set _timingReads to false. It needs to go through at least one tick
+                // before pausing, otherwise _readTimingElapsed might never be updated if PauseTimingReads() is always
+                // called before the next tick.
```
halter73: This means that elapsed time is being added while there is no read callback configured. This means about a second of elapsed time could be added for each pause when the client actually can't upload additional data.

If we error, we should error the other way.

src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/FrameConnection.cs
```diff
@@ -275,5 +327,35 @@ private void AssignTimeout(long ticks, TimeoutAction timeoutAction)
             // Add Heartbeat.Interval since this can be called right before the next heartbeat.
             Interlocked.Exchange(ref _timeoutTimestamp, _lastTimestamp + ticks + Heartbeat.Interval.Ticks);
         }
+
+        public void StartTimingReads()
+        {
+            Interlocked.Exchange(ref _readTimingElapsed, 0);
+            Interlocked.Exchange(ref _bytesRead, 0);
+            Interlocked.Exchange(ref _readTimingStartTicks, _lastTimestamp);
```
halter73: Why are these Interlocked.Exchange's helping? If we're worried about concurrent access/modification of these fields, wouldn't these fields being momentarily inconsistent with each other be a problem?

You can't safely do `+=` operations to fields that are reset concurrently even with Interlocked.Exchange, so `_readTimingElapsed += timestamp - _lastTimestamp;` seems unsafe. I don't think setting `_timingReads` to true last saves you here.

src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/FrameConnection.cs
```diff
@@ -275,5 +327,35 @@ private void AssignTimeout(long ticks, TimeoutAction timeoutAction)
...
+
+        public void StopTimingReads()
+        {
+            _timingReads = false;
+        }
+
+        public void PauseTimingReads()
+        {
+            Log.RequestBodyTimingPause(_context.ConnectionId, _frame.TraceIdentifier, TimeSpan.FromTicks(_readTimingElapsed), TimeSpan.FromTicks(_lastTimestamp - _readTimingStartTicks));
+            _pauseTimingReads = true;
```
halter73: I'm pretty sure you're going to need to add a lock to FrameConnection. The locks would be per-connection and should always be for quick operations, but I think we could use Monitor.TryEnter in Tick if need be.

src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/FrameConnection.cs
```diff
@@ -231,21 +255,49 @@ private void DisposeAdaptedConnections()
...
+                _readTimingElapsed += timestamp - _lastTimestamp;
+
+                if (_readTimingElapsed > _frame.RequestBodyTimeout.Ticks)
+                {
+                    Log.RequestBodyTimeout(_context.ConnectionId, _frame.TraceIdentifier, _frame.RequestBodyTimeout);
                     Timeout();
                 }
+                else if (_frame.RequestBodyMinimumDataRate > 0)
+                {
+                    var rate = (_bytesRead / (_readTimingElapsed / TimeSpan.TicksPerSecond));
```
halter73: I would clarify this line of code by splitting it into a couple lines, ex:

```
var elapsedSecondsFloored = _readTimingElapsed / TimeSpan.TicksPerSecond;
var rate = _bytesRead / elapsedSecondsFloored ;
```

Technically the rate is floored too which isn't exactly ideal, but bytes/second is a precise enough measurement that I doubt it matters.

src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/FrameConnection.cs
cesarbs: Yeah these are unnecessary. Setting `_timingReads` last should help since the other properties won't be read until that is true.

src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/FrameConnection.cs
cesarbs: Doesn't need to be floored really, I'll cast to `double`.

src/Microsoft.AspNetCore.Server.Kestrel.Core/Features/IHttpRequestBodyTimeoutFeature.cs
```diff
@@ -0,0 +1,39 @@
...
+{
+    /// <summary>
+    /// Represents a timeout for the request body in an HTTP request.
+    /// </summary>
+    public interface IHttpRequestBodyTimeoutFeature
+    {
+        /// <summary>
+        /// The maximum amount of time in which the request body should be fully received.
+        /// </summary>
+        TimeSpan RequestBodyTimeout { get; set; }
```
Tratcher: Are you mirroring apache's incremental timeout behavior?
"If the client sends data, increase the timeout by 1 second for every 1000 bytes received, with no upper limit for the timeout" - https://httpd.apache.org/docs/2.4/mod/mod_reqtimeout.html
Otherwise this feature is going to be very difficult to use.

cesarbs: Yes. We check the rate every second and time out if it drops below the minimum required, so that's effectively the same thing mod_reqtimeout does.

Tratcher: That doesn't sound the same at all. apache's full body timeout gets longer so long as data is still being received.

src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/FrameConnection.cs
```diff
@@ -242,10 +267,40 @@ public void Tick(DateTimeOffset now)
...
+                    SetTimeoutResponse();
+                }
+
+                Timeout();
+            }
+            else if (_timingReads)
+            {
+                _readTimingElapsed += timestamp - _lastTimestamp;
+
+                if (_readTimingElapsed > _frame.RequestBodyTimeout.Ticks)
```
Tratcher: Where do you grow RequestBodyTimeout like Apache's?

halter73: My understanding is that RequestBodyTimeout is unaffected by rate. This is the absolute max timeout. If you want a timeout that grows with more data, you could just set a rate and leave this unlimited.

Tratcher: Ah, those docs are just talking about the rate.

I don't see how an absolute timeout is workable for variable payload sizes. You'd always have to change the size limit and the time limit in tandem, and the time limit is only a multiple of the data rate and the size limit. What value does RequestBodyTimeout add beyond the size and rate limits?

halter73: This along with RequestHeadersTimeout gives the developer a simple/convenient way to set an absolute timeout for a client to send a complete request. Even if it's possible to achieve the same thing by doing some math and selecting the appropriate min rate and max size, that seems more complicated than necessary for many scenarios.

src/Microsoft.AspNetCore.Server.Kestrel.Core/KestrelServerLimits.cs
```diff
@@ -207,5 +210,26 @@ public TimeSpan RequestHeadersTimeout
...
+
+        /// <summary>
+        /// Gets or sets the default timeout period for receiving the request body. Defaults to 2 minutes.
+        /// </summary>
+        public TimeSpan DefaultRequestBodyTimeout
+        {
+            get => _defaultRequestBodyTimeout;
+            set
+            {
+                if (value <= TimeSpan.Zero)
```
halter73: Shouldn't we support `Timeout.InfiniteTimeSpan`? `Timeout.InfiniteTimeSpan <= TimeSpan.Zero` BTW.

halter73: Add this check to KeepAliveTimeout and RequestHeadersTimeout while you're at it.

src/Microsoft.AspNetCore.Server.Kestrel.Core/KestrelServerLimits.cs
```diff
@@ -207,5 +210,26 @@ public TimeSpan RequestHeadersTimeout
                 _maxConcurrentUpgradedConnections = value;
             }
         }
+
+        /// <summary>
+        /// Gets or sets the default timeout period for receiving the request body. Defaults to 2 minutes.
+        /// </summary>
```
halter73: Nit: Default values are in the remarks section of the other properties.

src/Microsoft.AspNetCore.Server.Kestrel.Core/KestrelServerLimits.cs
```diff
@@ -207,5 +210,26 @@ public TimeSpan RequestHeadersTimeout
...
+        public TimeSpan DefaultRequestBodyTimeout
```
halter73: Rename to just RequestBodyTimeout since it can be modified to a non-default value. I get that this value can be modified on a per-request basis, but that can be called out in the doc comments [like this](https://github.com/aspnet/KestrelHttpServer/blob/3dbfb003bb064e117b52d6336ee145cf602363ea/src/Microsoft.AspNetCore.Server.Kestrel.Core/KestrelServerLimits.cs#L145). This is better because it explains how to change the "default".

src/Microsoft.AspNetCore.Server.Kestrel.Core/KestrelServerLimits.cs
```diff
@@ -207,5 +210,26 @@ public TimeSpan RequestHeadersTimeout
...
+                    throw new ArgumentOutOfRangeException(nameof(value), CoreStrings.PositiveTimeSpanRequired);
+                }
+                _defaultRequestBodyTimeout = value;
+            }
+        }
+
+        /// <summary>
+        /// Gets or sets the default request body minimum data rate in bytes/second.
+        /// </summary>
+        public MinimumDataRate DefaultRequestBodyMinimumDataRate { get; set; } = null;
```
halter73: Also remove "Default" from this property name.

src/Microsoft.AspNetCore.Server.Kestrel.Core/KestrelServerLimits.cs
```diff
@@ -207,5 +210,26 @@ public TimeSpan RequestHeadersTimeout
...
+                {
+                    throw new ArgumentOutOfRangeException(nameof(value), CoreStrings.PositiveTimeSpanRequired);
+                }
+                _defaultRequestBodyTimeout = value;
+            }
+        }
+
+        /// <summary>
+        /// Gets or sets the default request body minimum data rate in bytes/second.
+        /// </summary>
```
halter73: The summary should mention that null means there is no minimum rate, and add a remark indicating that null is the default value.

src/Microsoft.AspNetCore.Server.Kestrel.Core/Features/IHttpRequestBodyTimeoutFeature.cs
cesarbs: Up to a limit. You give it a range (e.g. 20-40 seconds) and it will increase the timeout by 1 second up to the upper bound as long as data is arriving at the specified rate.

src/Microsoft.AspNetCore.Server.Kestrel.Core/KestrelServerLimits.cs
Tratcher: Timeout.InfiniateTimeSpan (-1s) to disable?

src/Microsoft.AspNetCore.Server.Kestrel.Core/KestrelServerLimits.cs
Tratcher: Timeout.InfiniteTimeSpan (-1s) to disable?

src/Microsoft.AspNetCore.Server.Kestrel.Core/KestrelServerLimits.cs
Tratcher: The spec calls for a non-null default.

src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/FrameConnection.cs
Tratcher: You have to do the math anyways don't you? When you set this you have to make sure you allow enough time for your expected data size to arrive at the expected rate. RequestBodyTimeout is completely redundant.

src/Microsoft.AspNetCore.Server.Kestrel.Core/KestrelServerLimits.cs
cesarbs: What spec?

src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/FrameConnection.cs
halter73: You don't have to do the math. You could just limit one of rate/timeout/size and set the others unlimited if that's want you want.

src/Microsoft.AspNetCore.Server.Kestrel.Core/KestrelServerLimits.cs
Tratcher: https://github.com/aspnet/Coherence-Signed/issues/427 calls for a default read timeout. You've replaced that with a data rate, but the rate needs to be enabled by default still.

src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/FrameConnection.cs
Tratcher: I'll accept as much data as you can send me in 60s? I don't see that as a realistic scenario.

A full request/response processing time limit seems more realistic. E.g. my SLA for this request is 1s, and if I can't fulfill it I need to cancel.

src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/FrameConnection.cs
```diff
@@ -242,10 +268,46 @@ public void Tick(DateTimeOffset now)
...
+                                Timeout();
+                            }
+                        }
+
+                        // PauseTimingReads() cannot just set _timingReads to false. It needs to go through at least one tick
+                        // before pausing, otherwise _readTimingElapsed might never be updated if PauseTimingReads() is always
+                        // called before the next tick.
+                        if (_readTimingPauseRequested)
+                        {
+                            _readTimingEnabled = false;
```
halter73: What happens if both PauseTimingReads and ResumeTimingReads are called between a single tick?.

cesarbs: fccfa37d24e52a9aede689f0b6bcf208f50f403b

cesarbs:
Updated to remove request body timeout, leaving only the min rate stuff in there.

src/Microsoft.AspNetCore.Server.Kestrel.Core/KestrelServerLimits.cs
```diff
@@ -234,5 +250,16 @@ public TimeSpan RequestHeadersTimeout
...
+        /// <summary>
+        /// Gets or sets the request body minimum data rate in bytes/second.
+        /// Setting this property to null indicates no minimum data rate should be enforced.
+        /// This limit has no effect on upgraded connections which are always unlimited.
+        /// This can be overridden per-request via <see cref="IHttpRequestBodyMinimumDataRateFeature"/>.
+        /// </summary>
+        /// <remarks>
+        /// Defaults to null.
+        /// </remarks>
+        public MinimumDataRate RequestBodyMinimumDataRate { get; set; } = new MinimumDataRate(1, TimeSpan.FromSeconds(5));
```
Tratcher: use named parameters when passing in constants

src/Microsoft.AspNetCore.Server.Kestrel.Core/MinimumDataRate.cs
```diff
@@ -0,0 +1,41 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+
+namespace Microsoft.AspNetCore.Server.Kestrel.Core
+{
+    public class MinimumDataRate
```
Tratcher: Put this in the feature namespace?

src/Microsoft.AspNetCore.Server.Kestrel.Core/Features/IHttpRequestBodyMinimumDataRateFeature.cs
```diff
@@ -0,0 +1,18 @@
...
+    /// Represents a minimum data rate for the request body of an HTTP request.
+    /// </summary>
+    public interface IHttpRequestBodyMinimumDataRateFeature
+    {
+        /// <summary>
+        /// The minimum data rate in bytes/second at which the request body should be received.
+        /// Setting this property to null indicates no minimum data rate should be enforced.
+        /// This limit has no effect on upgraded connections which are always unlimited.
+        /// </summary>
+        MinimumDataRate MinimumDataRate { get; set; }
```
Tratcher: Design note: Using local concrete data types on a feature abstraction makes it hard to generalize / replace. This is tolerable so long as this remains a kestrel exclusive feature, but we'd need a different design if this were ever moved to HttpAbstractions.

cesarbs: Couldn't we move `MinimumDataRate` to HttpAbstractions if that were the case? But yeah this is what I had in mind when I was opposing the new data type.

Tratcher: Yes, the type could move with it, but it's existance still limits the extensibility of the feature. I think we only have one other local concrete type exposed from a feature, CookieOptions.

src/Microsoft.AspNetCore.Server.Kestrel.Core/KestrelServerLimits.cs
```diff
@@ -169,7 +171,14 @@ public int MaxRequestHeaderCount
         public TimeSpan KeepAliveTimeout
         {
             get => _keepAliveTimeout;
-            set => _keepAliveTimeout = value;
+            set
+            {
+                if (value <= TimeSpan.Zero && value != Timeout.InfiniteTimeSpan)
```
halter73: InfiniteTimeSpan needs to be special cased in the timeout logic. I'd add a functional test that verifies that connections aren't killed immediate when InfiniteTimeSpan is configured.

src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/FrameConnection.cs
```diff
@@ -29,13 +29,25 @@ public class FrameConnection : IConnectionContext, ITimeoutControl
         private long _timeoutTimestamp = long.MaxValue;
         private TimeoutAction _timeoutAction;
 
+        private object _readTimingLock = new object();
+        private bool _readTimingEnabled;
+        private bool _readTimingPauseRequested;
+        private long _readTimingStartTicks;
```
halter73: No longer needed.

src/Microsoft.AspNetCore.Server.Kestrel.Core/KestrelServerLimits.cs
cesarbs: I'm checking for it and setting to `TimeSpan.MaxValue`. Added unit tests.

PR #1877, opened 2017-06-01
Add a request body size limit

halter73:
- Add IHttpMaxRequestBodySizeFeature
- Improve TestConnection.Receive timeout errors


src/Microsoft.AspNetCore.Server.Kestrel.Core/KestrelServerLimits.cs
```diff
@@ -20,6 +20,9 @@ public class KestrelServerLimits
         // Matches the default large_client_header_buffers in nginx.
         private int _maxRequestHeadersTotalSize = 32 * 1024;
 
+        // Disabled by default.
+        private long? _maxRequestBodySize = null;
```
davidfowl: good default 😉

src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/Http/MessageBody.cs
```diff
@@ -499,13 +528,14 @@ private void ParseChunkedPrefix(ReadableBuffer buffer, out ReadCursor consumed,
...
 
-                do
+                while (reader.ConsumedBytes < 10)
                 {
                     if (ch1 == ';')
                     {
                         consumed = reader.Cursor;
                         examined = reader.Cursor;
 
+                        AddAndCheckConsumedBytes(reader.ConsumedBytes);
```
Tratcher: Counting the framing is going to cause strange inconsistencies between Chunked, Content-Length, and HttpSysServer. @shirhatti I assume IIS does not count the chunked framing because it doesn't even see it, Http.Sys removes it, right?

src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/Http/MessageBody.cs
```diff
@@ -548,39 +574,45 @@ private void ParseExtension(ReadableBuffer buffer, out ReadCursor consumed, out
...
                 consumed = buffer.Start;
                 examined = buffer.Start;
+
                 do
                 {
                     ReadCursor extensionCursor;
                     if (ReadCursorOperations.Seek(buffer.Start, buffer.End, out extensionCursor, ByteCR) == -1)
                     {
                         // End marker not found yet
+                        CheckExaminedBytes(buffer.Length);
```
Tratcher: This is the wrong limit to apply to extensions. Let's review offline.

test/Microsoft.AspNetCore.Server.Kestrel.Core.Tests/MessageBodyTests.cs
```diff
@@ -130,24 +116,79 @@ public class MessageBodyTests
...
+            using (var input = new TestInput())
+            {
+                var body = MessageBody.For(HttpVersion.Http11, new FrameRequestHeaders { HeaderTransferEncoding = "chunked" }, input.FrameContext);
+                var stream = new FrameRequestStream();
+                stream.StartAcceptingReads(body);
+
+                input.Add("80000000\r\n");
+
+                var buffer = new byte[1024];
+                var ex = await Assert.ThrowsAsync<OverflowException>(async () =>
```
Tratcher: It should be wrapped in an IOException

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/MaxRequestBodySizeTests.cs
```diff
@@ -0,0 +1,384 @@
...
+                using (var connection = server.CreateConnection())
+                {
+                    await connection.Send(
+                        "POST / HTTP/1.1",
+                        "Host:",
+                        "Content-Length: " + (globalMaxRequestBodySize + 1),
+                        "",
+                        "");
+                    await connection.ReceiveForcedEnd(
+                        "HTTP/1.1 413 Payload Too Large",
```
Tratcher: The BadHttpRequestException is catchable by the app, correct? You only get this 413 if they do not catch it and it bubbles back out from the app pipeline?

src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/Http/MessageBody.cs
```diff
@@ -215,7 +213,24 @@ private void TryProduceContinue()
             }
         }
 
-        protected abstract bool Read(ReadableBuffer readableBuffer, WritableBuffer writableBuffer, out ReadCursor consumed, out ReadCursor examined);
+        private void TryInit()
+        {
+            if (!_context.HasStartedConsumingRequestBody)
+            {
+                _context.HasStartedConsumingRequestBody = true;
+                OnReadStart();
```
davidfowl: Start was cleaner tbh. I understand why it was removed but it meant we avoid this lazy initialize once logic.

src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/Http/MessageBody.cs
```diff
@@ -378,6 +383,14 @@ protected override bool Read(ReadableBuffer readableBuffer, WritableBuffer writa
 
                 return _inputLength == 0;
             }
+
+            protected override void OnReadStart()
+            {
+                if (_contentLength > _context.MaxRequestBodySize)
+                {
+                    _context.RejectRequest(RequestRejectionReason.RequestBodyTooLarge);
```
davidfowl: This is typical kestrel code though.

benaadams:
Have déjà vu 😄 

Good to standardize it as a common cross server feature though

src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/Http/Frame.cs
```diff
@@ -1370,9 +1372,9 @@ private IPipe CreateRequestBodyPipe()
             => ConnectionInformation.PipeFactory.Create(new PipeOptions
             {
                 ReaderScheduler = ServiceContext.ThreadPool,
-                WriterScheduler = ConnectionInformation.InputWriterScheduler,
-                MaximumSizeHigh = ServiceContext.ServerOptions.Limits.MaxRequestBufferSize ?? 0,
-                MaximumSizeLow = ServiceContext.ServerOptions.Limits.MaxRequestBufferSize ?? 0
+                WriterScheduler = InlineScheduler.Default,
+                MaximumSizeHigh = 1,
+                MaximumSizeLow = 1
```
cesarbs: Why?

src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/Http/MessageBody.cs
```diff
@@ -215,7 +213,24 @@ private void TryProduceContinue()
...
         }
 
-        protected abstract bool Read(ReadableBuffer readableBuffer, WritableBuffer writableBuffer, out ReadCursor consumed, out ReadCursor examined);
+        private void TryInit()
+        {
+            if (!_context.HasStartedConsumingRequestBody)
+            {
+                _context.HasStartedConsumingRequestBody = true;
+                OnReadStart();
+                var ignore = PumpAsync();
```
pakrym: Observe exceptions?

src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/Http/MessageBody.cs
```diff
@@ -499,13 +528,14 @@ private void ParseChunkedPrefix(ReadableBuffer buffer, out ReadCursor consumed,
                 var chunkSize = CalculateChunkSize(ch1, 0);
                 ch1 = ch2;
 
-                do
+                while (reader.ConsumedBytes < 10)
```
natemcmaster: can you give this magic number a name by setting it as a const?

src/Microsoft.AspNetCore.Server.Kestrel.Core/KestrelServerLimits.cs
```diff
@@ -134,6 +137,28 @@ public int MaxRequestHeaderCount
...
+        /// </summary>
+        /// <remarks>
+        /// Defaults to null (unlimited).
+        /// </remarks>
+        public long? MaxRequestBodySize
+        {
+            get => _maxRequestBodySize;
+            set
+            {
+                if (value <= 0)
```
natemcmaster: Quirky, but technically 0 is a valid limit, right?

src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/Http/IHttpMaxRequestBodySizeFeature.cs
```diff
@@ -0,0 +1,23 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
+{
+    /// <summary>
+    ///
```
natemcmaster: Nit: add a description

src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/Http/IHttpMaxRequestBodySizeFeature.cs
```diff
@@ -0,0 +1,23 @@
...
+{
+    /// <summary>
+    /// 
+    /// </summary>
+    public interface IHttpMaxRequestBodySizeFeature
+    {
+        // TODO: Explain how to override
+        /// <summary>
+        /// Gets or sets the maximum allowed size of the current request body in bytes.
+        /// When set to null, the maximunm request body size is unlimited.
```
natemcmaster: Typo: maximum

src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/Http/MessageBody.cs
natemcmaster: I've been misled by this too. At one point, someone compared this to explicit throws (e.g. `throw MethodThatCreatesException()`) and concluded that it doesn't inline as well, enough that it hurts perf. I think that was @pakrym or @benaadams

src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/Http/MessageBody.cs
halter73: Yep. I definitely understand avoiding lazy initialization is cleaner. It's not even too inefficient to greedily start the pump now that the "max" size of the body pipe is 1 byte.

The problem is that the lazy init logic was required for the request feature to work nicely with Content-Length bodies, and if you are going to do some check only on the first read anyway, it's no longer cleaner to greedily start the pump.

src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/Http/MessageBody.cs
halter73: I was tempted to change the name of RejectRequest to ThrowBadHttpRequestException as part of this change. I could still do that if we're in agreement that it makes this and all the other calling code at least slightly more readable.

src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/Http/MessageBody.cs
halter73: It's super trivial to change this to just count the "data" bytes if that's what we decide. I decided to do the more difficult thing first.

I'm not too concerned about breaking any apps by counting extension bytes towards the request body size limit since chunk extensions are so rare. Counting the framing bytes might make Kestrel behave differently that HttpSysServer, but the framing can easily take up to ~92% of the uploaded bytes without any chunk extensions or trailing headers.  We can discuss offline how much we care about that.

test/Microsoft.AspNetCore.Server.Kestrel.Core.Tests/MessageBodyTests.cs
halter73: This is just a test I added for existing behavior, so *technically* that's a breaking change but not really in practice.

src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/Http/MessageBody.cs
halter73: PumpAsync is wrapped in a giant try/catch which will Complete the body pipe writer with the error, so the app can observe the exception. I'll add a comment though.

src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/Http/IHttpMaxRequestBodySizeFeature.cs
halter73: Got lazy since I knew this would be moved to HttpAbstractions 😄

src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/Http/MessageBody.cs
benaadams: `RejectRequestAndThrow`?

benaadams: Renaming to `ThrowXxx` probably makes sense to indicate control is not returned to caller.

src/Microsoft.AspNetCore.Server.Kestrel.Core/KestrelServerLimits.cs
```diff
@@ -134,6 +137,28 @@ public int MaxRequestHeaderCount
...
         /// <summary>
+        /// Gets or sets the maximum allowed size of any request body in bytes.
+        /// When set to null, the maximunm request body size is unlimited.
+        /// This limit has no affect on upgraded connections which are always unlmited.
+        /// This can be overridden by <see cref="Internal.Http.IHttpMaxRequestBodySizeFeature"/>
+        /// </summary>
+        /// <remarks>
+        /// Defaults to null (unlimited).
+        /// </remarks>
+        public long? MaxRequestBodySize
```
cesarbs: Add tests for default value, valid values and invalid values in `KestrelServerLimitsTests`.

src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/Http/Frame.FeatureCollection.cs
```diff
@@ -202,6 +203,24 @@ string IHttpRequestIdentifierFeature.TraceIdentifier
             set => TraceIdentifier = value;
         }
 
+        long? IHttpMaxRequestBodySizeFeature.MaxRequestBodySize
```
cesarbs: Add tests in `FrameTests`.

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/MaxRequestBodySizeTests.cs
halter73: Mostly correct. All future reads would also fail, so even if the app swallowed all exceptions and didn't write to the response body, the request draining code would see the exception which would still trigger a 413. If the app writes to the response body before the request is drained (or it's a non-keepalive request that doesn't get drained), then there might not be a 413 response.

halter73:
⬆️ 📅

src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/Http/Frame.cs
natemcmaster: +1, can you explain the change?

src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/Http/MessageBody.cs
Tratcher: Doesn't look like this got fixed.

src/Microsoft.AspNetCore.Server.Kestrel.Core/KestrelServerLimits.cs
```diff
@@ -20,6 +21,10 @@ public class KestrelServerLimits
         // Matches the default large_client_header_buffers in nginx.
         private int _maxRequestHeadersTotalSize = 32 * 1024;
 
+        // Matches the default maxAllowedContentLength in IIS (~28.6 MB)
+        // https://www.iis.net/configreference/system.webserver/security/requestfiltering/requestlimits#005
+        private long? _maxRequestBodySize = 30000000;
```
Tratcher: We decided on 16mb in the meeting

halter73: I remember deciding the highest default of IIS/nginx which would be this.

halter73:
```
> dotnet run -- -n CopyToAsync  -o Benchmarks@halter73/react-to-logging -o KestrelHttpServer@rel/2.0.0-preview2
[05:40:07.608] Starting scenario CopyToAsync on benchmark server...
[05:40:30.061] Warmup
[05:40:30.102] Starting scenario CopyToAsync on benchmark client...
[05:40:47.397] Scenario CopyToAsync completed on benchmark client
[05:40:47.397] RPS: 10562.37
[05:40:47.398] Stopping scenario CopyToAsync on benchmark client...
[05:40:47.407] Benchmark
[05:40:47.408] Starting scenario CopyToAsync on benchmark client...
[05:41:03.579] Scenario CopyToAsync completed on benchmark client
[05:41:03.579] RPS: 10638.63
[05:41:03.580] Stopping scenario CopyToAsync on benchmark client...
[05:41:03.583] Stopping scenario CopyToAsync on benchmark server...

> dotnet run -- -n CopyToAsync -o Benchmarks@halter73/react-to-logging -o KestrelHttpServer@halter73/max-read
[05:56:08.048] Starting scenario CopyToAsync on benchmark server...
[05:56:31.534] Warmup
[05:56:31.562] Starting scenario CopyToAsync on benchmark client...
[05:56:47.802] Scenario CopyToAsync completed on benchmark client
[05:56:47.802] RPS: 10438.51
[05:56:47.803] Stopping scenario CopyToAsync on benchmark client...
[05:56:47.806] Benchmark
[05:56:47.806] Starting scenario CopyToAsync on benchmark client...
[05:57:04.024] Scenario CopyToAsync completed on benchmark client
[05:57:04.024] RPS: 10563.52
[05:57:04.025] Stopping scenario CopyToAsync on benchmark client...
[05:57:04.030] Stopping scenario CopyToAsync on benchmark server...
```

src/Microsoft.AspNetCore.Server.Kestrel.Core/KestrelServerLimits.cs
Tratcher: No, go read the meeting notes.

halter73: I see it in the meeting notes, but there's no reasoning. The decision was to use the larger default which is ~28.6 MB. I think that was written before we looked up what the largest default was. I can change it in the notes if you think that helps.

Tratcher: @DamianEdwards @blowdart

src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/Http/Frame.cs
halter73: This prevents double buffering. Since the request body is only exposed directly as a Stream, it's always greedily consumed (i.e. it's impossible to inspect data from the request body via a Stream without consuming the data from the pipe). This means there's no functional reason to buffer here.

There could theoretically be a perf impact which is why I posted the CopyToAsync benchmark results. This is still pretty efficient since backpressure doesn't get applied to the client until the connection's Input pipe buffers up to MaxRequestBufferSize.

src/Microsoft.AspNetCore.Server.Kestrel.Core/KestrelServerLimits.cs
Tratcher: To clarify, clear it with Damian and Barry, don't just change it.

src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/Http/Frame.FeatureCollection.cs
```diff
@@ -202,6 +203,28 @@ string IHttpRequestIdentifierFeature.TraceIdentifier
...
+        {
+            get => MaxRequestBodySize;
+            set
+            {
+                if (HasStartedConsumingRequestBody)
```
davidfowl: How does the application check this to make sure it's safe to do to avoid throwing?

src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/Http/Frame.FeatureCollection.cs
```diff
@@ -202,6 +203,28 @@ string IHttpRequestIdentifierFeature.TraceIdentifier
...
+                {
+                    throw new InvalidOperationException(CoreStrings.MaxRequestBodySizeCannotBeModifiedAfterRead);
+                }
+                if (_wasUpgraded)
```
davidfowl: Same here.

src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/Http/Frame.cs
davidfowl: In theory, this should really be set to `MaxRequestBufferSize` and this limit should be decided by the transport. Basically our limits for the server shouldn't apply here. The problem is that the request is split for us between start line and headers and the body size...

src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/Http/MessageBody.cs
```diff
@@ -215,7 +213,24 @@ private void TryProduceContinue()
...
         }
 
-        protected abstract bool Read(ReadableBuffer readableBuffer, WritableBuffer writableBuffer, out ReadCursor consumed, out ReadCursor examined);
+        private void TryInit()
+        {
+            if (!_context.HasStartedConsumingRequestBody)
+            {
+                OnReadStart();
+                _context.HasStartedConsumingRequestBody = true;
+                var ignore = PumpAsync();
```
davidfowl: nit: _ = PumpAsync()

src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/Http/Frame.cs
halter73: @davidfowl Are you saying that ideally the request body Pipe would buffer instead of the connection "Input" Pipe?

src/Microsoft.AspNetCore.Server.Kestrel.Core/KestrelServerLimits.cs
halter73: Done

src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/Http/Frame.FeatureCollection.cs
halter73: Done

src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/Http/Frame.FeatureCollection.cs
halter73: Right now, the app has to keep track of whether or not it's read from the request body itself. I know this might break down a little if you are writing some reusable component where you want to change the upload limit, but only if nothing has been uploaded yet. That seems pretty niche though, and nothing uses this feature today ofc.

Do you want to add HasStartedConsumingRequestBody to some request feature? Do you want to do the same for WasUpgraded?

davidfowl: Yes, the problem is you can add middleware anywhere in the pipeline that reads the body. There needs to be a way to not throw here.

> Do you want to add HasStartedConsumingRequestBody to some request feature? Do you want to do the same for WasUpgraded?

Yes, or something more generally named like our other bools of this nature (HasResponseStarted etc).

Tratcher: Could add it to the new IHttpMaxRequestBodySizeFeature feature

davidfowl: Both of those names are terrible but I agree with the sentiment.

src/Microsoft.AspNetCore.Server.Kestrel.Core/KestrelServerLimits.cs
```diff
@@ -134,6 +139,28 @@ public int MaxRequestHeaderCount
         }
 
         /// <summary>
+        /// Gets or sets the maximum allowed size of any request body in bytes.
+        /// When set to null, the maximum request body size is unlimited.
+        /// This limit has no affect on upgraded connections which are always unlimited.
+        /// This can be overridden by <see cref="IHttpMaxRequestBodySizeFeature"/>.
```
cesarbs: I would change this to "This can be overridden per-request via <see cref="IHttpMaxRequestBodySizeFeature"/>."

src/Microsoft.AspNetCore.Server.Kestrel.Core/KestrelServerLimits.cs
```diff
@@ -134,6 +139,28 @@ public int MaxRequestHeaderCount
         }
 
         /// <summary>
+        /// Gets or sets the maximum allowed size of any request body in bytes.
+        /// When set to null, the maximum request body size is unlimited.
+        /// This limit has no affect on upgraded connections which are always unlimited.
```
cesarbs: s/affect/effect

Issue #1875, opened 2017-05-31
Update kestrel config binding


HaoK:
1. Switch to ConfigureDefaultOptions instead of IConfigureOptions
2. Bind to section "Microsoft:AspNetCore:Hosting:Kestrel" instead of just Kestrel

cesarbs:
Did you mean to file this in MetaPackages?

Tratcher:
No, it's moving to Kestrel.

PR #1878, opened 2017-06-01
Add a Configuration package and tests.

Tratcher:
#1875, https://github.com/aspnet/Hosting/pull/1095
This moves config code from Microsoft.AspNetCore.dll down to Kestrel's own config package. A reaction PR will follow for MetaPackages.
The cross targeting is due to the src package.


src/Microsoft.AspNetCore.Server.Kestrel.Configuration/KestrelConfigurationExtensions.cs
```diff
@@ -0,0 +1,34 @@
...
+        /// <summary>
+        /// Specify Kestrel as the server to be used by the web host and bind its settings from configuration.
+        /// </summary>
+        /// <param name="hostBuilder">
+        /// The Microsoft.AspNetCore.Hosting.IWebHostBuilder to configure.
+        /// </param>
+        /// <returns>
+        /// The Microsoft.AspNetCore.Hosting.IWebHostBuilder.
+        /// </returns>
+        public static IWebHostBuilder UseKestrelWithConfiguration(this IWebHostBuilder hostBuilder)
```
davidfowl: What is this?

src/Microsoft.AspNetCore.Server.Kestrel.Configuration/KestrelServerOptionsSetup.cs
```diff
@@ -0,0 +1,111 @@
...
+using Microsoft.AspNetCore.Hosting;
+using Microsoft.AspNetCore.Server.Kestrel.Core;
+using Microsoft.Extensions.Configuration;
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Options;
+using Microsoft.Extensions.Options.Infrastructure;
+
+namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal
+{
+    internal class KestrelServerConfigureOptions : ConfigureDefaultOptions<KestrelServerOptions>
```
davidfowl: Make this public.

src/Microsoft.AspNetCore.Server.Kestrel.Configuration/KestrelConfigurationExtensions.cs
HaoK: Is there no way to takeover the default UseKestrel() which would make it match all of the other AddXyz()'s that also add default AddConfigureOptions?

Can we just add a reference to Microsoft.AspNetCore.Certificates.Configuration.Sources and ConfigurationBinder to Microsoft.AspNetCore.Server.Kestrel and do this inside of UseKestrel?

Tratcher: Not without pulling in the .Https dependency.

davidfowl: Lets just make the .Kestrel package the uber one and nuke this package.

src/Microsoft.AspNetCore.Server.Kestrel.Configuration/KestrelServerOptionsSetup.cs
HaoK: Do we want all of these `ConfigureDefaultOptions` public?   Or just Kestrel's since its doing real work unlike the rest which are usually one liners that only call GetSection().Bind which are mostly private.

davidfowl: It should be pubternal yea.

HaoK: Ok I'll file an issue to make them all pubternal post preview2 if I don't end up touching them again this week.  We need to standardize the naming then.

`ConfigureDefaultXyzOptions`?

davidfowl: .Internal

src/Microsoft.AspNetCore.Server.Kestrel/Microsoft.AspNetCore.Server.Kestrel.csproj
```diff
@@ -4,19 +4,22 @@
...
-    <TargetFramework>netstandard2.0</TargetFramework>
+    <TargetFrameworks>netstandard2.0;netcoreapp2.0</TargetFrameworks>
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
     <PackageTags>aspnetcore;kestrel</PackageTags>
     <NoWarn>CS1591;$(NoWarn)</NoWarn>
     <EnableApiCheck>false</EnableApiCheck>
   </PropertyGroup>
 
   <ItemGroup>
+    <PackageReference Include="Microsoft.AspNetCore.Certificates.Configuration.Sources" Version="$(AspNetCoreVersion)" />
```
davidfowl: PrivateAssets=All

src/Microsoft.AspNetCore.Server.Kestrel/WebHostBuilderKestrelExtensions.cs
```diff
@@ -28,6 +29,7 @@ public static IWebHostBuilder UseKestrel(this IWebHostBuilder hostBuilder)
             return hostBuilder.ConfigureServices(services =>
             {
                 services.AddTransient<IConfigureOptions<KestrelServerOptions>, KestrelServerOptionsSetup>();
+                services.AddTransient<IConfigureOptions<KestrelServerOptions>, KestrelServerConfigureOptions>();
```
davidfowl: This doesn't look right.

Tratcher: Doing it this way temporarily to work around https://github.com/aspnet/DependencyInjection/issues/500

davidfowl: @HaoK are all of the other sub systems doing this for preview2?

HaoK: That works, the other alternative would be to register

`services.AddTransient<IConfigureOptions<KestrelServerOptions>, ConfigureDefaults<KestrelServerConfigureOptions>()`

Which would let us just remove this line once the DI is fixed.

For preview 2 I don't think it really matters which one we do.  Its slightly easier to cleanup if we go with lining up all the defaults for now.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/KestrelServerConfigureOptions.cs
```diff
@@ -0,0 +1,110 @@
...
+            KestrelServerOptions options,
+            IConfigurationSection endPoint,
+            CertificateLoader certificateLoader)
+        {
+            var configAddress = endPoint.GetValue<string>("Address");
+            var configPort = endPoint.GetValue<string>("Port");
+
+            if (!IPAddress.TryParse(configAddress, out var address))
+            {
+                throw new InvalidOperationException($"Invalid IP address in configuration: {configAddress}");
```
cesarbs: User facing strings should go in the resx file.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/KestrelServerConfigureOptions.cs
```diff
@@ -0,0 +1,110 @@
...
+                {
+                    try
+                    {
+                        try
+                        {
+                            certificate = certificateLoader.Load(certificateConfig).FirstOrDefault();
+                        }
+                        catch (KeyNotFoundException) when (certificateConfig.Value.Equals(DevelopmentSSLCertificateName, StringComparison.Ordinal) && _hostingEnvironment.IsDevelopment())
+                        {
+                            var storeLoader = new CertificateStoreLoader();
```
cesarbs: I know you're just moving the code, but it still bothers me that we're newing this up here. I think loading the default cert should be a method on `CertificateLoader`. Thoughts?

src/Microsoft.AspNetCore.Server.Kestrel/Internal/KestrelServerConfigureOptions.cs
```diff
@@ -0,0 +1,110 @@
...
+                            certificate = storeLoader.Load(DefaultCertificateSubjectName, "My", StoreLocation.CurrentUser, validOnly: false) ??
+                                storeLoader.Load(DefaultCertificateSubjectName, "My", StoreLocation.LocalMachine, validOnly: false);
+
+                            if (certificate == null)
+                            {
+                                var logger = _loggerFactory.CreateLogger("Microsoft.AspNetCore.Server.Kestrel.KestrelServerConfigureOptions");
+                                logger.LogError("No HTTPS certificate was found for development. For information on configuring HTTPS see https://go.microsoft.com/fwlink/?linkid=848054.");
```
cesarbs: resx

src/Microsoft.AspNetCore.Server.Kestrel/Internal/KestrelServerConfigureOptions.cs
```diff
@@ -0,0 +1,110 @@
...
+                            }
+                        }
+
+                        if (certificate == null)
+                        {
+                            throw new InvalidOperationException($"No certificate found for endpoint '{endPoint.Key}'.");
```
cesarbs: resx

src/Microsoft.AspNetCore.Server.Kestrel/Internal/KestrelServerConfigureOptions.cs
```diff
@@ -0,0 +1,110 @@
...
+                        }
+                    }
+                    catch (Exception ex)
+                    {
+                        throw new InvalidOperationException("Unable to configure HTTPS endpoint. For information on configuring HTTPS see https://go.microsoft.com/fwlink/?linkid=848054.", ex);
```
cesarbs: resx

src/Microsoft.AspNetCore.Server.Kestrel/Microsoft.AspNetCore.Server.Kestrel.csproj
```diff
@@ -4,19 +4,22 @@
 
   <PropertyGroup>
     <Description>ASP.NET Core Kestrel cross-platform web server.</Description>
-    <TargetFramework>netstandard2.0</TargetFramework>
+    <TargetFrameworks>netstandard2.0;netcoreapp2.0</TargetFrameworks>
```
cesarbs: ?

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/ConfigurationTests.cs
```diff
@@ -0,0 +1,115 @@
...
+
+                var response = await HttpClientSlim.GetStringAsync($"https://127.0.0.1:{port}", validateCertificate: false);
+                Assert.Equal($"https://127.0.0.1:{port}/", response);
+            }
+        }
+
+        private static int GetWebHostPort(IWebHost webHost)
+            => webHost.ServerFeatures.Get<IServerAddressesFeature>().Addresses
+                .Select(serverAddress => new Uri(serverAddress).Port)
+                .FirstOrDefault();
```
halter73: nit: Single()

src/Microsoft.AspNetCore.Server.Kestrel/Internal/KestrelServerConfigureOptions.cs
Tratcher: This really isn't the time to refactor it.

src/Microsoft.AspNetCore.Server.Kestrel/Microsoft.AspNetCore.Server.Kestrel.csproj
Tratcher: The cert loader sources package requires it

src/Microsoft.AspNetCore.Server.Kestrel/Internal/KestrelServerConfigureOptions.cs
HaoK: Yeah just file an issue for post preview 2 :)

src/Microsoft.AspNetCore.Server.Kestrel/Microsoft.AspNetCore.Server.Kestrel.csproj
davidfowl: Where is the ifdef?

src/Microsoft.AspNetCore.Server.Kestrel/WebHostBuilderKestrelExtensions.cs
HaoK: chatted a bit offline with @davidfowl lets use `services.AddTransient<IConfigureOptions<KestrelServerOptions>, ConfigureDefaults<ConfigureDefaultKestrelServerOptions>()` so when the DI issues is fixed, we only have to remove this line, and won't have to update the ConfigureDefaultOptions class

src/Microsoft.AspNetCore.Server.Kestrel/Microsoft.AspNetCore.Server.Kestrel.csproj
Tratcher: Here: https://github.com/aspnet/Hosting/pull/1095/files#diff-b877901ee14c476e81bad19896b94775R182

davidfowl: I see, what happens on .NET Framework?

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/ConfigurationTests.cs
```diff
@@ -0,0 +1,115 @@
...
+using Microsoft.AspNetCore.Hosting;
+using Microsoft.AspNetCore.Hosting.Server.Features;
+using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Http.Extensions;
+using Microsoft.AspNetCore.Testing;
+using Microsoft.Extensions.Configuration;
+using Microsoft.Extensions.DependencyInjection;
+using Xunit;
+
+namespace Microsoft.AspNetCore.Server.Kestrel.FunctionalTests
```
halter73: Any tests verifying the behavior when something goes wrong like being given an invalid address value or a bad/missing cert? Would the host fail to start even if there are other addresses that are successfully configured?

src/Microsoft.AspNetCore.Server.Kestrel/Microsoft.AspNetCore.Server.Kestrel.csproj
Tratcher: It uses the netstandard version that doesn't have this new feature

davidfowl: Which does what?

src/Microsoft.AspNetCore.Server.Kestrel/Internal/KestrelServerConfigureOptions.cs
```diff
@@ -0,0 +1,110 @@
...
+using Microsoft.AspNetCore.Certificates.Configuration;
+using Microsoft.AspNetCore.Hosting;
+using Microsoft.AspNetCore.Server.Kestrel.Core;
+using Microsoft.Extensions.Configuration;
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Options;
+
+namespace Microsoft.AspNetCore.Server.Kestrel.Internal
+{
+    public class KestrelServerConfigureOptions : IConfigureOptions<KestrelServerOptions>
```
davidfowl: No, lets keep core clean. I don't want configuration leaking into the core APIs. The meta package is where all of the dependencies sit.

src/Microsoft.AspNetCore.Server.Kestrel/Microsoft.AspNetCore.Server.Kestrel.csproj
halter73: Just move this to the Core project. Currently only the "meta" project only contains WebHostBuilderKestrelExtensions.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/KestrelServerConfigureOptions.cs
HaoK: You can't combine them as one is a true default that you always want to apply, while the config binding one is a 'squishy' default, that we only want to apply by default via WebHostBuilder

Tratcher: It requires .Https

halter73: It wouldn't work anyway since this calls UseHttps() 😢

src/Microsoft.AspNetCore.Server.Kestrel/Microsoft.AspNetCore.Server.Kestrel.csproj
halter73: nvm. I just realized this depends on both the Core and Https projects.

src/Microsoft.AspNetCore.Server.Kestrel/WebHostBuilderKestrelExtensions.cs
Tratcher: It's getting fixed: https://github.com/aspnet/DependencyInjection/issues/530
That said, @halter73 says we can't rely on that because it doesn't guarantee the same ordering and KestrelServerConfigureOptions requires KestrelServerOptionsSetup to have run first.

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/ConfigurationTests.cs
Tratcher: It throws. I added a test.

src/Microsoft.AspNetCore.Server.Kestrel/Microsoft.AspNetCore.Server.Kestrel.csproj
```diff
@@ -4,21 +4,30 @@
...
   </ItemGroup>
 
   <ItemGroup>
     <ProjectReference Include="..\Microsoft.AspNetCore.Server.Kestrel.Core\Microsoft.AspNetCore.Server.Kestrel.Core.csproj" />
+    <ProjectReference Include="..\Microsoft.AspNetCore.Server.Kestrel.Https\Microsoft.AspNetCore.Server.Kestrel.Https.csproj" />
     <ProjectReference Include="..\Microsoft.AspNetCore.Server.Kestrel.Transport.Libuv\Microsoft.AspNetCore.Server.Kestrel.Transport.Libuv.csproj" />
     <ProjectReference Include="..\Microsoft.AspNetCore.Server.Kestrel.Transport.Sockets\Microsoft.AspNetCore.Server.Kestrel.Transport.Sockets.csproj" />
   </ItemGroup>
 
+  <ItemGroup>
```
davidfowl: Remove?

Tratcher:
That should be everything, Please review.

Tratcher:
rebased and merged.

PR #1885, opened 2017-06-02
Add request body read timeout

cesarbs:
This is a per read timeout similar to nginx's [`client_body_timeout`](http://nginx.org/en/docs/http/ngx_http_core_module.html#client_body_timeout). Most other servers offer this feature. It's a simpler alternative to #1874.

TODO:

- [ ] Don't enforce on upgrade requests
- [ ] Make it configurable per request
- [ ] Tests

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Kestrel.Core/KestrelServerLimits.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel.Core/KestrelServerLimits.cs
@@ -32,6 +32,8 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core
         private long? _maxConcurrentConnections = null;
         private long? _maxConcurrentUpgradedConnections = null;
 
+        private TimeSpan _requestBodyReadTimeout = TimeSpan.FromSeconds(60);
+
         /// <summary>
         /// Gets or sets the maximum size of the response buffer before write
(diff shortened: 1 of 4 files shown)


Tratcher:
This is the super simple one like we discussed in the meeting today, right?

cesarbs:
@Tratcher Yes.

src/Microsoft.AspNetCore.Server.Kestrel.Core/KestrelServerLimits.cs
```diff
@@ -207,5 +209,14 @@ public TimeSpan RequestHeadersTimeout
                 _maxConcurrentUpgradedConnections = value;
             }
         }
+
+        public TimeSpan RequestBodyReadTimeout
+        {
+            get => _requestBodyReadTimeout;
```
natemcmaster: But if we go with this, it doesn't look like we need a backing field. Can be simplified to 
```c#
public TimeSpan RequestBodyReadTimeout { get; set; } = TimeSpan.FromSeconds(60);
```

cesarbs: No validation yet. If it's added we need a backing field.

halter73:
@cesarbs You're closing this, right?

PR #1886, opened 2017-06-05
Remove ConfigureDefaultOptions

HaoK:


Tratcher:
@DamianEdwards @danroth27 are we sure about this one? The default binders do not work for Kestrel.

danroth27:
Yes, the config support for kestrel is getting pulled for 2.0.

src/Microsoft.AspNetCore.Server.Kestrel/Microsoft.AspNetCore.Server.Kestrel.csproj
```diff
@@ -14,7 +14,6 @@
   <ItemGroup>
     <PackageReference Include="Microsoft.AspNetCore.Certificates.Configuration.Sources" Version="$(AspNetCoreVersion)" PrivateAssets="All" />
```
Tratcher: No longer used?
@javiercn are you going to use Microsoft.AspNetCore.Certificates.Configuration.Sources in Identity? If not, should it be removed from Hosting for now?

Tratcher: Remove Kestrel.Https as well, it was only added for config binding

Tratcher (changes requested):

HaoK:
Reverted the original change instead https://github.com/aspnet/KestrelHttpServer/issues/1875

PR #1889, opened 2017-06-08
Remove usage of TaskCache

pranavkm:


pranavkm:
Do we want to keep `TaskCache<T>` or just make this a field on the type and remove the package reference: https://github.com/aspnet/KestrelHttpServer/blob/dev/src/Microsoft.AspNetCore.Server.Kestrel.Https/Internal/ClosedStream.cs#L59?

natemcmaster:
Let's remove it completely. +1 for making it a static field in ClosedStream.

By the way, TaskCache has this comment:

>  We're caching this in a static readonly field to make it more inlinable and avoid the volatile lookup done by Task.CompletedTask.

Can you run some benchmarks to see if there is any perf impact of this change?

pranavkm:
@natemcmaster that was fixed in https://github.com/dotnet/coreclr/pull/8846/files. @rynowak to confirm

src/Microsoft.AspNetCore.Server.Kestrel.Https/Internal/ClosedStream.cs
```diff
@@ -56,7 +57,7 @@ public override int Read(byte[] buffer, int offset, int count)
 
         public override Task<int> ReadAsync(byte[] buffer, int offset, int count, CancellationToken cancellationToken)
         {
-            return TaskCache<int>.DefaultCompletedTask;
+            return CompletedIntTask;
```
pranavkm: ❓

src/Microsoft.AspNetCore.Server.Kestrel.Https/Internal/ClosedStream.cs
```diff
@@ -5,12 +5,13 @@
 using System.IO;
 using System.Threading;
 using System.Threading.Tasks;
-using Microsoft.Extensions.Internal;
 
 namespace Microsoft.AspNetCore.Server.Kestrel.Https.Internal
 {
     internal class ClosedStream : Stream
     {
+        private static readonly Task<int> CompletedIntTask = Task.FromResult(default(int));
```
benaadams: That's pretty weird isn't it? Its explicitly a zero return - rather than a `default(int)`; also `CompletedIntTask` doesn't really specify what it is.

cesarbs: Let's call it `ZeroResultTask`.

pranavkm:
🆙 📅

src/Microsoft.AspNetCore.Server.Kestrel.Https/Internal/ClosedStream.cs
```diff
@@ -5,12 +5,13 @@
 using System.IO;
 using System.Threading;
 using System.Threading.Tasks;
-using Microsoft.Extensions.Internal;
 
 namespace Microsoft.AspNetCore.Server.Kestrel.Https.Internal
 {
     internal class ClosedStream : Stream
     {
+        private static readonly Task<int> ZeroResultTask = Task.FromResult(default(int));
```
cesarbs: `FromResult(0)`

pranavkm:
🆙 📅

PR #1891, opened 2017-06-08
Update defaults in KestrelServerLimits doc comments

halter73:


src/Microsoft.AspNetCore.Server.Kestrel.Core/KestrelServerLimits.cs
```diff
@@ -147,7 +147,7 @@ public int MaxRequestHeaderCount
         /// This can be overridden per-request via <see cref="IHttpMaxRequestBodySizeFeature"/>.
         /// </summary>
         /// <remarks>
-        /// Defaults to null (unlimited).
+        /// Defaults to 30000000, which is approximately 28.6MB.
```
cesarbs: 30,000,000 bytes

PR #1892, opened 2017-06-08
Implement a test for illegal trailing whitespace

rjmooney:
TFS-221201: Illegal trailing whitespace should cause subsequent headers with leading whitespace to be ignored

@natemcmaster @halter73 @muratg


cesarbs:
We already cover that in `BadHttpRequestTests.TestInvalidHeaders`.

rjmooney:
Whoops, you're right.

PR #1893, opened 2017-06-09
Always complete RequestBodyPipe.Reader

halter73:
This is part of an effort to stop leaking memory pool blocks.

Semi related: https://github.com/dotnet/corefxlab/pull/1601


cesarbs:
CI is upset. But the change itself looks good.

PR #1894, opened 2017-06-09
Complete IPipeWriter in OutputProducer.Abort()

halter73:
- This prevents MemoryPoolBlocks from leaking when writes fail

The leak can be observed by running LibuvOutputConsumerTests.FailedWriteCompletesOrCancelsAllPendingTasks with BLOCK_LEASE_TRACKING defined in Transport.Abstractions. Once https://github.com/dotnet/corefxlab/pull/1601 is merged, defining BLOCK_LEASE_TRACKING won't be necessary to see the leak occurred.


davidfowl:
My bad 😄

Issue #1896, opened 2017-06-14
Both reader and writer need to be completed to be able to reset 


davidfowl:
Seems like there's an issue resetting the pipe in for the request body. This happend in the SignalR sample when I closed the browser.

```
SocketsSample> info: Microsoft.AspNetCore.Hosting.Internal.WebHost[2]
SocketsSample>       Request finished in 3017.3625ms 101 
SocketsSample> warn: Microsoft.AspNetCore.Server.Kestrel[0]
SocketsSample>       Connection processing ended abnormally.
SocketsSample> System.InvalidOperationException: Both reader and writer need to be completed to be able to reset 
SocketsSample>    at Microsoft.AspNetCore.Server.Kestrel.Internal.System.IO.Pipelines.Pipe.Reset()
SocketsSample>    at Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http.Frame`1.<ProcessRequestsAsync>d__2.MoveNext()
```

/cc @halter73

halter73:
Do you see this consistently? I think @anurse saw this before in the Autobahn test suite, but it was sporadic.

I tested the Sockets, Hubs and Streaming pages of SocketSample with WebSockets and then closed the browser (Edge) completely. In each case, I saw a "Socket closed" log from WebSocketsTransport prior to "Request finished in ...". There were no "Connection processing ended abnormally" logs.

At first I thought this might be caused by a rogue ongoing ReadAsync call when the middleware completed, but if that were the case you would get a different InvalidOperationException ("Concurrent reads or writes are not supported."). If this happened in dev after https://github.com/aspnet/KestrelHttpServer/commit/d879518a18c765a0571b4583ba85be705ca889fe went in, that would mean that the Writer wasn't completed since the Reader definitely was.

There are a couple approaches we could take to fix it. The most obvious to me is to await MessageBody's PumpAsync task again. Alternatively, not calling RequestBodyPipe.Reset for non-keepalive requests like before is probably also sufficient.

Some more repro details could verify the fix though.

anurse:
Looks like the same (though the exception message is a little different, I think mine was just about readers).

With WebSockets, it is definitely possible (though incorrect) for a user to release control in the App Func while still holding on to the socket (by pushing it over to a background thread). This should probably be guarded a little better to ensure that the error messaging is clearer.

PR #1902, opened 2017-06-15
Wait until the writer is complete to reset RequestBodyPipe

halter73:
#1896


src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/Http/FrameOfT.cs
```diff
@@ -202,6 +194,11 @@ public Frame(IHttpApplication<TContext> application, FrameContext frameContext)
                             {
                                 RequestBodyPipe.Reader.Complete();
 
+                                // Wait for MessageBody.PumpAsync() to call RequestBodyPipe.Writer.Complete().
+                                messageBody.Cancel();
+                                Input.CancelPendingRead();
```
davidfowl: Should messageBody.Cancel call Input.CancelPendingRead as well?

Issue #1904, opened 2017-06-17
Min rate applies to upgraded requests as well?


davidfowl:
I can't remember if this was intentional but now it seems like websockets is broken. The application sent a single packet and then stopped (which is typical in these kinds of applications) and kestrel kills the connection because it dropped below the min rate.

```
info: Microsoft.AspNetCore.Server.Kestrel[27]
      Connection id "0HL5LFRPK4FA2", Request id "0HL5LFRPK4FA2:00000001": request body incoming data rate dropped below 1 bytes/second.
fail: Microsoft.AspNetCore.Sockets.ConnectionManager[0]
      Failed disposing connection 89673b87-85ff-447f-9ccd-e487c8377922
System.Net.WebSockets.WebSocketException (0x80004005): The remote party closed the WebSocket connection without completing the close handshake. ---> Microsoft.AspNetCore.Server.Kestrel.Core.BadHttpRequestException: Request timed out.
   at Microsoft.AspNetCore.Server.Kestrel.Internal.System.IO.Pipelines.PipeCompletion.ThrowFailed()
   at Microsoft.AspNetCore.Server.Kestrel.Internal.System.IO.Pipelines.Pipe.GetResult(ReadResult& result)
   at Microsoft.AspNetCore.Server.Kestrel.Internal.System.IO.Pipelines.Pipe.Microsoft.AspNetCore.Server.Kestrel.Internal.System.IO.Pipelines.IReadableBufferAwaiter.GetResult()
   at Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http.MessageBody.<ReadAsync>d__22.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http.FrameRequestStream.<ReadAsyncInternal>d__20.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at System.Net.WebSockets.ManagedWebSocket.<EnsureBufferContainsAsync>d__70.MoveNext()
[...]

PR #1907, opened 2017-06-19
Don't pause and resume read timing on upgrade requests (#1904)

cesarbs:
#1904

@muratg @Eilon @DamianEdwards This should be in preview2. Enforcing the rate on WebSocket requests breaks things.

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/Http/MessageBody.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/Http/MessageBody.cs
@@ -8,6 +8,7 @@ using System.Threading.Tasks;
 using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Infrastructure;
 using Microsoft.AspNetCore.Server.Kestrel.Internal.System.IO.Pipelines;
+using Microsoft.Extensions.Logging;
 
 namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
 {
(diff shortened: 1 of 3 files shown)


src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/Http/MessageBody.cs
```diff
@@ -273,6 +274,22 @@ private void TryStartTimingReads()
             }
         }
 
+        private void TryPauseTimingReads()
+        {
+            if (!RequestUpgrade)
```
natemcmaster: Should this only apply after the request has actually been upgraded? This value is true even if the app does not accept the upgrade request.

cesarbs: Reads return 0 bytes on upgrade requests if you try to read from them before upgrading. So the app wouldn't hang on a read. Is that your concern?

halter73:
Why not add a functional test?

src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/Http/MessageBody.cs
natemcmaster: Was just asking. At the moment, we don't support upgradable requests with request bodies so this should be fine.

Btw, it would be cleaner to put this in the `ForUpgrade` class.

cesarbs:
@halter73 Added functional test.

muratg:
This bug was NOT approved for preview2 yet.  @DamianEdwards, how important this is to get in preview2?

DamianEdwards:
It breaks WebSockets, I don't think we can ship preview2 with regressed WebSockets support.

Issue #1876, opened 2017-06-01
Make Microsoft.AspNetCore.Server.Kestrel.Transport.Abstractions pubternal


davidfowl:
We need to do this before we ship 2.0

/cc @halter73

halter73:
@muratg I added this to the 2.0 milestone.

PR #1911, opened 2017-06-21
Make Transport.Abstractions pubternal

halter73:
#1876

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Kestrel/Microsoft.AspNetCore.Server.Kestrel.csproj
+++ b/src/Microsoft.AspNetCore.Server.Kestrel/Microsoft.AspNetCore.Server.Kestrel.csproj
@@ -18,7 +18,6 @@
   <ItemGroup>
     <ProjectReference Include="..\Microsoft.AspNetCore.Server.Kestrel.Core\Microsoft.AspNetCore.Server.Kestrel.Core.csproj" />
     <ProjectReference Include="..\Microsoft.AspNetCore.Server.Kestrel.Transport.Libuv\Microsoft.AspNetCore.Server.Kestrel.Transport.Libuv.csproj" />
-    <ProjectReference Include="..\Microsoft.AspNetCore.Server.Kestrel.Transport.Sockets\Microsoft.AspNetCore.Server.Kestrel.Transport.Sockets.csproj" />
   </ItemGroup>
 
 </Project>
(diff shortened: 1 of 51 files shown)


src/Microsoft.AspNetCore.Server.Kestrel/Microsoft.AspNetCore.Server.Kestrel.csproj
```diff
@@ -18,7 +18,6 @@
   <ItemGroup>
     <ProjectReference Include="..\Microsoft.AspNetCore.Server.Kestrel.Core\Microsoft.AspNetCore.Server.Kestrel.Core.csproj" />
     <ProjectReference Include="..\Microsoft.AspNetCore.Server.Kestrel.Transport.Libuv\Microsoft.AspNetCore.Server.Kestrel.Transport.Libuv.csproj" />
-    <ProjectReference Include="..\Microsoft.AspNetCore.Server.Kestrel.Transport.Sockets\Microsoft.AspNetCore.Server.Kestrel.Transport.Sockets.csproj" />
```
natemcmaster: Seems unrelated to the pubternal change. What's the reason for this removal?

halter73:
We are pubternalizing this API because we don't want to commit to no breaking changes post 2.0. All of the pipe-related APIs are being shipped as pubternal APIs already, so there wasn't going to be any guarantee of API stability to begin with. This just makes that more clear.

src/Microsoft.AspNetCore.Server.Kestrel/Microsoft.AspNetCore.Server.Kestrel.csproj
halter73: It's pretty unrelated, but it's required for https://github.com/aspnet/Coherence/pull/231. We're not shipping the package because it isn't tested well enough yet, so we can't depend on it from a shipping package. I can move this into another PR if need be.

natemcmaster: It's fine here. Thanks for clarifying.

natemcmaster:
Looks like we'll need to update the Travis systemd tests. In the meantime, we can use the deprecated image by adding this to the .travis.yml file

```yml
group: deprecated-2017Q2
```

https://blog.travis-ci.com/2017-06-21-trusty-updates-2017-Q2-launch

Issue #1901, opened 2017-06-15
Rename min rate APIs for consistency


cesarbs:
We should do this before RTM:

- Rename `KestrelServerLimits.RequestBodyMinimumDataRate` to `MinRequestBodyDataRate` (consistent with MaxRequestBodySize, and other limits that all start with “Max”)
- Rename `IHttpRequestBodyMinimumDataRateFeature` to `IHttpMinRequestBodyDataRateFeature` (consistent with `IHttpMaxRequestBodySizeFeature`)
- Rename the `rate` parameter of the `MinimumDataRate` ctor to `bytesPerSecond`, so that the rate’s unit of measurement is clear

cesarbs:
@muratg

muratg:
@Eilon @DamianEdwards @shirhatti FYI -- proposed breaking change between preview2 and RTM to make the names more consistent.

Eilon:
Can someone show the full list of before/after values? Hard to tell what's consistent ec.

muratg:
@cesarbs could you make a table?

cesarbs:
### KestrelServerLimits

Before | After
------- | ------
MaxResponseBufferSize | MaxResponseBufferSize
MaxRequestBufferSize | MaxRequestBufferSize
MaxRequestLineSize | MaxRequestLineSize
MaxRequestHeadersTotalSize | MaxRequestHeadersTotalSize
MaxRequestHeaderCount | MaxRequestHeaderCount
MaxRequestBodySize | MaxRequestBodySize
**RequestBodyMinimumDataRate** | **MinRequestBodyDataRate**
MaxConcurrentConnections | MaxConcurrentConnections
MaxConcurrentUpgradedConnections | MaxConcurrentUpgradedConnections
KeepAliveTimeout | KeepAliveTimeout
RequestHeadersTimeout | RequestHeadersTimeout

### Features

Before | After
------- | ------
IHttpMaxRequestBodySizeFeature | IHttpMaxRequestBodySizeFeature
**IHttpRequestBodyMinimumDataRateFeature** | **IHttpMinRequestBodyDataRateFeature**

### MinimumDataRate

#### Before

```csharp
public class MinimumDataRate
{
    public MinimumDataRate(double rate, TimeSpan gracePeriod)
    {
        ...
    }

    public double Rate { get; }

    public TimeSpan GracePeriod { get; }
}
```

#### After

```csharp
public class MinimumDataRate
{
[...]

PR #1915, opened 2017-06-23
Rename request body min rate APIs (#1901)

cesarbs:
#1901


src/Microsoft.AspNetCore.Server.Kestrel.Core/KestrelServerLimits.cs
```diff
@@ -253,13 +253,52 @@ public TimeSpan RequestHeadersTimeout
...
+        /// This limit can be overridden per-request via <see cref="IHttpMinRequestBodyDataRateFeature"/>.
         /// </summary>
         /// <remarks>
-        /// Defaults to 1 byte/second with a 5 second grace period.
+        /// Defaults to 1 byte/second with a 5 second grace period during which the minimum data rate is not enforced.
         /// </remarks>
-        public MinimumDataRate RequestBodyMinimumDataRate { get; set; } = new MinimumDataRate(rate: 1, gracePeriod: TimeSpan.FromSeconds(5));
+        public MinimumDataRate MinRequestBodyDataRate { get; } = new MinimumDataRate { BytesPerSecond = 1, GracePeriod = TimeSpan.FromSeconds(5) };
+
+        public class MinimumDataRate
```
halter73: I don't like having a public nested class.

src/Microsoft.AspNetCore.Server.Kestrel.Core/KestrelServerLimits.cs
```diff
@@ -253,13 +253,52 @@ public TimeSpan RequestHeadersTimeout
...
+
+            /// <summary>
+            /// The amount of time to delay enforcement of <see cref="MinimumDataRate" />.
+            /// </summary>
+            public TimeSpan GracePeriod
+            {
+                get => _gracePeriod;
+                set
+                {
+                    if (value < TimeSpan.Zero)
```
halter73: I wonder if we should throw for TimeSpan.Zero too, since that's just asking for problems.

src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/Http/Frame.FeatureCollection.cs
```diff
@@ -229,10 +229,32 @@ string IHttpRequestIdentifierFeature.TraceIdentifier
...
+                MinRequestBodyDataRate.BytesPerSecond = value;
+            }
+        }
+
+        TimeSpan IHttpMinRequestBodyDataRateFeature.GracePeriod
+        {
+            get => MinRequestBodyDataRate.GracePeriod;
+            set
+            {
+                if (value < TimeSpan.Zero)
```
halter73: Is there a way to share the validation logic between this and the MinimumDataRate class?

src/Microsoft.AspNetCore.Server.Kestrel.Core/Features/IHttpMinRequestBodyDataRateFeature.cs
```diff
@@ -1,18 +1,25 @@
...
         /// The minimum data rate in bytes/second at which the request body should be received.
-        /// Setting this property to null indicates no minimum data rate should be enforced.
+        /// Setting this property to zero effectively means no minimum data rate should be enforced.
         /// This limit has no effect on upgraded connections which are always unlimited.
         /// </summary>
-        MinimumDataRate MinimumDataRate { get; set; }
+        double BytesPerSecond { get; set; }
+
+        /// <summary>
+        /// The amount of time to delay enforcement of the minimum data rate.
```
Tratcher: Starting from when? After the request arrives? The first read?

src/Microsoft.AspNetCore.Server.Kestrel.Core/KestrelServerLimits.cs
Tratcher: The feature uses two fields but the options use a custom type? Why is this type needed for options?

cesarbs: @Tratcher `MinRequestBodyDataRate.BytesPerSecond` matches how the feature is used better than `MinRequestBodyDataRateBytesPerSecond`. It's just aesthetics though.

src/Microsoft.AspNetCore.Server.Kestrel.Core/KestrelServerLimits.cs
cesarbs: It's still a legal value. Means the rate will be enforced right away. It's not recommended, but there's no technical reason to disallow it.

halter73: The technical reason to disallow it is because setting it to zero creates a race that could cause any connection to be killed no matter the actual upload rate

cesarbs: Then we should require it to be larger than 1 second.

cesarbs: Or, more precisely, larger than `Heartbeat.Interval`, which right now is one second.

src/Microsoft.AspNetCore.Server.Kestrel.Core/Features/IHttpMinRequestBodyDataRateFeature.cs
cesarbs: Will add ", starting at the time data is first read or written." Written as well since this same type will (likely) be used for the response rate.

cesarbs:
- Reverted 5c63a585a66368c2c232602c6147d8b26e347786 to stop the bikeshedding
- Moved `MinimumDataRate` to the base Core namespace so people don't need a weird `using` when setting it in `KestrelServerLimits`
- Allow 0 bytes/second
- Require the grace period to be larger than `Heartbeat.Interval`

cesarbs:
📢

src/Microsoft.AspNetCore.Server.Kestrel.Core/Features/IHttpMinRequestBodyDataRateFeature.cs
```diff
@@ -6,7 +6,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Features
     /// <summary>
     /// Represents a minimum data rate for the request body of an HTTP request.
     /// </summary>
-    public interface IHttpRequestBodyMinimumDataRateFeature
+    public interface IHttpMinRequestBodyDataRateFeature
```
halter73: I'm not sure we should be abbreviating Minimum to Min

halter73: After thinking about this for a little bit, I think we should change all usages of "Minimum" in public APIs to "Min" for consistency considering how many limits already use "Max".

So that means the MinimumDataRate class changes to MinDataRate, IHttpMinRequestBodyDataRateFeature.MinimumDataRate changes to IHttpMinRequestBodyDataRateFeature.MinDataRate, etc...

I don't care as much about private/internal names, but I think some of that could also change to match.

cesarbs: I already change the property/param to bytesPerSecond, so that's taken care of. I'll change the type name now.

I went with Min because we already have Max on other properties.

halter73:
I meant to leave a comment with my approval. I approve this pending changing "Minimum" to "Min" in all public APIs.

cesarbs:
Updated, will merge when CI is green.

src/Microsoft.AspNetCore.Server.Kestrel.Core/KestrelServerLimits.cs
```diff
@@ -255,11 +255,11 @@ public TimeSpan RequestHeadersTimeout
...
         /// Setting this property to null indicates no minimum data rate should be enforced.
         /// This limit has no effect on upgraded connections which are always unlimited.
-        /// This can be overridden per-request via <see cref="IHttpRequestBodyMinimumDataRateFeature"/>.
+        /// This can be overridden per-request via <see cref="IHttpMinRequestBodyDataRateFeature"/>.
         /// </summary>
         /// <remarks>
         /// Defaults to 1 byte/second with a 5 second grace period.
         /// </remarks>
-        public MinimumDataRate RequestBodyMinimumDataRate { get; set; } = new MinimumDataRate(rate: 1, gracePeriod: TimeSpan.FromSeconds(5));
+        public MinDataRate MinRequestBodyDataRate { get; set; } = new MinDataRate(bytesPerSecond: 1, gracePeriod: TimeSpan.FromSeconds(5));
```
halter73: Change this to match the IIS default. I believe it's [240 bytes/second](https://www.iis.net/configreference/system.applicationhost/weblimits#005).

halter73:
I know I already approved this. Hopeful changing the default is the last thing.

src/Microsoft.AspNetCore.Server.Kestrel.Core/KestrelServerLimits.cs
cesarbs: Separate change. Let's discuss it in the office tomorrow.

halter73: And here I was trying to avoid creating a new issue 😢

halter73: https://github.com/aspnet/KestrelHttpServer/issues/1929

Issue #1521, opened 2017-03-20
Add flag to disable synchronous IO 


davidfowl:
Because blocking threads is bad so we want a way to turn this off synchronous IO by default to let you shoot yourself in the foot if you want this behavior.

benaadams:
`throw new NotWiseException()`?

Drawaes:
Throw new Bang (Target.Foot)

DamianEdwards:
@davidfowl you planning to get this added this week?

davidfowl:
@DamianEdwards I'm not sure we can do this for preview. There's too much churn and too many other important things to do in Kestrel right now (like stabilize the build).

muratg:
@davidfowl @DamianEdwards We'd like to punt this out to 2.1.  That OK with you guys?

muratg:
Talked to @DamianEdwards in person, we'll try to do this in 2.0.0.

PR #1919, opened 2017-06-27
Add feature to optionally disallow synchronous IO

halter73:
#1521

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/Http/Frame.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/Http/Frame.cs
@@ -306,7 +306,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
         {
             if (_frameStreams == null)
             {
-                _frameStreams = new Streams(this, this);
+                _frameStreams = new Streams(bodyControl: this, frameControl: this);
             }
 
             (RequestBody, ResponseBody) = _frameStreams.Start(messageBody);
--- a/src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/Http/Frame.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/Http/Frame.cs
@@ -122,6 +122,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
         public string ConnectionIdFeature { get; set; }
         public bool HasStartedConsumingRequestBody { get; set; }
         public long? MaxRequestBodySize { get; set; }
+        public bool AllowSynchronousIO { get; set; }
 
         /// <summary>
         /// The request id. <seealso cref="HttpContext.TraceIdentifier"/>
@@ -305,7 +306,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
         {
             if (_frameStreams == null)
             {
-                _frameStreams = new Streams(this);
+                _frameStreams = new Streams(this, this);
             }
 
             (RequestBody, ResponseBody) = _frameStreams.Start(messageBody);
@@ -329,6 +330,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
 
             HasStartedConsumingRequestBody = false;
             MaxRequestBodySize = ServerOptions.Limits.MaxRequestBodySize;
+            AllowSynchronousIO = ServerOptions.AllowSynchronousIO;
             TraceIdentifier = null;
             Scheme = null;
             Method = null;
--- a/src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/Http/FrameRequestStream.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/Http/FrameRequestStream.cs
@@ -59,7 +59,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
         {
             if (!_bodyControl.AllowSynchronousIO)
             {
-                throw new InvalidOperationException("Synchronous operations are disallowed. Call ReadAsync or set AllowSynchronousIO to true instead.");
+                throw new InvalidOperationException(CoreStrings.SynchronousReadsDisallowed);
             }
 
             return ReadAsync(buffer, offset, count).GetAwaiter().GetResult();
--- a/src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/Http/FrameRequestStream.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/Http/FrameRequestStream.cs
@@ -7,17 +7,20 @@ using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Infrastructure;
 using Microsoft.AspNetCore.Server.Kestrel.Transport.Abstractions.Internal;
+using Microsoft.AspNetCore.Http.Features;
 
 namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
 {
     internal class FrameRequestStream : ReadOnlyStream
     {
+        private readonly IHttpBodyControlFeature _bodyControl;
         private MessageBody _body;
         private FrameStreamState _state;
         private Exception _error;
 
-        public FrameRequestStream()
+        public FrameRequestStream(IHttpBodyControlFeature bodyControl)
         {
+            _bodyControl = bodyControl;
             _state = FrameStreamState.Closed;
         }
 
@@ -34,13 +37,12 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
 
         public override void Flush()
         {
-            // No-op.
+            throw new NotSupportedException();
         }
 
         public override Task FlushAsync(CancellationToken cancellationToken)
         {
-            // No-op.
-            return Task.CompletedTask;
+            throw new NotSupportedException();
         }
 
         public override long Seek(long offset, SeekOrigin origin)
@@ -55,8 +57,12 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
 
         public override int Read(byte[] buffer, int offset, int count)
         {
-            // ValueTask uses .GetAwaiter().GetResult() if necessary
-            return ReadAsync(buffer, offset, count).Result;
+            if (!_bodyControl.AllowSynchronousIO)
+            {
+                throw new InvalidOperationException("Synchronous operations are disallowed. Call ReadAsync or set AllowSynchronousIO to true instead.");
+            }
+
+            return ReadAsync(buffer, offset, count).GetAwaiter().GetResult();
         }
 
         public override IAsyncResult BeginRead(byte[] buffer, int offset, int count, AsyncCallback callback, object state)
--- a/src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/Http/FrameResponseStream.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/Http/FrameResponseStream.cs
@@ -34,7 +34,6 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
             set => throw new NotSupportedException();
         }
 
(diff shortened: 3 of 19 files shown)


src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/Http/FrameResponseStream.cs
```diff
@@ -31,6 +34,7 @@ public override long Position
             set => throw new NotSupportedException();
         }
 
+        // REVIEW: Should Flush() throw?
```
Tratcher: Yes

src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/Http/Frame.cs
```diff
@@ -330,6 +331,7 @@ public void Reset()
 
             HasStartedConsumingRequestBody = false;
             MaxRequestBodySize = ServerOptions.Limits.MaxRequestBodySize;
+            AllowSynchronousIO = false;
```
Tratcher: Initialize from KestrelServerOptions?

src/Microsoft.AspNetCore.Server.Kestrel.Core/KestrelServerOptions.cs
```diff
@@ -36,6 +36,11 @@ public class KestrelServerOptions
         public SchedulingMode ApplicationSchedulingMode { get; set; } = SchedulingMode.Default;
 
         /// <summary>
+        /// Gets or sets a value that controls whether synchronous IO is allowed for the <see cref="Http.Features.IHttpRequestFeature.Body"/> and <see cref="Http.Features.IHttpResponseFeature.Body"/>
```
Tratcher: HttpContext.Request/Reponse.Body

src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/Http/FrameResponseStream.cs
davidfowl: Hmmm, we discussed this yesterday and there are some reasons it shouldn't. `StreamWriter` *always* calls Flush on Dispose https://github.com/dotnet/corefx/blob/6dd451f51451a7d0ceea6104b51bd17005e9a0e6/src/System.Runtime.Extensions/src/System/IO/StreamWriter.cs#L188.

Tratcher: Which will call Write and throw: https://github.com/dotnet/corefx/blob/6dd451f51451a7d0ceea6104b51bd17005e9a0e6/src/System.Runtime.Extensions/src/System/IO/StreamWriter.cs#L253

Tratcher: Kestrel's flush needs to throw for all the same reasons write is throwing, sync over async. HttpSys's flush no-ops unless you haven't sent the headers, so throwing could be conditional there but that would be even more confusing.

davidfowl: It means people can't use `StreamWriter` over the body even if they wrote *everything* async.

Tratcher: Which is our recommendation for response streams anyways. E.g. we already make Dispose no-op, it's the servers job to clean up the stream.

That said, StreamWriter calls Flush even for LeaveOpen:
https://github.com/dotnet/corefx/blob/6dd451f51451a7d0ceea6104b51bd17005e9a0e6/src/System.Runtime.Extensions/src/System/IO/StreamWriter.cs#L184

davidfowl: Well, we need to tell people to *not* dispose the StreamWriter at all.

halter73: StreamWriter.Dispose calls Stream.Flush no matter what. You can prevent StreamWriter.Dispose from calling Stream.Write by either calling StreamWriter.FlushAsync immediately before or using "leaveOpen".

src/Microsoft.AspNetCore.Server.Kestrel.Core/KestrelServerOptions.cs
```diff
@@ -36,6 +37,11 @@ public class KestrelServerOptions
         public SchedulingMode ApplicationSchedulingMode { get; set; } = SchedulingMode.Default;
 
         /// <summary>
+        /// Gets or sets a value that controls whether synchronous IO is allowed for the <see cref="HttpContext.Request"/> and <see cref="HttpContext.Response"/>
```
halter73: I changed the doc comments to reference HttpContext instead of IHttpRequest/ResponseFeature. The problem is if I reference HttpContext.Request/Response.**Body**, I get the following error:

```
KestrelServerOptions.cs(40,101): error CS1574: XML comment has cref attribute 'Body' that could not be resolved [D:\dev\aspnet\KestrelHttpServer\src\Microsoft.AspNetCore.Server.Kestrel.Core\Microsoft.AspNetCore.Server.Kestrel.Core.csproj]
KestrelServerOptions.cs(40,144): error CS1574: XML comment has cref attribute 'Body' that could not be resolved [D:\dev\aspnet\KestrelHttpServer\src\Microsoft.AspNetCore.Server.Kestrel.Core\Microsoft.AspNetCore.Server.Kestrel.Core.csproj]
```

Any ideas why that is?

Tratcher: I assume it can't navigate multiple properties deep.

halter73: That makes sense. At the same time, we want to reference HttpContext in the doc comment, right? I'll just leave it as is then.

halter73:
⬆️ 📅

src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/Http/Frame.cs
```diff
@@ -305,7 +306,7 @@ public void InitializeStreams(MessageBody messageBody)
         {
             if (_frameStreams == null)
             {
-                _frameStreams = new Streams(this);
+                _frameStreams = new Streams(this, this);
```
cesarbs: Use keyword args.

src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/Http/FrameRequestStream.cs
```diff
@@ -55,8 +57,12 @@ public override void SetLength(long value)
 
         public override int Read(byte[] buffer, int offset, int count)
         {
-            // ValueTask uses .GetAwaiter().GetResult() if necessary
-            return ReadAsync(buffer, offset, count).Result;
+            if (!_bodyControl.AllowSynchronousIO)
+            {
+                throw new InvalidOperationException("Synchronous operations are disallowed. Call ReadAsync or set AllowSynchronousIO to true instead.");
```
cesarbs: Move string to CoreStrings.

src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/Http/FrameResponseStream.cs
```diff
@@ -58,6 +61,11 @@ public override void SetLength(long value)
 
         public override void Write(byte[] buffer, int offset, int count)
         {
+            if (!_bodyControl.AllowSynchronousIO)
+            {
+                throw new InvalidOperationException("Synchronous operations are disallowed. Call WriteAsync or set AllowSynchronousIO to true instead.");
```
cesarbs: CoreStrings

test/Microsoft.AspNetCore.Server.Kestrel.Core.Tests/MessageBodyTests.cs
```diff
@@ -28,18 +29,18 @@ public class MessageBodyTests
...
-                var stream = new FrameRequestStream();
+                var stream = new FrameRequestStream(Mock.Of<IHttpBodyControlFeature>());
                 stream.StartAcceptingReads(body);
 
                 input.Add("Hello");
 
                 var buffer = new byte[1024];
 
-                var count = stream.Read(buffer, 0, buffer.Length);
+                var count = await stream.ReadAsync(buffer, 0, buffer.Length);
```
cesarbs: This test (and similar-named ones) are specifically testing sync reads. Pass a mock feature that allows sync I/O.

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/ResponseTests.cs
```diff
@@ -510,12 +510,11 @@ public Task ResponseStatusCodeSetBeforeHttpContextDisposedRequestMalformedReadIg
             var testLogger = new TestApplicationErrorLogger();
             var serviceContext = new TestServiceContext { Log = new TestKestrelTrace(testLogger) };
 
-            using (var server = new TestServer(httpContext =>
+            using (var server = new TestServer(async httpContext =>
```
cesarbs: This test is specifically checking behavior on sync I/O (note there are tests with Write in the name, and other tests with WriteAsync). Enable sync I/O here.

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/ResponseTests.cs
```diff
@@ -967,12 +965,11 @@ public Task ResponseStatusCodeSetBeforeHttpContextDisposedRequestMalformedReadIg
         {
             var flushed = new SemaphoreSlim(0, 1);
 
-            using (var server = new TestServer(httpContext =>
+            using (var server = new TestServer(async httpContext =>
```
cesarbs: Enable sync I/O here - test is specific to sync I/O.

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/ResponseTests.cs
```diff
@@ -1248,7 +1245,7 @@ public Task ResponseStatusCodeSetBeforeHttpContextDisposedRequestMalformedReadIg
         [Fact]
         public async Task FirstWriteVerifiedAfterOnStarting()
         {
-            using (var server = new TestServer(httpContext =>
+            using (var server = new TestServer(async httpContext =>
```
cesarbs: This test wants to use sync I/O.

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/ResponseTests.cs
```diff
@@ -1289,7 +1285,7 @@ public Task ResponseStatusCodeSetBeforeHttpContextDisposedRequestMalformedReadIg
         [Fact]
         public async Task SubsequentWriteVerifiedAfterOnStarting()
         {
-            using (var server = new TestServer(httpContext =>
+            using (var server = new TestServer(async httpContext =>
```
cesarbs: This test wants to use sync I/O.

halter73:
@cesarbs 🆙 📆ed to address your feedback

test/Microsoft.AspNetCore.Server.Kestrel.Core.Tests/MessageBodyTests.cs
```diff
@@ -249,13 +256,13 @@ public class MessageBodyTests
...
                 var body = MessageBody.For(httpVersion, new FrameRequestHeaders(), input.Frame);
-                var stream = new FrameRequestStream();
+                var stream = new FrameRequestStream(Mock.Of<IHttpBodyControlFeature>());
                 stream.StartAcceptingReads(body);
 
                 input.Add("Hello");
 
                 var buffer = new byte[1024];
-                Assert.Equal(0, stream.Read(buffer, 0, buffer.Length));
+                Assert.Equal(0, await stream.ReadAsync(buffer, 0, buffer.Length));
```
cesarbs: This one should be sync.

halter73: Fixed

PR #1925, opened 2017-06-29
Attempting to fix systemd activation tests with minimal changes

onyxmaster:
This is an attempt to improve/fix on #1912 along with preparation for #1922 tests.
Fixes inspired mostly by https://github.com/solita/docker-systemd

PR #1922, opened 2017-06-28
Domain socket handles

onyxmaster:
This is the PR to solve #1906 
It deviates a little from the originally described implementation plan, as I believe that adding new ListenTypes isn't actually a great idea since the proposed changes represented a different "dimension" of file handle, so the FileHandleType (auto/tcp/pipe) was introduced. The worst (in my opinion) part of the PR is reliance on IEndpointInformation.HandleType being mutable (but since IPEndpoint was mutable and since I've added some sanity checks in place) I believe it's okay. The other, not-so-great part is requirement that both listen and accept listeners do share the same IEndpointInformation reference, but I don't believe there are reasons this should change (and if it would, the only case that will suffer is systemd one).
I tried to keep the changes to a minimum, but had to split the ListenTcp/ListenPipe/ListenHandle and its accept counterparts' due to switch code becoming cluttered after try/catch was applied.

The LibuvConstants.ENOTSUP has no value on Windows intentionally.

Tested with stock Ubuntu 16.04.2, simple cases work, more complex testing will be done tomorrow.

cc: @halter73 

Thank you for considering this.

PR #1930, opened 2017-06-30
Improved systemd activation tests

onyxmaster:
While working on activation tests for #1922 I found a [better way to test systemd activation](https://www.freedesktop.org/software/systemd/man/systemd-socket-activate.html) that allows not running systemd itself in a container (which, at the current point, is a more or less fragile hack). Also it allows for non-privileged container which does not need access to LXC.
This PR radically simplifies the docker part of the equation, along with introducing BASE_PORT envvar to allow for specifying base ports for non-systemd listens, also testing both the systemd socket activation and explicity configured ports (this seems useful, because there is a possibility that one thing can break another).

I understand that there already was a fix at #1925, but it just got it working, this one is much less likely to break in the future, if docker or systemd get upgraded in the CI environment.

I already have #1922 test ready, but it needs this one to work better.

Thank you for reviewing this.

### DIFF
--- a/samples/SampleApp/Startup.cs
+++ b/samples/SampleApp/Startup.cs
@@ -9,6 +9,7 @@ using Microsoft.AspNetCore.Builder;
 using Microsoft.AspNetCore.Hosting;
 using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.Server.Kestrel.Transport.Abstractions.Internal;
+using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.Logging;
 
 namespace SampleApp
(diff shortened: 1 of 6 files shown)


samples/SampleApp/Startup.cs
```diff
@@ -40,6 +41,12 @@ public static void Main(string[] args)
                 Console.WriteLine("Unobserved exception: {0}", e.Exception);
             };
 
+            var configuration = new ConfigurationBuilder()
+                .AddEnvironmentVariables()
+                .Build();
+            if (!ushort.TryParse(configuration["BASE_PORT"], System.Globalization.NumberStyles.None, System.Globalization.CultureInfo.InvariantCulture, out var basePort))
```
cesarbs: Style nits:

- Add new line before the `if`
- Use a block `{ }`
- `using System.Globalization;`

halter73:
This definitely looks way less hacky. I also really like that this tests that directly binding to a port still works.

I wait to see if @cesarbs has any more comments before I merge, but I'm feeling a :shipit:

halter73:
Thanks!

Issue #1929, opened 2017-06-30
Increase default MinRequestBodyDataRate to match IIS's default


halter73:
I believe the ISS default is [240 bytes/second](https://www.iis.net/configreference/system.applicationhost/weblimits#005).

halter73:
@muratg @blowdart ping.

blowdart:
Yup, do it.

cesarbs:
Note that the IIS min rate is the response min rate.

240 bytes/second might be too aggressive and drop legit connections from slow clients. The current default ensures there is at least *some* (1 byte) network activity every second, without dropping clients with limited network resources.

PR #1931, opened 2017-06-30
Increase default request min rate to 240 bytes/second (#1929)

cesarbs:
#1929


src/Microsoft.AspNetCore.Server.Kestrel.Core/KestrelServerLimits.cs
```diff
@@ -258,8 +258,8 @@ public TimeSpan RequestHeadersTimeout
         /// This can be overridden per-request via <see cref="IHttpMinRequestBodyDataRateFeature"/>.
         /// </summary>
         /// <remarks>
-        /// Defaults to 1 byte/second with a 5 second grace period.
```
halter73: Can we add a regular non-doc comment indicating that this default comes from IIS like we do for a lot of our other limits?

PR #1932, opened 2017-06-30
Avoid torn reads from Frame.MinReadBodyDataRate in FrameConnection.Tick()

cesarbs:
If the rate is changed via `IHttpMinRequestBodyDataRateFeature` and `FrameConnection.Tick()` is running at the same time, the rate and the grace period might be read from different `MinDataRate` instances. This PR fixes that by reading `Frame.MinRequestBodyDataRate` into a local variable in `FrameConnection.Tick()`.


halter73:
Good catch. Did you see something break because of this?

cesarbs:
@halter73 No, I was just going through the code and realized that could happen.

Issue #1766, opened 2017-04-25
Flaky Test: ConnectionResetMidRequestIsLoggedAsDebug


mikeharder:
```
[12:12:42] :		 [Test collection for Microsoft.AspNetCore.Server.Kestrel.FunctionalTests.RequestTests (1)] ConnectionResetMidRequestIsLoggedAsDebug (1m:09s)
[12:12:42] :	 [Step 3/4]   The active test run was aborted. Reason: Unable to communicate with test host process.

ConnectionResetMidRequestIsLoggedAsDebug [FAIL]
  Assert.True() Failure
  Expected: True
  Actual:   False
  Stack Trace:
    RequestTests.cs(441,0): at Microsoft.AspNetCore.Server.Kestrel.FunctionalTests.RequestTests.<ConnectionResetMidRequestIsLoggedAsDebug>d__14.MoveNext()
    --- End of stack trace from previous location where exception was thrown ---
       at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
       at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
    --- End of stack trace from previous location where exception was thrown ---
       at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
       at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
    --- End of stack trace from previous location where exception was thrown ---
       at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
       at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
```

## Repros
4/25/17: http://aspnetci/viewLog.html?buildId=208284&buildTypeId=Lite_UniverseTest&tab=buildLog&filter=err

davidfowl:
I'm guessing this is https://github.com/aspnet/KestrelHttpServer/issues/1761

muratg:
@mikeharder did you see this after #1761 was resolved?

pakrym:
Still happening a log in Windows kestrel stress test configuration

mikeharder:
Just happened again today: http://aspnetci/viewLog.html?buildId=255643&tab=buildResultsDiv&buildTypeId=Lite_UniverseTest

mikeharder:
[...]

PR #1933, opened 2017-07-03
Fix flaky test: ConnectionResetMidRequestIsLoggedAsDebug

natemcmaster:
Resolves https://github.com/aspnet/KestrelHttpServer/issues/1766

This test was flaky because there was a race condition in the test. The call to `await context.Request.Body.ReadAsync(new byte[1], 0, 1);` does not guarantee the request will remain open long enough for the connection to be killed externally. I changed the body of the test server to ensure it waits long enough for this to happen.

### DIFF
--- a/test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/RequestTests.cs
+++ b/test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/RequestTests.cs
@@ -28,6 +28,7 @@ using Moq;
 using Newtonsoft.Json;
 using Newtonsoft.Json.Linq;
 using Xunit;
+using Xunit.Abstractions;
 
 namespace Microsoft.AspNetCore.Server.Kestrel.FunctionalTests
 {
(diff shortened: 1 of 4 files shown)


test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/RequestTests.cs
```diff
@@ -440,10 +447,11 @@ public void CanUpgradeRequestWithConnectionKeepAliveUpgradeHeader()
                 // This check MUST come before disposing the server, otherwise there's a race where the RST
                 // is still in flight when the connection is aborted, leading to the reset never being received
                 // and therefore not logged.
-                Assert.True(await connectionReset.WaitAsync(TimeSpan.FromSeconds(10)));
+                Assert.True(await connectionReset.WaitAsync(TimeSpan.FromSeconds(30)), "Connection reset event should have been logged");
+                connectionClosing.Release();
             }
 
-            Assert.False(loggedHigherThanDebug);
+            Assert.False(loggedHigherThanDebug, "Logged event should have been higher than debug.");
```
cesarbs: nit: should NOT have been higher than debug

natemcmaster: updated

natemcmaster:
Also, removing RuntimeIdentifier from test projects as it causes VS Code project load errors and is no longer required.

PR #1937, opened 2017-07-06
[Bad perf] Add response minimum data rate feature

cesarbs:
I need to fix a few details (like exception messages), but it's reviewable. I'm working on more tests.


test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/ResponseTests.cs
```diff
@@ -2437,6 +2437,53 @@ public Task ResponseStatusCodeSetBeforeHttpContextDisposedRequestMalformedReadIg
...
+                {
+                    connection.Socket.ReceiveBufferSize = 1;
+
+                    await connection.Send(
+                        "GET / HTTP/1.1",
+                        "Host:",
+                        "",
+                        "");
+
+                    Assert.True(aborted.Wait(TimeSpan.FromSeconds(60)));
```
natemcmaster: Can we also assert something about the connection state? it should be forcefully closed

test/Microsoft.AspNetCore.Server.Kestrel.Core.Tests/FrameConnectionTests.cs
```diff
@@ -364,5 +365,80 @@ public void ReadTimingNotPausedWhenResumeCalledBeforeNextTick()
...
+            var aborted = new ManualResetEventSlim();
+
+            _frameConnectionContext.ServiceContext.ServerOptions.Limits.MinResponseDataRate =
+                new MinDataRate(bytesPerSecond: 100, gracePeriod: TimeSpan.FromSeconds(2));
+            _frameConnectionContext.ServiceContext.SystemClock = systemClock;
+
+            var mockLogger = new Mock<IKestrelTrace>();
+            _frameConnectionContext.ServiceContext.Log = mockLogger.Object;
+
+            _frameConnection.CreateFrame(new DummyApplication(context => Task.CompletedTask), _frameConnectionContext.Input.Reader, _frameConnectionContext.Output);
```
natemcmaster: you can use the default `ctor` for `DummyApplication`

natemcmaster:
Can you share the benchmarks?

cesarbs:
This change kills perf.

**dev**: 1,196,131.76 RPS
**PR (with inline scheduler)**: 891,422.92
**PR (with threadpool scheduler)**: 469,364.33

I'm working on a different implementation without the loop in `OutputProducer` that has better perf. I'll send a PR shortly.

src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/Http/Frame.cs
```diff
@@ -410,15 +419,15 @@ private void CancelRequestAbortedToken()
         /// <summary>
         /// Immediate kill the connection and poison the request and response streams.
         /// </summary>
-        public void Abort(Exception error)
+        public void Abort(Exception error, bool abortOutputWithError = false)
```
cesarbs: It's not really about force-closing the output, it's just about propagating the error. With that in mind, do you have another suggestion?

I wouldn't want to make it a required parameter, since it's called with `true` in a single place. How about an overload?

cesarbs:
I've opened #1939 for an implementation that doesn't kill perf.

cesarbs:
Closing this in factor of the other PR.

PR #1939, opened 2017-07-07
Add response minimum data rate feature

cesarbs:
Alternative to #1937 that doesn't kill perf.

Instead of piping writes to a loop, it keeps track of each write's timeout in a queue of timeouts.

@natemcmaster I've addressed your feedback here (except the bit about connection state in `ResponseTests`, I'm currently working on that).


cesarbs:
Perf:

**dev**: 1,196,131.76 RPS
**PR**: 1,191,081.72

src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/FrameConnection.cs
```diff
@@ -381,5 +430,36 @@ public void BytesRead(int count)
...
+                var minResponseDataRate = _frame.MinResponseDataRate;
+
+                if (minResponseDataRate != null)
+                {
+                    var timeoutTicks = Math.Max(
+                        minResponseDataRate.GracePeriod.Ticks,
+                        TimeSpan.FromSeconds(size / minResponseDataRate.BytesPerSecond).Ticks);
+                    var timeoutTimestamp = (_writeTimingTimeouts.Count == 0
+                            ? _lastTimestamp
+                            : _writeTimingTimeouts.Last()) + timeoutTicks;
```
natemcmaster: This would be clearer if `timeoutTimestamp  += timeoutTicks` was on its own line. I missed it at first because it looks like it's part of the `false` condition.

Also, I believe the call to `Last()` is going to be expensive. It calls a `Queue.GetEnumerator()` and walks the collection to the end.

src/Microsoft.AspNetCore.Server.Kestrel.Core/Features/IHttpMinRequestBodyDataRateFeature.cs
```diff
@@ -4,7 +4,7 @@
 namespace Microsoft.AspNetCore.Server.Kestrel.Core.Features
 {
     /// <summary>
-    /// Represents a minimum data rate for the request body of an HTTP request.
+    /// Feature to set the minimum data rate at which the the request body should be received.
```
halter73: I think I would go with "should be received" to "must be sent".

src/Microsoft.AspNetCore.Server.Kestrel.Core/Features/IHttpMinResponseDataRateFeature.cs
```diff
@@ -0,0 +1,18 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.AspNetCore.Server.Kestrel.Core.Features
+{
+    /// <summary>
+    /// Feature to set the minimum data rate at which the response should be sent.
```
halter73: I would change "should be sent" to "must be received".

src/Microsoft.AspNetCore.Server.Kestrel.Core/Features/IHttpMinResponseDataRateFeature.cs
```diff
@@ -0,0 +1,18 @@
...
+    /// </summary>
+    public interface IHttpMinResponseDataRateFeature
+    {
+        /// <summary>
+        /// The minimum data rate in bytes/second at which the response should be sent.
```
halter73: diit.

src/Microsoft.AspNetCore.Server.Kestrel.Transport.Libuv/Internal/LibuvOutputConsumer.cs
```diff
@@ -82,6 +88,17 @@ public class LibuvOutputConsumer
             }
         }
 
+        private static void DisposeSocket(Exception ex, object state)
+        {
+            // If the pipe writer was closed with a TimeoutException, a write timed out.
+            // Disposing the socket will cancel the pending UvWriteReq.
+            if (ex is TimeoutException)
```
halter73: I think it's reasonable to have the contract be to close the socket immediate if the output/response pipe is completed with any exception. That would allow us to just make this a null check.

test/shared/TestServiceContext.cs
```diff
@@ -32,7 +33,11 @@ public TestServiceContext(ILoggerFactory loggerFactory, IKestrelTrace kestrelTra
             HttpParserFactory = frameAdapter => new HttpParser<FrameAdapter>(frameAdapter.Frame.ServiceContext.Log.IsEnabled(LogLevel.Information));
             ServerOptions = new KestrelServerOptions
             {
-                AddServerHeader = false
+                AddServerHeader = false,
+                Limits =
+                {
+                    MinResponseDataRate = new MinDataRate(bytesPerSecond: 1, gracePeriod: TimeSpan.MaxValue)
```
halter73: This is super scary. If we have to do this, let's try to make this change to specific tests.

src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/FrameConnection.cs
```diff
@@ -277,42 +302,66 @@ public void Tick(DateTimeOffset now)
                     Timeout();
                 }
             }
-            else
+        }
+
+        private void CheckForReadTimeout(long timestamp)
```
halter73: Nit: I would rename this methods to CheckForRead*Rate*Timeout and CheckForWriteRateTimeout to make it clear that these aren't absolute timeouts.

src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/FrameConnection.cs
```diff
@@ -277,42 +302,66 @@ public void Tick(DateTimeOffset now)
...
+        {
+            if (TimedOut || _timeoutTimestamp < long.MaxValue)
```
halter73: I think this deserves a comment.

We don't check for read timeouts if there's a _timeoutTimestamp set? Why not? Isn't this taken care of by _readTimingEnabled?

If we really don't want to do this, It would probably be both more correct an easier to read as `if (TimedOut || Interlocked.Read(_timeoutTimestamp) != long.MaxValue)`

src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/FrameConnection.cs
```diff
@@ -381,5 +430,38 @@ public void BytesRead(int count)
...
+
+                    if (_writeTimingTimeouts.Count == 0)
+                    {
+                        _writeTimingLastTimeoutTimestamp = _lastTimestamp;
+                    }
+
+                    // Add Heartbeat.Interval since this can be called right before the next heartbeat.
+                    _writeTimingLastTimeoutTimestamp += timeoutTicks + Heartbeat.Interval.Ticks;
+
+                    _writeTimingTimeouts.Enqueue(_writeTimingLastTimeoutTimestamp);
```
halter73: Instead of maintaining the queue, I would just add to the previous write timeout timestamp. While this could make a timeout take longer than it would with the queue, it's no different than a larger buffer being written.

cesarbs: Sounds good. It was my initial thought, but I figured it would be easier to remove the queue if I got that feedback than add it later 😁

src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/FrameConnection.cs
cesarbs: Added comment and a test.

src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/Http/Frame.cs
```diff
@@ -1382,6 +1385,15 @@ private IPipe CreateRequestBodyPipe()
                 MaximumSizeLow = 1
             });
 
+        private IPipe CreateResponsePipe()
```
natemcmaster: Is this still used anywhere?

natemcmaster:
What do benchmarks look like that we've dropped the queue ?

src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/FrameConnection.cs
```diff
@@ -381,5 +430,37 @@ public void BytesRead(int count)
...
+
+                if (minResponseDataRate != null)
+                {
+                    var timeoutTicks = Math.Max(
+                        minResponseDataRate.GracePeriod.Ticks,
+                        TimeSpan.FromSeconds(size / minResponseDataRate.BytesPerSecond).Ticks);
+
+                    if (_writeTimingWrites == 0)
+                    {
+                        _writeTimingTimeoutTimestamp = _lastTimestamp;
```
halter73: I think you can just add `Heartbeat.Interval.Ticks` here and not below.

cesarbs:
@natemcmaster Still good, 1,194,528.36 RPS.

src/Microsoft.AspNetCore.Server.Kestrel.Core/MinDataRate.cs
```diff
@@ -16,9 +16,9 @@ public class MinDataRate
         /// starting at the time data is first read or written.</param>
         public MinDataRate(double bytesPerSecond, TimeSpan gracePeriod)
         {
-            if (bytesPerSecond < 0)
+            if (bytesPerSecond <= 0)
             {
-                throw new ArgumentOutOfRangeException(nameof(bytesPerSecond), CoreStrings.NonNegativeNumberRequired);
+                throw new ArgumentOutOfRangeException(nameof(bytesPerSecond), CoreStrings.PositiveNumberRequired);
```
halter73: I would like to include something like the following in the exception message. "If you want to disable a minimum rate, use a null MinDataRate instead."

test/shared/TestServiceContext.cs
```diff
@@ -1,6 +1,7 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System;
```
halter73: Nit: remove using.

PR #1940, opened 2017-07-07
Validate certificate EKU when it is provided

### LINKED ISSUE #1764: Validate certificate EKU
When setting up SSL with Kestrel if the cert has an EKU we should validate that the cert is for server usage.

natemcmaster:
When a certificate has an enhanced key usage section, this validates that this section contains `serverAuth`.

RFC https://tools.ietf.org/html/rfc3280#section-4.2.1.13
See also https://blogs.msdn.microsoft.com/kaushal/2012/02/17/client-certificates-vs-server-certificates/

Resolves #1764


src/Microsoft.AspNetCore.Server.Kestrel.Https/HttpsConnectionAdapter.cs
```diff
@@ -119,6 +131,32 @@ public Task<IAdaptedConnection> OnConnectionAsync(ConnectionAdapterContext conte
...
+
+            foreach (var extension in certificate.Extensions.OfType<X509EnhancedKeyUsageExtension>())
+            {
+                hasEkuExtension = true;
+                foreach (var oid in extension.EnhancedKeyUsages)
+                {
+                    if (oid.Value.Equals(ServerAuthenticationOid.Value, StringComparison.Ordinal))
+                    {
+                        allowedForServerAuth = true;
+                        goto finish;
```
cesarbs: Can you replace this second `foreach` with a Linq query (use `Any()`) so you don't need the `goto`?

src/Microsoft.AspNetCore.Server.Kestrel.Https/HttpsConnectionAdapter.cs
```diff
@@ -37,6 +44,11 @@ public HttpsConnectionAdapter(HttpsConnectionAdapterOptions options, ILoggerFact
                 throw new ArgumentException(HttpsStrings.ServiceCertificateRequired, nameof(options));
             }
 
+            // capture the certificate now so it can be switched after validation
+            _serverCertificate = options.ServerCertificate;
+
+            ValidateEku(_serverCertificate);
```
cesarbs: I'm not a fan of having logic in a ctor. Have you considered validating in the `HttpConnectionAdapterOptions.ServerCertificate` setter?

src/Microsoft.AspNetCore.Server.Kestrel.Https/HttpsConnectionAdapter.cs
natemcmaster: I wish. X509EnhancedKeyUsageExtension.EnhancedKeyUsages is an [OidCollection](https://docs.microsoft.com/en-us/dotnet/api/system.security.cryptography.oidcollection?view=netstandard-2.0), which doesn't implement IEnumerable\<T>.

src/Microsoft.AspNetCore.Server.Kestrel.Https/HttpsConnectionAdapter.cs
natemcmaster: They're functionally equivalent, so I don't care either way...but normally we like our "Options" typos to be POCOs, so I opted to put the validation here.

natemcmaster:
Hmm, on macOS this throws
```c#
private static readonly Oid ServerAuthenticationOid = Oid.FromOidValue("1.3.6.1.5.5.7.3.1", OidGroup.EnhancedKeyUsage)
```

cc @bartonjs ideas?

**Error**
```
System.Security.Cryptography.CryptographicException : The OID value is invalid.
```

bartonjs:
Windows localizes the friendly name, so it isn't in our static lookup table.  macOS doesn't expose a OID to name lookup function, so we throw with it being unresolvable.

If no one ever sees the Oid object you can just do `new Oid(value, value)` or `new Oid(value, null)`.  Or, since you're not using the Oid object, just the const string of the dotted decimal :smile:.

natemcmaster:
Thanks @bartonjs!

natemcmaster:
:up: :date: passing on macOS, linux and Windows now.

src/Microsoft.AspNetCore.Server.Kestrel.Https/HttpsConnectionAdapter.cs
natemcmaster: Nvm. TIL there is away to do this. `IEnumerable<T> Cast<T>(this ICollection collection)`

natemcmaster:
@blowdart On a related note, should we also be checking [`X509​Key​Usage​Extension.KeyUsages`](https://docs.microsoft.com/en-us/dotnet/api/system.security.cryptography.x509certificates.x509keyusageextension.keyusages?view=netframework-4.7#System_Security_Cryptography_X509Certificates_X509KeyUsageExtension_KeyUsages)? This is separate from EKU. Values include:

```c#
    [Flags]
    public enum X509KeyUsageFlags
    {
        None = 0x0000,
        EncipherOnly = 0x0001,
        CrlSign = 0x0002,
        KeyCertSign = 0x0004,
        KeyAgreement = 0x0008,
        DataEncipherment = 0x0010,
        KeyEncipherment = 0x0020,
        NonRepudiation = 0x0040,
        DigitalSignature = 0x0080,
        DecipherOnly = 0x8000,
    }
```

blowdart:
yes, look for Digital Signature & Key Encipherment (a0)

KestrelHttpServer.sln
```diff
@@ -54,6 +54,17 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Microsoft.AspNetCore.Server
 EndProject
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "TestResources", "TestResources", "{2822C132-BFFB-4D53-AC5B-E7E47DD81A6E}"
 	ProjectSection(SolutionItems) = preProject
+		test\shared\TestResources\eku.none.ini = test\shared\TestResources\eku.none.ini
```
halter73: Nit: Might as well rename the folder to TestCertificates at this point.

natemcmaster: updated.

natemcmaster:
@blowdart okay thanks, I'll add a separate PR to validate KU as well.

bartonjs:
@blowdart not... quite.  Key Encipherment is only used in SSL3, I think.   It's definitely not involved in DSA/ECDSA certs.  And not involved in ECDHE-RSA or DHE-RSA cipher suites.

natemcmaster:
FYI - https://github.com/aspnet/KestrelHttpServer/issues/1946. We decided we need to look into this more before we add Key Usage restrictions to kestrel.

Drawaes:
Its used in TLS 1.x if you don't use any of the forward secrecy algos (ECDHE and DHE) but will be used if it is ECDH, DH or RSA directly. TLS1.3 is the first that would not require this at all for any of the supported ciphers.

Drawaes:
Actually I could be wrong on the ECDH/DH, anyway it seems to be pretty standard to accept a cert with EITHER of those... not requiring both

PR #1942, opened 2017-07-07
Add XML docs for the public HTTPs APIs shipping in 2.0

natemcmaster:
Add XML docs for ClientCertificateMode and HttpsConnectionAdapterOptions.

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Kestrel.Https/HttpsConnectionAdapterOptions.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel.Https/HttpsConnectionAdapterOptions.cs
@@ -8,18 +8,43 @@ using System.Security.Cryptography.X509Certificates;
 
 namespace Microsoft.AspNetCore.Server.Kestrel.Https
 {
+    /// <summary>
+    /// Settings for how Kestrel should handle HTTPS connections.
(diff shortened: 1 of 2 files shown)


src/Microsoft.AspNetCore.Server.Kestrel.Https/HttpsConnectionAdapterOptions.cs
```diff
@@ -8,18 +8,43 @@
...
+        /// </summary>
         public ClientCertificateMode ClientCertificateMode { get; set; }
+
+        /// <summary>
+        /// Specifies a callback for additional client certificate validation that will be invoked during authentication.
+        /// </summary>
         public Func<X509Certificate2, X509Chain, SslPolicyErrors, bool> ClientCertificateValidation { get; set; }
+
+        /// <summary>
+        /// Specifies allowable SSL protocols to be used for authentication. Defaults to either <see cref="SslProtocols.Tls12" /> or <see cref="SslProtocols.Tls11"/>.
```
cesarbs: I think this should be "and" instead of "or".

natemcmaster: updated.

PR #1944, opened 2017-07-07
Validate key usages on certificates when the extension is provided

natemcmaster:
Adds additional server certificate validation.
 #1940 validates the Extended Key Usage extension (RFC 3280 section 4.2.1.**13**). 
This change validates the Key Usage extension. (RFC 3280 section 4.2.1.**3**).

When the Key Usage extension extension is present, HttpsConnectionAdapter will assert the server certificate is marked for both Digital Signature & Key Encipherment usages. Additional usages may be present. If the assertions fail, this will throw `CryptographicException` before the server starts.


natemcmaster:
Closing for now. Needs more investigation. We'll use https://github.com/aspnet/KestrelHttpServer/issues/1946 to track this.

Issue #1850, opened 2017-05-22
Test failure on .NET 4.6.1: MaxRequestBufferSizeTests.LargeUpload


cesarbs:
Failing occasionally specifically with these inputs: `LargeUpload(maxRequestBufferSize: 20971519, connectionAdapter: False, expectPause: False)`:

```
[xUnit.net 00:01:14.9070146]     LargeUpload(maxRequestBufferSize: 20971519, connectionAdapter: False, expectPause: False) [FAIL]
[xUnit.net 00:01:14.9985686]       System.IO.IOException : Unable to write data to the transport connection: An existing connection was forcibly closed by the remote host.
[xUnit.net 00:01:14.9986601]       ---- System.Net.Sockets.SocketException : An existing connection was forcibly closed by the remote host
[xUnit.net 00:01:15.1096310]       Stack Trace:
[xUnit.net 00:01:15.1413459]            at System.Net.Sockets.NetworkStream.BeginWrite(Byte[] buffer, Int32 offset, Int32 size, AsyncCallback callback, Object state)
[xUnit.net 00:01:15.1691890]            at System.IO.Stream.<>c.<BeginEndWriteAsync>b__53_0(Stream stream, ReadWriteParameters args, AsyncCallback callback, Object state)
[xUnit.net 00:01:15.2289030]            at System.Threading.Tasks.TaskFactory`1.FromAsyncTrim[TInstance,TArgs](TInstance thisRef, TArgs args, Func`5 beginMethod, Func`3 endMethod)
[xUnit.net 00:01:15.2290481]            at System.IO.Stream.BeginEndWriteAsync(Byte[] buffer, Int32 offset, Int32 count)
[xUnit.net 00:01:15.2291153]            at System.IO.Stream.WriteAsync(Byte[] buffer, Int32 offset, Int32 count, CancellationToken cancellationToken)
[xUnit.net 00:01:15.2291854]            at System.IO.Stream.WriteAsync(Byte[] buffer, Int32 offset, Int32 count)
[xUnit.net 00:01:15.2292631]         MaxRequestBufferSizeTests.cs(111,0): at Microsoft.AspNetCore.Server.Kestrel.FunctionalTests.MaxRequestBufferSizeTests.<>c__DisplayClass4_1.<<LargeUpload>b__0>d.MoveNext()
[xUnit.net 00:01:15.2293695]         --- End of stack trace from previous location where exception was thrown ---
[...]

PR #1950, opened 2017-07-10
Fix flakiness in MaxRequestBufferSizeTests.LargeUpload (#1850)

cesarbs:
#1850

Turns out it was just a timeout issue. I suspected when I saw it passing when either increasing the timeout or increasing the client send size from 4KB to something large like 1MB per send (total payload size is 20MB). Also, it only reproed under load (running all tests).

After increasing the timeout, I measure how long it took for the task to be completed and it was around 1 minute. Giving it 2 minutes should be enough margin to avoid future issues.

Since logging helped me debug this, I'd like to leave it enabled in case we run into any issues with these tests again in the future.

### DIFF
--- a/test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/MaxRequestBufferSizeTests.cs
+++ b/test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/MaxRequestBufferSizeTests.cs
@@ -32,7 +32,10 @@ namespace Microsoft.AspNetCore.Server.Kestrel.FunctionalTests
 
         private readonly Action<ILoggingBuilder> _configureLoggingDelegate;
 
-        public MaxRequestBufferSizeTests(ITestOutputHelper output) =>_configureLoggingDelegate = builder => builder.AddXunit(output);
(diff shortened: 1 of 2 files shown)


test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/MaxRequestBufferSizeTests.cs
```diff
@@ -28,6 +30,10 @@ public class MaxRequestBufferSizeTests
             "\r\n"
         };
 
+        private readonly Action<ILoggingBuilder> _configureLoggingDelegate;
+
+        public MaxRequestBufferSizeTests(ITestOutputHelper output) =>_configureLoggingDelegate = builder => builder.AddXunit(output);
```
natemcmaster: Took me a second to realize this is a constructor. imo the arrow-body syntax obscures this a little too much. Would be better to use brackets.

cesarbs: 👍 Also changed the `AddressRegistrationTests` ctor.

natemcmaster:
:shipit:

Issue #1666, opened 2017-04-13
Make StopAsync multi-thread safe


Tratcher:
@davidfowl wants any parallel calls to KestrelServer.StopAsync to gate on the first call rather than no-op. IWebHost.StopAsync may need this logic, though putting it in both places seems redundant.

davidfowl:
It's not that I want it, I just want it to be right 👍

muratg:
@cesarbs We need to do the same in HttpSysServer.

cesarbs:
I've opened https://github.com/aspnet/HttpSysServer/issues/377 to track this on HttpSysServer.

PR #1951, opened 2017-07-11
Make StopAsync multi-thread safe (#1666)

cesarbs:
#1666

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Kestrel.Core/KestrelServer.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel.Core/KestrelServer.cs
@@ -157,34 +157,42 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core
         {
             if (Interlocked.Exchange(ref _stopping, 1) == 1)
             {
-                await _stoppedTcs.Task;
+                await _stoppedTcs.Task.ConfigureAwait(false);
                 return;
             }
 
-            var tasks = new Task[_transports.Count];
-            for (int i = 0; i < _transports.Count; i++)
+            try
             {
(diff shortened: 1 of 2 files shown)


src/Microsoft.AspNetCore.Server.Kestrel.Core/KestrelServer.cs
```diff
@@ -154,8 +155,9 @@ private static ServiceContext CreateServiceContext(IOptions<KestrelServerOptions
         // Graceful shutdown if possible
         public async Task StopAsync(CancellationToken cancellationToken)
         {
-            if (Interlocked.Exchange(ref _stopped, 1) == 1)
+            if (Interlocked.Exchange(ref _stopping, 1) == 1)
             {
+                await _stoppedTcs.Task;
```
davidfowl: `ConfigureAwait(false)`

src/Microsoft.AspNetCore.Server.Kestrel.Core/KestrelServer.cs
```diff
@@ -183,6 +185,8 @@ private static ServiceContext CreateServiceContext(IOptions<KestrelServerOptions
             await Task.WhenAll(tasks).ConfigureAwait(false);
 
             _heartbeat.Dispose();
+
+            _stoppedTcs.TrySetResult(null);
```
Tratcher: this needs to be in a finally block, or it may not get released on failures.

cesarbs: Instead of putting it in a finally, I'm catching any exception and completing the task with it and rethrowing.

Issue #1742, opened 2017-04-22
Flaky Test: RejectsRequestWithContentLengthAndUpgrade


davidfowl:
Saw this once on AppVeyor after @natemcmaster 's change.

```
[xUnit.net 00:00:00.6503592]   Discovering: Microsoft.AspNetCore.Server.Kestrel.FunctionalTests
[xUnit.net 00:00:02.0149756]   Discovered:  Microsoft.AspNetCore.Server.Kestrel.FunctionalTests
[xUnit.net 00:00:02.5873241]   Starting:    Microsoft.AspNetCore.Server.Kestrel.FunctionalTests
The running command stopped because the preference variable "ErrorActionPreference" or common parameter is set to Stop: [xUnit.net 00:00:12.8264301]     RejectsRequestWithContentLengthAndUpgrade [FAIL]
```
Not sure why it failed. If it doesn't happen again we can close this.

Exception from CI build log:

```
[Microsoft.AspNetCore.Server.Kestrel.FunctionalTests.UpgradeTests.RejectsRequestWithContentLengthAndUpgrade]
System.IO.IOException : Unable to read data from the transport connection:
An established connection was aborted by the software in your host machine.
---- System.Net.Sockets.SocketException : An established connection was aborted by the software in your host machine
   at System.Net.Sockets.NetworkStream.BeginRead(Byte[] buffer, Int32 offset, Int32 size, AsyncCallback callback, Object state)
   at System.IO.Stream.<>c.<BeginEndReadAsync>b__43_0(Stream stream, ReadWriteParameters args, AsyncCallback callback, Object state)
   at System.Threading.Tasks.TaskFactory`1.FromAsyncTrim[TInstance,TArgs](TInstance thisRef, TArgs args, Func`5 beginMethod, Func`3 endMethod)
   at System.IO.Stream.BeginEndReadAsync(Byte[] buffer, Int32 offset, Int32 count)
   at System.IO.Stream.ReadAsync(Byte[] buffer, Int32 offset, Int32 count, CancellationToken cancellationToken)
   at System.IO.Stream.ReadAsync(Byte[] buffer, Int32 offset, Int32 count)
   at System.IO.StreamReader.<ReadBufferAsync>d__97.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
[...]

PR #1952, opened 2017-07-11
Fix flakiness in RejectsRequestWithContentLengthAndUpgrade (#1742)

cesarbs:
Sorry @halter73 for poaching #1742 from you, but I saw it fail a number of times today and figured the fix was very simple.

Although the stack trace shown when this test fails always comes from a receive call, the problem is really in trying to send that byte in the request body. If the server sends the response and closes the connection while that byte is in still flight, the connection is reset and any subsequent operation on the socket fails.

Since we're only testing that an error is returned when there are payload headers along with a `Connection: upgrade` header, not sending a payload doesn't affect the purpose of this test. We don't even attempt to read it in the app (which never really gets to run anyways).

### DIFF
--- a/test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/UpgradeTests.cs
+++ b/test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/UpgradeTests.cs
@@ -163,7 +163,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.FunctionalTests
                     "Content-Length: 1",
                     "Connection: Upgrade",
                     "",
-                    "1");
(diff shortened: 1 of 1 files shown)


natemcmaster:
Would it be better to preserve the correctness of the POST body and wrap the send in a try/catch that swallows?

cesarbs:
@natemcmaster We could do that, but we'd have to do the same for the receive. I don't think that's a better alternative, though.

halter73:
Never apologize for fixing flaky tests!

Interesting that not sending the body fixes this considering that Send() completes successfully and that it's ReceiveForcedEnd that fails, but if it works, great 😄

Issue #640, opened 2016-02-18
Create a logging scope per connection


halter73:
After seeing [HostingLogScope](https://github.com/aspnet/Hosting/blob/906ed5f0fb5408a0af5726e6bb201f7e67f88fe7/src/Microsoft.AspNetCore.Hosting/Internal/HostingLoggerExtensions.cs#L91), I realized we need the equivalent for Kestrel connections.

11 of the 13 special log functions defined in `KestrelTrace` take the connect id as a logging parameter. Of the remaining 2, `KestrelTrace.ApplicationError` _really should_ log the connection id but doesn't. It would be far more elegant to create a logging scope that includes the connection id assuming that these can be nested. This will also prevent future oversights like the one in `ApplicationError`.

An added benefit is that we could lazily serialize the connection id like we do with the request id since we're not constantly passing a connection id string to logging methods. See #638

_Note:_ we will have to be careful when logging outside of the connection's async loop (such as in callbacks on the libuv thread) since the AsyncLocal backing for the scope (I assume that what's being used) won't be available. In those cases I imagine we log connection id's like we do today, but check the loglevel before calling the log method so we don't have to serialize the connection id unnecessarily. That's assuming we can't delay the logging until we get back on the async loop.

benaadams:
Isn't `AsyncLocal` for `HostingLogScope` as `AsyncLocal` modifies the execution context from default and would be seeing lots of `ExectionContext` allocations like you see when using the `HttpContextAccessor`

halter73:
Ultimately I think it's up to the logger how they implement scopes. The [console logger uses AsyncLocal/CallContext](https://github.com/aspnet/Logging/blob/46c6d523fd323eba101f230b96111d3fa4dcfa46/src/Microsoft.Extensions.Logging.Console/ConsoleLogScope.cs#L48), but it seems you would only pay for it if you use it.

[...]

PR #1953, opened 2017-07-11
Add support for connection scopes if logging is enabled

davidfowl:
- Don't create a scope if logging isn't on
- Copied the pattern we use in Hosting

Questions:
- Should we remove the connection id parameter from the logs? I'd suggesting removing it only if the log message is readable without it. For e.g we could say "Connected started" instead of "Connection {ConnectionId} started" but it doesn't really hurt.

Before:

image

After:

image

#640

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/ConnectionLogScope.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/ConnectionLogScope.cs
@@ -1,4 +1,7 @@
-﻿using System;
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
 using System.Collections;
 using System.Collections.Generic;
 using System.Globalization;
--- a/src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/ConnectionLogScope.cs
+++ b/src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/ConnectionLogScope.cs
@@ -1,0 +1,60 @@
+﻿using System;
+using System.Collections;
+using System.Collections.Generic;
+using System.Globalization;
+
+namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal
+{
+    public class ConnectionLogScope : IReadOnlyList<KeyValuePair<string, object>>
+    {
+        private readonly string _connectionId;
+
+        private string _cachedToString;
+
+        public ConnectionLogScope(string connectionId)
+        {
+            _connectionId = connectionId;
+        }
+
+        public KeyValuePair<string, object> this[int index]
+        {
+            get
+            {
+                if (index == 0)
+                {
+                    return new KeyValuePair<string, object>("ConnectionId", _connectionId);
+                }
+
+                throw new ArgumentOutOfRangeException(nameof(index));
+            }
+        }
+
+        public int Count => 1;
+
+        public IEnumerator<KeyValuePair<string, object>> GetEnumerator()
+        {
+            for (int i = 0; i < Count; ++i)
+            {
+                yield return this[i];
+            }
+        }
+
+        IEnumerator IEnumerable.GetEnumerator()
+        {
+            return GetEnumerator();
+        }
+
+        public override string ToString()
+        {
+            if (_cachedToString == null)
+            {
+                _cachedToString = string.Format(
+                    CultureInfo.InvariantCulture,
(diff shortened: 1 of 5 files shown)


cesarbs:
Should we have a scope per request as well?

Tratcher:
Hosting creates one right?

src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/FrameConnection.cs
```diff
@@ -454,5 +460,68 @@ public void StopTimingWrite()
...
+        {
+            if (Log.IsEnabled(LogLevel.Critical))
+            {
+                return Log.BeginScope(new ConnectionScope(ConnectionId));
+            }
+
+            return null;
+        }
+
+        private class ConnectionScope : IReadOnlyList<KeyValuePair<string, object>>
```
halter73: Move to another file.

src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/FrameConnection.cs
```diff
@@ -84,67 +86,71 @@ public void StartRequestProcessing<TContext>(IHttpApplication<TContext> applicat
...
-                Log.ConnectionStart(ConnectionId);
-                KestrelEventSource.Log.ConnectionStart(this, _context.ConnectionInformation);
+                try
+                {
 
-                AdaptedPipeline adaptedPipeline = null;
-                var adaptedPipelineTask = Task.CompletedTask;
-                var input = _context.Input.Reader;
-                var output = _context.Output;
+                    Log.ConnectionStart(ConnectionId);
```
halter73: Nit. No need for newline after `try {`

src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/FrameConnection.cs
```diff
@@ -454,5 +460,68 @@ public void StopTimingWrite()
...
+                return GetEnumerator();
+            }
+
+            public override string ToString()
+            {
+                if (_cachedToString == null)
+                {
+                    _cachedToString = string.Format(
+                        CultureInfo.InvariantCulture,
+                        "ConnectionId:{0}",
```
halter73: Should we remove `@"Connection id ""{ConnectionId}""` from KestrelTrace and LibuvTrace for logs we no we make from the async local scope? It seems a bit redundant if we don't remove it, but it could lead to a worse experience with loggers that don't support scopes.

halter73: Also now that we have this should we log any extra metadata about the connection (e.g. remote IP) when a connection starts?

natemcmaster: How well do our [own loggers](https://github.com/aspnet/Logging) support scopes? e.g. console, Azure app services, etc. If these don't easily display the name of the scope along with the message, we should keep the connection id in the message for now.

halter73: @natemcmaster That's what I was concerned about. Assuming this is for 2.1, maybe we should still do this and add scope support to our loggers.

src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/FrameConnection.cs
davidfowl: Why? It's a private class literally only used in this file.

src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/FrameConnection.cs
davidfowl: I think we should keep it in the logs.

src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/FrameConnection.cs
halter73: That's also true for all the dozens of what used to be private classes we moved to their own files in Kestrel. FrameConnection is already pretty big.

src/Microsoft.AspNetCore.Server.Kestrel.Core/Internal/ConnectionLogScope.cs
```diff
@@ -0,0 +1,60 @@
+﻿using System;
```
natemcmaster: Nit: copyright header

Issue #1956, opened 2017-07-12
Response.Body.FlushAsync() closes connection if a connection adapter is enabled


halter73:
If you change the [call to Response.WriteAsync(response)](https://github.com/aspnet/KestrelHttpServer/blob/7ebbdad9746f21b5966c62d68717fa68e21df72d/samples/SampleApp/Startup.cs#L34) in Kestrel's sample app to the following, only "hello" will be written. As soon as Response.Body.FlushAsync() is called, Kestrel closes the connection.

```C#
await context.Response.WriteAsync(response.Substring(0, 5));
await context.Response.Body.FlushAsync();
await context.Response.WriteAsync(response.Substring(5));
```

Here is part of the log output:

```
System.ArgumentNullException: Value cannot be null.
Parameter name: array
   at Microsoft.AspNetCore.Server.Kestrel.Internal.System.ThrowHelper.ThrowArgumentNullException(ExceptionArgument argument) in C:\Users\shalter\.nuget\packages\microsoft.aspnetcore.internal.corefxlab.sources\2.0.0-preview2-21341\contentFiles\cs\netstandard1.0\System.Memory\src\System\ThrowHelper.cs:line 36
   at Microsoft.AspNetCore.Server.Kestrel.Internal.System.ReadOnlySpan`1..ctor(T[] array, Int32 start, Int32 length) in C:\Users\shalter\.nuget\packages\microsoft.aspnetcore.internal.corefxlab.sources\2.0.0-preview2-21341\contentFiles\cs\netstandard1.0\System.Memory\src\System\ReadOnlySpan.cs:line 92
   at Microsoft.AspNetCore.Server.Kestrel.Internal.System.ReadOnlySpan`1.op_Implicit(ArraySegment`1 arraySegment) in C:\Users\shalter\.nuget\packages\microsoft.aspnetcore.internal.corefxlab.sources\2.0.0-preview2-21341\contentFiles\cs\netstandard1.0\System.Memory\src\System\ReadOnlySpan.cs:line 281
   at Microsoft.AspNetCore.Server.Kestrel.Core.Adapter.Internal.RawStream.<WriteAsync>d__19.MoveNext() in D:\dev\aspnet\KestrelHttpServer\src\Microsoft.AspNetCore.Server.Kestrel.Core\Adapter\Internal\RawStream.cs:line 88
--- End of stack trace from previous location where exception was thrown ---
[...]

PR #1957, opened 2017-07-12
Fix FlushAsync when a ConnectionAdapter is configured

halter73:
#1956


src/Microsoft.AspNetCore.Server.Kestrel.Transport.Sockets/SocketConnection.cs
```diff
@@ -38,7 +38,7 @@ internal SocketConnection(Socket socket, SocketTransport transport)
             _remoteEndPoint = (IPEndPoint)_socket.RemoteEndPoint;
         }
 
-        public async void Start(IConnectionHandler connectionHandler)
+        public async Task Start(IConnectionHandler connectionHandler)
```
cesarbs: Can we rename this to `StartAsync`?

PR #1963, opened 2017-07-20
React to pipeline changes

pakrym:
https://github.com/dotnet/corefxlab/pull/1668

### DIFF
--- a/test/Microsoft.AspNetCore.Server.Kestrel.Performance/PipeThroughputBenchmark.cs
+++ b/test/Microsoft.AspNetCore.Server.Kestrel.Performance/PipeThroughputBenchmark.cs
@@ -38,7 +38,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Performance
 
             var reading = Task.Run(async () =>
             {
-                int remaining = InnerLoopCount * _writeLenght;
+                long remaining = InnerLoopCount * _writeLenght;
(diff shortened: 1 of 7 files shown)


test/Microsoft.AspNetCore.Server.Kestrel.Performance/PipeThroughputBenchmark.cs
```diff
@@ -38,7 +38,7 @@ public void ParseLiveAspNetTwoTasks()
 
             var reading = Task.Run(async () =>
             {
-                int remaining = InnerLoopCount * _writeLenght;
+                long remaining = InnerLoopCount * _writeLenght;
```
halter73: Not new, but nit: _writeLeng*th*.

halter73: Also why? It looks like you're multiplying two pretty small consts.

pakrym: Because of `remaining -= result.Buffer.Length;` it looked better changing this `int` to `long` vs casting,

pakrym:
https://github.com/aspnet/KestrelHttpServer/commit/fd6617d1019b8e5009ad3f8a8991483e8bff7eb1

PR #1976, opened 2017-07-31
 	Consume corefxlab packages and private build of C# compiler

pakrym:
Depends on: https://github.com/dotnet/corefxlab/pull/1674
Compiler VSIX: https://dotnet.myget.org/F/roslyn/vsix/0b48e25b-9903-4d8b-ad39-d4cca196e3c7-2.6.0.6191501.vsix


build/common.props
```diff
@@ -11,10 +11,12 @@
     <PublicSign Condition="'$(OS)' != 'Windows_NT'">true</PublicSign>
     <VersionSuffix Condition="'$(VersionSuffix)'!='' AND '$(BuildNumber)' != ''">$(VersionSuffix)-$(BuildNumber)</VersionSuffix>
     <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
+    <LangVersion>7.2</LangVersion>
```
natemcmaster: Does `<LangVersion>latest</LangVersion>` work?

samples/LargeResponseApp/LargeResponseApp.csproj
```diff
@@ -11,4 +11,8 @@
     <ProjectReference Include="..\..\src\Kestrel\Kestrel.csproj" />
   </ItemGroup>
 
+  <ItemGroup>
+    <PackageReference Include="Microsoft.NETCore.Compilers" Version="$(MicrosoftNETCoreCompilersVersion)" PrivateAssets="All" />
```
natemcmaster: Is this just for consistency within the repro? Or does these mean all consumers of Kestrel would need a new compiler?

pakrym: This is required because when we build entire repo in one invocation having two versions of compilers in a single process causes load exception.

build/common.props
pakrym: Yes

test/Kestrel.FunctionalTests/SystemdActivation/docker.sh
```diff
@@ -3,22 +3,23 @@
 set -e
 
 scriptDir=$(dirname "${BASH_SOURCE[0]}")
-~/.dotnet/dotnet publish -f netcoreapp2.0 ./samples/SampleApp/
+PATH="$HOME/.dotnet/:$PATH"
```
pakrym: /cc @halter73

PR #1987, opened 2017-08-09
Remove dependency on a custom task to detect when compiling on macOS

natemcmaster:
MSBuild.IsOSPlatform is new to MSBuild 15.3 and can be used to determine when we are building on macOS.


natemcmaster:
Failing on https://github.com/aspnet/KestrelHttpServer/issues/1684, but I verified locally that this builds correctly on macOS and Windows

PR #1989, opened 2017-08-10
[WIP]: Bedrock refactoring (don't code review the code but look at the PR comment)

davidfowl:
So I hacked together a branch (non working) to see what approach we could take to refactor Kestrel further to prepare for bedrock and @cesarbs HTTP2 work.

There are a bunch of high level decisions we need to make about the approach before we go full steam ahead:

- Breaking changes and deprecating old APIs
 - Should IConnectionAdapters be deprecated or removed completely? We moved them to the .Internal namespace to make sure we sent a clear signal that they were going to be changing.
 - We have some top level types that are not in the internal namespace but are exposed from Kestrel.Core as first class APIs. We have a few options here:
   - Introduce a new "Core" assembly that Kestrel.Core uses under the covers. We could preserve all of the top  level API but swap the implementation to use this new core assembly.
   - We can use type forwards to move implementations to new assemblies. We need to preserve the namespace and type names but the implementations would be elsewhere.
   - We break things (non starter).
- We should deprecate Transport.Abstractions. I think these types should live in Protocols.Abstractions. I've moved a few of the primitives there and it seems to fit well. IEndPointInformation needs clean up.
- `IConnectionHandler` should take the `ConnectionContext` directly https://github.com/aspnet/KestrelHttpServer/blob/davidfowl/bedrock-plus-plus/src/Kestrel.Transport.Abstractions/Internal/IConnectionHandler.cs#L10.
- Transports should create the `ConnectionContext` and `IFeatureCollection` instead of passing an `IConnectionInformation` to the `IConnectionHandler`  https://github.com/aspnet/KestrelHttpServer/blob/davidfowl/bedrock-plus-plus/src/Kestrel.Transport.Libuv/Internal/LibuvConnection.cs#L52. This also means that settings for scheduling and pipe limits live with the transport instead of with the dispatcher (in this case Kestrel.Core).
- `IKestrelTrace` is a monolith and needs to be broken into micro interfaces 😄 .
- Where does the ASP.NET `IServer` implementation live? If we decide to keep Kestrel.Core and introduce a new core assembly, the decision is easy. We keep it in the Core and use the underlying server API.
- Introducing a new top level `ServerBuilder` API. To decouple everything, we need a new top level server API that pretty much matches the extension we have on `IWebHostBuilder` today. See https://github.com/aspnet/KestrelHttpServer/blob/davidfowl/bedrock-plus-plus/samples/SampleApp/Startup.cs#L58-L76 for a sample. This could be used as the underlying implementation of the the extension methods.
- Who keeps track of the outstanding connections? Does the connection middleware need to handle this on its own? If so, we need a way to signal shutdown so it can clean up.
- The HTTP server implementation (which I'd like to decouple from the Core) needs things like the Heartbeat, ISystemClock, FrameConnectionManager and a few other cross cutting concerns. How should it get those? The entry point is here https://github.com/aspnet/KestrelHttpServer/blob/davidfowl/bedrock-plus-plus/src/Protocols.Http/HttpServerConnectionBuilderExtensions.cs.
- We don't have a good way to have different settings per binding. If you want to wire up different server instances with different settings that come from DI, we can use named options, but we don't support that today. For example, [HttpServerLimits](https://github.com/aspnet/KestrelHttpServer/blob/davidfowl/bedrock-plus-plus/src/Protocols.Http/HttpServerLimits.cs) and [HttpServerOptions](https://github.com/aspnet/KestrelHttpServer/blob/davidfowl/bedrock-plus-plus/src/Protocols.Http/HttpServerOptions.cs). Another example is the Logger name used by different instances of the HttpServer. If you have more than one, we have no good way of distinguishing what the logger name should be. This might be outside the scope of this project though but it is an interesting issue.

After some discussion:
- Make a new assembly with the server building API call Microsoft.AspNetCore.Protocols.Server
-  Move the cross cutting concerns there (Heartbeat, ISystemClock, ConnectionManager etc)
- Introduce Protocols.Http that will depend on Protocols.Abstractions and Protocols.Server
- Kestrel.Core will depend on Protocols.Server and Protocols.Http
- Remove connection adapters 
- Introduce Protocols.Tls and have Kestrel.Https depend on that. In the implementation of the UseHttps extension method, call UseTls.

See https://github.com/aspnet/KestrelHttpServer/pull/1995

/cc @benaadams


samples/SampleApp/Startup.cs
```diff
@@ -51,6 +54,33 @@ public static void Main(string[] args)
                 basePort = 5000;
             }
 
+
+            var server = new ServerBuilder()
+                    .Listen("localhost", 8085, builder =>
```
pakrym: 👍

PR #1990, opened 2017-08-10
Stop using ConcurrentBag in tests for complex types

halter73:
dotnet/corefx#23068

### DIFF
--- a/test/shared/TestApplicationErrorLogger.cs
+++ b/test/shared/TestApplicationErrorLogger.cs
@@ -3,7 +3,6 @@
 
 using System;
 using System.Collections.Concurrent;
-using System.Threading.Tasks;
 using System.Linq;
 using Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Infrastructure;
 using Microsoft.Extensions.Logging;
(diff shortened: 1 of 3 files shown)


benaadams:
`ConcurrentQueue` is always my favorite; also has a `.ToArray` method on it

test/shared/TestApplicationErrorLogger.cs
```diff
@@ -29,7 +50,11 @@ public class TestApplicationErrorLogger : ILogger
 
         public IDisposable BeginScope<TState>(TState state)
         {
-            Scopes.Add(state);
+            lock (_lockObj)
+            {
+                _scopes.Add(state);
+            }
+
             return new Disposable(() => { });
```
davidfowl: May as well remove it here.

halter73:
@benaadams  Then we probably wouldn't need the new locks or ToArrays(). But now I'm afraid of the Concurrent types that aren't ConcurrentDictionary 😞

benaadams:
ConcurrentBag is the only one that uses thread local storage; unordered and less contended (via thread local)

halter73:
Attempt number 2.

PR #1991, opened 2017-08-11
Initial HTTP/2 support

cesarbs:
## What works

- HTTP/2 over TLS1.2 with ALPN (see `TlsStream` - temporary class that will be gone once `SslStream` supports what we need, **NOT FOR PRODUCTION USE**)
- Request and response flow
  - Headers are compressed and decompressed with HPACK
  - Request body can be read by streams (if present)
  - MVC template app with individual auth works fine
- PRIORITY frames are validated
- RST_STREAM frames are validated and abort streams
- SETTINGS frames are validated and ACKed
- PING frames are validated and ACKed
- GOAWAY frames stop connections
- WINDOW_UPDATE frames are validated
- CONTINUATION frames are sent for large header blocks

## What doesn't work yet

- Flow control in either direction
- It's not possible to encode a single header across more than one frame
  - Affects only a very large header (name and value combined ~16KB long)
  - Multiple headers that exceed a single frame's payload size are OK i.e. CONTINUATION frames are sent for those
- Request trailers
- Response trailers
- Limits and timeouts in `KestrelServerLimits` are not enforced on HTTP/2
- HPACK use is very limited on the send side
  - Literals are not Huffman-encoded
  - Common headers (e.g. "server: Kestrel") are never indexed
- Honoring client settings
- Some error checking is still missing (e.g. validating incoming frame size)

## Immediate future work

- Consolidate `Frame`/`Frame<T>` and `Http2Stream`/`Http2Stream<T>`

## Future work (no ETA)

- Use priority and dependency values sent by client to order outgoing frames
- Server push


samples/TlsApp/Startup.cs
```diff
@@ -0,0 +1,60 @@
...
+using System.IO;
+using System.Net;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Builder;
+using Microsoft.AspNetCore.Hosting;
+using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Server.Kestrel.Transport.Abstractions;
+using Microsoft.Extensions.Logging;
+
+namespace TlsApp
```
davidfowl: Any reason we need another application instead of using the same app?

src/Kestrel.Core/Internal/FrameConnection.cs
```diff
@@ -20,10 +22,16 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal
 {
     public class FrameConnection : IConnectionContext, ITimeoutControl
     {
+        private const int Http2ConnectionNotStarted= 0;
```
davidfowl: nit: spacing

src/Kestrel.Core/Internal/FrameConnection.cs
```diff
@@ -178,7 +204,14 @@ public Task StopAsync()
         {
             Debug.Assert(_frame != null, $"{nameof(_frame)} is null");
 
-            _frame.Stop();
+            if (Interlocked.Exchange(ref _http2ConnectionState, Http2ConnectionClosed) == Http2ConnectionStarted)
```
davidfowl: Should this be compare exchange?

tools/CodeGenerator/CodeGenerator.csproj
```diff
@@ -20,7 +20,7 @@
 
   <PropertyGroup>
     <StartWorkingDirectory>$(MSBuildThisFileDirectory)..\..\src\Kestrel.Core</StartWorkingDirectory>
-    <StartArguments>Internal\Http\FrameHeaders.Generated.cs Internal\Http\Frame.Generated.cs Internal\Infrastructure\HttpUtilities.Generated.cs</StartArguments>
+    <StartArguments>Internal/Http/FrameHeaders.Generated.cs Internal/Http/Frame.Generated.cs Internal/Http2/Http2Stream.Generated.cs Internal/Infrastructure/HttpUtilities.Generated.cs</StartArguments>
```
davidfowl: What's this for?

samples/TlsApp/Startup.cs
cesarbs: The cert format and the connection adapter are different. This was just for convenience while writing the code. I'll remove it before merging.

src/Kestrel.Core/Internal/FrameConnection.cs
cesarbs: No. This one should always replace the value in `_http2ConnectionState`.

Issue #1772, opened 2017-04-26
Remove OnConnectionClosed and Abort from IConnectionContext


davidfowl:
Once this is merged https://github.com/dotnet/corefxlab/pull/1491, we should be able to remove all use the pipe as a single communication mechanism for transports.

muratg:
This is not part of the public API surface, so moving this out to 2.1.0.

davidfowl:
Fixed in https://github.com/aspnet/KestrelHttpServer/commit/7854c0604a3e6e50a0bea84c4fa9d9e6ca5d7057

PR #1995, opened 2017-08-12
Initial bedrock refactoring

davidfowl:
Initial bedrock refactoring
 - Added Protocols.Abstractions
 - IConnectionHandler.OnConnection takes an IFeatureCollection instead of
 IConnectionInfo
 - Removed IConnectionContext and IConnectionInformation replaced with
 IConnectionTransportFeature
 - Updated FrameConnectionContext and FrameContext to have the relevant
 state instead of flowing the ConnectionInformation.
 - Updated tests


src/Kestrel.Core/Internal/FrameConnection.cs
```diff
@@ -44,6 +47,9 @@ public class FrameConnection : IConnectionContext, ITimeoutControl
         public FrameConnection(FrameConnectionContext context)
         {
             _context = context;
+
+            Input.OnReaderCompleted(_inputCompleted, this);
```
davidfowl: This is https://github.com/aspnet/KestrelHttpServer/issues/1772

src/Kestrel.Core/Internal/Infrastructure/KestrelEventSource.cs
```diff
@@ -26,15 +29,17 @@ private KestrelEventSource()
...
             // avoid allocating strings unless this event source is enabled
             if (IsEnabled())
             {
+                var info = context.Features.Get<IHttpConnectionFeature>();
+
                 ConnectionStart(
                     context.ConnectionId,
-                    information.LocalEndPoint.ToString(),
-                    information.RemoteEndPoint.ToString());
+                    new IPEndPoint(info.LocalIpAddress, info.LocalPort).ToString(),
```
davidfowl: Kinda gross.

src/Kestrel.Transport.Libuv/Internal/LibuvConnectionContext.Features.cs
```diff
@@ -0,0 +1,204 @@
...
+using System.Net;
+using System.Text;
+using Microsoft.AspNetCore.Http.Features;
+using Microsoft.AspNetCore.Protocols.Features;
+
+namespace Microsoft.AspNetCore.Server.Kestrel.Transport.Libuv.Internal
+{
+    // TODO: Autogenerate like we do with frame
+    public partial class LibuvConnectionContext : IFeatureCollection,
+                                                  IHttpConnectionFeature,
```
davidfowl: Feels like implementing `IHttpConnectionFeature` is kinda hacky.

src/Protocols.Abstractions/Features/IConnectionTransportFeature.cs
```diff
@@ -0,0 +1,21 @@
+﻿using System;
+using System.Collections.Generic;
+using System.IO.Pipelines;
+using System.Text;
+
+namespace Microsoft.AspNetCore.Protocols.Features
+{
+    public interface IConnectionTransportFeature
```
davidfowl: This interface needs some love. We need to represent 2 points of view:

1. Transport -> Application
2. Application -> Transport

The `ConnectionContext` is the thing the middleware sees, it's the application facing feature. This feature should be the Transport -> Application view. It should be what the application needs from the transport.

There should be another feature that gets added when OnConnection is called that does Transport -> Application. Maybe `IApplicationConnectionFeature`.

davidfowl: I also preserved the semantics for now but I think we should let the transport create the pipe but we'd have to invert the flow and have the kestrel provide the schedulers. More thought is required here.

src/Kestrel.Transport.Libuv/Internal/LibuvConnectionContext.Features.cs
benaadams: Inherit from something that already implements it? (e.g. have a default implementation parent class)

benaadams: As per HttpAbstractions 

`
```csharp
public class DefaultConnectionFeature : IHttpConnectionFeature
{
    // ...
}

public class LibuvConnectionContext : DefaultConnectionFeature, 
                                      IFeatureCollection, 
                                      IConnectionIdFeature, 
                                      IConnectionTransportFeature
{
    // ...
}
```

src/Protocols.Abstractions/Features/IConnectionTransportFeature.cs
davidfowl: Latest commit adds an IConnectionApplicationFeature

src/Kestrel.Transport.Libuv/Internal/LibuvConnectionContext.Features.cs
davidfowl: It's fine, the problem is really that this feature should have been HTTP agnostic to begin with.

src/Kestrel.Transport.Sockets/SocketConnection.Features.cs
```diff
@@ -0,0 +1,222 @@
...
+using System.Collections.Generic;
+using System.IO.Pipelines;
+using System.Net;
+using System.Text;
+using Microsoft.AspNetCore.Http.Features;
+using Microsoft.AspNetCore.Protocols.Features;
+
+namespace Microsoft.AspNetCore.Server.Kestrel.Transport.Sockets
+{
+    internal partial class SocketConnection : IFeatureCollection,
```
davidfowl: This duplication sucks. I'll make a base class.

davidfowl:
Why is this an error:

> C:\projects\kestrelhttpserver\src\Protocols.Abstractions\Protocols.Abstractions.csproj : error NU1603: System.Buffers.Primitives 0.1.0-e170811-2 depends on System.Numerics.Vectors (>= 4.4.0-preview2-25317-01) but System.Numerics.Vectors 4.4.0-preview2-25317-01 was not found. An approximate best match of System.Numerics.Vectors 4.4.0-preview2-25405-01 was resolved. [C:\projects\kestrelhttpserver\KestrelHttpServer.sln]
    0 Warning(s)
    1 Error(s)
Time Elapsed 00:00:29.90

benaadams:
Does [TreatWarningsAsErrors](https://github.com/aspnet/KestrelHttpServer/blob/d3cf0494e79807c00136f3f66fbdaefb80e231fb/build/common.props#L13) apply to restores?

davidfowl:
@benaadams I think so but I can't reproduce it locally. It might be because I have packages cached locally...

davidfowl:
Passed!

src/Protocols.Abstractions/Protocols.Abstractions.csproj
```diff
@@ -0,0 +1,30 @@
...
+    <!-- TODO: remove when https://github.com/dotnet/sdk/pull/1270 is fixed -->
+    <GenerateResourceMSBuildRuntime>CurrentRuntime</GenerateResourceMSBuildRuntime>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.AspNetCore.Http.Features" Version="$(AspNetCoreVersion)" />
+    <PackageReference Include="System.Numerics.Vectors" Version="$(CoreFxVersion)" />
+    <PackageReference Include="System.Buffers" Version="$(CoreFxVersion)" />
+    <PackageReference Include="System.IO.Pipelines" Version="$(CoreFxLabVersion)" />
+    <PackageReference Include="System.Text.Encodings.Web.Utf8" Version="$(CoreFxLabVersion)" />
```
davidfowl: This shouldn't be here

src/Kestrel.Core/Internal/FrameConnection.cs
```diff
@@ -10,15 +10,15 @@
...
 using Microsoft.AspNetCore.Hosting.Server;
 using Microsoft.AspNetCore.Http.Features;
+using Microsoft.AspNetCore.Protocols.Features;
 using Microsoft.AspNetCore.Server.Kestrel.Core.Adapter.Internal;
 using Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http;
 using Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Infrastructure;
-using Microsoft.AspNetCore.Server.Kestrel.Transport.Abstractions.Internal;
 using Microsoft.Extensions.Logging;
 
 namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal
```
benaadams: Should be `Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http` as it contains a `Frame` in that namespace which makes it specific to Http?

src/Kestrel.Core/Internal/FrameConnectionContext.cs
```diff
@@ -14,7 +14,7 @@ public class FrameConnectionContext
         public long FrameConnectionId { get; set; }
```
benaadams: Is this still a `FrameConnectionContext` and a `FrameConnectionId` as they are Http specific; is this context?

src/Kestrel.Core/Internal/Http/Frame.cs
```diff
@@ -345,11 +345,11 @@ public void Reset()
             StatusCode = StatusCodes.Status200OK;
             ReasonPhrase = null;
 
-            RemoteIpAddress = RemoteEndPoint?.Address;
-            RemotePort = RemoteEndPoint?.Port ?? 0;
+            RemoteIpAddress = HttpConnectionFeature?.RemoteIpAddress;
+            RemotePort = HttpConnectionFeature?.RemotePort ?? 0;
```
benaadams: Do these need to be separate things (Address and Port) rather than `IPEndPoint`? As its moving from a 4x property lookups to 4x interface property lookups

benaadams: Also `HttpConnectionFeature` is a feature lookup property itself; so this is even worse... need a function local cache e.g.
```csharp
var httpConnectionFeature = HttpConnectionFeature;
// ... Do the stuff 
```
Else you are evaluating `ConnectionInformation.Features.Get<IHttpConnectionFeature>()` also 4x

benaadams: Also will the `IHttpConnectionFeature` Ip:Ports change from the initial value across resets? Cache locally then set to local values on Reset?

src/Kestrel.Core/Internal/FrameConnection.cs
davidfowl: Not changing any of this as part of this PR. I plan to move all of the Internal.Http to another assembly. But baby steps 😄

src/Kestrel.Core/Internal/FrameConnectionContext.cs
davidfowl: For now. See the above comment.

src/Kestrel.Core/Internal/Http/Frame.cs
davidfowl: > Do these need to be separate things (Address and Port) rather than IPEndPoint? As its moving from a 4x property lookups to 4x interface property lookups
 
That's how it's expressed by the feature. We could have another feature but not sure if it's worth it yet.

> Also will the IHttpConnectionFeature Ip:Ports change from the initial value across resets? Cache locally then set to local values on Reset?

Good idea.

davidfowl: Done

src/Kestrel.Core/Internal/Infrastructure/KestrelEventSource.cs
```diff
@@ -26,15 +29,17 @@ private KestrelEventSource()
         // - Avoid renaming methods or parameters marked with EventAttribute. EventSource uses these to form the event object.
 
         [NonEvent]
-        public void ConnectionStart(IConnectionContext context, IConnectionInformation information)
+        public void ConnectionStart(ConnectionContext context)
         {
             // avoid allocating strings unless this event source is enabled
             if (IsEnabled())
             {
+                var info = context.Features.Get<IHttpConnectionFeature>();
```
davidfowl: We can cache the IPEndPoint on FrameConnection and pass that directly to the Frame.

src/Kestrel.Core/Internal/ConnectionHandler.cs
```diff
@@ -24,38 +28,73 @@ public ConnectionHandler(ListenOptions listenOptions, ServiceContext serviceCont
...
+            var inputPipe = transportFeature.PipeFactory.Create(GetInputPipeOptions(transportFeature.InputWriterScheduler));
+            var outputPipe = transportFeature.PipeFactory.Create(GetOutputPipeOptions(transportFeature.OutputReaderScheduler));
 
             var connectionId = CorrelationIdGenerator.GetNextId();
             var frameConnectionId = Interlocked.Increment(ref _lastFrameConnectionId);
 
+            // Set the transport and connection id
+            connectionContext.ConnectionId = connectionId;
+            transportFeature.Connection = new PipeConnection(inputPipe.Reader, outputPipe.Writer);
+            var applicationConnection = new PipeConnection(outputPipe.Reader, inputPipe.Writer);
```
halter73: Semantically, wouldn't the application want the output writer and the input reader?

src/Kestrel.Core/Internal/ConnectionHandler.cs
```diff
@@ -24,38 +28,73 @@ public ConnectionHandler(ListenOptions listenOptions, ServiceContext serviceCont
...
         }
 
-        public IConnectionContext OnConnection(IConnectionInformation connectionInfo)
+        public void OnConnection(IFeatureCollection features)
         {
-            var inputPipe = connectionInfo.PipeFactory.Create(GetInputPipeOptions(connectionInfo.InputWriterScheduler));
-            var outputPipe = connectionInfo.PipeFactory.Create(GetOutputPipeOptions(connectionInfo.OutputReaderScheduler));
+            var connectionContext = new DefaultConnectionContext(features);
+
+            var transportFeature = connectionContext.Features.Get<IConnectionTransportFeature>();
```
davidfowl: When do we ever do that? I could move `DefaultConnectionContext` to Kestrel.Core and put whatever common properties in there though.

src/Kestrel.Core/Internal/ConnectionHandler.cs
```diff
@@ -24,38 +28,73 @@ public ConnectionHandler(ListenOptions listenOptions, ServiceContext serviceCont
...
+                    frameConnectionContext.RemoteEndPoint = new IPEndPoint(connectionFeature.RemoteIpAddress, connectionFeature.RemotePort);
+                }
+            }
+
+            var connection = new FrameConnection(frameConnectionContext)
+            {
+                Connection = applicationConnection
+            };
+
+            connectionContext.Features.Set<IConnectionApplicationFeature>(connection);
```
halter73: Why doesn't this have a property on the DefaultConnectionContext?

src/Kestrel.Core/Internal/ConnectionHandler.cs
davidfowl: The naming is either wrong or right depending on how you look at it. It's either who is consuming it or what it represents. Transport is what the applications gets a handle on, and application is what the transport gets a handle on.

src/Kestrel.Core/Internal/ConnectionHandler.cs
davidfowl: See earlier comment

src/Kestrel.Core/Internal/ConnectionHandler.cs
halter73: That's the weird part to me. I would expect the *application* to get a handle on the `applicationConnection`.

src/Kestrel.Core/Internal/ConnectionHandler.cs
halter73: I though `DefaultConnectionContext.ConnectionTransportFeature` was already public. Didn't read that carefully enough. I guess there will only be one ConnectionHandler, so maybe it's not a big deal that it's ugly and hard-to-discover.

src/Kestrel.Core/Internal/ConnectionHandler.cs
davidfowl: No, the application reads from and writes to the transport thats why the application gets a handle to the transport.

The transport reads from and writes to the applications, thats why the transport gets a handle to the application.

src/Kestrel.Core/Internal/FrameConnection.cs
halter73: The other thing is that very little of this file has anything to do with HTTP. Everything to do with connection/socket lifetime, connection adapters, scopes, logging, efficient timeout management, etc. will likely be abstracted out to the connection level. But not yet I guess... @davidfowl what are you waiting for? 😝

davidfowl: I'm not making any changes outside the scope of what I actually changed. This PR is big enough already and all tests pass. There will be several more PRs that move and rename these types around.

src/Kestrel.Core/Internal/ConnectionHandler.cs
halter73: I totally understand your perspective on things. I wish I had a better way of naming it that was less ambiguous. I'll think about it.

anurse: We went through this in SignalR too. Definitely open to naming suggestions :).

src/Kestrel.Transport.Abstractions/Internal/TransportConnection.Features.cs
```diff
@@ -0,0 +1,215 @@
...
+        private static readonly Type IConnectionApplicationFeatureType = typeof(IConnectionApplicationFeature);
+
+        private object _currentIHttpConnectionFeature;
+        private object _currentIConnectionIdFeature;
+        private object _currentIConnectionTransportFeature;
+        private object _currentIConnectionApplicationFeature;
+
+        private int _featureRevision;
+
+        private List<KeyValuePair<Type, object>> MaybeExtra;
```
anurse: `MaybeExtra`? How about just `ExtraFeatures`

src/Kestrel.Transport.Abstractions/Internal/TransportConnection.cs
```diff
@@ -0,0 +1,33 @@
...
+    public abstract partial class TransportConnection
+    {
+        public TransportConnection()
+        {
+            _currentIConnectionIdFeature = this;
+            _currentIConnectionTransportFeature = this;
+            _currentIHttpConnectionFeature = this;
+        }
+
+        public IPAddress RemoteAddress { get; set; }
```
anurse: Why do this pattern instead of just using the same property names as `IHttpConnectionFeature`, etc. ?

src/Kestrel.Transport.Libuv/Internal/LibuvConnectionContext.cs
```diff
@@ -7,24 +7,17 @@
...
-        public IPEndPoint RemoteEndPoint { get; set; }
-        public IPEndPoint LocalEndPoint { get; set; }
-
-        public PipeFactory PipeFactory => ListenerContext.Thread.PipeFactory;
-        public IScheduler InputWriterScheduler => ListenerContext.Thread;
-        public IScheduler OutputReaderScheduler => ListenerContext.Thread;
+        
+        public override PipeFactory PipeFactory => ListenerContext.Thread.PipeFactory;
+        public override IScheduler InputWriterScheduler => ListenerContext.Thread;
+        public override IScheduler OutputReaderScheduler => ListenerContext.Thread;
```
anurse: Just `InputScheduler` and `OutputScheduler`?

src/Protocols.Abstractions/Features/IConnectionTransportFeature.cs
```diff
@@ -0,0 +1,18 @@
...
+    {
+        PipeFactory PipeFactory { get; }
+
+        IPipeConnection Connection { get; set; }
+
+        IScheduler InputWriterScheduler { get; }
```
anurse: See comment above on simplifying scheduler names.

src/Kestrel.Transport.Libuv/Internal/LibuvConnectionContext.cs
anurse: Or `WriterScheduler`/`ReaderScheduler` if that's more appropriate?

src/Kestrel.Transport.Abstractions/Internal/TransportConnection.Features.cs
halter73: The Maybe in MaybeExtra makes is meant to make it abundantly clear that the field can be null. I'm pretty sure it's a haskellism carried over from when @lodejard first code generated our FrameHeaders class. Personally, I like it a lot. Starting a private field name with a capital letter though...

src/Kestrel.Transport.Abstractions/Internal/TransportConnection.cs
davidfowl: Copied from what kestrel does today.

src/Kestrel.Transport.Abstractions/Internal/TransportConnection.Features.cs
anurse: I think I'd prefer something cleaner like having a static readonly empty instance in the null case, but whatever.

PR #1999, opened 2017-08-15
Add fixes for problems discovered by xunit.analyzers

natemcmaster:
The new version of xunit.analyzers found a few interesting places where we had theory parameters that were never actually used. These changes will be required before upgrading to xunit 2.3.0-beta4 (cref https://github.com/aspnet/Universe/pull/540)


test/Kestrel.FunctionalTests/RequestTests.cs
```diff
@@ -1183,9 +1183,8 @@ public void AbortingTheConnectionSendsFIN()
             }
         }
 
-        [Theory]
-        [MemberData(nameof(ConnectionAdapterData))]
-        public async Task HeadersAndStreamsAreReusedAcrossRequests(ListenOptions listenOptions)
```
natemcmaster: This one is interesting. We've been running this test twice, but listenOptions was never used to make the test inputs different. I traced this one back to a massive refactoring in our tests added by @halter73 here: https://github.com/aspnet/KestrelHttpServer/pull/1280

test/Kestrel.Core.Tests/FrameTests.cs
```diff
@@ -335,7 +335,9 @@ public void InitializeStreamsResetsStreams()
             string requestLine,
             string expectedMethod,
             string expectedRawTarget,
+#pragma warning disable xUnit1026 // Theory methods should use all of their parameters
```
halter73: Comment that it's sharing theory data with with HttpParserTests.ParsesRequestLine.

test/Kestrel.Core.Tests/HttpParserTests.cs
```diff
@@ -26,8 +26,10 @@ public class HttpParserTests
             string expectedMethod,
             string expectedRawTarget,
             string expectedRawPath,
+#pragma warning disable xUnit1026 // Theory methods should use all of their parameters
```
halter73: Comment that it's sharing theory data with FrameTests.TakeStartLineSetsFrameProperties.

Issue #1980, opened 2017-08-03
 Project Bedrock


davidfowl:
## Project Bedrock

Project bedrock is about further decoupling the components of Kestrel so that we can use it as the foundation for our non-http networking stack. 
We want to build on the primitives, patterns and cross cutting concerns that exist today in ASP.NET Core applications. The goal is to enable higher level frameworks (like SignalR or WCF and even ASP.NET Core itself) to build on top of abstractions that don't tie them to a specific connection implementation (OWIN for Connections). As an example, it allows SignalR to run both on top of TCP or Websockets without having to understand what the underlying transport is. We also want to enable building raw low level protocol servers to handle things like MQTT for IOT scenarios.

There are 3 main actors in this server side programming model:
- **Applications/Middleware/Frameworks** - The application code that handles connections and implement protocol parsing logic or other logic that modifies the stream of data (http, TLS as an example)
- **Transports** - Transports provide an implementation of an `IFeatureCollection` that implements the underlying connection semantics. In short, transports provide a concrete implementation of the `ConnectionContext` that flows through the dispatcher to the application. 
- **Dispatchers** - The dispatcher is the component that brings the transport layer and application layer together. It's job is to manage the lifetime of the transport connection and application running on top. The dispatcher will expose the `ISocketBuilder` for a particular binding relevant to the transport. For example, the http dispatcher will expose the `ISocketBuilder` based on a particular route, while the TCP dispatcher will expose an `ISocketBuilder` based on an ip address and port.

## Applications/Middleware/Frameworks

At the center of this work is a new set of primitives that represent an underlying connection:

```C#
public abstract class ConnectionContext
{
    public abstract string ConnectionId { get; set; }

    public abstract IFeatureCollection Features { get; }
    
    public abstract IPipe Transport { get; set; }
}
```

```C#
public interface IConnectionIdFeature
{
    string ConnectionId { get; set; }
}
```

```C#
public interface IConnectionTransportFeature
{
    public abstract PipeFactory PipeFactory { get; set; }

    public abstract IPipe Transport { get; set; }
}
```

The `ConnectionContext` is the "HttpContext" of the connection universe. It's an abstraction that represents a persistent connection of some form. This could
be a TCP connection, a websocket connection or something more hybrid (like a connection implemented over a non duplex protocol like server sent events + http posts). The feature collection
is there for the same reason it's there on the `HttpContext`, the server or various pieces of "middleware" can add, augment or remove features
[...]

Issue #1654, opened 2017-04-11
Delay config of listen options until server starts


JunTaoLuo:
Currently ListenOptions are configured immediately when Listen is called. We should delay this until we have determined to use the ListenOptions or IServerAddressesFeature since we may be throwing away the configured endpoints in https://github.com/aspnet/KestrelHttpServer/blob/99278c469161370b1015c67d18ae9388591a49af/src/Microsoft.AspNetCore.Server.Kestrel.Core/KestrelServer.cs#L119. This could potentially help when configuring HTTPS for Kestrel, see https://github.com/aspnet/KestrelHttpServer/issues/1290. 

cc @glennc @danroth27 @muratg

Tratcher:
Scenario: Kestrel is configured for https with a cert file for development, but then the app is published to IIS and that cert is not needed/appropriate. Even running the app in IIS Express would fail today if the self-host cert was not present, which is unnecessary.

Tratcher:
Fixed by https://github.com/aspnet/KestrelHttpServer/pull/2003/files#diff-319ef184f2f31b202bc2f4566a710afe ?

davidfowl:
Yes

PR #2003, opened 2017-08-18
Added initial connection middleware pipeline

davidfowl:
- Implemented IConnectionBuilder on ListenOptions. Kept IConnectionAdapter for now. 
- Delay the configure callback for ListenOptions until the server has started. 
- Added ConnectionLimitMiddleware and HttpConnectionMiddleware 
- Expose ConnectionAborted and ConnectionClosed on ConnectionContext and IConnectionTransportFeature 
- Updated the tests

Progress on #1980 
Progress on #1772


benchmarks/Kestrel.Performance/FrameWritingBenchmark.cs
```diff
@@ -109,8 +109,8 @@ private TestFrame<object> MakeFrame(PipeFactory pipeFactory)
             {
                 ServiceContext = serviceContext,
                 PipeFactory = pipeFactory,
-                Input = input.Reader,
-                Output = _outputPipe 
+                Application = pair.Application,
+                Transport = pair.Transport
```
halter73: I think you want to use `_outputPipe` here so the Cleanup method works.

src/Protocols.Abstractions/Features/IConnectionTransportFeature.cs
```diff
@@ -2,17 +2,24 @@
...
+
+        IPipeConnection Application { get; set; }
 
         IScheduler InputWriterScheduler { get; }
 
         IScheduler OutputReaderScheduler { get; }
+
+        Task ConnectionAborted { get; }
+
+        Task ConnectionClosed { get; }
```
halter73: Why can't a single Task represent aborted and closed?

Didn't we intend to communicate when/how the connection was closed via the IPipeConnections?

src/Protocols.Abstractions/PipeFactoryExtensions.cs
```diff
@@ -0,0 +1,21 @@
...
+            return pipeFactory.CreateConnectionPair(new PipeOptions(), new PipeOptions());
+        }
+
+        public static (IPipeConnection Transport, IPipeConnection Application) CreateConnectionPair(this PipeFactory pipeFactory, PipeOptions inputOptions, PipeOptions outputOptions)
+        {
+            var input = pipeFactory.Create(inputOptions);
+            var output = pipeFactory.Create(outputOptions);
+
+            var transportToApplication = new PipeConnection(output.Reader, input.Writer);
+            var applicationToTransport = new PipeConnection(input.Reader, output.Writer);
```
halter73: I like the TransportToApplication and ApplicationToTransport naming. Can we use this everywhere instead of of just Transport and Application?

src/Kestrel.Core/Internal/ConnectionHandler.cs
```diff
@@ -34,89 +28,45 @@ public void OnConnection(IFeatureCollection features)
 
             var transportFeature = connectionContext.Features.Get<IConnectionTransportFeature>();
 
-            var inputPipe = transportFeature.PipeFactory.Create(GetInputPipeOptions(transportFeature.InputWriterScheduler));
-            var outputPipe = transportFeature.PipeFactory.Create(GetOutputPipeOptions(transportFeature.OutputReaderScheduler));
+            // REVIEW: Unfortunately, we still need to use the service context to create the pipes since the settings
+            // for the scheduler and limits are specified here
```
halter73: Could you just pass in the KestrelServerOptions? Do you need to pass in a scheduler and the "ThreadPool"?

src/Kestrel.Core/Internal/ConnectionHandler.cs
```diff
@@ -34,89 +28,45 @@ public void OnConnection(IFeatureCollection features)
...
-            connectionContext.Features.Set<IConnectionApplicationFeature>(connection);
+            // This *must* be set before returning from OnConnection
+            transportFeature.Application = pair.Application;
 
-            // Since data cannot be added to the inputPipe by the transport until OnConnection returns,
-            // Frame.ProcessRequestsAsync is guaranteed to unblock the transport thread before calling
-            // application code.
-            connection.StartRequestProcessing(_application);
+            // REVIEW: This task should be tracked by the server for graceful shutdown
+            // Today it's handled specifically for http but not for aribitrary middleware
```
halter73: Can we at least log a critical error if something goes wrong? I know that shouldn't/won't hapeen today, but this would also help avoid potential unobserved exceptions in the future.

src/Kestrel.Core/Internal/ConnectionHandler.cs
```diff
@@ -34,89 +28,45 @@ public void OnConnection(IFeatureCollection features)
...
+            _ = _connectionDelegate(connectionContext);
         }
 
         // Internal for testing
-        internal PipeOptions GetInputPipeOptions(IScheduler writerScheduler) => new PipeOptions
+        internal static PipeOptions GetInputPipeOptions(ServiceContext serviceContext, IScheduler writerScheduler) => new PipeOptions
         {
-            ReaderScheduler = _serviceContext.ThreadPool,
+            ReaderScheduler = serviceContext.ThreadPool,
```
halter73: Maybe we should pass the reader scheduler explicitly like the writer scheduler.

src/Kestrel.Core/Internal/ConnectionLimitMiddleware.cs
```diff
@@ -0,0 +1,47 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Protocols;
+using Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Infrastructure;
+
+namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal
+{
+    public class ConnectionLimitMiddleware
```
halter73: 👍

src/Kestrel.Core/Internal/Http/OutputProducer.cs
```diff
@@ -103,8 +106,8 @@ public void Abort(Exception error)
                 _log.ConnectionDisconnect(_connectionId);
                 _completed = true;
 
-                _pipe.Reader.CancelPendingRead();
-                _pipe.Writer.Complete(error);
+                _outputPipeReader.CancelPendingRead();
```
halter73: I would love if we could find a way of getting rid of the IPipeReader dependency in this class. Not that this needs to be done as part of this PR, but something to think about.

src/Kestrel.Core/Internal/HttpConnectionMiddleware.cs
```diff
@@ -0,0 +1,111 @@
...
+
+            // The order here is important, start request processing so that
+            // the frame is created before this yields. Events need to be wired up
+            // afterwards
+            var processingTask = connection.StartRequestProcessing(_application);
+
+            // Wire up the events an forward calls to the frame connection
+            // It's important that these execute synchronously because graceful
+            // connection close is order sensative (for now)
+            connectionContext.ConnectionAborted.ContinueWith((task, state) =>
```
halter73: ContinueWith! I think I remember you claiming there's never a good excuse for this 😉

src/Kestrel.Core/KestrelServer.cs
```diff
@@ -135,7 +135,16 @@ private static ServiceContext CreateServiceContext(IOptions<KestrelServerOptions
 
                 async Task OnBind(ListenOptions endpoint)
                 {
-                    var connectionHandler = new ConnectionHandler<TContext>(endpoint, ServiceContext, application);
+                    // Add the connection limit middleware
+                    endpoint.UseConnectionLimit(ServiceContext);
+
+                    // Configure the user delegate
+                    endpoint?.Configure(endpoint);
```
halter73: Why `?.`?

src/Kestrel.Core/ListenOptions.cs
```diff
@@ -126,6 +129,10 @@ public FileHandleType HandleType
         /// </remarks>
         public List<IConnectionAdapter> ConnectionAdapters { get; } = new List<IConnectionAdapter>();
 
+        public IServiceProvider ApplicationServices => KestrelServerOptions?.ApplicationServices;
```
halter73: Not again! Where is this used?

src/Kestrel.Core/KestrelServerOptions.cs
```diff
@@ -100,8 +100,7 @@ public void Listen(IPEndPoint endPoint, Action<ListenOptions> configure)
                 throw new ArgumentNullException(nameof(configure));
             }
 
-            var listenOptions = new ListenOptions(endPoint) { KestrelServerOptions = this };
-            configure(listenOptions);
+            var listenOptions = new ListenOptions(endPoint) { KestrelServerOptions = this, Configure = configure };
```
halter73: Instead of adding this weird Configure callback, why not keep it the way it was and invoke configure directly in Listen()? Is this just for testing? Seems super confusing to me.

src/Kestrel.Core/Internal/HttpConnectionMiddleware.cs
```diff
@@ -0,0 +1,111 @@
...
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Hosting.Server;
+using Microsoft.AspNetCore.Http.Features;
+using Microsoft.AspNetCore.Protocols;
+using Microsoft.AspNetCore.Protocols.Features;
+using Microsoft.AspNetCore.Server.Kestrel.Core.Adapter.Internal;
+
+namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal
+{
+    public class HttpConnectionMiddleware<TContext>
```
halter73: 👍 I love that you were able to factor this out of ConnectionHandler.

src/Kestrel.Core/Internal/HttpConnectionMiddleware.cs
```diff
@@ -0,0 +1,111 @@
...
+                // Unwrap the aggregate exception
+                ((FrameConnection)state).OnConnectionClosed(task.Exception?.InnerException);
+            },
+            connection, TaskContinuationOptions.ExecuteSynchronously);
+
+            return processingTask;
+        }
+    }
+
+    public static class HttpConnectionBuilderExtensions
```
halter73: Move to a new file.

src/Kestrel.Core/ListenOptions.cs
```diff
@@ -14,9 +16,10 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core
     /// Describes either an <see cref="IPEndPoint"/>, Unix domain socket path, or a file descriptor for an already open
     /// socket that Kestrel should bind to or open.
     /// </summary>
-    public class ListenOptions : IEndPointInformation
+    public class ListenOptions : IEndPointInformation, IConnectionBuilder
     {
         private FileHandleType _handleType;
+        private readonly List<Func<ConnectionDelegate, ConnectionDelegate>> _components = new List<Func<ConnectionDelegate, ConnectionDelegate>>();
```
halter73: 🎉

src/Kestrel.Core/Internal/ConnectionLimitMiddleware.cs
```diff
@@ -0,0 +1,47 @@
...
+                connection.Transport.Input.Complete();
+                connection.Transport.Output.Complete();
+                return Task.CompletedTask;
+            }
+
+            return _next(connection);
+        }
+    }
+
+    public static class ConnectionLimitBuilderExtensions
```
halter73: Move to new file.

src/Kestrel.Core/KestrelServer.cs
davidfowl: I guess the delegate is never null.

src/Kestrel.Core/ListenOptions.cs
davidfowl: `ApplicationServices` is part of the interface. It will be used to activate middleware, just like `IApplicationBuilder. ApplicationServices`

src/Kestrel.Core/KestrelServerOptions.cs
davidfowl: I need to run this later so that I can wrap the connection limit around the entire pipeline.

benchmarks/Kestrel.Performance/FrameWritingBenchmark.cs
davidfowl: Good catch.

src/Protocols.Abstractions/Features/IConnectionTransportFeature.cs
davidfowl: It's broken because of the dispatching. Ordering problems cause issues that I don't want to fix while changing everything else.

src/Protocols.Abstractions/PipeFactoryExtensions.cs
davidfowl: No, it's too verbose. Also the only thing that usually sees both sides are when you're implementing the transport. Notice the `DefaultConnectionContext` still has the `Transport` exposed.

src/Kestrel.Core/Internal/ConnectionHandler.cs
davidfowl: I don't see the point in doing that as part of this change.

src/Kestrel.Core/Internal/ConnectionHandler.cs
davidfowl: Sure, eventually.

src/Kestrel.Core/Internal/Http/OutputProducer.cs
davidfowl: Agreed.

src/Kestrel.Core/KestrelServer.cs
davidfowl: Well unit tests manually new up these things. I'll set the delegate in the ctor.

src/Kestrel.Core/Internal/ConnectionHandler.cs
davidfowl: Will do.

src/Kestrel.Core/Internal/HttpConnectionMiddleware.cs
davidfowl: There isn't, it's always worse 😄

davidfowl:
🆙 📅

src/Kestrel.Core/KestrelServerOptions.cs
Tratcher: https://github.com/aspnet/KestrelHttpServer/issues/1654

PR #2006, opened 2017-08-18
React to changes in BenchmarkDotNet 0.10.9

natemcmaster:
setup/cleanup are obsolete in favor of globalsetup/globalcleanup and iterationsetup/cleanup https://github.com/dotnet/BenchmarkDotNet/issues/325

cc @halter73

### DIFF
--- a/benchmarks/Kestrel.Performance/Kestrel.Performance.csproj
+++ b/benchmarks/Kestrel.Performance/Kestrel.Performance.csproj
@@ -23,7 +23,7 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="BenchmarkDotNet" />
+    <PackageReference Include="BenchmarkDotNet" Version="0.10.9" NoWarn="KRB4002" />
   </ItemGroup>
 
 </Project>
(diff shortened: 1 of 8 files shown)


natemcmaster:
FYI - CI is expected to fail until the lineup from https://github.com/aspnet/Universe/pull/544 is available

benchmarks/Kestrel.Performance/Kestrel.Performance.csproj
```diff
@@ -23,7 +23,7 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="BenchmarkDotNet" />
+    <PackageReference Include="BenchmarkDotNet" Version="0.10.9" NoWarn="KRB4002" />
```
ryanbrandenburg: I know we don't do it elsewhere, but maybe we should start adding a short note about what the warn-codes we're suppressing are?

natemcmaster: This is a workaround for the lineup not being available until the build passes. I can remove it and Kestrel will just be busted for the next 20 minutes

Issue #2008, opened 2017-08-21
Frame.Reset allocates when TLS Feature is present


benaadams:
`foreach`ing the `ConnectionFeatures`

1.1GB for 17.6M requests (15 secs)

![1.1GB](https://aoa.blob.core.windows.net/aspnet/tls.png)

Multibox!

/cc @davidfowl @halter73

PR #2009, opened 2017-08-21
Don't alloc IEnumerable in Reset for ITlsConnectionFeature

benaadams:
Fixes https://github.com/aspnet/KestrelHttpServer/issues/2008

/cc @davidfowl @halter73

### DIFF
--- a/src/Kestrel.Core/Internal/Http/Frame.cs
+++ b/src/Kestrel.Core/Internal/Http/Frame.cs
@@ -367,6 +367,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
                 var tlsFeature = ConnectionFeatures?[typeof(ITlsConnectionFeature)];
                 _scheme = tlsFeature != null ? "https" : "http";
             }
+
             Scheme = _scheme;
 
             _manuallySetRequestAbortToken = null;
--- a/src/Kestrel.Core/Internal/Http/Frame.cs
+++ b/src/Kestrel.Core/Internal/Http/Frame.cs
@@ -83,6 +83,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
 
         private HttpRequestTarget _requestTargetForm = HttpRequestTarget.Unknown;
         private Uri _absoluteRequestTarget;
+        private string _scheme = null;
 
         public Frame(FrameContext frameContext)
         {
@@ -338,7 +339,6 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
             MaxRequestBodySize = ServerOptions.Limits.MaxRequestBodySize;
             AllowSynchronousIO = ServerOptions.AllowSynchronousIO;
             TraceIdentifier = null;
-            Scheme = null;
             Method = null;
             PathBase = null;
             Path = null;
@@ -362,13 +362,12 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
             RequestHeaders = FrameRequestHeaders;
             ResponseHeaders = FrameResponseHeaders;
 
-            var tlsFeature = ConnectionFeatures?[typeof(ITlsConnectionFeature)];
-            if (tlsFeature != null)
+            if (_scheme == null)
             {
-                // Set the scheme to https if there's an ITlsConnectionFeature
-                Scheme = "https";
-                FastFeatureSet(typeof(ITlsConnectionFeature), tlsFeature);
+                var tlsFeature = ConnectionFeatures?[typeof(ITlsConnectionFeature)];
+                _scheme = tlsFeature != null ? "https" : "http";
             }
+            Scheme = _scheme;
 
             _manuallySetRequestAbortToken = null;
             _abortedCts = null;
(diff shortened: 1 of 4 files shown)


davidfowl:
My fear is that this won't scale. We want to flow features from the connection to the request.

davidfowl:
Actually it might be possible to just delegate instead of copy per request.

benaadams:
Not sure what you mean? It it because there will be more than one `ITlsConnectionFeature` in the `ConnectionFeatures` and the lookup will return first?

So you need to enumerate every feature to pickup the last?

davidfowl:
@benaadams features from the connection's feature collection should be accessible from the request's feature collection.

benaadams:
I don't understand why enumerating would "scale" better than a look up though?

benaadams:
Is it the stacking of FeatureCollections and you want to get to the bottom one?

davidfowl:
> I don't understand why enumerating would "scale" better than a look up though?

https://github.com/aspnet/KestrelHttpServer/pull/2009#issuecomment-323650306

That's why I said it would probably be fine to just delegate instead of copy.

benaadams:
You just linked to earlier comment that suggested an alternative implementation; but didn't provide a more information on the issue.

This change is doing the same thing as the enumerator; i.e. using the `ConnectionFeatures` to look up and set the `ITlsConnectionFeature`.

Where I could see the behavior differing is when the `ConnectionFeatures` is multiple stacked `FeatureCollection`s; so the enumerator will keep going until it gets to the bottom `FeatureCollection` that has a `ITlsConnectionFeature` (and keep going until its enumerated everything) whereas the lookup will use the `ITlsConnectionFeature` on the top `FeatureCollection`.

So is that the issue?

davidfowl:
My initial reaction was that this was a problem because it was no longer copying features from the connection to the request feature collection. That's what I meant by "it won't scale". I realized that it would probably be better to delegate to the ConnectionFeatures itself rather than copy them to the request features on a per request basis.

TL;DR: This change is fine 😄

benaadams:
Ahh... By delegate, just set to null and null ?? coalese in getter?

davidfowl:
> Ahh... By delegate, just set to null and null ?? coalese in getter?

I meant fallback to ConnectionFeatures when asking for request features if not found in the current feature collection.

Frame.Features.cs

```C#
IFeatureCollection.this[Type key]
{
   get => return GetFeature(key) ?? ConnectionFeatures[key];
}
```

benaadams:
Then can just null it an pay cost on lookup rather than for every Reset. Makes sense.

benaadams:
Its an expensive way to set scheme for https though; will have a play

davidfowl:
> Its an expensive way to set scheme for https though; will have a play

I think special casing the scheme like this is fine.

benaadams:
I think I made it unhappy

src/Kestrel.Core/Internal/Http2/Http2Stream.FeatureCollection.cs
```diff
@@ -246,13 +246,13 @@ MinDataRate IHttpMinResponseDataRateFeature.MinDataRate
 
         object IFeatureCollection.this[Type key]
         {
-            get => FastFeatureGet(key);
```
Tratcher: Beware of exposing the underlying features directly. Any mutable fields should be reset per request.

benaadams: They are reset further up by `ResetFeatureCollection();` in `Reset`

Tratcher: That resets the presence of a feature, not the values on that feature.

benaadams: Ah, like if you change `ClientCertificate` on `ITlsConnectionFeature`

halter73: I'm not so sure connection-level features should be reset be request. Then again, I'm not sure why you would want to mutate a connection-level feature via the request's feature collection. If there was a way to make these features immutable when accessed this way, we probably would.

src/Kestrel.Core/Internal/Http/Frame.cs
```diff
@@ -362,19 +362,12 @@ public void Reset()
...
-                    if (feature.Key == typeof(ITlsConnectionFeature))
-                    {
-                        Scheme = "https";
-                    }
-
-                    FastFeatureSet(feature.Key, feature.Value);
-                }
+                var tlsFeature = ConnectionFeatures?[typeof(ITlsConnectionFeature)];
+                _scheme = tlsFeature != null ? "https" : "http";
             }
```
halter73: Nit: newline.

benaadams: Added

halter73:
Thanks!

src/Kestrel.Core/Internal/Http2/Http2Stream.FeatureCollection.cs
Tratcher: Will remote IP be a connection level feature? We mutate that all the time using x-forwarded-for. It should be reset per request as each request may have different forwarders.

So far I think you're handling the IPs correctly, copying them when you create the frame, but I'm less sure about the other features and their mutable fields.

Tratcher: Yes, client cert is a good example.

PR #2010, opened 2017-08-21
Improve ConnectionLimitMiddleware and connection pipeline building

halter73:
We will probably have to go back to using ConnectionManager in the
ConnectionLimitMiddleware, but this shows that ListenOptions.Configure
is unnecessary.


src/Kestrel.Core/KestrelServer.cs
```diff
@@ -135,15 +135,17 @@ private static ServiceContext CreateServiceContext(IOptions<KestrelServerOptions
...
                     // Add the HTTP middleware as the terminal connection middleware
                     endpoint.UseHttpServer(endpoint.ConnectionAdapters, ServiceContext, application);
 
+                    var connectionDelegate = endpoint.Build();
+
+                    // Add the connection limit middleware
+                    if (Options.Limits.MaxConcurrentConnections.HasValue)
+                    {
+                        connectionDelegate = new ConnectionLimitMiddleware(connectionDelegate, Trace, Options.Limits.MaxConcurrentConnections.Value).OnConnectionAsync;
+                    }
```
halter73: @davidfowl The rest of this PR will likely need to be undone because upgraded connections need to be removed from the normal connection count, but the change here to remove the call to `endpoint.Configure` I think is solid.

halter73:
@pakrym @davidfowl updated.

src/Kestrel.Core/Internal/Http/Frame.FeatureCollection.cs
```diff
@@ -294,7 +294,8 @@ void IHttpResponseFeature.OnCompleted(Func<object, Task> callback, object state)
 
             _wasUpgraded = true;
 
-            ServiceContext.ConnectionManager.NormalConnectionCount.ReleaseOne();
+            var connectionCountFeature = ConnectionFeatures[typeof(IDecrementConcurrentConnectionCountFeature)] as IDecrementConcurrentConnectionCountFeature;
```
davidfowl: Can you use the generic?

davidfowl:
Fixes #2031

src/Kestrel.Core/Internal/ConnectionLimitMiddleware.cs
```diff
@@ -1,32 +1,74 @@
...
                 connection.Transport.Output.Complete();
-                return Task.CompletedTask;
+                return;
+            }
+
+            var releasor = new ConnectionReleasor(_concurrentConnectionCounter);
+
+            try
+            {
+                connection.Features[typeof(IDecrementConcurrentConnectionCountFeature)] = releasor;
```
davidfowl: Use the generic set?

src/Kestrel.Core/Internal/ConnectionLimitMiddleware.cs
```diff
@@ -1,32 +1,74 @@
...
+            private bool _connectionReleased;
+
+            public ConnectionReleasor(ResourceCounter normalConnectionCounter)
+            {
+                _concurrentConnectionCounter = normalConnectionCounter;
+            }
+
+            public void ReleaseConnection()
+            {
+                if (!_connectionReleased)
```
davidfowl: Does this need to be thread safe?

halter73: I almost commented on the interface that this isn't thread safe, but I figure that was expected. Most connection/request features like the header dictionaries aren't thread safe.

src/Kestrel.Core/Internal/ConnectionLimitMiddleware.cs
```diff
@@ -1,32 +1,74 @@
...
 
-        public Task OnConnectionAsync(ConnectionContext connection)
+        public async Task OnConnectionAsync(ConnectionContext connection)
         {
-            if (!_serviceContext.ConnectionManager.NormalConnectionCount.TryLockOne())
+            if (!_concurrentConnectionCounter.TryLockOne())
             {
                 KestrelEventSource.Log.ConnectionRejected(connection.ConnectionId);
-                _serviceContext.Log.ConnectionRejected(connection.ConnectionId);
+                _trace?.ConnectionRejected(connection.ConnectionId);
```
davidfowl: trace can be null?

halter73: No unless someone constructs it manually. It could have been before when there was the IConnectionBuilder extension method.

PR #2012, opened 2017-08-22
Faster IFeatureCollection.Get<TFeature>

benaadams:
Speed up the generic `IFeatureCollection` `Get<T>`/`Set<T>`

Pre
```
                 Method |     Mean |         Op/s |
----------------------- |---------:|-------------:|
     GetViaTypeOf_First | 12.96 ns | 77,136,481.9 | 
    GetViaGeneric_First | 22.38 ns | 44,679,211.4 | IHttpRequestFeature
    
      GetViaTypeOf_Last | 69.91 ns | 14,304,944.1 |
     GetViaGeneric_Last | 80.70 ns | 12,392,225.5 | IHttpSendFileFeature
     
    GetViaTypeOf_Custom | 76.84 ns | 13,014,603.2 |
   GetViaGeneric_Custom | 84.15 ns | 11,882,891.0 |
   
  GetViaTypeOf_NotFound | 76.29 ns | 13,108,494.2 |
 GetViaGeneric_NotFound | 84.44 ns | 11,843,124.1 |
```

Post
```
                 Method |     Mean |         Op/s |
----------------------- |---------:|-------------:|
     GetViaTypeOf_First | 12.70 ns | 78,733,231.9 |
    GetViaGeneric_First | 19.96 ns | 50,103,862.6 | IHttpRequestFeature
    
      GetViaTypeOf_Last | 66.28 ns | 15,087,099.9 |
     GetViaGeneric_Last | 33.59 ns | 29,771,919.4 | IHttpSendFileFeature
     
    GetViaTypeOf_Custom | 75.68 ns | 13,214,141.0 |
   GetViaGeneric_Custom | 44.21 ns | 22,617,022.7 |
   
  GetViaTypeOf_NotFound | 75.45 ns | 13,254,289.3 |
 GetViaGeneric_NotFound | 44.53 ns | 22,454,601.1 |
```

### DIFF
--- a/benchmarks/Kestrel.Performance/HttpProtocolFeatureCollection.cs
+++ b/benchmarks/Kestrel.Performance/HttpProtocolFeatureCollection.cs
@@ -1,8 +1,10 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
-using System;
 using System.IO.Pipelines;
+using System.Runtime.CompilerServices;
+using System.Threading;
+using System.Threading.Tasks;
 using BenchmarkDotNet.Attributes;
 using Microsoft.AspNetCore.Http.Features;
 using Microsoft.AspNetCore.Server.Kestrel.Core;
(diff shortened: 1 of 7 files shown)


src/Kestrel.Core/Internal/Http/Frame.FeatureCollection.cs
```diff
@@ -257,7 +257,13 @@ MinDataRate IHttpMinResponseDataRateFeature.MinDataRate
...
         TFeature IFeatureCollection.Get<TFeature>()
         {
-            return (TFeature)(FastFeatureGet(typeof(TFeature)) ?? ConnectionFeatures?[typeof(TFeature)]);
+            var feature = FastFeatureGet<TFeature>();
+            if (feature != null)
+            {
+                return feature;
+            }
+
+            return (TFeature)ConnectionFeatures?[typeof(TFeature)];
```
benaadams: ?? gets confused and doesn't complile, says can't convert TFeature to TFeature

benaadams:
Shame about the `Interface<Generic>` cost, because the generic function itself is pretty good

benaadams:
Updated, rebased

benaadams:
Using two arrays of struct wrappers for `Type` and `object` with a for loop (for high density on Type matching and to switch covariance off on the array) rather than a branching if/else is slower than the `typeof` version, so way slower than the generic version for `_Last`.

davidfowl:
This PR needs to do over. We refactored everything for http2. This change is interesting though, we should look at it again.

benaadams:
Rebased https://github.com/aspnet/KestrelHttpServer/pull/2290

PR #2013, opened 2017-08-22
Create the connection logging scope in ConnectionHandler

davidfowl:
- Instead of doing it on the FrameConnection only. This will
make sure all middleware logs get the connection id as part of their scope.

### DIFF
--- a/src/Kestrel.Core/Internal/ConnectionHandler.cs
+++ b/src/Kestrel.Core/Internal/ConnectionHandler.cs
@@ -24,6 +24,8 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal
             _connectionDelegate = connectionDelegate;
         }
 
+        private IKestrelTrace Log => _serviceContext.Log;
+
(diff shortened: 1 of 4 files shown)


src/Kestrel.Core/Internal/ConnectionHandler.cs
```diff
@@ -51,14 +53,31 @@ public void OnConnection(IFeatureCollection features)
...
+                try
+                {
+                    await _connectionDelegate(connectionContext);
+                }
+                catch (Exception ex)
+                {
+                    Log.LogCritical(0, ex, $"{nameof(ConnectionHandler)}.{nameof(Execute)}() {connectionContext.ConnectionId}");
+                }
+
+                Log.ConnectionStop(connectionContext.ConnectionId);
```
halter73: Would this be better in a finally?

davidfowl: Will move it in the next change.

PR #2014, opened 2017-08-23
Less StringValue struct copies for header existence checks

benaadams:
Less StringValue struct copies for header existence checks
- check struct in-place

Lazy calc transfer coding
- calc once existence is confirmed rather than always (since existence is checked anyway)


benaadams:
Pre

pre

Post

post

davidfowl:
This PR needs to do over. We refactored everything for http2. This looks like and easy win though.

Issue #1955, opened 2017-07-12
Flaky test: ConnectionClosedWhenResponseDoesNotSatisfyMinimumDataRate


natemcmaster:
Saw this test failure on Travis today. Appears to have failed this assertion:
```c#
Assert.True(aborted.Wait(TimeSpan.FromSeconds(60)));
```

```
Failed   ConnectionClosedWhenResponseDoesNotSatisfyMinimumDataRate
Error Message:
 Assert.True() Failure
Expected: True
Actual:   False
Stack Trace:
   at Microsoft.AspNetCore.Server.Kestrel.FunctionalTests.ResponseTests.<ConnectionClosedWhenResponseDoesNotSatisfyMinimumDataRate>d__64.MoveNext() in /home/travis/build/aspnet/KestrelHttpServer/test/Kestrel.FunctionalTests/ResponseTests.cs:line 2499
```

https://travis-ci.org/aspnet/KestrelHttpServer/jobs/252952635#L994-L1000

cc @cesarbs

mikeharder:
Caused 2 failures of the Ubuntu mirror:

* http://aspnetci/viewLog.html?buildId=290526&tab=buildResultsDiv&buildTypeId=CoreCLR_Linux
* http://aspnetci/viewLog.html?buildId=290537&tab=buildResultsDiv&buildTypeId=CoreCLR_Linux

We need to prioritize fixing this test since it's causing slowdowns and manual work in build steps like the mirror.

muratg:
@cesarbs please take a look.

davidfowl:
I'm seeing this fail again and I can't tell if it's because of my changes or if it's because this test is generally flaky.

natemcmaster:
I think this test is still flaky. Just failed again.

http://aspnetci/viewLog.html?buildId=339030&tab=buildResultsDiv&buildTypeId=Lite_KestrelStressTest

```
The request was not aborted within the timeout period.
Expected: True
Actual:   False
   at Microsoft.AspNetCore.Server.Kestrel.FunctionalTests.ResponseTests.ConnectionClosedWhenResponseDoesNotSatisfyMinimumDataRate() in C:\b\w\33bdfc1cae7b2a38\.r\KestrelHttpServer\test\Kestrel.FunctionalTests\ResponseTests.cs:line 2496
```

ryanbrandenburg:
[...]

PR #2015, opened 2017-08-24
Fix flakiness in ConnectionClosedWhenResponseDoesNotSatisfyMinimumDataRate (#1955)

cesarbs:
#1955

Switched to a raw socket to avoid the extra layer of buffering in the `StreamReader` used by `TestConnection`. This seems to stabilize the test.

### DIFF
--- a/test/Kestrel.FunctionalTests/ResponseTests.cs
+++ b/test/Kestrel.FunctionalTests/ResponseTests.cs
@@ -2493,14 +2493,22 @@ namespace Microsoft.AspNetCore.Server.Kestrel.FunctionalTests
                     Assert.True(messageLogged.Wait(TimeSpan.FromSeconds(60)));
                     Assert.True(requestAborted.Wait(TimeSpan.FromSeconds(60)));
 
-                    var buffer = new byte[chunkSize];
                     var totalReceived = 0;
-                    var received = socket.Receive(buffer);
+                    var received = 0;
 
-                    while (received > 0)
+                    try
                     {
-                        totalReceived += received;
-                        received = socket.Receive(buffer);
+                        var buffer = new byte[chunkSize];
+
+                        do
+                        {
+                            received = socket.Receive(buffer);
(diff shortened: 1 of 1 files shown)


test/Kestrel.FunctionalTests/ResponseTests.cs
```diff
@@ -2471,46 +2471,40 @@ public Task ResponseStatusCodeSetBeforeHttpContextDisposedRequestMalformedReadIg
...
-                    await Assert.ThrowsAsync<EqualException>(async () => await connection.ReceiveForcedEnd(
-                        new string('a', chunks * chunkSize)));
+                    var buffer = new byte[chunkSize];
+                    var totalReceived = 0;
+                    var received = socket.Receive(buffer);
+
+                    while (received > 0)
+                    {
+                        totalReceived += received;
+                        received = socket.Receive(buffer);
```
halter73: Maybe put this in a try/catch and break out of the loop in the case Receive throws an IOException.

test/Kestrel.FunctionalTests/ResponseTests.cs
```diff
@@ -2471,46 +2471,48 @@ public Task ResponseStatusCodeSetBeforeHttpContextDisposedRequestMalformedReadIg
...
 
-                    await connection.Receive(
-                        "HTTP/1.1 200 OK",
-                        "");
-                    await connection.ReceiveStartsWith("Date: ");
-                    await connection.Receive(
-                        $"Content-Length: {chunks * chunkSize}",
-                        "",
-                        "");
+                    Assert.True(messageLogged.Wait(TimeSpan.FromSeconds(60)));
```
natemcmaster: Consider adding a messages when using Assert.True/False so when the assertion fails, we know which one. eg
```c#
Assert.True(messageLogged.Wait(TimeSpan.FromSeconds(60)), "The message should have been received but timed out");
```

PR #2018, opened 2017-08-25
Only await when needed

benaadams:
Full on silly...


benaadams:
Closing as is distracting

Issue #1909, opened 2017-06-21
[Patch 1.1.3] Error -16 EBUSY resource busy or locked


Jamesxql:
I have seven web api applications on a same server hosted by `systemd` as a daemon . Sometimes when I stop my application, one of these applications will report this error:

```c#
Unhandled Exception: System.AggregateException: One or more errors occurred. (Error -16 EBUSY resource busy or locked) ---> Microsoft.AspNetCore.Server.Kestrel.Internal.Networking.UvException: Error -16 EBUSY resource busy or locked
    at Microsoft.AspNetCore.Server.Kestrel.Internal.Networking.Libuv.ThrowError(Int32 statusCode)
    at Microsoft.AspNetCore.Server.Kestrel.Internal.Networking.Libuv.loop_close(UvLoopHandle handle)
    at Microsoft.AspNetCore.Server.Kestrel.Internal.Networking.UvLoopHandle.ReleaseHandle()
    at System.Runtime.InteropServices.SafeHandle.InternalDispose()
    at Microsoft.AspNetCore.Server.Kestrel.Internal.KestrelThread.ThreadStart(Object parameter)
--- End of stack trace from previous location where exception was thrown ---
    at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
    at Microsoft.AspNetCore.Server.Kestrel.Internal.KestrelThread.<StopAsync>d__51.MoveNext()
--- End of inner exception stack trace ---
    at System.ThrowHelper.ThrowAggregateException(List`1 exceptions)
    at System.Threading.Tasks.Task.WaitAll(Task[] tasks, Int32 millisecondsTimeout, CancellationToken cancellationToken)
    at Microsoft.AspNetCore.Server.Kestrel.Internal.KestrelEngine.Dispose()
    at Microsoft.AspNetCore.Server.Kestrel.KestrelServer.Dispose()
    at Microsoft.Extensions.DependencyInjection.ServiceProvider.Dispose()
    at Microsoft.AspNetCore.Hosting.Internal.WebHost.Dispose()
    at Microsoft.AspNetCore.Hosting.WebHostExtensions.Run(IWebHost host, CancellationToken token, String shutdownMessage)
    at Microsoft.AspNetCore.Hosting.WebHostExtensions.Run(IWebHost host)
    at Oathkeeper.Api.Program.Main(String[] args)
```

The Kestrel version is :  1.1.2 
The Web Server: Ubuntu 16.04.2

[...]

PR #1633, opened 2017-04-08
Don't dispose WriteReqPool and PipeFactory too soon

halter73:
#1231

PR #2019, opened 2017-08-26
Fix EBUSY errors on uv_loop_close (#1909).

cesarbs:
#1909

Backports #1798 and #1633 to 1.1.x.


cesarbs:
Added one more test. Will merge when CI is green.

PR #2021, opened 2017-08-26
Use preparsed HttpVersion, rather than reparsing

benaadams:
Current path is:

 Parser -> HttpVersion Enum -> Get Version String -> Set `HttpVersion` with string -> Convert back to Enum

Can just set the enum directly since its already parsed.

Also using the const rather than strings; where the strings are in code

Resolves https://github.com/aspnet/KestrelHttpServer/issues/2022

### DIFF
--- a/src/Kestrel.Core/Internal/Http/Frame.cs
+++ b/src/Kestrel.Core/Internal/Http/Frame.cs
@@ -158,11 +158,11 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
             {
                 if (_httpVersion == Http.HttpVersion.Http11)
                 {
-                    return "HTTP/1.1";
+                    return HttpUtilities.Http11Version;
                 }
                 if (_httpVersion == Http.HttpVersion.Http10)
                 {
-                    return "HTTP/1.0";
+                    return HttpUtilities.Http10Version;
                 }
 
                 return string.Empty;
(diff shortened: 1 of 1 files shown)


src/Kestrel.Core/Internal/Http/Frame.cs
```diff
@@ -1224,7 +1224,7 @@ public void OnStartLine(HttpMethod method, HttpVersion version, Span<byte> targe
             Method = method != HttpMethod.Custom
                 ? HttpUtilities.MethodToString(method) ?? string.Empty
                 : customMethod.GetAsciiStringNonNullCharacters();
-            HttpVersion = HttpUtilities.VersionToString(version);
```
benaadams: This is the main change

src/Kestrel.Core/Internal/Http/Frame.cs
```diff
@@ -1224,7 +1224,7 @@ public void OnStartLine(HttpMethod method, HttpVersion version, Span<byte> targe
...
+            _httpVersion = version;
```
benaadams: This is the main change

benaadams:
Bad compiler?

> /home/travis/.nuget/packages/microsoft.netcore.compilers/2.6.0-rdonly-ref-61915-01/tools/Microsoft.CSharp.Core.targets(84,5): error MSB6006: "RunCsc" exited with code 131. 
> [/home/travis/build/aspnet/KestrelHttpServer/src/Protocols.Abstractions/Protocols.Abstractions.csproj]

cesarbs:
@natemcmaster just fixed the compiler.

PR #2023, opened 2017-08-26
Remove the events on ConnectionContext

davidfowl:
- Things seem to mostly work. ~One test needed to be commented out
because of the racy behavior.~
- ~Made IConnectionHandler.OnConnectionAsync return task
to ease testing~

### DIFF
--- a/src/Kestrel.Core/Internal/HttpConnectionMiddleware.cs
+++ b/src/Kestrel.Core/Internal/HttpConnectionMiddleware.cs
@@ -72,6 +72,12 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal
             },
             connection);
 
+            connectionContext.Transport.Output.OnReaderCompleted((error, state) =>
+            {
+                ((FrameConnection)state).OnConnectionClosed(error);
+            }, 
+            connection);
+
             return processingTask;
         }
     }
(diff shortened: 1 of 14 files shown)


src/Kestrel.Core/Internal/HttpConnectionMiddleware.cs
```diff
@@ -70,21 +70,14 @@ public Task OnConnectionAsync(ConnectionContext connectionContext)
 
             // Wire up the events an forward calls to the frame connection
             // It's important that these execute synchronously because graceful
-            // connection close is order sensative (for now)
-            connectionContext.ConnectionAborted.ContinueWith((task, state) =>
+            // connection close is order sensitive (for now)
+            connectionContext.InputClosed.ContinueWith((task, state) =>
```
davidfowl: @pakrym This is a problem with the callbacks we have today...

src/Kestrel.Core/Internal/HttpConnectionMiddleware.cs
```diff
@@ -63,27 +63,20 @@ public Task OnConnectionAsync(ConnectionContext connectionContext)
...
-            // the frame is created before this yields. Events need to be wired up
-            // afterwards
             var processingTask = connection.StartRequestProcessing(_application);
 
-            // Wire up the events an forward calls to the frame connection
-            // It's important that these execute synchronously because graceful
-            // connection close is order sensative (for now)
-            connectionContext.ConnectionAborted.ContinueWith((task, state) =>
+            // Abort the frame when the transport writer completes
+            connectionContext.Transport.Input.OnWriterCompleted((error, state) =>
```
davidfowl: @pakrym I think this is rooting the FrameConnection causing the test to fail. We never remove the callbacks it seems.

davidfowl: hmm, actually it looks like we do clear them...

Issue #2020, opened 2017-08-26
Unnecessary string equality checks in Frame


benaadams:
For GET method `IsHead` test is performed on the string `Method` which since `GET` isn't reference equal to `HEAD` drops through to `String.Equals` -> `OrdinalIgnoreCaseComparer.Equals`

![](https://aoa.blob.core.windows.net/aspnet/ishead.png)

However the Parser starts with it as an Enum, which then gets converted to a string; so it can just be tested with the enum.

Same thing happens with `HttpVersion`; where the `HttpVersion` enum is used to get a `HttpVersion` string to set the `HttpVersion` which then converts it back to a `HttpVersion` enum. It can just be set directly.

PR #2027, opened 2017-08-27
Use enum for method rather than string compares

benaadams:
For GET method `IsHead` test is performed on the string `Method` which since `GET` isn't reference equal to `HEAD` drops through to `String.Equals` -> `OrdinalIgnoreCaseComparer.Equals`

However the Parser starts with it as an Enum, which then gets converted to a string; so it can just be tested with the enum.

Http/2 change isn't so good; but ti gets method from header collection, so that probably needs a more general revisit

Resolves https://github.com/aspnet/KestrelHttpServer/issues/2020

### DIFF
--- a/src/Kestrel.Core/Internal/Http/HttpMethod.cs
+++ b/src/Kestrel.Core/Internal/Http/HttpMethod.cs
@@ -16,5 +16,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
         Options,
 
         Custom,
+
+        Unknown = byte.MaxValue
     }
 }
--- a/src/Kestrel.Core/Internal/Http2/Http2Stream.FeatureCollection.cs
+++ b/src/Kestrel.Core/Internal/Http2/Http2Stream.FeatureCollection.cs
@@ -11,6 +11,7 @@ using System.Threading.Tasks;
 using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.Http.Features;
 using Microsoft.AspNetCore.Server.Kestrel.Core.Features;
+using Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Infrastructure;
 using Microsoft.Extensions.Primitives;
 
 namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http2
@@ -92,8 +93,21 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http2
 
         string IHttpRequestFeature.Method
         {
-            get => Method;
-            set => Method = value;
+            get
+            {
+                if (CustomMethod != null)
+                {
+                    return CustomMethod;
+                }
+
+                CustomMethod = HttpUtilities.MethodToString(Method);
+                return CustomMethod;
+            }
+            set
+            {
+                CustomMethod = HttpUtilities.GetKnownMethod(value, out var method) ? null : value.ToUpperInvariant();
+                Method = method;
+            }
         }
 
         string IHttpRequestFeature.PathBase
--- a/src/Kestrel.Core/Internal/Infrastructure/HttpUtilities.cs
+++ b/src/Kestrel.Core/Internal/Infrastructure/HttpUtilities.cs
@@ -142,6 +142,112 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Infrastructure
             }
         }
 
+        /// <summary>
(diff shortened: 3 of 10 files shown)


arespr:
I think that enum types should inherit from lowest possible type. HttpMethod can inherit from sbyte.

benaadams:
> I think that enum types should inherit from lowest possible type. HttpMethod can inherit from sbyte.

Rather than `byte` as it does currently? what advantage would it confer?

blowdart:
It'd mean adding values is an even bigger breaking change than normal if people have decided to store it in a database somewhere.

src/Kestrel.Core/Internal/Http/HttpMethod.cs
```diff
@@ -16,5 +16,7 @@ public enum HttpMethod: byte
         Options,
 
         Custom,
+
+        Unknown = byte.MaxValue
```
Tratcher: It's best practice to make Unknown/None the zero value, the default for uninitialized fields.

src/Kestrel.Core/Internal/Http2/Http2Stream.FeatureCollection.cs
```diff
@@ -92,8 +93,21 @@ string IHttpRequestFeature.Scheme
...
+                {
+                    return CustomMethod;
+                }
+
+                CustomMethod = HttpUtilities.MethodToString(Method);
+                return CustomMethod;
+            }
+            set
+            {
+                CustomMethod = HttpUtilities.GetKnownMethod(value, out var method) ? null : value.ToUpperInvariant();
```
Tratcher: This mirrors the Http.Sys APIs

src/Kestrel.Core/Internal/Http/HttpMethod.cs
benaadams: Its used as the [index into `_methodNames` array](https://github.com/aspnet/KestrelHttpServer/blob/5bf7d3b28a2fcca56a7d24fd7e79c13dedae657b/src/Kestrel.Core/Internal/Infrastructure/HttpUtilities.Generated.cs#L41) and does some [pretty weird things](https://github.com/aspnet/KestrelHttpServer/blob/5bf7d3b28a2fcca56a7d24fd7e79c13dedae657b/src/Kestrel.Core/Internal/Infrastructure/HttpUtilities.Generated.cs#L65-L71) based on with the number/order

benaadams: Could just reset it to `Get` rather than `Unknown`; isn't a path where it can't get set to something

cesarbs:
> Http/2 change isn't so good; but ti gets method from header collection, so that probably needs a more general revisit

Yep, what's there now is just what's need to make it work. Still needs to be optimized (not just the method but all the pseudo-header field reads).

src/Kestrel.Core/Internal/Infrastructure/HttpUtilities.cs
```diff
@@ -142,6 +142,112 @@ public static unsafe bool GetKnownMethod(this Span<byte> span, out HttpMethod me
...
+            {
+                throw new ArgumentException(nameof(value));
+            }
+
+            // Clear the 6th bit to get Upper case char. We are testing direct equality 
+            // so doesn't matter that it would garble chars outside [A-Za-z] range
+            var firstUppercaseChar = (char)(value[0] & ~0x20);
+            if (length == 3)
+            {
+                if (firstUppercaseChar == 'G' && string.Equals(value, "GET", StringComparison.OrdinalIgnoreCase))
```
cesarbs: > `OrdinalIgnoreCase`

https://tools.ietf.org/html/rfc7230#section-3.1.1:

> The request method is case-sensitive.

benaadams: Somethings wonder if they flip coins to determine what is case sensitive in the specs. Will change

Tratcher: When it doubt, use case in-sensitive for parsing, regardless of the spec.

halter73: @Tratcher I would go with stricter parsing when in doubt. The risk is that Kestrel might interpret a request differently than a proxy. For instance, imagine a proxy is configured with rules that should be applied to all POST requests to a given endpoint. With case-insensitive parsing in Kestrel, these rules could possibly be bypassed by sending a pOsT request.

davidfowl:
This PR needs to do over. We refactored everything for http2

PR #2033, opened 2017-08-27
Fix typo in FrameTests

DamirAinullin:
Fix typo in FrameTests.


benaadams:
Might be more major than a typo :)

/cc @davidfowl

davidfowl:
Hah, that just looks like a bug 😉

davidfowl:
Thanks!

PR #2034, opened 2017-08-28
Shorter LibuvThread locks

benaadams:
Create the objects out of locks
Different locks for the different queues

Burning some time in `Post`


halter73:
:+1: We know this can be a pretty highly contested lock when there's a lot of IO.

benaadams:
Renamed

halter73:
Thanks

PR #2038, opened 2017-08-28
Add waits in ListenerPrimaryTests to mitigate race with List.Add

cesarbs:
Same fix @natemcmaster applied in #1769.

### DIFF
--- a/test/Microsoft.AspNetCore.Server.KestrelTests/ListenerPrimaryTests.cs
+++ b/test/Microsoft.AspNetCore.Server.KestrelTests/ListenerPrimaryTests.cs
@@ -189,8 +189,8 @@ namespace Microsoft.AspNetCore.Server.KestrelTests
 
                 await kestrelThreadPrimary.PostAsync(_ => pipe.Dispose(), null);
 
-                // Wait up to 10 seconds for error to be logged
-                for (var i = 0; i < 10 && primaryTrace.Logger.TotalErrorsLogged == 0; i++)
+                // Wait up to 30 seconds for error to be logged
+                for (var i = 0; i < 300 && primaryTrace.Logger.TotalErrorsLogged == 0; i++)
                 {
                     await Task.Delay(100);
                 }
@@ -265,8 +265,8 @@ namespace Microsoft.AspNetCore.Server.KestrelTests
                 var listenerSecondary = new TcpListenerSecondary(serviceContextSecondary);
                 await listenerSecondary.StartAsync(pipeName, Guid.NewGuid().ToByteArray(), address, kestrelThreadSecondary);
 
-                // Wait up to 10 seconds for error to be logged
-                for (var i = 0; i < 10 && primaryTrace.Logger.TotalErrorsLogged == 0; i++)
+                // Wait up to 30 seconds for error to be logged
+                for (var i = 0; i < 300 && primaryTrace.Logger.TotalErrorsLogged == 0; i++)
                 {
                     await Task.Delay(100);
                 }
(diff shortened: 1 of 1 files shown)


cesarbs:
We should do this across all branches (1.0, 1.1 and dev), to all applicable tests.

src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/ListenerPrimary.cs
```diff
@@ -33,6 +33,9 @@ protected ListenerPrimary(ServiceContext serviceContext) : base(serviceContext)
         {
         }
 
+        // For testing purposes.
+        internal int UvPipeCount => _dispatchPipes.Count;
```
pranavkm: cc @muratg since there's a src code change (even though nothing's changing 👍 )

muratg: This looks fine to me (adding internal property in an internal namespace). @Eilon is this OK?

Eilon: Not crazy about this. I would rather not have any API change at all.

halter73:
@cesarbs I think #1230 was the more important post-1.0 fix to he flakiness in ListenerPrimaryTests.

Are you sure #1769 fixed any flakiness? I'm suspect it might not have considering @natemcmaster said he couldn't reproduce the issue even in a tight loop prior to the changes.

As far as I can tell, the only way #1769 could have possibly helped is by effectively increasing the timeout for the secondary listener to connect from 10 seconds (that used by #1769) to 30 seconds (the `maxWait` in #1769). My reasoning for this based on the premise that `AssertResponseEventually` can't possibly complete until after the `UvPipeHandle` is added to `_dispatchPipes`.

I would just backport #1230 and increase the default `maxRetries` to 300 to get the increased 30 second timeout. I would also change `maxRetries` to 300 in the dev branch and revert #1769 unless someone can explain how #1769 helps reduce flakiness given `AssertResponseEventually`.

cesarbs:
Updated with backport of #1230. @Eilon This doesn't involve any product code change and stabilizes tests.

test/Microsoft.AspNetCore.Server.KestrelTests/ListenerPrimaryTests.cs
```diff
@@ -277,6 +289,32 @@ public class ListenerPrimaryTests
             Assert.Contains("Bad data", errorMessage.Exception.ToString());
         }
 
+        private static async Task AssertResponseEventually(
+            string address,
+            string expected,
+            string[] allowed = null,
+            int maxRetries = 100,
```
halter73: I think upping `maxRetries` to 300 for a 30 second timeout should be better.

test/Microsoft.AspNetCore.Server.KestrelTests/ListenerPrimaryTests.cs
```diff
@@ -189,6 +189,12 @@ public class ListenerPrimaryTests
 
                 await kestrelThreadPrimary.PostAsync(_ => pipe.Dispose(), null);
 
+                // Wait up to 10 seconds for error to be logged
+                for (var i = 0; i < 10 && primaryTrace.Logger.TotalErrorsLogged == 0; i++)
```
halter73: I know you just copied this from #1230, but this only waits up to 1 second not 10 seconds. I would update `< 10` to `< 300` for a 30 second timeout to match `AssertResponseEventually`.

Eilon: Derp, didn't see this until now, but I made a similar comment in the other PR.

cesarbs:
Updated with 30 second timeouts.

PR #2040, opened 2017-08-28
Add wait to NonListenerPipeConnectionsAreLoggedAndIgnored to mitigate race with List.Add

cesarbs:
`NonListenerPipeConnectionsAreLoggedAndIgnored` is subject to the same race @natemcmaster mitigated in #1769.


cesarbs:
This was mitigated in #1230.

Issue #2017, opened 2017-08-24
Merge code paths duplicated between HTTP/1.1 and HTTP/2


cesarbs:
The initial HTTP/2 implementation contains a lot of duplicated code from the existing HTTP/1.1 stuff. That duplication should be eliminated so we can share as much code as possible between the two protocol implementations.

PR #2041, opened 2017-08-29
Merge code paths duplicated between HTTP/1.x and HTTP/2 implementations (#2017)

cesarbs:
#2017

## A note on naming

I'm intentionally not renaming things now, so they look a bit off. I'm planning on doing some renames after addressing feedback (see below).

## Summary

This change makes `Frame` the base class for the HTTP/1.1 and HTTP/2 implementations.

The HTTP/1.x-specific code in `Frame` is in a subclass called `Http1Frame` (so now `Frame<T>` subclasses `Http1Frame`).

`Http2Stream` is a subclass of `Frame`.

`Frame` exposes a few more abstract properties and methods that must be implemented by protocol-specific classes.

Producing HTTP/1.1 and HTTP/2 output is done via the same interface (`IHttpOutputProducer`), which `OutputProducer` and `Http2OutputProducer` implement.

## Things I'm not crazy about

1) Multiple `.FeatureCollection.cs` files. I think the ideal would be to have a single one (`Frame.FeatureCollection.cs`). It would implement HTTP/1.1 and HTTP/2-specific features (upgrade and stream ID, respectively), but those could be nulled out in `OnReset()` where they're not supposed to be implemented. What do you guys think?

1) The way `Frame.Output` is instantiated by each subclass. It's all internal but ugly still. What do you guys think of moving it to `IHttpProtocolContext` i.e. have it passed to `Frame` instead of instantiated by it?

1) Protected member variables. They were already there, but now they're uglier IMO because some are not accessed by a direct subclass of `Frame` (e.g. `Frame<T>` accesses `Frame._applicationException`).

## Planned renames

* `Frame` -> `HttpProtocol` (I wanted to name it `HttpStream` but @davidfowl didn't like that)
* `FrameAdapter` -> `Http1FrameAdapter`
* `IFrameControl` -> `IHttpResponseControl`
* `FrameRequestStream` -> `HttpRequestStream`
* `FrameResponseStream` -> `HttpResponseStream`
* `FrameRequestHeaders` -> `HttpRequestHeaders`
* `FrameResponseHeaders` -> `HttpResponseHeaders`
* `FrameConnection` -> `HttpConnection`
* `OutputProducer` -> `Http1OutputProducer`

I believe these would be better done as part of the Bedrock refactoring:

* `FrameConnectionManager` -> `ConnectionManager` or `HttpConnectionManager` (if HTTP-specific)
* `FrameHeartbeatManager` -> `HeartbeatManager` or `HttpHeartbeatManager` (if HTTP-specific)


halter73:
Why don't you like multiple .FeatureCollection.cs files? You still removed the duplication, right?

halter73:
If you want to replace Frame with Http in our types I would do it everywhere.

cesarbs:
> Why don't you like multiple .FeatureCollection.cs files? You still removed the duplication, right?

It's personal taste, really. If everyone is fine with it I'll leave it like that.

src/Kestrel.Core/Internal/Http/OutputProducer.cs
```diff
@@ -49,7 +54,7 @@ public class OutputProducer : IDisposable
             _flushCompleted = OnFlushCompleted;
         }
 
-        public Task WriteAsync(ArraySegment<byte> buffer, bool chunk = false, CancellationToken cancellationToken = default(CancellationToken))
+        public Task WriteDataAsync(ArraySegment<byte> buffer, bool chunk = false, CancellationToken cancellationToken = default(CancellationToken))
```
halter73: WriteToBodyAsync?

src/Kestrel.Core/Internal/Http2/Http2FrameWriter.cs
```diff
@@ -50,6 +51,39 @@ public Task Write100ContinueAsync(int streamId)
...
+
+                var done = _hpackEncoder.BeginEncode(statusCode, EnumerateHeaders(headers), _outgoingFrame.Payload, out var payloadLength);
+                _outgoingFrame.Length = payloadLength;
+
+                if (done)
+                {
+                    _outgoingFrame.HeadersFlags = Http2HeadersFrameFlags.END_HEADERS;
+                }
+
+                Write(_outgoingFrame.Raw);
```
halter73: Can this Write block on IO? It better not. Flushing headers never blocks for HTTP/1.1.

src/Kestrel.Core/Internal/Http2/Http2FrameWriter.cs
```diff
@@ -50,6 +51,39 @@ public Task Write100ContinueAsync(int streamId)
             return Task.CompletedTask;
         }
 
+        public void WriteHeaders(int streamId, int statusCode, IHeaderDictionary headers)
+        {
+            lock (_writeLock)
```
halter73: What is this lock protecting against? In the async version of this method, the lock is released before the write tasks complete.

src/Kestrel.Core/Internal/Http2/Http2FrameWriter.cs
```diff
@@ -168,6 +202,18 @@ public Task WriteGoAwayAsync(int lastStreamId, Http2ErrorCode errorCode)
             }
         }
 
+        private void Write(ArraySegment<byte> data)
```
halter73: Does this need to be called with the _writeLock. If so,  I'd add a comment.

src/Kestrel.Core/Internal/Http2/Http2OutputProducer.cs
```diff
@@ -0,0 +1,57 @@
...
+        }
+
+        public Task WriteStreamSuffixAsync(CancellationToken cancellationToken)
+        {
+            return _frameWriter.WriteDataAsync(_streamId, _emptyData, endStream: true, cancellationToken: cancellationToken);
+        }
+
+        public void WriteResponseHeaders(int statusCode, string ReasonPhrase, FrameResponseHeaders responseHeaders)
+        {
+            _frameWriter.WriteHeaders(_streamId, statusCode, responseHeaders);
```
halter73: Nit: Why does IHttp2FrameWriter and IHttpOutputProducer have different method names for seemingly the same method?

src/Kestrel.Core/Internal/Http2/Http2OutputProducer.cs
```diff
@@ -0,0 +1,57 @@
...
+
+        public Task FlushAsync(CancellationToken cancellationToken) => _frameWriter.FlushAsync(cancellationToken);
+
+        public Task Write100ContinueAsync(CancellationToken cancellationToken) => _frameWriter.Write100ContinueAsync(_streamId);
+
+        public Task WriteDataAsync(ArraySegment<byte> data, bool chunk, CancellationToken cancellationToken)
+        {
+            if (chunk)
+            {
+                throw new ArgumentException("Chunked transfer coding is not supported in HTTP/2", nameof(chunk));
```
halter73: There was a short time where we passed the chunk bool to WriteAsync in order to allow chunks to be aggregated, but now OutputProducer.WriteAsync just adds the chunking inline.

So instead of adding the chunk param to the shared IHttpOutputProducer interface, I would just add the chunks in Frame.WriteChunkedAsync.

src/Kestrel.Core/Internal/Http/Http1Frame.cs
```diff
@@ -0,0 +1,447 @@
...
+                }
+            }
+        }
+
+        protected override string CreateRequestId()
+            => StringUtilities.ConcatAsHexSuffix(ConnectionId, ':', _requestCount);
+
+        protected override void OnReset()
+        {
+            ExtraFeatureSet(typeof(IHttpUpgradeFeature), this);
```
halter73: Is this why you don't like the current design? I definitely don't like allocating the MaybeExtra by default for each Frame object.

src/Kestrel.Core/Internal/Http2/Http2Stream.cs
```diff
@@ -1,1025 +1,54 @@
...
-            await WriteSuffix();
-        }
-
-        private Task WriteSuffix()
+        protected override Task WriteSuffix()
         {
             if (HttpMethods.IsHead(Method) && _responseBytesWritten > 0)
             {
                 Log.ConnectionHeadResponseBodyWrite(ConnectionId, _responseBytesWritten);
             }
```
halter73: Couldn't the above go in a virtual base method? It looks exactly the same as the log and condition used in Http1Frame.WriteSuffix().

I'm not morally opposed to calling base methods. I get the appeal of the "OnEvent" abstract methods, but I'd rather have calls to base methods than duplicate code like this.

src/Kestrel.Core/Internal/Http2/Http2Stream.cs
```diff
@@ -1,1025 +1,54 @@
...
-                    {
-                        await entry.Key.Invoke(entry.Value);
-                    }
-                    catch (Exception ex)
-                    {
-                        ReportApplicationError(ex);
-                    }
-                }
-            }
+            ExtraFeatureSet(typeof(IHttp2StreamIdFeature), this);
```
halter73: I think we need to generate fields and have virtual calls to FastFeatureGet/Set instead of always creating a MaybeExtra list.

src/Kestrel.Core/Internal/Http2/Http2StreamOfT.cs
```diff
@@ -22,18 +22,20 @@ public Http2Stream(IHttpApplication<TContext> application, Http2StreamContext co
             _application = application;
         }
 
-        public override async Task ProcessRequestAsync()
+        public override async Task ProcessRequestsAsync()
```
halter73: Can we not share more code from this method? The error handling code in particular looks extremely complex and similar to FrameOfT.

src/Kestrel.Core/Internal/Http2/Http2FrameWriter.cs
cesarbs: It protects access to `_outgoingFrame`.

src/Kestrel.Core/Internal/Http2/Http2FrameWriter.cs
cesarbs: It's not different than the sync `Write` in `OutputProducer`, except there's no callback involved.

src/Kestrel.Core/Internal/Http/Http1Frame.cs
cesarbs: Actually I hadn't realized this would allocate every time, but I felt the multiple `.FeatureCollection.cs` files kinda smelled. Better have it all in `Frame.FeatureCollection.cs` and null out features that don't belong to a protocol version, right?

src/Kestrel.Core/Internal/Http2/Http2StreamOfT.cs
cesarbs: Agreed. I'll look into it.

src/Kestrel.Core/Internal/Http/OutputProducer.cs
cesarbs: WriteResponseBodyAsync?

src/Kestrel.Core/Internal/Http2/Http2FrameWriter.cs
davidfowl: So it more like append?

halter73: Yeah. From looking at it Write doesn't called flush so there's no back pressure. I would probably rename Write to Append in this method to make it clear it's not doing any IO.

I don't see who's calling WriteHeadersAsync which does flush. I would probably remove that method. Today, we never handle IO errors or back pressure when writing headers. If anything like that happens we let the application deal with it when they flush the first write to the body.

src/Kestrel.Core/Internal/Http/Http1Frame.cs
halter73: I don't see the problem with multiple FeatureCollection.cs files.

src/Kestrel.Core/Internal/Http/Frame.FeatureCollection.cs
```diff
@@ -282,7 +280,7 @@ void IHttpResponseFeature.OnCompleted(Func<object, Task> callback, object state)
                 throw new InvalidOperationException(CoreStrings.CannotUpgradeNonUpgradableRequest);
             }
 
-            if (_wasUpgraded)
+            if (IsUpgraded)
             {
                 throw new InvalidOperationException(CoreStrings.UpgradeCannotBeCalledMultipleTimes);
```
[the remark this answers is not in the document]
cesarbs: Just rebased and took care of that.

src/Kestrel.Core/Internal/Http2/Http2Stream.cs
cesarbs: I put it in `Frame` (no need to override): 7f15df1ac614401d172569feab45ba842a474edc

Still dupes the HEAD check but that follows the pattern of other methods which have an `Awaited` version. I could eliminate the duplication by moving the check to the top of `WriteSuffix`. It's not a big deal to change the order of log messages in that situation.

src/Kestrel.Core/Internal/Http2/Http2OutputProducer.cs
cesarbs: 974e5532083fa546237b93459c0bdc0beb04d5c7

I prefer this to not having the `chunk` parameter in the interface, but it's still protocol-specific in that `Http2OutputProducer` doesn't implement the new `WriteAsync` method.

src/Kestrel.Core/Internal/Http/Frame.FeatureCollection.cs
```diff
@@ -42,7 +41,7 @@ public void ResetFeatureCollection()
             _featureRevision++;
         }
 
-        private object ExtraFeatureGet(Type key)
+        protected object ExtraFeatureGet(Type key)
```
halter73: As far as I can tell this can be made private again.

cesarbs:
Updated list of planned renames (updates in **bold**):

* `Frame` -> `HttpProtocol`
* `FrameAdapter` -> **`Http1ParsingHandler`** (since it's now only used in parsing-related code)
* `IFrameControl` -> `IHttpResponseControl`
* `FrameRequestStream` -> `HttpRequestStream`
* `FrameResponseStream` -> `HttpResponseStream`
* `FrameRequestHeaders` -> `HttpRequestHeaders`
* `FrameResponseHeaders` -> `HttpResponseHeaders`
* `FrameConnection` -> `HttpConnection`
* **`Http1Frame`** -> **`Http1Connection`**
* `OutputProducer` -> `Http1OutputProducer`
* `FrameConnectionManager` -> **`HttpConnectionManager`**
* `FrameHeartbeatManager` -> **`HttpHeartbeatManager`**

cesarbs:
A few more necessary renames that I had overlooked before:

`FrameHeaders` -> `HttpHeaders`
`FrameDuplexStream` -> `HttpUpgradeStream`
`FrameConnectionReference` -> `HttpConnectionReference`
`FrameStreamState` -> `HttpStreamState`
`FrameContext` -> `Http1ConnectionContext`

cesarbs:
Commit b9cf5d80e8325115f39557456543b39815656605 does the renames.

Issue #1673, opened 2017-04-13
IHttpSendFileFeature provided by Transport


tmds:
Sending files can be offloaded to the kernel. The Transport can expose this as a sendfile feature.

Some implementations support encrypting (BSD sendfile, Http.Sys) while others do not (Linux sendfile, corefx Socket).

https://github.com/aspnet/KestrelHttpServer/issues/1593 tracks Transport independent implementation. 

cc @davidfowl @Tratcher @Drawaes @benaadams

tmds:
Does someone have an idea of how this might work? I'm pondering about it, but I can't figure out how the the pipeline data flow and the sendfile system call work together.

Tratcher:
SendFile bypasses the normal datastream and writes directly to the socket, though it should flush the datastream fist to ensure proper ordering. Higher layers that would behave incorrectly if bypassed (SSL, caching, etc.) should fake the feature by wrapping the datastream, or remove the feature.

tmds:
@Tratcher makes sense. SendFile implementation would indeed need to flush the datastream, also it also needs to take into account the Connection lifetime. HTTP 2 is also a higher layer, right?

benaadams:
HTTP/2 is on HTTPS (for all practical purposes). Is OS sendfile non-blocking/async, e.g. will it callback on send complete so the connection can be resumed for keepalive?

tmds:
> (for all practical purposes)

I've heard that too. Hard to imagine no-one wants to cut out the cost of encryption for some use-cases, but I'm no expert.

> Is OS sendfile non-blocking/async, e.g. will it callback on send complete so the connection can be resumed for keepalive?

On Linux, it's a blocking call which returns when the file is sent. nginx offloads sendfile to a threadpool to avoid blocking the main worker thread.
Once the sendfile call returns, the connection can be used for other requests.

Tratcher:
SendFile is async on Windows. We avoided implemented it in the past due to the sync API on linux.

davidfowl:
This doesn't work well at the transport layer.

tmds:
> This doesn't work well at the transport layer.

@davidfowl In the Pipelines API review issue you said this was Kestrel specific. Now you say it doesn't work at the Transport layer. Is the conclusion, Kestrel/Pipelines cannot leverage the `sendfile` system call?

davidfowl:
IHttpSendFileFeature is ASP.NET specific.

[...]

PR #2049, opened 2017-09-07
[PoC] Implement IHttpSendFileFeature

benaadams:
With "Add Memory-based Stream overloads to coreclr" https://github.com/dotnet/coreclr/pull/13769

Can read directly from the `FileStream` into the `Pipeline` output buffer to implement a lower copy SendFile than is done by [`StreamCopyOperation.CopyToAsync`](https://github.com/aspnet/HttpAbstractions/blob/1e8a22dae3dd2828bcaca936821be8efa47824e2/src/Microsoft.AspNetCore.Http.Extensions/StreamCopyOperation.cs#L36-L84)

i.e. Read Stream -> array -> Write Stream -> copy to Pipeline

May want to bypass the feature if the Transport has something https://github.com/aspnet/KestrelHttpServer/issues/1673

/cc @halter73 @davidfowl @stephentoub @tmds @Tratcher

### DIFF
--- a/src/Kestrel.Core/Internal/Http/Http1OutputProducer.cs
+++ b/src/Kestrel.Core/Internal/Http/Http1OutputProducer.cs
@@ -4,6 +4,7 @@
 using System;
 using System.Buffers;
 using System.Diagnostics;
+using System.IO;
 using System.IO.Pipelines;
 using System.Runtime.CompilerServices;
 using System.Text;
@@ -24,8 +25,8 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
         private readonly ITimeoutControl _timeoutControl;
         private readonly IKestrelTrace _log;
 
-        // This locks access to to all of the below fields
-        private readonly object _contextLock = new object();
+        // This locks access to the output writing fields below
+        protected readonly SemaphoreSlim _contextSemaphore = new SemaphoreSlim(1);
 
         private bool _completed = false;
 
@@ -76,7 +77,8 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
 
         public void Write<T>(Action<PipeWriter, T> callback, T state)
         {
-            lock (_contextLock)
+            _contextSemaphore.Wait();
+            try
             {
                 if (_completed)
                 {
@@ -87,11 +89,38 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
                 callback(buffer, state);
                 buffer.Commit();
             }
+            finally
+            {
+                _contextSemaphore.Release();
+            }
         }
 
         public Task WriteAsync<T>(Action<PipeWriter, T> callback, T state)
         {
-            lock (_contextLock)
+            var task = _contextSemaphore.WaitAsync();
+#if NETCOREAPP2_1
+            if (!task.IsCompletedSuccessfully)
+            {
+                return WriteAsync(task, callback, state);
+            }
+#else
+            if (!task.IsCompleted || task.IsFaulted || task.IsCanceled)
+            {
+                return WriteAsync(task, callback, state);
+            }
+#endif
+            return WriteAsyncWithLock(callback, state);
+        }
+
+        private async Task WriteAsync<T>(Task task, Action<PipeWriter, T> callback, T state)
+        {
+            await task;
+            await WriteAsyncWithLock(callback, state);
+        }
+
+        private Task WriteAsyncWithLock<T>(Action<PipeWriter, T> callback, T state)
+        {
+            try
             {
                 if (_completed)
                 {
@@ -102,13 +131,72 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
                 callback(buffer, state);
                 buffer.Commit();
             }
+            finally
+            {
+                _contextSemaphore.Release();
+            }
 
             return FlushAsync();
         }
 
+#if NETCOREAPP2_1
+        public async Task WriteAsync(Stream stream, long? count, CancellationToken cancellationToken)
+        {
+            const int maxBlockSize = 4032; // Should be: Environment.SystemPageSize https://github.com/dotnet/corefxlab/pull/2099
+
+            var buffer = _pipeWriter;
+            var remaining = count ?? long.MaxValue;
+            int outputBytes;
+            do
+            {
+                await _contextSemaphore.WaitAsync();
+                try
+                {
+                    if (_completed)
+                    {
+                        return;
+                    }
+
+                    var memory = buffer.GetMemory(remaining > maxBlockSize ? maxBlockSize : (int)remaining);
+                    // Read the stream directly into the output buffer
+                    if (memory.Length > remaining)
+                    {
+                        memory = memory.Slice((int)remaining);
+                    }
+                    outputBytes = await stream.ReadAsync(memory, cancellationToken);
+                    remaining -= outputBytes;
+                    if (outputBytes > 0)
+                    {
+                        buffer.Advance(outputBytes);
+                        buffer.Commit();
+                    }
+                }
+                finally
+                {
+                    _contextSemaphore.Release();
+                }
+
+                // WriteAsync() - not forced flush here
+                await buffer.FlushAsync(cancellationToken);
+            } while (outputBytes > 0 && !cancellationToken.IsCancellationRequested);
+            // FlushAsync() - forced flush; or better move it to outer loop
+
+            if (count.HasValue && remaining > 0)
+            {
+                // throw
+            }
+        }
+#else
+        public Task WriteAsync(Stream stream, long? count, CancellationToken cancellationToken)
+        {
+            throw new NotImplementedException();
+        }
+#endif
+
         public void WriteResponseHeaders(int statusCode, string reasonPhrase, HttpResponseHeaders responseHeaders)
         {
-            lock (_contextLock)
+            _contextSemaphore.Wait();
+            try
             {
                 if (_completed)
                 {
@@ -125,11 +213,16 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
                 writer.Write(_bytesEndHeaders);
                 buffer.Commit();
             }
+            finally
+            {
+                _contextSemaphore.Release();
+            }
         }
 
         public void Dispose()
         {
-            lock (_contextLock)
+            _contextSemaphore.Wait();
+            try
             {
                 if (_completed)
                 {
@@ -140,11 +233,16 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
                 _completed = true;
                 _pipeWriter.Complete();
             }
+            finally
+            {
+                _contextSemaphore.Release();
+            }
         }
 
         public void Abort(Exception error)
         {
-            lock (_contextLock)
+            _contextSemaphore.Wait();
+            try
             {
                 if (_completed)
                 {
@@ -156,6 +254,10 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
 
                 _outputPipeReader.CancelPendingRead();
                 _pipeWriter.Complete(error);
+            }
+            finally
+            {
+                _contextSemaphore.Release();
             }
         }
 
@@ -170,7 +272,9 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
         {
             var writableBuffer = default(PipeWriter);
             long bytesWritten = 0;
-            lock (_contextLock)
+
+            _contextSemaphore.Wait();
+            try
             {
                 if (_completed)
                 {
@@ -186,6 +290,10 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
                 }
 
                 writableBuffer.Commit();
+            }
+            finally
+            {
+                _contextSemaphore.Release();
             }
 
             return FlushAsync(writableBuffer, bytesWritten, cancellationToken);
--- a/src/Kestrel.Core/Internal/Http/HttpProtocol.SendFileFeature.cs
+++ b/src/Kestrel.Core/Internal/Http/HttpProtocol.SendFileFeature.cs
@@ -1,0 +1,38 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.IO;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Http.Features;
+
+namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
+{
+    public partial class HttpProtocol : IHttpSendFileFeature
+    {
+        async Task IHttpSendFileFeature.SendFileAsync(string path, long offset, long? count, CancellationToken cancellationToken)
+        {
+            var contentLength = ResponseHeaders.ContentLength;
+            if (contentLength.HasValue && count.HasValue && contentLength.Value < _responseBytesWritten + count.Value)
+            {
+                // throw
+            }
+
+            using (var fileStream = new FileStream(path,
+                                 FileMode.Open,
+                                 FileAccess.Read,
+                                 FileShare.ReadWrite,
+                                 bufferSize: 1, // Don't create internal buffer
+                                 FileOptions.Asynchronous | FileOptions.SequentialScan))
+            {
+                if (offset > 0)
+                {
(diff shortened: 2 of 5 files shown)


src/Kestrel.Core/Internal/Http/OutputProducer.cs
```diff
@@ -148,6 +149,41 @@ public void Abort(Exception error)
             return FlushAsync(writableBuffer, cancellationToken);
         }
 
+        public async Task WriteAsync(
+            Stream stream,
+            CancellationToken cancellationToken)
+        {
+
+            Monitor.Enter(_contextLock);
```
benaadams: Not sure about lock here :-/

src/Kestrel.Core/Internal/Http/Frame.FeatureCollection.cs
```diff
@@ -321,5 +322,26 @@ void IHttpRequestLifetimeFeature.Abort()
         {
             Abort(error: null);
         }
+
+        async Task IHttpSendFileFeature.SendFileAsync(string path, long offset, long? count, CancellationToken cancellation)
+        {
+            ResponseHeaders.ContentLength = count;
```
davidfowl: So the idea is that SendFileAsync takes over the whole response?

src/Kestrel.Core/Internal/Http/OutputProducer.cs
davidfowl: Yea, I'm not sure about this either. async + Monitor... is a bad combination

src/Kestrel.Core/Internal/Http/Frame.FeatureCollection.cs
tmds: Leaving this up to the caller, makes it usable in more cases.

The caller needs to figure out the length of the file, so perhaps a `FileStream` instead of string path argument makes more sense?

src/Kestrel.Core/Internal/Http/Frame.FeatureCollection.cs
```diff
@@ -321,5 +322,26 @@ void IHttpRequestLifetimeFeature.Abort()
...
+            using (var fileStream = new FileStream(path,
+                                             FileMode.Open,
+                                             FileAccess.Read,
+                                             FileShare.ReadWrite,
+                                             4096, // Buffer size - not currently pooled
+                                             FileOptions.Asynchronous | FileOptions.SequentialScan))
+            {
+                if (offset > 0)
+                {
+                    // No SeekAsync?
```
tmds: Seeking adjust an internal offset field, so no Async as this isn't doing much.

src/Kestrel.Core/Internal/Http/Frame.FeatureCollection.cs
```diff
@@ -321,5 +322,26 @@ void IHttpRequestLifetimeFeature.Abort()
...
+                    fileStream.Seek(offset, SeekOrigin.Begin);
+                }
+
+                // Write the filestream directly into the output
+                await Output.WriteAsync(fileStream, cancellation);
```
tmds: This isn't using the `count` parameter.

src/Kestrel.Core/Internal/Http/Frame.FeatureCollection.cs
tmds: "Range Requests": https://tools.ietf.org/html/rfc7233
Probably not a major use-case for Kestrel, but would be nice if it could be implemented using `IHttpSendFileFeature`.

davidfowl: I'm not sure this belongs at the send file feature level. It's currently implemented in higher level components.

tmds: I meant `IHttpSendFileFeature` should not set ContentLength and leave it up to the higher level components. e.g. to implement multi-range responses as described in the rfc.

davidfowl: Ah, yes I agree

benaadams: > "Range Requests"

StaticFiles will use `IHttpSendFileFeature` for [these if available](https://github.com/aspnet/StaticFiles/blob/1e0fe87c204b0199b97bc1d74f095e9746e9e674/src/Microsoft.AspNetCore.StaticFiles/StaticFileContext.cs#L369-L374)

benaadams: I removed the header; but its at a low level; so it needs to know if chunking and then chunk ;-/

benaadams: nvm `StaticFiles` sets the header before calling `SendFile` so this works

src/Kestrel.Core/Internal/Http/OutputProducer.cs
```diff
@@ -148,6 +149,47 @@ public void Abort(Exception error)
...
+            long length,
+            CancellationToken cancellationToken)
+        {
+
+            var writableBuffer = default(WritableBuffer);
+            var remaining = length;
+            int count;
+            do
+            {
+                lock (_contextLock)
```
benaadams: @halter73 would this locking be enough?

halter73: Given our current locking strategy which is to ensure that the contents of concurrent calls to Response.Body.Write and similar don't interleave, then no.

I'm not in love with the strictness of this strictness of this strategy, but at least now it should be low contention unlike in the 1.x versions of Kestrel where the _contextLock would be shared with the consumer on the libuv thread.

davidfowl: We might have to change to a semaphore

src/Kestrel.Core/Internal/Http/OutputProducer.cs
```diff
@@ -148,6 +149,47 @@ public void Abort(Exception error)
...
+
+                count = await stream.ReadAsync(buffer, cancellationToken);
+                remaining -= count;
+                if (count > 0)
+                {
+                    writableBuffer.Advance(count);
+
+                    await writableBuffer.FlushAsync(cancellationToken);
+                }
+            } while (count > 0);
```
tmds: What should happen when we reach EOF before we read `count` (SendFileAsync arg)? Should this throw?

Tratcher: Yes, both offset and count must be in range. Otherwise you'd end up sending a different amount of data than the caller expected you to and this would break Content-Length.

src/Kestrel.Core/Internal/Http/OutputProducer.cs
benaadams: Semaphore is allocately. Gate that works for sync and async?

tmds: This semaphore, is that something that a user of `IHttpSendFileFeature` must manually lock/unlock to ensure the entire response doesn't interleave with other output?

davidfowl: Today we prevent overlapping writes by locking. SendFile is basically a write.

benaadams: Would be automatic

tmds: If you have some code like:
```
WriteHeaders
SendFile
```
This unlocks in-between. Can http2 interleave responses in a way that this would be an issue?

davidfowl: What is WriteHeaders?

> This unlocks in-between. Can http2 interleave responses in a way that this would be an issue?

Nope.

tmds: > What is WriteHeaders?

What ever code writing the headers before sendfile is writing the content.

> Nope.

In general, why don't responses interleaved? Does a response get packed as a "WritableBuffer"? Or is there some other lock?

davidfowl: > What ever code is writing the headers before sendfile is writing the content.

Unlocking in between those calls is fine. We're just making sure the buffer isn't corrupted (i.e. 2 calls don't write to the WritableBuffer in parallel). 

> In general, why don't responses interleave? Does a response get packed as a "WritableBuffer"? Or is there some other lock?

I'm not sure what you mean?

tmds: For a response, something is 'grouping' the data that needs to be sent together (e.g on Flush). Then this 'group' is sent as a a whole (so no data from other responses can interleave).
In general: How is this implemented? Is this grouping done by `WritableBuffer` and is `_contextLock` responsible for appending a `WritableBuffer` to the output stream? Or something else?
Here specifically: How does `SendFileAsync` get 'grouped' with the response headers?

benaadams: I think you are going too deep; it doesn't compile yet 😉 

Though it should check the headers and trigger them prior to sending the file data? (i.e. they are normal triggered by a `Response.Body.Write` which this bypasses)

Transport level send file wouldn't be able to do that; so it should intercept even if the transport supported `SendFile` for prep-work?

If `ContentLength` is not set it should also chunk? Not sure how large a single chunk can be (i.e. whole file) as it shouldn't chunk inside the file data, or is altering `SendFile`'s expected behaviour too much (i.e. just top and tail)?

tmds: > Though it should check the headers and trigger them prior to sending the file data?

Yes, if headers are not sent I think they should be at that point.

> If ContentLength is not set it should also chunk? Not sure how large a single chunk can be (i.e. whole file) as it shouldn't chunk inside the file data, or is altering SendFile's expected behaviour too much (i.e. just top and tail)?

I think that would be appropriate. So when `long? count` is `null`, it should be set it to `File.Length - offset`. Then count would be the size of the chunk.

For the Transport provided SendFile, I think we could use a lower-level IHttp**Transport**SendFileFeature which doesn't have to deal with *interleaving* and *connection lifetime*.
Where you are now putting the user-level copy (under appropriate locking), we can check the lower-level feature and use that instead if available.

benaadams:
Rebased, reworked

benaadams:
What a beautiful stack trace
```
Failed   RegisterIPEndPoint_DynamicPort_Success(endPoint: [::1]:0, testUrl: "http://[::1]")
Error Message:
 System.Net.Sockets.SocketException : Operation timed out
Stack Trace:
   at Microsoft.AspNetCore.Testing.HttpClientSlim.GetSocket(Uri requestUri)
   at Microsoft.AspNetCore.Testing.HttpClientSlim.GetStream(Uri requestUri, Boolean validateCertificate)
   at Microsoft.AspNetCore.Testing.HttpClientSlim.GetStringAsync(Uri requestUri, Boolean validateCertificate)
   at Microsoft.AspNetCore.Testing.HttpClientSlim.GetStringAsync(String requestUri, Boolean validateCertificate)
   at Microsoft.AspNetCore.Server.Kestrel.FunctionalTests.AddressRegistrationTests.RegisterIPEndPoint_Success(IPEndPoint endPoint, String testUrl, Int32 testPort) in /_/test/Kestrel.FunctionalTests/AddressRegistrationTests.cs:line 287
   at Microsoft.AspNetCore.Server.Kestrel.FunctionalTests.AddressRegistrationTests.RegisterIPEndPoint_DynamicPort_Success(IPEndPoint endPoint, String testUrl) in /_/test/Kestrel.FunctionalTests/AddressRegistrationTests.cs:line 104
```

davidfowl:
@benaadams this change needs tests. Lots of tests. And we need to make sure that using the semaphore didn't regress performance of normal WriteAsync code paths.

benaadams:
Happy to pass it over 😄 

Was mostly a PoC and an example of where WriteAsync method type would be used 😉

src/Kestrel.Core/Internal/Http/Http1OutputProducer.cs
```diff
@@ -102,13 +131,72 @@ public Task WriteAsync<T>(Action<PipeWriter, T> callback, T state)
...
+                        buffer.Commit();
+                    }
+                }
+                finally
+                {
+                    _contextSemaphore.Release();
+                }
+
+                // WriteAsync() - not forced flush here
+                await buffer.FlushAsync(cancellationToken);
```
benaadams: Here, would prefer a 
```
await buffer.WriteAsync(cancellationToken);
```

benaadams:
MVC is similar, ditch its buffering TextWritter (sitting on array pool arrays) and have it use a TextWriter that writes straight through to the Response.Stream onto to the pipe (holding back a char or two for the encoder); which will be far more efficient.

However, that will only not degrade performance for MVC if every write is not an automatic flush (like behaviour of its buffering TextWriter); then the Transport can be deciding when it flushes on a WriteAsync as is suitable for packet stuffing and data density on the wire. 

When no more input; either end of request (or for more fancy, no more input requests); then the top level e.g. parser does a Flush which clears out every layer through to the transport which puts everything remaining on the wire.

src/Kestrel.Core/Internal/Http/HttpProtocol.SendFileFeature.cs
```diff
@@ -0,0 +1,38 @@
...
+            if (contentLength.HasValue && count.HasValue && contentLength.Value != count.Value)
+            {
+                // throw
+            }
+
+            using (var fileStream = new FileStream(path,
+                                 FileMode.Open,
+                                 FileAccess.Read,
+                                 FileShare.ReadWrite,
+                                 8 * 4096, // Buffer size - not currently pooled
```
davidfowl: We should use a super small buffer size here so we can avoid using the internal FileStreamBuffer.

benaadams: Is zero valid? 😄

davidfowl: Nope 😄

stephentoub: The checks are for <, so if you use a side of 1, the internal buffer won't be created.

benaadams:
Might be able to copypasta some of the StaticFiles tests?

Tratcher:
Or you could borrow tests from HttpSys

src/Kestrel.Core/Internal/Http/HttpProtocol.SendFileFeature.cs
benaadams: 1 it is :)

benaadams:
OSX
```
Failed   NonListenerPipeConnectionsAreLoggedAndIgnored
Error Message:
 System.IO.IOException : Unable to transfer data on the transport connection: Connection reset by peer.
---- System.Net.Sockets.SocketException : Connection reset by peer
Stack Trace:
   at System.IO.StreamReader.ReadBufferAsync()
   at System.IO.StreamReader.ReadToEndAsyncInternal()
   at Microsoft.AspNetCore.Testing.HttpClientSlim.ReadResponse(Stream stream)
   at Microsoft.AspNetCore.Testing.HttpClientSlim.GetStringAsync(Uri requestUri, Boolean validateCertificate)
   at Microsoft.AspNetCore.Server.Kestrel.Transport.Libuv.Tests.ListenerPrimaryTests.AssertResponseEventually(Uri address, String expected, String[] allowed, Int32 maxRetries, Int32 retryDelay) in /_/test/Kestrel.Transport.Libuv.Tests/ListenerPrimaryTests.cs:line 281
   at Microsoft.AspNetCore.Server.Kestrel.Transport.Libuv.Tests.ListenerPrimaryTests.NonListenerPipeConnectionsAreLoggedAndIgnored() in /_/test/Kestrel.Transport.Libuv.Tests/ListenerPrimaryTests.cs:line 139
----- Inner Stack Trace -----
```

src/Kestrel.Core/Internal/Http/HttpProtocol.SendFileFeature.cs
```diff
@@ -0,0 +1,38 @@
...
+using System.IO;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Http.Features;
+
+namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
+{
+    public partial class HttpProtocol : IHttpSendFileFeature
+    {
+        async Task IHttpSendFileFeature.SendFileAsync(string path, long offset, long? count, CancellationToken cancellationToken)
```
Tratcher: Does this trigger on-first-write activities like Response.OnStarting callbacks, header finalization, etc? It Should.

muratg:
We won't be doing this in 2.1, so closing.

PR #2064, opened 2017-09-18
Fix flakiness in KeepAliveTimeoutTests.ConnectionKeptAliveBetweenRequests (#1684)

cesarbs:
#1684

[`Task.Delay` is unreliable.](https://github.com/aspnet/KestrelHttpServer/issues/1684#issuecomment-330285740)

### DIFF
--- a/test/Kestrel.FunctionalTests/KeepAliveTimeoutTests.cs
+++ b/test/Kestrel.FunctionalTests/KeepAliveTimeoutTests.cs
@@ -77,6 +77,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.FunctionalTests
                         "",
                         "");
 
+                    // Don't change this to Task.Delay. See https://github.com/aspnet/KestrelHttpServer/issues/1684#issuecomment-330285740.
                     Thread.Sleep(_shortDelay);
                 }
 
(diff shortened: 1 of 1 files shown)


mikeharder:
Would another option be to just use `Thread.Sleep()` instead of `Task.Delay()`?  I wonder if `Thread.Sleep()` is more reliable.

cesarbs:
`Thread.Sleep` would block a thread that could be used by other tests.

cesarbs:
Using `Thread.Sleep` as discussed.

test/Kestrel.FunctionalTests/KeepAliveTimeoutTests.cs
```diff
@@ -90,11 +77,7 @@ public Task TestKeepAliveTimeout()
                         "",
                         "");
 
-                    var stopWatch = new Stopwatch();
-                    stopWatch.Start();
-                    await Task.Delay(_shortDelay);
-                    stopWatch.Stop();
-                    _logger.LogDebug($"Short delay lasted {TimeSpan.FromMilliseconds(stopWatch.ElapsedMilliseconds).TotalSeconds} seconds instead of {_shortDelay.TotalSeconds} seconds.");
+                    Thread.Sleep(_shortDelay);
```
pakrym: Comment, so no one "fixes" it back

PR #2066, opened 2017-09-19
HTTP/2: fix reading DATA frames with maximum length

cesarbs:
Miscalculated lengths led to bugs.

### DIFF
--- a/src/Kestrel.Core/Internal/Http2/Http2FrameWriter.cs
+++ b/src/Kestrel.Core/Internal/Http2/Http2FrameWriter.cs
@@ -35,6 +35,11 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http2
         {
             lock (_writeLock)
             {
+                if (_completed)
+                {
+                    return;
+                }
+
                 _completed = true;
                 _outputReader.CancelPendingRead();
                 _outputWriter.Complete(ex);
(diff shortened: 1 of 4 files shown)


src/Kestrel.Core/Internal/Http2/Http2FrameWriter.cs
```diff
@@ -35,6 +35,11 @@ public void Abort(Exception ex)
         {
             lock (_writeLock)
             {
+                if (_completed)
```
lillo42: Is not better use double-checked locking ?
like this:

```
if(!_completed)
{
      lock(_writeLock)
      {
            if(!_completed)
            {
                    _outputReader.CancelPendingRead();
                    _outputWriter.Complete(ex);
                   _completed = true;
            }
      }
}
```

benaadams: Likely only ever called once if at all, so wouldn't make much difference?

benaadams: Likely only ever called once if at all, so wouldn't make much difference?

Also the second call would complete before its finished as it wouldn't wait for the lock

lillo42: I don't know, I suggested that code because I saw lock and  double-checked locking can be used in this situation , but if is only ever called once maybe doesn't make much difference, OK

cesarbs: It's not on a hot path, so unlikely to benefit from double-checked locking.

cesarbs:
:rage1:

Issue #2067, opened 2017-09-19
HTTP/2: add flag to configure an endpoint as HTTP/2-only


cesarbs:
Support HTTP/2 with prior knowledge per [Section 3.4](http://httpwg.org/specs/rfc7540.html#known-http) of RFC 7540. This can be done per-endpoint.

Tratcher:
Your link is broken.

So you'd end up with some sort of protocol negotiation enum like HTTP1, HTTP2, HTTP1 | HTTP2... that you could translate to ALPN values.

cesarbs:
Fixed the link.

No, the spec allows for direct HTTP/2 cleartext connections. There's no protocol negotiation in that case.

Tratcher:
Sure, the same enum could be used for plaintext too, though the combine values wouldn't be valid.

cesarbs:
@muratg I wouldn't call this a bug. It's more like an enhancement, or even a task (since we haven't shipped HTTP/2 support yet, this would look a bit odd showing up in the release notes - just HTTP/2 should suffice).

PR #2068, opened 2017-09-19
Configure enabled protocols per endpoint and add HTTP/2 w/ prior knowledge support (#2067)

cesarbs:
#2067


cesarbs:
Still kind of a WIP - need to discuss the design and add more tests.

src/Kestrel.Core/EndPointProtocols.cs
```diff
@@ -0,0 +1,14 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+
+namespace Microsoft.AspNetCore.Server.Kestrel.Core
+{
+    [Flags]
+    public enum EndPointProtocols
```
davidfowl: HttpProtocols

Tratcher: Or just Protocols? In the specs they are ApplicationProtocols.

Tratcher: No combined value?

Tratcher: Flags enums usually have a default `None = 0` value.

anurse: Yeah, I'd add `None` and `Http1AndHttp2` (don't call it `Both`, as that means we're screwed if we need to add more later ;)). `None` is a little weird because if you specify it, we should throw out an error, but it can be useful for initialization, comparison, null values, etc.

src/Kestrel.Core/Internal/HttpConnection.cs
```diff
@@ -143,15 +144,30 @@ public Task StartRequestProcessing<TContext>(IHttpApplication<TContext> applicat
...
-                if (_http1Connection.ConnectionFeatures.Get<ITlsApplicationProtocolFeature>()?.ApplicationProtocol == "h2" &&
-                    Interlocked.CompareExchange(ref _http2ConnectionState, Http2ConnectionStarted, Http2ConnectionNotStarted) == Http2ConnectionNotStarted)
+                var hasTls = _context.ConnectionFeatures.Get<ITlsConnectionFeature>() != null;
+                var applicationProtocol = _context.ConnectionFeatures.Get<ITlsApplicationProtocolFeature>()?.ApplicationProtocol;
+                var http1Enabled = (_context.Protocols & EndPointProtocols.Http1) == EndPointProtocols.Http1;
+                var http2Enabled = (_context.Protocols & EndPointProtocols.Http2) == EndPointProtocols.Http2;
+
+                if (!hasTls && http1Enabled && http2Enabled)
+                {
+                    throw new InvalidOperationException("HTTP/1.x and HTTP/2 cannot both be served on the same cleartext endpoint.");
```
anurse: Suggest rephrase: "Using both HTTP/1.x and HTTP/2 on the same endpoint requires the use of TLS"

src/Kestrel.Core/Internal/HttpConnection.cs
```diff
@@ -143,15 +144,30 @@ public Task StartRequestProcessing<TContext>(IHttpApplication<TContext> applicat
...
+                }
+
+                var useHttp2 = http2Enabled && (!hasTls || applicationProtocol == "h2");
```
anurse: Should we use byte arrays for these values? As discussed with @Tratcher yesterday, ALPN just uses byte strings for protocols rather than specifying strings and encodings and such.

Tratcher: I'm not immediately concerned with this as long as it doesn't leak into public APIs. We'll have to refactor when we get the real ALPN support anyways.

anurse: `ITlsApplicationProtocolFeature` is (and will be released someday as) a public API. It's a generic ALPN feature for TLS connections.

src/Kestrel.Core/Internal/HttpConnection.cs
```diff
@@ -143,15 +144,30 @@ public Task StartRequestProcessing<TContext>(IHttpApplication<TContext> applicat
...
                     await _http2Connection.ProcessAsync(httpApplication);
                 }
-                else
+                else if (http1Enabled && (applicationProtocol == null || applicationProtocol == "http/1.1"))
                 {
                     await _http1Connection.ProcessRequestsAsync();
                 }
+                else
+                {
+                    throw new InvalidOperationException("Unable to select a protocol for the endpoint.");
```
anurse: I generally dislike "Unable to do thing" messages that don't say why. If a few more `if` clauses are needed to figure out a more specific error, I say we do that work so consumers don't have to.

src/Kestrel.Core/Internal/HttpConnection.cs
Tratcher: There's a new type coming that we can use for the datatype here so it doesn't have to be raw bytes.
https://github.com/dotnet/corefx/issues/23177 or we can go down one layer to `ReadOnlyMemory<byte>`.

anurse: I think we'd need to go to `ReadOnlyMemory<byte>`. We'd want this interface to be `netstandard` even though the implementation will be `netcoreapp2.x` until the SslStream features reach `netstandard`. So unless the exchange type is moved to a different `netstandard`-compatible assembly, I think we'll have to bypass it.

src/Kestrel.Core/EndPointProtocols.cs
cesarbs: I'm aware flags typically have a `None`, but it doesn't make sense here. Why add it just to check for it and throw? I don't see it being useful for the cases you pointed out (when would that happen?).

`Http1AndHttp2` is ugly but I'm willing to put it there. I don't see what's wrong with `|` though.

Tratcher: The default value for enum fields is 0, None provides a name for that value. That way it's more obvious when you're working with an uninitialized field and you don't have to check for 0 in code.

anurse: `Http1AndHttp2` is definitely ugly. I'm not entirely opposed to encouraging `|`, but most of our (the greater 'our', the one that includes .NET ;)) flags enums have common combinations as specific values and `Both` is limiting.

Regarding `None`, @Tratcher basically has the reasoning. Since you can cast any int to an enum and `0` is a common enough value for uninitialized enums/ints, it's very useful for debugging purposes to have a name for it. If you don't want an explicit error case, that's OK. I'm sure some kind of error would occur if neither flag was set. Specifying `None` can have the same error case as specifying `(HttpProtocol)42`, but having the name is a useful debugging error.

cesarbs:
I've addressed most of your feedback. I'm now working on setting up ALPN on `TlsStream` based on the value of `ListenOptions.Protocols`.

cesarbs:
Latest commit uses `ListenOptions.Protocols` to set up ALPN on `TlsStream`.

test/Kestrel.FunctionalTests/Http2Tests.cs
```diff
@@ -0,0 +1,42 @@
...
+using Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http2;
+using Microsoft.AspNetCore.Testing;
+using Xunit;
+
+namespace Microsoft.AspNetCore.Server.Kestrel.FunctionalTests
+{
+    public class Http2Tests
+    {
+        [Fact]
+        public async Task ServesHttp2WithPriorKnowledge()
```
Tratcher: No ALPN test? or do you not have an ALPN client?

src/Kestrel.Tls/TlsConnectionAdapter.cs
```diff
@@ -47,6 +47,17 @@ public TlsConnectionAdapter(TlsConnectionAdapterOptions options, ILoggerFactory
 
             _options = options;
             _logger = loggerFactory?.CreateLogger(nameof(TlsConnectionAdapter));
+
+            // Order is important. If HTTP/2 is enabled, we prefer it over HTTP/1.1. So add it first.
+            if ((options.Protocols & HttpProtocols.Http2) == HttpProtocols.Http2)
```
Tratcher: You could #ifdef this section in preparation for the new core API. Then you'd be able to test the net46 behavior.

test/Kestrel.FunctionalTests/Http2Tests.cs
cesarbs: Currently working on more tests. Not sure I'll be able to final an ALPN client.

cesarbs:
I've renamed `Http2Tests` to `HttpProtocolSelectionTests`. For now there are only cleartext tests there since I couldn't make `HttpClient` use HTTP/2 over TLS on neither Windows nor Linux. We can add TLS tests once we have ALPN on `SslStream`.

I've added `ListenOptionsTests` with a test to check that endpoints are HTTP/1.x-only by default.

src/Kestrel.Core/Internal/HttpConnection.cs
```diff
@@ -170,17 +170,15 @@ public Task StartRequestProcessing<TContext>(IHttpApplication<TContext> applicat
...
-                throw;
-            }
             catch (Exception ex)
             {
                 Log.LogError(0, ex, $"Unexpected exception in {nameof(HttpConnection)}.{nameof(ProcessRequestsAsync)}.");
             }
             finally
             {
+                // Complete the output in case an exception was thrown before request processing could start.
+                _context.Transport.Output.Complete();
```
cesarbs: @davidfowl I'd like your input here.

src/Kestrel.Tls/TlsConnectionAdapter.cs
cesarbs: Since I can't properly test with ALPN now, I'll do that in a future change.

cesarbs:
Ping.

src/Kestrel.Core/Internal/HttpConnectionBuilderExtensions.cs
```diff
@@ -11,12 +11,22 @@ public static class HttpConnectionBuilderExtensions
     {
         public static IConnectionBuilder UseHttpServer<TContext>(this IConnectionBuilder builder, ServiceContext serviceContext, IHttpApplication<TContext> application)
         {
-            return builder.UseHttpServer(Array.Empty<IConnectionAdapter>(), serviceContext, application);
+            return builder.UseHttpServer(serviceContext, application, protocols: HttpProtocols.Http1);
```
Tratcher: why the named parameters here?

cesarbs: Talked to @davidfowl. I'm removing those extension methods and leaving only the new ones (with the new `HttpProtocols` param).

src/Kestrel.Core/Internal/HttpConnection.cs
```diff
@@ -290,16 +297,50 @@ private void DisposeAdaptedConnections()
...
+
+            if (!http1Enabled && http2Enabled && hasTls && applicationProtocol != "h2")
+            {
+                error = CoreStrings.EndPointHttp2NotNegotiated;
+            }
+
+            if (error != null)
+            {
+                Log.LogError(0, error);
+                Abort(ex: null);
```
davidfowl: SIDE EFFECTS! Do this outside.

PR #2072, opened 2017-09-21
HTTP/2: connection error when receiving frames disallowed by stream states

cesarbs:
http://httpwg.org/specs/rfc7540.html#rfc.section.5.1

> idle:
> ...
> Receiving any frame other than HEADERS or PRIORITY on a stream in this state MUST be treated as a connection error (Section 5.4.1) of type PROTOCOL_ERROR.

### DIFF
--- a/src/Kestrel.Core/Internal/Http2/Http2Connection.cs
+++ b/src/Kestrel.Core/Internal/Http2/Http2Connection.cs
@@ -249,16 +249,35 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http2
 
             ThrowIfIncomingFrameSentToIdleStream();
 
-            if (_streams.TryGetValue(_incomingFrame.StreamId, out var stream) && !stream.HasReceivedEndStream)
+            if (_streams.TryGetValue(_incomingFrame.StreamId, out var stream) && !stream.EndStreamReceived)
             {
                 return stream.OnDataAsync(_incomingFrame.DataPayload,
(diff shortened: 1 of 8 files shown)


src/Kestrel.Core/Internal/Http2/Http2Connection.cs
```diff
@@ -238,6 +238,11 @@ private Task ProcessDataFrameAsync()
                 throw new Http2ConnectionErrorException(Http2ErrorCode.PROTOCOL_ERROR);
             }
 
+            if (_incomingFrame.StreamId > _highestOpenedStreamId)
```
Tratcher: This one may need some explaining.

cesarbs: Added comments and moved the check to a separate method, since it applies to three different frame types.

cesarbs:
Ping.

cesarbs:
Ping.

cesarbs:
Just added more stuff here. Instead of handling just the particular scenario in the first commit, I'm now handling all errors described in http://httpwg.org/specs/rfc7540.html#rfc.section.5.1.

src/Kestrel.Core/Internal/Http2/Http2MessageBody.cs
```diff
@@ -29,9 +29,9 @@ public override Task StopAsync()
             HttpRequestHeaders headers,
             Http2Stream context)
         {
-            if (!context.ExpectData)
```
halter73: Is ExpectData still used anywhere?

cesarbs: No. Good catch 👍

PR #2073, opened 2017-09-21
HTTP/2: do not ACK PING frames received with ACK set

cesarbs:
http://httpwg.org/specs/rfc7540.html#rfc.section.6.7

> The PING frame defines the following flags:
> 
> ACK (0x1): When set, bit 0 indicates that this PING frame is a PING response. An endpoint MUST set this flag in PING responses. **An endpoint MUST NOT respond to PING frames containing this flag.**


cesarbs:
Ping.

PR #2077, opened 2017-09-21
HTTP/2: close connection with PROTOCOL_ERROR when a stream depends on itself

cesarbs:
http://httpwg.org/specs/rfc7540.html#rfc.section.5.3.1

> A stream cannot depend on itself. An endpoint MUST treat this as a stream error (Section 5.4.2) of type PROTOCOL_ERROR.

### DIFF
--- a/src/Kestrel.Core/Internal/Http2/Http2Connection.cs
+++ b/src/Kestrel.Core/Internal/Http2/Http2Connection.cs
@@ -3,7 +3,6 @@
 
 using System;
 using System.Collections.Concurrent;
-using System.Collections.Generic;
 using System.IO.Pipelines;
 using System.Text;
 using System.Threading.Tasks;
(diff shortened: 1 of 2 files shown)


src/Kestrel.Core/Internal/Http2/Http2Connection.cs
```diff
@@ -264,6 +262,11 @@ private Task ProcessHeadersFrameAsync<TContext>(IHttpApplication<TContext> appli
                 throw new Http2ConnectionErrorException(Http2ErrorCode.PROTOCOL_ERROR);
             }
 
+            if (_incomingFrame.HeadersHasPriority && _incomingFrame.HeadersStreamDependency == _incomingFrame.StreamId)
```
Tratcher: What's the relation between priority and dependency?

cesarbs: Priority tells the server how much of its resources should be dedicated to a given stream. Dependencies tell the server what streams to work on first.

cesarbs: And here `HeadersHasPriority` checks for the priority flag on the frame. When its set on a HEADERS frame, that means the frame contains both priority and stream dependency on it (you can't send just one of those values).

PR #2078, opened 2017-09-21
HTTP/2: close connection with PROTOCOL_ERROR when receiving a frame with an even stream ID

cesarbs:
http://httpwg.org/specs/rfc7540.html#rfc.section.5.1.1

> Streams initiated by a client MUST use odd-numbered stream identifiers;

> An endpoint that receives an unexpected stream identifier MUST respond with a connection error (Section 5.4.1) of type PROTOCOL_ERROR.

### DIFF
--- a/src/Kestrel.Core/Internal/Http2/Http2Connection.cs
+++ b/src/Kestrel.Core/Internal/Http2/Http2Connection.cs
@@ -196,6 +196,11 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http2
 
         private Task ProcessFrameAsync<TContext>(IHttpApplication<TContext> application)
         {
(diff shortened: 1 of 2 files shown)


src/Kestrel.Core/Internal/Http2/Http2Connection.cs
```diff
@@ -196,6 +196,11 @@ private bool ParsePreface(ReadableBuffer readableBuffer, out ReadCursor consumed
 
         private Task ProcessFrameAsync<TContext>(IHttpApplication<TContext> application)
         {
+            if (_incomingFrame.StreamId != 0 && (_incomingFrame.StreamId & 1) == 0)
```
Tratcher: This one needs a comment. It's ok to quote the spec. Even better if you link to the section.

PR #2079, opened 2017-09-21
HTTP/2: close connection with PROTOCOL_ERROR when receiving GOAWAY frame with non-zero stream ID

cesarbs:
http://httpwg.org/specs/rfc7540.html#rfc.section.6.8

> An endpoint MUST treat a GOAWAY frame with a stream identifier other than 0x0 as a connection error (Section 5.4.1) of type PROTOCOL_ERROR.

### DIFF
--- a/src/Kestrel.Core/Internal/Http2/Http2Connection.cs
+++ b/src/Kestrel.Core/Internal/Http2/Http2Connection.cs
@@ -406,6 +406,11 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http2
                 throw new Http2ConnectionErrorException(Http2ErrorCode.PROTOCOL_ERROR);
             }
 
+            if (_incomingFrame.StreamId != 0)
+            {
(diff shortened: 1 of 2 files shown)


src/Kestrel.Core/Internal/Http2/Http2Connection.cs
```diff
@@ -399,6 +399,11 @@ private Task ProcessGoAwayFrameAsync()
                 throw new Http2ConnectionErrorException(Http2ErrorCode.PROTOCOL_ERROR);
             }
 
+            if (_incomingFrame.StreamId != 0)
+            {
+                throw new Http2ConnectionErrorException(Http2ErrorCode.PROTOCOL_ERROR);
```
Tratcher: For our logging purposes we should include messages describing these errors.

cesarbs: Yes, I will revisit that (#2080).

PR #2081, opened 2017-09-21
Remove explicit LangVersion

Tratcher:
This is now set globally to 7.1: aspnet/BuildTools#408


natemcmaster:
This repo has a custom version of roslyn. I think LangVersion needs to be 7.2 cc @pakrym

pakrym:
@natemcmaster is right, Kestrel needs 7.2 features (Span support)

PR #2084, opened 2017-09-27
HPACK fixes and improvements

cesarbs:
* Tests for all header field representations and dynamic table updates (http://httpwg.org/specs/rfc7541.html#rfc.section.6)
* Throw on error conditions
  * Incomplete header block (http://httpwg.org/specs/rfc7540.html#rfc.section.4.3)
  * Invalid header index
  * String length larger than internal buffer
  * Dynamic table update to a size larger than the maximum allowed by server settings (http://httpwg.org/specs/rfc7541.html#rfc.section.6.3)
  * Huffman decoding errors (read last two paragraphs of http://httpwg.org/specs/rfc7541.html#rfc.section.5.2)
    * Padding longer than 7 bits
    * Incomplete symbols
    * EOS


src/Kestrel.Core/Internal/Http2/HPack/HPackDecoder.cs
```diff
@@ -265,13 +325,36 @@ private void OnStringLength(int length, State nextState)
...
+                    return Huffman.Decode(_stringOctets, 0, _stringLength);
+                }
+                catch (HuffmanDecodingException ex)
+                {
+                    throw new HPackDecodingException(CoreStrings.HPackHuffmanError, ex);
+                }
+            }
+            else
+            {
+                return Encoding.ASCII.GetString(_stringOctets, 0, _stringLength);
```
davidfowl: Don't we have a faster version of this?

src/Kestrel.Core/Internal/Http2/HPack/HeaderField.cs
```diff
@@ -12,6 +12,9 @@ public HeaderField(string name, string value)
         }
 
         public string Name { get; }
+
         public string Value { get; }
+
+        public int Length => Name.Length + Value.Length + 32;
```
davidfowl: 32?

cesarbs: http://httpwg.org/specs/rfc7541.html#rfc.section.4.1

davidfowl: Can you generally put more comments in the code that are related to the spec.

src/Kestrel.Core/Internal/Http2/HPack/HeaderField.cs
```diff
@@ -1,17 +1,26 @@
...
     {
-        public HeaderField(string name, string value)
+        public HeaderField(Span<byte> name, Span<byte> value)
         {
-            Name = name;
-            Value = value;
+            Name = new byte[name.Length];
+            name.CopyTo(Name);
+
+            Value = new byte[value.Length];
```
halter73: Can these arrays be pooled?

src/Kestrel.Core/Internal/Http2/Http2Connection.cs
```diff
@@ -141,6 +141,12 @@ public void Stop()
                 error = ex;
                 errorCode = ex.ErrorCode;
             }
+            catch (HPackDecodingException ex)
```
halter73: Does every HPackDecodingException that gets thrown get caught here? If so, you might as well log all the new exception messages. Should be easy, right?

src/Kestrel.Core/Internal/Http2/HPack/DynamicTable.cs
```diff
@@ -37,33 +39,48 @@ public DynamicTable(int maxSize)
...
+            _size += entry.Length;
             _count++;
         }
 
         public void Resize(int maxSize)
         {
             _maxSize = maxSize;
-            EnsureSize(_maxSize);
+
+            if (maxSize > _maxSize)
```
halter73: Isn't this condition always false b/c of `_maxSize = maxSize` above?

src/Kestrel.Core/Internal/Http2/HPack/DynamicTable.cs
```diff
@@ -37,33 +39,48 @@ public DynamicTable(int maxSize)
...
+                _buffer = newBuffer;
+            }
+            else
+            {
+                EnsureSize(_maxSize);
+            }
         }
 
-        public void EnsureSize(int size)
+        private void EnsureSize(int size)
```
halter73: So this ensures the total size of the entries in the dynamic table is *no more* than `size`? Wouldn't it be easier to change this method to `EnsureFree`?

Then `EnsureSize(_maxSize - entryLength)` would become `EnsureFree(entryLength)` and `EnsureSize(_maxSize)` would become `EnsureFree(0)`.  I'm still not sure what the `EnsureSize(_maxSize)` is for.

src/Kestrel.Core/Internal/Http2/HPack/DynamicTable.cs
```diff
@@ -37,33 +39,48 @@ public DynamicTable(int maxSize)
             }
         }
 
-        public void Insert(string name, string value)
+        public void Insert(Span<byte> name, Span<byte> value)
         {
-            var entrySize = name.Length + value.Length + 32;
-            EnsureSize(_maxSize - entrySize);
+            var entryLength = (name.Length + value.Length + 32);
```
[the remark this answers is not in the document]
halter73: Maybe this should be a static method on HeaderField since this math is repeated in a few places.

src/Kestrel.Core/Internal/Http2/HPack/DynamicTable.cs
```diff
@@ -7,23 +7,25 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http2.HPack
...
+        private int _maxSize;
         private int _size;
         private int _count;
         private int _insertIndex;
         private int _removeIndex;
 
         public DynamicTable(int maxSize)
         {
-            _buffer = new HeaderField[maxSize];
+            _buffer = new HeaderField[maxSize / 32];
```
halter73: The header field names and values need to be at least one octet too, right? Could you divide by 34?

cesarbs: The value may be empty.

src/Kestrel.Core/Internal/Http2/HPack/DynamicTable.cs
cesarbs: Lol yeah.

src/Kestrel.Core/Internal/Http2/Http2Connection.cs
cesarbs: There are TODOs for logs in a bunch of places. I'm going to address those as a separate change (need to define log messages for all the possible connection errors, etc.).

Issue #1290, opened 2017-01-07
Make it easier to configure Kestrel endpoints from config


Tratcher:
https://github.com/aspnet/KestrelHttpServer/pull/1280 added new KestrelServerOptions.Listen APIs for directly configuring kestrel to listen on IPs and ports. However, it's non-trivial to parse these values from config so they can be passed to the new methods. @shirhatti @DamianEdwards 

Basic scenarios:
1. http://localhost:port/ (IPv4 and IPv6 loopback, ANCM)
2. http://*:port/ (IPv6Any)
3. http://IP:port/
4. https://*:port/ with given SSL cert & password
5. https://IP:port/ with given SSL cert & password
6. 2 and 4 at the same time.
7. A list of IPs and ports

Update the samples to use values passed in via config rather than hardcoded values.

Current API/Sample:
```
options.Listen(IPAddress.Loopback, 5000);
options.Listen(IPAddress.Loopback, 5001, listenOptions =>
{
       listenOptions.UseHttps("testCert.pfx", "testPassword");
});
```

Proposals:
- It's easy enough for the user to infer the scheme from the presence of cert information.
  - `if (!string.IsNullOrEmpty(config["certPath"])) { listenOptions.UseHttps(config["certPath"], config["certPassword"]); }`
- If the app only expects to be loopback or *, they only need to pass in a port. Adding Loopback and Any versions with string overloads that parse the port would help.
  - `options.ListenLoopback(int.Parse(config["port"], CultureInfo.InvariantCulture));` vs `options.ListenLoopback(config["port"]);`
  - `options.ListenAny(int.Parse(config["port"], CultureInfo.InvariantCulture));` vs `options.ListenAny(config["port"]);`
- For an IP and port, IPEndpoint does not have a Parse method. If we don't provide one then the config needs to store the parameters separately.
  - `options.Listen(IPAddress.Parse(config["IP"]), int.Parse(config["port"], CultureInfo.InvariantCulture));` vs `options.Listen(config["IPandPort"]);`
- Most of the above do not scale well if you need to pass in multiple values, especially if the count varies.
  - `options.Listen(config["IPandPort1"]); options.Listen(config["IPandPort2"]); options.Listen(config["IPandPort3"]);`
  - The user may be able to do a string split on semicolon and call Listen in a loop.
  - `config["IPsAndPorts"].Split(';').ForEach(endpoint => options.Listen(endpoint));`

muratg:
[...]

PR #2093, opened 2017-10-03
Add "zero config" HTTPS support using local development certificate.

cesarbs:
This change enables automatic HTTPS using a development certificate found in the user store. Any URLs containing the https scheme will be configured with that certificate.

Paging @Eilon since this is a breaking change (previously we would throw upon seeing an https URL).

So now if you set `ASPNETCORE_URLS` (or use `UseUrls()`) to e.g. `https://localhost:5001`, you'll get HTTPS if you add Kestrel with no listen options i.e.

```csharp
var host = new WebHostBuilder()
    ...
    .UseKestrel()
    ...
```

This turned out to be more convoluted than I anticipated. I'm hoping someone can suggest a simpler way to achieve the same result.


src/Kestrel/DefaultHttpsProvider.cs
```diff
@@ -0,0 +1,63 @@
...
+        }
+
+        private static X509Certificate2 FindDevelopmentCertificate()
+        {
+            using (var store = new X509Store(StoreName.My, StoreLocation.CurrentUser))
+            {
+                store.Open(OpenFlags.ReadOnly);
+
+                var certificates = store.Certificates.OfType<X509Certificate2>();
+                var certificate = certificates
```
javiercn: You need to check that it has an OID, a private key and hasn't expired

javiercn:
> Paging @Eilon since this is a breaking change (previously we would throw upon seeing an https URL).

I don't see this as a breaking change. At most a behavior change, and you are weakening the precondition (allowing something that wasn't allowed before) so I don't think this is a breaking change.

muratg:
To me this looks like a minor breaking change, but an acceptable one.  If you specify https for your URL, your app would fail to start before this. Now it'll do what you (presumably) wanted anyway.

src/Kestrel.Core/KestrelServer.cs
```diff
@@ -52,6 +54,8 @@ internal KestrelServer(ITransportFactory transportFactory, ServiceContext servic
             Features = new FeatureCollection();
             _serverAddresses = new ServerAddressesFeature();
             Features.Set(_serverAddresses);
+
+            _defaultHttpsProvider = serviceContext.ServerOptions.ApplicationServices.GetService<IDefaultHttpsProvider>();
```
halter73: Don't call GetService for this. Just make IDefaultHttpsProvider a param to the public ctor.

src/Kestrel.Core/Internal/AddressBinder.cs
```diff
@@ -19,7 +19,9 @@ internal class AddressBinder
     {
         public static async Task BindAsync(IServerAddressesFeature addresses,
             List<ListenOptions> listenOptions,
+            KestrelServerOptions serverOptions,
```
halter73: It doesn't look like this is actually used by the DefaultHttpsProvider or anything else. I see that you are setting this on ListenOptions before calling ConfigureHttps, but I don't see why that's necessary. Save some lines of code and don't bother plumbing this through here.

src/Kestrel/DefaultHttpsProvider.cs
cesarbs: When do you think you will have https://github.com/aspnet/Common/pull/263 merged? We shouldn't replicate all that logic here. I should be able to get the certificate by calling `CertificateManager.ListCertificates(CertificatePurpose.HTTPS, StoreName.My, StoreLocation.CurrentUser, isValid: true)`, right?

cesarbs:
@halter73 To get the test cert set up on your machine, clone this PR https://github.com/aspnet/Common/pull/263/files#diff-1f595d265c92ca9d7a059ed3d0d223fbR33 and run the tests. @javiercn can give you more info.

src/Kestrel.Core/Internal/AddressBinder.cs
cesarbs: It blows up here: https://github.com/aspnet/KestrelHttpServer/blob/83e208436858a2b435e7a9dd0f6ea228bc842e4b/src/Kestrel.Https/ListenOptionsHttpsExtensions.cs#L89

cesarbs: I may have misunderstood what you said. There's indeed no need to pass it as an arg to `BindAsync`, since I can get the `KestrelServerOptions` instance from one of the `ListenOptions` in the list. I still need that object though, otherwise a NRE is thrown at the line above.

halter73: You're right. I forgot about the call to UseHttps().

cesarbs:
Injecting `IDefaultHttpsProvider` in the `KestrelServer` ctor is a breaking change - I had to add an exception for ApiCheck. However, given these are injected parameters, is that an acceptable breaking change for a minor version? @muratg @Eilon

halter73:
@cesarbs Couldn't you just add a second constructor that calls the original one with a null IDefaultHttpsProvider? Of course, you'd want to keep the old behavior if IDefaultHttpsProvider is null.

Eilon:
Yeah just do what @halter73 suggested and there's no breaking change at all in terms of source compat, binary compat, runtime behavior, etc. It's nearly always fine to throw *fewer* exceptions and be more permissive.

src/Kestrel.Core/Internal/AddressBinder.cs
```diff
@@ -31,7 +32,9 @@ internal class AddressBinder
             {
                 Addresses = addresses.Addresses,
                 ListenOptions = listenOptions,
+                ServerOptions = listenOptions.FirstOrDefault()?.KestrelServerOptions,
```
davidfowl: This is pretty gross

Tratcher:
New/updated sample?

src/Kestrel.Core/Internal/AddressBinder.cs
```diff
@@ -131,7 +137,14 @@ private static Task BindEndpointAsync(IPEndPoint endpoint, AddressBindContext co
...
             try
             {
-                await BindEndpointAsync(new IPEndPoint(IPAddress.Loopback, address.Port), context).ConfigureAwait(false);
+                var options = new ListenOptions(new IPEndPoint(IPAddress.Loopback, address.Port));
+                await BindEndpointAsync(options, context).ConfigureAwait(false);
+                
+                if (https)
+                {
+                    options.KestrelServerOptions = context.ServerOptions;
+                    context.DefaultHttpsProvider?.ConfigureHttps(options);
```
Tratcher: We want this to throw if DefaultHttpsProvider is null, no? You asked for https and you're not getting it.

src/Kestrel/DefaultHttpsProvider.cs
```diff
@@ -0,0 +1,74 @@
...
+                    .FirstOrDefault(c => HasOid(c, AspNetHttpsOid) && !IsExpired(c) /*&& HasPrivateKey(c)*/);
+
+                if (certificate == null)
+                {
+                    throw new InvalidOperationException("Unable to find ASP.NET Core development certificate.");
```
Tratcher: This level of magic deserves an FW link explaining what to do when it fails.

src/Kestrel/DefaultHttpsProvider.cs
```diff
@@ -0,0 +1,74 @@
...
+        {
+            // TODO: replace this with call to CertificateManager.FindCertificates(CertificatePurpose.HTTPS, StoreName.My, StoreLocation.CurrentUser, isValid: true)
+            // when that becomes available.
+            using (var store = new X509Store(StoreName.My, StoreLocation.CurrentUser))
+            {
+                store.Open(OpenFlags.ReadOnly);
+
+                var certificates = store.Certificates.OfType<X509Certificate2>();
+                var certificate = certificates
+                    .FirstOrDefault(c => HasOid(c, AspNetHttpsOid) && !IsExpired(c) /*&& HasPrivateKey(c)*/);
```
Eilon: Is the commented code needed?

cesarbs: Yes. Thanks for catching that, I forgot I had commented it out while testing something.

src/Kestrel/DefaultHttpsProvider.cs
cesarbs: @danroth27

src/Kestrel/DefaultHttpsProvider.cs
javiercn: I'm working on it

cesarbs:
Updated to use https://github.com/aspnet/Common/blob/b9bc076c782af4a9664b34b561fd91d37e6edda4/shared/Microsoft.AspNetCore.Certificates.Generation.Sources/CertificateManager.cs to find development certificate.

src/Kestrel/Kestrel.csproj
```diff
@@ -12,10 +12,13 @@
 
   <ItemGroup>
     <PackageReference Include="Microsoft.AspNetCore.Hosting" />
+    <PackageReference Include="Microsoft.AspNetCore.Certificates.Generation.Sources" PrivateAssets="All" />
+    <PackageReference Include="System.Security.Cryptography.Cng" />
```
davidfowl: What's this for?

javiercn: He’s using the same code we have in tooling to locate the certificate (that’s why it’s a shared source library)

cesarbs: We need that reference for the `RSACng` type.

cesarbs:
Rebased.

test/Kestrel.Core.Tests/KestrelServerTests.cs
```diff
@@ -37,20 +38,17 @@ public void StartWithInvalidAddressThrows()
...
                 server.Features.Get<IServerAddressesFeature>().Addresses.Add("https://127.0.0.1:0");
 
-                var exception = Assert.Throws<InvalidOperationException>(() => StartDummyApplication(server));
+                StartDummyApplication(server);
 
-                Assert.Equal(
-                    $"HTTPS endpoints can only be configured using {nameof(KestrelServerOptions)}.{nameof(KestrelServerOptions.Listen)}().",
-                    exception.Message);
-                Assert.Equal(1, testLogger.CriticalErrorsLogged);
+                mockDefaultHttpsProvider.Verify(provider => provider.ConfigureHttps(It.IsAny<ListenOptions>()), Times.Once);
```
anurse: Not a fan of tests that verify that a certain method was called. Is there a way to build a simple dummy DefaultHttpsProvider and then verify the certificate on the ListenOptions?

cesarbs: What is the issue with this kind of test? To test with `DefaultHttpsProvider`, I'd have to make the test project depend on the `Kestrel` project, instead of just `Kestrel.Core`.

anurse: Hmm. I'll take a bit of a look at it and see if we can make something work. Tests that use mock verification tend to be brittle whenever you change the implementation and they aren't actually testing the desired behavior of the function, they're just testing that the code does what is written. It's a little like a comment over `x += 1` that says `// Adds 1 to x` :). If that's all that we can do, then I suppose it's OK, but I'd like to try and avoid it.

anurse:
Any further comments beyond my own re:testing? I'm going to spend a few quick cycles seeing if I can tidy up the test, but otherwise it looks OK to me.

javiercn:
I'm no expert but it looks good to me.

Tratcher:
I'm wondering how well this will generalize when we implement the config file scenarios. E.g. If you specify a default cert in config that should take precedence over the dev one from the cert store but still work with https entries in URLs.
Devs also want the top-level UseHttps back where they can configure it in code, which could plug into this HttpsProvider somehow. It's not clear if the new dev cert and config file approaches will cover this need.

anurse:
:up: :date: I gave up with the testing refinement. But I did notice that if for some reason DefaultHttpsProvider isn't provided (because it is an **optional** dependency specified in a separate ctor, even if `.UseKestrel` adds it by default), the user gets a rando null-ref exception. So, I added a few tests and some logic to ensure that when there isn't an `IDefaultHttpsProvider` you get a nice error in the right place.

anurse:
> Devs also want the top-level UseHttps back where they can configure it in code, which could plug into this HttpsProvider somehow

Agree, but I think that's something we can iterate on when we have the cert config story worked out. Is there an issue tracking that?

test/Kestrel.FunctionalTests/TestHelpers/TestServer.cs
```diff
@@ -11,10 +11,12 @@
 using Microsoft.AspNetCore.Hosting.Server;
 using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.Server.Kestrel.Core;
+using Microsoft.AspNetCore.Server.Kestrel.Core.Internal;
 using Microsoft.AspNetCore.Server.Kestrel.Transport.Abstractions.Internal;
 using Microsoft.AspNetCore.Testing;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Options;
+using Moq;
```
halter73: Nit: fix usings.

src/Kestrel/Internal/DefaultHttpsProvider.cs
```diff
@@ -0,0 +1,77 @@
...
+        private static void DisposeCertificates(IEnumerable<X509Certificate2> certificates)
+        {
+            foreach (var certificate in certificates)
+            {
+                try
+                {
+                    certificate.Dispose();
+                }
+                catch
+                {
```
halter73: Why no logging here?

src/Kestrel/Internal/DefaultHttpsProvider.cs
```diff
@@ -0,0 +1,77 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Security.Cryptography.X509Certificates;
+using Microsoft.AspNetCore.Hosting;
+using Microsoft.AspNetCore.Server.Kestrel.Core;
+using Microsoft.AspNetCore.Server.Kestrel.Core.Internal;
+
+namespace Microsoft.AspNetCore.Server.Kestrel.Internal
+{
+    public class DefaultHttpsProvider : IDefaultHttpsProvider
```
halter73: Why two classes with the same name in the same assembly? They look almost (but not exactly) the same.

Tratcher:
https://github.com/aspnet/KestrelHttpServer/issues/1290

src/Kestrel/Internal/DefaultHttpsProvider.cs
```diff
@@ -0,0 +1,77 @@
...
+        {
+            listenOptions.UseHttps(DefaultCertificateResolver());
+        }
+
+        private static X509Certificate2 FindDevelopmentCertificate()
+        {
+            // TODO: replace this with call to
+            // CertificateManager.FindCertificates(CertificatePurpose.HTTPS, StoreName.My, StoreLocation.CurrentUser, isValid: true)
+            // when that becomes available.
+            using (var store = new X509Store(StoreName.My, StoreLocation.CurrentUser))
```
cesarbs: Weird, I was sure I had pushed the change that removed all this logic and just called [`CertificateManager.ListCertificate`](https://github.com/aspnet/Common/blob/9154a85af62b4e461733f32e062498140ecd7cad/shared/Microsoft.AspNetCore.Certificates.Generation.Sources/CertificateManager.cs#L42). You can just call that like this:

```csharp
CertificateManager.ListCertificates(CertificatePurpose.HTTPS, StoreName.My, StoreLocation.CurrentUser, isValid: true, requireExportable: false);
```

anurse: It's possible something blew up during a rebase. I'll bring it back in :).

src/Kestrel/Internal/DefaultHttpsProvider.cs
anurse: Yeah it does look like a perfect place for some logs. I'll add some.

anurse:
:up: :date: Using CertificateManager shared-source package now. Also, added some more logging and exception messages.

src/Kestrel/KestrelStrings.resx
```diff
@@ -0,0 +1,123 @@
...
+    <value>2.0</value>
+  </resheader>
+  <resheader name="reader">
+    <value>System.Resources.ResXResourceReader, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
+  </resheader>
+  <resheader name="writer">
+    <value>System.Resources.ResXResourceWriter, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
+  </resheader>
+  <data name="HttpsUrlProvidedButNoDevelopmentCertificateFound" xml:space="preserve">
+    <value>An 'https' URL was provided, but a development certificate could not be found.</value>
```
javiercn: Can you add something along the lines of the text that I sent you on email and add the forwarding link that @danroth27  provided?

src/Kestrel/Internal/DefaultHttpsProvider.cs
```diff
@@ -0,0 +1,72 @@
...
+
+        public DefaultHttpsProvider(ILogger<DefaultHttpsProvider> logger)
+        {
+            _logger = logger;
+            DefaultCertificateResolver = FindDevelopmentCertificate;
+        }
+
+        public void ConfigureHttps(ListenOptions listenOptions)
+        {
+            var cert = DefaultCertificateResolver();
```
halter73: Why the Func? Was this replaced in a test at one point? It seems like you could just call FindDevelopmentCertificate directly.

src/Kestrel/Internal/DefaultHttpsProvider.cs
```diff
@@ -0,0 +1,72 @@
...
+            listenOptions.UseHttps(cert);
+        }
+
+        private X509Certificate2 FindDevelopmentCertificate()
+        {
+            var certificate = _certificateManager.ListCertificates(CertificatePurpose.HTTPS, StoreName.My, StoreLocation.CurrentUser, isValid: true)
+                .FirstOrDefault();
+            if (certificate != null)
+            {
+                _logger.LogDebug("Using development certificate: {certificateSubjectName} (Thumbprint: {certificateThumbprint})", certificate.Subject, certificate.Thumbprint);
```
halter73: For common/expected logs, I think it's best to use resource strings.

src/Kestrel/Internal/DefaultHttpsProvider.cs
```diff
@@ -0,0 +1,72 @@
...
+                return certificate;
+            }
+            else
+            {
+                _logger.LogDebug("Development certificate could not be found");
```
halter73: Should this be given a higher log level? I'm assuming this would normally cause the application to crash on startup, but in the event the exception is somehow swallowed, I think it's a good idea to log this as an error.

src/Kestrel/Internal/DefaultHttpsProvider.cs
anurse: Yeah, I was wondering that too. I suppose so that you could construct it and replace the func, but that doesn't really make sense now that it's pubternal rather than public. It's also a simple enough interface to implement. I'll yank the func.

src/Kestrel/Internal/DefaultHttpsProvider.cs
anurse: Hmm, we use literals for all the other logs (see KestrelTrace) so I wasn't sure exactly what the right pattern was here. I thought about adding some `LoggerMessage.Define` action here but even with that, I've not seen our code use resource strings for that. I guess we could though.

src/Kestrel/Internal/DefaultHttpsProvider.cs
anurse: Well, an exception will be thrown by `ConfigureHttps`, which calls this. But when refactoring to remove the delegate, I'll probably just do exactly as you suggest here :).

test/Kestrel.Core.Tests/KestrelServerTests.cs
```diff
@@ -37,20 +38,60 @@ public void StartWithInvalidAddressThrows()
...
+            }
+        }
+ 
+        [Fact]
+        public void KestrelServerThrowsUsefulExceptionIfDefaultHttpsProviderNotAdded()
+        {
+            var mockDefaultHttpsProvider = new Mock<IDefaultHttpsProvider>();
```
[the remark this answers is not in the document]
anurse: yep, copy🍝

test/Kestrel.Core.Tests/KestrelServerTests.cs
```diff
@@ -274,6 +315,11 @@ private static KestrelServer CreateServer(KestrelServerOptions options, ILogger
             return new KestrelServer(Options.Create(options), new MockTransportFactory(), new LoggerFactory(new[] { new KestrelTestLoggerProvider(testLogger) }));
         }
 
+        private static KestrelServer CreateServer(KestrelServerOptions options, IDefaultHttpsProvider defaultHttpsProvider)
+        {
+            return new KestrelServer(Options.Create(options), new MockTransportFactory(), new LoggerFactory(new[] { new KestrelTestLoggerProvider(Mock.Of<ILogger>()) }), defaultHttpsProvider);
```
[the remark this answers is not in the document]
anurse: Makes sense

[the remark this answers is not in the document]
anurse: Well. After changing that I'm now getting this error:

```
Failed   KestrelServerThrowsUsefulExceptionIfDefaultHttpsProviderNotAdded
Error Message:
 Assert.Throws() Failure
Expected: typeof(System.InvalidOperationException)
Actual:   typeof(System.AggregateException): An error occurred while writing to logger(s).
Stack Trace:
   at Microsoft.Extensions.Logging.Logger.Log[TState](LogLevel logLevel, EventId eventId, TState state, Exception exception, Func`3 formatter)
   at Microsoft.AspNetCore.Server.Kestrel.Core.Internal.KestrelTrace.Log[TState](LogLevel logLevel, EventId eventId, TState state, Exception exception, Func`3 formatter) in C:\Users\anurse\Code\aspnet\KestrelHttpServer\src\Kestrel.Core\Internal\Infrastructure\KestrelTrace.cs:line 198
   at Microsoft.Extensions.Logging.LoggerExtensions.LogCritical(ILogger logger, EventId eventId, Exception exception, String message, Object[] args)
   at Microsoft.AspNetCore.Server.Kestrel.Core.KestrelServer.<StartAsync>d__24`1.MoveNext() in C:\Users\anurse\Code\aspnet\KestrelHttpServer\src\Kestrel.Core\KestrelServer.cs:line 166
   at Microsoft.AspNetCore.Server.Kestrel.Core.Tests.KestrelServerTests.StartDummyApplication(IServer server) in C:\Users\anurse\Code\aspnet\KestrelHttpServer\test\Kestrel.Core.Tests\KestrelServerTests.cs:line 321
   at Microsoft.AspNetCore.Server.Kestrel.Core.Tests.KestrelServerTests.<>c__DisplayClass2_0.<KestrelServerThrowsUsefulExceptionIfDefaultHttpsProviderNotAdded>b__0() in C:\Users\anurse\Code\aspnet\KestrelHttpServer\test\Kestrel.Core.Tests\KestrelServerTests.cs:line 62
```

I think I'm going to put it back for now :)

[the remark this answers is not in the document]
anurse: Ah, it's because the default ctor uses TestApplicationErrorLogger which throws an exception when a Critical message is logged by default. I'm going to change the test to use a NullLogger instance though instead of the Mock.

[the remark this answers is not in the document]
anurse: I changed `KestrelTestLoggerProvider` to set `ThrowOnCriticalErrors` to `false`, since we don't seem to use that behavior anywhere.

anurse:
:up: :date: Responded to feedback, including @javiercn 's comments re messaging.

test/shared/KestrelTestLoggerProvider.cs
```diff
@@ -11,7 +11,7 @@ public class KestrelTestLoggerProvider : ILoggerProvider
         private readonly ILogger _testLogger;
 
         public KestrelTestLoggerProvider()
-            : this(new TestApplicationErrorLogger())
+            : this(new TestApplicationErrorLogger() { ThrowOnCriticalErrors = false })
```
halter73: Don't do this for all tests For tests where you are expecting a critical errors to be logged, I would just call `new KestrelTestLoggerProvider(new TestApplicationErrorLogger() { ThrowOnCriticalErrors = false })` explicitly.

The idea is that most tests should cause critical errors to be logged, so having this can help us catch issues that might have otherwise been missed.

anurse:
:up: :date: resolved last few pieces of feedback

PR #2097, opened 2017-10-04
Improve TLS handshake

halter73:
#2096


src/Microsoft.AspNetCore.Server.Kestrel.Https/HttpsConnectionAdapterOptions.cs
```diff
@@ -51,5 +57,21 @@ public HttpsConnectionAdapterOptions()
         /// Specifies whether the certificate revocation list is checked during authentication.
         /// </summary>
         public bool CheckCertificateRevocation { get; set; }
+
+        /// <summary>
+        /// Specifies the maximum amount of time allowed for the TLS/SSL handshake.
+        /// </summary>
+        public TimeSpan HandshakeTimeout
```
Eilon: Hmm have we added public API surface in patches before?

arespr:
I think that you could simlify logic to
if(handshakeTask.Wait(_options.HandshakeTimeout) == false)
instead of combination
Task.Delay and Task.WhenAny

davidfowl:
Wait is blocking, we do not want to block.

davidfowl:
Wait is blocking, we do not want to block.

src/Microsoft.AspNetCore.Server.Kestrel.Https/HttpsConnectionAdapterOptions.cs
halter73: Not that I know of, but I think we should keep this property to mitigate the risk of this patch.

Eilon: Can we add a quirks switch instead that reverts to the 2.0.0 RTM behavior?

halter73:
@Eilon I added a quirks switch and removed the HandshakeTimeout option.

test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/HttpsTests.cs
```diff
@@ -258,6 +260,80 @@ public void ConnectionFilterDoesNotLeakBlock()
...
+                    {
+                        listenOptions.UseHttps(new HttpsConnectionAdapterOptions
+                        {
+                            ServerCertificate = new X509Certificate2(TestResources.TestCertificatePath, "testPassword")
+                        });
+                    });
+                })
+                .Configure(app => app.Run(httpContext => Task.CompletedTask));
+
+            AppContext.SetSwitch(HttpsConnectionAdapter.DisableHandshakeTimeoutSwitch, true);
```
davidfowl: ZOMG statics...

src/Microsoft.AspNetCore.Server.Kestrel.Https/Internal/HttpsConnectionAdapter.cs
```diff
@@ -110,8 +113,43 @@ public Task<IAdaptedConnection> OnConnectionAsync(ConnectionAdapterContext conte
...
+                if (AppContext.TryGetSwitch(DisableHandshakeTimeoutSwitch, out var handshakeDisabled) && handshakeDisabled)
+                {
+                    await sslStream.AuthenticateAsServerAsync(
+                        _serverCertificate, certificateRequired, _options.SslProtocols, _options.CheckCertificateRevocation);
+                }
+                else
+                {
+                    try
+                    {
+                        var handshakeTask = sslStream.AuthenticateAsServerAsync(
```
davidfowl: Should we pass in a cancellation token anyways? You can imagine a scenario where this is fixed in .NET Core 2.1 where people run ASP.NET Core 1.x on .NET Core 2.1. It would just work like magic 😄

Tratcher: Yes, they are adding cancellation support.

PR #2098, opened 2017-10-05
HTTP/2: validate request headers prior to starting new stream

cesarbs:
Addresses the error cases described in [Section 8.1.2](http://httpwg.org/specs/rfc7540.html#rfc.section.8.1.2).

The error cases are:

* Uppercase characters in header name
* Pseudo-header fields (header names starting with ':') appearing after regular headers (i.e. they should appear first in a header block)
* Missing mandatory request pseudo-header fields (:method, :path and :scheme - :authority is optional)
* Response pseudo-header field (:status) in request headers
* Duplicate pseudo-header fields
* Connection-specific headers (`Connection`, or `TE` with value different than `trailers`)

### DIFF
--- a/src/Kestrel.Core/Internal/Http2/Http2Connection.cs
+++ b/src/Kestrel.Core/Internal/Http2/Http2Connection.cs
@@ -357,7 +357,17 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http2
                 _streams[_incomingFrame.StreamId] = _currentHeadersStream;
 
                 var endHeaders = (_incomingFrame.HeadersFlags & Http2HeadersFrameFlags.END_HEADERS) == Http2HeadersFrameFlags.END_HEADERS;
-                _hpackDecoder.Decode(_incomingFrame.HeadersPayload, endHeaders, handler: this);
+
+                try
+                {
+                    _hpackDecoder.Decode(_incomingFrame.HeadersPayload, endHeaders, handler: this);
+                }
+                catch (Http2StreamErrorException ex)
+                {
+                    // TODO: log
+                    await _frameWriter.WriteRstStreamAsync(ex.StreamId, Http2ErrorCode.PROTOCOL_ERROR);
+                    return;
+                }
 
                 if (endHeaders)
                 {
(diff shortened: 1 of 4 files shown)


src/Kestrel.Core/Internal/Http2/Http2Connection.cs
```diff
@@ -581,9 +655,167 @@ void IHttp2StreamLifetimeHandler.OnStreamCompleted(int streamId)
...
+                        {
+                            headerField = PseudoHeaderFields.Path;
+                        }
+                        break;
+                    case 6:
+                        if (*intPtr == _methInt && *shortPtr == _odShort)
+                        {
+                            headerField = PseudoHeaderFields.Method;
+                        }
+                        else if (*intPtr == _scheInt && *shortPtr == _meShort)
```
halter73: This reminds me of FrameRequestHeaders.Append which is generated [here](https://github.com/aspnet/KestrelHttpServer/blob/2ce47196906a39b2821c998c109ce1aec235a806/tools/CodeGenerator/KnownHeaders.cs#L36) in KnownHeaders.

I think it might be easier to read/review/edit a code generator instead of manually bit manipulation.

cesarbs: I'll leave that unoptimized for now then, since I want to prioritize finishing HTTP/2 features instead of performance work.

src/Kestrel.Core/Internal/Infrastructure/HttpUtilities.cs
```diff
@@ -59,7 +59,19 @@ private static unsafe ulong GetAsciiStringAsLong(string str)
             }
         }
 
-        private static unsafe uint GetAsciiStringAsInt(string str)
+        public static unsafe ushort GetAsciiStringAsShort(string str)
```
halter73: Since you removed the optimized comparisons, does this file still need to change?

test/Kestrel.Core.Tests/Http2ConnectionTests.cs
```diff
@@ -29,16 +28,14 @@ public class Http2ConnectionTests : IDisposable, IHttpHeadersHandler
         {
             new KeyValuePair<string, string>(":method", "POST"),
             new KeyValuePair<string, string>(":path", "/"),
-            new KeyValuePair<string, string>(":authority", "127.0.0.1"),
-            new KeyValuePair<string, string>(":scheme", "https"),
+            new KeyValuePair<string, string>(":scheme", "http"),
```
halter73: Why did these change?

cesarbs: No strong reason. I just changed those to http because those tests are not using https.

Issue #1694, opened 2017-04-16
Transport specification tests


davidfowl:
We should provide a package that transport authors can use to verify their transport works. Similar to the model we use for DI adapters:

https://github.com/aspnet/DependencyInjection/tree/37380ce62fc2ed3c93490d1111dc88ff16bcc81b/src/Microsoft.Extensions.DependencyInjection.Specification.Tests

This can be done as part of the work to enable sockets and libuv tests to run.

PR #2100, opened 2017-10-05
Improve testing of Sockets transport

halter73:
This PR also makes Sockets the default. This will allow us to see the quickly see the impact of selecting the Sockets transport in ServerTests/MusicStore/Entropy/etc.

#1694


src/Kestrel/WebHostBuilderKestrelExtensions.cs
```diff
@@ -23,10 +26,11 @@ public static class WebHostBuilderKestrelExtensions
         /// </returns>
         public static IWebHostBuilder UseKestrel(this IWebHostBuilder hostBuilder)
         {
-            hostBuilder.UseLibuv();
-
             return hostBuilder.ConfigureServices(services =>
             {
+                // Don't override an already-configured transport
+                services.TryAddSingleton<ITransportFactory, SocketTransportFactory>();
```
davidfowl: Why not call UseSockets()

halter73: Calling UseSockets here would cause UseKestrel override any previously configured transport. This is different than the old behavior with the libuv default, but it makes the TransportSelectors easier to implement.

KestrelHttpServer.sln
```diff
@@ -112,6 +110,17 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "build", "build", "{C2910A13
...
+Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "SystemdActivation", "SystemdActivation", "{B7B0EA74-528F-46B8-9BC4-909D9A67C194}"
+	ProjectSection(SolutionItems) = preProject
+		test\SystemdActivation\docker-entrypoint.sh = test\SystemdActivation\docker-entrypoint.sh
+		test\SystemdActivation\docker.sh = test\SystemdActivation\docker.sh
+		test\SystemdActivation\Dockerfile = test\SystemdActivation\Dockerfile
+	EndProjectSection
+EndProject
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Kestrel.Trasnport.Libuv.FunctionalTests", "test\Kestrel.Transport.Libuv.FunctionalTests\Kestrel.Trasnport.Libuv.FunctionalTests.csproj", "{74032D79-8EA7-4483-BD82-C38370420FFF}"
+EndProject
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Kestrel.Trasnport.Sockets.FunctionalTests", "test\Kestrel.Transport.Sockets.FunctionalTests\Kestrel.Trasnport.Sockets.FunctionalTests.csproj", "{9C7B6B5F-088A-436E-834B-6373EA36DEEE}"
```
cesarbs: typo: Trasnport

src/Kestrel.Transport.Sockets/Internal/ISocketTrace.cs
```diff
@@ -0,0 +1,23 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.AspNetCore.Server.Kestrel.Transport.Sockets.Internal
+{
+    public interface ISocketTrace : ILogger
```
cesarbs: I think this should be called ISocket**s**Trace, given the transport is called Socket**s**.

src/Kestrel.Transport.Sockets/Internal/SocketTrace.cs
```diff
@@ -0,0 +1,93 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.AspNetCore.Server.Kestrel.Transport.Sockets.Internal
+{
+    public class SocketTrace : ISocketTrace
```
cesarbs: Isn't this exactly (or almost exactly) the same as `LibuvTrace`?

src/Kestrel.Transport.Sockets/Internal/SocketTrace.cs
```diff
@@ -0,0 +1,93 @@
...
+    {
+        // ConnectionRead: Reserved: 3
+
+        private static readonly Action<ILogger, string, Exception> _connectionPause =
+            LoggerMessage.Define<string>(LogLevel.Debug, 4, @"Connection id ""{ConnectionId}"" paused.");
```
cesarbs: Won't these IDs clash with other IDs? What's the scope of an event ID?

src/Kestrel/Kestrel.csproj
```diff
@@ -16,7 +16,9 @@
 
   <ItemGroup>
     <ProjectReference Include="..\Kestrel.Core\Kestrel.Core.csproj" />
-    <ProjectReference Include="..\Kestrel.Transport.Libuv\Kestrel.Transport.Libuv.csproj" />
+    <!-- Even though the Libuv transport is no longer used by default, it's remains for back-compat -->
```
cesarbs: it

KestrelHttpServer.sln
```diff
@@ -112,6 +110,17 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "build", "build", "{C2910A13
...
 	EndProjectSection
 EndProject
+Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "SystemdActivation", "SystemdActivation", "{B7B0EA74-528F-46B8-9BC4-909D9A67C194}"
+	ProjectSection(SolutionItems) = preProject
+		test\SystemdActivation\docker-entrypoint.sh = test\SystemdActivation\docker-entrypoint.sh
+		test\SystemdActivation\docker.sh = test\SystemdActivation\docker.sh
+		test\SystemdActivation\Dockerfile = test\SystemdActivation\Dockerfile
+	EndProjectSection
+EndProject
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Kestrel.Trasnport.Libuv.FunctionalTests", "test\Kestrel.Transport.Libuv.FunctionalTests\Kestrel.Trasnport.Libuv.FunctionalTests.csproj", "{74032D79-8EA7-4483-BD82-C38370420FFF}"
```
cesarbs: typo: Trasnport

src/Kestrel.Transport.Sockets/Internal/SocketTrace.cs
halter73: The IDs match mostly to make the cross-transport tests easier. It's in the "Microsoft.AspNetCore.Server.Kestrel.Transport.Sockets" scope, so there shouldn't be any conflicts.

src/Kestrel.Transport.Sockets/Internal/SocketTrace.cs
halter73: Yes. Minus a few things.

src/Kestrel.Transport.Sockets/SocketConnection.cs
```diff
@@ -223,7 +244,13 @@ private void SetupSendBuffers(ReadableBuffer buffer)
             }
             finally
             {
+                // Now, complete the input so that no more reads can happen
+                Input.Complete(error ?? new ConnectionAbortedException());
```
davidfowl: Do we need to do this here? It happens on the input side. It's a bit cleaner than the libuv transport.

halter73: https://github.com/aspnet/KestrelHttpServer/blob/dev/test/Kestrel.FunctionalTests/RequestTests.cs#L1081-L1091

If we don't complete the input with a ConnectionAbortedException here prior to initiating a connection shutdown, the test will close the client half of the connection which will complete the input gracefully.

If something is still trying to read after the output is closed, that's a problem. The connection is aborted at this point. The exact same code is there for libuv: https://github.com/aspnet/KestrelHttpServer/blob/dev/src/Kestrel.Transport.Libuv/Internal/LibuvConnection.cs#L72-L82

src/Kestrel.Transport.Sockets/Internal/SocketTrace.cs
cesarbs: Then how about having a common base class in `Transport.Abstractions`? Do you think that might not suit future transports?

src/Kestrel.Transport.Sockets/SocketConnection.cs
davidfowl: The feature is fine I just want it to happen at the "right place" which IMO is the end of the input loop. 

https://github.com/aspnet/KestrelHttpServer/blob/dev/src/Kestrel.Transport.Sockets/SocketConnection.cs#L208-L215

halter73: I think you linked to the end of the output loop. Just like the LibuvConnection, Sockets completes the input both when the output is closed and the input is closed. This means input is always completed twice in both transports.

src/Kestrel.Transport.Sockets/Internal/SocketTrace.cs
halter73: It could be a good idea, but it's not 100% clear exactly which log methods belong in the base class. Some transports might not be TCP-based and not have a concept of things like sending vs receiving a FIN.

I created #2101 to discuss this idea further.

src/Kestrel.Transport.Sockets/SocketConnection.cs
davidfowl: I meant this:

https://github.com/aspnet/KestrelHttpServer/blob/dev/src/Kestrel.Transport.Sockets/SocketConnection.cs#L111-L122

halter73: I see. Well, it's removed now and we're back to one call to IPipeWriter.Complete(). I'm wondering if we should change the libuv transport as well.

davidfowl: It's easier here since there's 2 async loops

halter73: True. If the libuv output loop exits first, it will call uv_close which prevents the read callback from receiving notification of the disconnect.

src/Kestrel.Transport.Sockets/SocketConnection.cs
```diff
@@ -247,10 +254,12 @@ private void SetupSendBuffers(ReadableBuffer buffer)
             }
             finally
             {
-                // Now, complete the input so that no more reads can happen
                 Output.Complete(error);
 
-                // Make sure to close the connection only after the input is aborted
+                // Make sure to close the connection only after the _aborted flag is set.
+                // Without this, the RequestsCanBeAbortedMidRead test will sometimes fail when
+                // a BadHttpRequestException is thrown instaed of a TaskCanceledException.
```
cesarbs: nit: instead

src/Kestrel.Transport.Sockets/SocketConnection.cs
```diff
@@ -40,6 +46,10 @@ internal SocketConnection(Socket socket, SocketTransport transport)
             RemotePort = remoteEndPoint.Port;
         }
 
+        public override PipeFactory PipeFactory { get; }
+        public override IScheduler InputWriterScheduler => InlineScheduler.Default;
```
davidfowl: Is this problematic?

davidfowl:
Why so Travis failing?

davidfowl:
Seems like there are a bunch of legit failures on OSX.

src/Kestrel.Transport.Sockets/SocketConnection.cs
```diff
@@ -111,25 +133,36 @@ internal SocketConnection(Socket socket, SocketTransport transport)
...
             catch (Exception ex)
             {
                 error = new IOException(ex.Message, ex);
+                _trace.ConnectionError(ConnectionId, error);
             }
             finally
             {
+                if (_aborted)
+                {
+                    error = error ?? new ConnectionAbortedException();
```
muratg: Should we log in the ConnectionAbortedException case?

halter73: Just because the error is being set here, doesn't mean that there is an ongoing call to `Request.Body.ReadAsync()` for the error to be thrown from. This error always gets set for a server-initiated connection close. If there is an ongoing `ReadAsync()` call when the server closes the connection, it's because the application called `Abort()`.

We log a disconnect in the `Abort()` call that sets `_aborted` flag [here](https://github.com/aspnet/KestrelHttpServer/blob/08c6c3866791334d22f6137d3a0548d42ba067c2/src/Kestrel.Core/Internal/Http/Http1OutputProducer.cs#L153). This covers both the libuv and sockets transports.

We could look to changing the log in Http1OutputProducer to make it more clear whether the connection was aborted or closed normally, though this is usually clear when the exception from `ReadAsync` bubbles up and gets logged.

PR #2106, opened 2017-10-10
HTTP/2: implement 100-continue

cesarbs:


src/Kestrel.Core/Internal/Http2/Http2MessageBody.cs
```diff
@@ -16,6 +16,11 @@ protected Http2MessageBody(Http2Stream context)
             _context = context;
         }
 
+        protected override void OnReadStarting()
+        {
+            TryProduceContinue();
```
halter73: It looks to me like the 100-continue response is being sent every time the body of an "Expect: 100-continue" request is read.

Shouldn't this instead behave like HTTP/1.1 where we do not send the 100-continue response if the request body is already available when the app starts reading it?

cesarbs:
Forgot to comment that I've updated this to have the same behavior as we have with HTTP/1.x.

anurse:
Ping on this @halter73 . If it looks good to you, I'll get it merged. Somewhat deferring to you here though since I haven't had a chance to catch up on this code and I don't want to hold this PR open any longer :).

src/Kestrel.Core/Internal/Http/MessageBody.cs
```diff
@@ -111,6 +112,27 @@ public virtual Task ConsumeAsync()
...
+                _context.HttpResponseControl.ProduceContinue();
+                _send100Continue = false;
+            }
+        }
+
+        protected ReadableBufferAwaitable ReadRequestBodyPipeAsync()
+        {
+            var awaitable = _context.RequestBodyPipe.Reader.ReadAsync();
+
+            if (!awaitable.IsCompleted)
```
halter73: @anurse Ideally we would only check this once per-request like we do for HTTP/1.1 instead of per-read like is done here. This might not be so easy with multiple streams per connection. I'll look into merging this tomorrow.

anurse: You mean checking `_send100Continue` I assume?

halter73:
@anurse Can you review? I updated this to check _send100Continue less. This should also prevent the PR from hurting HTTP/1.1 perf.

PR #2107, opened 2017-10-10
HTTP/2: support trailers

cesarbs:


Tratcher:
Do we really want to perpetuate the existing trailers problem? Can we discard these trailers for now until we design a proper feature for them across protocol versions?

cesarbs:
They are part of the spec. We need to at least read the HEADERS frame containing trailers to look at the END_STREAM flag and do something about it (error if it's not there, or complete the stream to which it was sent so any pending `ReadAsync`/`CopyToAsync` is unblocked).

The way I implemented it here, the behavior is pretty much the same as what we have for HTTP/1.x.

Tratcher:
Yes we need to read them, but we can discard them rather than continuing the 1.1 bug of adding them to the request headers.

cesarbs:
I wouldn't call the 1.1 behavior a bug. We just don't separate headers and trailers in two separate collections, which would be better.

Tratcher:
*shrug* we already know it's undesirable behavior that we plan to change, so why continue it here?

cesarbs:
I can leave out the part where it's added to the header dictionary - but the rest should be done (process frames, decode HPACK and throw on errors, etc.).

anurse:
I don't think we should remove the part where it's added to the header dictionary. I'd rather this PR get HTTP/2 up to parity with how HTTP/1 trailers are handled and then handle the trailer problem across the board with a new feature. Leaving HTTP/2 with a half-way implementation where trailers are discarded seems like a recipe for something getting punted and users having no way to actually read trailers.

cesarbs:
Rebased.

PR #2108, opened 2017-10-11
HTTP/2: add exception error messages and log them

cesarbs:
TODO:

- [x] Actually log exceptions
- [x] Verify exceptions in tests

### DIFF
--- a/src/Kestrel.Core/Internal/Http2/Http2Connection.cs
+++ b/src/Kestrel.Core/Internal/Http2/Http2Connection.cs
@@ -166,9 +166,14 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http2
                     }
                 }
             }
-            catch (ConnectionAbortedException ex)
+            catch (ConnectionResetException ex)
             {
-                // TODO: log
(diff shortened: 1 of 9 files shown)


cesarbs:
This is ready for review.

src/Kestrel.Core/Internal/Http2/Http2Connection.cs
```diff
@@ -589,6 +604,7 @@ private Task DecodeHeadersAsync(bool endHeaders, Span<byte> payload)
             }
             catch (Http2StreamErrorException ex)
             {
+                // TODO: log
```
halter73: Why not do this a part of this PR?

cesarbs: Because I totally forgot about it 😆

cesarbs:
Updated w/ logs for `Http2StreamErrorException`.

I've also done one minor change, which is to always close the connection when receiving a WINDOW_UPDATE frame. You can see the change here, as well as the explanation for it in the comments: https://github.com/aspnet/KestrelHttpServer/pull/2108/commits/1f2b55587504bfc0a9d384d66a263434a752d61a#diff-4ab77515238e5e964ae9a3a4615205b1R553

PR #2111, opened 2017-10-11
Set SO_REUSEADDR on managed listen sockets on Unix

halter73:
- https://github.com/dotnet/corefx/issues/24562

### DIFF
--- a/src/Kestrel.Transport.Sockets/SocketTransport.cs
+++ b/src/Kestrel.Transport.Sockets/SocketTransport.cs
@@ -6,10 +6,12 @@ using System.Diagnostics;
 using System.IO.Pipelines;
 using System.Net;
 using System.Net.Sockets;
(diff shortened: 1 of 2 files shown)


src/Kestrel.Transport.Sockets/SocketTransport.cs
```diff
@@ -128,5 +132,32 @@ public Task StopAsync()
                 }
             }
         }
+
+        [DllImport("libc", SetLastError = true)]
+        private static extern int setsockopt(IntPtr socket, int level, int option_name, IntPtr option_value, uint option_len);
+
+        private const int SOL_SOCKET = 0xffff;
+        private const int SO_REUSEADDR = 0x0004;
```
tmds: On Linux you want these to be:
SOL_SOCKET=1
SO_REUSEADDR=2

PR #2122, opened 2017-10-17
Treat SocketError.InvalidArgument from Socket.ReceiveAsync as abort

halter73:
Disposing after calling ReceiveAsync can result in a "SocketError.InvalidArgument" SocketException on macOS and Linux. This is making the RequestsCanBeAbortedMidRead test flaky.


davidfowl:
Add a comment then LGTM

halter73:
This works around dotnet/corefx#24677.

Issue #2085, opened 2017-09-28
Perf Regression: Merge code paths duplicated between HTTP/1.x and HTTP/2 implementations


mikeharder:
"Merge code paths duplicated between HTTP/1.x and HTTP/2 implementations" (https://github.com/aspnet/KestrelHttpServer/issues/2017) seems to have caused a 5% regression in Plaintext on both Cloud and Physical hardware.  You can repro the regression by comparing branch `mikeharder/perf-test-before` to 1470ee0.

Hardware | mikeharder/perf-test-before | 1470ee0 | Ratio
------------ | ------------- | ------------ | ------------
Cloud | 560k | 534k | 0.95
Physical | 1.30M | 1.22M | 0.94

anurse:
It does look like the RequestParsingBenchmark has also regressed by 5%. On mikeharder/perf-test-before I get these results: 

|               Method |     Mean |     Error |    StdDev |      Op/s | Scaled |  Gen 0 | Allocated |
|--------------------- |---------:|----------:|----------:|----------:|-------:|-------:|----------:|
| PlaintextTechEmpower | 1.093 us | 0.0258 us | 0.0386 us | 915,216.8 |   1.00 | 0.0019 |     416 B |

But on `dev` I get:

|               Method |     Mean |     Error |    StdDev |      Op/s | Scaled |  Gen 0 | Allocated |
|--------------------- |---------:|----------:|----------:|----------:|-------:|-------:|----------:|
| PlaintextTechEmpower | 1.144 us | 0.0296 us | 0.0443 us | 874,088.6 |   1.00 | 0.0019 |     416 B |

Going to start exploring that a little bit.

anurse:
Resolved by #2123 (30010a103bb20f913d00f64d49a9914d37eb8650)

PR #2123, opened 2017-10-18
Fix for #2085 - "The Detaskening"

anurse:
PR #2041 ended up introducing a 5% perf regression in the HTTP/1.1 path. After investigating, I tracked it down to the refactoring of parsing logic into a helper function. Prior to #2041, the HTTP/1.1 and HTTP/2 implementations of `Frame` (what is now `HttpProtocol`) had their own request processing loops and the parsing was done inline in this function (in fact, much of the logic was done in-line, or in functions that intentionally avoid creating an async state machine until necessary). In #2041, by refactoring the parsing logic into a helper function, an async state machine was created in every request (within the `Http1Connection.ParseRequestAsync` method).

So, unfortunately this means we can't have nice things and we have to unrefactor that logic. It also means that `Http1Connection` and `Http2Stream` (the two implementations of `HttpProtocol`) have to have mostly-similar `ProcessRequestsAsync` methods but we can't refactor them up to the base class (because it would create more state machines).

Open to suggestions on how to add the code sharing back in a performant way.

Perf Data:

| Hardware | Benchmark | `dev` | This change | Ratio |
|-|-|-|-|-|
| Physical | Plaintext | 1.246M | 1.307M | 1.246 / 1.307 = **0.953** |
| Cloud | Plaintext | 501,510.91 | 537,988.42 | **0.933** (but Cloud data is also a bit volatile) |

/fyi @cesarbs, in case you're curious what the fix ended up being ;).

### DIFF
--- a/src/Kestrel.Core/Internal/Http/Http1Connection.cs
+++ b/src/Kestrel.Core/Internal/Http/Http1Connection.cs
@@ -413,67 +413,77 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
         protected override MessageBody CreateMessageBody()
             => Http1MessageBody.For(_httpVersion, HttpRequestHeaders, this);
 
-        protected override async Task<bool> ParseRequestAsync()
+        protected override void InitializeParsing()
         {
             Reset();
             TimeoutControl.SetTimeout(_keepAliveTicks, TimeoutAction.StopProcessingNextRequest);
+        }
 
-            while (_requestProcessingStatus != RequestProcessingStatus.AppStarted)
+        protected override bool BeginRead(out ReadableBufferAwaitable awaitable)
+        {
+            awaitable = Input.ReadAsync();
+            return true;
+        }
+
+        protected override void FinishParsing()
+        {
+            EnsureHostHeaderExists();
+        }
+
+        protected override bool TryParseRequest(ReadResult result, out bool endConnection)
+        {
+            var examined = result.Buffer.End;
+            var consumed = result.Buffer.End;
+
+            try
             {
-                var result = await Input.ReadAsync();
-
-                var examined = result.Buffer.End;
-                var consumed = result.Buffer.End;
-
-                try
+                ParseRequest(result.Buffer, out consumed, out examined);
+            }
+            catch (InvalidOperationException)
+            {
+                if (_requestProcessingStatus == RequestProcessingStatus.ParsingHeaders)
                 {
-                    ParseRequest(result.Buffer, out consumed, out examined);
+                    throw BadHttpRequestException.GetException(RequestRejectionReason
+                        .MalformedRequestInvalidHeaders);
                 }
-                catch (InvalidOperationException)
-                {
-                    if (_requestProcessingStatus == RequestProcessingStatus.ParsingHeaders)
-                    {
-                        throw BadHttpRequestException.GetException(RequestRejectionReason
-                            .MalformedRequestInvalidHeaders);
-                    }
-                    throw;
-                }
-                finally
-                {
-                    Input.Advance(consumed, examined);
-                }
-
-                if (result.IsCompleted)
-                {
-                    switch (_requestProcessingStatus)
-                    {
-                        case RequestProcessingStatus.RequestPending:
-                            return false;
-                        case RequestProcessingStatus.ParsingRequestLine:
-                            throw BadHttpRequestException.GetException(
-                                RequestRejectionReason.InvalidRequestLine);
-                        case RequestProcessingStatus.ParsingHeaders:
-                            throw BadHttpRequestException.GetException(
-                                RequestRejectionReason.MalformedRequestInvalidHeaders);
-                    }
-                }
-                else if (!_keepAlive && _requestProcessingStatus == RequestProcessingStatus.RequestPending)
-                {
-                    // Stop the request processing loop if the server is shutting down or there was a keep-alive timeout
-                    // and there is no ongoing request.
-                    return false;
-                }
-                else if (RequestTimedOut)
-                {
-                    // In this case, there is an ongoing request but the start line/header parsing has timed out, so send
-                    // a 408 response.
-                    throw BadHttpRequestException.GetException(RequestRejectionReason.RequestTimeout);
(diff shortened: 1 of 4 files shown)


anurse:
Welp, in tidying this up to focus just on removing the async state machine, I reintroduced the perf regression, indicating it's not actually that specific part of this change that caused the improvement.

anurse:
False alarm! I'll have an update with the refactor. It preserves the same perf improvement as before (in my benchmarks)

anurse:
:up: :date: @davidfowl @halter73 @pakrym - Managed to save the refactoring!

src/Kestrel.Core/Internal/Http/Http1Connection.cs
```diff
@@ -413,67 +413,77 @@ protected override string CreateRequestId()
...
+                return true;
+            }
+            else if (RequestTimedOut)
+            {
+                // In this case, there is an ongoing request but the start line/header parsing has timed out, so send
+                // a 408 response.
+                throw BadHttpRequestException.GetException(RequestRejectionReason.RequestTimeout);
+            }
 
-            EnsureHostHeaderExists();
```
halter73: Nit: Why not continue to call EnsureHostHeaderExists here if `_requestProcessingStatus == RequestProcessingStatus.AppStarted`? I think it cleans up HttpProtocol a bit to remove FinishParsing.

src/Kestrel.Core/Internal/Http/Http1Connection.cs
```diff
@@ -413,67 +413,77 @@ protected override string CreateRequestId()
         protected override MessageBody CreateMessageBody()
             => Http1MessageBody.For(_httpVersion, HttpRequestHeaders, this);
 
-        protected override async Task<bool> ParseRequestAsync()
+        protected override void InitializeParsing()
```
halter73: I also think that calling InitializeParsing at the beginning of the first call to BeginRead would be nice for simplifying the HttpProtocol class. Just be sure to only call it when _requestProcessingStatus is RequestPending.

src/Kestrel.Core/Internal/Http/Http1Connection.cs
anurse: Fair point. Will add that in.

src/Kestrel.Core/Internal/Http/Http1Connection.cs
anurse: If that's a reliable way to detect the first call to BeginRead, sure.

halter73: It's reliable. There are four ways for ReadAsync to complete.

1. With request data: If the ReadableBuffer is anything but empty, it will move out of RequestPending and into the ParsingRequestLine state.
2. With a graceful connection close: This of course exits the while loop by returning true from the TryParseRequest method and setting the endConnection out param.
3. With a connection error: ReadAsync throws and exits the loop.
4. Because CancelPendingRead: This is the most unusual scenario. CancelPendingRead is only called on the connection's Input pipe during server shutdown. In this scenario, _keepAlive will be set to false first. The loop will exit due to the `else if (!_keepAlive && _requestProcessingStatus == RequestProcessingStatus.RequestPending)` condition in TryParseRequest.

I'm only going into such excruciating detail b/c you're ramping up on Kestrel.

anurse: Looks like it's actually `Reset` that sets the state back to `RequestPending`, so that wouldn't quite work as-written. Instead, I've moved the code to reset the state up to the `ProcessRequestsAsync` loop itself.

anurse:
:up: :date: (running build/tests right now though, I may have derped something, but I think the feedback has been addressed)

src/Kestrel.Core/Internal/Http/Http1Connection.cs
anurse: Though to be fair `Reset` is on `HttpProtocol`. So in theory we could run it in the outer loop rather than in `BeginRead`. Need to see where HTTP/2 does it to see how that compares. The timeout stuff is HTTP/1.1 only (for now)

anurse:
Well, I got an assert about "Concurrent Timeouts not supported" on my build by AppVeyor succeeded. Might have derped something. Investigating.

anurse:
Looks like there are cases where BeginRead can be called a second time for a single request but `_requestProcessStatus` will still be `RequestPending`. Investigating further.

anurse:
Yep, it seems that if the first buffer is empty this can happen. Not sure what test is causing this situation. Think I might just go back to the `InitializeParsing` method. Any other solution I can come up with just ends up the same way (with some kind of boolean flag, which may as well just be done via an Initialize method since the caller already knows when the first call is)

test/Kestrel.Core.Tests/Kestrel.Core.Tests.csproj
```diff
@@ -23,4 +23,8 @@
     <PackageReference Include="Microsoft.AspNetCore.Http" />
   </ItemGroup>
 
+  <ItemGroup>
+    <Service Include="{82a7f48d-3b50-4b1e-b82e-3ada8210c358}" />
+  </ItemGroup>
```
halter73: Nit: remove this.

src/Kestrel.Core/Internal/Http/Http1Connection.cs
```diff
@@ -413,67 +413,81 @@ protected override string CreateRequestId()
         protected override MessageBody CreateMessageBody()
             => Http1MessageBody.For(_httpVersion, HttpRequestHeaders, this);
 
-        protected override async Task<bool> ParseRequestAsync()
+        protected override void BeginRequestProcessing()
```
halter73: 👍  It does feel better to not rely on the state always transitioning on the first iteration of the loop.

The other idea didn't work because of [this CancelPendingRead in Http1MessageBody.StopAsync()](https://github.com/aspnet/KestrelHttpServer/blob/dev/src/Kestrel.Core/Internal/Http/Http1MessageBody.cs#L129). I forgot that this also called CancelPendingRead.

PR #2124, opened 2017-10-19
[WIP] Migrate to new pipe APIs 

pakrym:


Directory.Build.props
```diff
@@ -10,7 +10,6 @@
     <PublicSign Condition="'$(OS)' != 'Windows_NT'">true</PublicSign>
     <VersionSuffix Condition="'$(VersionSuffix)'!='' AND '$(BuildNumber)' != ''">$(VersionSuffix)-$(BuildNumber)</VersionSuffix>
     <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
-    <LangVersion>latest</LangVersion>
```
halter73: reminder to revert

muratg: Double reminder

NuGet.config
```diff
@@ -5,5 +5,7 @@
     <add key="AspNetCore" value="https://dotnet.myget.org/F/aspnetcore-ci-dev/api/v3/index.json" />
     <add key="NuGet" value="https://api.nuget.org/v3/index.json" />
     <add key="roslyn" value="https://dotnet.myget.org/F/roslyn/api/v3/index.json" />
+    <add key="corefxlab" value="https://dotnet.myget.org/F/dotnet-corefxlab/api/v3/index.json" />
+    <add key="corefx" value="https://dotnet.myget.org/F/dotnet-core/api/v3/index.json" />
```
muratg: These to be removed as well, I assume... (once the bits are consumed?)

src/Kestrel.Transport.Sockets/Internal/SocketConnection.cs
```diff
@@ -240,5 +243,16 @@ internal SocketConnection(Socket socket, PipeFactory pipeFactory, ISocketsTrace
                 _socket.Shutdown(SocketShutdown.Both);
             }
         }
+
+        private static ArraySegment<byte> GetArraySegment(Memory<byte> buffer)
```
halter73: Is this called still?

PR #2130, opened 2017-10-24
Add some tracing and increase timeout for flaky test

anurse:
This will help with #1955 , but I'm not yet sure I've identified why it's flaky. My hope is that when it does flake out again, we'll have some logs to go back through.


src/Kestrel.Core/Adapter/ListenOptionsConnectionLoggingExtensions.cs
```diff
@@ -30,6 +30,11 @@ public static ListenOptions UseConnectionLogging(this ListenOptions listenOption
         public static ListenOptions UseConnectionLogging(this ListenOptions listenOptions, string loggerName)
         {
             var loggerFactory = listenOptions.KestrelServerOptions.ApplicationServices.GetRequiredService<ILoggerFactory>();
+            return listenOptions.UseConnectionLogging(loggerFactory, loggerName);
+        }
+
+        internal static ListenOptions UseConnectionLogging(this ListenOptions listenOptions, ILoggerFactory loggerFactory, string loggerName)
```
anurse: I didn't want to create a new public API here, but having the connection logging in the test I was tracing seemed somewhat useful. I'm open to this being public, as it doesn't seem to hurt, but I don't really know the public use-case for it :). I needed to add it because in the tests we create `ListenOptions` which do not have an associated `KestrelServerOptions`.

muratg: Perhaps pubternal instead? (I'm sure @davidfowl has an opinion on this.)

anurse: It can't really be pubternal because it's an extension method. We could put it in a separate namespace, but we generally put extension methods like this in the same namespace as the type being extended.

halter73: Instead of adding an extension method, I'd just call `listenOptions.ConnectionAdapters.Add(new LoggingConnectionAdapter(logger));` directly from the test.

test/Kestrel.Core.Tests/HttpConnectionTests.cs
```diff
@@ -1,13 +1,15 @@
...
 
 using System;
 using System.Collections.Generic;
 using System.IO.Pipelines;
 using System.Threading;
+using System.Threading.Tasks;
 using Microsoft.AspNetCore.Http.Features;
 using Microsoft.AspNetCore.Server.Kestrel.Core.Adapter.Internal;
 using Microsoft.AspNetCore.Server.Kestrel.Core.Internal;
+using Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http;
```
halter73: Nit: it doesn't look like these new usings are needed.

test/Kestrel.FunctionalTests/ResponseTests.cs
```diff
@@ -19,23 +20,28 @@
...
 using Microsoft.AspNetCore.Http.Features;
 using Microsoft.AspNetCore.Server.Kestrel.Core;
+using Microsoft.AspNetCore.Server.Kestrel.Core.Adapter.Internal;
 using Microsoft.AspNetCore.Server.Kestrel.Core.Internal;
 using Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http;
 using Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Infrastructure;
 using Microsoft.AspNetCore.Server.Kestrel.Https;
 using Microsoft.AspNetCore.Server.Kestrel.Https.Internal;
+using Microsoft.AspNetCore.Server.Kestrel.Transport.Abstractions.Internal;
+using Microsoft.AspNetCore.Server.Kestrel.Transport.Libuv;
```
halter73: Nit: extra usings.

test/Kestrel.FunctionalTests/TestHelpers/TestServer.cs
```diff
@@ -14,6 +14,8 @@
 using Microsoft.AspNetCore.Server.Kestrel.Transport.Abstractions.Internal;
 using Microsoft.AspNetCore.Testing;
 using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Logging.Abstractions;
```
halter73: Nit: more extra usings.

src/Kestrel.Core/Internal/Infrastructure/KestrelTrace.cs
```diff
@@ -174,7 +174,7 @@ public void RequestBodyMininumDataRateNotSatisfied(string connectionId, string t
             _requestBodyMinimumDataRateNotSatisfied(_logger, connectionId, traceIdentifier, rate, null);
         }
 
-        public void ResponseMininumDataRateNotSatisfied(string connectionId, string traceIdentifier)
+        public virtual void ResponseMininumDataRateNotSatisfied(string connectionId, string traceIdentifier)
```
halter73: This whole class is now a mix of virtual and non-virtual. I tried to see what would happen if I made everything non-virtual in dev, and the build still passes. It turns out that none of the other methods actually need to be virtual.

Can you make *all* the logging methods either virtual or not for consistency? Thanks.

anurse: Did you run the whole build or just compilation? My understanding is that they are overridden by Moq mocks. I can make them all virtual for consistency though.

I'm not super happy with that model of verification, but other things used it so I didn't want to rewrite everything.

anurse: Hang on, this is on KestrelTrace itself. Yeah, I'm going to remove this mock entirely. I want the test logging, hence using KestrelTrace, but the test also verified that the necessary method was called, which I don't think is necessary in this case.

halter73:
I like verifying that ResponseMininumDataRateNotSatisfied is logged. It ensures that it's really the min data rate that's causing the connection to get aborted, and I don't think it makes the test anymore fragile or difficult to maintain.

I understand wanting to capture the logs too, which is why I'm alright with making all the KestrelTrace methods virtual.

anurse:
Alright, sounds OK to me. It's just unfortunately awkward that it has to be done via `KestrelTrace` rather than `IKestrelTrace` here (in order for the real logging to flow through). And even then, the ResponseMinimumDataRate event never actually makes it to the real log because of how Moq works.

test/Kestrel.FunctionalTests/ResponseTests.cs
```diff
@@ -2447,73 +2451,94 @@ public Task ResponseStatusCodeSetBeforeHttpContextDisposedRequestMalformedReadIg
...
+                var responseSize = chunks * chunkSize;
 
-            var testContext = new TestServiceContext
-            {
-                Log = mockKestrelTrace.Object,
-                SystemClock = new SystemClock(),
-                ServerOptions =
+                var requestAborted = new ManualResetEventSlim();
+                var messageLogged = new ManualResetEventSlim();
+                var mockKestrelTrace = new Mock<KestrelTrace>(loggerFactory.CreateLogger("Microsoft.AspNetCore.Server.Kestrel")) { CallBase = true };
```
halter73: :+1: I see you found `CallBase`. I don't think you can change this on a per-method basis though.

anurse: I had CallBase there before, but it requires that the method be virtual, hence the whole virtual/non-virtual discussion above :). I actually don't want it per-method. I want it on for all methods including ones I add Callbacks to (i.e. run the base logic after the callback or something), but Moq doesn't support that, so for that one method that I actually mock out, the base behavior won't run.

test/Kestrel.Core.Tests/HttpConnectionTests.cs
```diff
@@ -1,4 +1,4 @@
-﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Copyright (c) .NET Foundation. All rights reserved.
```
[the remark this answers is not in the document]
anurse: I'm not really sure. VS shows the encoding of this, and other files in the repo that I didn't touch, as "Western European Code Page 1252" which is the ANSI Latin-1 code page (aka ASCII, for all intents and purposes). I don't know what it was before. Perhaps because I edited it and saved it at some point it converted it?

PR #2132, opened 2017-10-24
add event names to Logging messages

anurse:
I was in KestrelTrace and noticed that we haven't done the "Update all use of EventID to include Event Names" update to this repo. So I ran my VsVim macro on these files (see @davidfowl , VsVim is great) and created this PR :). Pretty mild update, shouldn't affect anything functional except that events now include a name as well as an integer Id.


anurse:
Weird. VS wasn't showing me the build errors I'm getting now. Stand-by for a fix. Just some simple name mismatches.

anurse:
There was a test that was using Moq to verify certain log messages were written. I changed it to use TestLogger since that's what it's designed for and it's not as brittle to this kind of refactoring.

PR #2133, opened 2017-10-24
Better handle Socket transport accept errors

halter73:
Also don't log server-aborted connections as errors.

### DIFF
--- a/src/Kestrel.Transport.Sockets/SocketTransport.cs
+++ b/src/Kestrel.Transport.Sockets/SocketTransport.cs
@@ -6,8 +6,10 @@ using System.Diagnostics;
 using System.IO.Pipelines;
 using System.Net;
 using System.Net.Sockets;
+using System.Runtime.ExceptionServices;
 using System.Runtime.InteropServices;
 using System.Threading.Tasks;
+using Microsoft.AspNetCore.Hosting;
 using Microsoft.AspNetCore.Protocols;
 using Microsoft.AspNetCore.Server.Kestrel.Transport.Abstractions.Internal;
 using Microsoft.AspNetCore.Server.Kestrel.Transport.Sockets.Internal;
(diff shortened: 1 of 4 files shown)


halter73:
This should fix the flaky ConnectionFilterDoesNotLeakBlock test.

src/Kestrel.Transport.Sockets/SocketTransport.cs
```diff
@@ -89,14 +94,22 @@ public Task BindAsync()
...
                 Debug.Assert(_listenTask != null);
                 await _listenTask.ConfigureAwait(false);
+
+                _unbinding = false;
+                _listenSocket = null;
                 _listenTask = null;
+
+                if (_listenException != null)
+                {
+                    var exInfo = ExceptionDispatchInfo.Capture(_listenException);
```
anurse: What's the point of capturing and then immediately rethrowing? Aren't you supposed to `.Capture` at the point where the exception occurs and then `.Throw`?

I may be misinterpreting how EDI works though.

pakrym: You don't have to capture at the point of exception. See https://github.com/dotnet/corefx/pull/17570

src/Kestrel.Transport.Sockets/SocketTransport.cs
```diff
@@ -112,23 +125,39 @@ public Task StopAsync()
...
+                        var acceptSocket = await _listenSocket.AcceptAsync();
+                        acceptSocket.NoDelay = _endPointInformation.NoDelay;
+
+                        var connection = new SocketConnection(acceptSocket, _pipeFactory, _trace);
+                        _ = connection.StartAsync(_handler);
+                    }
+                    catch (SocketException ex) when (ex.SocketErrorCode == SocketError.ConnectionReset)
+                    {
+                        // REVIEW: Should there be a seperate log message for a connection reset this early?
+                        _trace.ConnectionReset(connectionId: "(null)");
```
pakrym: Do you want to log connection reset when unbinding?

src/Kestrel.Transport.Sockets/SocketTransport.cs
anurse: I withdraw my concerns then :). Thanks for clarifying.

src/Kestrel.Transport.Sockets/SocketTransport.cs
halter73: Yes. AFAICT unbinding doesn't cause a SocketError.ConnectionReset. If it happens to occur during shutdown, I don't see the harm in this info level log message.

PR #2136, opened 2017-10-26
Pass setsockopt fd as int

tmds:
Even on 64-bit, Unix file descriptors are 32-bit.

CC @halter73

### DIFF
--- a/src/Kestrel.Transport.Sockets/SocketTransport.cs
+++ b/src/Kestrel.Transport.Sockets/SocketTransport.cs
@@ -154,7 +154,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Transport.Sockets
                 {
                     _trace.LogCritical($"Unexpected exeption in {nameof(SocketTransport)}.{nameof(RunAcceptLoopAsync)}.");
                     _listenException = ex;
-                    
+
                     // Request shutdown so we can rethrow this exception
                     // in Stop which should be observable.
                     _appLifetime.StopApplication();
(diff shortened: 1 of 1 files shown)


halter73:
Thanks!

src/Kestrel.Transport.Sockets/SocketTransport.cs
```diff
@@ -163,7 +163,7 @@ public Task StopAsync()
         }
 
         [DllImport("libc", SetLastError = true)]
-        private static extern int setsockopt(IntPtr socket, int level, int option_name, IntPtr option_value, uint option_len);
+        private static extern int setsockopt(int socket, int level, int option_name, IntPtr option_value, uint option_len);
```
halter73: @tmds Do you think the same change should be made [in corefx](https://github.com/dotnet/corefx/blob/bffef76f6af208e2042a2f27bc081ee908bb390b/src/Common/src/Interop/Unix/System.Native/Interop.SetSockOpt.cs#L17)?

tmds: corefx converts the intptr to an int in native code

https://github.com/dotnet/corefx/blob/473463e644cba090257b6fde42a17af5bc01b01b/src/Native/Unix/Common/pal_utilities.h#L143-L148

tmds: (corefx is using IntPtr between managed and native is because SafeHandles marshal to intptr)

PR #2137, opened 2017-10-27
Pin tool and package versions to make builds more repeatable

natemcmaster:
Part of https://github.com/aspnet/Universe/issues/575

This removes floating versions of dependencies. This includes the korebuild version and versions used by PackageReference's.


natemcmaster:
🔔

davidfowl:
How do we avoid commits that just update the version of dependencies?

davidfowl:
I think this is a bad idea. Will discuss offline.

natemcmaster:
Floating dependencies is worse, trust me.

davidfowl:
I lived in both worlds and it’s possible you’re right but I need to understand what the main thing is that we’re fixing. Having managed this in the past for 3 repositories, where things went very poorly I thought floating was an improvement.

How do we deal with package versions being deleted from myget? Maybe we don’t care if builds are *that* repeatable?

How do we handle cross repository breaking changes? Does the ci catch these or is it up to the dev to manually update things? 


When are devs supposed to update things?

natemcmaster:
All good questions. Let me add some context.

A few months ago, we decided to tackle problems listed in https://github.com/aspnet/KoreBuild/issues/239, "auto-update and version pinning". The problem: floating package versions frequently fail. Devs were frequently getting broken restore results due to caching race conditions. Plus, it floating made restores almost impossible to debug as restore results could vary from minute to minute. The result of this was PackageLineup. This reduced the 'floating' to a single point: the lineup package, but introduced a new requirement for you to run 'build.cmd /t:Pin' before you could work on a repo in VS. This continues to irritate devs on our team and confusing external contributors, testers, and other teams.

So, the current iteration on the plan to address https://github.com/aspnet/KoreBuild/issues/239 is to remove PackageLineup, but to forbid floating versions. I wrote tools to automate updates to the dependencies.props file. See https://github.com/aspnet/BuildTools/wiki/PackageReference-management. Forbidding floating versions addresses https://github.com/aspnet/KoreBuild/issues/239 and should help devs isolate their work from the rapid pace of change coming out of CI builds.

> How do we deal with package versions being deleted from myget? Maybe we don’t care if builds are that repeatable?

We plan to increase myget retention. We will only delete packages from myget after we RTM a stable release. At that point, we expect our source code to have converged to stable versions.

> How do we handle cross repository breaking changes? Does the ci catch these or is it up to the dev to manually update things?

CI will catch these. Each run of CI builds latest against latest. For local development, I'm working on enhancements to aspnet/Universe to reduce the friction of cross-repo changes.

> When are devs supposed to update things?

When they need new implementations or new API from other packages. Otherwise, let the automation handle it.

davidfowl:
So we only care about being able to build commits during a release but after we ship we don’t care. This mostly reduces churn for developers but doesn’t change the ci right?

I’d the ci runs against the latest (meaning different version than what’s in source control) how does a developer reproduce the failure? Does the ci spit out a list of versions that were used or are we doing project to project references? 

So as a developer, if I’m never consuming a need API, I’ll need to run that upgrade script.

So repositories with little churn wil just have a bunch of upgrade dependencies commits. I guess that’s the tradeoff... I hate that but I don’t know a way around it. 

I don’t see a big deal with running build.cmd once. The bigger issue was how often you had to re-run it(new projects, branch changes). I never looked at how /t:Pin was implemented but why wasn’t it good enough to get the dependencies file once and have it be petty much identical to this proposal.

natemcmaster:
> So we only care about being able to build commits during a release but after we ship we don’t care. This mostly reduces churn for developers but doesn’t change the ci right?

Correct, one goal is reducing churn for developers working on the `dev` branch. 

> I’d the ci runs against the latest (meaning different version than what’s in source control) how does a developer reproduce the failure? Does the ci spit out a list of versions that were used or are we doing project to project references?

I'm working on making builds of aspnet/Universe repeatable too.  https://github.com/aspnet/Universe/pull/621. You should be able to reproduce failures locally by cloning universe, setting submodules to the commits CI was trying to build, and running `build.cmd`.

> So as a developer, if I’m never consuming a need API, I’ll need to run that upgrade script.

If you're never consuming new API between repos, you should not need to run the upgrade script. 

> I don’t see a big deal with running build.cmd once. The bigger issue was how often you had to re-run it(new projects, branch changes). 

Bingo. PackageLineup cannot handle branch changes or update on a git-pull because it writes files into the repo that are ignored by source-control. Committing versions to source means changing branches, git-pull, and building old commits should work.

Also, PackageLineup was hard to work with on new projects. Devs had to wait for a new build of universe before they could use new packages or new package versions. These restrictions don't apply to local builds anymore.

pakrym:
> If you're never consuming new API between repos, you should not need to run the upgrade script.
We need to stop thinking only about API changes, behavior changes are important too and happen all the time.

natemcmaster:
True. My main point is: this gives devs enough control to isolate their work from those changes, and gives them a simple way to pull in updates when they want them. Less magic, more automation.

build/dependencies.props
```diff
@@ -0,0 +1,44 @@
+﻿<Project>
+  <PropertyGroup>
+    <MSBuildAllProjects>$(MSBuildAllProjects);$(MSBuildThisFileFullPath)</MSBuildAllProjects>
+  </PropertyGroup>
+  <!-- This files is typically managed by automation. Execute 'run.ps1 upgrade deps' to update these variables to the last-known-good versions. -->
+  <PropertyGroup Label="Package Versions">
+    <BenchmarkDotNetPackageVersion>0.10.9</BenchmarkDotNetPackageVersion>
+    <InternalAspNetCoreSdkPackageVersion>2.1.0-preview1-15544</InternalAspNetCoreSdkPackageVersion>
+    <LibuvPackageVersion>1.10.0</LibuvPackageVersion>
+    <MicrosoftAspNetCoreBenchmarkRunnerSourcesPackageVersion>2.1.0-preview1-27450</MicrosoftAspNetCoreBenchmarkRunnerSourcesPackageVersion>
```
davidfowl: Can't we have a shared version for AspNetCore things?

natemcmaster: While there is certainly duplication here, we've been burned in the past by trying to combine variables. Patch builds, one-off version changes, 'special packages', etc. There were enough edge cases that we found it's simpler and more reliable to do `packageId.ToPascalCase() + "PackageVersion"`.

natemcmaster:
> Were the changes generated with a script?

Yes. See https://github.com/aspnet/BuildTools/blob/dev/modules/KoreBuild.Tasks/GenerateDependenciesPropsFile.cs. It was reviewed by @ryanbrandenburg in https://github.com/aspnet/BuildTools/pull/434

natemcmaster:
FYI - I would like to merge this soon, so please take a look now if you want to add additional feedback.

There were some concerns discussed in-person about how this affects cross-repo changes, especially when there are breaking changes. (See aspnet/Coherence-Signed#680). I don't believe these concerns should prevent us from striving for reproducible builds. It will require devs to take explicit actions to upgrade downstream repos when breaking API changes are made, but this is actually something that I believe we want (as opposed to repos that suddenly can't build because dependencies changed.)

cc @mikeharder @muratg

muratg:
cc @halter73 

Let's get this in and give it a go. We need to make progress towards ProdCon and if we find out that this doesn't work or work worse than before, we can regroup and go back to the drawing board.

build/dependencies.props
davidfowl: See https://github.com/aspnet/HttpClientFactory/pull/22/files#diff-30dcc9789f8ea3c7632e4f054e60c072. I think it's ridiculous...

PR #2147, opened 2017-11-02
Use SocketAsyncEventArgs to improve perf

halter73:
Using SocketAsyncEventArgs in the Kestrel managed socket transport instead of the TAP APIs helps narrow the performance gap between the libuv and socket transports on both Windows and Linux.

Here are some RPS numbers I got from running the plaintext benchmark with no pipelining against the [PlatformLevelTechempower app](https://github.com/halter73/PlatformLevelTechempower/commit/428d30d08fb7d0d09d47938966bfbc2fde28b6cf). I used this instead of the the official [Benchmarks app](https://github.com/aspnet/Benchmarks) because PlatformLevelTechempower is written directly against the transport abstraction.

### Benchmark results:

image

### Benchmark environment info:
```
PlatformLevelTechempower halter73/2.1 commit: 428d30d08fb7d0d09d47938966bfbc2fde28b6cf
Kestrel dev commit: 6c82f78c657145404bcdbe403024b2d93ca7699a
Kestrel halter73/socket-event-args commit: 69d7c4c3affd2373ad33d11784c26977c693aee0
Windows server, Linux server and Linux client VMs: F4S
Windows OS: Windows Server 2016 Datacenter
Linux server and client OS: Ubuntu 16.04.2 LTS

Sample server commands:
dotnet ./bin/Release/netcoreapp2.0/PlatformLevelTechempower.dll --mode raw --port 5000 --transport sockets
dotnet ./bin/Release/netcoreapp2.0/PlatformLevelTechempower.dll --mode raw --port 5000

Sample client command:
wrk -t 32 -c 256 -d 30 http://host:5000/plaintext
```

### Copyable benchmark results:

  | Windows Sockets Dev | Windows SocketAsyncEventArgs | Windows Libuv |Linux Sockets Dev | Linux SocketAsyncEventArgs | Linux Libuv
-- | -- | -- | -- | -- |-- | --
Run 1 | 163209 | 186672 | 233858 | 188827 | 202882 | 231730
Run 2 | 163330 | 187990 | 236810 | 190314 | 205801 | 238674
Run 3 | 166695 | 187526 | 238348 | 189439 | 206955 | 235977
AVG: | 164411 | 187396 | 236338 | 189527 | 205213 | 235460

### DIFF
--- a/src/Kestrel.Transport.Sockets/Internal/SocketAwaitable.cs
+++ b/src/Kestrel.Transport.Sockets/Internal/SocketAwaitable.cs
@@ -2,6 +2,7 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.Diagnostics;
 using System.Net.Sockets;
 using System.Runtime.CompilerServices;
 using System.Threading;
@@ -21,6 +22,8 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Transport.Sockets.Internal
 
         public int GetResult()
         {
+            Debug.Assert(_callback == _callbackCompleted);
+
             _callback = null;
 
             if (_error != SocketError.Success)
--- a/src/Kestrel.Transport.Sockets/Internal/SocketAwaitable.cs
+++ b/src/Kestrel.Transport.Sockets/Internal/SocketAwaitable.cs
@@ -1,0 +1,55 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Net.Sockets;
+using System.Runtime.CompilerServices;
+using System.Threading;
+
+namespace Microsoft.AspNetCore.Server.Kestrel.Transport.Sockets.Internal
+{
+    public class SocketAwaitable : ICriticalNotifyCompletion
+    {
+        private static readonly Action _callbackCompleted = () => { };
+
+        private Action _callback;
+        private int _bytesTransfered;
+        private SocketError _error;
+
+        public SocketAwaitable GetAwaiter() => this;
+        public bool IsCompleted => _callback == _callbackCompleted;
+
+        public int GetResult()
+        {
+            _callback = null;
+
+            if (_error != SocketError.Success)
+            {
+                throw new SocketException((int)_error);
+            }
+
+            return _bytesTransfered;
+        }
+
+        public void OnCompleted(Action continuation)
+        {
+            if (_callback == _callbackCompleted ||
+                Interlocked.CompareExchange(ref _callback, continuation, null) == _callbackCompleted)
+            {
+                continuation();
+            }
+        }
+
+        public void UnsafeOnCompleted(Action continuation)
+        {
+            OnCompleted(continuation);
+        }
+
+        public void Complete(int bytesTransferred, SocketError socketError)
+        {
+            _error = socketError;
+            _bytesTransfered = bytesTransferred;
+            Interlocked.Exchange(ref _callback, _callbackCompleted)?.Invoke();
+        }
+    }
+}
--- a/src/Kestrel.Transport.Sockets/Internal/SocketReceiver.cs
+++ b/src/Kestrel.Transport.Sockets/Internal/SocketReceiver.cs
@@ -1,0 +1,44 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Net.Sockets;
+
+namespace Microsoft.AspNetCore.Server.Kestrel.Transport.Sockets.Internal
+{
+    public class SocketReceiver
+    {
+        private readonly Socket _socket;
+        private readonly SocketAsyncEventArgs _eventArgs = new SocketAsyncEventArgs();
+        private readonly SocketAwaitable _awaitable = new SocketAwaitable();
+
+        public SocketReceiver(Socket socket)
+        {
+            _socket = socket;
+            _eventArgs.UserToken = this;
+            _eventArgs.Completed += (_, e) => ((SocketReceiver)e.UserToken).ReceiveCompleted(e);
+        }
+
+        public SocketAwaitable ReceiveAsync(Buffer<byte> buffer)
+        {
+            if (!buffer.TryGetArray(out var segment))
+            {
+                throw new InvalidOperationException();
+            }
+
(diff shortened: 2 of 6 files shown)


src/Kestrel.Transport.Sockets/Internal/SocketReceiver.cs
```diff
@@ -0,0 +1,44 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Net.Sockets;
+
+namespace Microsoft.AspNetCore.Server.Kestrel.Transport.Sockets.Internal
+{
+    public class SocketReceiver
```
davidfowl: This might be of interest:

https://github.com/dotnet/corefx/blob/2427e05827c6e6f1eeaa77b01a3596f9a03582ff/src/System.Net.Sockets/src/System/Net/Sockets/NetworkStream.cs#L964

src/Kestrel.Transport.Sockets/Internal/SocketReceiver.cs
```diff
@@ -0,0 +1,44 @@
...
+
+            if (!_socket.ReceiveAsync(_eventArgs))
+            {
+                ReceiveCompleted(_eventArgs);
+            }
+
+            return _awaitable;
+        }
+
+        private void ReceiveCompleted(SocketAsyncEventArgs e)
```
davidfowl: Any reason this needs to be another method?

src/Kestrel.Transport.Sockets/Internal/SocketSender.cs
```diff
@@ -0,0 +1,108 @@
...
+            }
+
+            return _bufferList;
+        }
+
+        private static ArraySegment<byte> GetArraySegment(Buffer<byte> buffer)
+        {
+            if (!buffer.TryGetArray(out var segment))
+            {
+                throw new InvalidOperationException();
```
anurse: If this is something that we think could happen, let's at least put a message in place (even if it's a little obscure). If it's never going to happen unless we screw something up, maybe an Assert is better?

src/Kestrel.Transport.Sockets/Internal/SocketAwaitable.cs
```diff
@@ -0,0 +1,55 @@
...
+
+        private Action _callback;
+        private int _bytesTransfered;
+        private SocketError _error;
+
+        public SocketAwaitable GetAwaiter() => this;
+        public bool IsCompleted => _callback == _callbackCompleted;
+
+        public int GetResult()
+        {
```
anurse: If this is called when the task hasn't been completed, what happens? You get the data from the last result? Seems sketchy even for internal code. If a check on `_callback` is going to impact perf, maybe an Assert?

anurse: Though I see that the corefx version linked below doesn't do that check either. I don't know, maybe I don't care. Just seems safer to slap an assert in there or something.

src/Kestrel.Transport.Sockets/Internal/SocketReceiver.cs
```diff
@@ -0,0 +1,44 @@
...
+            _socket = socket;
+            _eventArgs.UserToken = this;
+            _eventArgs.Completed += (_, e) => ((SocketReceiver)e.UserToken).ReceiveCompleted(e);
+        }
+
+        public SocketAwaitable ReceiveAsync(Buffer<byte> buffer)
+        {
+            if (!buffer.TryGetArray(out var segment))
+            {
+                throw new InvalidOperationException();
```
anurse: Same as below comment. Put a message in here.

src/Kestrel.Transport.Sockets/Internal/SocketSender.cs
pakrym: There is an extension method for this: https://github.com/aspnet/KestrelHttpServer/blob/659fa967a1900653f7a82f02624c7c7995a3b786/src/Kestrel.Core/Internal/Http/PipelineExtensions.cs#L27

src/Kestrel.Transport.Sockets/Internal/SocketReceiver.cs
pakrym: https://github.com/aspnet/KestrelHttpServer/blob/659fa967a1900653f7a82f02624c7c7995a3b786/src/Kestrel.Core/Internal/Http/PipelineExtensions.cs#L27

anurse: Even better :)

src/Kestrel.Transport.Sockets/Internal/SocketReceiver.cs
halter73: No. It's a little more consistent with the sender class which does more in the `Completed` callback. Do you care?

src/Kestrel.Transport.Sockets/Internal/SocketAwaitable.cs
halter73: This class makes no attempt to validate that it's being called/used correctly. This is the same as LibuvAwaitable and AwaitableSocketAsyncEventArgs as you point out.

Would a `Debug.Assert(_callback == _callbackCompleted)` work for you? I guess we could do the same in LibuvAwaitable for consistency.

src/Kestrel.Transport.Sockets/Internal/SocketReceiver.cs
halter73: Defnitely similar. Is there anything you prefer about that implementation? Ours will always be slightly different. For example, we wrap our exceptions in IOExceptions after the awaiting the IO operation.

src/Kestrel.Transport.Sockets/Internal/SocketReceiver.cs
davidfowl: Yea, remove it.

src/Kestrel.Transport.Sockets/Internal/SocketAwaitable.cs
anurse: It would certainly make me feel a little better. But I get it, it's an internal component and we use it "right" so it's not a big thing.

src/Kestrel.Transport.Sockets/Internal/SocketReceiver.cs
davidfowl: I'm wondering if there's anywhere we can save allocations (like avoiding an extra object by deriving from SAEA)

PR #2149, opened 2017-11-07
Update "temporary" OpenSSL wrapper to support 1.1, and add HTTP/2 sample with docker

anurse:
Useful for reproducing issues, debugging issues, etc.

This includes a test cert in the right format and the necessary configuration options. Should just be `dotnet run`-able on a Linux machine with OpenSSL.


anurse:
:up: :date: ping @pakrym @halter73 

I also was able to quickly update the sample to have docker scripts. That'll be useful for automated spec testing (because we can use docker to run the spec tests), even when we are using the real ALPN support in SslStream. As part of that though, I had to update to OpenSSL 1.1 which is what our docker images use. It was pretty trivial though (the big change is that the initialization functions were removed and made automatic, which we can detect with exceptions, which is gross but works for this sample/hack)

src/Kestrel.Tls/TlsStream.cs
```diff
@@ -181,9 +182,16 @@ public override Task WriteAsync(byte[] buffer, int offset, int count, Cancellati
                     {
                         var error = OpenSsl.SSL_get_error(_ssl, ret);
 
-                        if (error != 2)
+                        if (error == 1)
                         {
-                            throw new IOException($"TLS handshake failed: {nameof(OpenSsl.SSL_do_handshake)} error {error}.");
+                            // SSL error, get it from the OpenSSL error queue
+                            error = OpenSsl.ERR_get_error();
+                            if (error == 336130204)
```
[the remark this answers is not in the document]
anurse: Enjoy :). I'll put a constant in before merging.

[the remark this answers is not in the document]
anurse: If this were anything except code that literally cannot be shipped I'd do something better. Also, it's just to clean up an ugly and hard to identify error message.

samples/Http2SampleApp/Dockerfile
```diff
@@ -0,0 +1,13 @@
+FROM microsoft/dotnet:2.0.0-runtime-stretch
```
halter73: Out of curiosity when would I choose this over `microsoft/aspnetcore:2.0.0-stretch` or vice versa?

samples/Http2SampleApp/Dockerfile
```diff
@@ -0,0 +1,13 @@
...
+
+RUN apt-get update && \
+    apt-get install -y --no-install-recommends libssl-dev && \
```
halter73: Why libssl-dev. Is libssl1.0.0 insufficient?

samples/Http2SampleApp/Program.cs
```diff
@@ -0,0 +1,50 @@
...
+                    factory.AddConsole();
+                })
+                .UseKestrel(options =>
+                {
+                    // Run callbacks on the transport thread
+                    options.ApplicationSchedulingMode = SchedulingMode.Inline;
+
+                    options.Listen(IPAddress.Any, basePort, listenOptions =>
+                    {
+                        listenOptions.Protocols = HttpProtocols.Http1AndHttp2;
```
halter73: Could we just add this to our main sample app and use a command line argument to control whether or not to support HTTP/2?

samples/Http2SampleApp/cert/testCert.key
```diff
@@ -0,0 +1,51 @@
+-----BEGIN RSA PRIVATE KEY-----
```
halter73: Assuming that TlsStream cannot be modified to accept testCert.pfx, I would just throw this in `test/shared/TestCertificates`.

samples/Http2SampleApp/scripts/build-docker.sh
```diff
@@ -0,0 +1,6 @@
+#!/usr/bin/env bash
+DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+
+dotnet publish --framework netcoreapp2.0 "$DIR/../Http2SampleApp.csproj"
```
halter73: Nit: Is `DIR` necessary since you've already `cd`d into it?

src/Kestrel.Core/Internal/Http2/Http2Frame.cs
```diff
@@ -66,5 +67,21 @@ public int StreamId
         }
 
         public Span<byte> Payload => new Span<byte>(_data, PayloadOffset, Length);
+
+        public void TraceFrame(ILogger logger, string connectionId)
```
halter73: Where is this called?

samples/Http2SampleApp/scripts/build-docker.sh
anurse: Nope, the above code doesn't `cd` in the main context, it creates a subshell with `$()`

src/Kestrel.Core/Internal/Http2/Http2Frame.cs
anurse: From a different change that I didn't clean up. Will remove.

samples/Http2SampleApp/cert/testCert.key
anurse: Sure. It may be possible to modify it but since a) it's temporary and b) I don't know how to OpenSSL very well, I don't feel like investing time to do it :).

samples/Http2SampleApp/Dockerfile
anurse: It may not be, I'll try that :).

samples/Http2SampleApp/Dockerfile
anurse: I suppose it's OK. That will bring in the runtime store, but I think that's probably fine.

samples/Http2SampleApp/scripts/run-docker.sh
```diff
@@ -0,0 +1,4 @@
+#!/usr/bin/env bash
+DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
```
halter73: Nit: DIR isn't used.

samples/Http2SampleApp/Program.cs
anurse: I'm hesitant to do that right now since the Sample app already has HTTPS configuration using PFX. Once we get the ALPN SslStream stuff merged we can look at merging the samples.

samples/Http2SampleApp/scripts/run-docker.sh
anurse: True, I'll remove it from this. Just copy 🍝

anurse:
Stand by for one more iteration that adds SSH necessary for remote debugging.

samples/Http2SampleApp/Program.cs
halter73: That's fine. We could switch from UseHttps to UseTls in the main sample when HTTP/2 is selected, but no big deal either way.

samples/Http2SampleApp/cert/testCert.key
Drawaes: @anurse hold my beer ;)
#2150

anurse:
Went around in circles enabling SSH and I gave up. Addressing PR feedback now

samples/Http2SampleApp/cert/testCert.key
anurse: Hey, if you want to invest the time into our never-gonna-ship code, I'm not going to complain :)

anurse:
> Went around in circles enabling SSH and I gave up.

Btw, turns out my problems with this may have been a problem with Docker for Windows and may work when I try it next time. But I'm not bogging down this PR with that yet again ;)

samples/Http2SampleApp/Http2SampleApp.csproj
```diff
@@ -0,0 +1,21 @@
...
+    <ProjectReference Include="..\..\src\Kestrel\Kestrel.csproj" />
+    <ProjectReference Include="..\..\src\Kestrel.Tls\Kestrel.Tls.csproj" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.Extensions.Logging.Console" Version="$(MicrosoftExtensionsLoggingConsolePackageVersion)" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <Content Include="../../test/shared/TestCertificates/testCert.pfx" CopyToOutputDirectory="PreserveNewest" />
```
halter73: You're changing the main SampleApp to use this cert to, right?

anurse: Yes, I'll do that as well

anurse:
Any further thoughts here? The build failures all seem to be unrelated flaky tests

Issue #2139, opened 2017-10-27
Add KestrelServerOptions.ListenLocalhost


halter73:

Tratcher:
Subset of https://github.com/aspnet/KestrelHttpServer/issues/1290 ?

halter73:
Not exactly a subset since it's largely orthogonal to config.

PR #2158, opened 2017-11-14
[WIP] Add ListenLocalhost and ListenPrefix

Tratcher:
#2139, #1290
[Widows update took over my machine before I could finish the tests. Starting this PR so we can review the overall approach.]

This adds `options.ListenLocalhost(int port);` and `options.ListenPrefix(string prefix);` and the associated config lambda overloads.

ListenLocalhost listens on both IPv4 & IPv6 loopback adapters, and can add https either with the default cert or a manual cert. This mirrors the functionality of .UseUrls("https://localhost:{port}") with the addition of ListenOptions.

ListenPrefix is a more generic version of ListenLocalhost that allows any syntax from UseUrls and adds the ability to configure the endpoint settings like protocol and certificate. This will be useful for the upcoming configuration work.

### DIFF
--- a/samples/SampleApp/Startup.cs
+++ b/samples/SampleApp/Startup.cs
@@ -73,7 +73,19 @@ namespace SampleApp
                     options.Listen(IPAddress.Loopback, basePort + 1, listenOptions =>
                     {
                         listenOptions.UseHttps("testCert.pfx", "testPassword");
-                        listenOptions.UseConnectionLogging();
+                    });
+
+                    options.ListenLocalhost(basePort + 2);
+
+                    // Use the default certificate
+                    options.ListenLocalhost(basePort + 3, useHttps: true);
+
+                    // Use the default certificate
+                    options.ListenPrefix($"https://*:{basePort + 4}");
+
+                    options.ListenPrefix($"https://*:{basePort + 5}", listenOptions =>
+                    {
+                        listenOptions.UseHttps("testCert.pfx", "testPassword");
                     });
 
                     options.UseSystemd();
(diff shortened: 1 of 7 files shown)


samples/SampleApp/Startup.cs
```diff
@@ -73,7 +73,19 @@ public static Task Main(string[] args)
...
-                        listenOptions.UseConnectionLogging();
+                    });
+
+                    options.ListenLocalhost(basePort + 2);
+
+                    // Use the default certificate
+                    options.ListenLocalhost(basePort + 3, useHttps: true);
+
+                    // Use the default certificate
+                    options.ListenPrefix($"https://*:{basePort + 4}");
```
halter73: What if you want to use the default certificate with `Listen(IPEndPoint)` or `ListenLocalhost`. Are you forced to convert your endpoint to a string and use `ListenPrefix`.

Why not just add a `UseHttps` overload that uses the default certificate? This way you can remove ListenPrefix entirely, and instead of adding `ListenType.Prefix` you can just add `ListenType.Loopback` which I prefer greatly.

test/Kestrel.FunctionalTests/AddressRegistrationTests.cs
```diff
@@ -184,10 +195,119 @@ public AddressRegistrationTests(ITestOutputHelper output)
...
+                    options.ListenPrefix(prefix + ":" + port);
+                })
+                .ConfigureLogging(_configureLoggingDelegate)
+                .Configure(ConfigureEchoAddress);
+
+            using (var host = hostBuilder.Build())
+            {
+                host.Start();
+
+                foreach (var testUrl in testUrls.Select(testUrl => $"{testUrl}:{port == 0 ? host.GetPort() : port}"))
```
halter73: You have to wrap the ternary expression in parens. Otherwise the compiler thinks the `:` is a format specifier delimiter.

PR #2159, opened 2017-11-14
Add ListenLocalhost and ListenAnyIP

Tratcher:
#2139 Still needs tests

This abstracts the address parser to work independently of the binder so it can be used for future config code. It also adds first class support for two scenarios formerly only available via UseUrls, ListenLocalhost and ListenAnyIP. Both of these provide dynamic fallback behavior that's not was not possible with a single ListenOptions before.

### DIFF
--- a/src/Kestrel.Core/Internal/AddressBinder.cs
+++ b/src/Kestrel.Core/Internal/AddressBinder.cs
@@ -47,17 +47,6 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal
             await strategy.BindAsync(context).ConfigureAwait(false);
         }
 
-        private class AddressBindContext
-        {
-            public ICollection<string> Addresses { get; set; }
-            public List<ListenOptions> ListenOptions { get; set; }
-            public KestrelServerOptions ServerOptions { get; set; }
-            public ILogger Logger { get; set; }
-            public IDefaultHttpsProvider DefaultHttpsProvider { get; set; }
-
-            public Func<ListenOptions, Task> CreateBinding { get; set; }
-        }
-
         private static IStrategy CreateStrategy(ListenOptions[] listenOptions, string[] addresses, bool preferAddresses)
         {
             var hasListenOptions = listenOptions.Length > 0;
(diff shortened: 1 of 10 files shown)


src/Kestrel.Transport.Abstractions/Internal/ListenType.cs
```diff
@@ -9,6 +9,8 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Transport.Abstractions.Internal
     public enum ListenType
     {
         IPEndPoint,
+        Localhost,
```
davidfowl: This makes me super sad

halter73: I understand the feeling. Still better than "Prefix" though.

Tratcher:
Rebased and updated with tests, ready for review.

src/Kestrel.Transport.Abstractions/Internal/ListenType.cs
davidfowl: I thought about it, I think this is too big of a hack to exist. It doesn't fit the current model. This enum needs to be something else and not part of ListenType.

Tratcher: Such as? I could drive from Listen options and type check. Or add a second enum IPMode with values LocalhostSplit, AnyIPFallback, and Single. Or derive from ListenOptions and put the bind method on that type. Hmm, that might work...

src/Kestrel.Core/ListenOptions.cs
```diff
@@ -182,5 +183,11 @@ public ConnectionDelegate Build()
 
             return app;
         }
+
+        internal virtual async Task BindAsync(AddressBinder.AddressBindContext context)
```
halter73: Nit: unnest AddressBindContext

src/Kestrel.Core/Internal/AddressBinder.cs
```diff
@@ -126,60 +123,7 @@ private static Task BindEndpointAsync(IPEndPoint endpoint, AddressBindContext co
...
-                throw new IOException(CoreStrings.FormatAddressBindingFailed(address), new AggregateException(exceptions));
-            }
-
-            // If StartLocalhost doesn't throw, there is at least one listener.
-            // The port cannot change for "localhost".
-            context.Addresses.Add(address.ToString());
-        }
-
-        private static async Task BindAddressAsync(string address, AddressBindContext context)
+        internal static ListenOptions ParseAddress(string address, KestrelServerOptions serverOptions, IDefaultHttpsProvider defaultHttpsProvider)
```
halter73: 👍

PR #2169, opened 2017-11-16
Update samples and tests to target netcoreapp2.1

pranavkm:


test/Directory.Build.props
```diff
@@ -2,6 +2,13 @@
   <Import Project="..\Directory.Build.props" />
 
   <PropertyGroup>
+    <DeveloperBuildTestTfms>netcoreapp2.1</DeveloperBuildTestTfms>
```
pranavkm: I recall you saying you preferred folks always ran desktop tests locally. Is that still a thing or would you rather we follow the same pattern we use elsewhere?

halter73: As long as it's easy to run desktop tests and the desktop tests are still run on the CI, I'm OK with making running desktop test non-default locally.

pranavkm: It defaults to running tests in 3 tfms. You can opt in to running it only on netcoreapp2.1.

PR #2170, opened 2017-11-16
Updating toolset compiler to 2.6.0-beta3-62316-02

VSadov:
Nothing significantly changed. 
Just a newer version to match CoreFxLab and CoreFx repos.


davidfowl:
@VSadov is there a version of the .NET CLI with this compiler yet? I'd like to stop using this package because it's making our builds extremely slow (no crossgen)

davidfowl:
See https://github.com/aspnet/KestrelHttpServer/commit/01e2c881317ef3473b313a01eb194f6f418b11b6

VSadov:
Thanks.
Now that there is no active feature development and there is an SDK with 7.2, using SDK seems the right thing to do.

PR #2173, opened 2017-11-20
Use new Memory<byte> APIs on SocketAsyncEventArgs

davidfowl:
- This should improve the performance of handling buffers by (eventually)
removing GCHandle churn for Kestrel's already pinned buffers.
- Made the Sockets transport target both netcoreapp2.1 and netstandard2.0 to use new
APIs

Contributes to https://github.com/aspnet/KestrelHttpServer/issues/1686

/cc @stephentoub @geoffkizer

### DIFF
--- a/src/Kestrel.Transport.Sockets/Kestrel.Transport.Sockets.csproj
+++ b/src/Kestrel.Transport.Sockets/Kestrel.Transport.Sockets.csproj
@@ -4,7 +4,7 @@
     <AssemblyName>Microsoft.AspNetCore.Server.Kestrel.Transport.Sockets</AssemblyName>
     <RootNamespace>Microsoft.AspNetCore.Server.Kestrel.Transport.Sockets</RootNamespace>
     <Description>Managed socket transport for the ASP.NET Core Kestrel cross-platform web server.</Description>
-    <TargetFramework>netstandard2.0</TargetFramework>
+    <TargetFrameworks>netstandard2.0;netcoreapp2.1</TargetFrameworks>
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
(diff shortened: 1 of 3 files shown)


davidfowl:
Seems like API check can't handle netcoreapp2.1.

src/Kestrel.Transport.Sockets/Kestrel.Transport.Sockets.csproj
```diff
@@ -4,11 +4,12 @@
     <AssemblyName>Microsoft.AspNetCore.Server.Kestrel.Transport.Sockets</AssemblyName>
     <RootNamespace>Microsoft.AspNetCore.Server.Kestrel.Transport.Sockets</RootNamespace>
     <Description>Managed socket transport for the ASP.NET Core Kestrel cross-platform web server.</Description>
-    <TargetFramework>netstandard2.0</TargetFramework>
+    <TargetFrameworks>netstandard2.0;netcoreapp2.1</TargetFrameworks>
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
     <PackageTags>aspnetcore;kestrel</PackageTags>
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
     <NoWarn>CS1591;$(NoWarn)</NoWarn>
+    <EnableAPICheck>false</EnableAPICheck>
```
davidfowl: I disabled this for 2 reasons:
- We haven't shipped this package yet.
- API check is broken for netcoreapp2.1 projects. I submitted a PR for this.

halter73: This PR didn't change anything APICheck would complain about, right? Either way, I created #2175 to track reenabling this once we ship.

PR #2176, opened 2017-11-20
Use MSBuild to set NuGet feeds instead of NuGet.config

natemcmaster:
Part of https://github.com/aspnet/Coherence-Signed/issues/682

Required for orchestrated build so we can conditionally exclude myget while restoring on CI.

### DIFF
--- a/NuGet.config
+++ b/NuGet.config
@@ -2,9 +2,6 @@
 <configuration>
   <packageSources>
     <clear />
-    <add key="AspNetCore" value="https://dotnet.myget.org/F/aspnetcore-ci-dev/api/v3/index.json" />
-    <add key="AspNetCoreTools" value="https://dotnet.myget.org/F/aspnetcore-tools/api/v3/index.json" />
-    <add key="roslyn" value="https://dotnet.myget.org/F/roslyn/api/v3/index.json" />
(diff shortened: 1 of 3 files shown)


NuGet.config
```diff
@@ -2,9 +2,6 @@
 <configuration>
   <packageSources>
     <clear />
-    <add key="AspNetCore" value="https://dotnet.myget.org/F/aspnetcore-ci-dev/api/v3/index.json" />
-    <add key="AspNetCoreTools" value="https://dotnet.myget.org/F/aspnetcore-tools/api/v3/index.json" />
-    <add key="roslyn" value="https://dotnet.myget.org/F/roslyn/api/v3/index.json" />
-    <add key="NuGet" value="https://api.nuget.org/v3/index.json" />
+    <!-- Restore sources should be defined in build/sources.props. -->
```
halter73: Will `dotnet restore` inside of a project still use the NuGet feeds defined in build/sources.props?

natemcmaster: Yes.

natemcmaster: To clarify, the `RestoreSources` property is a feature of NuGet, not KoreBuild or any particular build system.

Issue #1334, opened 2017-01-31
Determining Hosting Environment While Configuring Kestrel and UseHttps


RehanSaeed:
In the ASP.NET Core Main method below, how can I determine the hosting environment, so I can switch between different certificate files for HTTPS?

```
public sealed class Program
{
    public static void Main(string[] args)
    {
        new WebHostBuilder()
            .UseContentRoot(Directory.GetCurrentDirectory())
            .UseKestrel(
                options =>
                {
                    if ([Development Hosting Environment])
                    {
                        options.UseHttps("DevelopmentCertificate.pfx");
                    }
                    else
                    {
                        options.UseHttps("ProductionCertificate.pfx");
                    }
                })
            .UseIISIntegration()
            .UseStartup<Startup>()
            .Build()
            .Run();
    }
}
```

GuardRex:
It was just answered on another thread that I just saw: https://github.com/aspnet/Home/issues/1915#issuecomment-276133606

... and if you don't use command-line config, env var config, a config file, etc., then you can read the `ASPNETCORE_ENVIRONMENT` env var directly after you manually set it somehow.

RehanSaeed:
It turns out you can use `ConfigureServices` to get hold of `IHostingEnvironment` like so:

```
public sealed class Program
{
    public static void Main(string[] args)
    {
        IHostingEnvironment hostingEnvironment = null;
        new WebHostBuilder()
            .UseContentRoot(Directory.GetCurrentDirectory())
            .ConfigureServices(
                services =>
                {
                    hostingEnvironment = services
                        .Where(x => x.ServiceType == typeof(IHostingEnvironment))
                        .Select(x => (IHostingEnvironment)x.ImplementationInstance)
                        .First();
                })
            .UseKestrel(
                options =>
                {
                    if (hostingEnvironment.IsDevelopment())
                    {
                        // Use a self-signed certificate to enable 'dotnet run' to work in development.
                        options.UseHttps("DevelopmentCertificate.pfx", "password");
                    }
                })
            .UseIISIntegration()
            .UseStartup<Startup>()
            .Build()
            .Run();
    }
}
```

It really feels like this should be easier.

muratg:
Assigning to @cesarbs for investigation

dotnetshadow:
@RehanSaeed How can you handle the .UseUrls() to wrap an if statment around it?

`
// Only add if development??
.UseUrls("https://localhost:5001")
`

[...]

PR #2177, opened 2017-11-21
Expose WebHostBuilderContext in UseKestrel

Tratcher:
#1334 @halter73 I've poached this because it will make the configuration work easier.

### DIFF
--- a/samples/SampleApp/Startup.cs
+++ b/samples/SampleApp/Startup.cs
@@ -43,22 +43,15 @@ namespace SampleApp
                 Console.WriteLine("Unobserved exception: {0}", e.Exception);
             };
 
-            var configuration = new ConfigurationBuilder()
-                .AddEnvironmentVariables()
-                .Build();
-
-            if (!ushort.TryParse(configuration["BASE_PORT"], NumberStyles.None, CultureInfo.InvariantCulture, out var basePort))
-            {
-                basePort = 5000;
-            }
-
             var hostBuilder = new WebHostBuilder()
                 .ConfigureLogging((_, factory) =>
                 {
                     factory.AddConsole();
                 })
-                .UseKestrel(options =>
+                .UseKestrel((context, options) =>
                 {
+                    var basePort = context.Configuration.GetValue<int?>("BASE_PORT") ?? 5000;
+
                     // Run callbacks on the transport thread
                     options.ApplicationSchedulingMode = SchedulingMode.Inline;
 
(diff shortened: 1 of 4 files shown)


src/Kestrel/WebHostBuilderKestrelExtensions.cs
```diff
@@ -57,5 +57,31 @@ public static IWebHostBuilder UseKestrel(this IWebHostBuilder hostBuilder, Actio
...
+        /// <param name="hostBuilder">
+        /// The Microsoft.AspNetCore.Hosting.IWebHostBuilder to configure.
+        /// </param>
+        /// <param name="configureOptions">A callback to configure Kestrel options.</param>
+        /// <returns>
+        /// The Microsoft.AspNetCore.Hosting.IWebHostBuilder.
+        /// </returns>
+        public static IWebHostBuilder UseKestrel(this IWebHostBuilder hostBuilder, Action<WebHostBuilderContext, KestrelServerOptions> configureOptions)
+        {
+            if (configureOptions == null)
```
JunTaoLuo: Should this null check be added to the extension above as well?

Tratcher: services.Configure does that check.

samples/SampleApp/Startup.cs
```diff
@@ -57,8 +57,8 @@ public static Task Main(string[] args)
                 {
                     factory.AddConsole();
                 })
-                .UseKestrel(options =>
                 {
+                .UseKestrel((context, options) =>
```
JunTaoLuo: This doesn't look right

davidfowl:
I wonder if this would be better to put on the KestrelServerOptions object itself.

Tratcher:
```
                .UseKestrel(options =>
                  {
                    var basePort = options.HostContext.Configuration.GetValue<int?>("BASE_PORT") ?? 5000;
                    if (options.HostContext.HostingEnvironment.IsDevelopment()) { ... }
```

*shrug*

It may mess up `services.Configure<KestrelServerOptions>(options => ...)`, or you'd have to capture it like you capture the services today.

davidfowl:
> It may mess up services.Configure<KestrelServerOptions>(options => ...), or you'd have to capture it like you capture the services today.

Is the WebHostBuilderContext in the container?

Tratcher:
No

davidfowl:
Then this is fine

PR #2179, opened 2017-11-23
Use ALPN support in SSLStream on .NET Core >= 2.1

davidfowl:
- Always add the TlsConnectionFeature when the HttpsConnectionAdapter runs
- Implemented the ITlsApplicationProtocolsFeature on the existing TlsConnectionFeature
- Removed Kestrel.Tls


src/Kestrel.Https/Internal/HttpsConnectionAdapter.cs
```diff
@@ -134,11 +163,24 @@ public Task<IAdaptedConnection> OnConnectionAsync(ConnectionAdapterContext conte
...
 
-            // Always set the feature even though the cert might be null
-            context.Features.Set<ITlsConnectionFeature>(new TlsConnectionFeature
+#if NETCOREAPP2_1
+            // Don't allocate in the common case, see https://github.com/dotnet/corefx/issues/25432
+            if (sslStream.NegotiatedApplicationProtocol == SslApplicationProtocol.Http11)
             {
-                ClientCertificate = ConvertToX509Certificate2(sslStream.RemoteCertificate)
-            });
+                feature.ApplicationProtocol = "http/1.1";
```
anurse: Should these be raw byte arrays? Or `SslApplicationProtocol` instances? The spec is actually pretty clear that ALPN values are "byte strings" and not text strings; the current known values just happen to map to UTF-8-encoded strings. If there's ever a reason to represent an ALPN value as a non-printable string sequence, our feature will have to change.

It's always compared against known constants anyway, so it's not like it matters much what the type is.

src/Kestrel.Https/Internal/HttpsConnectionAdapter.cs
```diff
@@ -114,8 +119,32 @@ public Task<IAdaptedConnection> OnConnectionAsync(ConnectionAdapterContext conte
 
             try
             {
+#if NETCOREAPP2_1
+                var sslOptions = new SslServerAuthenticationOptions()
+                {
+                    ServerCertificate = _serverCertificate,
```
Tratcher: There's also a new option to disable renegotiate that you need to set.

src/Kestrel.Https/Internal/HttpsConnectionAdapter.cs
davidfowl: That's ok, I'm not doing that as a part of this change. We can file an issue if we think it's important.

1. It could be a byte[] or Span\<byte\> or whatever makes sense.
2. It can't be SslApplicationProtocol because it needs to be .NET Standard 2.0.

Tratcher: Yes, file it.

anurse: I figured (2) was the case but thought I'd throw it out there. I'll file a bug though since it's a trivial fix and seems appropriate. Probably has to be `Memory<byte>` because it's an interface?

davidfowl: https://github.com/aspnet/KestrelHttpServer/issues/2181

Issue #2151, opened 2017-11-08
BadHttpRequestException at Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http.HttpParser`1.RejectRequestLine


npehrsson:
We're experiencing some strange behavior a couple of times per day.

The behavior is that we are receiving exceptions about bad requests somewhere in the core when it tries to parse the Http Method.

We see it from both an integration and from random web UI requests.
Good to know is that these errors are always occurring during higher load (from the integration) and in bursts.

We have not been able to reproduce it, the burst of errors occur 1-2 a day and almost every day.

The integration is using HttpClient.PostAsync, we have tried to resend one of those requests (the integration is logging what it sends) that have failed, but we couldn't reproduce it that way.

Stack
```
{"assembly":"Microsoft.AspNetCore.Server.Kestrel.Core, Version=2.0.0.0, Culture=neutral, PublicKeyToken=adb9793829ddae60","method":"Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http.HttpParser`1.RejectRequestLine","level":0,"line":0}

{"assembly":"Microsoft.AspNetCore.Server.Kestrel.Core, Version=2.0.0.0, Culture=neutral, PublicKeyToken=adb9793829ddae60","method":"Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http.HttpParser`1.GetUnknownMethod","level":1,"line":0}

{"assembly":"Microsoft.AspNetCore.Server.Kestrel.Core, Version=2.0.0.0, Culture=neutral, PublicKeyToken=adb9793829ddae60","method":"Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http.HttpParser`1.ParseRequestLine","level":2,"line":0}

{"assembly":"Microsoft.AspNetCore.Server.Kestrel.Core, Version=2.0.0.0, Culture=neutral, PublicKeyToken=adb9793829ddae60","method":"Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http.HttpParser`1.ParseRequestLine","level":3,"line":0}

{"assembly":"Microsoft.AspNetCore.Server.Kestrel.Core, Version=2.0.0.0, Culture=neutral, PublicKeyToken=adb9793829ddae60","method":"Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http.Frame.TakeStartLine","level":4,"line":0}

[...]

PR #2180, opened 2017-11-23
Proposed fix for #2151

davidfowl:
- ~Tweak the SegmentEnumerator to avoid looking at empty segments~
- Changed the ReadableBufferReader's Cursor computation to avoid special
casing end.
- The issue was caused by a mix of things and the interaction with pipes.
The ReadableBufferReader was moving to the next block but also skipping
the _index update since it special cases when you've reached the end of the
buffer. As a result, the cursor ended up pointing to the end of the empty
block and the pipe ended up having a negative length for the buffer.

PS: I know we need to go through the patch process but it was easier to review this PR since the code was all in a single place.


davidfowl:
I think I know what the issue is. Will update later.

pakrym:
I like this so much more then the previous iteration but you should do it in corefxlab first so it runs through pipelines unit tests.

shared/corefxlab/System.IO.Pipelines/ReadableBufferReader.cs
```diff
@@ -92,12 +87,13 @@ private void MoveNext()
         {
             while (_enumerator.MoveNext())
             {
+                // Always reset the index to 0 if we changed the current segment
+                _index = 0;
                 var part = _enumerator.Current;
                 var length = part.Length;
                 if (length != 0)
                 {
                     _currentSpan = part.Segment.Buffer.Span.Slice(part.Start, length);
```
pakrym: Should `_currentSpan` be reset to `default` if length if 0.

shared/corefxlab/System.IO.Pipelines/ReadableBufferReader.cs
```diff
@@ -43,11 +43,6 @@ public ReadCursor Cursor
             {
                 var part = _enumerator.Current;
 
-                if (_end)
-                {
-                    return new ReadCursor(part.Segment, part.Start + _currentSpan.Length);
```
pakrym: This is really strange, now when I look at it.

davidfowl:
> I like this so much more then the previous iteration but you should do it in corefxlab first so it runs through pipelines unit tests.

I'm currently doing that, writing a test a bit tricky but that's it's coming.

shared/corefxlab/System.IO.Pipelines/ReadableBufferReader.cs
davidfowl: The assumption was that end = end of the current segment. Turns out it can also mean beginning of an empty segment.

shared/corefxlab/System.IO.Pipelines/ReadableBufferReader.cs
```diff
@@ -123,7 +119,11 @@ public void Skip(int length)
                     break;
                 }
 
-                length -= (_currentSpan.Length - _index);
+                var remaining = (_currentSpan.Length - _index);
```
halter73: Nit: the parens are weird now.

davidfowl: Yah.

davidfowl:
🆙 📅

PR #2184, opened 2017-11-27
Fix for #2129

### LINKED ISSUE #2129: Kestrelserver ReasonPhrase formatting
I know that reason phrase is not in Http/2 anymore and even optional in Http/1.1, but according to RFC spec the current calculation is wrong:
https://github.com/aspnet/KestrelHttpServer/blob/3fbfba63f813de004030615546546360b565b8d5/src/Kestrel.Core/Internal/Http/ReasonPhrases.cs#L223-L227

According to the RFC https://tools.ietf.org/html/rfc7230#section-3.1.2 you need to print the space regardless if the reason phrase is empty.
>  status-line = HTTP-version SP status-code SP reason-phrase CRLF

I know this is nit-picking, but I actually stumbled across while writing my http batch handler middleware.

Tornhoof:
This pull request fixes #2129.
- Add the whitespace even if the reason phrase is empty
- Add a comment to explain it
- Add tests (in new class, couldn't find any test class for it)

### DIFF
--- a/test/Kestrel.Core.Tests/ReasonPhrasesTests.cs
+++ b/test/Kestrel.Core.Tests/ReasonPhrasesTests.cs
@@ -3,26 +3,20 @@
 
 using Xunit;
 using Microsoft.AspNetCore.Http;
+using System.Text;
 
 namespace Microsoft.AspNetCore.Server.Kestrel.Core.Tests
 {
     public class ReasonPhraseTests
     {
         [Theory]
-        [InlineData(999, "Unknown")]
-        [InlineData(999, null)]
-        public void UnknownStatusCodes(int statusCode, string reasonPhrase)
+        [InlineData(999, "Unknown", "999 Unknown")]
+        [InlineData(999, null, "999 ")]
+        public void UnknownStatusCodes(int statusCode, string reasonPhrase, string expectedResult)
         {
             var bytes = Internal.Http.ReasonPhrases.ToStatusBytes(statusCode, reasonPhrase);
             Assert.NotNull(bytes);
-            if (string.IsNullOrEmpty(reasonPhrase))
-            {
-                Assert.Equal(0x20, bytes[bytes.Length - 1]);
-            }
-            else
-            {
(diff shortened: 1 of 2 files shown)


test/Kestrel.Core.Tests/ReasonPhrasesTests.cs
```diff
@@ -0,0 +1,38 @@
...
+using Xunit;
+using Microsoft.AspNetCore.Http;
+
+namespace Microsoft.AspNetCore.Server.Kestrel.Core.Tests
+{
+    public class ReasonPhraseTests
+    {
+        [Theory]
+        [InlineData(999, "Unknown")]
+        [InlineData(999, null)]
```
halter73: Could put the entire expected output in the InlineData? Then you could just do `Assert.Equal(expectedString, Encoding.ASCII.GetString(bytes))`.

Tornhoof: done.

test/Kestrel.Core.Tests/ReasonPhrasesTests.cs
```diff
@@ -0,0 +1,32 @@
...
+        {
+            var bytes = Internal.Http.ReasonPhrases.ToStatusBytes(statusCode, reasonPhrase);
+            Assert.NotNull(bytes);
+            Assert.Equal(expectedResult, Encoding.ASCII.GetString(bytes));
+        }
+
+        [Theory]
+        [InlineData(StatusCodes.Status200OK, "OK")]
+        [InlineData(StatusCodes.Status200OK, null)]
+        public void KnownStatusCodes(int statusCode, string reasonPhrase)
```
Tratcher: Test KnownStatusCodes the same way you test UnknownStatusCodes

Tornhoof: Yes, makes sense. Since both methods are then the same, I consolidated them and renamed the test appropriately.

halter73:
Thanks @Tornhoof

Issue #2167, opened 2017-11-16
No way to configure the endpoint options when using PreferServerUrls


Tratcher:
I wanted to turn on `listenOptions.UseConnectionLogging();` behind ANCM but there's no way to do this, PreferServerUrls overrides all Kestrel ListenOptions.

halter73:
Hmm. This brings me back to the idea of having something like `KestrelServerOptions.ConfigureAllEndpoints(Action<ListenOptions>)`.

I don't love it, but it's my best idea so far other than just not supporting connection middleware for "ServerUrls".

Tratcher:
Ordering will be an interesting problem for shared middleware no matter what.

Many of the other options will be simpler, we can use that callback (first, so you can override it) or KSO can have properties for defaults like DefaultProtocols.

Issue #2166, opened 2017-11-15
Add a UseHttps overload to allow using the default certificate


Tratcher:
Today the default dev certificate is only accessible via the UseUrls configuration path. It would be helpful if the Listen APIs could also request the default certificate.

We may also want to provide a first class way of replacing the default certificate in code. This would allow the app to use the dev default in development and specify a different default in production.

Issue #2016, opened 2017-08-24
HTTPS binding configuration for Kestrel


javiercn:
We need to figure out how config for HTTPS will work for Kestrel in development. (Config format for setting up an HTTPS binding and how will tooling setup the bindings during development). 

This is part of our plan for HTTPS in 2.1

muratg:
@cesarbs assigning this to you.  I suspect we'll need to design this.  cc @shirhatti and @davidfowl

Tratcher:
How much of this is just bringing back the config feature we cut from 2.0?

muratg:
@shirhatti will schedule a design meeting to plan this work.

muratg:
@shirhatti where are we with this one?

muratg:
@anurse could you sync-up with @Tratcher?

anurse:
I haven't really had a chance to work on this beyond what I've already sent out. We can sync up on Monday though if there are any gaps that haven't been filled.

javiercn:
We have the dev part done. We need to figure out configuration for production scenarios

danroth27:
@Tratcher Is this a dup of #1290?

danroth27:
Nevermind - https://github.com/aspnet/KestrelHttpServer/pull/2186 clarifies that #1290 is tracking "config and code parity".

PR #2186, opened 2017-11-28
Implement configuration support

Tratcher:
This includes the updated design. If you like it then I'll start working on the tests.

Related issues:
https://github.com/aspnet/KestrelHttpServer/issues/2016 - Config support
https://github.com/aspnet/KestrelHttpServer/issues/2167 - Default settings
https://github.com/aspnet/KestrelHttpServer/issues/2166 - UseHttps()
https://github.com/aspnet/KestrelHttpServer/issues/1290 - Config and code parity
https://github.com/aspnet/KestrelHttpServer/issues/2188 - Listen on https by default

Major features:
- KestrelServerOptions.ConfigureEndpointDefaults sets a callback that will be called for every new endpoint no matter what source.
- KestrelServerOptions.ConfigureHttpsDefaults sets a callback that will be called to configured every new Https adapter. This is primarily useful for endpoints config.
- KestrelServerOptions.Configure(IConfiguration) creates a builder that reads config and registers endpoints. This includes the .Endpoint method to configure additional settings for individually named endpoints from config.
- New ListenOptions.UseHttps extensions to provide code and config functional symmetry, and overloads with configureOptions so you don't have to re-invent cert loading when you want to configure other settings.

Supported configurations:

- No config: Listens on http://localhost:5000 and https://localhost:5001 (if a default cert is available)
- Configure Urls: Urls can be specified by ASNETCORE_URLS environment variable, Urls command line variable, or Urls top level key in hosting's configuration, or via the `UseUrls` IWebHostBuilder extension. These can be http or https (if a default cert is available). This may be a semicolon separated list. E.g. `"Urls": "http://localhost:5000;http://localhost:5001"`.
- Default cert from config: The default certificate can be replaced from configuration. Call `serverOptions.Configure(context.Configuration.GetSection("Kestrel"))` to specify the configuration section (this will be added to the metapackage). Then add the following config:
```
{
  "Kestrel": {
    "Certificates": {
       "Default": {
          "Path": "...",
          "Password": ""
       }
    }
  }
}
```
Certificates can also be loaded from the machine store by specifying the following parameters instead of Path and Password:
```
  "Subject": "Test Subject", // Required
  "Store": "My", // StoreName defaults to My
  "Location": "CurrentUser/LocalMachine", // StoreLocation defaults to CurrentUser
  "AllowInvalid": true // defaults to false. Setting this to true allows for self-signed or other non-trusted certificates.
```
- Change defaults in code: ConfigureEndpointDefaults and ConfigureHttpsDefaults can be used to change default settings for ListenOptions and HttpsConnectionAdapterOptions, including overriding the default certificate specified in the prior scenario.
```
options.ConfigureEndpointDefaults(opt =>
{
    opt.Protocols = HttpProtocols.Http1;
});

options.ConfigureHttpsDefaults(httpsOptions =>
{
    httpsOptions.SslProtocols = SslProtocols.Tls12;
});
```
- Configure endpoints: Named endpoints can be added via configuration with the following format:
```
  "Kestrel": {
	"Endpoints": {
	   "HTTP": { "Url": "http://*:5000" },
           "HTTPS": {
              "Url": "https://*:5463",
              "Certificate": {
                  "Path": "testCert.pfx",
                  "Password": "testPassword"
              }
          }
    }
```
Schema details:
1. The endpoints above are named HTTP and HTTPS. These names are arbitrary but will become useful in the next example. These names are case-insensitive.
2. The Url parameter is required for each endpoint. The format for this parameter is the same as the top level Urls configuration parameter except that it's limited to a single value.
3. These endpoints replace those defined in the top level Urls configuration rather than adding to them. Endpoints defined in code via Listen are cumulative with the endpoints defined in this configuration section.
4. The Certificate section is optional, if not specified then the defaults defined in earlier scenarios will be used. If no defaults are available then the server will throw and fail to start.
5. The Certificate section supports both the Path&Password and the Subject&Store type of certificates.
6. Any number of endpoints may be defined in this way so long as they don't cause port conflicts.
- `serverOptions.Configure(context.Configuration.GetSection("Kestrel"))` returns a KestrelConfigurationLoader with an `.Endpoint(string name, options => { })` method that can be used to supplement configured endpoint's settings.
```
  serverOptions.Configure(context.Configuration.GetSection("Kestrel"))
    .Endpoint("HTTPS", opt =>
    {
         opt.ListenOptions.Protocols = HttpProtocols.Http1;
         opt.HttpsOptions.SslProtocols = SslProtocols.Tls12;
    });
```
- The configuration section for each endpoint is a available on the options in the Endpoint method so that custom settings may be read.
- Multiple configurations may be loaded by calling `serverOptions.Configure(context.Configuration.GetSection("Kestrel"))` again with another section. Only the last configuration will be used unless Load is explicitly called on prior instances. The metapackage will not call Load so that it's default configuration section may be replaced.
- KestrelConfigurationLoader mirrors the Listen family of APIs from KestrelServerOptions as Endpoint overloads so code and config endpoints may be configured in the same place. These overloads do not use names and only consume default settings from configuration.


src/Kestrel/Internal/ConfigReader.cs
```diff
@@ -0,0 +1,114 @@
...
+                    "Certificate": {
+                        "Path": "testCert.pfx",
+                        "Password": "testPassword"
+                    }
+                }
+                */
+                var url = endpointConfig["Url"];
+                if (string.IsNullOrEmpty(url))
+                {
+                    // TODO: Log?
```
halter73: Why not throw in this case? I would log at a minimum though.

src/Kestrel/KestrelConfigBuilder.cs
```diff
@@ -0,0 +1,122 @@
...
+
+            EndpointConfigurations[name] = configureOptions ?? throw new ArgumentNullException(nameof(configureOptions));
+            return this;
+        }
+
+        public void Build()
+        {
+            if (Options.ConfigurationBuilder == null)
+            {
+                // The builder has already been built.
```
halter73: When does Build get called twice?

src/Kestrel/KestrelServerOptionsConfigurationExtensions.cs
```diff
@@ -0,0 +1,15 @@
...
+
+using Microsoft.Extensions.Configuration;
+
+namespace Microsoft.AspNetCore.Server.Kestrel.Core
+{
+    public static class KestrelServerOptionsConfigurationExtensions
+    {
+        public static KestrelConfigBuilder Configure(this KestrelServerOptions options, IConfiguration config)
+        {
+            return new KestrelConfigBuilder(options, config); // Assigns itself to options.ConfigurationBuilder
```
halter73: Why not just create a local variable and assign it to options.ConfigurationBuilder in this method instead? It feels slightly better than assigning a partially constructed object even if it's functionally the same.

src/Kestrel/Internal/ConfigReader.cs
```diff
@@ -0,0 +1,114 @@
...
+            foreach (var endpointConfig in endpointsConfig)
+            {
+                /*
+                 "EndpointName": {
+                    "Url": "https://*:5463",
+                    "Certificate": {
+                        "Path": "testCert.pfx",
+                        "Password": "testPassword"
+                    }
+                }
```
halter73: Nit: I don't mind the comment, but use single line comments instead.

src/Kestrel/KestrelConfigBuilder.cs
```diff
@@ -0,0 +1,122 @@
...
+                Options.EndpointDefaults(listenOptions);
+
+                HttpsConnectionAdapterOptions httpsOptions = null;
+                if (https)
+                {
+                    httpsOptions = new HttpsConnectionAdapterOptions();
+                    Options.GetHttpsDefaults()(httpsOptions);
+
+                    var certInfo = new CertificateConfig(endpoint.CertConfig);
+                    if (certInfo.IsFileCert)
```
halter73: Are any errors thrown if a file path and a store location are both specified for the same endpoint?

src/Kestrel.Core/KestrelServerOptions.cs
```diff
@@ -55,6 +57,31 @@ public class KestrelServerOptions
...
+
+        /// <summary>
+        /// A default configuration action for all endpoints. Use for Listen, configuration, the default url, and URLs.
+        /// </summary>
+        internal Action<ListenOptions> EndpointDefaults { get; set; } = _ => { };
+
+        /// <summary>
+        /// Used to flow settings for connection adapters and other extensions.
+        /// </summary>
+        public IDictionary<string, object> AdapterData { get; } = new Dictionary<string, object>(0);
```
halter73: Yuck. IFeatureCollection?

src/Kestrel.Https/KestrelServerOptionsHttpsExtensions.cs
```diff
@@ -0,0 +1,25 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using Microsoft.AspNetCore.Server.Kestrel.Core;
+
+namespace Microsoft.AspNetCore.Server.Kestrel.Https
+{
+    public static class KestrelServerOptionsHttpsExtensions
+    {
+        public static void ConfigureHttpsDefaults(this KestrelServerOptions serverOptions, Action<HttpsConnectionAdapterOptions> configureOptions)
```
halter73: I would add doc comments to this and `ConfigureEndpointDefaults` need to be called prior to the Configure/Listen calls it is providing default values for.

src/Kestrel/KestrelConfigBuilder.cs
Tratcher: I don't expect a given builder to be called twice, but it is a public API. If it was allowed to run twice you'd get duplicate endpoints and port conflicts.

src/Kestrel.Core/KestrelServerOptions.cs
Tratcher: A dictionary in sheep's clothing... You prefer the typed keys?

src/Kestrel/KestrelServerOptionsConfigurationExtensions.cs
Tratcher: I did it this way because Build then un-assigns itself, this way you can see the whole lifecycle in one class. I can change it.

src/Kestrel/KestrelConfigBuilder.cs
Tratcher: No, I don't want to second guess extraneous data. If we have what we need then there's no reason to fail. Nor do I expect people to mix those two up.

samples/SampleApp/SampleApp.csproj
```diff
@@ -12,11 +12,24 @@
...
     <PackageReference Include="Microsoft.Extensions.Logging.Console" Version="$(MicrosoftExtensionsLoggingConsolePackageVersion)" />
   </ItemGroup>
 
   <ItemGroup>
     <Content Include="../../test/shared/TestCertificates/testCert.pfx" CopyToOutputDirectory="PreserveNewest" />
   </ItemGroup>
 
+  <ItemGroup>
+    <Content Update="appsettings.Development.json">
+      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
```
davidfowl: nit: Make these properties

samples/SampleApp/Startup.cs
```diff
@@ -71,11 +104,17 @@ public static Task Main(string[] args)
...
-                        listenOptions.UseHttps("testCert.pfx", "testPassword");
+                        // Use default dev cert
+                        listenOptions.UseHttps();
                     });
 
                     options.ListenAnyIP(basePort + 3);
 
+                    options.ListenAnyIP(basePort + 4, listenOptions =>
+                    {
+                        listenOptions.UseHttps(StoreName.My, "aspnet.test", StoreLocation.CurrentUser);
```
davidfowl: Will this explode?

src/Kestrel.Core/KestrelServerOptions.cs
```diff
@@ -55,6 +57,31 @@ public class KestrelServerOptions
         /// </summary>
         public KestrelServerLimits Limits { get; } = new KestrelServerLimits();
 
+        /// <summary>
+        /// Provides a configuration source where endpoints will be loaded from on server start.
+        /// The default is null.
+        /// </summary>
+        public IKestrelConfigBuilder ConfigurationBuilder { get; set; }
```
davidfowl: nit: `IKestrelConfigurationBuilder`

src/Kestrel.Core/KestrelServerOptions.cs
davidfowl: This is pretty gross. How is it used?

src/Kestrel/KestrelConfigBuilder.cs
halter73: I see no reason not to fail if both a file path and a store location is provided.

What's the logic behind choosing file path over the store location when both are specified? If you don't expect people to mix the two, what's wrong with throwing?

src/Kestrel.Core/KestrelServerOptions.cs
halter73: @davidfowl It's there to allow extension methods defined by connection adapters in separate assemblies to store data between calls to extension methods.

In this case it is used by KestrelServerOptions.ConfigureHttpsDefaults() to store away a default HttpsConnectionAdapterOptions for later use by ListenOptions.UseHttps() and KestrelConfigBuilder.Build().

samples/SampleApp/SampleApp.csproj
Tratcher: I'd rather not hand customizes VS generated settings.

samples/SampleApp/Startup.cs
Tratcher: If the cert is missing? Yes

Tratcher:
@halter73 Updated with tests and ready to review. Let me know if you want me to trim down the new sample content.

Tratcher:
Filed https://github.com/aspnet/KestrelHttpServer/issues/2216 to consider what other settings would make sense to read from config. These would all be easy to add on top of the existing design.

Tratcher:
*Poke* please review this today.

src/Kestrel.Core/Internal/IDefaultHttpsProvider.cs
```diff
@@ -1,10 +1,21 @@
...
+using System.Security.Cryptography.X509Certificates;
+
 namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal
 {
     public interface IDefaultHttpsProvider
     {
+        /// <summary>
+        /// Returns the default certificate, if available, otherwise null.
+        /// </summary>
+        X509Certificate2 Certificate { get; }
```
halter73: When does this need to be set?

src/Kestrel.Core/KestrelServerOptions.cs
```diff
@@ -55,6 +57,31 @@ public class KestrelServerOptions
...
+        /// </summary>
+        public IDictionary<string, object> AdapterData { get; } = new Dictionary<string, object>(0);
+
+        /// <summary>
+        /// Specifies a configuration Action to run for each newly created endpoint.
+        /// </summary>
+        /// <param name="configureOptions"></param>
+        public void ConfigureEndpointDefaults(Action<ListenOptions> configureOptions)
+        {
+            EndpointDefaults = configureOptions ?? throw new ArgumentNullException(nameof(configureOptions));
```
halter73: Should we allow this to stack? Either way the doc comments should mention whether this overrides previous calls to ConfigureEndpointDefaults.

src/Kestrel.Https/Internal/CertificateLoader.cs
```diff
@@ -0,0 +1,103 @@
...
+                finally
+                {
+                    if (foundCertificate != null)
+                    {
+                        storeCertificates.Remove(foundCertificate);
+                        foundCertificates.Remove(foundCertificate);
+                    }
+
+                    DisposeCertificates(storeCertificates);
+                    DisposeCertificates(foundCertificates);
```
halter73: After you're done with a X509Store, you seriously have to enumerate X509Store.Certificates and dispose all the ones you didn't use? That's ridiculous if true.

halter73: Also isn't everything in foundCertificates also in storeCertificates, so the second call is just double disposing?

src/Kestrel/EndpointConfiguration.cs
```diff
@@ -0,0 +1,26 @@
...
+        internal EndpointConfiguration(bool isHttps, ListenOptions listenOptions, HttpsConnectionAdapterOptions httpsOptions, IConfigurationSection configSection)
+        {
+            IsHttps = isHttps;
+            Listener = listenOptions ?? throw new ArgumentNullException(nameof(listenOptions));
+            Https = httpsOptions ?? throw new ArgumentNullException(nameof(httpsOptions));
+            ConfigSection = configSection ?? throw new ArgumentNullException(nameof(configSection));
+        }
+
+        public bool IsHttps { get; }
+        public ListenOptions Listener { get; }
```
halter73: Nit: name it ListenOptions.

src/Kestrel/KestrelConfigurationBuilder.cs
```diff
@@ -0,0 +1,306 @@
...
+        }
+
+        /// <summary>
+        /// Bind to given IP address and port.
+        /// </summary>
+        public KestrelConfigurationBuilder Endpoint(IPAddress address, int port) => Endpoint(address, port, _ => { });
+
+        /// <summary>
+        /// Bind to given IP address and port.
+        /// The callback configures endpoint-specific settings.
```
halter73: Nit: If we're going to include the sentence about the configure callback, we should do so for all applicable overloads.

samples/SampleApp/Startup.cs
```diff
@@ -48,10 +53,28 @@ public static Task Main(string[] args)
                 {
                     factory.AddConsole();
                 })
+                .ConfigureAppConfiguration((hostingContext, config) =>
+                {
+                    var env = hostingContext.HostingEnvironment;
+                    config.AddJsonFile("appsettings.json", optional: true)
+                          .AddJsonFile($"appsettings.{env.EnvironmentName}.json", optional: true);
```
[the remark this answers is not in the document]
Tratcher: No, they're optional in the sample. The config is additive on top of the endpoints defined in code.

src/Kestrel.Core/Internal/AddressBinder.cs
```diff
@@ -169,10 +164,33 @@ private class DefaultAddressStrategy : IStrategy
...
+                options.KestrelServerOptions = context.ServerOptions;
+                context.ServerOptions.EndpointDefaults(options);
+                await options.BindAsync(context).ConfigureAwait(false);
+
+                // Conditional https default, only if a cert is available
+                options = ParseAddress(Constants.DefaultServerHttpsAddress, out https);
+                options.KestrelServerOptions = context.ServerOptions;
+                context.ServerOptions.EndpointDefaults(options);
+
+                if (!options.ConnectionAdapters.Any(f => f.IsHttps))
```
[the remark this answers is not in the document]
Tratcher: If you add HTTPS in ConfigureEndpointDefaults then you end up listening on both "https://localhost:5000" and "https://localhost:5001". This if only skips adding the https adapter twice, it doesn't skip binding to this endpoint.

src/Kestrel.Core/Internal/IDefaultHttpsProvider.cs
Tratcher: You found it over at https://github.com/aspnet/KestrelHttpServer/pull/2186/files#diff-590565eed2ca0d80acae41edb7fde1a7R230. Continuing the discussion there.

src/Kestrel/KestrelConfigurationBuilder.cs
```diff
@@ -0,0 +1,306 @@
...
+                var httpsOptions = new HttpsConnectionAdapterOptions();
+                if (https)
+                {
+                    // Defaults
+                    httpsOptions.ServerCertificate = listenOptions.KestrelServerOptions.GetOverriddenDefaultCertificate();
+                    Options.GetHttpsDefaults()(httpsOptions);
+                    if (httpsOptions.ServerCertificate == null)
+                    {
+                        var provider = Options.ApplicationServices.GetRequiredService<IDefaultHttpsProvider>();
+                        httpsOptions.ServerCertificate = provider.Certificate; // May be null.
```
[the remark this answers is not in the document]
Tratcher: ConfigureHttps doesn't let you provide your own httpsOptions, or even a callback. The dev needs to be able to configure httpsOptions in configureEndpoint below before it's applied.

IDefaultHttpsProvider's signature is constrained by its use in Kestrel.Core, it can't expose any of the Https types directly.

src/Kestrel.Core/Internal/IKestrelConfigurationBuilder.cs
```diff
@@ -0,0 +1,13 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal
+{
+    /// <summary>
+    /// Used to provide delayed, replaceable configuration of KestrelServerOptions
+    /// </summary>
+    public interface IKestrelConfigurationBuilder
```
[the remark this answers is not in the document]
Tratcher: Verses an abstract class? Or something else?

Kestrel.Core needs a reference to the builder, but it can't include any of the https types.

src/Kestrel/KestrelServerOptionsConfigurationExtensions.cs
```diff
@@ -0,0 +1,35 @@
...
+    public static class KestrelServerOptionsConfigurationExtensions
+    {
+        /// <summary>
+        /// Creates a configuration builder for setting up Kestrel.
+        /// </summary>
+        /// <param name="options"></param>
+        /// <returns></returns>
+        public static KestrelConfigurationBuilder Configure(this KestrelServerOptions options)
+        {
+            var builder = new KestrelConfigurationBuilder(options, new ConfigurationBuilder().Build());
```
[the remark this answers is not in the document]
Tratcher: Not that I see. @HaoK ?

src/Kestrel.Https/Internal/CertificateLoader.cs
Tratcher: @javiercn? You've played with the cert store a lot lately.

Note this was copied from Hosting with an added EKU check.
https://github.com/aspnet/Hosting/blob/a2962d54f18dece7ec57ee42f3e565f3727168e1/shared/Microsoft.AspNetCore.Certificates.Configuration.Sources/CertificateStoreLoader.cs#L50

I think you're right about the double dispose, unless Find is making a deep copy for some reason.

javiercn: @halter73 I think you need to dispose the ones that you don't use. I would have to ask, but I believe that when you open the store and enumerate the certificates you get some unmanaged references to the actual underlying keys (when present) so I believe you need to dispose those.

I believe that the dotnet crypto folks gave us explicit feedback on Cesar's original PR about this. We can probably dig up that PR

Tratcher: Native resources would all have finalizers, but it's a good habit to explicitly dispose them when you're done.

src/Kestrel.Core/Internal/IKestrelConfigurationBuilder.cs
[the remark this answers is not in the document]
HaoK: Maybe you should replace Configuration with something like Settings to disambiguate between Config

IKestrelConfigurationBuilder => KestrelSettingsBuilder
EndpointConfiguration => EndpointSettings

src/Kestrel.Core/KestrelServerOptions.cs
HaoK: This looks pretty weird, I'm still catching up with this PR and I'll see if I have any better alternatives for what you guys are trying to do

samples/SampleApp/Startup.cs
```diff
@@ -71,11 +94,34 @@ public static Task Main(string[] args)
...
+                    options.ListenAnyIP(basePort + 4, listenOptions =>
+                    {
+                        listenOptions.UseHttps(StoreName.My, "aspnet.test", StoreLocation.CurrentUser, allowInvalid: true);
+                    });
+
+                    options
+                        .Configure()
+                        .Endpoint(IPAddress.Loopback, basePort + 5)
+                        .LocalhostEndpoint(basePort + 6)
+                        .Build();
```
HaoK: So, this I'm guessing Build() is required because that's what triggers the endpoint to actually get added to the options?  

Related question, there is no Build() missing on the Kestrel binding one below, so would that just basically be a no-op giving you an unused builder?

I think it might make things simpler if you just add a separate dedicated options for the defaults, i.e. KestralServerDefaultOptions and configure it via options normally.  Thent he ConfigureXyzDefault methods would just be configuring this other option instead of manipulating the builder.   

You'd still need the existing builder logic to be able to create endpoints from config, but its no longer part of the KestralServerOptions.  The server can just get new default options along with the normal options and continue doing the same thing that Build() does.  The only change that might be required is all of the defaults configuration would need to be done earlier, as they would need to be in the service container.

I definitely don't have a ton of context in this code or how everything fits together, but using a dedicated options class for the defaults seems to be an potential alternative path to consider

src/Kestrel/Internal/ConfigurationReader.cs
```diff
@@ -0,0 +1,136 @@
...
+    internal class CertificateConfig
+    {
+        public CertificateConfig(IConfigurationSection configSection)
+        {
+            ConfigSection = configSection;
+        }
+
+        public IConfigurationSection ConfigSection { get; }
+
+        public bool Exists => ConfigSection.GetChildren().Any();
```
HaoK: Nit: you could consider using the Exists extension method which is similar.
 https://github.com/aspnet/Configuration/blob/dev/src/Config.Abstractions/ConfigurationExtensions.cs#L76

You can consider using section.Bind(this) if you wanted to avoid all the ConfigSection property lookups where the name is the property name.

src/Kestrel/Internal/ConfigurationReader.cs
```diff
@@ -0,0 +1,136 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using Microsoft.Extensions.Configuration;
+
+namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal
+{
+    internal class ConfigurationReader
```
HaoK: Nit: you could consider just making this into two internal static extension methods on IConfiguration

`IConfiguration.ReadCertificates(this IConfiguration)`
`IConfiguration.ReadEndpoints(this IConfiguration)`

samples/SampleApp/Startup.cs
Tratcher: Kestrel calls Build on the last builder at Start. The idea is that the metapackage will assign a default configuration but it won't call Build so that that default can be replaced by user code. If you want to add two configurations you need to call Build on at least the first one.

HaoK: Requiring them to call build sometimes seems a bit awkward, can't you just expose a list that they can control and remove/add as needed instead... options.Builders.Add instead of options.Configure

RE duplication, you should be able to reuse whatever sub options you want in the defaults, why wouldn't you be able to use the ListenOptions/Adapters for the default options as well.  But even if you wanted to have a complete identical surface area, you could just use named options

// Store defaults via "default" name
```
   var defaults = IOptionsMonitor<KestrelServerOptions>.Get("defaults");
   // ConfigureXDefault methods just pass extra name parameter
```

Tratcher: In practice I expect them to almost never call build, they either use the config defined in the metapackage or they use the one from their own code. Keeping a list does not seem necessary. We'll see what the preview1 feedback is like.

From your examples I'm not sure you understand what defaults we're talking about configuring. It's not that there are multiple KestrelServerOptions with a set of defaults, it's that a given KestrelServerOptions has multiple endpoints / ListenOptions / httpsOptions and you can define defaults for the options on those. The connection adapters are the most tricky here, you'd need to create a fresh one for each endpoint so you can't create one as a default without implementing some kind of cloning logic. With the current model you can re-use all of the extensions off of ListenOptions for creating connection adapters.

src/Kestrel.Core/Internal/AddressBinder.cs
```diff
@@ -169,10 +164,33 @@ private class DefaultAddressStrategy : IStrategy
...
             public async Task BindAsync(AddressBindContext context)
             {
-                context.Logger.LogDebug(CoreStrings.BindingToDefaultAddress, Constants.DefaultServerAddress);
+                var options = ParseAddress(Constants.DefaultServerAddress, out var https);
+                options.KestrelServerOptions = context.ServerOptions;
+                context.ServerOptions.EndpointDefaults(options);
+                await options.BindAsync(context).ConfigureAwait(false);
+
+                // Conditional https default, only if a cert is available
+                options = ParseAddress(Constants.DefaultServerHttpsAddress, out https);
```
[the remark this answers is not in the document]
Tratcher: We use httpsOptions for HttpsConnectionAdapterOptions everywhere else. I've renamed these to httpDefault and httpsDefault.

src/Kestrel.Core/Internal/IKestrelConfigurationBuilder.cs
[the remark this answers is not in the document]
Tratcher: It's about loading information from config, what's to disambiguate?

src/Kestrel.Core/KestrelServerOptions.cs
Tratcher: I don't think more that one default action is required. Comments updated.

src/Kestrel/KestrelConfigurationBuilder.cs
Tratcher: Removed

src/Kestrel/Internal/ConfigurationReader.cs
Tratcher: No need when this is only for internal usage.

src/Kestrel.Core/Internal/IKestrelConfigurationBuilder.cs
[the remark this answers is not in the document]
HaoK: The naming pattern for this still doesn't match what all the other builders do.  Namely they Build something, IWebHostBuilder builds IWebHost.  Or the builders that return from AddXyz off of services don't have a build but they are implicitly building the XyzOptions for the feature, i.e. Mvc/Identity.

This looks more like a KestrelConfigurationLoader with a void Load() method.

The builders are basically about having a mutable setting object (i.e. Options, ConfigurationSource, XyzBuilder), that builds something immutable that like a ConfigurationProvider, scheme instance, or in this case a KestrelConfiguration that KestrelServer uses.  So a Build() that returns nothing doesn't seem to follow that pattern.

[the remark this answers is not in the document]
Tratcher: Interesting thought. Do we have any other Loaders in the stack?

Tratcher:
Note many of the strange design decisions here are due to the assembly separation between K.Core and K.Https. We've opted to merge the https libraries into core and then revisit this PR.

samples/SampleApp/Startup.cs
HaoK: So let me try to get on the same page...

So the KestrelServer eventually wants to produce a bunch of endpoints which has options in a ListenOptions instance, (Endpoints are created from these 1:1?) 

Https is doing weird stuff which I guess relates to the fact that its configuring listen option instances and doing its own options like stuff.  The EndpointDefaults action is basically equivalent to doing a ConfigureAll on ListenOptions if you were using IOptions.

Part of the confusion here I think is that KestrelServerOptions are using IOptions/Configure, while there are a few options types that are used more like EndpointSettings/AdapterSettings and not configured like normal options.

With named endpoints, that matches exactly named options, you can just use `IOptions<EndpointOptions>.Get(name)` and just bind against `config.GetSection("Kestrel:Endpoints")`.  Your defaults could then just be sugar for services.ConfigureAll<EndpointOptions>(o =>  o.Protocols = HttpProtocols.Http1)


Assuming you can just hang all Endpoint operations off of `.Endpoints`, maybe something like?   But I'm sure I don't grok what's going on around the adapter/https yet.

```C#
   UseKestrel(context, o => {

    // Default wireup (but showing what the sugar methods call)
    o.ConfigureAll<EndpointOptions>(opt => opt.Protocols = HttpProtocols.Http1);
    o.ConfigureAll<HttpsDefaultOptions>(o => o.SslProtocols = SslProtocols.Tls12);

      // just binds all the endpoint options to their names using Configure (and defaults?)
      o.Endpoints.Configure(context.Configuration.GetSection("Kestrel:Endpoints")); 
      o.Endpoints.Add(IPAddress.Loopback, basePort + 5);
      o.Endpoints.AddLocalhost(basePort + 6);
      o.Endpoints.AddNamed("NamedEndpoint", opt =>
                        {
                             opt.Listener.Protocols = HttpProtocols.Http1;
                         })
      o.Endpoints.AddNamed("NamedHttpsEndpoint", opt =>
                         {
                             opt.Https.SslProtocols = SslProtocols.Tls12;
                         });
   }
```

src/Kestrel.Core/Internal/IKestrelConfigurationBuilder.cs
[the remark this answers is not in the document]
HaoK: Sure, its basically a the worker/instance half of the configuration providers:

https://github.com/aspnet/Configuration/blob/dev/src/Config.Abstractions/IConfigurationProvider.cs

But basically you can do what you want, I just think there's an implicit usage/behavior pattern involved when you name it a "Buidler"

Tratcher:
Rebased on the new merged assemblies and code consolidated. Ready for a final pass @halter73

samples/SampleApp/Startup.cs
Tratcher: You're still missing quite a bit of context. Stop by if you want to walk through it.

The question I'd originally sent you was if configuration has a static helper for the scenario where I only want an empty placeholder instance like `new ConfigurationBuilder().Build()`

HaoK: No because there's always config, its pretty strange to pass in empty configuration to a configuration loader class though...  Hence why I think the naming is a bit weird.

Like this doesn't use config at all right?
```
options
                         .Configure()
                         .Endpoint(IPAddress.Loopback, basePort + 5)
                         .LocalhostEndpoint(basePort + 6)
                         .Load();
```

Tratcher: Damian wanted to mirror the Listen APIs so you could still configure everything in one place. Yes it makes for an slightly odd fraken type.

HaoK: That's fine, my main point is that this seems more than just a configuration loader, this is public endpoint management API basically, where the ctor takes an optional config and Load() is basically committing the set of endpoints.  

I think building that dedicated API where configuration is just one way to import a chunk of endpoint settings is maybe a simpler way to expose this...

I still think this would be easier to understand if the API just exposed the endpoints on the options.  Having to call configure/endpoint/load and then another configure/endpoint block just seems strange.

Seems like it'd be better to let them express it more concisely, without any weird loads.

```
options.AddEndpoint(IPAddress.Loopback, basePort + 5)
    .AddLocalhostEndpoint(basePort + 6)
    .LoadConfig(context.Configuration.GetSection("Kestrel")
    .AddEndpoint("NamedEndpoint", opt =>
                         {
                             opt.ListenOptions.Protocols = HttpProtocols.Http1;
                         })
                         .Endpoint("NamedHttpsEndpoint", opt =>
                         {
                             opt.HttpsOptions.SslProtocols = SslProtocols.Tls12;
                         }
```

And you can always just let them Remove/Update existing endpoints.

src/Kestrel.Core/KestrelConfigurationLoader.cs
```diff
@@ -0,0 +1,297 @@
...
+                Options.ListenUnixSocket(socketPath, configure);
+            });
+
+            return this;
+        }
+
+        /// <summary>
+        /// Open a socket file descriptor.
+        /// </summary>
+        public KestrelConfigurationLoader HandleEndpoint(ulong handle) => HandleEndpoint(handle, _ => { });
```
muratg: Just curious - did we expose `ulong`s in the public API before?

Tratcher: Yes these are all mirrors of the existing Listen APIs.

halter73: Yes. `KestrelServerOptions.ListenHandle(ulong)`.

muratg: :+1:

guardrex:
@Tratcher 

In the initial post above `Certificates` is a child of `Kestrel` ...

```json
{
  "Kestrel": {
    "Certificates": {
       "Default": {
          "Path": "...",
          "Password": ""
       }
    }
  }
}
```
... but in the blog post, it's not ...

https://blogs.msdn.microsoft.com/webdev/2018/02/02/asp-net-core-2-1-roadmap#security

```json
{
  "Kestrel": {
    
  },
  "Certificates": {
    "Default": {
      "Path": "<file>",
      "Password": "<password>"
    },

    "Default": {
      "Subject": "",
      "Store": "",
      "Location": "",
      "AllowInvalid": ""
    }
  }
}
```

... no love from schema store ... http://json.schemastore.org/appsettings

I assume the blog post is incorrect. cc/ @danroth27

Tratcher:
@guardrex you're correct. We've already fixed that in another blog post, but it looks like we should update this one too. I left him a comment directly on the that post.

Issue #1778, opened 2017-04-26
Update to a release version of Roslyn


halter73:
See https://github.com/aspnet/KestrelHttpServer/pull/1748#issuecomment-297556985.

muratg:
We'll reopen if we decide to this in the future.

PR #2193, opened 2017-11-30
Remove the custom C# compiler now that we're using the 2.2 SDK

davidfowl:
- It should speed builds up considerably

/cc @VSadov We're going to be getting compilers from the SDK from now on. So we should make sure the changes propagate there regularly.

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -18,9 +18,4 @@
     <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
     <LangVersion>latest</LangVersion>
   </PropertyGroup>
-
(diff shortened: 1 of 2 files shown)


pakrym:
This has to be done in `SignalR` and `Benchmarks` too

natemcmaster:
Do we need to set `LangVersion` to 7.2 at some point?

davidfowl:
> Do we need to set LangVersion to 7.2 at some point?

If we don't plan to always use latest.

halter73:
I was going to close #1778, but the issue was already closed back in June. Good thing we remembered 😅

PR #2217, opened 2017-12-13
Allow Content-Length 0 for Websocket-Connection

forki:
mojo is currently sending Content-Length 0 for websockets. This fails against ASP.NET Core/Kestrel.
I tried to patch mojo (see https://github.com/kraih/mojo/pull/1163), but I think this should patched server-side as well.

### DIFF
--- a/src/Kestrel.Core/Internal/Http/Http1MessageBody.cs
+++ b/src/Kestrel.Core/Internal/Http/Http1MessageBody.cs
@@ -228,6 +228,16 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
                 keepAlive = (connectionOptions & ConnectionOptions.KeepAlive) == ConnectionOptions.KeepAlive;
             }
 
+            if (upgrade)
+            {
+                if (headers.HeaderTransferEncoding.Count > 0 || (headers.ContentLength.HasValue && headers.ContentLength.Value != 0))
+                {
+                    context.ThrowRequestRejected(RequestRejectionReason.UpgradeRequestCannotHavePayload);
+                }
+
+                return new ForUpgrade(context);
+            }
+
             var transferEncoding = headers.HeaderTransferEncoding;
             if (transferEncoding.Count > 0)
             {
@@ -244,11 +254,6 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
                     context.ThrowRequestRejected(RequestRejectionReason.FinalTransferCodingNotChunked, transferEncoding.ToString());
                 }
 
-                if (upgrade)
-                {
-                    context.ThrowRequestRejected(RequestRejectionReason.UpgradeRequestCannotHavePayload);
-                }
-
                 return new ForChunkedEncoding(keepAlive, context);
             }
 
@@ -259,10 +264,6 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
                 if (contentLength == 0)
                 {
                     return keepAlive ? MessageBody.ZeroContentLengthKeepAlive : MessageBody.ZeroContentLengthClose;
-                }
-                else if (upgrade)
-                {
-                    context.ThrowRequestRejected(RequestRejectionReason.UpgradeRequestCannotHavePayload);
                 }
 
                 return new ForContentLength(keepAlive, contentLength, context);
@@ -278,11 +279,6 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
                     var requestRejectionReason = httpVersion == HttpVersion.Http11 ? RequestRejectionReason.LengthRequired : RequestRejectionReason.LengthRequiredHttp10;
                     context.ThrowRequestRejected(requestRejectionReason, context.Method);
                 }
-            }
-
-            if (upgrade)
-            {
-                return new ForUpgrade(context);
             }
 
             return keepAlive ? MessageBody.ZeroContentLengthKeepAlive : MessageBody.ZeroContentLengthClose;
--- a/test/Kestrel.Core.Tests/MessageBodyTests.cs
+++ b/test/Kestrel.Core.Tests/MessageBodyTests.cs
@@ -504,6 +504,32 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Tests
         }
 
         [Fact]
+        public async Task UpgradeConnectionAcceptsContentLengthZero()
+        {
+            // https://tools.ietf.org/html/rfc7230#section-3.3.2
+            // "A user agent SHOULD NOT send a Content-Length header field when the request message does not contain
+            // a payload body and the method semantics do not anticipate such a body."
+            //  ==> it can actually send that header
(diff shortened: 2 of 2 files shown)


forki:
the travis fail looks unrelated

forki:
@khellang thanks for finding this. Now I need it as nupkg?! What's the easiest way?

davidfowl:
We should test some other servers to see what they do:
- IIS
- nginx
- apache 

If they all work then it seems like a no brainer.

khellang:
Until anything's merged, I guess building locally is your only option. It would be nice if AppVeyor exposed artifacts. Filed https://github.com/aspnet/BuildTools/issues/518

forki:
@khellang yes building locally (rebased on 2.0.1)

@davidfowl I asusme the others don't care. otherwise mojo would probably be already fixed, For the record I sent https://github.com/kraih/mojo/pull/1163 (hoping it works), but even if it gets merged I will have HUGE trouble to get it deployed. The client side is in ecar charging stations (hardware) and these stations have trouble to connect to our backend

davidfowl:
> @davidfowl I asusme the others don't care. otherwise mojo would probably be already fixed, For the record I sent kraih/mojo#1163 (hoping it works), but even if it gets merged I will have HUGE trouble to get it deployed. The client side is in ecar charging stations (hardware) and these stations have trouble to connect to our backend

Sure, I'd still like to know what others do before merging this.

forki:
Giphy gave me this:

mojo

Needed to share

forki:
@part-timeDev found the relevant part of the spec in https://tools.ietf.org/html/rfc7230#page-30

    A user agent SHOULD send a Content-Length in a request message when
    no Transfer-Encoding is sent and the request method defines a meaning
    for an enclosed payload body.  For example, a Content-Length header
    field is normally sent in a POST request even when the value is 0
    (indicating an empty payload body).  A user agent SHOULD NOT send a
    Content-Length header field when the request message does not contain
    a payload body and the method semantics do not anticipate such a
    body.

So if I read that correct then mojo SHOULD NOT send it, but it's not strictly violating the spec if it does. In other words kestrel needs to accept it. Right?

Tratcher:
`Content-Length: 0` should be treated as a no-op for any request that's not expected to have a body.

forki:
Currently it's not.

Am 14.12.2017 17:23 schrieb "Chris Ross" <notifications@github.com>:

Tratcher:
Understood, I was only agreeing with your assessment.

src/Kestrel.Core/Internal/Http/Http1MessageBody.cs
```diff
@@ -258,6 +258,10 @@ private void TryStopTimingReads()
 
                 if (contentLength == 0)
                 {
+                    if (upgrade && object.ReferenceEquals(context.Method, HttpMethods.Get))
+                    {
+                        return new ForUpgrade(context);
+                    }
```
halter73: I don't love having `return new ForUpgrade(context)` in two places. I think it's simpler to just handle upgrade requests earlier in the method. How does this look instead @forki @Tratcher?

```diff
+            if (upgrade)
+            {
+                if (headers.ContentLength.HasValue && headers.ContentLength.Value != 0)
+                {
+                    context.ThrowRequestRejected(RequestRejectionReason.UpgradeRequestCannotHavePayload);
+                }
+
+                return new ForUpgrade(context);
+            }
+
             if (headers.ContentLength.HasValue)
             {
                 var contentLength = headers.ContentLength.Value;
 
                 if (contentLength == 0)
                 {
                     return keepAlive ? MessageBody.ZeroContentLengthKeepAlive : MessageBody.ZeroContentLengthClose;
                 }
-                else if (upgrade)
-                {
-                    context.ThrowRequestRejected(RequestRejectionReason.UpgradeRequestCannotHavePayload);
-                }
 
                 return new ForContentLength(keepAlive, contentLength, context);
             }
 
             // Avoid slowing down most common case
             if (!object.ReferenceEquals(context.Method, HttpMethods.Get))
             {
                 // If we got here, request contains no Content-Length or Transfer-Encoding header.
                 // Reject with 411 Length Required.
                 if (HttpMethods.IsPost(context.Method) || HttpMethods.IsPut(context.Method))
                 {
                     var requestRejectionReason = httpVersion == HttpVersion.Http11 ? RequestRejectionReason.LengthRequired : RequestRejectionReason.LengthRequiredHttp10;
                     context.ThrowRequestRejected(requestRejectionReason, context.Method);
                 }
             }
 
-            if (upgrade)
-            {
-                return new ForUpgrade(context);
-            }
-
             return keepAlive ? MessageBody.ZeroContentLengthKeepAlive : MessageBody.ZeroContentLengthClose;
         }
```

The only behavioral difference between the changes I propose above vs the change in this PR is that now POST and GET requests with a "Connection: upgrade" header will now also get upgraded instead of rejected with a 411 response. I see this as a slight improvement.

davidfowl:
Nobody wants to test it with another sever? It’s been a couple of days now....

src/Kestrel.Core/Internal/Http/Http1MessageBody.cs
Tratcher: Makes sense.

forki:
@davidfowl a couple of days = 1 ;-)
And I cited the relevant part of the spec above.

forki:
I added a testcase and applied the proposed changes. Please review

forki:
@davidfowl we are going to test apache and iis on Monday.
Unfortunately the client side is not going to accept the patch and therefore I hope we can somehow solve it here.
@tratcher said "Content-Length: 0 should be treated as a no-op for any request that's not expected to have a body"

So I assume it should be independent of upgrade/websocket stuff. So if apache and iis accept simple GET request (without upgrade) that send content-lengrh zero, then the test is already considered as "2e need to change kestrel"? Is that correct?

davidfowl:
No, I’d like to see what happens when an upgrade is sent with a content length. I’m interested for 2 reasons:
- I’m curious what happens from a purity perspective 
- I’m curious is what happens when those severs act as a reverse proxy to kestrel.

I don’t think the change is bad or good, we’ve had to tweak this logic several times based on clients and proxies doing bizzare things. I’d just like to see what other severs do

forki:
Ok good. So scenario is to test upgrade. Got it. Thanks

Am 15.12.2017 20:14 schrieb "David Fowler" <notifications@github.com>:

forki:
Ok I just set up an nginx reverse proxy to my patched kestrel (this PR applied) and everything works with websockets.

forki:
ping.

forki:
@halter73 @tratcher is there a chance this can be backported to the next hotfix? That would help me to get rid of self built kestrel on our servers. What can I do to help?

Tratcher:
Only severe bugs get fixes backported, like hangs and crashes.

forki:
fair enough. so is there a rough estimate when this hits nuget?

davidfowl:
When 2.1 ships, we haven't announced a date yet. We don't ship these packages individually, it's part of the .NET Core 2.1 release.

PR #2233, opened 2017-12-22
Use common BDN config

pakrym:
That doesn't fail in debug.


benchmarks/Kestrel.Performance/StringUtilitiesBenchmark.cs
```diff
@@ -6,7 +6,7 @@
 
 namespace Microsoft.AspNetCore.Server.Kestrel.Performance
 {
-    [ParameterizedJobConfig(typeof(CoreConfig))]
+    [AspNetCoreBenchmark]
```
pranavkm: Can this be put on the assembly?

pakrym: Good idea, I'll try it

PR #2234, opened 2017-12-22
React to pipeline changes

pakrym:


src/Kestrel.Core/Internal/Infrastructure/IThreadPool.cs
```diff
@@ -7,9 +7,9 @@
 
 namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Infrastructure
 {
-    public interface IThreadPool : IScheduler
+    public abstract class IThreadPool: Scheduler
```
davidfowl: Rename to something else?

test/Kestrel.Core.Tests/HttpParserTests.cs
```diff
@@ -2,8 +2,9 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.Buffers;
 using System.IO.Pipelines;
-using System.IO.Pipelines.Testing;
+//using System.IO.Pipelines.Testing;
```
davidfowl: Remove dead code

src/Kestrel.Core/Internal/Infrastructure/IThreadPool.cs
pakrym: Yep, I'll clean up the code, just wanted to see if I broke anything large during  refactorings.

NuGet.config
```diff
@@ -2,6 +2,7 @@
 <configuration>
   <packageSources>
     <clear />
+    <add key="local" value="d:\localnuget" />
```
muratg: temporary?

build/sources.props
```diff
@@ -8,6 +8,8 @@
       https://dotnet.myget.org/F/aspnetcore-dev/api/v3/index.json;
       https://dotnet.myget.org/F/aspnetcore-tools/api/v3/index.json;
       https://dotnet.myget.org/F/roslyn/api/v3/index.json;
+      https://dotnet.myget.org/F/dotnet-core/api/v3/index.json;
+      https://dotnet.myget.org/f/dotnet-corefxlab/api/v3/index.json;
```
muratg: temporary?

NuGet.config
pakrym: Yes, all feed changes in reaction PRs are temporary and would be undone before the commit.

Issue #2238, opened 2017-12-29
Add functional tests for http2


Tratcher:
There are not currently any functional test for the Http2 features. We accidentally disabled the entire feature when merging the https and core packages due to a missing TFM and didn't notice. https://github.com/aspnet/KestrelHttpServer/pull/2224

The big blocker here is the lack of consistent client support. However, even if there's one client on one of the OS's then that would be a good start.

linianhui:
Does client mean web browser or web server?

Tratcher:
Browsers are clients, yes. System.Net.Http.HttpClient is the client we prefer for functional tests.

linianhui:
@Tratcher But now System.Net.Http.HttpClient does not support http2. (￣▽￣)"

benaadams:
HttpClient supports http2 on .NET Core for UWP, Linux https://github.com/dotnet/corefx/pull/4756 and Windows https://github.com/dotnet/corefx/pull/10948 (but not Full Framework as yet)

Tratcher:
Note the Windows version restriction as well.
Windows 10 Anniversary release a.k.a Windows 10 Version 1607

Tratcher:
I did some initial work in #2665 but only for Win10 WinHttpHandler. Linux and Mac have too many dependency issues, we may have to wait for SocketsHttpHandler.

muratg:
The remaining test we'd like to add is using new HttpClient.

Issue #2182, opened 2017-11-27
Change ITlsApplicationProtocolFeature.ApplicationProtocol to Span<byte>


anurse:
https://github.com/aspnet/KestrelHttpServer/blob/76de77746dd900058f7bd49f66debdd245cf9260/src/Kestrel.Core/Features/ITlsApplicationProtocolFeature.cs#L7-L10

This property should not be represented as a string because the ALPN spec intentionally defines it as an "opaque byte string" and not a textual string:

> Protocols are named by IANA-registered, opaque, non-empty byte strings, as described further in Section 6 ("IANA Considerations") of this document.
>
> https://tools.ietf.org/html/rfc7301#section-3.1

The fact that the currently-registered values are all valid UTF-8 strings is not required by the spec so we shouldn't require it in our type.

We could use `byte[]`, which is relatively safe performance-wise since in fast paths this will almost always be set to an existing constant value. Or we could go with the new hotness and make it return `ReadOnlyMemory<byte>`. Probably the latter is best even if we always just end up backing it with a predefined constant array.

/cc @benaadams @Tratcher @davidfowl

anurse:
Dupe of #2181

anurse:
Also of note: `SslStream` chose not to use a `string` here, so I don't think we should arbitrarily limit it more than that. They used a custom type wrapping a byte array. We could go that far if we wanted to, it would make equality checks clearer, but I don't really mind either way.

benaadams:
Its a getter only, you want to stash it on heap; single instance them - should be `ReadOnlyMemory<byte>` rather than Span

Also `SslApplicatonProtocol` is an option, except doesn't work for non-Ssl ApplicatonProtocols... (in name)

davidfowl:
SslApplicatonProtocol isn't an option because it's .NET Core 2.0 only.

benaadams:
> SslApplicatonProtocol isn't an option because it's .NET Core 2.0 only.

There's an obvious question here; but I'm not going to ask it 😄

benaadams:
Ah gowan... 

So Apln will be Core 2.x and not Framework, until Framework's SslStream catches up - or an Apln shim for Framework?

Guessing as its security/crypto; will need Framework's SslStream to catch up (i.e. a future version of Framework rather than 4.7.1)

anurse:
Correct, ALPN is a .NET Core 2.x only feature.

[...]

PR #2239, opened 2017-12-29
Change ApplicationProtocol to ReadOnlyMemory<byte> 

Tratcher:
#2182 
In the process I realized the prior change for merging the https and core packages had dropped the new netcoreapp2.1 TFM so all of the HTTP/2 code was disabled. Filed https://github.com/aspnet/KestrelHttpServer/issues/2238 for followup.


samples/Http2SampleApp/Http2SampleApp.csproj
```diff
@@ -15,7 +15,9 @@
   </ItemGroup>
 
   <ItemGroup>
-    <Content Include="../../test/shared/TestCertificates/testCert.pfx" CopyToOutputDirectory="PreserveNewest" />
+    <Content Include="testCert.pfx">
+      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
```
halter73: Out of curiosity, does CopyToOutputDirectory need to be a tag, or does it still work as a attribute?

src/Kestrel.Core/Kestrel.Core.csproj
```diff
@@ -4,7 +4,7 @@
     <AssemblyName>Microsoft.AspNetCore.Server.Kestrel.Core</AssemblyName>
     <RootNamespace>Microsoft.AspNetCore.Server.Kestrel.Core</RootNamespace>
     <Description>Core components of ASP.NET Core Kestrel cross-platform web server.</Description>
-    <TargetFramework>netstandard2.0</TargetFramework>
```
halter73: How was this working before without the 2.1 target?

src/Kestrel.Core/Internal/HttpConnection.cs
```diff
@@ -303,7 +305,8 @@ private void DisposeAdaptedConnections()
         private HttpProtocols SelectProtocol()
         {
             var hasTls = _context.ConnectionFeatures.Get<ITlsConnectionFeature>() != null;
-            var applicationProtocol = _context.ConnectionFeatures.Get<ITlsApplicationProtocolFeature>()?.ApplicationProtocol;
+            var applicationProtocol = _context.ConnectionFeatures.Get<ITlsApplicationProtocolFeature>()?.ApplicationProtocol
```
halter73: Can't we now just get rid of ITlsApplicationProtocolFeature and pull the application protocal from the ITlsConnectionFeature we access a line above?

src/Kestrel.Core/Kestrel.Core.csproj
Tratcher: It wasn't. See the PR summary.

src/Kestrel.Core/Internal/HttpConnection.cs
Tratcher: ITlsConnectionFeature is from HttpAbstractions, it's a breaking change to merge them.

samples/Http2SampleApp/Http2SampleApp.csproj
Tratcher: It might work as an attribute but this is what VS consistently generates.

davidfowl: But it looks cleaner as an attribute 😄

src/Kestrel.Core/Internal/HttpConnection.cs
Tratcher: Oh, I see what you mean. The implementation is already merged but we're accessing it via interfaces and can't tell.

halter73: Just delete ITlsApplicationProtocolFeature.cs

halter73: I see. ITlsConnectionFeature itself doesn't include ApplicationProtocol though the implementation does. Just get rid of the TODO then.

Issue #2223, opened 2017-12-18
Flaky test on Ubuntu caused by null-ref in cleanup


ryanbrandenburg:
Kestrel tests [failed on Ubuntu](http://aspnetci/viewLog.html?buildId=370016&buildTypeId=XPlat_Linux_UbuntuUniverseCoreCLR&tab=buildLog&_focus=55343). It seems like the test failed in startup or shutdown due to a null-ref on a `Span`?

pakrym:
https://github.com/dotnet/corefxlab/issues/1978

halter73:
I saw another null-ref (actually a NullReferenceException instead of an ArgumentNullException this time) post the #2247 and #2279 fixes. [Here's the commit](https://github.com/aspnet/KestrelHttpServer/commit/14332c5daf94a7b2d771dc5ba8a03e4b2f63a604) and some details:

```
shalter@halter73-mbp Kestrel.Transport.Sockets.FunctionalTests (dev* u=)$ dotnet build
Microsoft (R) Build Engine version 15.6.54.9755 for .NET Core
Copyright (C) Microsoft Corporation. All rights reserved.
…
Time Elapsed 00:00:15.38
shalter@halter73-mbp Kestrel.Transport.Sockets.FunctionalTests (dev* u=)$ xunit -method Microsoft.AspNetCore.Server.Kestrel.FunctionalTests.MaxRequestBodySizeTests.DoesNotRejectBodylessGetRequestWithZeroMaxRequestBodySize
xUnit.net Console Runner (64-bit .NET Core 4.6.26130.05)
  Discovering: Microsoft.AspNetCore.Server.Kestrel.Transport.Sockets.FunctionalTests
  Discovered:  Microsoft.AspNetCore.Server.Kestrel.Transport.Sockets.FunctionalTests
  Starting:    Microsoft.AspNetCore.Server.Kestrel.Transport.Sockets.FunctionalTests
 
Unhandled Exception: System.NullReferenceException: Object reference not set to an instance of an object.
   at System.Net.Sockets.SocketAsyncContext.BufferMemoryReceiveOperation.InvokeCallback()
   at System.Net.Sockets.SocketAsyncContext.OperationQueue`1.ProcessQueue(SocketAsyncContext context)
   at System.Net.Sockets.SocketAsyncContext.OperationQueue`1.<>c.<.cctor>b__15_0(Object o)
[...]

PR #2247, opened 2018-01-09
Improve server shutdown logic

halter73:
- Ensure connections abort if they don't close gracefully.
- Ensure tests fail if connections don't close during shutdown.

#2223

### DIFF
--- a/src/Kestrel.Core/Internal/Infrastructure/KestrelTrace.cs
+++ b/src/Kestrel.Core/Internal/Infrastructure/KestrelTrace.cs
@@ -45,7 +45,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal
             LoggerMessage.Define<string>(LogLevel.Information, new EventId(20, nameof(RequestProcessingError)), @"Connection id ""{ConnectionId}"" request processing ended abnormally.");
 
         private static readonly Action<ILogger, Exception> _notAllConnectionsAborted =
-            LoggerMessage.Define(LogLevel.Debug, new EventId(21, nameof(NotAllConnectionsAborted)), "Some connections failed to abort during server shutdown.");
+            LoggerMessage.Define(LogLevel.Critical, new EventId(21, nameof(NotAllConnectionsAborted)), "Some connections failed to abort during server shutdown.");
 
         private static readonly Action<ILogger, TimeSpan, DateTimeOffset, Exception> _heartbeatSlow =
             LoggerMessage.Define<TimeSpan, DateTimeOffset>(LogLevel.Warning, new EventId(22, nameof(HeartbeatSlow)), @"Heartbeat took longer than ""{interval}"" at ""{now}"".");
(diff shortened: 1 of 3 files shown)


src/Kestrel.Core/Internal/Infrastructure/KestrelTrace.cs
```diff
@@ -45,7 +45,7 @@ public class KestrelTrace : IKestrelTrace
             LoggerMessage.Define<string>(LogLevel.Information, new EventId(20, nameof(RequestProcessingError)), @"Connection id ""{ConnectionId}"" request processing ended abnormally.");
 
         private static readonly Action<ILogger, Exception> _notAllConnectionsAborted =
-            LoggerMessage.Define(LogLevel.Debug, new EventId(21, nameof(NotAllConnectionsAborted)), "Some connections failed to abort during server shutdown.");
+            LoggerMessage.Define(LogLevel.Critical, new EventId(21, nameof(NotAllConnectionsAborted)), "Some connections failed to abort during server shutdown.");
```
pakrym: Seems like a pretty severe log level jump.

test/shared/TestApplicationErrorLogger.cs
```diff
@@ -44,14 +45,23 @@ public void Log<TState>(LogLevel logLevel, EventId eventId, TState state, Except
...
 
                 if (logLevel == LogLevel.Critical && ThrowOnCriticalErrors)
                 {
-                    throw new Exception("Unexpected critical error.", exception);
+                    throw new Exception($"Unexpected critical error. {log}", exception);
                 }
             }
 
+            // Fail tests where not all the connections close during server shutdown.
+            if (eventId.Id == 21 && eventId.Name == nameof(KestrelTrace.NotAllConnectionsAborted))
```
pakrym: Ugh. Do we need both name and id here?

pakrym:
Nullrefs on CI

test/shared/TestApplicationErrorLogger.cs
halter73: It's the easiest way to not accidentally collide with a LibuvTrace.

natemcmaster:
@muratg @pakrym  what do you think about cherry-picking this into preview1? I'm seeing consistent test failures with the ArgumentNullException from System.Memory.

For example, I've run the release/2.1 branch three times in a row on Travis, and all of them fail with this.
https://travis-ci.org/aspnet/KestrelHttpServer/builds/339575583
https://travis-ci.org/aspnet/KestrelHttpServer/builds/339566949
https://travis-ci.org/aspnet/KestrelHttpServer/builds/339575583

halter73:
@natemcmaster I think this is a good/safe change to cherry-pick.

If we're trying to improve test pass rate, I might consider not cherry-picking the TestApplicationErrorLogger.cs part of this change, since that change verify the server shuts down gracefully for every functional test. I think this is good to verify, but maybe not for a preview1 backport to reduce test failures.

davidfowl:
We should do the other one too then right? The null ref fix?

halter73:
@davidfowl is referring to #2279. I agree that if we're cherry-picking this we should do the same for that. Both PRs fix different causes of the same null ref.

natemcmaster:
I think its worth doing. Kestrel tests fail on every run of macOS for preview1. The especially bad part about this bug is it causes the test runner to abort, so we don't even get a list of tests that passed or failed.

natemcmaster:
http://aspnetci/viewLog.html?tab=buildLog&logTab=tree&filter=debug&expand=all&buildId=406837&_focus=379410

PR #2264, opened 2018-01-20
Hide the Http2 opt-in option.

Tratcher:
The Http/2 implementation is not yet ready for consumption. Hide the setting as to not confuse users. I've added a hidden switch advanced users can use for targeted testing.

### DIFF
--- a/src/Kestrel.Core/ListenOptions.cs
+++ b/src/Kestrel.Core/ListenOptions.cs
@@ -19,6 +19,9 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core
     /// </summary>
     public class ListenOptions : IEndPointInformation, IConnectionBuilder
     {
+        private const string Http2ExperimentSwitch = "Switch.Microsoft.AspNetCore.Server.Kestrel.Experiential.Http2";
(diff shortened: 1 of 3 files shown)


src/Kestrel.Core/ListenOptions.cs
```diff
@@ -19,13 +19,25 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core
     /// </summary>
     public class ListenOptions : IEndPointInformation, IConnectionBuilder
     {
+        private const string Http2ExperimentSwitch = "Switch.Microsoft.AspNetCore.Server.Kestrel.Experiential.Http2";
```
davidfowl: The only beef I have with this is that it's app domain global.

Tratcher: It does mess up non-tls endpoints.
https://github.com/aspnet/KestrelHttpServer/blob/f4d27e67bbe770b0290f1a08c2e3401bae8686f0/src/Kestrel.Core/Internal/HttpConnection.cs#L347-L355

What if we left the API public but threw a NotImplementedException with an fwlink that said to set the AppContext switch to allow it to work? Then you'd still have endpoint specific settings.

nil4: `Experiential` -> `Experimental` ❓

PR #2265, opened 2018-01-22
Mark HTTP/2 as not supported with an AppContext switch override.

Tratcher:
Alternative to https://github.com/aspnet/KestrelHttpServer/pull/2264. With this approach you can enable HTTP/2 per endpoint and it doesn't mess up non-TLS connections.

### DIFF
--- a/src/Kestrel.Core/ListenOptions.cs
+++ b/src/Kestrel.Core/ListenOptions.cs
@@ -19,13 +19,18 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core
     /// </summary>
     public class ListenOptions : IEndPointInformation, IConnectionBuilder
     {
+        internal const string Http2ExperimentSwitch = "Switch.Microsoft.AspNetCore.Server.Kestrel.Experimental.Http2";
+
         private FileHandleType _handleType;
+        private HttpProtocols _protocols = HttpProtocols.Http1;
+        internal bool _isHttp2Supported;
         private readonly List<Func<ConnectionDelegate, ConnectionDelegate>> _components = new List<Func<ConnectionDelegate, ConnectionDelegate>>();
 
         internal ListenOptions(IPEndPoint endPoint)
         {
             Type = ListenType.IPEndPoint;
             IPEndPoint = endPoint;
+            AppContext.TryGetSwitch(Http2ExperimentSwitch, out _isHttp2Supported);
         }
 
         internal ListenOptions(string socketPath)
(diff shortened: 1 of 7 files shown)


test/Kestrel.FunctionalTests/HttpProtocolSelectionTests.cs
```diff
@@ -16,8 +16,18 @@
...
+    public class HttpProtocolSelectionTests : IDisposable
     {
+        public HttpProtocolSelectionTests()
+        {
+            AppContext.SetSwitch(ListenOptions.Http2ExperimentSwitch, isEnabled: true);
+        }
+
+        public void Dispose()
+        {
+            AppContext.SetSwitch(ListenOptions.Http2ExperimentSwitch, isEnabled: false);
```
halter73: This makes it impossible to add another functional test class for HTTP/2 while keeping the experimental flag. Can we create an internal flag like we did for the other switches so that we can test without using the switch itself?

Tratcher: What if we turn the switch on in tests and don't turn it off?

halter73: It would be nice to test the switch functionality by just setting the internal flag without risking the failure of other tests.

Tratcher: *shrug* The switch is pretty trivial, an automated test does not seem necessary. We should opt the test project into HTTP/2 support, there's not much risk for cross contamination as you still need to make a deliberate setting in each test to enable it.

halter73: This is dev, not a patch branch. I think we should have some test for the switch. If we want to have this test project opt-in to the switch, and then have another test project verify it's disabled by default that's also OK I guess.

Tratcher:
*ping*

test/Kestrel.Core.Tests/ListenOptionsTests.cs
```diff
@@ -14,5 +15,13 @@ public void ProtocolsDefault()
             var listenOptions = new ListenOptions(new IPEndPoint(IPAddress.Loopback, 0));
             Assert.Equal(HttpProtocols.Http1, listenOptions.Protocols);
         }
+
+        [Fact]
+        public void Http2DisabledByDefault()
+        {
+            var listenOptions = new ListenOptions(new IPEndPoint(IPAddress.Loopback, 0));
+            Assert.Throws<NotSupportedException>(() => listenOptions.Protocols = HttpProtocols.Http1AndHttp2);
```
halter73: Can you assert the exception message matches what's in CoreStrings.Http2NotSupported?

src/Kestrel.Core/ListenOptions.cs
```diff
@@ -19,13 +19,18 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core
     /// </summary>
     public class ListenOptions : IEndPointInformation, IConnectionBuilder
     {
+        internal const string Http2ExperimentSwitch = "Switch.Microsoft.AspNetCore.Server.Kestrel.Experimental.Http2";
+
         private FileHandleType _handleType;
+        private HttpProtocols _protocols = HttpProtocols.Http1;
+        internal bool _isHttp2Supported;
```
halter73: Nit: Supported -> Enabled.

Tratcher: Supported seems more accurate. _protocols tracks what's enabled.

Issue #1809, opened 2017-05-04
Remove Core project reference from CodeGenerator


halter73:
The dependency was added in #1418 to access to some types like HttpMethod. The downside is that if the Core project doesn't compile (perhaps due to a code generation error), the CodeGenerator also can't compile.

muratg:
@jkotalik Please ping @halter73 for this.

PR #2275, opened 2018-01-25
React to pipelines changes

pakrym:


samples/SampleApp/Startup.cs
```diff
@@ -76,7 +76,7 @@ public static Task Main(string[] args)
                     });
 
                     // Run callbacks on the transport thread
-                    options.ApplicationSchedulingMode = SchedulingMode.Inline;
+                    //options.ApplicationSchedulingMode = SchedulingMode.Inline;
```
davidfowl: ?

samples/SampleApp/Startup.cs
```diff
@@ -130,7 +130,7 @@ public static Task Main(string[] args)
                 .UseContentRoot(Directory.GetCurrentDirectory())
                 .UseStartup<Startup>();
 
-            if (string.Equals(Process.GetCurrentProcess().Id.ToString(), Environment.GetEnvironmentVariable("LISTEN_PID")))
+            //if (string.Equals(Process.GetCurrentProcess().Id.ToString(), Environment.GetEnvironmentVariable("LISTEN_PID")))
```
davidfowl: ?

pakrym:
:up::date:

src/Kestrel.Core/Internal/Http/Http1MessageBody.cs
```diff
@@ -62,19 +63,19 @@ protected Http1MessageBody(Http1Connection context)
 
                         if (!readableBuffer.IsEmpty)
                         {
-                            var writableBuffer = _context.RequestBodyPipe.Writer.Alloc(1);
                             bool done;
 
                             try
                             {
-                                done = Read(readableBuffer, writableBuffer, out consumed, out examined);
+                                _context.RequestBodyPipe.Writer.GetMemory();
```
davidfowl: Is this the same?

tools/CodeGenerator/CodeGenerator.csproj
```diff
@@ -7,9 +7,6 @@
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
   </PropertyGroup>
 
-  <ItemGroup>
-    <ProjectReference Include="..\..\src\Kestrel.Core\Kestrel.Core.csproj" />
-  </ItemGroup>
```
halter73: :tada: #1809

src/Kestrel.Core/Internal/Http/HttpParser.cs
```diff
@@ -279,16 +280,20 @@ public unsafe bool ParseHeaders(TRequestHandler handler, ReadOnlyBuffer buffer,
                             else
                             {
                                 var current = reader.Position;
+                                var currentSlice = buffer.Slice(current, bufferEnd);
```
davidfowl: @pakrym Won't this performance? This is why I didn't understand why we were removing seek that just took cursors. This means we end up slicing more than we need to.

test/Kestrel.Core.Tests/HttpParserTests.cs
```diff
@@ -384,16 +384,16 @@ public void ExceptionDetailNotIncludedWhenLogLevelInformationNotEnabled()
             Assert.Single(pairs);
             Assert.Equal(headerName, pairs[0].Key);
             Assert.Equal(expectedHeaderValue, pairs[0].Value);
-            Assert.Equal(buffer.End, consumed);
-            Assert.Equal(buffer.End, examined);
+            Assert.True(buffer.Slice(consumed).IsEmpty);
```
davidfowl: This is pretty bad.

pakrym: We can add `ROB.PositionsEqual(p1, p2)`

davidfowl:
Are you meeting this into dev?

src/Kestrel.Core/Internal/Http/HttpParser.cs
davidfowl: Hopefully we don’t regress performance any more than we already have

pakrym:
Yes, I need to prepare other repos first, versions would be different.

PR #2278, opened 2018-01-26
Update kestrel error message

javiercn:


javiercn:
The message was provided by @danroth27 

@natemcmaster Is the command to install the tool correct? that's the minimal thing that worked on my machine (plus a source parameter, but we won't need that).

src/Kestrel.Core/CoreStrings.resx
```diff
@@ -490,7 +490,9 @@
     <value>The endpoint {endpointName} is missing the required 'Url' parameter.</value>
   </data>
   <data name="NoCertSpecifiedNoDevelopmentCertificateFound" xml:space="preserve">
-    <value>Unable to configure HTTPS endpoint. No server certificate was specified and the default developer certificate could not be found. Try running 'dotnet developercertificates https -t' to setup a developer certificate for use with localhost. For information on configuring HTTPS see https://go.microsoft.com/fwlink/?linkid=848054</value>
+    <value>Unable to configure HTTPS endpoint. No server certificate was specified, and the default developer certificate could not be found.
+To install the developer certificate first install the dev-certs tool by running 'dotnet install tool dotnet-dev-certs -g --version 2.1.0-preview1' and then run 'dotnet-dev-certs https'. To trust the certificate (Windows and macOS only) run 'dotnet-dev-certs https --trust'.
```
natemcmaster: `--version 2.1.0-preview1-*` or `--version 2.1.0-preview1-final`. Start would work better for nightly builds.

natemcmaster:
You should probably target the <kbd>release/2.1</kbd> branch. 

cc @muratg

javiercn:
@natemcmaster Yeah, don't worry. I know the drill. I'll put the change on top of 2.1, merge it to dev, update this branch and push dev

src/Kestrel.Core/CoreStrings.resx
javiercn: I tried * and didn't work. Will the final version be sufixed final?

halter73: Does `dotnet-dev-certs https --trust` need elevated permissions?

javiercn: No. The tool elevates permissions when it needs to and asks you for credentials

natemcmaster: Our released version should be 2.1.0-preview1-final.

Hmm, so  `--version 2.1.0-*` doesn't work? Seems like a good thing for dotnet-install-tool to support.

PR #2279, opened 2018-01-27
Fixed race in sockets transport

davidfowl:
Testing a theory.

@halter73 I can't seem to reproduce the error with sockets alone but my guess is that we're still shutting down to early. Based on the changes you made earlier (https://github.com/aspnet/KestrelHttpServer/commit/f4d27e67bbe770b0290f1a08c2e3401bae8686f0), we trigger OnConnectionClosed before the socket is disposed in the SocketTransport. This moves the call to Output.Complete to happen after and thus fixes the race.

I don't like returning the exception but it's the cleanest thing I can think of to not expose all of the exception handling to the caller. I'll see if we can make it a bit cleaner.

### DIFF
--- a/src/Kestrel.Transport.Sockets/Internal/SocketConnection.cs
+++ b/src/Kestrel.Transport.Sockets/Internal/SocketConnection.cs
@@ -57,6 +57,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Transport.Sockets.Internal
 
         public async Task StartAsync(IConnectionHandler connectionHandler)
         {
+            Exception error = null;
             try
             {
                 connectionHandler.OnConnection(this);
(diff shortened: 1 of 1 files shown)


davidfowl:
OSX failure: Sending100ContinueDoesNotStartResponse

src/Kestrel.Transport.Sockets/Internal/SocketConnection.cs
```diff
@@ -57,13 +57,14 @@ internal SocketConnection(Socket socket, MemoryPool memoryPool, ISocketsTrace tr
 
         public async Task StartAsync(IConnectionHandler connectionHandler)
         {
+            Exception error = null;
```
halter73: Nit: rename to sendError.

PR #2280, opened 2018-01-28
Use the OnWriterCompleted and OnReaderCompleted events directly.

davidfowl:
Since the leak is fixed this should be fine now.


davidfowl:
Failure is fixed by https://github.com/aspnet/KestrelHttpServer/pull/2279

Issue #740, opened 2016-04-12
Change LoggingThreadPool to run actions that take a state argument


cesarbs:
Right now it only runs `Action`, we should run `Action<T>` to prevent creating closures in places like https://github.com/aspnet/KestrelHttpServer/blob/7b6384635796d99bdb74ba24d80dbdc50731fd5e/src/Microsoft.AspNetCore.Server.Kestrel/Http/Connection.cs#L157

benaadams:
Issue is it already takes a object as parameter which is the logger :cry:

cesarbs:
I'm thinking of closing this. Not worth the time investment. The closure is not created on a critical code path so there's no perf impact.

cc @muratg @halter73 @mikeharder @davidfowl

davidfowl:
Sounds fine to me

muratg:
:+1:

benaadams:
ofc could always have an event on the threadpool that could be listened to rather than crashing the process, then it would free up the slot https://github.com/dotnet/coreclr/issues/2144

PR #2281, opened 2018-01-29
Remove LoggingThreadPool Action allocations

benaadams:
Not sure us worth it, 

but Resolves https://github.com/aspnet/KestrelHttpServer/issues/740 (closed issue)

Pre
image

Post
image

 LoggingThreadPool Action allocations are gone


benaadams:
Isn't really worth it though as is such a small allocation

PR #2287, opened 2018-02-02
Avoid long path 

ryanbrandenburg:
This test fails on the CI due to long path like [this](http://aspnetci/viewLog.html?buildId=399124&tab=buildResultsDiv&buildTypeId=Lite_KestrelStressTest). Let's shorten the name so that it doesn't.

### DIFF
--- a/test/Kestrel.FunctionalTests/ResponseTests.cs
+++ b/test/Kestrel.FunctionalTests/ResponseTests.cs
@@ -2495,10 +2495,9 @@ namespace Microsoft.AspNetCore.Server.Kestrel.FunctionalTests
         [Fact]
         public void ConnectionClosedWhenResponseDoesNotSatisfyMinimumDataRate()
         {
-            using (StartLog(out var loggerFactory))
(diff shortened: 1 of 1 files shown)


test/Kestrel.FunctionalTests/ResponseTests.cs
```diff
@@ -2497,8 +2497,7 @@ public void ConnectionClosedWhenResponseDoesNotSatisfyMinimumDataRate()
         {
             using (StartLog(out var loggerFactory))
             {
-                var logger = loggerFactory.CreateLogger($"{typeof(ResponseTests).FullName}.{nameof(ConnectionClosedWhenResponseDoesNotSatisfyMinimumDataRate)}");
-
+                var logger = loggerFactory.CreateLogger("ConnClosedWhenRespDoesNotSatisfyMinDataRate");
```
anurse: This is changing the logger name, not the test name used to generate the file path. You need to edit the `StartLog` call to use a shorter name rather than the `CallerMemberName` value it currently uses.

ryanbrandenburg:
:up::date:

PR #2293, opened 2018-02-03
Sanitize and centralize exception throws

benaadams:

PR #2290, opened 2018-02-03
Faster IFeatureCollection.Get<TFeature>

benaadams:
Go via Generic rather than wrapping with `(TFeature)Get(typeof(TFeature))` as it scales better

pre
```
                 Method |     Mean |     Error |    StdDev |         Op/s | Allocated |
----------------------- |---------:|----------:|----------:|-------------:|----------:|
     GetViaTypeOf_First | 12.56 ns | 0.1568 ns | 0.2347 ns | 79,608,881.2 |       0 B |
    GetViaGeneric_First | 22.93 ns | 0.1275 ns | 0.1908 ns | 43,613,651.3 |       0 B |
      GetViaTypeOf_Last | 75.97 ns | 0.1121 ns | 0.1678 ns | 13,163,676.0 |       0 B |
     GetViaGeneric_Last | 85.30 ns | 0.2586 ns | 0.3870 ns | 11,723,594.4 |       0 B |
    GetViaTypeOf_Custom | 77.44 ns | 0.2239 ns | 0.3351 ns | 12,912,708.4 |       0 B |
   GetViaGeneric_Custom | 86.67 ns | 0.3066 ns | 0.4590 ns | 11,538,511.2 |       0 B |
  GetViaTypeOf_NotFound | 77.56 ns | 0.2899 ns | 0.4339 ns | 12,893,115.2 |       0 B |
 GetViaGeneric_NotFound | 86.23 ns | 0.2085 ns | 0.3121 ns | 11,597,065.2 |       0 B |
```
post
```
                 Method |     Mean |     Error |    StdDev |         Op/s | Allocated |
----------------------- |---------:|----------:|----------:|-------------:|----------:|
     GetViaTypeOf_First | 11.85 ns | 0.0565 ns | 0.0846 ns | 84,418,379.7 |       0 B |
    GetViaGeneric_First | 21.11 ns | 0.4077 ns | 0.6102 ns | 47,376,257.7 |       0 B |
      GetViaTypeOf_Last | 74.42 ns | 0.3818 ns | 0.5714 ns | 13,437,263.4 |       0 B |
     GetViaGeneric_Last | 40.41 ns | 4.6554 ns | 6.9680 ns | 24,748,132.8 |       0 B |
    GetViaTypeOf_Custom | 76.88 ns | 0.3606 ns | 0.5397 ns | 13,006,456.8 |       0 B |
[...]

PR #2294, opened 2018-02-03
Use enum for method rather than string compares

benaadams:
For GET method `IsHead` test is performed on the string `Method` which since `GET` isn't reference equal to `HEAD` drops through to `String.Equals` -> `OrdinalIgnoreCaseComparer.Equals`

However the Parser starts with it as an Enum, which then gets converted to a string; so it can just be tested with the enum.

Rebased https://github.com/aspnet/KestrelHttpServer/pull/2027

Resolves: https://github.com/aspnet/KestrelHttpServer/issues/2020

### DIFF
--- a/src/Kestrel.Core/Internal/Http/Http1Connection.cs
+++ b/src/Kestrel.Core/Internal/Http/Http1Connection.cs
@@ -181,13 +181,16 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
                 OnAuthorityFormTarget(method, target);
             }
 
-            Method = method != HttpMethod.Custom
-                ? HttpUtilities.MethodToString(method) ?? string.Empty
-                : customMethod.GetAsciiStringNonNullCharacters();
+            Method = method;
+            if (method == HttpMethod.Custom)
+            {
+                _methodText = customMethod.GetAsciiStringNonNullCharacters();
+            }
+
             _httpVersion = version;
 
             Debug.Assert(RawTarget != null, "RawTarget was not set");
-            Debug.Assert(Method != null, "Method was not set");
+            Debug.Assert(((IHttpRequestFeature)this).Method != null, "Method was not set");
             Debug.Assert(Path != null, "Path was not set");
             Debug.Assert(QueryString != null, "QueryString was not set");
             Debug.Assert(HttpVersion != null, "HttpVersion was not set");
--- a/src/Kestrel.Core/Internal/Http/HttpProtocol.FeatureCollection.cs
+++ b/src/Kestrel.Core/Internal/Http/HttpProtocol.FeatureCollection.cs
@@ -11,6 +11,7 @@ using System.Threading.Tasks;
 using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.Http.Features;
 using Microsoft.AspNetCore.Server.Kestrel.Core.Features;
+using Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Infrastructure;
 
 namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
 {
@@ -89,8 +90,21 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
 
         string IHttpRequestFeature.Method
         {
-            get => Method;
-            set => Method = value;
+            get
+            {
+                if (_methodText != null)
+                {
+                    return _methodText;
+                }
+
+                _methodText = HttpUtilities.MethodToString(Method) ?? string.Empty;
+                return _methodText;
+            }
+            set
+            {
+                Method = HttpUtilities.GetKnownMethod(value);
+                _methodText = value;
+            }
         }
 
         string IHttpRequestFeature.PathBase
--- a/src/Kestrel.Core/Internal/Http/HttpProtocol.FeatureCollection.cs
+++ b/src/Kestrel.Core/Internal/Http/HttpProtocol.FeatureCollection.cs
@@ -102,7 +102,6 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
             }
             set
             {
-                Method = HttpUtilities.GetKnownMethod(value);
                 _methodText = value;
             }
         }
--- a/src/Kestrel.Core/Internal/Http/HttpProtocol.cs
+++ b/src/Kestrel.Core/Internal/Http/HttpProtocol.cs
@@ -62,6 +62,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
 
         private readonly IHttpProtocolContext _context;
 
+        protected string _methodText = null;
         private string _scheme = null;
 
         public HttpProtocol(IHttpProtocolContext context)
@@ -119,7 +120,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
         public IPAddress LocalIpAddress { get; set; }
         public int LocalPort { get; set; }
         public string Scheme { get; set; }
-        public string Method { get; set; }
+        public HttpMethod Method { get; set; }
         public string PathBase { get; set; }
         public string Path { get; set; }
         public string QueryString { get; set; }
@@ -324,7 +325,8 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
             MaxRequestBodySize = ServerOptions.Limits.MaxRequestBodySize;
             AllowSynchronousIO = ServerOptions.AllowSynchronousIO;
             TraceIdentifier = null;
-            Method = null;
+            Method = HttpMethod.None;
+            _methodText = null;
             PathBase = null;
             Path = null;
             RawTarget = null;
@@ -905,7 +907,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
         {
             var responseHeaders = HttpResponseHeaders;
 
-            if (!HttpMethods.IsHead(Method) &&
+            if (Method != HttpMethod.Head &&
                 StatusCode != StatusCodes.Status304NotModified &&
                 !responseHeaders.HasTransferEncoding &&
                 responseHeaders.ContentLength.HasValue &&
@@ -1074,7 +1076,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
                 Log.ConnectionKeepAlive(ConnectionId);
             }
 
-            if (HttpMethods.IsHead(Method) && _responseBytesWritten > 0)
+            if (Method == HttpMethod.Head && _responseBytesWritten > 0)
             {
                 Log.ConnectionHeadResponseBodyWrite(ConnectionId, _responseBytesWritten);
             }
@@ -1094,7 +1096,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
                 Log.ConnectionKeepAlive(ConnectionId);
             }
 
-            if (HttpMethods.IsHead(Method) && _responseBytesWritten > 0)
+            if (Method == HttpMethod.Head && _responseBytesWritten > 0)
             {
                 Log.ConnectionHeadResponseBodyWrite(ConnectionId, _responseBytesWritten);
             }
@@ -1124,7 +1126,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
             }
 
             // Set whether response can have body
-            _canHaveBody = StatusCanHaveBody(StatusCode) && Method != "HEAD";
+            _canHaveBody = StatusCanHaveBody(StatusCode) && Method != HttpMethod.Head;
 
             // Don't set the Content-Length or Transfer-Encoding headers
             // automatically for HEAD requests or 204, 205, 304 responses.
@@ -1261,7 +1263,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
         public void HandleNonBodyResponseWrite()
         {
             // Writes to HEAD response are ignored and logged at the end of the request
-            if (Method != "HEAD")
+            if (Method != HttpMethod.Head)
             {
                 ThrowWritingToResponseBodyNotSupported();
             }
--- a/src/Kestrel.Core/Internal/Infrastructure/HttpUtilities.cs
+++ b/src/Kestrel.Core/Internal/Infrastructure/HttpUtilities.cs
@@ -188,7 +188,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Infrastructure
         /// <returns><see cref="HttpMethod"/></returns>
         public static HttpMethod GetKnownMethod(string value)
         {
-            // Called by http/2 and if user code overrides and sets the Method on IHttpRequestFeature
+            // Called by http/2
             if (value == null)
             {
                 throw new ArgumentNullException(nameof(value));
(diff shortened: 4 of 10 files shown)


src/Kestrel.Core/Internal/Http/Http1Connection.cs
```diff
@@ -182,13 +182,16 @@ public void OnStartLine(HttpMethod method, HttpVersion version, Span<byte> targe
...
+            if (method == HttpMethod.Custom)
+            {
+                _customMethod = customMethod.GetAsciiStringNonNullCharacters();
+            }
+
             _httpVersion = version;
 
             Debug.Assert(RawTarget != null, "RawTarget was not set");
-            Debug.Assert(Method != null, "Method was not set");
+            Debug.Assert(((IHttpRequestFeature)this).Method != null, "Method was not set");
```
davidfowl: Why this change?

davidfowl: Ah, because method is an enum.

benaadams: To get the string version

benaadams:
OSX
```
Failed   RegisterAddresses_IPv6_Success(addressInput: "http://*:0/", testUrls: ["http://127.0.0.1", "http://[::1]"])
Error Message:
 System.Net.Sockets.SocketException : Operation timed out
Stack Trace:
   at Microsoft.AspNetCore.Testing.HttpClientSlim.GetSocket(Uri requestUri)
   at Microsoft.AspNetCore.Testing.HttpClientSlim.GetStream(Uri requestUri, Boolean validateCertificate)
   at Microsoft.AspNetCore.Testing.HttpClientSlim.GetStringAsync(Uri requestUri, Boolean validateCertificate)
   at Microsoft.AspNetCore.Testing.HttpClientSlim.GetStringAsync(String requestUri, Boolean validateCertificate)
   at Microsoft.AspNetCore.Server.Kestrel.FunctionalTests.AddressRegistrationTests.RegisterAddresses_Success(String addressInput, String[] testUrls, Int32 testPort) in /_/test/Kestrel.FunctionalTests/AddressRegistrationTests.cs:line 179
   at Microsoft.AspNetCore.Server.Kestrel.FunctionalTests.AddressRegistrationTests.RegisterAddresses_IPv6_Success(String addressInput, String[] testUrls) in /_/test/Kestrel.FunctionalTests/AddressRegistrationTests.cs:line 121
```

benaadams:
Reworded commit to retrigger CI

src/Kestrel.Core/Internal/Infrastructure/HttpUtilities.cs
```diff
@@ -177,6 +177,110 @@ internal static unsafe HttpMethod GetKnownMethod(byte* data, int length, out int
...
+                throw new ArgumentNullException(nameof(value));
+            }
+
+            var length = value.Length;
+            if (length == 0)
+            {
+                throw new ArgumentException(nameof(value));
+            }
+
+            var firstChar = value[0];
```
davidfowl: Why this implementation if its no performance sensitive? Just use the http abstractions methods

benaadams: Is used by http2 so may be perf sensitive :-/

Also are aspnet dlls ngen'd in the newest versions? Does mean none of the methods can be inlined if so.

Did you mean
```csharp
if (HttpMethods.IsGet(value))
{
    method = HttpMethod.Get;
    return true;
}
```

src/Kestrel.Core/Internal/Http/HttpProtocol.FeatureCollection.cs
```diff
@@ -90,8 +90,21 @@ string IHttpRequestFeature.Scheme
...
-            get => Method;
-            set => Method = value;
+            get
+            {
+                if (_customMethod != null)
+                {
+                    return _customMethod;
+                }
+
+                _customMethod = HttpUtilities.MethodToString(Method);
```
halter73: The old code did `HttpUtilities.MethodToString(method) ?? string.Empty`. I'd either keep that or have MethodToString throw instead of return null neither of which should ever actually happen in practice.

src/Kestrel.Core/Internal/Infrastructure/HttpUtilities.cs
halter73: How is this used by HTTP/2? If this is only hit by applications setting the request method themselves, I too would prefer a more readable slightly slower version of this method.

benaadams: HTTP/2 goes via string header
```csharp
((IHttpRequestFeature)this).Method = RequestHeaders[":method"];
```
Though it probably shouldn't use an interface cast for it https://github.com/dotnet/coreclr/issues/16198

benaadams: e.g. first line of `Http2Stream.TryParseRequest(ReadResult result, out bool endConnection)`

src/Kestrel.Core/Internal/Http/HttpProtocol.FeatureCollection.cs
Tratcher: Why store this in _costomMethod? It's a static lookup. You store null for Set.

src/Kestrel.Core/Internal/Infrastructure/HttpUtilities.cs
benaadams: Improved the `TryParseRequest` not to cast via interface and use method directly

src/Kestrel.Core/Internal/Http/HttpProtocol.FeatureCollection.cs
benaadams: done

src/Kestrel.Core/Internal/Http/HttpProtocol.cs
```diff
@@ -324,7 +325,8 @@ public void Reset()
             MaxRequestBodySize = ServerOptions.Limits.MaxRequestBodySize;
             AllowSynchronousIO = ServerOptions.AllowSynchronousIO;
             TraceIdentifier = null;
-            Method = null;
+            Method = HttpMethod.Get;
```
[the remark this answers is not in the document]
benaadams: Added

src/Kestrel.Core/Internal/Http/HttpProtocol.FeatureCollection.cs
benaadams: `GetKnownMethod` returns an `HttpMethod` enum for known methods, but not the string so null.

The field is there anyway for a non-known method, so once you've looked it up once, might as well store it if its checked again rather than doing another lookup,

Tratcher: So why didn't you store it in the Set code path?

Tratcher: If you're going to use it for both then call it `_methodText`

benaadams: Ahh.. that's a very good point 😄

src/Kestrel.Core/Internal/Infrastructure/HttpUtilities.cs
benaadams: Also changed to use the `HttpMethods` strings

src/Kestrel.Core/Internal/Infrastructure/HttpUtilities.cs
```diff
@@ -177,6 +178,110 @@ internal static unsafe HttpMethod GetKnownMethod(byte* data, int length, out int
...
 
+        /// <summary>
+        /// Parses string <paramref name="value"/> for a known HTTP method.
+        /// </summary>
+        /// <remarks>
+        /// A "known HTTP method" can be an HTTP method name defined in the HTTP/1.1 RFC.
+        /// The Known Methods (CONNECT, DELETE, GET, HEAD, PATCH, POST, PUT, OPTIONS, TRACE)
+        /// </remarks>
+        /// <returns><c>true</c> if the input matches a known string, <c>false</c> otherwise.</returns>
+        public static bool GetKnownMethod(string value, out HttpMethod method)
```
JamesNK: Probably should be called `TryGetKnownMethod` since it is following the try pattern of returning a bool with an out param.

src/Kestrel.Core/Internal/Http/HttpProtocol.cs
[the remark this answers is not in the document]
benaadams: Had to change `None = byte.MaxValue` as so indexing and hashing uses the values

src/Kestrel.Core/Internal/Infrastructure/HttpUtilities.cs
benaadams: Changed

benaadams:
K, done - I think

src/Kestrel.Core/Internal/Infrastructure/HttpUtilities.cs
```diff
@@ -135,17 +136,17 @@ public static string GetAsciiStringEscaped(this Span<byte> span, int maxChars)
...
             {
-                method = GetKnownMethod(data, span.Length, out length);
+                method = TryGetKnownMethod(data, span.Length, out length);
                 return method != HttpMethod.Custom;
             }
         }
 
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
-        internal static unsafe HttpMethod GetKnownMethod(byte* data, int length, out int methodLength)
+        internal static unsafe HttpMethod TryGetKnownMethod(byte* data, int length, out int methodLength)
```
davidfowl: This shouldn't be Try

benaadams: Because it always returns a value in `out`?

Reverted

src/Kestrel.Core/Internal/Infrastructure/HttpUtilities.cs
benaadams: Reverted, always returns value in `out`

benaadams:
K, now done... hopefully :)

src/Kestrel.Core/Internal/Infrastructure/HttpUtilities.cs
davidfowl: What's the bool for then?

benaadams: Ah good point can clean this up; return the string

src/Kestrel.Core/Internal/Infrastructure/HttpUtilities.cs
```diff
@@ -177,6 +178,87 @@ internal static unsafe HttpMethod GetKnownMethod(byte* data, int length, out int
...
 
+        /// <summary>
+        /// Parses string <paramref name="value"/> for a known HTTP method.
+        /// </summary>
+        /// <remarks>
+        /// A "known HTTP method" can be an HTTP method name defined in the HTTP/1.1 RFC.
+        /// The Known Methods (CONNECT, DELETE, GET, HEAD, PATCH, POST, PUT, OPTIONS, TRACE)
+        /// </remarks>
+        /// <returns><c>true</c> if the input matches a known string, <c>false</c> otherwise.</returns>
+        public static string GetKnownMethod(string value, out HttpMethod method)
```
JamesNK: This method is taking string `value` and then returning it back again. Why not just return HttpMethod instead of an out param?

Also XML docs need to be updated.

benaadams: Done

src/Kestrel.Core/Internal/Http/HttpProtocol.FeatureCollection.cs
```diff
@@ -90,8 +90,21 @@ string IHttpRequestFeature.Scheme
...
+                {
+                    return _methodText;
+                }
+
+                _methodText = HttpUtilities.MethodToString(Method) ?? string.Empty;
+                return _methodText;
+            }
+            set
+            {
+                Method = HttpUtilities.GetKnownMethod(value);
```
halter73: Out of curiosity, what would happen if we treated every request method set by the app via IHttpRequestFeature.Method as a custom method whether or not it known in order to elide this call to GetKnownMethod?

I don't see how this could be a problem, because none of Kestrel's logic should be predicated on the request method set by the application.

benaadams: So the app couldn't change request behavior based on method? If you changed method from `GET` -> `HEAD` the would mean if the app produced no body, but a content length it would throw with an 500 error.

However that may be better than responding with a `HEAD` response to a `GET` request and then hanging the browser while waiting for data?

Tratcher: While the app should be able to change all of the properties of HttpContext to lie to itself, the server must maintain a source of truth so it can properly interact with the client.

benaadams: Changed

benaadams:
Changed and rebased

src/Kestrel.Core/Internal/Http/HttpProtocol.FeatureCollection.cs
benaadams: This change is really messing up the tests :(

benaadams:
Rebased

halter73:
@pakrym you want to take a look?

davidfowl:
@benaadams can you rebase?

src/Kestrel.Core/Internal/Infrastructure/HttpUtilities.cs
```diff
@@ -177,6 +178,87 @@ internal static unsafe HttpMethod GetKnownMethod(byte* data, int length, out int
...
+        /// Parses string <paramref name="value"/> for a known HTTP method.
+        /// </summary>
+        /// <remarks>
+        /// A "known HTTP method" can be an HTTP method name defined in the HTTP/1.1 RFC.
+        /// The Known Methods (CONNECT, DELETE, GET, HEAD, PATCH, POST, PUT, OPTIONS, TRACE)
+        /// </remarks>
+        /// <returns><see cref="HttpMethod"/></returns>
+        public static HttpMethod GetKnownMethod(string value)
+        {
+            // Called by http/2 and if user code overrides and sets the Method on IHttpRequestFeature
```
davidfowl: This is only used for http2 now, not setting the method directly right?

benaadams: Yes, will change comment

benaadams:
Rebased, changed comment

halter73:
Thanks!

sebastienros:
7% improvement on plaintext running on our physical.

image

benaadams:
😄  What does the `CF` mean on those graphs?

sebastienros:
Old benchmark to analyze the cost of connection adapters.
 
Used here: https://github.com/aspnet/benchmarks/blob/dev/src/Benchmarks/Program.cs#L234
With this filter: https://github.com/aspnet/benchmarks/blob/dev/src/Benchmarks/PassthroughConnectionFilter.cs#L12

halter73:
This really shows what a difference one extra string comparison per request makes. I think `(object.ReferenceEquals(context.ServerSetMethod, HttpMethods.Head))` would have also worked, but moving to enum comparisons was much cleaner.

Thanks @benaadams for finding and fixing this.

benaadams:
Don't think its just down to this one as there were 5 PRs merged at the same time; but glad they had an effect 😄 

Use enum for method rather than string compares (#2294)
Sanitize and centralize exception throws (#2293)
Flatten exception handling (#2313) 
Speed up TryGetAsciiString (#1973)
Faster IFeatureCollection.Get<TFeature> (#2290)

benaadams:
Also none of them in the area of recent regression, so everything still to play for 😉

PR #2295, opened 2018-02-04
Pass StringValues via in

benaadams:
Depends on https://github.com/aspnet/Common/pull/311 (for perf)

Resolves https://github.com/aspnet/KestrelHttpServer/issues/2301

### DIFF
--- a/src/Kestrel.Core/Internal/Http/HttpHeaders.Generated.cs
+++ b/src/Kestrel.Core/Internal/Http/HttpHeaders.Generated.cs
@@ -1252,7 +1252,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
             return MaybeUnknown?.TryGetValue(key, out value) ?? false;
         }
 
-        protected override void SetValueFast(string key, StringValues value)
+        protected override void SetValueFast(string key, in StringValues value)
         {
             switch (key.Length)
             {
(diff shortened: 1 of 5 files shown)


halter73:
This is already compiling despite aspnet/Common#311 not being merged. Does the struct not need to be marked as readonly to use the in keyword for parameters? Do we still need to merge both PRs to avoid the copy?

benaadams:
> Does the struct not need to be marked as readonly to use the in keyword for parameters?

Nope :-|

You can read *fields* of a non-readonly struct without it causing a copy; but not properties or methods (will cause copy)

Also changing to `in` is a code compatible change (not binary compatible).


> Do we still need to merge both PRs to avoid the copy?

Yep

benaadams:
As aside it will give an error when `in` is used with a generic constraint https://github.com/dotnet/roslyn/issues/24601#issuecomment-362699734 because its what can be specified on an interface; which is only properties and methods and no fields; thus is generally bad. Is talk in that issue of adding a `readonly struct` constraint to allow it when it wouldn't be counter productive.

halter73:
I'm OK to merge this as soon as aspnet/Common#311 is merged. I'll want to see benchmark numbers ofc.

davidfowl:
I'll merge it once the CI runs against the latest Common

PR #2298, opened 2018-02-07
Clean up threadpool overrides

benaadams:
* Call through to common non-virtual method (rather than virtual method)
* Use `QueueUserWorkItem(..., preferLocal: true)` when available https://github.com/dotnet/corefx/issues/12442
* Be more explicit about closure capture


davidfowl:
@benaadams @halter73 is working on a change in this area so I suggest we hold off on this until then.

muratg:
Thanks @benaadams!  @halter73 will be looking in this in general (#1662) so closing this PR.

PR #2300, opened 2018-02-07
Removed automatic Upgrade header logic from UpgradeAsync (#2086).

### LINKED ISSUE #2086: Remove automatic Upgrade header logic from UpgradeAsync
https://github.com/aspnet/KestrelHttpServer/blob/a31d1e024ca86a4a9053e0be105fde17c5ccae10/src/Kestrel.Core/Internal/Http/Http1Connection.FeatureCollection.cs#L42-L50
This code copies the request Upgrade header to the response. However, the request Upgrade header is a list of suggestions, while the response is the ordered list of actually selected protocols. Setting any of these response headers is the caller's responsibility. The WebSocket middleware sets these.
https://github.com/aspnet/WebSockets/blob/31c1e9d7d727fda23ae657cb3598c7c3f3950d15/src/Microsoft.AspNetCore.WebSockets/Internal/HandshakeHelpers.cs#L66-L75

https://tools.ietf.org/html/rfc7230#section-6.7

This method is also missing a HasResponseStarted check. It will fail only when it tries to set the StatusCode.

markuspalme:
Fixes #2086.

### DIFF
--- a/src/Kestrel.Core/Internal/Http/Http1Connection.FeatureCollection.cs
+++ b/src/Kestrel.Core/Internal/Http/Http1Connection.FeatureCollection.cs
@@ -8,7 +8,6 @@ using System.Threading.Tasks;
 using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.Http.Features;
 using Microsoft.AspNetCore.Server.Kestrel.Core.Features;
-using Microsoft.Extensions.Primitives;
 
 namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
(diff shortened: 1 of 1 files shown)


markuspalme:
Will get to those tests tomorrow.

muratg:
@markuspalme thanks for the contribution!  Are you still looking into this?

markuspalme:
@muratg Sorry for the late response. I could not look into this, but am not sure this is needed any more because of PR #2329.

davidfowl:
Will close this out

PR #2304, opened 2018-02-07
Include limits default values in intellisense comments.

### LINKED ISSUE #1972: Make Intellisense comments on Limits include default values
When setting Kestrel's limits, the default values for each should show up in the IntelliSense. For example
image

(sorry for image size) is confusing because by setting the property to null, there is no minimum data rate, making it seem like the default minimum data rate is 0 (which it may or may not be).

Redouane64:
Fixes #1972

### DIFF
--- a/src/Kestrel.Core/KestrelServerLimits.cs
+++ b/src/Kestrel.Core/KestrelServerLimits.cs
@@ -43,12 +43,12 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core
         /// Gets or sets the maximum size of the response buffer before write
         /// calls begin to block or return tasks that don't complete until the
         /// buffer size drops below the configured limit.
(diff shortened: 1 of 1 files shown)


src/Kestrel.Core/KestrelServerLimits.cs
```diff
@@ -145,9 +145,9 @@ public int MaxRequestHeaderCount
         /// When set to null, the maximum request body size is unlimited.
         /// This limit has no effect on upgraded connections which are always unlimited.
         /// This can be overridden per-request via <see cref="IHttpMaxRequestBodySizeFeature"/>.
+        /// Defaults to 30,000,000 bytes, which is approximately 28.6MB.
         /// </summary>
         /// <remarks>
```
natemcmaster: I know I'm late to the party...but FYI, you can delete empty remarks blocks

PR #2305, opened 2018-02-08
Add clearer exception message to HttpResponseStream ODE

### LINKED ISSUE #2299: Add clearer exception message to FrameResponseStream ODE
A frequent issue with WebSocket usage is that the user hands the socket off to a background processor and then allows the middleware pipeline to exit. The exception they get in that case is an unpleasant and unhelpful `ObjectDisposedException` for `FrameResponseStream` (a type which the user has never encountered):

```
System.ObjectDisposedException: Cannot access a disposed object.
Object name: 'FrameResponseStream'.
```

The exception is correct, but could be made clearer. I suggest a custom message be added to the exception, something like:

```
System.ObjectDisposedException: Cannot write to the response because the response has ended.
Object name: 'FrameResponseStream'.
```

clarkis117:
This adds a clearer message to the ObjectDisposedException thrown from HttpResponseStream, when a write is attempted after the HTTP response has ended

fixes #2299

### DIFF
--- a/src/Kestrel.Core/CoreStrings.resx
+++ b/src/Kestrel.Core/CoreStrings.resx
@@ -500,7 +500,7 @@ For more information on configuring HTTPS see https://go.microsoft.com/fwlink/?l
   <data name="Http2NotSupported" xml:space="preserve">
     <value>HTTP/2 support is experimental, see https://go.microsoft.com/fwlink/?linkid=866785 to enable it.</value>
   </data>
-  <data name="WritingToResponseAfterResponseEnded" xml:space="preserve">
-    <value>Cannot write to the response because the response has ended.</value>
+  <data name="WritingToResponseBodyAfterResponseCompleted" xml:space="preserve">
+    <value>Cannot write to the response body, the response has completed.</value>
   </data>
 </root>
--- a/src/Kestrel.Core/CoreStrings.resx
+++ b/src/Kestrel.Core/CoreStrings.resx
@@ -500,4 +500,7 @@ For more information on configuring HTTPS see https://go.microsoft.com/fwlink/?l
   <data name="Http2NotSupported" xml:space="preserve">
     <value>HTTP/2 support is experimental, see https://go.microsoft.com/fwlink/?linkid=866785 to enable it.</value>
   </data>
+  <data name="WritingToResponseAfterResponseEnded" xml:space="preserve">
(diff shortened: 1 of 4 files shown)


src/Kestrel.Core/CoreStrings.resx
```diff
@@ -500,4 +500,7 @@ For more information on configuring HTTPS see https://go.microsoft.com/fwlink/?l
   <data name="Http2NotSupported" xml:space="preserve">
     <value>HTTP/2 support is experimental, see https://go.microsoft.com/fwlink/?linkid=866785 to enable it.</value>
   </data>
+  <data name="WritingToResponseAfterResponseEnded" xml:space="preserve">
+    <value>Cannot write to the response because the response has ended.</value>
```
anurse: Wordsmithing time! (I put this text in the bug but I wasn't 100% with it happy yet either ;)). Having `response` in there twice is kinda awkward-sounding.

How about `Cannot write to the response body because the response has completed.` It still has `response` twice, but adding `body` makes it clear that there are two things in play here (the "response" and the "response body")

test/Kestrel.Core.Tests/HttpResponseStreamTests.cs
```diff
@@ -99,8 +99,9 @@ public void StopAcceptingWritesCausesWriteToThrowObjectDisposedException()
             var stream = new HttpResponseStream(Mock.Of<IHttpBodyControlFeature>(), Mock.Of<IHttpResponseControl>());
             stream.StartAcceptingWrites();
             stream.StopAcceptingWrites();
-            Assert.Throws<ObjectDisposedException>(() => { stream.WriteAsync(new byte[1], 0, 1); });
-        }
+			var ex = Assert.Throws<ObjectDisposedException>(() => { stream.WriteAsync(new byte[1], 0, 1); });
```
davidfowl: Tabs!??!!?

clarkis117: looks like my VS settings reset again 🤷🏻 will fix soon

src/Kestrel.Core/CoreStrings.resx
clarkis117: How about "Cannot write to the response body, the response has completed"?

halter73:
Thanks!

PR #2312, opened 2018-02-10
Add Plaintext sample

benaadams:
Keep creating this every time I want a clean test to profiling and testing; have run out of new and interesting variants, so maybe just a standard version...

Benchmarks repo is overly complicated if just interested in a simple implementation


davidfowl:
Looks good

PR #2313, opened 2018-02-11
Flatten exception handling

benaadams:
3 nested try blocks  with 3 finallys (4 total) and 7 catches in same function O_o

Simplifies the request processing loop function by separating it from the connection cleanup, and reducing the blocks to a single try around the user code call.

### DIFF
--- a/src/Kestrel.Core/Internal/Http/HttpProtocol.cs
+++ b/src/Kestrel.Core/Internal/Http/HttpProtocol.cs
@@ -46,7 +46,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
         private CancellationToken? _manuallySetRequestAbortToken;
 
         protected RequestProcessingStatus _requestProcessingStatus;
-        protected volatile bool _keepAlive = true; // volatile, see: https://msdn.microsoft.com/en-us/library/x13ttww7.aspx
+        protected volatile bool _keepAlive; // volatile, see: https://msdn.microsoft.com/en-us/library/x13ttww7.aspx
         protected bool _upgradeAvailable;
         private bool _canHaveBody;
         private bool _autoChunk;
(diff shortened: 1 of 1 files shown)


muratg:
@benaadams is this more of an experiment/wip?

benaadams:
> is this more of an experiment/wip?

Not really; its a proper PR 😄 

Currently `ProcessRequestsAsync` creates for the main loop statemachine il that gets an impressive 7 `try`s deep
```
.try
{
  .try
  {
    .try
    {
      .try
      {
        .try
        {
          .try
          {
            .try
            {
```
It also generates a chunky statemachine struct to lift the values across the exception handling and extra stuff in the method; each object reference requiring a GC memory barrier assign.

image


This PR extracts the main loop into a different method, that only deals with the loop; away from the extra processing that only happens once per connection; also simplifies the exception handling in the loop.

After this change the statemachine for the main loop is only gets two `try`s deep
```
.try
{
  .try
  {
```
And the statemachine looses 6 wrap field (3 object refs); though gains one Exception field; so the state machine struct is reduced by 28 bytes per await and 2 GC memory barriers.

image


Also the main process requests loop is has more clarity of focus and is easier to understand as it is only dealing with processing requests in a loop; rather than additionally handling connection cleanup and failure (which is handled in the outer method)

muratg:
@benaadams thanks that clarifies quite a bit!

benaadams:
> that adding another state machine doesn't regress the performance

Second statemachine is created once and executed once per connection, was paying close attention 😄 
Is why my second commit puts `ConsumeAsync` back inline as that introduced a third statemachine that could be created and executed multiple times per connection.

davidfowl:
I see, it's ProcessRequests(), I thought this was per request but it's still per connection. We'll do a sanity test run.

davidfowl:
@benaadams can you rebase?

benaadams:
Done

src/Kestrel.Core/Internal/Http/HttpProtocol.cs
```diff
@@ -615,6 +483,156 @@ public void OnHeader(Span<byte> name, Span<byte> value)
...
+                        // IHttpContextFactory.Dispose(HttpContext) is called.
+                        await ProduceEnd();
+
+                        // ForZeroContentLength does not complete the reader nor the writer
+                        if (!messageBody.IsEmpty)
+                        {
+                            try
+                            {
+                                // Finish reading the request body in case the app did not.
+                                await messageBody.ConsumeAsync();
```
halter73: How confident are we that this is the only thing that can throw after invoking the application and verifying the response body length?

Personally, I'm fairly confident this is the case, but I want to make extra sure that we dispose the HttpContext in all the common scenarios.

muratg:
Thanks @halter73!

src/Kestrel.Core/Internal/Http/HttpProtocol.cs
benaadams: Was thinking `FireOnStarting` and `FireOnCompleted` if the user code the executed threw; but they don't let exceptions escape.

src/Kestrel.Core/Internal/Http/HttpProtocol.cs
```diff
@@ -615,6 +483,156 @@ public void OnHeader(Span<byte> name, Span<byte> value)
...
+
+                if (badRequestException != null)
+                {
+                    // Handle BadHttpRequestException thrown during app execution or remaining message body consumption.
+                    // This has to be caught here so StatusCode is set properly before disposing the HttpContext
+                    // (DisposeContext logs StatusCode).
+                    SetBadRequestState(badRequestException);
+                }
+
+                application.DisposeContext(httpContext, _applicationException);
```
davidfowl: This not being in a finally makes me nervous, @halter73 I'm hoping our tests are complete here.

src/Kestrel.Core/Internal/Http/HttpProtocol.cs
```diff
@@ -440,139 +440,7 @@ public void OnHeader(Span<byte> name, Span<byte> value)
...
-                            await messageBody.StopAsync();
-
-                            // At this point both the request body pipe reader and writer should be completed.
-                            RequestBodyPipe.Reset();
-                        }
-                    }
-                }
+                await ProcessRequests(application);
             }
             catch (BadHttpRequestException ex)
```
davidfowl: Can this still happen? At this level all error are connection oriented right?

benaadams: Parsing headers and request line can throw a `BadHttpRequestException` which gets caught here

src/Kestrel.Core/Internal/Http/HttpProtocol.cs
benaadams: Only thing would be if `await ProduceEnd` could throw a `BadRequestException`

src/Kestrel.Core/Internal/Http/HttpProtocol.cs
davidfowl: I see

benaadams: Is before any request has started so it catches here and tears down the connection

src/Kestrel.Core/Internal/Http/HttpProtocol.cs
davidfowl: Seems like it could? What happens if you set a utf8 response header before the response starts then you throw from the application?

benaadams: They are check when they are set and would throw then in the application. When they are written out they use `PipelineExtensions.WriteAsciiNoValidation`

benaadams: Taking of which `EncodeAsciiCharsToBytes` looks like it could go faster :)

halter73: I don't think ProduceEnd should ever throw.

src/Kestrel.Core/Internal/Http/HttpProtocol.cs
halter73: Yep. I think this change is safe. I just wanted an extra pair of eyes.

Issue #2276, opened 2018-01-26
Always consume request body for valid non-keep-alive requests before closing connection


halter73:
I'm suggesting that we revert the behavior introduced to fix #418 that doesn't drain the request body for non-keep-alive connections. We should be careful to ensure this doesn't adversely affect nginx scenarios, but I don't think we'd see hangs anymore since we've since gotten rid of `MessageBody.ForRemainingData`.

I'm suggesting changing this now because of the behavior described in [this comment](https://github.com/dotnet/corefx/issues/23504#issuecomment-360639626) in which some clients blow up due to RSTs sent after the entire response was flushed.

alefranz:
I'd like to work on this one. I've submitted PR #2314 and waiting for feedback.
Is there any scripts and nginx config around to easily test #418 after this change?

halter73:
Thanks @alefranz!

PR #2314, opened 2018-02-11
Consumes request before closing connection

alefranz:
Hi,
first time contributor here, I'm trying to work on #2276.

I've spent quite a bit of time on this but I can't see any better way to write a test. Any suggestion?

I've also marked the Upgrade message body as empty, as there is no message to be consumed in an Upgrade request.

### DIFF
--- a/src/Kestrel.Core/Internal/Http/HttpProtocol.cs
+++ b/src/Kestrel.Core/Internal/Http/HttpProtocol.cs
@@ -533,7 +533,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
                             await ProduceEnd();
 
                             // ForZeroContentLength does not complete the reader nor the writer
-                            if (!messageBody.IsEmpty && _keepAlive)
+                            if (!messageBody.IsEmpty)
                             {
                                 // Finish reading the request body in case the app did not.
                                 await messageBody.ConsumeAsync();
--- a/test/Kestrel.Core.Tests/Http1ConnectionTests.cs
+++ b/test/Kestrel.Core.Tests/Http1ConnectionTests.cs
@@ -12,6 +12,7 @@ using System.Linq;
 using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
+using Microsoft.AspNetCore.Hosting.Server;
 using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.Http.Features;
 using Microsoft.AspNetCore.Server.Kestrel.Core.Features;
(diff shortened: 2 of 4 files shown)


src/Kestrel.Core/Internal/Http/HttpProtocol.cs
```diff
@@ -533,7 +533,7 @@ public void OnHeader(Span<byte> name, Span<byte> value)
                             await ProduceEnd();
 
                             // ForZeroContentLength does not complete the reader nor the writer
-                            if (!messageBody.IsEmpty && _keepAlive)
```
halter73: I'm not sure I like setting the IsEmpty property true for upgraded connections since it feels a middle misleading. Maybe we can leave the ForUpgrade class as is and change this condition to `if (!messageBody.IsEmpty && !messageBody.RequestUpgrade)` to make things more explicit.

alefranz: The reason why there is no need to consume the body is only because there is no request body.
The upgrade request has always no body as far as I understand, in fact there is a check for the `Content-Length` to be zero [here](https://github.com/alefranz/KestrelHttpServer/blob/de2ecaa53051e669fc4ba9309b9646c30320b826/src/Kestrel.Core/Internal/Http/Http1MessageBody.cs#L233-L238),
so to me it makes sense to set the body as empty. What do you think?

However in the [RFC](https://tools.ietf.org/html/rfc7230#section-6.7) there is no mention that an upgrade request must have no payload (but I'm no expert on this), so I guess this behaviour can potentially change in the feature, but if then an upgrade request can have a body, it will makes sense to always ensure the body is consumed.

halter73: @alefranz You've convinced me. Technically we're not dealing with a message body for upgraded requests.

test/Kestrel.Core.Tests/Http1ConnectionTests.cs
```diff
@@ -792,6 +793,34 @@ public void ThrowsWhenMaxRequestBodySizeIsSetToANegativeValue()
...
+
+            var requestCompleted = false;
+            _application.Output.OnReaderCompleted((exception, state) => { requestCompleted = true; }, null);
+
+            var data = Encoding.ASCII.GetBytes("POST / HTTP/1.1\r\nHost:\r\nConnection: close\r\ncontent-length: 1\r\n\r\n");
+            await _application.Output.WriteAsync(data);
+
+            // Wait for ProcessRequestsAsync to process the request
+            // Unfortunately there is no way to know when it is waiting on the request body
+            await Task.Delay(150);
```
halter73: Get rid of the delay and instead use a ManualResetEventSlim and wait for that with a (very long) timeout. E.g:

```c#
var requestCompletedMre = new ManualResetEventSlim();

// ...

_application.Output.OnWriterCompleted((exception, state) => { requestCompletedMre.Set() }, null);

// ...

Assert.True(requestCompletedMre.Wait(TestConstants.DefaultTimeout);
```

halter73: Ignore the previous comment. I misread the test. I see you're verifying that OnReaderCompleted *isn't* called. The problem I see is that unless you wire up something to consume the application output, there will never be a reader to get completed, so OnReaderCompleted can't be called.

alefranz: Sorry it's not really clear this test.
The idea is that, if it is not waiting for the body [here](https://github.com/aspnet/KestrelHttpServer/pull/2314/files/de2ecaa53051e669fc4ba9309b9646c30320b826#diff-71145e700087089041944298eab15d6cR539) it will carry on and complete the pipe [here](https://github.com/aspnet/KestrelHttpServer/pull/2314/files/de2ecaa53051e669fc4ba9309b9646c30320b826#diff-71145e700087089041944298eab15d6cR603) given it is not a connection to be kept alive.

In the end this is the same as checking if the requestProcessingTask has complete, but I thought it was better to check the pipe given that the test is about making sure it doesn't close the connection before consuming the body. The task termination feels like an implementation detail.

This is both testing that the request is not completed before receiving the body and that it is completed afterwards, otherwise it would timeout on waiting the requestProcessingTask. Probably for symmetry it can be tested `Assert.True(requestCompleted);` but testing that the connection is closed after it's not really the responsibility of this test.

I don't like this test at all and I'm happy to scrap it.

I can have another go tomorrow to see if I can mock the pipe to be notified when the processing task is waiting for data. Or any other suggestion is welcome.

halter73: This test is making more sense to me now. I have to admit the transport/application pipe names always mess me up, so that was part of the confusion.

I wonder if you would be better off creating a custom mock Http1Connection that overrides CreateMessageBody() to return a mock MessageBody.

At the end of the test after the requestProcessingTask completes, you can then verify that ConsumeAsync did not got called on the mock message body. This way we don't need to introduce a new arbitrary timeouts that always seem to cause problems no matter how conservative we make them.

Thanks!

alefranz: Good call! I didn't notice `CreateMessageBody` was virtual.
I haven't created a separate test object or injected a `MessageBody` factory in the test object as it seems overkill.

I have renamed the test as it is only verified that the body get consumed, not that it actually happens before closing the connection but I believe this is enough and it makes the test a lot more clean.

halter73:
Thanks @alefranz! I used Trainline EU a lot during a recent trip to Germany. I really like the app.

PR #2318, opened 2018-02-14
[WIP] Allow transport to influence application scheduler selection

halter73:
This  started as an effort to avoid extra threadpool dispatching in the socket transport's default configuration. You can already choose to disable threadpool dispatching via KestrelServerOptions, but this PR intended to provide a mechanism for the transport itself to disable threadpool dispatching.

The idea was that the socket transport would default to running app code inline even if the app developer didn't opt in because it's safe and it's faster than dispatching. Unfortunately this posed a few issues such as deciding which scheduler to use for the request body reader so the PumpAsync loop keeps going even if the application does a synchronous read.

@davidfowl and I discussed the possibility of using the inline scheduler for sockets only if the request has no body because that case wouldn't be susceptible to deadlocking due to a synchronous read, but that raised other issues.

We decided if we're going to selectively switch schedulers mid-connection maybe we should always run the parser inline even for the libuv transport. But the logic for the libuv transport can't be exactly like the socket transport, because we don't want to run app code inline even if the request doesn't have a body in the case of the libuv transport.

Then we started discussing how a transport could communicate these kind of subtle scheduling preferences in a way that doesn't make these scheduling preferences something that's specific to the HTTP protocol and not applicable to other connection middleware.

After all this discussion I put this PR on the back burner since we don't have a great design that addresses all of the above concerns. Here's what I have so far. It doesn't do any inline "scheduling" by default yet due to the aforementioned issues. Thoughts?

@benaadams

https://github.com/aspnet/KestrelHttpServer/issues/1662


benaadams:
Associated thoughts...

The `QueueUserWorkItem(..., preferLocal: true)` from https://github.com/aspnet/KestrelHttpServer/pull/2298 means if the thread queuing to the threadpool is already a threadpool thread; it queues to its local queue, so its likely the same thread that will pick it up the work if it finishes quickly after queuing (wouldn't effect libuv, but should effect sockets)

Other thing you could also try is if `Thread.IsThreadPoolThread` then run inline; otherwise queue to threadpool?

benaadams:
Testing `Thread.IsThreadPoolThread` is probably better of the two as then you wouldn't need to allocate a closure for the action or for logging

benaadams:
Something like https://github.com/aspnet/KestrelHttpServer/pull/2319?

src/Kestrel.Core/Internal/Http/HttpProtocol.cs
```diff
@@ -420,7 +420,13 @@ public void Abort(Exception error)
...
 
                 // Potentially calling user code. CancelRequestAbortedToken logs any exceptions.
-                ServiceContext.ThreadPool.UnsafeRun(state => ((HttpProtocol)state).CancelRequestAbortedToken(), this);
+
+                // REVIEW: Should this still QUWI even if the user selects the inline scheduler?
+                // Should we remove the closure allocation? Or maybe allocate the closure once at the start of the connection?
+                // I'm using the ApplicationScheduler for now so we don't have to introduce timeouts in LibuvOutputConsumerTests.
+
+                //ThreadPool.QueueUserWorkItem(state => ((HttpProtocol)state).CancelRequestAbortedToken(), this);
+                _context.ApplicationScheduler.Schedule(CancelRequestAbortedToken);
```
benaadams: Let the scheduler do the closure?
```csharp
_context.ApplicationScheduler.Schedule(state => ((HttpProtocol)state).CancelRequestAbortedToken(), this);
```

benaadams:
Moving to the `PipeScheduler`s rather than Kestrel local ones should pick up the `, preferLocal: true`?

So its worth testing if this already improves sockets; even without going inline?

src/Kestrel.Transport.Abstractions/Internal/TransportConnection.cs
```diff
@@ -25,6 +25,8 @@ public TransportConnection()
         public virtual MemoryPool MemoryPool { get; }
         public virtual PipeScheduler InputWriterScheduler { get; }
         public virtual PipeScheduler OutputReaderScheduler { get; }
+        // REVIEW: Should we make all properties non-virtual with get and set?
```
davidfowl: File an issue for this. Why isn't this like the others?

src/Protocols.Abstractions/Features/IConnectionTransportFeature.cs
```diff
@@ -15,5 +15,7 @@ public interface IConnectionTransportFeature
         PipeScheduler InputWriterScheduler { get; }
 
         PipeScheduler OutputReaderScheduler { get; }
+
+        PipeScheduler ApplicationScheduler { get; set; }
```
davidfowl: Why is this get;set;?

src/Kestrel.Core/Internal/Http/HttpProtocol.cs
davidfowl: Yea, let's remove the allocation.

davidfowl:
Ok after looking at this a bit I think we should take it and expose something on SocketOptions that lets us toggle the default behavior. That way we can run in the unsafe mode (where a synchronous Read can deadlock) for benchmarks or in cases where we own all of the code.

src/Protocols.Abstractions/Features/IConnectionTransportFeature.cs
davidfowl: Also, I think we should move this feature (being able to schedule the "application") to Transport abstractions. It really is a contract between Kestrel's transport and Kestrel itself. It doesn't belong in Protocol.Abstractions.

davidfowl:
> Moving to the PipeSchedulers rather than Kestrel local ones should pick up the , preferLocal: true?

> So its worth testing if this already improves sockets; even without going inline?

Good call. That's true.

benaadams:
Thinking about it, if you wanted a true doevents/setimmidate/settimeout(0) type thing

1. Queue an action to the threadpool `(, preferLocal: true)` that had a sidecar awaitable
2. Immediately await the awaitable
3. When it finished have it run the awaitable completion inline

For bonus points you could reuse the awaitable+action for the next request/action

benaadams:
Nah; bad idea, use `(, preferLocal: true)` then it will keep running the `PumpAsync` and if then that awaits the thread will get the item its just queued and run it (almost inline).

If it doesn't await then another thread will run it.

davidfowl:
Closing this PR in favor of #2348

PR #2319, opened 2018-02-14
Run inline when already on ThreadPool

benaadams:
> Unfortunately this posed a few issues such as deciding which scheduler to use for the request body reader so the PumpAsync loop keeps going even if the application does a synchronous read.

Might not work due to this subtly? https://github.com/aspnet/KestrelHttpServer/pull/2318


benaadams:
Oof; yea that's kills it...

PR #2321, opened 2018-02-14
Allows Content-Length for 304 Not Modified response

### LINKED ISSUE #2099: Allow 304 responses with a Content-Length but no body
304 Not Modified responses are not allowed to contain a body, but they may still contain an informative Content-Length header similar to a Head response.

https://tools.ietf.org/html/rfc7230#section-3.3.2
>  A server MAY send a Content-Length header field in a 304 (Not
   Modified) response to a conditional GET request (Section 4.1 of
   [RFC7232]); a server MUST NOT send Content-Length in such a response
unless its field-value equals the decimal number of octets that would
   have been sent in the payload body of a 200 (OK) response to the same
   request.
   A server MUST NOT send a Content-Length header field in any response
   with a status code of 1xx (Informational) or 204 (No Content).  A
   server MUST NOT send a Content-Length header field in any 2xx
   (Successful) response to a CONNECT request (Section 4.3.6 of
   [RFC7231]).

MVC started doing this in 2.0 and running into the `Response Content-Length mismatch: too few bytes written (0 of 42).` error.
https://github.com/aspnet/Mvc/issues/6875
They've patched it for 2.0.1
https://github.com/aspnet/Mvc/issues/6887

alefranz:
Resolves #2099

### DIFF
--- a/src/Kestrel.Core/Internal/Http/HttpProtocol.cs
+++ b/src/Kestrel.Core/Internal/Http/HttpProtocol.cs
@@ -875,6 +875,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
             var responseHeaders = HttpResponseHeaders;
 
             if (!HttpMethods.IsHead(Method) &&
+                StatusCode != StatusCodes.Status304NotModified &&
                 !responseHeaders.HasTransferEncoding &&
                 responseHeaders.ContentLength.HasValue &&
                 _responseBytesWritten < responseHeaders.ContentLength.Value)
--- a/test/Kestrel.FunctionalTests/ResponseTests.cs
+++ b/test/Kestrel.FunctionalTests/ResponseTests.cs
@@ -2662,6 +2662,35 @@ namespace Microsoft.AspNetCore.Server.Kestrel.FunctionalTests
             }
         }
 
+        [Fact]
+        public async Task NonZeroContentLengthFor304StatusCodeIsAllowed()
+        {
+            using (var server = new TestServer(httpContext =>
+            {
+                var response = httpContext.Response;
+                response.StatusCode = StatusCodes.Status304NotModified;
+                response.ContentLength = 42;
+
+                return Task.CompletedTask;
+            }))
+            {
+                using (var connection = server.CreateConnection())
+                {
+                    await connection.Send(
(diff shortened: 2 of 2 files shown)


halter73:
@Tratcher Care to review? Feel free to merge.

Tratcher:
Nice and simple

halter73:
@Tratcher I was thinking is that maybe we should verify that the actual length of the response body is zero for 304 responses instead of just skipping the response-body-too-short validation.

I didn't bring this up in the PR review because it's a separate issue and we'll probably want to do the same validation for HEAD requests. Do you think this is worth addressing?

halter73:
I'm not too familiar with the spec requires wrt this. Is a 304 response ever allowed to have a non-empty response body?

Tratcher:
304 is pretty special, it's allot like a head request. It explicitly must not have a response body, but it is allowed to describe the body it would have had if it were a 200.

alefranz:
@halter73 I thought about the response length validation as well, but as you said is a separate issue. Also it can be considered a breaking change.
I'm happy to have a look if I find a clean way to implement it.
Should I open a issue for discussion?

halter73:
We're good then. We already log and ignore response body writes for HEAD requests and 204/205/304 responses.

I'm assuming we should *not* allow a non-zero content length to be set for 204 and 205 responses.

halter73:
See [here](https://github.com/alefranz/KestrelHttpServer/blob/73256056ed25877db886a23f7c6a0ae423b0dfdd/src/Kestrel.Core/Internal/Http/HttpProtocol.cs#L1096) for the logic that decides whether or not to ignore writes to the response body.

alefranz:
304 is include, see [here](https://github.com/alefranz/KestrelHttpServer/blob/73256056ed25877db886a23f7c6a0ae423b0dfdd/src/Kestrel.Core/Internal/Http/HttpProtocol.cs#L1173)
and it's already handled as write throws if it can't have body.
Tests [here](https://github.com/aspnet/KestrelHttpServer/blob/dev/test/Kestrel.Core.Tests/Http1ConnectionTests.cs#L511-L530)

Nothing else to be done, right?

halter73:
I think we have everything covered. I just wanted someone to agree that allowing a non-zero Content-Length response header for 204 and 205 responses is silly.

PR #2322, opened 2018-02-15
Fixed race in sockets transport (#2279)

halter73:
@muratg We were previously discussing this backport at #2247 which is a closed PR. That particular PR is already merged into the release/2.1 branch. David's change (this one) is all that's remaining to completely rid us of these ArgumentNullExceptions.


muratg:
Let's take this and reduce the noise.

Issue #2035, opened 2017-08-28
Response.OnCompleted not fired AFTER the Response is sent


Tratcher:
_From @JeanCollas on August 28, 2017 0:8_

Using ASP.NET Core 2.0
A simple code like this shows the issue:

    public async Task<IActionResult> TestOnComplete()
    {
            Response.OnCompleted(async () =>
            {
                throw new Exception();
            });
    }

Response.OnCompleted is said to be invoked after the response is sent to the user, so the user should already have got the answer, whatever happens there.

https://github.com/aspnet/HttpAbstractions/blob/master/src/Microsoft.AspNetCore.Http.Abstractions/HttpResponse.cs#L75

>         /// <summary>
>         /// Adds a delegate to be invoked after the response has finished being sent to the client.
>         /// </summary>
>         /// <param name="callback">The delegate to invoke.</param>
>         /// <param name="state">A state object to capture and pass back to the delegate.</param>
>         public abstract void OnCompleted(Func<object, Task> callback, object state);

But this breaks the response. You can also set a breakpoint on the OnComplete action, the user will wait for the response until the breakpoint is passed.

This means that the OnCompleted is fired before the response is actually sent.

_Copied from original issue: aspnet/Home#2170_

Tratcher:
Which server? Kestrel? HttpSys? IIS/Express + Kestrel?
Which version?

Tratcher:
_From @JeanCollas on August 28, 2017 9:2_

IIS Express for the debugging, I haven't verified if it was the same behaviour on Full IIS+Kestrel.

            var host = new WebHostBuilder()
                .UseKestrel()
                .UseContentRoot(Directory.GetCurrentDirectory())
                .UseIISIntegration()
                .UseStartup<Startup>()
                .Build();

Tratcher:
Exceptions from OnCompleted should only ever be logged, they should not interfere with the processing of the response or even prevent other OnCompleted callbacks from running.

halter73:
Looks like the issue is that the `OnCompleted` callbacks are called prior to sending the chunked terminator. When an `OnCompleted` callback throws, `_applicationException` is set and `Frame` which then prevents the chunked terminator from being sent.

Not that this makes things much better, but I've verified that the response is unaffected if a Content-Length is specified

JeanCollas:
The OnCompleted is said to be executed after the sent to the user, so the response time and performances should not be affected by the call of this method.
[...]

PR #2324, opened 2018-02-16
#2035 Do not await OnCompleted handlers before sending the Response

justinwyer:
We no longer await OnCompleted handlers running as part of the Response. This required some minor modifications to existing tests, adding two new tests and changing the intent of one test ThrowingInOnCompletedIsLoggedAndClosesConnection -> ThrowingInOnCompletedIsLogged. Look forward to some feedback 😄

### DIFF
--- a/src/Kestrel.Core/Internal/Http/HttpProtocol.cs
+++ b/src/Kestrel.Core/Internal/Http/HttpProtocol.cs
@@ -737,10 +737,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
             {
                 return Task.CompletedTask;
             }
-            else
-            {
-                return FireOnCompletedAwaited(onCompleted);
-            }
+            return FireOnCompletedAwaited(onCompleted);
         }
 
         private async Task FireOnCompletedAwaited(Stack<KeyValuePair<Func<object, Task>, object>> onCompleted)
--- a/src/Kestrel.Core/Internal/Http/HttpProtocol.cs
+++ b/src/Kestrel.Core/Internal/Http/HttpProtocol.cs
@@ -552,11 +552,6 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
 
                 PauseStreams();
 
-                if (_onCompleted != null)
-                {
-                    await FireOnCompleted();
-                }
-
                 if (badRequestException == null)
                 {
                     // If _requestAbort is set, the connection has already been closed.
@@ -597,6 +592,11 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
                         // meaningful status code to log.
                         StatusCode = 0;
                     }
+                }
+
+                if (_onCompleted != null)
+                {
+                    await FireOnCompleted();
                 }
 
                 if (badRequestException != null)
--- a/src/Kestrel.Core/Internal/Http/HttpProtocol.cs
+++ b/src/Kestrel.Core/Internal/Http/HttpProtocol.cs
@@ -737,6 +737,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
             {
                 return Task.CompletedTask;
             }
+
             return FireOnCompletedAwaited(onCompleted);
         }
 
--- a/src/Kestrel.Core/Internal/Http/HttpProtocol.cs
+++ b/src/Kestrel.Core/Internal/Http/HttpProtocol.cs
@@ -724,7 +724,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
             }
         }
 
-        protected Task FireOnCompleted()
+        protected void FireOnCompleted()
         {
             Stack<KeyValuePair<Func<object, Task>, object>> onCompleted;
             lock (_onCompletedSync)
@@ -733,29 +733,25 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
                 _onCompleted = null;
             }
 
-            if (onCompleted == null)
-            {
-                return Task.CompletedTask;
-            }
-            else
-            {
-                return FireOnCompletedAwaited(onCompleted);
-            }
+            FireOnCompletedAwaited(onCompleted);
         }
 
-        private async Task FireOnCompletedAwaited(Stack<KeyValuePair<Func<object, Task>, object>> onCompleted)
+        private void FireOnCompletedAwaited(Stack<KeyValuePair<Func<object, Task>, object>> onCompleted)
         {
-            foreach (var entry in onCompleted)
+            Task.Run(async () =>
             {
-                try
+                foreach (var entry in onCompleted)
                 {
-                    await entry.Key.Invoke(entry.Value);
+                    try
+                    {
+                        await entry.Key.Invoke(entry.Value);
+                    }
+                    catch (Exception ex)
+                    {
+                        Log.ApplicationError(ConnectionId, TraceIdentifier, ex);
+                    }
                 }
-                catch (Exception ex)
-                {
-                    ReportApplicationError(ex);
-                }
-            }
+            });
         }
 
         public Task FlushAsync(CancellationToken cancellationToken = default(CancellationToken))
--- a/src/Kestrel.Core/Internal/Http/HttpProtocol.cs
+++ b/src/Kestrel.Core/Internal/Http/HttpProtocol.cs
@@ -724,7 +724,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
             }
         }
 
-        protected void FireOnCompleted()
+        protected Task FireOnCompleted()
         {
             Stack<KeyValuePair<Func<object, Task>, object>> onCompleted;
             lock (_onCompletedSync)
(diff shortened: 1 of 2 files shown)


src/Kestrel.Core/Internal/Http/HttpProtocol.cs
```diff
@@ -715,29 +717,25 @@ protected Task FireOnCompleted()
...
-                return FireOnCompletedAwaited(onCompleted);
-            }
+            FireOnCompletedAwaited(onCompleted);
         }
 
-        private async Task FireOnCompletedAwaited(Stack<KeyValuePair<Func<object, Task>, object>> onCompleted)
+        private void FireOnCompletedAwaited(Stack<KeyValuePair<Func<object, Task>, object>> onCompleted)
         {
-            foreach (var entry in onCompleted)
+            Task.Run(async () =>
```
halter73: We don't want to fire and forget the OnCompleted callbacks. It's important to wait on the callbacks in the request processing loop to ensure that we don't dispose the HttpContext prior to the callbacks completing.

I think the fix for #2035 will be as simple as moving the call to `await FireOnCompleted();` to after the call to `await ProduceEnd();`, but I haven't tried it yet.

justinwyer:
@halter73 thanks for the feedback, made the change you requested

test/Kestrel.FunctionalTests/ResponseTests.cs
```diff
@@ -173,6 +174,7 @@ public class ResponseTests : LoggedTest
 
                     Assert.Equal(HttpStatusCode.InternalServerError, response.StatusCode);
                     Assert.False(onStartingCalled);
+                    onCompletedTask.Wait();
```
davidfowl: Why aren't you awaiting this task?

test/Kestrel.FunctionalTests/ResponseTests.cs
```diff
@@ -2034,9 +2095,15 @@ public Task ResponseStatusCodeSetBeforeHttpContextDisposedRequestMalformedReadIg
             }
 
             // All OnCompleted callbacks should be called even if they throw.
+            try
+            {
+                Task.WaitAll(onCompletedCalled1, onCompletedCalled2, Task.Delay(TimeSpan.FromSeconds(1)));
```
davidfowl: We have a TimeoutAfter extension method that other tests use.

test/Kestrel.FunctionalTests/ResponseTests.cs
justinwyer: derp 🙁

test/Kestrel.FunctionalTests/ResponseTests.cs
justinwyer: Thanks will remember for next time

test/Kestrel.FunctionalTests/ResponseTests.cs
```diff
@@ -147,6 +147,7 @@ public class ResponseTests : LoggedTest
         {
             var onStartingCalled = false;
             var onCompletedCalled = false;
+            var onCompleted = Task.Run(() => onCompletedCalled = true);
```
davidfowl: This test modification isn't correct. This task could run immediately and you wouldn't be able to tell if OnCompleted was called or not.

justinwyer:
thanks for taking the time @davidfowl really appreciate it

halter73:
I think there's an issue here where an invalid chunked encoding prevents OnCompleted callbacks from being called. We should add a test for this.

@justinwyer Do you have the time/inclination to write a test for this and try fixing the issue (assuming I'm right) in the next 24 hours? If not, I'll use this PR as a starting point and write the test and fix myself tomorrow. I want to get this fix in before I leave for vacation Saturday.

Thanks!

src/Kestrel.Core/Internal/Http/HttpProtocol.cs
```diff
@@ -719,10 +719,7 @@ protected Task FireOnCompleted()
             {
                 return Task.CompletedTask;
             }
-            else
-            {
-                return FireOnCompletedAwaited(onCompleted);
-            }
+            return FireOnCompletedAwaited(onCompleted);
```
halter73: Nit: missing newline above.

test/Kestrel.FunctionalTests/ResponseTests.cs
```diff
@@ -294,6 +300,67 @@ public Task ResponseStatusCodeSetBeforeHttpContextDisposedRequestMalformedReadIg
...
+        {
+            var hostBuilder = TransportSelector.GetWebHostBuilder()
+                .UseKestrel()
+                .UseUrls("http://127.0.0.1:0/")
+                .Configure(app =>
+                {
+                    app.Run(async context =>
+                    {
+#pragma warning disable CS1998 // Async method lacks 'await' operators and will run synchronously
+                        context.Response.OnCompleted(async () => throw new Exception());
```
halter73: Nit: I think you can get rid of the pragmas by simply removing `async` from your callbacks.

test/Kestrel.FunctionalTests/ResponseTests.cs
```diff
@@ -294,6 +300,67 @@ public Task ResponseStatusCodeSetBeforeHttpContextDisposedRequestMalformedReadIg
...
+                    var response = await client.GetAsync($"http://127.0.0.1:{host.GetPort()}/");
+                    Assert.Equal(HttpStatusCode.OK, response.StatusCode);
+                }
+            }
+        }
+
+        [Fact]
+        public async Task OnCompletedShouldNotBlockAResponse()
+        {
+            var delay = Task.Delay(TimeSpan.FromSeconds(20));
```
halter73: Nit: Use `TestConstants.DefaultTimeout`. If that's too short we might want to consider making the constant larger.

test/Kestrel.FunctionalTests/ResponseTests.cs
```diff
@@ -2035,8 +2094,6 @@ public Task ResponseStatusCodeSetBeforeHttpContextDisposedRequestMalformedReadIg
 
             // All OnCompleted callbacks should be called even if they throw.
             Assert.Equal(2, testLogger.ApplicationErrorsLogged);
-            Assert.True(onCompletedCalled1);
-            Assert.True(onCompletedCalled2);
```
halter73: Why remove these checks?

justinwyer:
@halter73 I have some time now, I'll take a look at it. Ping on gitter for a progress update if you like, or I'll leave further feedback here.

test/Kestrel.FunctionalTests/ResponseTests.cs
justinwyer: OnCompleted takes a Task as it's parameter so I can change this to `Task.Run(() => throw Exception())`.

test/Kestrel.FunctionalTests/ResponseTests.cs
justinwyer: Will do

test/Kestrel.FunctionalTests/ResponseTests.cs
justinwyer: Oh right I can use the overload, silly me.

test/Kestrel.FunctionalTests/ResponseTests.cs
justinwyer: If the OnCompleted handlers are called we should have exactly 2 exceptions. I guess those exceptions could be from elsewhere, so I'll add it back.

justinwyer:
@halter73 hopefully this covers everything 👍 I've included the test, everything appears to be working fine, but please review if the test case is correct.

test/Kestrel.FunctionalTests/ResponseTests.cs
```diff
@@ -173,6 +178,7 @@ public class ResponseTests : LoggedTest
 
                     Assert.Equal(HttpStatusCode.InternalServerError, response.StatusCode);
                     Assert.False(onStartingCalled);
+                    await onCompletedTcs.Task.TimeoutAfter(TestConstants.DefaultTimeout);
                     Assert.True(onCompletedCalled);
```
halter73: Nit: Isn't this assert redundant?

halter73: I think we can get rid of the onCompletedCalled bool.

test/Kestrel.FunctionalTests/ResponseTests.cs
```diff
@@ -294,6 +300,98 @@ public Task ResponseStatusCodeSetBeforeHttpContextDisposedRequestMalformedReadIg
...
+                        "gg");
+                    await connection.ReceiveForcedEnd(
+                        "HTTP/1.1 200 OK",
+                        $"Date: {server.Context.DateHeaderValue}",
+                        "Content-Length: 0",
+                        "",
+                        "");
+                }
+            }
+            Assert.True(onCompletedCalled);
```
halter73: I think you want to use a TaskCompletionSource here like you did for the OnCompleteCalledEvenWhenOnStartingNotCalled test since the response can be fully written prior to OnCompleted being called.

test/Kestrel.FunctionalTests/ResponseTests.cs
```diff
@@ -294,6 +300,98 @@ public Task ResponseStatusCodeSetBeforeHttpContextDisposedRequestMalformedReadIg
...
+                        "",
+                        "gg");
+                    await connection.ReceiveForcedEnd(
+                        "HTTP/1.1 200 OK",
+                        $"Date: {server.Context.DateHeaderValue}",
+                        "Content-Length: 0",
+                        "",
+                        "");
+                }
+            }
```
halter73: Nit: newline.

halter73:
@justinwyer Thanks! It looks great. Just a few minor comments on the tests.

halter73:
Note to others. Don't merge until the minor test changes are in.

test/Kestrel.FunctionalTests/ResponseTests.cs
```diff
@@ -294,6 +297,99 @@ public Task ResponseStatusCodeSetBeforeHttpContextDisposedRequestMalformedReadIg
...
+                    var response = await client.GetAsync($"http://127.0.0.1:{host.GetPort()}/");
+                    Assert.Equal(HttpStatusCode.OK, response.StatusCode);
+                }
+            }
+        }
+
+        [Fact]
+        public async Task OnCompletedShouldNotBlockAResponse()
+        {
+            var delay = Task.Delay(TestConstants.DefaultTimeout);
```
davidfowl: Do you really want this delay to star this early?

halter73: I think it's OK. It makes it easy to access the `delay` Task from server and client parts of the test. Unless we're worried about this test ever taking over 30 seconds.

halter73:
Thanks!

justinwyer:
My pleasure @halter73 thanks for all of the great feedback from yourself and @davidfowl. Sorry I wasn't able to get any keyboard time yesterday.

PR #2326, opened 2018-02-16
Remove unnecessary usings

mikeharder:


tools/CodeGenerator/HttpProtocolFeatureCollection.cs
```diff
@@ -4,8 +4,6 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
-using Microsoft.AspNetCore.Http.Features;
-using Microsoft.AspNetCore.Http.Features.Authentication;
 //using Microsoft.AspNetCore.Server.Kestrel.Core.Features;
```
Eilon: Also unused?

mikeharder: Will remove.  I didn't check for commented-out usings :).

mikeharder: Good news is that was the only one

PR #2329, opened 2018-02-17
Removed automatic Upgrade header logic from UpgradeAsync to fix issue 2086

brightcr:
Removed automatic Upgrade header logic from UpgradeAsync to fix issue #2086

- Removed code to set response headers -  connection and upgrade as it should be  set by caller e.g. WebSockets Middleware
- Fixed test cases - as connection would be set to "close" instead of "upgrade" after removing headers

### DIFF
--- a/src/Kestrel.Core/Internal/Http/Http1Connection.FeatureCollection.cs
+++ b/src/Kestrel.Core/Internal/Http/Http1Connection.FeatureCollection.cs
@@ -8,7 +8,6 @@ using System.Threading.Tasks;
 using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.Http.Features;
 using Microsoft.AspNetCore.Server.Kestrel.Core.Features;
-using Microsoft.Extensions.Primitives;
 
 namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
 {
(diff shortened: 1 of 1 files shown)


src/Kestrel.Core/Internal/Http/Http1Connection.FeatureCollection.cs
```diff
@@ -39,15 +38,6 @@ public partial class Http1Connection : IHttpUpgradeFeature
 
             StatusCode = StatusCodes.Status101SwitchingProtocols;
             ReasonPhrase = "Switching Protocols";
-            ResponseHeaders["Connection"] = "Upgrade";
```
davidfowl: I'm not sure we're supposed to be removing the Connection: Upgrade response header.

/cc @Tratcher @halter73

Tratcher: Websockets set that, this is redundant

davidfowl: Who's job is it to set it in general? Why shouldn't the server set the response header? I get removing the other piece but this seems like it should be the case if an upgrade was requested, not a connection close header.

Tratcher: Websocket is the only consumer so there's no clear boundary. Websocket sets a list of headers it needs, including this one.

davidfowl: This is super pedantic but I would keep the connection header, it’s more correct imo. Getting a connection close header after calling upgrade is just incorrect.

khellang:
Wouldn't it be decent to cherry-pick @markuspalme's commit from https://github.com/aspnet/KestrelHttpServer/pull/2300 here? 🤔

src/Kestrel.Core/Internal/Http/Http1Connection.FeatureCollection.cs
brightcr: Agree, updated commits - haven't removed response header of connection upgrade. No change in test cases required.

markuspalme:
Sorry, I didn't have a chance to work on this sooner. Is there anything left to do now?

davidfowl:
Nope, all good.

PR #2330, opened 2018-02-17
Use ReferenceEquals not MulticastDelegate.equals

benaadams:
It never inlines because its chunky equals method 
[MulticastDelegate.cs#L57-L132](https://github.com/dotnet/coreclr/blob/df5a0272078d3102a9e1f84c1b46af3b128b7012/src/mscorlib/src/System/MulticastDelegate.cs#L57-L132)


benaadams:
Added a coreclr PR as everyone keeps doing this https://github.com/dotnet/coreclr/pull/16431 though the inline `ReferenceEquality` will still be faster as it skips the null checks (also crossgen won't inline across assembly boundaries)

PR #2337, opened 2018-02-20
 React to pipes in corefx

pakrym:
I'll move BufferReader and Writer to common.

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -17,5 +17,6 @@
     <PublicSign Condition="'$(OS)' != 'Windows_NT'">true</PublicSign>
     <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
     <LangVersion>latest</LangVersion>
+    <EnableApiCheck>false</EnableApiCheck>
   </PropertyGroup>
 </Project>
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -17,6 +17,7 @@
     <PublicSign Condition="'$(OS)' != 'Windows_NT'">true</PublicSign>
     <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
     <LangVersion>latest</LangVersion>
+    <!-- https://github.com/aspnet/KestrelHttpServer/issues/2350 -->
     <EnableApiCheck>false</EnableApiCheck>
   </PropertyGroup>
 </Project>
--- a/build/sources.props
+++ b/build/sources.props
@@ -7,7 +7,8 @@
       $(RestoreSources);
       https://dotnet.myget.org/F/aspnetcore-dev/api/v3/index.json;
       https://dotnet.myget.org/F/aspnetcore-tools/api/v3/index.json;
-      https://dotnet.myget.org/F/roslyn/api/v3/index.json;
+      https://dotnet.myget.org/F/dotnet-core/api/v3/index.json;
+      d:\localnuget
     </RestoreSources>
     <RestoreSources Condition="'$(DotNetBuildOffline)' != 'true'">
       $(RestoreSources);
--- a/build/sources.props
+++ b/build/sources.props
@@ -7,8 +7,7 @@
       $(RestoreSources);
       https://dotnet.myget.org/F/aspnetcore-dev/api/v3/index.json;
       https://dotnet.myget.org/F/aspnetcore-tools/api/v3/index.json;
-      https://dotnet.myget.org/F/dotnet-core/api/v3/index.json;
-      d:\localnuget
+      https://dotnet.myget.org/F/roslyn/api/v3/index.json;
     </RestoreSources>
     <RestoreSources Condition="'$(DotNetBuildOffline)' != 'true'">
       $(RestoreSources);
--- a/src/Kestrel.Core/Adapter/Internal/AdaptedPipeline.cs
+++ b/src/Kestrel.Core/Adapter/Internal/AdaptedPipeline.cs
@@ -62,7 +62,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Adapter.Internal
 
                     try
                     {
-                        if (result.IsCancelled)
+                        if (result.IsCanceled)
                         {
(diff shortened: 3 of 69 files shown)


build/sources.props
```diff
@@ -8,6 +8,7 @@
       https://dotnet.myget.org/F/aspnetcore-dev/api/v3/index.json;
       https://dotnet.myget.org/F/aspnetcore-tools/api/v3/index.json;
       https://dotnet.myget.org/F/roslyn/api/v3/index.json;
+      D:\localnuget;
```
davidfowl: 🤣

src/Kestrel.Core/Adapter/Internal/AdaptedPipeline.cs
```diff
@@ -126,20 +126,13 @@ public class AdaptedPipeline : IDuplexPipe
                     var outputBuffer = Input.Writer.GetMemory(MinAllocBufferSize);
 
                     var array = outputBuffer.GetArray();
-                    try
```
davidfowl: fix whitespace

src/Kestrel.Transport.Abstractions/baseline.netcore.json
```diff
@@ -3176,7 +3176,7 @@
       "Kind": "Struct",
       "Sealed": true,
       "ImplementedInterfaces": [
```
davidfowl: uh, nah. undo

build/sources.props
benaadams:
```
C:\Users\appveyor\.dotnet\x64\sdk\2.1.300-preview2-008173\NuGet.targets(104,5): error : 
The local source 'D:\localnuget' doesn't exist. [C:\projects\kestrelhttpserver\KestrelHttpServer.sln]
Build FAILED.
```

src/Kestrel.Core/Internal/Http/HttpProtocol.cs
```diff
@@ -899,7 +899,7 @@ private Task WriteChunkedAsync(ArraySegment<byte> data, CancellationToken cancel
 
         private static void WriteChunk(PipeWriter writableBuffer, ArraySegment<byte> buffer)
         {
-            var writer = OutputWriter.Create(writableBuffer);
+            var writer = new BufferWriter(writableBuffer);
```
benaadams: `BufferWriter` will throw an exception if `buffer.Count` > max `MemoryPool` block size?

When its Span isn't big enough to write to it asks for a larger span via  [`Resize`](https://github.com/dotnet/corefxlab/blob/38f273b5383c2eb6a72d7137ce527af484e928d2/src/System.Buffers.Experimental/System/Buffers/Text/BufferWriter.cs#L49-L59) and then copies over the previously written data.
```csharp
private void Resize(int desiredBufferSize = 0)
{
    if (Enlarge == null) throw new BufferTooSmallException();
    // double the size; add one to ensure that an empty buffer still resizes
    if (desiredBufferSize <= 0) desiredBufferSize = _buffer.Length * 2 + 1;
    else if (desiredBufferSize < _buffer.Length + 1) throw new ArgumentOutOfRangeException(nameof(desiredBufferSize));
    var newBuffer = Enlarge(desiredBufferSize).Span;
    if (newBuffer.Length <= _buffer.Length) throw new Exception("Enlarge delegate created too small buffer");
    Written.CopyTo(newBuffer);
    _buffer = newBuffer;
}
```
If its already the max size for the pool it will throw an exception?

/cc @KrzysztofCwalina @aahsonkhan

benaadams: `OutputWriter` would probably be better here as it `Advance`s the pipe leaving previously written data when it runs out of `Span`, then gets a new `Span` for the rest (and repeats)

benaadams: Nevermind, is a different `BufferWriter`

davidfowl:
@pakrym you need to rebase

Directory.Build.props
```diff
@@ -17,5 +17,6 @@
     <PublicSign Condition="'$(OS)' != 'Windows_NT'">true</PublicSign>
     <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
     <LangVersion>latest</LangVersion>
+    <EnableApiCheck>false</EnableApiCheck>
```
davidfowl: ?

src/Kestrel.Transport.Abstractions/Internal/KestrelMemoryPool.cs
```diff
@@ -0,0 +1,12 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Buffers;
+
+namespace Microsoft.AspNetCore.Server.Kestrel.Transport.Abstractions.Internal
+{
+    public static class KestrelMemoryPool
+    {
+        public static MemoryPool<byte> Create() => new MemoryPool();
```
davidfowl: SlabMemoryPool?

Directory.Build.props
davidfowl: You need to undo this.

pakrym: It crashed before, let me try.

src/Kestrel.Core/Internal/Http/HttpResponseHeaders.cs
```diff
@@ -46,10 +46,10 @@ public void CopyTo(ref OutputWriter<PipeWriter> output)
                     {
                         if (value != null)
                         {
-                            output.Write(_CrLf);
-                            PipelineExtensions.WriteAsciiNoValidation(ref output, kv.Key);
-                            output.Write(_colonSpace);
-                            PipelineExtensions.WriteAsciiNoValidation(ref output, value);
+                            buffer.Write(_CrLf);
+                            PipelineExtensions.WriteAsciiNoValidation(ref buffer, kv.Key);
```
davidfowl: Use them like extension methods now

Issue #2341, opened 2018-02-21
FrameRequestStream/HttpRequestStream is not compatible with CryptoStream


kekekeks:
```
System.NotSupportedException: Specified method is not supported.
   at Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http.FrameRequestStream.Flush()
   at System.Security.Cryptography.CryptoStream.FlushFinalBlock()
   at System.Security.Cryptography.CryptoStream.Dispose(Boolean disposing)
```

`CryptoStream.Dispose` is calling `Flush` on the underlying stream even if said stream is read only - https://github.com/dotnet/corefx/issues/27326

Apparently this is intended behavior of `CryptoStream` and `Flush` shouldn't ever throw exceptions: https://github.com/dotnet/corefx/pull/27327#pullrequestreview-98384813

davidfowl:
Fixed by #2342

PR #2342, opened 2018-02-21
Don't throw in HttpRequestStream.Flush

kekekeks:
Because read-only streams apparently can have Flush semantics and this behavior is expected by some of built-in stream wrappers (e. g. CryptoStream)

https://github.com/dotnet/corefx/pull/27327#pullrequestreview-98384813
https://github.com/aspnet/KestrelHttpServer/issues/2341

### DIFF
--- a/src/Kestrel.Core/Internal/Http/HttpRequestStream.cs
+++ b/src/Kestrel.Core/Internal/Http/HttpRequestStream.cs
@@ -38,12 +38,11 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
 
         public override void Flush()
         {
-            throw new NotSupportedException();
         }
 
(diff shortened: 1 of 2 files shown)


davidfowl:
@kekekeks you need to fix the tests

src/Kestrel.Core/Internal/Http/HttpRequestStream.cs
```diff
@@ -38,7 +38,6 @@ public override long Position
 
         public override void Flush()
         {
-            throw new NotSupportedException();
         }
 
         public override Task FlushAsync(CancellationToken cancellationToken)
```
bartonjs: Presumably this should return Task.CompletedTask instead of throwing?

kekekeks:
:ok_hand:

davidfowl:
Thanks!

PR #2343, opened 2018-02-22
Drop Pipelines.Testing dependency that we don't use

pakrym:


natemcmaster:
@muratg @Eilon this is currently targeting release/2.1. It shouldn't really matter for preview1 if you don't want to take it. But we do need this for dev to unblock prodcon

Eilon:
Too late for preview1, we are in escrow.

pakrym:
Rebased

Issue #2340, opened 2018-02-21
Create request body pipe reader by decorating the connection pipe reader


halter73:
This should be more efficient than pumping request body data into a new pipe.

muratg:
Let's punt this away.

jkotalik:
@halter73 Assigning both of us for now. Based on costing, we can decide who actually does this later.

PR #2345, opened 2018-02-24
[WIP] Add request body PipeReader decorator

halter73:
#2340

I'll get some benchmark results as soon as I have a chance.

### DIFF
--- a/src/Kestrel.Core/Internal/Http/Http1MessageBody.cs
+++ b/src/Kestrel.Core/Internal/Http/Http1MessageBody.cs
@@ -133,6 +133,11 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
 
         protected override async Task OnConsumeAsync()
         {
+            if (_context.RequestTimedOut)
+            {
+                _context.ThrowRequestRejected(RequestRejectionReason.RequestTimeout);
+            }
+
             _context.TimeoutControl.SetTimeout(Constants.RequestBodyDrainTimeout.Ticks, TimeoutAction.SendTimeoutResponse);
 
             try
@@ -143,6 +148,12 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
                 do
                 {
                     result = await pipeReader.ReadAsync();
+
+                    if (_context.RequestTimedOut)
+                    {
+                        _context.ThrowRequestRejected(RequestRejectionReason.RequestTimeout);
+                    }
+
                     pipeReader.AdvanceTo(result.Buffer.End);
                 } while (!result.IsCompleted);
             }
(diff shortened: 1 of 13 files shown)


src/Kestrel.Core/Internal/Http2/Http2MessageBody.cs
```diff
@@ -16,22 +18,36 @@ protected Http2MessageBody(Http2Stream context)
...
             }
         }
 
-        protected override Task OnConsumeAsync() => Task.CompletedTask;
-
-        public override Task StopAsync()
+        public override void Stop()
         {
-            _context.RequestBodyPipe.Reader.Complete();
             _context.RequestBodyPipe.Writer.Complete();
```
davidfowl: Shouldn't the RequestBodyPipe go away entirely?

davidfowl: Ah http2

src/Kestrel.Core/Internal/Http/Http1MessageBody.cs
```diff
@@ -236,7 +73,9 @@ private void TryStopTimingReads()
                     BadHttpRequestException.Throw(RequestRejectionReason.UpgradeRequestCannotHavePayload);
                 }
 
-                return new ForUpgrade(context);
+                var body = new ForUpgrade(context);
+                context.RequestBodyPipeReader = new Http1MessageBodyPipeReader(context, body);
```
davidfowl: This allocates a new PipeReader per request. Ideally we could just reuse this like we reuse the Stream. It should be resettable.

src/Kestrel.Core/Internal/Http/Http1MessageBodyPipeReader.cs
```diff
@@ -0,0 +1,275 @@
...
+using System.Collections;
+using System.IO;
+using System.IO.Pipelines;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Infrastructure;
+
+namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
+{
+    public class Http1MessageBodyPipeReader : PipeReader, IAwaiter<ReadResult>
```
davidfowl: Where do we throw                      BadHttpRequestException.Throw(RequestRejectionReason.UnexpectedEndOfRequestContent);
 now?

src/Kestrel.Core/Internal/Http/Http1MessageBodyPipeReader.cs
```diff
@@ -0,0 +1,275 @@
...
+                _lastReadResult = new ReadResult(ReadOnlyBuffer<byte>.Empty, isCancelled: false, isCompleted: false);
+            }
+        }
+
+        private void AdvanceMessageBody(in SequencePosition consumed, in SequencePosition examined)
+        {
+            long consumedBytes;
+
+            // REVIEW: Is it possible for this condition to be false when the entire buffer was consumed?
+            if (consumed == _lastReadResult.Buffer.End)
```
davidfowl: @pakrym is this comparison still safe?

pakrym:
When are we planning to expose body pipe?

muratg:
We decided we're not doing this in PR.

PR #2354, opened 2018-03-02
Add benchmark using InMemoryTransport

mikeharder:
Results from my dev machine:

Method |     Mean |     Error |    StdDev |      Op/s | Allocated |
------------------- |---------:|----------:|----------:|----------:|----------:|
Plaintext | 9.734 us | 0.5443 us | 0.8147 us | 102,729.6 |     232 B |
PlaintextPipelined | 6.002 us | 0.0678 us | 0.1015 us | 166,607.3 |      21 B |

For comparison, here are the `RequestParsing` benchmarks on the same machine:

Method |       Mean |     Error |    StdDev |        Op/s | Scaled | ScaledSD |  Gen 0 | Allocated |
----------------------------------------- |-----------:|----------:|----------:|------------:|-------:|---------:|-------:|----------:|
PlaintextTechEmpower | 1,380.3 ns |  8.413 ns |  12.59 ns |   724,481.9 |   1.00 |     0.00 | 0.0010 |     344 B |
PipelinedPlaintextTechEmpower | 1,137.2 ns | 33.750 ns |  50.52 ns |   879,351.3 |   0.82 |     0.04 | 0.0013 |     344 B |

CC: @benaadams

### DIFF
--- a/benchmarks/Kestrel.Performance/InMemoryTransportBenchmark.cs
+++ b/benchmarks/Kestrel.Performance/InMemoryTransportBenchmark.cs
@@ -52,17 +52,17 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Performance
         }
 
         [Benchmark]
-        public void Plaintext()
+        public async Task Plaintext()
         {
-            _connection.SendRequestAsync(RequestParsingData.PlaintextTechEmpowerRequest).Wait();
-            _connection.GetResponseAsync(_plaintextExpectedResponseLength).Wait();
+            await _connection.SendRequestAsync(RequestParsingData.PlaintextTechEmpowerRequest);
+            await _connection.GetResponseAsync(_plaintextExpectedResponseLength);
         }
 
         [Benchmark(OperationsPerInvoke = RequestParsingData.Pipelining)]
-        public void PlaintextPipelined()
+        public async Task PlaintextPipelined()
         {
-            _connection.SendRequestAsync(RequestParsingData.PlaintextTechEmpowerPipelinedRequests).Wait();
-            _connection.GetResponseAsync(_plaintextPipelinedExpectedResponseLength).Wait();
+            await _connection.SendRequestAsync(RequestParsingData.PlaintextTechEmpowerPipelinedRequests);
+            await _connection.GetResponseAsync(_plaintextPipelinedExpectedResponseLength);
         }
 
         public class InMemoryTransportFactory : ITransportFactory
--- a/benchmarks/Kestrel.Performance/InMemoryTransportBenchmark.cs
+++ b/benchmarks/Kestrel.Performance/InMemoryTransportBenchmark.cs
@@ -1,0 +1,206 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using BenchmarkDotNet.Attributes;
+using Microsoft.AspNetCore.Builder;
+using Microsoft.AspNetCore.Hosting;
+using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Server.Kestrel.Transport.Abstractions.Internal;
+using Microsoft.Extensions.DependencyInjection;
+using System;
+using System.Collections.Generic;
+using System.IO.Pipelines;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Microsoft.AspNetCore.Server.Kestrel.Performance
+{
+    public class InMemoryTransportBenchmark
+    {
+        // Must use explicit line endings to ensure identical string on all platforms
+        private static readonly byte[] _plaintextRequest = Encoding.UTF8.GetBytes(
+            "GET /plaintext HTTP/1.1\r\n" +
+            "Host: localhost\r\n" +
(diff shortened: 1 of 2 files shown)


benaadams:
Add the other Transports then group them; baselining against the InMemory?
```csharp
[CategoriesColumn]
[GroupBenchmarksBy(BenchmarkLogicalGroupRule.ByCategory)]
public class TransportBenchmark
{
    [BenchmarkCategory("Plaintext"), Benchmark(Baseline = true, Description = "InMemory")]
    public void PlaintextInMemory() {}

    [BenchmarkCategory("Plaintext"), Benchmark(Description = "Libuv")]
    public void PlaintextInLibuv() {}

    [BenchmarkCategory("Plaintext"), Benchmark(Description = "Sockets")]
    public void PlaintextSockets() {}

    [BenchmarkCategory("PlaintextPipelined"), Benchmark(Baseline = true, Description = "InMemory")]
    public void PlaintextPipelinedInMemory() {}

    [BenchmarkCategory("PlaintexPipelinedt"), Benchmark(Description = "Libuv")]
    public void PlaintextPipelinedInLibuv() {}

    [BenchmarkCategory("PlaintextPipelined"), Benchmark(Description = "Sockets")]
    public void PlaintextPipelinedSockets() {}
}
```

benaadams:
Actually that doesn't quite work as you'd need something else generating the socket traffic :-/

benchmarks/Kestrel.Performance/InMemoryTransportBenchmark.cs
```diff
@@ -0,0 +1,196 @@
...
+        {
+            _host.Dispose();
+        }
+
+        [Benchmark]
+        public void Plaintext()
+        {
+            _connection.SendRequestAsync(RequestParsingData.PlaintextTechEmpowerRequest).Wait();
+            _connection.GetResponseAsync(_plaintextExpectedResponseLength).Wait();
+        }
```
benaadams: Why not async?
```csharp
public async Task Plaintext()
{
    await _connection.SendRequestAsync(RequestParsingData.PlaintextTechEmpowerRequest);
    await _connection.GetResponseAsync(_plaintextExpectedResponseLength);
}
```

mikeharder: Results are about the same so I'll make this change.  I originally made the method sync to match the `RequestParsing` benchmark methods.

# Sync
Method |      Mean |     Error |    StdDev |      Op/s | Allocated |
------------------- |----------:|----------:|----------:|----------:|----------:|
Plaintext | 11.262 us | 0.5450 us | 0.8157 us |  88,795.7 |     247 B |
PlaintextPipelined |  7.509 us | 0.4845 us | 0.7252 us | 133,177.1 |      21 B |

# Async
Method |      Mean |     Error |    StdDev |      Op/s | Allocated |
------------------- |----------:|----------:|----------:|----------:|----------:|
Plaintext | 11.744 us | 0.5272 us | 0.7890 us |  85,148.1 |     372 B |
PlaintextPipelined |  7.107 us | 0.2767 us | 0.4142 us | 140,696.9 |      30 B |

benchmarks/Kestrel.Performance/InMemoryTransportBenchmark.cs
```diff
@@ -0,0 +1,196 @@
...
+{
+    public class InMemoryTransportBenchmark
+    {
+        private const int _plaintextExpectedResponseLength = 132;
+        private const int _plaintextPipelinedExpectedResponseLength = _plaintextExpectedResponseLength * RequestParsingData.Pipelining;
+
+        private IWebHost _host;
+        private InMemoryConnection _connection;
+
+        [GlobalSetup(Target = nameof(Plaintext) + "," + nameof(PlaintextPipelined))]
```
pakrym: Why do you have Target here but not on cleanup?

pakrym: And why have it here at all?

benchmarks/Kestrel.Performance/InMemoryTransportBenchmark.cs
```diff
@@ -0,0 +1,196 @@
...
+                while (true)
+                {
+                    var result = await Output.ReadAsync();
+                    var buffer = result.Buffer;
+                    var consumed = buffer.Start;
+                    var examined = buffer.End;
+
+                    try
+                    {
+                        if (buffer.Length >= length)
```
pakrym: Let's change this to consume (AdvanceTo end) every time, it reflects better what socket/libuv layer is doing.

benchmarks/Kestrel.Performance/InMemoryTransportBenchmark.cs
mikeharder: Future-proofing for scenarios other than Plaintext.  Setup is specific to Plaintext, but Cleanup just disposes the host so it would apply to all scenarios.

benchmarks/Kestrel.Performance/InMemoryTransportBenchmark.cs
mikeharder: Can you specify exactly what code should be changed to what other code?

pakrym: Roughly:

```
while (length > 0)
{
var result = await Output.ReadAsync();
var buffer = result.Buffer;
length -= buffer.Lenght;
Output.AdvanceTo(buffer.End}
}

if (length< 0 ) throw InvalidOperationException("Unexpected responce");
```

PR #2355, opened 2018-03-02
Limit which features Transport is checked for

benaadams:
e.g. currently its checked for `IServiceProvidersFeature` when hosting is setting up `RequestServices`

/cc @davidfowl

### DIFF
--- a/src/Kestrel.Core/Internal/Http/HttpProtocol.Generated.cs
+++ b/src/Kestrel.Core/Internal/Http/HttpProtocol.Generated.cs
@@ -89,23 +89,23 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
                 object feature = null;
                 if (key == IHttpRequestFeatureType)
                 {
-                    feature = _currentIHttpRequestFeature;
+                    return _currentIHttpRequestFeature;
                 }
                 else if (key == IHttpResponseFeatureType)
                 {
-                    feature = _currentIHttpResponseFeature;
+                    return _currentIHttpResponseFeature;
                 }
                 else if (key == IHttpRequestIdentifierFeatureType)
                 {
-                    feature = _currentIHttpRequestIdentifierFeature;
+                    return _currentIHttpRequestIdentifierFeature;
                 }
                 else if (key == IServiceProvidersFeatureType)
                 {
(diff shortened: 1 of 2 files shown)


src/Kestrel.Core/Internal/Http/HttpProtocol.Generated.cs
```diff
@@ -89,87 +89,87 @@ private void FastReset()
...
+                    return _currentIServiceProvidersFeature;
                 }
                 else if (key == IHttpRequestLifetimeFeatureType)
                 {
-                    feature = _currentIHttpRequestLifetimeFeature;
+                    return _currentIHttpRequestLifetimeFeature;
                 }
                 else if (key == IHttpConnectionFeatureType)
                 {
                     feature = _currentIHttpConnectionFeature;
```
halter73: What's this supposed to achieve? Is returning directly out of the else/if faster or not? Isn't it functionally the same? I don't see why you'd change the pattern for this feature.

benaadams: Its because its also defined on the `ConnectionFeatures`; so if its not picked up from the `HttpProtocol` it then needs to check the `ConnectionFeatures`, so its maintaining previous behavior.

This change isn't inline return vs single return; its don't check `ConnectionFeatures` for everything you can't find on `HttpProtocol`.

In particular at the start of the request Hosting checks for the current `IServiceProvidersFeature` which is unset; so it can set up the request scope object. Currently this first checks `HttpProtocol` and since its `null` it then checks `ConnectionFeatures` which then also returns `null`. After this it will skip the second generic interface dispatch call.

halter73: I see. I forgot about the connection feature fallback. It would be nice if this wasn't so informal.

It makes me think maybe we should go back to explicitly copying connection features to the request feature dictionary like we did before with the PrepareRequest callback in the connection filter.

benaadams: @davidfowl has made a PR for hosting that would resolve this differently https://github.com/aspnet/Hosting/pull/1360

benaadams:
Closing as hosting change was merged

PR #2356, opened 2018-03-02
More detailed request timeout exceptions

Tratcher:
#2245


davidfowl:
@halter73 can you review this please?

halter73:
LGTM. The only other thing I might consider is whether it makes sense to include the configured minimum request data rate in the exception message like we do for the logs.

I also noticed that we don't log the configured minimum *response* data rate when it doesn't get reached like we do for the request data rate. This is all in KestrelTrace. Maybe that could be fixed up before merging this PR.

PR #2367, opened 2018-03-06
Be explicit about PipeScheduler.Inline

davidfowl:
When @pakrym upgraded to the new pipelines, we missed a few places that we were using the inline scheduler before. This change makes that explicit since the new default is thread pool. Hopefully, it'll make the tests less flaky.


pakrym:
Is it required in all these cases?

davidfowl:
@pakrym I have no idea, I'm preserving old behavior.

PR #2371, opened 2018-03-09
Remove KestrelThreadPool abstraction and replaced it with PipeScheduler

davidfowl:
- This should avoid allocations per call to Schedule on .NET Core 2.1 as well.


davidfowl:
Once we figure out the background crash we'll get rid of these allocations:

image

davidfowl:
OneToTenThreads is the failing test

test/Kestrel.Transport.Libuv.Tests/LibuvTransportTests.cs
```diff
@@ -128,6 +131,12 @@ public class LibuvTransportTests
             }
 
             await transport.UnbindAsync();
+
+            if (!await serviceContext.ConnectionManager.CloseAllConnectionsAsync(default).ConfigureAwait(false))
```
halter73: So this is what was causing test errors?

davidfowl: Yep

PR #2381, opened 2018-03-13
Clean up protocol abstractions

davidfowl:
- This change aims to clean up the feature interfaces
 used by kestrel and exposed by protocol absractions. It splits out the
 IConnectionTransportFeature into smaller features that may or may
 not be implemented on the connection.
- Added all of the features from Socket.Abstractions
in an attempt to make it go away completely. As a result
the helper methods and extensions have all been added here.

### DIFF
--- a/src/Kestrel.Core/Internal/ConnectionHandler.cs
+++ b/src/Kestrel.Core/Internal/ConnectionHandler.cs
@@ -28,31 +28,25 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal
 
         private IKestrelTrace Log => _serviceContext.Log;
 
-        public void OnConnection(IFeatureCollection features)
+        public void OnConnection(TransportConnection connection)
         {
-            var connectionContext = new DefaultConnectionContext(features);
-
-            var applicationFeature = connectionContext.Features.Get<IApplicationTransportFeature>();
-            var memoryPoolFeature = connectionContext.Features.Get<IMemoryPoolFeature>();
-            var schedulingFeatue = connectionContext.Features.Get<ITransportSchedulerFeature>();
-
             // REVIEW: Unfortunately, we still need to use the service context to create the pipes since the settings
             // for the scheduler and limits are specified here
-            var inputOptions = GetInputPipeOptions(_serviceContext, memoryPoolFeature.MemoryPool, schedulingFeatue.InputWriterScheduler);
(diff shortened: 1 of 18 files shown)


src/Kestrel.Core/Internal/ConnectionHandler.cs
```diff
@@ -32,12 +32,14 @@ public void OnConnection(IFeatureCollection features)
         {
             var connectionContext = new DefaultConnectionContext(features);
 
-            var transportFeature = connectionContext.Features.Get<IConnectionTransportFeature>();
+            var applicationFeature = connectionContext.Features.Get<IApplicationTransportFeature>();
```
davidfowl: We could expose these as first class properties on a KestrelConnection object but it's only used in this file. The HttpConnectionMiddleware is written against the `ConnectionContext` abstraction.

src/Kestrel.Core/Internal/ConnectionHandler.cs
```diff
@@ -32,12 +32,14 @@ public void OnConnection(IFeatureCollection features)
         {
```
davidfowl: I was thinking about this more and since it's actually the Kestrel specific interface, we should just change this to the `TransportConnection` base class. That's basically the primitive that Kestrel requires to be implemented in order to plug in a transport. It does mean that the `IConnectionHandler` is coupled to this model but that seems like an ok tradeoff.

davidfowl: I'll make this change.

src/Protocols.Abstractions/Features/IConnectionInherentKeepAliveFeature.cs
```diff
@@ -0,0 +1,24 @@
...
+    /// Indicates if the connection transport has an "inherent keep-alive", which means that the transport will automatically
+    /// inform the client that it is still present.
+    /// </summary>
+    /// <remarks>
+    /// The most common example of this feature is the Long Polling HTTP transport, which must (due to HTTP limitations) terminate
+    /// each poll within a particular interval and return a signal indicating "the server is still here, but there is no data yet".
+    /// This feature allows applications to add keep-alive functionality, but limit it only to transports that don't have some kind
+    /// of inherent keep-alive.
+    /// </remarks>
+    public interface IConnectionInherentKeepAliveFeature
```
anurse: Mwahahahaha, the name lives on!

halter73: Got to love some inherency.

PR #2383, opened 2018-03-13
Feature/clean up proto abstractions

davidfowl:
This is a do-over of https://github.com/aspnet/KestrelHttpServer/pull/2381. I need to react in SignalR before I merge this.

Blocked by https://github.com/aspnet/SignalR/pull/1583

### DIFF
--- a/src/Kestrel.Transport.Abstractions/Internal/TransportConnection.Features.cs
+++ b/src/Kestrel.Transport.Abstractions/Internal/TransportConnection.Features.cs
@@ -12,15 +12,24 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Transport.Abstractions.Internal
     public partial class TransportConnection : IFeatureCollection,
                                                IHttpConnectionFeature,
                                                IConnectionIdFeature,
-                                               IConnectionTransportFeature
+                                               IConnectionTransportFeature,
+                                               IMemoryPoolFeature,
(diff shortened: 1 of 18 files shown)


src/Kestrel.Transport.Abstractions/Internal/TransportConnection.Features.cs
```diff
@@ -160,18 +199,30 @@ IDuplexPipe IConnectionTransportFeature.Application
 
         TFeature IFeatureCollection.Get<TFeature>()
         {
-            if (typeof(TFeature) == typeof(IHttpConnectionFeature))
+            if (typeof(TFeature) == IHttpConnectionFeatureType)
```
benaadams: The generic version should use `typeof` for the Jit so its `T` == `T`
`typeof(TFeature) == typeof(IHttpConnectionFeature)`

benaadams: `Type` version is different as its `variable == `

davidfowl: Good to know

davidfowl: I'll fix this in a subsequent commit.

PR #2385, opened 2018-03-13
Clean up some left overs

davidfowl:
- Moved ConnectionBuilder to the root
- Use typeof instead of the type member


davidfowl:
Failures are related to https://github.com/aspnet/KestrelHttpServer/issues/2386

Issue #2386, opened 2018-03-13
Handle posting to the LibuvThread after shutdown


davidfowl:
Since doing https://github.com/aspnet/KestrelHttpServer/commit/e7d3b0c5f971d6367ccec438407e91807a31447c, it exposed a crash we have on shutdown where we're posting to the libuv thread after it's disposed.

```
The active test run was aborted. Reason: Unhandled Exception: System.AggregateException: One or more errors occurred. (Safe handle has been closed) ---> System.ObjectDisposedException: Safe handle has been closed
   at System.Runtime.InteropServices.SafeHandle.DangerousAddRef(Boolean& success)
   at System.StubHelpers.StubHelpers.SafeHandleAddRef(SafeHandle pHandle, Boolean& success)
   at Microsoft.AspNetCore.Server.Kestrel.Transport.Libuv.Internal.Networking.LibuvFunctions.NativeMethods.uv_async_send(UvAsyncHandle handle)
   at Microsoft.AspNetCore.Server.Kestrel.Transport.Libuv.Internal.Networking.LibuvFunctions.async_send(UvAsyncHandle handle) in /_/src/Kestrel.Transport.Libuv/Internal/Networking/LibuvFunctions.cs:line 179
   at Microsoft.AspNetCore.Server.Kestrel.Transport.Libuv.Internal.LibuvThread.Post[T](Action`1 callback, T state) in /_/src/Kestrel.Transport.Libuv/Internal/LibuvThread.cs:line 193
   at System.IO.Pipelines.Pipe.CancelPendingRead()
   at Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http.Http1OutputProducer.Abort(Exception error) in /_/src/Kestrel.Core/Internal/Http/Http1OutputProducer.cs:line 157
   at Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http.HttpProtocol.Abort(Exception error) in /_/src/Kestrel.Core/Internal/Http/HttpProtocol.cs:line 420
   at Microsoft.AspNetCore.Server.Kestrel.Core.Internal.HttpConnection.Abort(Exception ex) in /_/src/Kestrel.Core/Internal/HttpConnection.cs:line 285
   at System.IO.Pipelines.PipeCompletionCallbacks.Execute()
   --- End of inner exception stack trace ---
   at System.IO.Pipelines.PipeCompletionCallbacks.Execute()
[...]

PR #2388, opened 2018-03-13
Handle posting to the libuv thread after StopAsync

davidfowl:
- Check if the post handle is disposed and noop if it is.
We also catch an ObjectDisposedException because it's an inherent race condition.

Fixes https://github.com/aspnet/KestrelHttpServer/issues/2386

### DIFF
--- a/src/Kestrel.Transport.Libuv/Internal/LibuvThread.cs
+++ b/src/Kestrel.Transport.Libuv/Internal/LibuvThread.cs
@@ -177,11 +177,17 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Transport.Libuv.Internal
 
         public void Post<T>(Action<T> callback, T state)
         {
+            // Handle is closed to don't bother scheduling anything
+            if (_post.IsClosed)
+            {
+                return;
+            }
+
             var work = new Work
             {
                 CallbackAdapter = CallbackAdapter<T>.PostCallbackAdapter,
                 Callback = callback,
-                //TODO: This boxes
+                // TODO: This boxes
                 State = state
             };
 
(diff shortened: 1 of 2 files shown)


benaadams:
Would it be better to handle the exception in an outer function; rather than every Post? (e.g. in Abort/Stop)

src/Kestrel.Transport.Libuv/Internal/LibuvThread.cs
```diff
@@ -212,7 +232,15 @@ public Task PostAsync<T>(Action<T> callback, T state)
...
-            _post.Send();
+
+            try
+            {
+                _post.Send();
+            }
+            catch (ObjectDisposedException)
+            {
+                // There's an inherent race here where we're in the middle of shutdown
+            }
```
halter73: nit: newline

benaadams:
i.e does the `catch` hit perf doing it on every post?

halter73:
I'm also a little curious how this affects perf. @davidfowl you're becoming a pro at this, so I'm sure you can get some numbers.

I also think it's just try/catch in Abort. On one hand, any attempt to write or even dispose a libuv handle after a non graceful shutdown will result in an ODE. On the other hand, this was also true before. I just so happens that HttpConnection.Abort was being dispatched to the LoggingThreadPool before so the previous ODEs are no longer being ignored.

davidfowl:
> Would it be better to handle the exception in an outer function; rather than every Post? (e.g. in Abort/Stop)

For performance maybe since it's called much less but I think it's fundamentally broken to allow posting to the disposed LibuvThread. I'll measure the performance.

davidfowl:
There's no performance difference, it's within noise.

benaadams:
Fair enough :)

PR #2389, opened 2018-03-13
Pass ReadOnlySequence via in

benaadams:
Its a chunky `readonly struct`

### DIFF
--- a/src/Kestrel.Core/Internal/Http/HttpParser.cs
+++ b/src/Kestrel.Core/Internal/Http/HttpParser.cs
@@ -50,7 +50,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
                 // No request line end
                 return false;
             }
-            else if (TryGetNewLine(in buffer, out var found))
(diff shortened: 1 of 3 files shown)


src/Kestrel.Core/Internal/Http/HttpParser.cs
```diff
@@ -52,7 +50,7 @@ public unsafe bool ParseRequestLine(TRequestHandler handler, ReadOnlySequence<by
                 // No request line end
                 return false;
             }
-            else if (TryGetNewLine(ref buffer, out var found))
+            else if (TryGetNewLine(in buffer, out var found))
```
muratg: Do we need `in` here? (I heard this would work without adding in on the call site.)

benaadams: Nope, will remove

benaadams: Done

KrzysztofCwalina:
Ah, just saw this one. Disregard my comment in the other PR.

PR #2391, opened 2018-03-14
Added transfer format to Protocol.Abstractions

davidfowl:
For transports that let to choose the mode (binary or text) and ones that only support either (like server sent events only supports text).

### DIFF
--- a/src/Protocols.Abstractions/TransferFormat.cs
+++ b/src/Protocols.Abstractions/TransferFormat.cs
@@ -1,0 +1,14 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+
+namespace Microsoft.AspNetCore.Protocols
+{
+    [Flags]
+    public enum TransferFormat
+    {
+        Binary = 0x01,
+        Text = 0x02
+    }
+}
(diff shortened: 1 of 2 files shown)


src/Protocols.Abstractions/TransferFormat.cs
```diff
@@ -0,0 +1,14 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+
+namespace Microsoft.AspNetCore.Protocols
+{
+    [Flags]
+    public enum TransferFormat
+    {
+        Binary = 0x01,
```
anurse: Binary literal? :). I don't really care either way, mostly just trolling.

anurse:
To clarify the behavior here (might actually be worth a doc comment I guess?). The absence of this feature indicates that the transport supports any data you want and doesn't care what format you're using (i.e. a raw TCP-based transport like Kestrel's transports would never need to set this feature). The only purposes of the feature are: to allow the transport to constrain the type data it's allowed to transfer (i.e. Server-Sent Events only supports text), and/or for the transport to request that the application inform it what kind of data it is sending (to allow the WebSockets transport to select an appropriate frame type)

davidfowl:
All of the features need doc comments. I’ll do it all at once and we can debate them meanings of life on that PR

anurse:
I mean, that was more for @halter73 who I thought you were also asking to review this ;)

halter73:
I only have a vague idea of what this is all about. I figure you have more context about what's required for an SSE transport or whatever.

Issue #2360, opened 2018-03-05
Switch default transport to sockets


natemcmaster:
Currently, it is our goal by 2.1.0 to make sockets the default transport. We made the [switch in early builds of 2.1.0](https://github.com/aspnet/KestrelHttpServer/pull/2100) but reverted that for preview1 (https://github.com/aspnet/KestrelHttpServer/issues/2220) due to performance issues.

Motivations:
 - broad platform support. Wherever managed sockets are supported, Kestrel should work. Doesn't require recompiling libuv to bring up a new OS.
 - source-build. (https://github.com/dotnet/source-build). In current form, Kestrel packages can't be produced in a source-build environment
 - ecosystem. Helps to drive improvement to the .NET sockets implementation, which benefits anyone else using managed sockets.

/cc @halter73 @davidfowl @muratg

Suchiman:
@halter73 shouldn't the announcement prefer `Microsoft.AspNetCore.App` over `Microsoft.AspNetCore.All`?

halter73:
@Suchiman Your edit is correct. That's the main thing that makes the change breaking.

ondravondra:
Do the managed sockets have similar performance compared to libuv?

NinoFloris:
@ondravondra yes that was the prerequisite for the switch and why preview1 still ships with libuv. I think it might even be.a little faster already

nickalbrecht:
Just a heads up, the announcement has the tag `2.0.0-preview2`, not `2.1.0-preview2`

natemcmaster:
@nickalbrecht thanks for spotting that. I updated the milestone on the announcement.

kasper3:
> Do the managed sockets have similar performance compared to libuv?
>> @ondravondra yes that was the prerequisite for the switch and why preview1 still ships with libuv. I think it might even be.a little faster already

Couple of months ago, it was indicated that performance regressed when switching to sockets https://github.com/aspnet/KestrelHttpServer/issues/2220. Are there any before/after performance measurements to bring about concreteness to the mix?

Suchiman:
@kasper3
There is http://aka.ms/aspnet/benchmarks
If you navigate to page 1 with the pager at the bottom, there's a graph that shows Socket I/O by over a million RPS ahead of Libuv.
If you go to page 2 it looks differently so i don't really understand how to read them.

PR #2392, opened 2018-03-14
Déjà vu

halter73:
Default to the managed socket transport

#2360

### DIFF
--- a/src/Kestrel/Kestrel.csproj
+++ b/src/Kestrel/Kestrel.csproj
@@ -17,9 +17,6 @@
   <ItemGroup>
     <ProjectReference Include="..\Kestrel.Core\Kestrel.Core.csproj" />
     <ProjectReference Include="..\Kestrel.Https\Kestrel.Https.csproj" />
-
-    <!-- Even though the Libuv transport is no longer used by default, it remains for back-compat -->
-    <ProjectReference Include="..\Kestrel.Transport.Libuv\Kestrel.Transport.Libuv.csproj" />
     <ProjectReference Include="..\Kestrel.Transport.Sockets\Kestrel.Transport.Sockets.csproj" />
   </ItemGroup>
 
--- a/samples/SampleApp/SampleApp.csproj
+++ b/samples/SampleApp/SampleApp.csproj
@@ -9,6 +9,7 @@
   <ItemGroup>
     <ProjectReference Include="..\..\src\Kestrel\Kestrel.csproj" />
     <ProjectReference Include="..\..\src\Kestrel.Https\Kestrel.Https.csproj" />
+    <ProjectReference Include="..\..\src\Kestrel.Transport.Libuv\Kestrel.Transport.Libuv.csproj" />
   </ItemGroup>
 
   <ItemGroup>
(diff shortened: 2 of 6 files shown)


src/Kestrel/Kestrel.csproj
```diff
@@ -17,9 +17,6 @@
   <ItemGroup>
     <ProjectReference Include="..\Kestrel.Core\Kestrel.Core.csproj" />
     <ProjectReference Include="..\Kestrel.Https\Kestrel.Https.csproj" />
-
-    <!-- Even though the Libuv transport is no longer used by default, it remains for back-compat -->
```
halter73: @natemcmaster I left this comment the first time I switched the default transport to managed sockets. Doesn't removing this dependency break back-compat?

src/Kestrel/Kestrel.csproj
```diff
@@ -17,9 +17,6 @@
...
-    <ProjectReference Include="..\Kestrel.Transport.Libuv\Kestrel.Transport.Libuv.csproj" />
```
halter73: @natemcmaster I left this comment the first time I switched the default transport to managed sockets. Doesn't removing this dependency break back-compat?

davidfowl: This is the break only we're willing to take for 2.1.

halter73: To be specific, I think this change will break any existing app that references just the main Kestrel package and calls UseLibuv.

natemcmaster: Yes, this is an intentional breaking change that we decided is worth taking. It should only affect a small group of people since we are keeping Transport.Libuv in the .All metapackage. It only breaks for people who depend directly on the .Kestrel package, and they can fix this break by adding a package reference to K.T.Libuv when they upgrade to 2.1. IMO we should post this to aspnet/announcements as a heads up.

davidfowl: Yep, that's known. This will be documented as a breaking change in 2.1. 

/cc @DamianEdwards @Eilon

davidfowl:
@halter73 can you also stage various reaction PRs to other repositories using a feature branch (if there are any). I'll send a PR to benchmarks that explicitly adds the libuv package.

halter73:
Yes sir! I get right to staging all the other reaction PRs :wink:

davidfowl:
:trollface:

src/Kestrel/Kestrel.csproj
Eilon: Yup.

halter73: aspnet/Announcements#296

aateeque: Why are you breaking semantic versioning by introducing breaking change in a minor version increment?

davidfowl: Because of the reasons listed here https://github.com/aspnet/KestrelHttpServer/issues/2360. A few things are changing in 2.1 as well. We’re introducing a shared framework for ASP.NET Core and we’d prefer it if libuv wasn’t part of that. Also we feel like it will break few enough people that it shouldn’t matter that much. The workaround for people is also trivial *and* if you’re upgrading from ASP.NET Core 2.0 to 2.1 and you’re using the .All package everything still works.

Where things will break is if you were using .net Framework with libuv and were configuring the API explicitly (to say tweak the number of threads)

PR #2394, opened 2018-03-14
Metadata is now a first class property on ConnectionContext

davidfowl:
- Make IConnectionMetadata a manatory top level feature on ConnectionContext
- TransportConnection will lazily manifest ConnectionMetadata on first access.
This should avoid allocations since Kestrel isn't using this today.

This basically exposes an Items property like what exists on HttpContext, it's extremely useful to round trip arbitrary state per connection from various actors in a system.

### DIFF
--- a/src/Kestrel.Transport.Abstractions/Internal/TransportConnection.cs
+++ b/src/Kestrel.Transport.Abstractions/Internal/TransportConnection.cs
@@ -1,4 +1,5 @@
 ﻿using System.Buffers;
+using System.Collections.Generic;
 using System.IO.Pipelines;
 using System.Net;
 using System.Threading;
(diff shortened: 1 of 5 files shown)


src/Kestrel.Transport.Abstractions/Internal/TransportConnection.cs
```diff
@@ -31,6 +35,19 @@ public TransportConnection()
         public IDuplexPipe Transport { get; set; }
         public IDuplexPipe Application { get; set; }
 
+        public IDictionary<object, object> Metadata
```
anurse: Should it just be called `Items` for consistency?

davidfowl: Yea, I though that a long time ago but I got attached to metadata. What do you think? Should it be Items?

davidfowl: Gah, I'm going to change it to items...

davidfowl:
Replaced by https://github.com/aspnet/KestrelHttpServer/pull/2394. I'll get the reaction ready in SignalR.

anurse:
Can't move. Stuck in an infinite loop...

screen shot 2018-03-14 at 8 57 04 am

PR #2395, opened 2018-03-14
Items is now a first class property on ConnectionContext

davidfowl:
- Make Items a mandatory top level feature on ConnectionContext
- IConnectionItemsFeature is now a requirement for transport layers (SignalR's http transports will implement this as well).
- TransportConnection will lazily manifest ConnectionItems on first access. This should avoid allocations since Kestrel isn't using this today.

This replaces https://github.com/aspnet/KestrelHttpServer/pull/2394 because I need to produce packages (had to change the branch name)

### DIFF
--- a/src/Protocols.Abstractions/Features/IConnectionItemsFeature.cs
+++ b/src/Protocols.Abstractions/Features/IConnectionItemsFeature.cs
@@ -6,8 +6,8 @@ using System.Collections.Generic;
 
 namespace Microsoft.AspNetCore.Protocols.Features
 {
-    public interface IConnectionMetadataFeature
+    public interface IConnectionItemsFeature
     {
-        IDictionary<object, object> Metadata { get; set; }
+        IDictionary<object, object> Items { get; set; }
     }
 }
--- a/src/Kestrel.Transport.Abstractions/Internal/ConnectionItems.cs
+++ b/src/Kestrel.Transport.Abstractions/Internal/ConnectionItems.cs
@@ -7,14 +7,14 @@ using System.Collections.Generic;
 
 namespace Microsoft.AspNetCore.Server.Kestrel.Transport.Abstractions.Internal
 {
-    internal class ConnectionMetadata : IDictionary<object, object>
+    internal class ConnectionItems : IDictionary<object, object>
     {
-        public ConnectionMetadata()
+        public ConnectionItems()
             : this(new Dictionary<object, object>())
         {
         }
 
-        public ConnectionMetadata(IDictionary<object, object> items)
+        public ConnectionItems(IDictionary<object, object> items)
         {
             Items = items;
         }
--- a/src/Kestrel.Transport.Abstractions/Internal/ConnectionMetadata.cs
+++ b/src/Kestrel.Transport.Abstractions/Internal/ConnectionMetadata.cs
@@ -1,0 +1,118 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections;
+using System.Collections.Generic;
+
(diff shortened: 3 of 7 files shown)


src/Protocols.Abstractions/Features/IConnectionItemsFeature.cs
```diff
@@ -6,8 +6,8 @@
 
 namespace Microsoft.AspNetCore.Protocols.Features
 {
-    public interface IConnectionMetadataFeature
+    public interface IConnectionItemsFeature
     {
-        IDictionary<object, object> Metadata { get; set; }
```
halter73: What's going to end up here? Why object -> object instead of string -> object?

davidfowl: https://github.com/aspnet/HttpAbstractions/blob/fbbbb16305f57025eca71964ac5d7c9f43c42150/src/Microsoft.AspNetCore.Http.Features/IItemsFeature.cs Also string keys are inferior.

halter73: I mean strings are just objects after all. I should be embracing the fact we're using objects for keys. It should serve as a bright neon sign telling developers just to create their own features for anything important.

benaadams: Strings are slower keys (due to the fact the whole value of them must be compared)

davidfowl: Yappp!

halter73: @benaadams surely that depends on the efficiency of the key's Equals and GetHashCode methods.

I concede that you could create an object type that's a faster key than string most of the time. Though I wonder if string *does* end up being used as the object key, is it just as fast in a `Dictionary<object, object>` as it is in `Dictionary<string, object>`? Or does the extra compile-time info about the key result in more efficient IL?

benaadams: String in `object` would be slower as it would use randomised hashing; whereas `string` does a bait`n`switch to non-randomised hashing, switching back when collisions get high.

halter73: Thinking about this more, if you're going to use the same object instance as your lookup value every time, strings don't make much sense. So I get your point there. Interesting info about the string hashing. I'll have to take a look at that.

benaadams: [public Dictionary(int capacity, IEqualityComparer<TKey> comparer)](https://github.com/dotnet/coreclr/blob/ffc16020fee4d866de568042fe085143332a2159/src/mscorlib/shared/System/Collections/Generic/Dictionary.cs#L80-L84)
```csharp
if (typeof(TKey) == typeof(string) && _comparer == null)
{
    // To start, move off default comparer for string which is randomised
    _comparer = (IEqualityComparer<TKey>)NonRandomizedStringEqualityComparer.Default;
}
```

halter73:
@davidfowl Interesting you point to IItemsFeature as the HTTP analogue for IConnectionItemsFeature. We do a lot of plumbing for IItemsFeature, but it doesn't look like we use it in any of our samples, (non-API) docs, or framework code.

https://github.com/search?p=1&q=org%3Aaspnet+IItemsFeature&type=Code

davidfowl:
> @davidfowl Interesting you point to IItemsFeature as the HTTP analogue for IConnectionItemsFeature. We do a lot of plumbing for IItemsFeature, but it doesn't look like we use it in any of our samples, (non-API) docs, or framework code.

Search for HttpContext.Items on the interwebs. I'm also plumbing it through SignalR right now.

halter73:
Lol DefaultHttpContext was on the front page of the search results.

halter73:
How could I forget the viewbag-for-middleware!

PR #2396, opened 2018-03-14
Create or update baselines to match 2.0

ryanbrandenburg:


Eilon:
Why did some of the baselines get removed?

ryanbrandenburg:
No baseline files should have been removed, if you expand the ones it looks like GitHub is telling you were deleted in here its just that there were a lot more removals than additions.

Eilon:
Ah ok I was too afraid to try to expand them 😄 So it looks like it's also just `.Internal` stuff got removed, which is fine.

PR #2407, opened 2018-03-19
Never run continuation inline in SocketAwaitable.OnCompleted

halter73:
In the very unlikely event that the awaitable is repeatedly completed between the read/write loop checking IsCompleted and calling OnCompleted, there's the potential the loop could stack dive without the Task.Run.

### DIFF
--- a/src/Kestrel.Transport.Sockets/Internal/SocketAwaitable.cs
+++ b/src/Kestrel.Transport.Sockets/Internal/SocketAwaitable.cs
@@ -7,6 +7,7 @@ using System.IO.Pipelines;
 using System.Net.Sockets;
 using System.Runtime.CompilerServices;
 using System.Threading;
+using System.Threading.Tasks;
 
 namespace Microsoft.AspNetCore.Server.Kestrel.Transport.Sockets.Internal
 {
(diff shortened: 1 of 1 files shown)


davidfowl:
@dotnet-bot build OSX 10.12 Release Build

davidfowl:
@dotnet-bot help

src/Kestrel.Transport.Sockets/Internal/SocketAwaitable.cs
```diff
@@ -47,7 +48,7 @@ public void OnCompleted(Action continuation)
             if (ReferenceEquals(_callback, _callbackCompleted) ||
                 ReferenceEquals(Interlocked.CompareExchange(ref _callback, continuation, null), _callbackCompleted))
             {
-                continuation();
+                Task.Run(continuation);
```
benaadams: QUWI?

benaadams: Task.Run is very expensive

pranavkm:

src/Kestrel.Transport.Sockets/Internal/SocketAwaitable.cs
davidfowl: But this never happens.

PR #2409, opened 2018-03-20
React to aspnet/BuildTools#611

halter73:
This should fix the travis linux builds.

aspnet/BuildTools#611


halter73:
@pranavkm So it seems that environment variables are not retained between travis script sections, so I change this just look into the repo root for the `.dotnet` directory.

PR #2414, opened 2018-03-21
Remove dotnet tool install from instructions on using dotnet-dev-certs

natemcmaster:
This tool ships in-box, so the error message shouldn't instruct users to install it separately.

### DIFF
--- a/src/Kestrel.Core/Properties/CoreStrings.Designer.cs
+++ b/src/Kestrel.Core/Properties/CoreStrings.Designer.cs
@@ -1748,7 +1748,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core
 
         /// <summary>
         /// Unable to configure HTTPS endpoint. No server certificate was specified, and the default developer certificate could not be found.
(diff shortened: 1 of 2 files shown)


src/Kestrel.Core/Properties/CoreStrings.Designer.cs
```diff
@@ -1748,7 +1748,7 @@ internal static string FormatEndpointMissingUrl(object endpointName)
 
         /// <summary>
         /// Unable to configure HTTPS endpoint. No server certificate was specified, and the default developer certificate could not be found.
-        /// To install the developer certificate first install the dev-certs tool by running 'dotnet install tool dotnet-dev-certs -g --version 2.1.0-preview1-final' and then run 'dotnet-dev-certs https'. To trust the certificate (Windows and macOS only) run 'dotnet-dev-certs https --trust'.
+        /// To generate the developer certificate, run 'dotnet dev-certs https'. To trust the certificate (Windows and macOS only) run 'dotnet dev-certs https --trust'.
```
javiercn: To generate a developer certificate run 'dotnet dev-certs https', and to trust the certificate (Windows and macOS only) run 'dotnet dev-certs https --trust'

javiercn: Get sign-off from @danroth27 as this affects the experience. We need to do the same change on the CLI, I was planning to do both next week. Do you want to also take care of the CLI?

javiercn: Probably ignore my comments. Feel free to merge at your earliest convenience

Issue #2350, opened 2018-02-27
Reenable ApiCheck when fixed


pakrym:
https://github.com/aspnet/BuildTools/issues/592

JunTaoLuo:
Follow up needed in #2424

davidfowl:
@pakrym why did you re-open this?

muratg:
@pakrym do we need anything in RC1? If so, we'll need a shiproom label. Otherwise, could you move it to 2.2.0-mq?

pakrym:
If we want to continue to validate API changes - yes, I don't know if it is fixed to be re-enabled.

Fix is being tracked in https://github.com/aspnet/BuildTools/issues/592

PR #2418, opened 2018-03-23
Re-enable api check

JunTaoLuo:
#2350 Unblocked by https://github.com/aspnet/BuildTools/commit/d944172721e77b6ca05a512303a0859634786746


JunTaoLuo:
API check ran but it found a breaking change: https://github.com/aspnet/KestrelHttpServer/commit/fdb4184dbfdbac6ee8b53a4d8d1fbfaa8f60c0c1#diff-dfd6e5bf3bc39e44af0761394afeb8ecR16 and https://github.com/aspnet/KestrelHttpServer/commit/c3ba875d12cd43eeee1a2ab3a3a4ee3adac197a8#diff-dfd6e5bf3bc39e44af0761394afeb8ecR18. Should I add an exception or do we need to add an overload for the constructor?
cc @halter73

davidfowl:
That package was never released to nuget.org so it's fine.

JunTaoLuo:
@davidfowl in that case I think we should remove the baseline.json file.

davidfowl:
> @davidfowl in that case I think we should remove the baseline.json file right?

Yep!

JunTaoLuo:
@dotnet-bot test this please

JunTaoLuo:

JunTaoLuo:
API breaking change in https://github.com/aspnet/KestrelHttpServer/commit/abdcb47b8f96d7bf6554575e2d3493a7e955fc68

davidfowl:
@JunTaoLuo I think you should skip libuv for API check and file a blocking bug in preivew2, then we need to discuss how we treat pubternal API changes that are in public APIs (specifically this public API).

/cc @DamianEdwards @Eilon

Eilon:
@davidfowl it isn't clear to me what the breaking change is here by looking at the linked diffs. Changes in public APIs that use pubternals are not considered breaking changes by us, so those would be OK.

JunTaoLuo:
@Eilon Actually it's not a pubternal breaking change. The method signature of the constructor of LibuvTransportFactory changed and it's a public type. I believe the discussion is that this type should have remained pubternal.

JunTaoLuo:
Filed follow up issue at: https://github.com/aspnet/KestrelHttpServer/issues/2424. I want to unblock the API checks for the other packages ASAP.

halter73:
IMHO The issue is that LibuvTransportFactory never should have been public. The only public way to use or configure libuv should be through the UseLibuv extension methods on IWebHostBuilder.

JunTaoLuo:

halter73:
@ryanbrandenburg Are you saying that `src/Kestrel.Transport.Sockets/baseline.netcore.json` should be regenerated from the release/2.1 branch since it hasn't been shipped yet? If so, do we just regenerate the baseline if we change public API prior to 2.1 RTM?

ryanbrandenburg:
@halter73 no, I'm saying that if we haven't shipped Kestrel.Transport.Sockets yet we should have a _blank_ file at `src/Kestrel.Transport.Sockets/baseline.netcore.json`. It indicates to future travelers "this space intentionally left blank", rather than that we just forgot to generate it.

halter73:
@ryanbrandenburg @pranavkm Will either of you be mad if I cherry-pick b90766d09755c663e2af868380006213af1171f4 into Kestrel's release/2.1 branch?

JunTaoLuo:
Absolutely furious, but I'll get over it. /s

ryanbrandenburg:
@halter73 not as long as you then merge the resulting commit back into dev.

PR #2420, opened 2018-03-23
Make DefaultConnectionContext a poco

davidfowl:
- Made TransportConnecton derive from ConnectionContext
- Less objects, less opinions about what the ConnectionContext is. This diverges from what we do with HttpContext but it seems better overall.

Note: I don't know if we should just have a poco implementation of a ConnectionContext, it's basically what exists in SignalR (https://github.com/aspnet/SignalR/blob/290f18ac760af99e9402f21568f0d06261f8e69b/src/Microsoft.AspNetCore.Sockets.Abstractions/DefaultConnectionContext.cs) but I'm not sure if it should just implement all of the features that exists... It's basically used for unit testing and benchmarks. Maybe keeping the default makes sense but we make it more of a poco? I'm torn.

Update: Changed my mind, we now have a poco.


src/Connections.Abstractions/ConnectionMetadata.cs
```diff
@@ -3,18 +3,19 @@
...
     {
-        public ConnectionItems()
-            : this(new Dictionary<object, object>())
+        public ConnectionMetadata()
+            : this(new ConcurrentDictionary<object, object>())
         {
         }
 
-        public ConnectionItems(IDictionary<object, object> items)
+        public ConnectionMetadata(IDictionary<object, object> items)
```
anurse: Why the rename back? Didn't we want to use `Items` for consistency with `HttpContext`?

davidfowl: Yah, I can change it back. I just copied things.

davidfowl:
@dotnet-bot test this please

Issue #2357, opened 2018-03-02
Support SNI to enable multiple HTTPS hosts on single server


DamianEdwards:
There are three parts to this:
1) Work with corefx to define the SslStream API: dotnet/corefx#24553
2) Expose the new configuration in Kestrel's Https APIs.
3) Expose some form of this in Kestrel's config file format

@Tratcher @muratg

Tratcher:
See https://github.com/dotnet/corefx/issues/24553, https://github.com/aspnet/KestrelHttpServer/issues/2103

JoshCollinsMSFT:
Assuming dotnet/corefx#24553 makes it in the .NET Core 2.1 release, will the next release of KestrelHttpServer have support for SNI?

Tratcher:
That's the idea, will see how it goes.

Eilon:
@DamianEdwards / @shirhatti / @Tratcher - is this critical for Preview2?

Tratcher:
Yes. We're trying to take a new CoreFx build today to unblock it.

Eilon:
Got it, thanks!

Issue #2422, opened 2018-03-23
Delay loading the default development certificate


Tratcher:
Today Kestrel first tries to load the default development certificate and assign that to all new HttpsConnectionAdapterOptions.ServerCertificate. 

This shows up as a measurable % of startup time. It also causes log noise.

It also causes additional complexities with the new SNI support were the cert and the cert selector are mutually exclusive. This may need to get refactored as part of https://github.com/aspnet/KestrelHttpServer/issues/2357

muratg:
Costing?

davidfowl:
Once we have the SNI callback then we can use it here.

Tratcher:
No, this is later than we'd want, and invoked per connection. You want to load the cert as the server starts, but only if no other cert was provided.

davidfowl:
Disagree. You want to load the cert only if you’re using https.

Tratcher:
Yes, but inside the SNI callback is way too late. Do it when you set up HTTPS. One of these two places seems the most likely:
https://github.com/aspnet/KestrelHttpServer/blob/a37fa83aee400f621b1f19fca6a777bd3d804f6f/src/Kestrel.Core/ListenOptionsHttpsExtensions.cs#L214
https://github.com/aspnet/KestrelHttpServer/blob/a37fa83aee400f621b1f19fca6a777bd3d804f6f/src/Kestrel.Core/Internal/HttpsConnectionAdapter.cs#L39

PR #2425, opened 2018-03-23
Add SNI Support

Tratcher:
#2357 Here's what Kestrel's SNI support might look like.
This is based on the initial CoreFx design in https://github.com/dotnet/corefx/pull/28278.

### DIFF
--- a/samples/SampleApp/Startup.cs
+++ b/samples/SampleApp/Startup.cs
@@ -12,6 +12,7 @@ using Microsoft.AspNetCore.Builder;
 using Microsoft.AspNetCore.Hosting;
 using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.Server.Kestrel.Core;
+using Microsoft.AspNetCore.Server.Kestrel.Https.Internal;
 using Microsoft.AspNetCore.Server.Kestrel.Transport.Abstractions.Internal;
 using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.Logging;
@@ -105,10 +106,23 @@ namespace SampleApp
                         listenOptions.UseHttps(StoreName.My, "localhost", allowInvalid: true);
                     });
 
+                    options.ListenAnyIP(basePort + 5, listenOptions =>
+                    {
+                        listenOptions.UseHttps(httpsOptions =>
+                        {
+                            var localhostCert = CertificateLoader.LoadFromStoreCert("localhost", "My", StoreLocation.CurrentUser, allowInvalid: true);
+                            httpsOptions.ServerCertificateSelector = (features, name) =>
+                            {
+                                // Here you would check the name, select an appropriate cert, and provide a fallback or fail for null names.
+                                return localhostCert;
+                            };
+                        });
+                    });
+
                     options
                         .Configure()
-                        .Endpoint(IPAddress.Loopback, basePort + 5)
-                        .LocalhostEndpoint(basePort + 6)
+                        .Endpoint(IPAddress.Loopback, basePort + 6)
+                        .LocalhostEndpoint(basePort + 7)
                         .Load();
 
                     options
--- a/src/Kestrel.Core/HttpsConnectionAdapterOptions.cs
+++ b/src/Kestrel.Core/HttpsConnectionAdapterOptions.cs
@@ -6,6 +6,7 @@ using System.Net.Security;
 using System.Security.Authentication;
 using System.Security.Cryptography.X509Certificates;
 using System.Threading;
+using Microsoft.AspNetCore.Http.Features;
 using Microsoft.AspNetCore.Server.Kestrel.Core;
 
 namespace Microsoft.AspNetCore.Server.Kestrel.Https
@@ -29,13 +30,24 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Https
 
         /// <summary>
         /// <para>
-        /// Specifies the server certificate used to authenticate HTTPS connections.
+        /// Specifies the server certificate used to authenticate HTTPS connections. This is ignored if ServerCertificateSelector is set.
         /// </para>
         /// <para>
         /// If the server certificate has an Extended Key Usage extension, the usages must include Server Authentication (OID 1.3.6.1.5.5.7.3.1).
         /// </para>
         /// </summary>
         public X509Certificate2 ServerCertificate { get; set; }
+
+        /// <summary>
+        /// <para>
+        /// A callback that will be invoked to dynamically select a server certificate. This is higher priority than ServerCertificate.
+        /// If SNI is not avialable then the name parameter will be null.
+        /// </para>
+        /// <para>
+        /// If the server certificate has an Extended Key Usage extension, the usages must include Server Authentication (OID 1.3.6.1.5.5.7.3.1).
+        /// </para>
+        /// </summary>
+        public Func<IFeatureCollection, string, X509Certificate2> ServerCertificateSelector { get; set; }
 
         /// <summary>
         /// Specifies the client certificate requirements for a HTTPS connection. Defaults to <see cref="ClientCertificateMode.NoCertificate"/>.
--- a/src/Kestrel.Core/Internal/HttpsConnectionAdapter.cs
+++ b/src/Kestrel.Core/Internal/HttpsConnectionAdapter.cs
@@ -22,6 +22,8 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Https.Internal
 
         private readonly HttpsConnectionAdapterOptions _options;
         private readonly X509Certificate2 _serverCertificate;
+        private readonly Func<IFeatureCollection, string, X509Certificate2> _serverCertificateSelector;
+
         private readonly ILogger _logger;
 
         public HttpsConnectionAdapter(HttpsConnectionAdapterOptions options)
@@ -36,15 +38,24 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Https.Internal
                 throw new ArgumentNullException(nameof(options));
             }
 
-            if (options.ServerCertificate == null)
+            // capture the certificate now so it can't be switched after validation
+            _serverCertificate = options.ServerCertificate;
+            _serverCertificateSelector = options.ServerCertificateSelector;
+            if (_serverCertificate == null && _serverCertificateSelector == null)
             {
-                throw new ArgumentException(CoreStrings.ServiceCertificateRequired, nameof(options));
+                throw new ArgumentException(CoreStrings.ServerCertificateRequired, nameof(options));
             }
 
-            // capture the certificate now so it can be switched after validation
-            _serverCertificate = options.ServerCertificate;
-
-            EnsureCertificateIsAllowedForServerAuth(_serverCertificate);
+            // If a selector is provided then ignore the cert, it may be a default cert.
+            if (_serverCertificateSelector != null)
+            {
+                // SslStream doesn't allow both.
+                _serverCertificate = null;
+            }
+            else
+            {
+                EnsureCertificateIsAllowedForServerAuth(_serverCertificate);
+            }
 
             _options = options;
             _logger = loggerFactory?.CreateLogger(nameof(HttpsConnectionAdapter));
@@ -115,9 +126,26 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Https.Internal
             try
             {
 #if NETCOREAPP2_1
+                // Adapt to the SslStream signature
+                ServerCertificateSelectionCallback selector = null;
+                if (_serverCertificateSelector != null)
+                {
+                    selector = (sender, name) =>
+                    {
+                        context.Features.Set(sslStream);
+                        var cert = _serverCertificateSelector(context.Features, name);
+                        if (cert != null)
+                        {
+                            EnsureCertificateIsAllowedForServerAuth(cert);
+                        }
+                        return cert;
+                    };
+                }
+
                 var sslOptions = new SslServerAuthenticationOptions()
                 {
                     ServerCertificate = _serverCertificate,
+                    ServerCertificateSelectionCallback = selector,
                     ClientCertificateRequired = certificateRequired,
                     EnabledSslProtocols = _options.SslProtocols,
                     CertificateRevocationCheckMode = _options.CheckCertificateRevocation ? X509RevocationMode.Online : X509RevocationMode.NoCheck,
@@ -137,7 +165,17 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Https.Internal
 
                 await sslStream.AuthenticateAsServerAsync(sslOptions, CancellationToken.None);
 #else
-                await sslStream.AuthenticateAsServerAsync(_serverCertificate, certificateRequired,
+                var serverCert = _serverCertificate;
+                if (_serverCertificateSelector != null)
+                {
+                    context.Features.Set(sslStream);
+                    serverCert = _serverCertificateSelector(context.Features, null);                
+                    if (serverCert != null)
+                    {
+                        EnsureCertificateIsAllowedForServerAuth(serverCert);
+                    }
+                }
+                await sslStream.AuthenticateAsServerAsync(serverCert, certificateRequired,
                         _options.SslProtocols, _options.CheckCertificateRevocation);
 #endif
             }
--- a/build/dependencies.props
+++ b/build/dependencies.props
@@ -7,39 +7,39 @@
     <BenchmarkDotNetPackageVersion>0.10.13</BenchmarkDotNetPackageVersion>
     <InternalAspNetCoreSdkPackageVersion>2.1.0-preview2-15749</InternalAspNetCoreSdkPackageVersion>
     <LibuvPackageVersion>1.10.0</LibuvPackageVersion>
-    <MicrosoftAspNetCoreBenchmarkRunnerSourcesPackageVersion>2.1.0-a-preview2-mpr-16443</MicrosoftAspNetCoreBenchmarkRunnerSourcesPackageVersion>
-    <MicrosoftAspNetCoreCertificatesGenerationSourcesPackageVersion>2.1.0-a-preview2-mpr-16443</MicrosoftAspNetCoreCertificatesGenerationSourcesPackageVersion>
-    <MicrosoftAspNetCoreHostingAbstractionsPackageVersion>2.1.0-preview2-30478</MicrosoftAspNetCoreHostingAbstractionsPackageVersion>
-    <MicrosoftAspNetCoreHostingPackageVersion>2.1.0-preview2-30478</MicrosoftAspNetCoreHostingPackageVersion>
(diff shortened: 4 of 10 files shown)


shirhatti:
What's the behavior for legacy clients that don't support SNI? I'm assuming since we fallback to using the ServerCertificate if specified or close the connection?

Drawaes:
To be fair it's not "legacy clients" SNI isn't required and if you go via an IP there is no SNI to compare. 

The problem you have of course is if you are multi-homing you can't just use "certificate x" it might be the wrong website. So I think a default should be allowed but not required. It's a MAY include extension.

As an aside I wonder if the possible cipher list could be supplied if you are reading the client hello anyway. That way say a smaller and more secure ECDSA cert could be supplied and an RSA supplied if the client doesn't support it.

Tratcher:
@Drawaes you'll have to ask on dotnet/corefx#28278 about exposing additional parameters. Right now they only give us a name.

@shirhatti The current SslStream PR will invoke the callback with a null name if SNI was not supplied by the client. The developer is free to return a default certificate to continue, or return null, or throw to abort. ServerCertificate is always ignored if a selector is provided (SslStream makes them mutually exclusive).

src/Kestrel.Core/Internal/HttpsConnectionAdapter.cs
```diff
@@ -36,15 +38,34 @@ public HttpsConnectionAdapter(HttpsConnectionAdapterOptions options, ILoggerFact
...
+            {
+                // SslStream doesn't allow both.
+                _serverCertificate = null;
+                // Adapt to the SslStream signature
+                _serverCertificateSelector = (sender, name) =>
+                {
+                    var cert = selector(name);
+                    if (cert != null)
+                    {
+                        EnsureCertificateIsAllowedForServerAuth(cert);
```
Tratcher: @blowdart it's potentially quite expensive to validate EKUs this on every new connection. Can we exclude that for the advanced SNI scenario?

src/Kestrel.Core/HttpsConnectionAdapterOptions.cs
```diff
@@ -29,14 +29,26 @@ public HttpsConnectionAdapterOptions()
...
+        /// <summary>
+        /// A callback that will be invoked to dynamically select a server certificate. This is higher priority than ServerCertificate.
+        /// </summary>
+        public Func<string, X509Certificate2> ServerCertificateSelector
+#if NETCOREAPP2_1
+        { get; set; }
+#else
+        {
+            get => null;
+            set => throw new PlatformNotSupportedException("This API requires targeting netcoreapp2.1");
```
davidfowl: I'm not a huge fan of this throwing. If we made another callback API that let you control the cert per connection what would that look like?

Tratcher: Updated. Now a callback will be allowed even for 2.0. The difference is that for 2.0 we'll always invoke it with a null name before the handshake. The developer won't be able to distinguish between the client and server failing to provide SNI.

src/Kestrel.Core/Internal/HttpsConnectionAdapter.cs
blowdart: Cache the result based on cert thumbprint?

krwq:
If server name could not be extracted or is not supported the callback will pass null string

src/Kestrel.Core/Internal/HttpsConnectionAdapter.cs
Tratcher: @davidfowl what are the odds you'd allow registering and injecting a MemoryCache here? 😁

davidfowl: Kestrel.Core shouldn't depend on the memory cache.

src/Kestrel.Core/Internal/HttpsConnectionAdapter.cs
```diff
@@ -137,7 +159,12 @@ public Task<IAdaptedConnection> OnConnectionAsync(ConnectionAdapterContext conte
 
                 await sslStream.AuthenticateAsServerAsync(sslOptions, CancellationToken.None);
 #else
-                await sslStream.AuthenticateAsServerAsync(_serverCertificate, certificateRequired,
+                var serverCert = _serverCertificate;
+                if (_serverCertificateSelector != null)
+                {
+                    serverCert = _serverCertificateSelector(null, null);
```
halter73: Do you think we should null check serverCert to get better errors when ServerCertificateSelector is not null but its return value is?

src/Kestrel.Core/Internal/HttpsConnectionAdapter.cs
```diff
@@ -36,15 +38,34 @@ public HttpsConnectionAdapter(HttpsConnectionAdapterOptions options, ILoggerFact
...
             }
 
-            // capture the certificate now so it can be switched after validation
-            _serverCertificate = options.ServerCertificate;
-
-            EnsureCertificateIsAllowedForServerAuth(_serverCertificate);
+            // If a selector is provided then ignore the cert, it may be a default cert.
+            if (selector != null)
+            {
+                // SslStream doesn't allow both.
```
halter73: I'm thinking it should be up to the caller of the HttpsConnectionAdapter ctor to ensure either ServerCertificateSelector or ServerCertificate is set but not both. I think it's better to throw if both are set instead of committing to this behavior of preferring the selector.

halter73: In UseHttps, is ServerCertificate already set by the time the config callback is called?

src/Kestrel.Core/Internal/HttpsConnectionAdapter.cs
```diff
@@ -36,15 +38,34 @@ public HttpsConnectionAdapter(HttpsConnectionAdapterOptions options, ILoggerFact
                 throw new ArgumentNullException(nameof(options));
             }
 
-            if (options.ServerCertificate == null)
+            // capture the certificate now so it can't be switched after validation
+            _serverCertificate = options.ServerCertificate;
+            var selector = options.ServerCertificateSelector;
+            if (_serverCertificate == null && selector == null)
             {
-                throw new ArgumentException(CoreStrings.ServiceCertificateRequired, nameof(options));
```
halter73: lol the good 'ol service certificate.

src/Kestrel.Core/HttpsConnectionAdapterOptions.cs
```diff
@@ -29,14 +29,25 @@ public HttpsConnectionAdapterOptions()
...
+        /// <summary>
+        /// <para>
+        /// A callback that will be invoked to dynamically select a server certificate. This is higher priority than ServerCertificate.
+        /// If SNI is not avialable then the name parameter will be null.
+        /// </para>
+        /// <para>
+        /// If the server certificate has an Extended Key Usage extension, the usages must include Server Authentication (OID 1.3.6.1.5.5.7.3.1).
+        /// </para>
+        /// </summary>
+        public Func<string, X509Certificate2> ServerCertificateSelector { get; set; }
```
Tratcher: Consider passing through a state object, either a Kestrel specific object or the SslStream itself. That's what SslStream passes to their callback (as an Object).

halter73: So more of a "sender" than a state object? I don't think a state object is necessary since any closures would be allocated only once for the lifetime of the server.

Tratcher: Yes, a sender. Is there any connection specific state we think would be interesting to expose to this callback?

Drawaes: Why not just send the SslStream any properties are likely to be exposed on that in the future and means you won't have to change the kestrel side for every change.

Tratcher: SslStream is largely uninitialized at this stage in the handshake. See https://github.com/dotnet/corefx/pull/28278#discussion_r177871697

Drawaes: I agree that is the way it is today. But as we have client hello parsing started it's just a backdoor to other parts being parsed and populated ;) (don't tell anyone) and if these features are populated then there will be no lightup required in kestrel.

halter73: Maybe the ConnectionContext would be useful.

Tratcher: Updated to pass through the sender. Trying to pass though per-connection state requires allocating a per-connection closure. We'll follow up in RC to see if that makes sense.

src/Kestrel.Core/Internal/HttpsConnectionAdapter.cs
Tratcher: The default certs mess that up. See https://github.com/aspnet/KestrelHttpServer/issues/2422

src/Kestrel.Core/HttpsConnectionAdapterOptions.cs
halter73: Not that we have to do this now, but a per-connection closure allocation isn't a big deal in the context of initializing an HTTPS connection in Kestrel right now.

Tratcher: Yeah, I realized that looking at the client cert validation code. I added it back for now.

Tratcher:
Everything appears to be working in manual and automated tests. There are a few unrelated flaky test failures. @halter73 ready for a final review.

The last bit is figuring out what CoreFx build to target. @pakrym is waiting for the next prodcon build before updating universe.

davidfowl:
These aren't flaky tests, these look like regressions:

```
11:53:27 Log Critical[0]: Unexpected exception in HttpConnection.ProcessRequestsAsync. System.ObjectDisposedException: Cannot access a disposed object.
11:53:27 Object name: 'MemoryPoolBlock'.
11:53:27    at System.Buffers.MemoryPoolBlock.Dispose()
11:53:27    at System.IO.Pipelines.BufferSegment.ResetMemory()
11:53:27    at System.IO.Pipelines.Pipe.CompletePipe()
11:53:27    at System.IO.Pipelines.Pipe.CompleteWriter(Exception exception)
11:53:27    at System.IO.Pipelines.Pipe.DefaultPipeWriter.Complete(Exception exception)
11:53:27    at Microsoft.AspNetCore.Server.Kestrel.Core.Adapter.Internal.AdaptedPipeline.<ReadInputAsync>d__16.MoveNext()
11:53:27 --- End of stack trace from previous location where exception was thrown ---
11:53:27    at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
11:53:27    at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
11:53:27    at Microsoft.AspNetCore.Server.Kestrel.Core.Adapter.Internal.AdaptedPipeline.<RunAsync>d__14.MoveNext()
11:53:27 --- End of stack trace from previous location where exception was thrown ---
11:53:27    at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
11:53:27    at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
11:53:27    at System.Runtime.CompilerServices.TaskAwaiter.ValidateEnd(Task task)
11:53:27    at Microsoft.AspNetCore.Server.Kestrel.Core.Internal.HttpConnection.<ProcessRequestsAsync>d__47`1.MoveNext()
11:53:27 Log Critical[0]: ConnectionDispatcher.Execute() 0HLCPIGNNCR8B System.Exception: Unexpected critical error. Log Critical[0]: Unexpected exception in HttpConnection.ProcessRequestsAsync. System.ObjectDisposedException: Cannot access a disposed object.
11:53:27 Object name: 'MemoryPoolBlock'.
11:53:27    at System.Buffers.MemoryPoolBlock.Dispose()
11:53:27    at System.IO.Pipelines.BufferSegment.ResetMemory()
11:53:27    at System.IO.Pipelines.Pipe.CompletePipe()
11:53:27    at System.IO.Pipelines.Pipe.CompleteWriter(Exception exception)
11:53:27    at System.IO.Pipelines.Pipe.DefaultPipeWriter.Complete(Exception exception)
11:53:27    at Microsoft.AspNetCore.Server.Kestrel.Core.Adapter.Internal.AdaptedPipeline.<ReadInputAsync>d__16.MoveNext()
11:53:27 --- End of stack trace from previous location where exception was thrown ---
11:53:27    at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
11:53:27    at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
11:53:27    at Microsoft.AspNetCore.Server.Kestrel.Core.Adapter.Internal.AdaptedPipeline.<RunAsync>d__14.MoveNext()
11:53:27 --- End of stack trace from previous location where exception was thrown ---
11:53:27    at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
11:53:27    at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
11:53:27    at System.Runtime.CompilerServices.TaskAwaiter.ValidateEnd(Task task)
11:53:27    at Microsoft.AspNetCore.Server.Kestrel.Core.Internal.HttpConnection.<ProcessRequestsAsync>d__47`1.MoveNext() ---> System.ObjectDisposedException: Cannot access a disposed object.
11:53:27 Object name: 'MemoryPoolBlock'.
11:53:27    at System.Buffers.MemoryPoolBlock.Dispose()
11:53:27    at System.IO.Pipelines.BufferSegment.ResetMemory()
11:53:27    at System.IO.Pipelines.Pipe.CompletePipe()
11:53:27    at System.IO.Pipelines.Pipe.CompleteWriter(Exception exception)
11:53:27    at System.IO.Pipelines.Pipe.DefaultPipeWriter.Complete(Exception exception)
11:53:27    at Microsoft.AspNetCore.Server.Kestrel.Core.Adapter.Internal.AdaptedPipeline.<ReadInputAsync>d__16.MoveNext()
11:53:27 --- End of stack trace from previous location where exception was thrown ---
11:53:27    at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
11:53:27    at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
11:53:27    at Microsoft.AspNetCore.Server.Kestrel.Core.Adapter.Internal.AdaptedPipeline.<RunAsync>d__14.MoveNext()
11:53:27 --- End of stack trace from previous location where exception was thrown ---
11:53:27    at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
11:53:27    at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
11:53:27    at System.Runtime.CompilerServices.TaskAwaiter.ValidateEnd(Task task)
11:53:27    at Microsoft.AspNetCore.Server.Kestrel.Core.Internal.HttpConnection.<ProcessRequestsAsync>d__47`1.MoveNext()
11:53:27    --- End of inner exception stack trace ---
11:53:27    at Microsoft.AspNetCore.Testing.TestApplicationErrorLogger.Log[TState](LogLevel logLevel, EventId eventId, TState state, Exception exception, Func`3 formatter)
11:53:27    at Microsoft.AspNetCore.Server.Kestrel.Core.Internal.KestrelTrace.Log[TState](LogLevel logLevel, EventId eventId, TState state, Exception exception, Func`3 formatter)
11:53:27    at Microsoft.Extensions.Logging.LoggerExtensions.LogCritical(ILogger logger, EventId eventId, Exception exception, String message, Object[] args)
11:53:27    at Microsoft.AspNetCore.Server.Kestrel.Core.Internal.HttpConnection.<ProcessRequestsAsync>d__47`1.MoveNext()
11:53:27 --- End of stack trace from previous location where exception was thrown ---
11:53:27    at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
11:53:27    at 
```

Tratcher:
@pakrym this is the stuff you're updating right? Is this a matter of mismatched dependencies?

samples/SampleApp/Startup.cs
```diff
@@ -105,10 +106,23 @@ public static Task Main(string[] args)
...
                     });
 
+                    options.ListenAnyIP(basePort + 5, listenOptions =>
+                    {
+                        listenOptions.UseHttps(httpsOptions =>
+                        {
+                            var localhostCert = CertificateLoader.LoadFromStoreCert("localhost", "My", StoreLocation.CurrentUser, allowInvalid: true);
+                            httpsOptions.ServerCertificateSelector = (features, name) =>
+                            {
+                                // TODO: Name check, multiple certs, null names.
```
halter73: Is this something we actually want to do in this sample?

src/Kestrel.Core/HttpsConnectionAdapterOptions.cs
```diff
@@ -29,14 +30,25 @@ public HttpsConnectionAdapterOptions()
 
         /// <summary>
         /// <para>
-        /// Specifies the server certificate used to authenticate HTTPS connections.
+        /// Specifies the server certificate used to authenticate HTTPS connections. This is ignored if ServerCertificateSelector is set.
```
halter73: I wish we didn't have to set an order of precedence. Are there any alternatives?

samples/SampleApp/Startup.cs
```diff
@@ -105,10 +106,23 @@ public static Task Main(string[] args)
                         listenOptions.UseHttps(StoreName.My, "localhost", allowInvalid: true);
                     });
 
+                    options.ListenAnyIP(basePort + 5, listenOptions =>
+                    {
+                        listenOptions.UseHttps(httpsOptions =>
+                        {
+                            var localhostCert = CertificateLoader.LoadFromStoreCert("localhost", "My", StoreLocation.CurrentUser, allowInvalid: true);
+                            httpsOptions.ServerCertificateSelector = (features, name) =>
```
halter73: I'm not sure I like this just being a settable property. What about adding a RegisterCertificateSelector method that takes a state object as well?

Tratcher: Offline note: this is per endpoint, not per connection.

samples/SampleApp/Startup.cs
Tratcher: No plans to do this in the sample, only showing where to do it. I'll remove the TODO.

src/Kestrel.Core/HttpsConnectionAdapterOptions.cs
Tratcher: There are several layers that can assign default certs, so the best option for now is to ignore it if a selector is provided. See https://github.com/aspnet/KestrelHttpServer/issues/2422

PR #2426, opened 2018-03-23
Make LibuvTransport(Factory) pubinternal

halter73:
This enables the ApiCheck for the Libuv transport.

The ApiCheck was [previously disabled for the Libuv project](https://github.com/aspnet/KestrelHttpServer/pull/2418#issuecomment-375769339) because the already-pubternal IConnectionHandler got renamed IConnectionDispatcher last night, but it was required to invoke a method on the should-have-never-been-public LibuvTransportFactory type.

There were never any methods on LibuvTransportFactory that could be called *without* pubternal types, and LibuvTransport could never even be constructed without pubternal types.

@Eilon @davidfowl So I looked at this more deeply and I found that apparently what happened is there was a glitch in the baseline department, so I just went ahead and fixed the glitch.'

Seriously though, does this really require a breaking change announcement given neither type could ever be used without other pubternal types?


halter73:
I can think of one way this could break someone who didn't reference any pubternal types. If someone manually registered LibuvTransportFactory with DI, this *would* break them. 😢

Issue #2390, opened 2018-03-13
Implement file based logging accross the board


muratg:
We did this for SignalR -- where we log each test failure in a file. This makes it easier to debug and fix test cases, reducing the time to look into CI logs, etc.

muratg:
@JunTaoLuo talk to me and @anurse about more details.  This should help with the rest of the flaky test investigations.

muratg:
:tada:

PR #2433, opened 2018-03-27
Add logging for functional tests WIP

JunTaoLuo:
Addresses #2390. Need to fix long paths and invalid path chars on Windows.


test/Kestrel.FunctionalTests/BadHttpRequestTests.cs
```diff
@@ -57,7 +67,8 @@ public Task BadRequestWhenHeaderNameContainsNonASCIIOrNullCharacters(string head
             return TestBadRequest(
                 $"GET / HTTP/1.1\r\n{header}\r\n\r\n",
                 "400 Bad Request",
-                expectedExceptionMessage);
+                expectedExceptionMessage,
+                testName: $"{nameof(BadRequestWhenHeaderNameContainsNonASCIIOrNullCharacters)}_{header}_{expectedExceptionMessage}".RemoveIllegalFileChars());
```
Tratcher: You should also filter out control characters like those in this test.

test/shared/ListenOptionsExtensions.cs
```diff
@@ -0,0 +1,24 @@
...
+using Microsoft.AspNetCore.Server.Kestrel.Core;
+
+namespace Microsoft.AspNetCore.Testing
+{
+    public static class ListenOptionsExtensions
+    {
+        public static string GetListenOptionsDetails(this ListenOptions options)
+        {
+            var sb = new StringBuilder();
+            sb.Append(options.GetDisplayName());
```
Tratcher: Add semicolon delineators

src/Kestrel.Core/Properties/AssemblyInfo.cs
```diff
@@ -4,7 +4,7 @@
...
-[assembly: InternalsVisibleTo("Microsoft.AspNetCore.Server.Kestrel.Transport.Libuv.FunctionalTests, PublicKey=0024000004800000940000000602000000240000525341310004000001000100f33a29044fa9d740c9b3213a93e57c84b472c84e0b8a0e1ae48e67a9f8f6de9d5f7f3d52ac23e48ac51801f1dc950abe901da34d2a9e3baadb141a17c77ef3c565dd5ee5054b91cf63bb3c6ab83f72ab3aafe93d0fc3c2348b764fafb0b1c0733de51459aeab46580384bf9d74c4e28164b7cde247f891ba07891c9d872ad2bb")]
+[assembly: InternalsVisibleTo("Libuv.FunctionalTests, PublicKey=0024000004800000940000000602000000240000525341310004000001000100f33a29044fa9d740c9b3213a93e57c84b472c84e0b8a0e1ae48e67a9f8f6de9d5f7f3d52ac23e48ac51801f1dc950abe901da34d2a9e3baadb141a17c77ef3c565dd5ee5054b91cf63bb3c6ab83f72ab3aafe93d0fc3c2348b764fafb0b1c0733de51459aeab46580384bf9d74c4e28164b7cde247f891ba07891c9d872ad2bb")]
```
davidfowl: ?

JunTaoLuo: Shortened project names to maximize the length allowance for test log file names.

JunTaoLuo: All because of long paths woes on Windows.

halter73: Wouldn't it be better to add truncating logic to the file logger than to start changing test project names?

halter73: It seems like truncating logic will be much more reliable. I could easily see this turning into whac-a-mole.

README.md
```diff
@@ -17,6 +17,10 @@ This project is part of ASP.NET Core. You can find samples, documentation and ge
 
 To run a complete build on command line only, execute `build.cmd` or `build.sh` without arguments. See [developer documentation](https://github.com/aspnet/Home/wiki) for more details.
 
+## File logging for functional test
+
+Turn on file logging for Kestrel functional tests by specifying the environment variable ASPNETCORE_TEST_LOG_DIR to the log output directory.
```
halter73: Why not make this enabled by default? Are we going to make sure the environment variable get set on travis/appveryor/teamcity?

test/Kestrel.FunctionalTests/AddressRegistrationTests.cs
```diff
@@ -18,23 +19,26 @@
...
+using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Logging.Testing;
 using Microsoft.Extensions.Options;
 using Xunit;
 using Xunit.Abstractions;
 using Xunit.Sdk;
 
-namespace Microsoft.AspNetCore.Server.Kestrel.FunctionalTests
+namespace FunctionalTests
```
halter73: Let's not rename anything if it's just because of Windows path issues.

test/Kestrel.FunctionalTests/AddressRegistrationTests.cs
```diff
@@ -161,66 +165,74 @@ public AddressRegistrationTests(ITestOutputHelper output)
             await RegisterAddresses_StaticPort_Success("http://localhost", new[] { "http://localhost", "http://127.0.0.1", "http://[::1]" });
         }
 
-        private async Task RegisterAddresses_Success(string addressInput, string[] testUrls, int testPort = 0)
+        private async Task RegisterAddresses_Success(string addressInput, string[] testUrls, int testPort = 0, [CallerMemberName] string testName = "")
```
halter73: > [CallerMemberName]

Good idea.

README.md
JunTaoLuo: Yes we have to set it on all the build configurations. The file logger looks for the environment variable to write the logs to file. If we want this to be automatic, then we need to update the logger to choose a default location. File an issue to Logging if we want that?

test/Kestrel.FunctionalTests/AddressRegistrationTests.cs
muratg: :+1:

pakrym:
I don't think we should do logging this way, it's very time consuming to add, makes code more complex and harder to support.

I think extending xunit, generating code or even rewriting IL in test assemblies scale much better then a person putting same code in hundreds of methods.

If in the end we still decide to rewrite code manually we should use much less verbose syntax and turn

```
         public async Task BadRequestLogsAreNotHigherThanInformation()
         {
              using (StartLog(out var loggerFactory, TestConstants.DefaultFunctionalTestLogLevel))

```
into

```
       public async Task BadRequestLogsAreNotHigherThanInformation() => RunWithLog(async log =>{
       });

```

anurse:
Sounds good. This was always a cobbled-together hack to solve a few flaky tests. I started working on https://github.com/anurse/xunit.injection a while ago to help here but haven't had time to clean it up.

> rewriting IL in test assemblies

not so sure about that approach ;)

pakrym:
> not so sure about that approach ;)

JunTaoLuo:
Is now a good time to reinvent functional test logging? Since it seems pretty critical for all the flaky test work I'd imagine we can make do with what we have now and look into improvements later.

pakrym:
Are all the tests logging was added to flaky?

pakrym:
If we need it to fix flaky tests lets add to flaky tests but if we decide to add it to everything let's think a bit before wasting tons of time.

JunTaoLuo:
Spoke to @anurse and it seems like it's going to be hard to avoid per method modifications easily. You'll need to configure test names for theory based tests and you will need to flow the loggerFactory to where it needs to go. I think what we can do is to use the current logger infrastructure for now and start the logging of the functional tests to unblock fixing flaky tests. We can take a closer look into a better design in the coming weeks before rolling this out to more functional tests elsewhere.

JunTaoLuo:
This will need some rework

Issue #2406, opened 2018-03-19
Broken test: RemoteIPv6Address


ryanbrandenburg:
This test [failed](http://aspnetci/viewLog.html?buildId=422648&buildTypeId=XPlat_Windows_Win8_Universe) with the following error:
```
======= Failed test run #1 ==========
System.Net.Http.HttpRequestException : Response status code does not indicate success: 400 (Bad Request).
   at System.Net.Http.HttpResponseMessage.EnsureSuccessStatusCode()
   at Microsoft.AspNetCore.Server.Kestrel.FunctionalTests.RequestTests.TestRemoteIPAddress(String registerAddress, String requestAddress, String expectAddress) in C:\BuildAgent\work\33bdfc1cae7b2a38\modules\KestrelHttpServer\test\Kestrel.FunctionalTests\RequestTests.cs:line 1650
--- End of stack trace from previous location where exception was thrown ---
```

Future instances of this failure will be tracked [here](http://aspnetci/httpAuth/app/rest/builds?locator=tag:NoSuccess400,sinceDate:201710010T000000-0001&fields=build(startDate,buildTypeId,status,branchName,webUrl,tags)).

Tratcher:
Partner bug. It looks like the new SocketHttpHandler is missing brackets for the IPv6 host header. I'll get a repro and file a corefx issue.

kasper3:
@Tratcher, there is also a bug in zoneId that IPv6 parser recognize index to only be integer, however, it can be if_name: https://github.com/dotnet/corefx/issues/27529. Can that fix be addressed at the same time?

Tratcher:
@kasper3 that's unrelated, zoneIds are not supposed to be transmitted in the host header.

PR #2436, opened 2018-03-27
Skip broken test

ryanbrandenburg:


Tratcher:
Doesn't look flaky, looks like it has failed every run in dev and release since yesterday morning. Product bug?

Tratcher:
Merge it. We'll need to get a fix from CoreFx and that will take a few days.

ryanbrandenburg:
cref aspnet/KestrelHttpServer#2406

Tratcher:
@ryanbrandenburg this is also broken in release

PR #2444, opened 2018-03-29
Add IConnection to Connections.Abstractions

davidfowl:
The end goal for 2.1 will be to merge `IConnection` and `ConnectionContext`, in the short term, we want to move this to a shipping assembly so that we can delete sockets.abstractions.

### DIFF
--- a/src/Connections.Abstractions/IConnection.cs
+++ b/src/Connections.Abstractions/IConnection.cs
@@ -1,0 +1,17 @@
+using System.Collections.Generic;
+using System.IO.Pipelines;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Http.Features;
+
+namespace Microsoft.AspNetCore.Connections
+{
+    public interface IConnection
+    {
+        IDuplexPipe Transport { get; }
+        IFeatureCollection Features { get; }
+
+        Task StartAsync();
+        Task StartAsync(TransferFormat transferFormat);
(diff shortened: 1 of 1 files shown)


halter73:
Since we already looked at this together in person, ping me if you change anything that you want me to look at.

src/Connections.Abstractions/IConnection.cs
```diff
@@ -0,0 +1,17 @@
...
+
+namespace Microsoft.AspNetCore.Connections
+{
+    public interface IConnection
+    {
+        IDuplexPipe Transport { get; }
+        IFeatureCollection Features { get; }
+
+        Task StartAsync();
+        Task StartAsync(TransferFormat transferFormat);
```
pakrym: TransferFormat.Text in general purpose connection interface? :/

davidfowl: This is for preview2. See the comment above there are plans to not have this interface for rc1

anurse: And `StartAsync` is going to be handled on the client by a separate "Connection Factory" component which won't need to be in these abstractions, so the Http.Connections.Client can be the only one that has to care about TransferFormat (aside from the feature and enum _existing_ in the abstractions)

anurse:
Merging to chase https://github.com/aspnet/SignalR/pull/1777

PR #2446, opened 2018-03-30
Reaction to *Memory changes

pakrym:


src/Kestrel.Transport.Libuv/Internal/LibuvConnection.cs
```diff
@@ -118,6 +118,8 @@ private static void ReadCallback(UvStreamHandle handle, int status, object state
 
         private void OnRead(UvStreamHandle handle, int status)
         {
+            // Cleanup state from last OnAlloc. This is safe even if OnAlloc wasn't called.
+            _bufferHandle.Dispose();
```
davidfowl: Did we have a bug before?

pakrym: Yeah, FlushAsync wakes up the reader, it consumes the block and it get's disposed, only after that we unpin it.

Issue #2449, opened 2018-03-31
Make Kestrel work with alternative MemoryPool implementations


davidfowl:
Today we make some assumptions that the underlying block size will be 4K and swapping out with another pool (like the default array pool) makes the tests fail.

/cc @jkotas

jkotas:
cc @GrabYourPitchforks

stephentoub:
I assume nothing in Kestrel assumes the Memory is already pinned and always calls Retain/Pin when needed?

davidfowl:
> I assume nothing in Kestrel assumes the Memory is already pinned and always calls Retain/Pin when needed?

Correct. But I'm seeing tons of errors.  Up to ~200 errors.

davidfowl:
What's worse is that when I run the test that fail in isolation they pass 😄

davidfowl:
Oh wait, I think I know what it is.

davidfowl:
See https://github.com/aspnet/KestrelHttpServer/pull/2450

davidfowl:
This is done

PR #2450, opened 2018-04-01
 Fixed a parser bug found when trying out the array pool

davidfowl:
- When using the array pool (https://github.com/aspnet/KestrelHttpServer/issues/2449), we get terrible block density (@pakrym we need to look into this as SignalR is using this pool currently with pipelines) and as a result
the header parser was failing.
- This fixes the case where the parser needed to read ahead 2 bytes at the end
(which is unrealistic) at the beginning of every header parse to determine if it's at the end. Previously we were comparing the current index to the reader index is incorrect
since we may end up at the same index in another segment. Now we just check a boolean to determine if we did read ahead or not.

Here's what debugging revealed:

image

/cc @jkotas

### DIFF
--- a/test/Kestrel.Core.Tests/HttpParserTests.cs
+++ b/test/Kestrel.Core.Tests/HttpParserTests.cs
@@ -386,6 +386,35 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Tests
             Assert.Equal(buffer.End, examined);
         }
 
+        [Fact]
+        public void ParseHeadersWithGratuitouslySplitBuffers()
+        {
+            var parser = CreateParser(_nullTrace);
+            var buffer = ReadOnlySequenceFactory.CreateSegments(
+                Encoding.ASCII.GetBytes("Host:\r\nConnecti"),
+                Encoding.ASCII.GetBytes("on"),
+                Encoding.ASCII.GetBytes(":"),
+                Encoding.ASCII.GetBytes(" "),
+                Encoding.ASCII.GetBytes("k"),
+                Encoding.ASCII.GetBytes("e"),
+                Encoding.ASCII.GetBytes("e"),
+                Encoding.ASCII.GetBytes("p"),
+                Encoding.ASCII.GetBytes("-"),
+                Encoding.ASCII.GetBytes("a"),
+                Encoding.ASCII.GetBytes("l"),
+                Encoding.ASCII.GetBytes("i"),
+                Encoding.ASCII.GetBytes("v"),
+                Encoding.ASCII.GetBytes("e"),
+                Encoding.ASCII.GetBytes("\r"),
+                Encoding.ASCII.GetBytes("\n\r"),
+                Encoding.ASCII.GetBytes("\n"));
+
+            var requestHandler = new RequestHandler();
+            var result = parser.ParseHeaders(requestHandler, buffer, out var consumed, out var examined, out _);
+
+            Assert.True(result);
+        }
+
         private void VerifyHeader(
             string headerName,
             string rawHeaderValue,
--- a/test/Kestrel.Core.Tests/HttpParserTests.cs
+++ b/test/Kestrel.Core.Tests/HttpParserTests.cs
@@ -27,8 +27,8 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Tests
             string expectedMethod,
             string expectedRawTarget,
             string expectedRawPath,
-// This warns that theory methods should use all of their parameters,
-// but this method is using a shared data collection with Http1ConnectionTests.TakeStartLineSetsHttpProtocolProperties and others.
+            // This warns that theory methods should use all of their parameters,
+            // but this method is using a shared data collection with Http1ConnectionTests.TakeStartLineSetsHttpProtocolProperties and others.
 #pragma warning disable xUnit1026
             string expectedDecodedPath,
             string expectedQueryString,
@@ -390,24 +390,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Tests
         public void ParseHeadersWithGratuitouslySplitBuffers()
         {
             var parser = CreateParser(_nullTrace);
-            var buffer = ReadOnlySequenceFactory.CreateSegments(
-                Encoding.ASCII.GetBytes("Host:\r\nConnecti"),
-                Encoding.ASCII.GetBytes("on"),
-                Encoding.ASCII.GetBytes(":"),
-                Encoding.ASCII.GetBytes(" "),
-                Encoding.ASCII.GetBytes("k"),
-                Encoding.ASCII.GetBytes("e"),
-                Encoding.ASCII.GetBytes("e"),
-                Encoding.ASCII.GetBytes("p"),
-                Encoding.ASCII.GetBytes("-"),
-                Encoding.ASCII.GetBytes("a"),
-                Encoding.ASCII.GetBytes("l"),
-                Encoding.ASCII.GetBytes("i"),
-                Encoding.ASCII.GetBytes("v"),
-                Encoding.ASCII.GetBytes("e"),
-                Encoding.ASCII.GetBytes("\r"),
-                Encoding.ASCII.GetBytes("\n\r"),
-                Encoding.ASCII.GetBytes("\n"));
+            var buffer = BytePerSegmentTestSequenceFactory.Instance.CreateWithContent("Host:\r\nConnection: keep-alive\r\n\r\n");
 
             var requestHandler = new RequestHandler();
             var result = parser.ParseHeaders(requestHandler, buffer, out var consumed, out var examined, out _);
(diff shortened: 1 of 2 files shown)


test/Kestrel.Core.Tests/HttpParserTests.cs
```diff
@@ -386,6 +386,59 @@ public void ParseRequestLineSplitBufferWithoutNewLineDoesNotUpdateConsumed()
             Assert.Equal(buffer.End, examined);
         }
 
+        [Fact]
+        public void ParseHeadersWithGratuitouslySplitBuffers()
```
davidfowl: I need to rename these tests

benaadams: Is a good name :)

test/Kestrel.Core.Tests/HttpParserTests.cs
```diff
@@ -386,6 +386,59 @@ public void ParseRequestLineSplitBufferWithoutNewLineDoesNotUpdateConsumed()
...
+        {
+            var parser = CreateParser(_nullTrace);
+            var buffer = ReadOnlySequenceFactory.CreateSegments(
```
pakrym: Use https://github.com/aspnet/Common/blob/dev/shared/Microsoft.Extensions.Buffers.Testing.Sources/ReadOnlySequenceFactory.cs#L16 ?

pakrym:
> When using the array pool (#2449), we get terrible block density (@pakrym we need to look into this as SignalR is using this pool currently with pipelines) and as a result
the header parser was failing.

Set MinSegmentSize to 2*libuv request size

benaadams:
> Set MinSegmentSize to 2*libuv request size

Has to be non-power of 2; or 1 byte will cause second byte to allocate new block

test/Kestrel.Core.Tests/HttpParserTests.cs
davidfowl: I dunno how to use your crazy libraries 😄

pakrym:
That what happens right now, we allocate when block is more then half full.

pakrym:
That's why we should set minimum block size on the pipe to 4k, so in any pool you'll get 4k when requesting 2k.

davidfowl:
> Use segment factory

I don't like the fact that it puts an empty array blocks between each byte, the behavior might be slightly different. I claim false advertising.

davidfowl:
> Standard Output Messages:
 | [2018-03-31T23:39:14] Microsoft.AspNetCore.Hosting.Internal.WebHost Debug: Hosting starting
 | [2018-03-31T23:39:14] Microsoft.AspNetCore.Server.Kestrel.Core.KestrelServer Debug: Unable to locate an appropriate development https certificate.
 | [2018-03-31T23:39:14] Microsoft.AspNetCore.Hosting.Internal.WebHost Debug: Hosting started
 | [2018-03-31T23:39:14] Microsoft.AspNetCore.Hosting.Internal.WebHost Debug: Loaded hosting startup assembly Microsoft.AspNetCore.Server.Kestrel.Transport.Sockets.FunctionalTests
 | [2018-03-31T23:39:15] Microsoft.AspNetCore.Server.Kestrel Debug: Connection id "0HLCNJD4FFKR6" started.
 | [2018-03-31T23:39:16] LoggingConnectionAdapter Debug: ReadAsync[5] 16 03 03 00 AE 
 |                                                       \x16\x03\x03\0r
 | [2018-03-31T23:39:16] LoggingConnectionAdapter Debug: ReadAsync[174] 01 00 00 AA 03 03 5A C0 7E 93 DA 8E 21 72 08 27 C8 48 5A EE E5 4A 61 1D B0 8F 63 B8 14 88 65 43 1B 6A 85 01 AD 2F 00 00 38 C0 2C C0 2B C0 30 C0 2F 00 9F 00 9E C0 24 C0 23 C0 28 C0 27 C0 0A C0 09 C0 14 C0 13 00 39 00 33 00 9D 00 9C 00 3D 00 3C 00 35 00 2F 00 0A 00 6A 00 40 00 38 00 32 00 13 01 00 00 49 00 00 00 0E 00 0C 00 00 09 6C 6F 63 61 6C 68 6F 73 74 00 0A 00 08 00 06 00 1D 00 17 00 18 00 0B 00 02 01 00 00 0D 00 14 00 12 04 01 05 01 02 01 04 03 05 03 02 03 02 02 06 01 06 03 00 23 00 00 00 17 00 00 FF 01 00 01 00 
 |                                                       \x01\0\0ª\x03\x03ZA~?U?!r\x08'EHZîåJa\x1d°?c,\x14?eC\x1bj?\x01-/\0\08A,A+A0A/\0?\0?A$A#A(A'A
A	A\x14A\x13\09\03\0?\0?\0=\0<\05\0/\0
\0j\0@\08\02\0\x13\x01\0\0I\0\0\0\x0e\0\0\0	localhost\0
\0\x08\0\x06\0\x1d\0\x17\0\x18\0\0\x02\x01\0\0
\0\x14\0\x12\x04\x01\x05\x01\x02\x01\x04\x03\x05\x03\x02\x03\x02\x02\x06\x01\x06\x03\0#\0\0\0\x17\0\0ÿ\x01\0\x01\0
 | [2018-03-31T23:39:16] Microsoft.AspNetCore.Server.Kestrel Error: Uncaught exception from the OnConnectionAsync method of an IConnectionAdapter.
 |    at System.Net.SSPIWrapper.AcquireCredentialsHandle(SSPIInterface secModule, String package, CredentialUse intent, SCHANNEL_CRED scc)
 |    at System.Net.Security.SslStreamPal.AcquireCredentialsHandle(CredentialUse credUsage, SCHANNEL_CRED secureCredential)
 |    at System.Net.Security.SslStreamPal.AcquireCredentialsHandle(X509Certificate certificate, SslProtocols protocols, EncryptionPolicy policy, Boolean isServer)
 |    at System.Net.Security.SecureChannel.AcquireServerCredentials(Byte[]& thumbPrint)
 |    at System.Net.Security.SecureChannel.GenerateToken(Byte[] input, Int32 offset, Int32 count, Byte[]& output)
 |    at System.Net.Security.SecureChannel.NextMessage(Byte[] incoming, Int32 offset, Int32 count)
 |    at System.Net.Security.SslState.StartSendBlob(Byte[] incoming, Int32 count, AsyncProtocolRequest asyncRequest)
 |    at System.Net.Security.SslState.ProcessReceivedBlob(Byte[] buffer, Int32 count, AsyncProtocolRequest asyncRequest)
 |    at System.Net.Security.SslState.StartReadFrame(Byte[] buffer, Int32 readBytes, AsyncProtocolRequest asyncRequest)
 |    at System.Net.Security.SslState.StartReceiveBlob(Byte[] buffer, AsyncProtocolRequest asyncRequest)
 |    at System.Net.Security.SslState.ForceAuthentication(Boolean receiveFirst, Byte[] buffer, AsyncProtocolRequest asyncRequest)
 |    at System.Net.Security.SslState.ProcessAuthentication(LazyAsyncResult lazyResult)
 |    at System.Net.Security.SslStream.BeginAuthenticateAsServer(SslServerAuthenticationOptions sslServerAuthenticationOptions, CancellationToken cancellationToken, AsyncCallback asyncCallback, Object asyncState)
 |    at System.Net.Security.SslStream.<>c.<AuthenticateAsServerAsync>b__48_0(SslServerAuthenticationOptions arg1, CancellationToken arg2, AsyncCallback callback, Object state)
 |    at System.Threading.Tasks.TaskFactory`1.FromAsyncImpl[TArg1,TArg2](Func`5 beginMethod, Func`2 endFunction, Action`1 endAction, TArg1 arg1, TArg2 arg2, Object state, TaskCreationOptions creationOptions)
 |    at System.Threading.Tasks.TaskFactory.FromAsync[TArg1,TArg2](Func`5 beginMethod, Action`1 endMethod, TArg1 arg1, TArg2 arg2, Object state, TaskCreationOptions creationOptions)
 |    at System.Threading.Tasks.TaskFactory.FromAsync[TArg1,TArg2](Func`5 beginMethod, Action`1 endMethod, TArg1 arg1, TArg2 arg2, Object state)
 |    at System.Net.Security.SslStream.AuthenticateAsServerAsync(SslServerAuthenticationOptions sslServerAuthenticationOptions, CancellationToken cancellationToken)
 |    at Microsoft.AspNetCore.Server.Kestrel.Https.Internal.HttpsConnectionAdapter.InnerOnConnectionAsync(ConnectionAdapterContext context) in C:\projects\kestrelhttpserver\src\Kestrel.Core\Internal\HttpsConnectionAdapter.cs:line 138
 |    at Microsoft.AspNetCore.Server.Kestrel.Core.Internal.HttpConnection.ApplyConnectionAdaptersAsync() in C:\projects\kestrelhttpserver\src\Kestrel.Core\Internal\HttpConnection.cs:line 307
 | [2018-03-31T23:39:16] Microsoft.AspNetCore.Server.Kestrel Debug: Connection id "0HLCNJD4FFKR6" disconnecting.
 | [2018-03-31T23:39:17] Microsoft.AspNetCore.Server.Kestrel.Transport.Sockets Debug: Connection id "0HLCNJD4FFKR6" sending FIN.
 | [2018-03-31T23:39:17] Microsoft.AspNetCore.Server.Kestrel Debug: Connection id "0HLCNJD4FFKR6" stopped.
 | [2018-03-31T23:39:18] Microsoft.AspNetCore.Hosting.Internal.WebHost Debug: Hosting shutdown

pakrym:
> I don't like the fact that it puts an empty array blocks between each byte, the behavior might be slightly different. I claim false advertising.

Doesn't it repro with empty block?

davidfowl:
It probably does, but I think it only happens if we skip empty blocks in the BufferReader

benaadams:
Am sure the ROS skipped empty blocks at one time. Don't like zero byte writes; but like zero byte reads? 😆 

Though; realistically, zero byte segments shouldn't get into the ROS

PR #2452, opened 2018-04-01
Use 4K as the minimum segment size

davidfowl:
- This normalizes the behavior for kestrel no matter what memory pool implementation is used. The transports should behave the same (ask for 1/2 full blocks) across pool implementations.

Contributes to #2449

### DIFF
--- a/src/Kestrel.Transport.Sockets/Internal/SocketConnection.cs
+++ b/src/Kestrel.Transport.Sockets/Internal/SocketConnection.cs
@@ -18,8 +18,8 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Transport.Sockets.Internal
 {
     internal sealed class SocketConnection : TransportConnection
     {
-        private const int MinAllocBufferSize = 2048;
-        public readonly static bool IsWindows = RuntimeInformation.IsOSPlatform(OSPlatform.Windows);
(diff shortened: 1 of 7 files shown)


davidfowl:
@dotnet-bot build Ubuntu 16.04 Release Build

davidfowl:

pakrym:
Const

pakrym:
And use it in transports too

davidfowl:

pakrym:
Is there a way to add a test to assert that we are allocating segments in an expected manner?

src/Kestrel.Transport.Sockets/Internal/SocketConnection.cs
```diff
@@ -18,8 +18,8 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Transport.Sockets.Internal
 {
     internal sealed class SocketConnection : TransportConnection
     {
-        private const int MinAllocBufferSize = 2048;
-        public readonly static bool IsWindows = RuntimeInformation.IsOSPlatform(OSPlatform.Windows);
+        private static readonly int MinAllocBufferSize = KestrelMemoryPool.MinimumSegmentSize / 2;
```
halter73: Nit: Might as well make this same change to AdaptedPipeline.

davidfowl: Yep, missed that one, just updated it.

PR #2453, opened 2018-04-01
Remove bogus test

davidfowl:
The reason this works with the SlabMemoryPool is because the array is actually a giant slab and blocks are all backed by the same array. We're not really verifying anything other than the fact that they came from the same slab. In fact, we're still incurring a copy from transport pipe into the request body pipe so if a pool implementation wasn't slab based, this test could never pass. I'm not sure if we want to preserve anything from this test, but let me know.

Contributes to #2449

### DIFF
--- a/test/Kestrel.Core.Tests/MessageBodyTests.cs
+++ b/test/Kestrel.Core.Tests/MessageBodyTests.cs
@@ -421,65 +421,6 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Tests
             select new[] { stream[0], request[0], request[1] };
 
         [Theory]
-        [MemberData(nameof(RequestData))]
-        public async Task CopyToAsyncDoesNotCopyBlocks(HttpRequestHeaders headers, string[] data)
-        {
-            var writeCount = 0;
-            var writeTcs = new TaskCompletionSource<byte[]>();
-            var mockDestination = new Mock<Stream>() { CallBase = true };
-
-            mockDestination
-                .Setup(m => m.WriteAsync(It.IsAny<byte[]>(), It.IsAny<int>(), It.IsAny<int>(), CancellationToken.None))
-                .Callback((byte[] buffer, int offset, int count, CancellationToken cancellationToken) =>
-                {
-                    writeTcs.SetResult(buffer);
-                    writeCount++;
-                })
-                .Returns(Task.CompletedTask);
-
-            using (var input = new TestInput())
-            {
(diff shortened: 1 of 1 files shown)


davidfowl:

davidfowl:

halter73:
Also I see the point about the test being broken today because of the copy from the transport to the body pipe happening within the same array and not being caught. It was indeed an oversight this test wasn't changed when we first moved to having a request body pipe.

We could have caught the issue if we verified the offset and length instead of just verifying `Assert.Same(block.Array, await writeTcs.Task);`, so we should do that now. This test will pass without slabs once we change it to populate RequestBodyPipe directly.

davidfowl:
> We could have caught the issue if we verified the offset and length instead of just verifying Assert.Same(block.Array, await writeTcs.Task);, so we should do that now. This test will pass without slabs once we change it to populate RequestBodyPipe directly.

I'm confused, are you saying that this test is valuable? What aspects of it? Given that blocks *are* being copied even when this test claims they aren't, what value is there? Keeping it around for when we do fix the problem?

halter73:
Yes. I think there's value. Our benchmarks aren't always the easiest/best way to catch performance regressions. I opened #2458 as an alternative to this PR.

halter73:
Additionally, do we even have a benchmark that we run regularly and pay attention to that calls CopyToAsync() on the request body stream?

davidfowl:
> Additionally, do we even have a benchmark that we run regularly and pay attention to that calls CopyToAsync() on the request body stream?

If we don't you should add one 😉

davidfowl:
> Yes. I think there's value. Our benchmarks aren't always the easiest/best way to catch performance regressions. I opened #2458 as an alternative to this PR.

I'll close this out

halter73:
> If we don't you should add one 😉

I [know it exists](https://github.com/aspnet/benchmarks/blob/923956038060b18f02a8a5aa5089da29498f7379/src/Benchmarks/benchmarks.plaintext.json#L32-L35) the question was more of do we track the results?

halter73:
That won't get it in Power BI, right? I'm not sure the scenario is common enough to go on a graph though. There are too many graphs to keep track of as it is.

davidfowl:
> That won't get it in Power BI, right? I'm not sure the scenario is common enough to go on a graph though. There are too many graphs to keep track of as it is.

It doesn't need to be in power BI, it's tracked in the database and you can query that if you think it's important to track.

Issue #2439, opened 2018-03-28
Move Kestrel benchmarks to this repository


davidfowl:
Move the platform benchmark and plaintext to this repository. It'll make testing changes a bit easier.

/cc @halter73 @sebastienros

sebastienros:
Or maybe you need a custom benchmark application here too, dedicated to kestrel and the flags you need for local testing? And without having to bother with database, ...

davidfowl:
True, it doesn't need to replace anything per se but it would be nice to avoid duplication. We can use this application.

https://github.com/aspnet/KestrelHttpServer/blob/46ab09ca363f22d37dc326c16d757c81b89f1f40/samples/PlaintextApp/Startup.cs

At this point I wonder if we should just consider benchmarks to be the techempower application and each of the repos should just own their own benchmarks.

sebastienros:
Yes, I think so, we should have the TE scenarios in the app, and have other ones have custom logic. Like IIS, DB, Jil/Utf8/Json, then if we think they make sense for TE just port them over.

We can still add all these specific app results in the charts.

davidfowl:
Right, I would also argue that the middleware tests like response caching should be in their own repositories.

muratg:
@sebastienros I was told that you have a PR on this already.

unknowncloud:
go
	Go source files.
.c, .h
	C source files.
	If the package uses cgo or SWIG, these will be compiled with the
	OS-native compiler (typically gcc); otherwise they will
	trigger an error.
.cc, .cpp, .cxx, .hh, .hpp, .hxx
	C++ source files. Only useful with cgo or SWIG, and always
	compiled with the OS-native compiler.
.m
	Objective-C source files. Only useful with cgo, and always
	compiled with the OS-native compiler.
.s, .S
	Assembler source files.
	If the package uses cgo or SWIG, these will be assembled with the
	OS-native assembler (typically gcc (sic)); otherwise they
	will be assembled with the Go assembler.
.swig, .swigcxx
	SWIG definition files.
.syso
	System object files.

davidfowl:
@sebastienros Can you close this bug?

PR #2457, opened 2018-04-02
Moving PlatformBenchmarks from aspnet/benchmarks

sebastienros:


benchmarks/PlatformBenchmarks/PlatformBenchmarks.csproj
```diff
@@ -0,0 +1,27 @@
+<Project Sdk="Microsoft.NET.Sdk.Web">
+
+  <PropertyGroup>
```
davidfowl: This is problematic. It needs to match what we do in our repos.

sebastienros: Good, exactly what we need to talk about. Can't we just think of this app to be for the Benchmarks Server only, and just ensure it builds correctly with KoreBuild?

sebastienros: Looks like we might have to put a dependencies file dynamically that would match the KoreBuild one, probably by getting/building one from the meta packages for a specific ASP.NET version.

davidfowl: No we just need to the project to support overrides specified by the benchmarks driver.

sebastienros:
Why is it failing? https://ci3.dot.net/job/aspnet_KestrelHttpServer/job/dev/job/linux-Configuration_Release_prtest/48/console

halter73:
Looks like some sort of benchmark validation. I'm pretty sure @pakrym added this.

```
16:23:38 /home/dotnet-bot/.dotnet/buildtools/korebuild/2.1.0-preview3-17002/modules/vstest/module.targets(119,5): error : Run exited with a non-zero exit code: /mnt/j/workspace/aspnet_KestrelHttpServer/dev/linux-Configuration_Release_prtest/.dotnet/dotnet /mnt/j/workspace/aspnet_KestrelHttpServer/dev/linux-Configuration_Release_prtest/benchmarks/PlatformBenchmarks/bin/Release/netcoreapp2.1/PlatformBenchmarks.dll * --validate-fast [/home/dotnet-bot/.dotnet/buildtools/korebuild/2.1.0-preview3-17002/KoreBuild.proj]
16:23:38     0 Warning(s)
16:23:38     1 Error(s)
16:23:38 
16:23:38 Time Elapsed 00:05:59.77
16:23:38 [0;31merror  : dotnet failed with exit code 1[0m
```

pakrym:
We expect everything in `benchmarks` directory to be BenchmarkDotNet microbenchmark project. We also try to validate that all of them run that's why you build fails.

benchmarksapps/PlatformBenchmarks/PlatformBenchmarks.csproj
```diff
@@ -0,0 +1,27 @@
+﻿<Project Sdk="Microsoft.NET.Sdk.Web">
+
+  <PropertyGroup>
+    <TargetFramework>netcoreapp2.1</TargetFramework>
+    <OutputType>Exe</OutputType>
+    <LangVersion>latest</LangVersion>
+    <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
+    <RuntimeFrameworkVersion>$(MicrosoftNETCoreApp21PackageVersion)</RuntimeFrameworkVersion>
```
sebastienros: This should not be necessary, it's just to emphasize where it will only work with 2.1

halter73: Why isn't this compatible with 2.0?

sebastienros: Actually I have always thought it would only work on 2.1 ... if this is wrong then we should definitely remove this line.

davidfowl:
@BrennanConroy whats the deal here?

BrennanConroy:
First time I'm seeing this, what are you asking?

davidfowl:
Oops, wrong issue.

sebastienros:
All green, time to merge this PR

davidfowl:
@sebastienros whats up with the benchmarksapps/ folder? Didn't we agree benchmarks?

pakrym:
@davidfowl 

>We expect everything in benchmarks directory to be BenchmarkDotNet microbenchmark project. We also try to validate that all of them run

So having a web app in there breaks the build.

pakrym:
@davidfowl validation just tries to run the project with  `* --validate-fast` arguments, if run completes successfully nothing fails. It seems that this particular app fails to start with `* --validate-fast` and fails the build.
We should consider moving signalr tests apps out too.

pakrym:
I'll take a look at signalr repo.

pakrym:
Validation target has a bug, it only runs validation on the last project in the list. In SignalR case it happens to be microbenchmark project so it passes.
https://github.com/aspnet/BuildTools/blob/0e523ad57cf84e546a50ab15df544e6ded8688a4/files/KoreBuild/modules/vstest/module.targets#L119

sebastienros:
Fixes #2439

PR #2465, opened 2018-04-04
Delay loading the dev cert 

Tratcher:
#2422 Loading the dev cert is now delayed until after all possible user code that would set a cert.

### DIFF
--- a/src/Kestrel.Core/KestrelServerOptions.cs
+++ b/src/Kestrel.Core/KestrelServerOptions.cs
@@ -3,12 +3,17 @@
 
 using System;
 using System.Collections.Generic;
+using System.Linq;
 using System.Net;
 using System.Security.Cryptography.X509Certificates;
+using Microsoft.AspNetCore.Certificates.Generation;
 using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.Server.Kestrel.Https;
+using Microsoft.AspNetCore.Server.Kestrel.Internal;
 using Microsoft.AspNetCore.Server.Kestrel.Transport.Abstractions.Internal;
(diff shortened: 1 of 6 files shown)


halter73:
Why not just delay until the first request? I agree with what @davidfowl said in the issue. Why not just use the SNI callback to load the default cert? I like there being less special cases. It's not like you have to go to disk for every connection. You can cache the result.

src/Kestrel.Core/KestrelServerOptions.cs
```diff
@@ -105,10 +115,50 @@ public void ConfigureHttpsDefaults(Action<HttpsConnectionAdapterOptions> configu
 
         internal void ApplyHttpsDefaults(HttpsConnectionAdapterOptions httpsOptions)
         {
-            httpsOptions.ServerCertificate = DefaultCertificate;
             HttpsDefaults(httpsOptions);
         }
 
+        internal void ApplyDefaultCert(HttpsConnectionAdapterOptions httpsOptions)
```
halter73: Nit: Rename to TryApplyDefaultCert of EnsureDefaultCert.

Tratcher: I thought about that but you'd normally use Try when you were going to return the status and act differently. Here it just no-ops.

Tratcher:
Rebased onto the SNI changes and added a matching check to not load the dev cert if the SNI callback is given.

halter73:
Still LGTM after rebase.

src/Kestrel.Core/KestrelServerOptions.cs
halter73: EnsureServerCert it is then?

PR #2476, opened 2018-04-10
Make HTTP/2 internal

Tratcher:
This won't be finished for RTM so we'll make the option internal.


samples/Http2SampleApp/Program.cs
```diff
@@ -31,7 +29,8 @@ public static void Main(string[] args)
 
                     options.Listen(IPAddress.Any, basePort, listenOptions =>
                     {
-                        listenOptions.Protocols = HttpProtocols.Http1AndHttp2;
+                        // listenOptions.Protocols must be made public to run this sample.
```
halter73: Should we make InternalsVisibleTo to this sample like we do for our test projects?

src/Kestrel.Core/ListenOptions.cs
```diff
@@ -127,17 +123,12 @@ public FileHandleType HandleType
         /// <summary>
         /// The protocols enabled on this endpoint.
         /// </summary>
-        /// <remarks>Defaults to HTTP/1.x only. HTTP/2 support is experimental, see
-        /// https://go.microsoft.com/fwlink/?linkid=866785 to enable it.</remarks>
-        public HttpProtocols Protocols
```
halter73: Nit: might as well just make this an auto getter/setter for now.

samples/Http2SampleApp/Program.cs
Tratcher: I'd rather not. In theory this is a temporary change until we start 2.2.

Tratcher:
Updated.

unknowncloud:
System.IO;
using System.Net.Security

samples/Http2SampleApp/Program.cs
halter73: I'd rather just leave a comment saying this only works because of InternalsVisableTo. I don't like having to modify the sample just to try out an HTTP/2 endpoint.

Tratcher:
Updated.

PR #2478, opened 2018-04-11
Flow the ConnectionContext to the SNI callback

davidfowl:


unknowncloud:
James h

unknowncloud:
System.IO;
using System.Net.Security

PR #2479, opened 2018-04-11
Add logging for functional tests

JunTaoLuo:
Addresses https://github.com/aspnet/KestrelHttpServer/issues/2390


test/Kestrel.Transport.Sockets.FunctionalTests/Kestrel.Transport.Sockets.FunctionalTests.csproj
```diff
@@ -1,8 +1,8 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <AssemblyName>Microsoft.AspNetCore.Server.Kestrel.Transport.Sockets.FunctionalTests</AssemblyName>
-    <RootNamespace>Microsoft.AspNetCore.Server.Kestrel.Transport.Sockets.FunctionalTests</RootNamespace>
+    <AssemblyName>Sockets.FunctionalTests</AssemblyName>
```
halter73: No we know if the messes with TC test tracking?

README.md
```diff
@@ -17,6 +17,10 @@ This project is part of ASP.NET Core. You can find samples, documentation and ge
 
 To run a complete build on command line only, execute `build.cmd` or `build.sh` without arguments. See [developer documentation](https://github.com/aspnet/Home/wiki) for more details.
 
+## File logging for functional test
+
+Turn on file logging for Kestrel functional tests by specifying the environment variable ASPNETCORE_TEST_LOG_DIR to the log output directory.
```
halter73: Can we default to dropping logs in `{TestProjectDirectory}/testlogs` or something like that? We could still have an environment variable to opt-out if it's really needed.

JunTaoLuo: Hmm I need to think about where this default setting should live. Might need to make this change in Logging.

test/Kestrel.Transport.Sockets.FunctionalTests/Kestrel.Transport.Sockets.FunctionalTests.csproj
JunTaoLuo: So @ryanbrandenburg and I are not completely sure about how test history are tracked but it seems to be the full name of the method which isn't affected by the assembly name. However, Ryan was okay with the test history being lost.

Tratcher: Why change just the assembly name and not the project or directory name?

halter73: Well *I'm* not OK with the history being lost if there's any reasonable alternative.

JunTaoLuo: @Tratcher only the assembly name is used to construct the log directory at {root}/artifacts/logs/... the logs are not stored in the test directory. @halter73 We believe that with this change, the history will not be lost, but we are not 100% sure.

halter73: Shouldn't you be able to test this with a feature branch?

JunTaoLuo: Ah didn't know feature branches ran tests. For some reason I thought they only ran build.

JunTaoLuo: Yup confirmed that test history is maintained

JunTaoLuo:
@dotnet-bot test OSX 10.12 Release Build please

README.md
JunTaoLuo: There's a couple of ways to do this so let's talk about what we really want here. But we don't need to necessarily wait to figure this out for this PR and we can do this separately afterwards.

JunTaoLuo:
Size of logs 45.9 MB
Build time without logging: 651.2450628s
Build time with logging: 700.1792463s

JunTaoLuo:
This change is ready to go and all tests are passing (flakily).

test/Kestrel.FunctionalTests/AddressRegistrationTests.cs
```diff
@@ -484,14 +470,16 @@ public void ThrowsWhenBindingToIPv4AddressInUse()
         [IPv6SupportedCondition]
         public void ThrowsWhenBindingToIPv6AddressInUse()
         {
+            TestApplicationErrorLogger.ThrowOnCriticalErrors = false;
```
davidfowl: ?

test/Kestrel.FunctionalTests/AddressRegistrationTests.cs
```diff
@@ -630,7 +616,10 @@ public void ThrowsWhenBindingLocalhostToIPv6AddressInUse()
         [Fact]
         public void ThrowsWhenBindingLocalhostToDynamicPort()
         {
+            TestApplicationErrorLogger.ThrowOnCriticalErrors = false;
```
pakrym: Consider a delegate to ignore "know" exceptions.

test/Kestrel.FunctionalTests/AddressRegistrationTests.cs
JunTaoLuo: We inject a TestApplicationErrorLogger by default and it wraps invalid operation exceptions with an aggregate exception. These tests require the invalid operation exceptions so I'm disabling the throwing behaviour.

Tratcher: This test is expecting an error log, and the test logger throws by default if you log an error.

Issue #1853, opened 2017-05-23
Authentication failed because the remote party has closed the transport stream


danroth27:
When I configure HTTPS and browse to my app I see this error logged when I close the browser:

```
      Failed to authenticate HTTPS connection.
System.IO.IOException: Authentication failed because the remote party has closed the transport stream.
   at System.Net.Security.SslState.StartReadFrame(Byte[] buffer, Int32 readBytes, AsyncProtocolRequest asyncRequest)
   at System.Net.Security.SslState.PartialFrameCallback(AsyncProtocolRequest asyncRequest)
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Net.Security.SslState.InternalEndProcessAuthentication(LazyAsyncResult lazyResult)
   at System.Net.Security.SslState.EndProcessAuthentication(IAsyncResult result)
   at System.Net.Security.SslStream.EndAuthenticateAsServer(IAsyncResult asyncResult)
   at System.Threading.Tasks.TaskFactory`1.FromAsyncCoreLogic(IAsyncResult iar, Func`2 endFunction, Action`1 endAction, Task`1 promise, Boolean requiresSynchronization)
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at Microsoft.AspNetCore.Server.Kestrel.Https.HttpsConnectionAdapter.<OnConnectionAsync>d__7.MoveNext()
```

```
~\Desktop\NewApp> dotnet --info
.NET Command Line Tools (2.0.0-preview1-005977)

Product Information:
 Version:            2.0.0-preview1-005977
 Commit SHA-1 hash:  414cab8a0b

Runtime Environment:
 OS Name:     Windows
 OS Version:  10.0.15063
 OS Platform: Windows
 RID:         win10-x64
 Base Path:   C:\Program Files\dotnet\sdk\2.0.0-preview1-005977\

Microsoft .NET Core Shared Framework Host

  Version  : 2.0.0-preview1-002111-00
  Build    : 1ff021936263d492539399688f46fd3827169983
```

halter73:
[...]

PR #2481, opened 2018-04-11
Https handshake read-ahead to supress errors

Tratcher:
#1853

Too grose?

Plan B, downgrade the original error to Debug.

### DIFF
--- a/src/Kestrel.Core/Internal/HttpsConnectionAdapter.cs
+++ b/src/Kestrel.Core/Internal/HttpsConnectionAdapter.cs
@@ -71,6 +71,30 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Https.Internal
 
         private async Task<IAdaptedConnection> InnerOnConnectionAsync(ConnectionAdapterContext context)
         {
+            var timeoutFeature = context.Features.Get<IConnectionTimeoutFeature>();
+            timeoutFeature.SetTimeout(_options.HandshakeTimeout);
+
+            var prefix = new byte[1]; // TODO: Consider buffering the size of an expected TLS frame.
+            var read = 0;
+            try
+            {
+                // Wait for data to become available.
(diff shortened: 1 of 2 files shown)


src/Kestrel.Core/Internal/HttpsConnectionAdapter.cs
```diff
@@ -253,5 +274,84 @@ public void Dispose()
...
+
+            private int ReadPrefix(Span<byte> destination)
+            {
+                if (destination.Length == 0)
+                {
+                    // Data is available
+                    return 0;
+                }
+
+                var read = Math.Min(destination.Length, _prefix.Length);
```
halter73: Since `destination.Length != 0` and `_prefix.Length == 1` this isn't necessary.

src/Kestrel.Core/Internal/HttpsConnectionAdapter.cs
```diff
@@ -253,5 +274,84 @@ public void Dispose()
...
+                var source = _prefix.Span.Slice(0, read);
+                source.CopyTo(destination);
+                _prefix = _prefix.Slice(read);
+                if (_prefix.IsEmpty)
```
halter73: This is already checked to be false before calling ReadPrefix, right?

src/Kestrel.Core/Internal/HttpsConnectionAdapter.cs
halter73: Nvm. future proofing I see. I agree we probably wan't to try to get a full TCP frame on the first read if we do this.

halter73:
This isn't an API change or anything, so I think I could live with this for now. Once we use a pipeline-based API for connection adapters/middleware, this logic should become *much* cleaner.

Tratcher:
Abandond. I'll lower the log level to Debug instead.

Issue #2331, opened 2018-02-17
Unix socket fails with Unexpected exception in FrameConnection.ProcessRequestsAsync


zpodlovics:
Hi,

I have tried to use Kestrel with unix socket behind Nginx, unfortunately after a successful startup the first unix socket connection will crash it. The same problem are reproducable with curl. The same test url on loopback connection works fine. The example project are attached ( [kestrelunixsocketnre.zip](https://github.com/aspnet/KestrelHttpServer/files/1733823/kestrelunixsocketnre.zip) ).

Steps to reproduce:

1) ```cd /tmp```
2) ```unzip path/to/kestrelunixsocketnre.zip ```
3) ```cd kestrelunixsocketnre```
4) ```dotnet run```
5) ```curl --no-buffer -XGET --unix-socket /tmp/kestrel-test.sock http://localhost/test```
( in an other console, please note this command will hang)

6) press ctrl+c to curl

Expected result:
works without error

Actual result:
```
Hosting environment: Production
Content root path: /tmp/kestrelunixsocketnre
Now listening on: http://127.0.0.1:5000
Now listening on: http://unix:/tmp/kestrel-test.sock
Application started. Press Ctrl+C to shut down.
fail: Microsoft.AspNetCore.Server.Kestrel[0]
      Unexpected exception in FrameConnection.ProcessRequestsAsync.
System.NullReferenceException: Object reference not set to an instance of an object.
   at Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Infrastructure.KestrelEventSource.ConnectionStart(IConnectionContext context, IConnectionInformation information)
   at Microsoft.AspNetCore.Server.Kestrel.Core.Internal.FrameConnection.<ProcessRequestsAsync>d__42`1.MoveNext()
fail: Microsoft.AspNetCore.Server.Kestrel.Transport.Libuv[0]
      UbReadCb
System.NullReferenceException: Object reference not set to an instance of an object.
   at Microsoft.AspNetCore.Server.Kestrel.Core.Internal.FrameConnection.Abort(Exception ex)
   at Microsoft.AspNetCore.Server.Kestrel.Transport.Libuv.Internal.LibuvConnection.OnRead(UvStreamHandle handle, Int32 status)
   at Microsoft.AspNetCore.Server.Kestrel.Transport.Libuv.Internal.LibuvConnection.<>c.<.cctor>b__33_0(UvStreamHandle handle, Int32 status, Object state)
[...]

PR #2483, opened 2018-04-11
 Avoid nullref in KestrelEventSource

halter73:
#2331


halter73:
This sped up the travis macOS runs from 30 to 15 minutes 😮

PR #2492, opened 2018-04-14
Make logging on by default

JunTaoLuo:
And other minor cleanup

Followup for #2479.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,10 +17,6 @@ This project is part of ASP.NET Core. You can find samples, documentation and ge
 
 To run a complete build on command line only, execute `build.cmd` or `build.sh` without arguments. See [developer documentation](https://github.com/aspnet/Home/wiki) for more details.
 
-## File logging for functional test
-
-Turn on file logging for Kestrel functional tests by specifying the environment variable ASPNETCORE_TEST_LOG_DIR to the log output directory.
-
 ## Packages
 
 Kestrel is available as a NuGet package.
--- a/build/repo.targets
+++ b/build/repo.targets
@@ -13,5 +13,8 @@
     </ItemGroup>
     <MSBuild Projects="$(MSBuildProjectFullPath)" Targets="TestProjects" Properties="FakeProperty=%(Iteration.Identity)" />
   </Target>
-</Project>
 
+  <Target Name="EnableTestLogging" BeforeTargets="_ExecuteTestAssemblies" Condition="'$(ASPNETCORE_TEST_LOG_DIR)' == '' AND '$(DisableFTestLogging)' != 'true'">
+    <SetEnvironmentVariable Variable="ASPNETCORE_TEST_LOG_DIR" Value="$(MSBuildThisFileDirectory)..\artifacts\logs" />
+  </Target>
+</Project>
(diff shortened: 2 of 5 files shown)


build/repo.targets
```diff
@@ -13,5 +13,8 @@
     </ItemGroup>
     <MSBuild Projects="$(MSBuildProjectFullPath)" Targets="TestProjects" Properties="FakeProperty=%(Iteration.Identity)" />
   </Target>
-</Project>
 
+  <Target Name="EnableTestLogging" BeforeTargets="_ExecuteTestAssemblies" Condition="'$(ASPNETCORE_TEST_LOG_DIR)' == '' AND '$(DisableFTestLogging)' != 'true'">
```
JunTaoLuo: Looks like test invocation works differently for ./build.ps1 and dotnet test. The workaround is to add the custom target that executes before _ExecuteTestAssemblies and VSTest.

natemcmaster:
Is there any reason we shouldn't put this in KoreBuild itself? It's used in multiple places now, if I'm not mistaken.

build/repo.targets
natemcmaster: 😬 this is like dependency on pubternal API. The underscore convention in MSBuild is meant to indicate we make no promises about removing or changing this target in the future.

JunTaoLuo:
I am trying to turn logging on by default here since this was the first time this was asked for but I don't mind putting it in KoreBuild. Let's figure out a good mechanism for this first because I don't really like the current way it's implemented.

build/repo.targets
JunTaoLuo: Right but I couldn't find a better target.

README.md
```diff
@@ -17,10 +17,6 @@ This project is part of ASP.NET Core. You can find samples, documentation and ge
 
 To run a complete build on command line only, execute `build.cmd` or `build.sh` without arguments. See [developer documentation](https://github.com/aspnet/Home/wiki) for more details.
 
-## File logging for functional test
```
halter73: Maybe we should include how to turn file logging off in this README.

test/shared/TestApplicationErrorLogger.cs
```diff
@@ -16,7 +16,7 @@ public class TestApplicationErrorLogger : ILogger
         // Application errors are logged using 13 as the eventId.
         private const int ApplicationErrorEventId = 13;
 
-        public List<Type> IgnoredExceptions { get; } = new List<Type>();
+        public HashSet<Type> IgnoredExceptions { get; } = new HashSet<Type>();
```
halter73: Is this just for perf? It looks like this collection is only modified at the very start of the test before anything else could even try to read it.

JunTaoLuo: Because @pakrym mentioned it

natemcmaster:
My concern remains the same. This relies on 'internal' API in KoreBuild. We have tentative plans to change this part of code in 2.2.0-mq, and this may break. We make no promised about targets or properties that are "internal" i.e. things that start with an underscore.

halter73:
I hate seeing a test failed only to realize that I didn't set the environment variable. I could put it in my profile, but that doesn't help when I'm ssh'd into a CI agent for example.

@natemcmaster I don't think this is the first 'internal' API from one aspnet repo being called from another. Couldn't we just update Kestrel when if and when we make the 2.2.0-mq changes in KoreBuild?

halter73:
After talking to @natemcmaster, it seems this only works if you run the tests as part of a build and not if you run from VS or "dotnet test". That's not ideal and this PR is pretty old, so I'm closing.

PR #2496, opened 2018-04-16
make IConnectionInherentKeepAliveFeature a boolean feature

anurse:
Not used in Kestrel, but it's used by SignalR. 

We don't ever use the `TimeSpan`, and instead use the presence of the feature. This means we have to actually create a class implementing this and put it in the feature collection. We want to change to implement this interface directly on the ConnectionContext. See https://github.com/aspnet/SignalR/issues/1976

SignalR will treat the absence of this feature as equivalent to `HasInherentKeepAlive` being `false`.

### DIFF
--- a/src/Connections.Abstractions/Features/IConnectionInherentKeepAliveFeature.cs
+++ b/src/Connections.Abstractions/Features/IConnectionInherentKeepAliveFeature.cs
@@ -19,6 +19,6 @@ namespace Microsoft.AspNetCore.Connections.Features
     /// </remarks>
     public interface IConnectionInherentKeepAliveFeature
     {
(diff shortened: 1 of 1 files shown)


davidfowl:
We need to get this in before branching. So prep the change in SignalR as soon as the package is here.

anurse:
```
Still waiting to schedule task
All nodes of label ‘osx-10.12||OSX.1012.Amd64.Open’ are offline
```

Are we actually "using" dotnet CI in this repo?

anurse:
@pranavkm says the OSX delay is probably spurious. I'm gonna merge and watch UpdateUniverse like a hawk :)

Issue #2225, opened 2017-12-19
Flaky test: MaxRequestBufferSizeTests.LargeUpload


Tratcher:
This has failed 50% of the time for me on recent local builds, always with the same actual result.
```
[xUnit.net 00:00:44.2552397]     LargeUpload(maxRequestBufferSize: 5242880, connectionAdapter: True, expectPause: True) [FAIL]
[xUnit.net 00:00:44.2588137]       Assert.InRange() Failure
[xUnit.net 00:00:44.2591272]       Range:  (5238785 - 20971519)
[xUnit.net 00:00:44.2592656]       Actual: 20971520
[xUnit.net 00:00:44.2619018]       Stack Trace:
[xUnit.net 00:00:44.2652259]         D:\github\AspNet\KestrelHttpServer\test\Kestrel.FunctionalTests\MaxRequestBufferSizeTests.cs(160,0): at Microsoft.AspNetCore.Server.Kestrel.FunctionalTests.MaxRequestBufferSizeTests.<LargeUpload>d__6.MoveNext()
[xUnit.net 00:00:44.2657057]         --- End of stack trace from previous location where exception was thrown ---
[xUnit.net 00:00:44.2658986]         --- End of stack trace from previous location where exception was thrown ---
[xUnit.net 00:00:44.2660844]         --- End of stack trace from previous location where exception was thrown ---
[xUnit.net 00:00:44.2682372]       Output:
[xUnit.net 00:00:44.2684638]         | Microsoft.AspNetCore.Hosting.Internal.WebHost Information: Request starting HTTP/1.0 POST http:///  20971520
Failed   LargeUpload(maxRequestBufferSize: 5242880, connectionAdapter: True, expectPause: True)
Error Message:
 Assert.InRange() Failure
Range:  (5238785 - 20971519)
Actual: 20971520
Stack Trace:
   at Microsoft.AspNetCore.Server.Kestrel.FunctionalTests.MaxRequestBufferSizeTests.<LargeUpload>d__6.MoveNext() in D:\github\AspNet\KestrelHttpServer\test\Kestrel.FunctionalTests\MaxRequestBufferSizeTests.cs:line 160
```

Tratcher:
Still failing, same error.

muratg:
cc @halter73

halter73:
@Tratcher What OS? Have you noticed it fail on the 2.0 or 2.1 runtime? Both?

ryanbrandenburg:
[...]

PR #2497, opened 2018-04-16
Clean up flaky MaxRequestBufferSize tests

Tratcher:
I was debugging #2225 and had a reliable repro before John's logging changes. After his changes I could only cause OOMs. This change addresses the OOMs and the test now reliably passes.

I also removed the data corruption validation, there's a seperate test for that. These tests are much faster without it.


halter73:
The separate request corruption test that @Tratcher referenced is [here.](https://github.com/aspnet/KestrelHttpServer/blob/6fd09af3742ea07c63894ec512cafabbca1a5d68/test/Kestrel.FunctionalTests/RequestTests.cs#L88)

Tratcher:
Blocked while branching.

PR #2514, opened 2018-04-22
[WIP] IHttpSendFileFeature (Windows)

benaadams:
For comment; need clean up - follow up to https://github.com/aspnet/KestrelHttpServer/pull/2049 but with much better IO

### DIFF
--- a/src/Kestrel.Core/Internal/Http/SendFile.Windows.cs
+++ b/src/Kestrel.Core/Internal/Http/SendFile.Windows.cs
@@ -2,8 +2,10 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.Diagnostics;
 using System.IO;
 using System.IO.Pipelines;
+using System.Runtime.CompilerServices;
 using System.Runtime.InteropServices;
 using System.Text;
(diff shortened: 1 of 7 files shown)


src/Kestrel.Core/Internal/Http/SendFile.Windows.cs
```diff
@@ -0,0 +1,434 @@
...
+        [DllImport("Kernel32.dll", SetLastError = true, CharSet = CharSet.Unicode)]
+        static extern SafeFileHandle CreateFile(
+            string fileName,
+            [MarshalAs(UnmanagedType.U4)] FileAccess fileAccess,
+            [MarshalAs(UnmanagedType.U4)] FileShare fileShare,
+            IntPtr securityAttributes,
+            [MarshalAs(UnmanagedType.U4)] FileMode creationDisposition,
+            [MarshalAs(UnmanagedType.U4)] FileAttributes flags,
+            IntPtr template
+        );
```
halter73: Why make this windows-only for just these three P/Invokes? Is it really that much faster than the System.IO equivalents?

muratg:
We’re closing this PR because we don’t feel that this change is a good fit for the product at this time. 

PR #2520, opened 2018-04-23
WIP: Convert non-serializable data to dictionary lookup

mikeharder:
@halter73: Do you agree with this pattern before I convert the remaining tests?


halter73:
LGTM 👍

PR #2531, opened 2018-04-30
Use TaskCreationOptions.RunContinuationsAsynchronously a lot

pakrym:
Carpet bombing with `TaskCreationOptions.RunContinuationsAsynchronously`


Tratcher:
To what end?

pakrym:
The common issue I discovered in our tests is completing TCS inside the appfunc and then awaiting it just before disposing the server.
This causes continuation to be run inline in appfunc and prevents graceful shutdown making tests slower and possibly more flaky.
There is also a race it that if TCS get's completed before being awaited we get different behaviour then if was awaited first and then completed.

PR #2535, opened 2018-05-01
Decouple connection objects from the server

davidfowl:
- Making progress towards being able to use the connection objects on the client side.
- Move the call to OnConnected outside of the TransportConnection derived classes


davidfowl:

src/Kestrel.Transport.Libuv/Internal/Listener.cs
```diff
@@ -181,7 +181,10 @@ private void OnConnection(UvStreamHandle listenSocket, int status)
 
         protected virtual void DispatchConnection(UvStreamHandle socket)
         {
-            var connection = new LibuvConnection(this, socket);
+            var connection = new LibuvConnection(socket, Log, Thread);
+
+            TransportContext.ConnectionDispatcher.OnConnection(connection);
```
halter73: How is this going to flow once OnConnection starts returning a lifetime Task?

halter73: I guess the transport doesn't need a task for the connection lifetime. I guess just rename to AcceptConnection or something like that.

Also, make sure we're confident that On/AcceptConnection doesn't throw since you moved all the calls out of a try catch.

Issue #2517, opened 2018-04-23
Microsoft.AspNetCore.Server.Kestrel.FunctionalTests.ResponseTests.IgnoreNullHeaderValues


Tratcher:
9/10 of the theories for this test fail sporadically.

The failures are limited to MacOS & High Sierra.

This started on 4/13 and is still failing today on 4/23.

There were no test changes then, but there was a product change https://github.com/aspnet/KestrelHttpServer/commit/e30a02cee5ff4919a3be82bf53bab6bb5146e7de that suspicously related.

```
System.Threading.Tasks.TaskCanceledException : A task was canceled.
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at System.Runtime.CompilerServices.ConfiguredTaskAwaitable`1.ConfiguredTaskAwaiter.GetResult()
   at System.Net.Http.HttpClient.<FinishSendAsyncBuffered>d__58.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at Microsoft.AspNetCore.Server.Kestrel.FunctionalTests.ResponseTests.<IgnoreNullHeaderValues>d__3.MoveNext() in /_/test/Kestrel.FunctionalTests/ResponseTests.cs:line 126
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
[...]

PR #2537, opened 2018-05-02
Remove netcoreapp2.0 test coverage

Tratcher:
#2517 A series of tests have been consistently failing on netcoreapp2.0 on Mac due to an HttpClient issue. We're already planning to remove netcoreapp2.0 test coverage for 2.2.0 so I'm preemptively removing it here to get rid of the failures.

### DIFF
--- a/test/Directory.Build.props
+++ b/test/Directory.Build.props
@@ -4,7 +4,7 @@
   <PropertyGroup>
     <DeveloperBuildTestTfms>netcoreapp2.1</DeveloperBuildTestTfms>
     <StandardTestTfms>$(DeveloperBuildTestTfms)</StandardTestTfms>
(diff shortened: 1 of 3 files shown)


pranavkm:
Would it make sense to replace `netcoreapp2.0` -> `netcoreapp2.1` in other places in code? e.g. https://github.com/aspnet/KestrelHttpServer/blob/dev/benchmarks/Kestrel.Performance/Kestrel.Performance.csproj#L7. You could remove installing the shared fx entirely in that case: https://github.com/aspnet/KestrelHttpServer/blob/dev/build/repo.props#L12. It'll speed up builds for future 2.1.x patches.

Tratcher:
I know you're planning a much more thorough cleanout for 2.2.0. I don't want to get into those details here, this is only about disabling failing tests.

PR #2543, opened 2018-05-04
Make new types that shouldn't be exposed internal

halter73:
#2544

### DIFF
--- a/src/Kestrel.Core/Internal/HttpProtocols.cs
+++ b/src/Kestrel.Core/Internal/HttpProtocols.cs
@@ -3,7 +3,7 @@
 
 using System;
 
-namespace Microsoft.AspNetCore.Server.Kestrel.Core
+namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal
 {
     [Flags]
     public enum HttpProtocols
--- a/samples/Http2SampleApp/Program.cs
+++ b/samples/Http2SampleApp/Program.cs
@@ -1,7 +1,7 @@
 using System.IO;
 using System.Net;
 using Microsoft.AspNetCore.Hosting;
-using Microsoft.AspNetCore.Server.Kestrel.Core.Internal;
+using Microsoft.AspNetCore.Server.Kestrel.Core;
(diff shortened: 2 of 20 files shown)


src/Kestrel.Core/Internal/HttpProtocols.cs
```diff
@@ -3,7 +3,7 @@
 
 using System;
 
-namespace Microsoft.AspNetCore.Server.Kestrel.Core
+namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal
 {
     [Flags]
     public enum HttpProtocols
```
Tratcher: This one can go full internal until 2.2. No need to change the namespace.

Tratcher (approved):

src/Kestrel.Core/Internal/HttpProtocols.cs
halter73: The only problem with that is that the UseHttpServer IConnectionBuilder methods will need to also change from pubternal to actually internal. @davidfowl Are you OK with that?

halter73: For reference, these are all the additional changes that would need to be made to make this internal: https://github.com/aspnet/KestrelHttpServer/commit/27c8ea4e562668ca22ba6c7f69ee47d3df22f10d

Tratcher: Though most of those namespace changes are reverts.

halter73: True. Focus on the methods/ctor/and property that are now also internal.

davidfowl: Ugh, I wanted to demo this thing but thats ok.

halter73: Or we could just keep it pubternal like it is in this PR.

Tratcher (approved):

davidfowl (approved):

halter73:

halter73:
Not approved for 2.1. We will have to make breaking change announcements for the APIs that will change in 2.2.

Issue #2546, opened 2018-05-04
When the server aborts a connection any readers should get an OperationCanceled exception instead of an IOException


halter73:
Sometimes when the connection is aborted by the server with the libuv transport, uv_write will fail with ECANCELED which will cause any attempt to read from the connection to throw an IOException instead of an OperationCanceledException like you would see from the Socket transport or even the libuv transport if the connection wasn't aborted mid-write.

skyyark:
Hi.

I am running an AspNetCore Server (Kestrel + IIS) on Windows 2012 R2, and it started to produce this ECANCELED exception in the websocket receiveasync loop after upgrading to Core 2.0. Its websocket connections are supposed to stay alive for weeks. The problem is when this exception occurs, client-side websocket does not recognize that the connection has been lost, and the client does not run its reconnection logic. I am wondering why the server is producing ECANCELED exception and if it is normal and expected behavior. 

> System.IO.IOException: Error -4081 ECANCELED operation canceled ---> Microsoft.AspNetCore.Server.Kestrel.Transport.Libuv.Internal.Networking.UvException: Error -4081 ECANCELED operation canceled
   위치: Microsoft.AspNetCore.Server.Kestrel.Transport.Libuv.Internal.LibuvOutputConsumer.<WriteOutputAsync>d__6.MoveNext()
--- 예외가 throw된 이전 위치의 스택 추적 끝 ---
   위치: System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   위치: System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   위치: Microsoft.AspNetCore.Server.Kestrel.Transport.Libuv.Internal.LibuvConnection.<Start>d__26.MoveNext()
   --- 내부 예외 스택 추적의 끝 ---
   위치: Microsoft.AspNetCore.Server.Kestrel.Internal.System.IO.Pipelines.PipeCompletion.ThrowFailed()
   위치: Microsoft.AspNetCore.Server.Kestrel.Internal.System.IO.Pipelines.Pipe.GetResult(ReadResult& result)
   위치: Microsoft.AspNetCore.Server.Kestrel.Internal.System.IO.Pipelines.Pipe.Microsoft.AspNetCore.Server.Kestrel.Internal.System.IO.Pipelines.IReadableBufferAwaiter.GetResult()
[...]

PR #2547, opened 2018-05-04
Throw ConnectionResetException when socket closed observed by write

halter73:
#2546

### DIFF
--- a/src/Kestrel.Core/Internal/HttpConnection.cs
+++ b/src/Kestrel.Core/Internal/HttpConnection.cs
@@ -105,6 +105,10 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal
         {
             try
             {
+                // TODO: When we start tracking all connection middleware for shutdown, go back
+                // to logging connections tart and stop in ConnectionDispatcher so we get these
+                // logs for all connection middleware.
+                Log.ConnectionStart(ConnectionId);
                 KestrelEventSource.Log.ConnectionStart(this);
 
                 AdaptedPipeline adaptedPipeline = null;
@@ -196,6 +200,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal
                     _context.ServiceContext.ConnectionManager.UpgradedConnectionCount.ReleaseOne();
                 }
 
+                Log.ConnectionStop(ConnectionId);
                 KestrelEventSource.Log.ConnectionStop(this);
             }
         }
--- a/src/Kestrel.Transport.Sockets/Internal/SocketConnection.cs
+++ b/src/Kestrel.Transport.Sockets/Internal/SocketConnection.cs
@@ -105,15 +105,16 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Transport.Sockets.Internal
             {
                 await ProcessReceives();
             }
-            catch (SocketException ex) when (ex.SocketErrorCode == SocketError.ConnectionReset)
+            catch (SocketException ex) when (IsConnectionResetError(ex.SocketErrorCode))
             {
-                error = new ConnectionResetException(ex.Message, ex);
-                _trace.ConnectionReset(ConnectionId);
+                // A connection reset can be reported as SocketError.ConnectionAborted on Windows
+                if (!_aborted)
+                {
+                    error = new ConnectionResetException(ex.Message, ex);
+                    _trace.ConnectionReset(ConnectionId);
+                }
             }
-            catch (SocketException ex) when (ex.SocketErrorCode == SocketError.OperationAborted ||
-                                             ex.SocketErrorCode == SocketError.ConnectionAborted ||
-                                             ex.SocketErrorCode == SocketError.Interrupted ||
-                                             ex.SocketErrorCode == SocketError.InvalidArgument)
+            catch (SocketException ex) when (IsConnectionAbortError(ex.SocketErrorCode))
             {
                 if (!_aborted)
                 {
@@ -197,7 +198,13 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Transport.Sockets.Internal
             {
                 await ProcessSends();
             }
-            catch (SocketException ex) when (ex.SocketErrorCode == SocketError.OperationAborted)
+            catch (SocketException ex) when (IsConnectionResetError(ex.SocketErrorCode))
+            {
+                // A connection reset can be reported as SocketError.ConnectionAborted on Windows
+                error = null;
+                _trace.ConnectionReset(ConnectionId);
+            }
+            catch (SocketException ex) when (IsConnectionAbortError(ex.SocketErrorCode))
             {
                 error = null;
             }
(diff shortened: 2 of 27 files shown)


halter73:
Temporarily rebased on #2574 to remove noise from DiagnsoticMemoryPool.

davidfowl:
Can we fix this properly all at once? Or is this just to work around the issue right now?

halter73:
What do you mean exactly @davidfowl? None of these fixes are "workarounds". Did you read the following that I wrote in #2547? 

> The prior strategy of waiting for the pipe completed callbacks doesn't work because blocks are returned to the memory pool after the callbacks are fired.

You could argue that ConnectionClosed should be a Task instead of a CancellationToken (and you'd be right), but I figured it doesn't hurt too much to use the existing public API instead of using a new one.

davidfowl:
I don't think the transport should be waiting on the application task. That should be in the server itself

halter73:
@davidfowl The only reason the transport is waiting on the application task is so that it can safely dispose the CTS without risking an ODE in HttpConnectionMiddleware.OnConnectionAsync(). The only other alternatives I see are:

1. Expose the CTS on the feature and have the middleware dispose it 🤢
2. Replace the CancellationToken on IConnectionLifetimeFeature with a Task (ideal, but this breaks a now-public API)
3. Have the transports trip a CTS and a TCS because why not?

@pakrym and I knew you would hate this. Switching to a TCS means that the transport doesn't need to know when the application is finished for now, but I feel that's probably just kicking the can down the road. Any transport feature that's backed by disposable stuff could benefit from observing application completion in order to prevent ODE's.

davidfowl:
Let’s tlsk about this tomorrow. I dunno if it needs to be this complicated but I appreciate that you have given it a lot of thought.

halter73:
I'm not saying those are the only options, I was basically just listing the things I've considered.

src/Kestrel.Core/Internal/HttpConnection.cs
```diff
@@ -105,6 +105,10 @@ public Task StartRequestProcessing<TContext>(IHttpApplication<TContext> applicat
         {
             try
             {
+                // TODO: When we start tracking all connection middleware for shutdown, go back
```
davidfowl: Why did you move this?

halter73: So tests can verify connection processing wrapped up by checking that ConnectionStop was called once WebHost.Dispose() completed. Since the Task returned by this method isn't tracked via the HttpConnectionManager, WebHost.Dispose() can complete before ConnectionStop was logged causing test failures.

The plan is to move it back once we track all connections including those not handled by the HttpMiddleware. It will probably makes sense to move the EventSource logs to here at the same time.

davidfowl: All of these temporary hacks to make the tests pass are beginning to bug me. Does this at least make the tests more green now?

muratg: @davidfowl would you recommend a different approach?

halter73: Personally, I think it will be trivial and obvious to switch back when we move connection tracking downstack. I can create a new issue to track it. My focus atm is making the tests more reliable.

halter73:
#2547

src/Kestrel.Transport.Sockets/Internal/SocketConnection.cs
```diff
@@ -105,16 +105,29 @@ public void Dispose()
...
                 await ProcessReceives();
             }
-            catch (SocketException ex) when (ex.SocketErrorCode == SocketError.ConnectionReset)
+            catch (SocketException ex) when (IsConnectionResetError(ex.SocketErrorCode))
             {
-                error = new ConnectionResetException(ex.Message, ex);
-                _trace.ConnectionReset(ConnectionId);
+                _trace.LogTrace("DoReceive Reset (_aborted? {aborted}) {status}: {ex}", _aborted, (int)ex.SocketErrorCode, ex);
+                // A connection reset can be reported as SocketError.ConnectionAborted on Windows
+                if (!_aborted)
```
pakrym: Merge with when condition when removing logs.

test/Kestrel.FunctionalTests/ResponseTests.cs
```diff
@@ -2485,23 +2519,83 @@ public Task ResponseStatusCodeSetBeforeHttpContextDisposedRequestMalformedReadIg
...
+                {
+                    await connection.Send(
+                        "GET / HTTP/1.1",
+                        "Host:",
+                        "",
+                        "");
 
-                    await appCompletedTcs.Task.TimeoutAfter(TestConstants.DefaultTimeout);
+                    // Force a reset
+                    connection.Socket.LingerState = new LingerOption(true, 0);
```
pakrym: Make it a TestConnection method.

test/Kestrel.FunctionalTests/ResponseTests.cs
```diff
@@ -2485,23 +2519,83 @@ public Task ResponseStatusCodeSetBeforeHttpContextDisposedRequestMalformedReadIg
...
+
+            var testContext = new TestServiceContext(mockLoggerFactory.Object);
+
+            // There's not guarantee that the app even gets invoked in this test. The connection reset can be observed
+            // as early as accept.
+            using (var server = new TestServer(context => Task.CompletedTask, testContext, listenOptions))
+            {
+                using (var connection = server.CreateConnection())
+                {
+                    await connection.Send(
```
pakrym: This test is a bit funky already, but what if we run a loop of these? Around 10-100 times?

src/Kestrel.Transport.Sockets/Internal/SocketConnection.cs
halter73: We're swallowing the exception if aborted, we don't want it to fall through to the next catch.

test/Kestrel.FunctionalTests/ResponseTests.cs
```diff
@@ -2485,23 +2519,83 @@ public Task ResponseStatusCodeSetBeforeHttpContextDisposedRequestMalformedReadIg
...
+                // On Linux and macOS, the connection close is still sometimes observed as a FIN despite the LingerState.
+                Assert.True(connectionResetLogged || (!TestPlatformHelper.IsWindows && connectionFinLogged), "Connection reset not logged.");
+            }
+
+            mockKestrelTrace.Verify(t => t.ConnectionStop(It.IsAny<string>()), Times.Once());
+            Assert.True(requestAborted, "RequestAborted token didn't fire.");
+            Assert.False(loggedHigherThanDebug, "Logged message with level higher than debug.");
+        }
+
+
```
pakrym: Nit: Newline

test/shared/TaskTimeoutExtensions.cs
```diff
@@ -0,0 +1,20 @@
...
+    public static class TaskTimeoutExtensions
+    {
+        public static Task<T> DefaultTimeout<T>(this Task<T> task)
+        {
+            return task.DefaultTimeout();
+        }
+
+        public static Task DefaultTimeout(this Task task)
+        {
+            return task.DefaultTimeout();
```
pakrym: `13:24:24   Process is terminated due to StackOverflowException.`

Issue #2545, opened 2018-05-04
Autogenerate parts of TransportConnection


halter73:
A lot of the feature collection implementation for TransportConnection can be auto generated like it is for HttpProtocol. We should do this, since writing the same code manually is prone to copy-and-paste bugs.

PR #2548, opened 2018-05-05
Generate repetitive TransportConnection FeatureCollection code

halter73:
#2545


src/Kestrel.Transport.Abstractions/Internal/TransportConnection.FeatureCollection.cs
```diff
@@ -0,0 +1,111 @@
...
+            }
+            set
+            {
+                this[key] = value;
+            }
+        }
+
+        TFeature IFeatureCollection.Get<TFeature>()
+        {
+            return Get<TFeature>();
```
benaadams: Why introduce a double call?

Just add `IFeatureCollection.` to the private ones in the generated as before? This call can't be inlined as its an interface, and the next can't be inlined as its a its too large (and has too many generic type casts); so its adding a double call cost vs single call cost.

src/Kestrel.Transport.Abstractions/Internal/TransportConnection.FeatureCollection.cs
```diff
@@ -0,0 +1,111 @@
...
+    {
+        // NOTE: When feature interfaces are added to or removed from this TransportConnection class implementation,
+        // then the list of `features` in the generated code project MUST also be updated.
+        // See also: tools/CodeGenerator/TransportConnectionFeatureCollection.cs
+
+        bool IFeatureCollection.IsReadOnly => false;
+
+        int IFeatureCollection.Revision => _featureRevision;
+
+        object IFeatureCollection.this[Type key]
```
benaadams: Same comment (double call)

src/Kestrel.Transport.Abstractions/Internal/TransportConnection.FeatureCollection.cs
```diff
@@ -0,0 +1,111 @@
...
+                this[key] = value;
+            }
+        }
+
+        TFeature IFeatureCollection.Get<TFeature>()
+        {
+            return Get<TFeature>();
+        }
+
+        void IFeatureCollection.Set<TFeature>(TFeature feature)
```
benaadams: Same comment (double call)

src/Kestrel.Transport.Abstractions/Internal/TransportConnection.FeatureCollection.cs
halter73: The idea is to share the code to generate HttpProtocol which has connection feature fallback and TransportConnection which doesn't have any fallback.

PR #2560, opened 2018-05-09
Enable pubternal analyzer

pakrym:


pakrym:
I was going to, deciding what to do with APIs is a separate conversation that might last a long time. I can file an issue but wouldn't like to block PR on it.

Tratcher:
Then what's the benefit of this PR? To discourage any further exposure?

pakrym:
Yes, for example we accidentally exposed https://github.com/aspnet/KestrelHttpServer/pull/2560/files#diff-4c4e31b7f64ff3373186c05d2ffa1ea1R46 in 2.1

src/Kestrel.Core/ListenOptions.cs
```diff
@@ -17,7 +17,9 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core
     /// Describes either an <see cref="IPEndPoint"/>, Unix domain socket path, or a file descriptor for an already open
     /// socket that Kestrel should bind to or open.
     /// </summary>
+#pragma warning disable PUB0001 // Pubternal type in public API
     public class ListenOptions : IEndPointInformation, IConnectionBuilder
+#pragma warning restore PUB0001 // Pubternal type in public API
```
halter73: It should be allowed to implement a pubternal interface like you can with an internal one.

halter73:
No one merge until https://github.com/aspnet/KestrelHttpServer/pull/2543 is in dev.

pakrym:
@halter73 what should I do with this PR now?

PR #2561, opened 2018-05-10
Add support for netcoreapp2.2 in benchmarks

sebastienros:


sebastienros:
Only impacts the benchmarks, and get them to work, 4 days without platform benchmarks so far ...

Issue #2554, opened 2018-05-08
An existing connection was forcibly closed by the remote host


halter73:
_From @sepehr1014 on May 8, 2018 10:25_

Using ASP.NET Core 2.1 RC1 the process is killed after an hour or so and we see these logs:

```
Application: dotnet.exe
CoreCLR Version: 4.6.26426.2
Description: The process was terminated due to an unhandled exception.
Exception Info: System.AggregateException: One or more errors occurred. (An existing connection was forcibly closed by the remote host) ---> System.Net.Sockets.SocketException: An existing connection was forcibly closed by the remote host
   at System.Net.Sockets.Socket.UpdateStatusAfterSocketErrorAndThrowException(SocketError error, String callerName)
   at System.Net.Sockets.Socket.Shutdown(SocketShutdown how)
   at Microsoft.AspNetCore.Server.Kestrel.Transport.Sockets.Internal.SocketConnection.Shutdown()
   at Microsoft.AspNetCore.Server.Kestrel.Transport.Sockets.Internal.SocketConnection.Abort()
   at Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http.Http1OutputProducer.Abort(Exception error)
   at Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http.HttpProtocol.Abort(Exception error)
   at Microsoft.AspNetCore.Server.Kestrel.Core.Internal.HttpConnection.Abort(Exception ex)
   at System.IO.Pipelines.PipeCompletionCallbacks.Execute()
   --- End of inner exception stack trace ---
   at System.IO.Pipelines.PipeCompletionCallbacks.Execute()
   at System.Threading.ExecutionContext.RunInternal(ExecutionContext executionContext, ContextCallback callback, Object state)
--- End of stack trace from previous location where exception was thrown ---
   at System.Threading.ThreadPoolWorkQueue.Dispatch()

```

```
Faulting application name: dotnet.exe, version: 2.1.26426.3, time stamp: 0x5ae1770d
Faulting module name: KERNELBASE.dll, version: 10.0.14393.2189, time stamp: 0x5abdad60
Exception code: 0xe0434352
Fault offset: 0x0000000000033c58
Faulting process id: 0x2e90
Faulting application start time: 0x01d3e6aa48e5f5c6
Faulting application path: C:\Program Files\dotnet\dotnet.exe
Faulting module path: C:\Windows\System32\KERNELBASE.dll
[...]

PR #2562, opened 2018-05-11
Handle exception in SocketConnection.Shutdown()

halter73:
#2554

### DIFF
--- a/src/Kestrel.Transport.Sockets/Internal/SocketConnection.cs
+++ b/src/Kestrel.Transport.Sockets/Internal/SocketConnection.cs
@@ -263,8 +263,16 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Transport.Sockets.Internal
                     _aborted = true;
                     _trace.ConnectionWriteFin(ConnectionId);
 
-                    // Try to gracefully close the socket even for aborts to match libuv behavior.
-                    _socket.Shutdown(SocketShutdown.Both);
+                    try
+                    {
+                        // Try to gracefully close the socket even for aborts to match libuv behavior.
+                        _socket.Shutdown(SocketShutdown.Both);
+                    }
+                    catch
+                    {
+                        // Ignore any errors from Socket.Shutdown since we're tearing down the connection anyway.
+                    }
+
                     _socket.Dispose();
                 }
             }
--- a/test/Kestrel.FunctionalTests/RequestTests.cs
+++ b/test/Kestrel.FunctionalTests/RequestTests.cs
@@ -22,6 +22,7 @@ using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.Http.Features;
 using Microsoft.AspNetCore.Server.Kestrel.Core;
 using Microsoft.AspNetCore.Server.Kestrel.Core.Features;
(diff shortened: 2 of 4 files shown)


src/Kestrel.Transport.Sockets/Internal/SocketConnection.cs
```diff
@@ -263,8 +263,16 @@ private void Shutdown()
...
+                    {
+                        // Try to gracefully close the socket even for aborts to match libuv behavior.
+                        _socket.Shutdown(SocketShutdown.Both);
+                    }
+                    catch
+                    {
+                        // Ignore any errors from Socket.Shutdown since we're tearing down the connection anyway.
+                    }
+
                     _socket.Dispose();
```
halter73: @stephentoub @tmds I'm a little worried Socket.Dispose() could throw too. I don't wan't to put it in the same try block as Shutdown because we still want to call Dispose() when Shutdown throws. Should we add another try/catch or are we confident enough that Socket.Dispose() will never throw that we're willing to let the process crash if it does? Thanks.

stephentoub: I've not seen exceptions emerge from Socket.Dispose, but that doesn't mean it couldn't happen.  @davidsh?

davidsh: Socket.Dispose() is never supposed to throw.  And if the Socket is already disposed, it should be a no-op.

halter73: Thanks.

test/Kestrel.FunctionalTests/RequestTests.cs
```diff
@@ -1208,6 +1209,147 @@ public void AbortingTheConnectionSendsFIN()
...
+            }
+
+            mockKestrelTrace.Verify(t => t.ConnectionStop(It.IsAny<string>()), Times.Once());
+        }
+
+        [Theory]
+        [MemberData(nameof(ConnectionAdapterData))]
+        public async Task AppCanHandleClientAbortingConnectionMidRequest(ListenOptions listenOptions)
+        {
+            var readTcs = new TaskCompletionSource<Exception>(TaskContinuationOptions.RunContinuationsAsynchronously);
```
halter73: @stephentoub In what situations will this not actually run continuations asynchronously? [This gist](https://gist.github.com/halter73/6270b0c90372df2e1d2302b026b6cb39) includes a program that demonstrates the continuation runs inline pretty frequently.

halter73:
@dotnet-bot test OSX 16.04 Release Build

halter73:

src/Kestrel.Transport.Sockets/Internal/SocketConnection.cs
tmds: Indeed. Socket.Dispose will never throw ODE. Only the first one to Dispose actually disposes and others noop: https://github.com/dotnet/corefx/blob/48ce3254e28f2dae2d9a5efe7dd42cfab119d287/src/System.Net.Sockets/src/System/Net/Sockets/Socket.cs#L4456

Issue #2565, opened 2018-05-11
HttpConnectionMiddleware can OnConnectionClosed before the transports input writer loop is complete.


halter73:
HttpConnectionMiddleware can OnConnectionClosed before the transports input writer loop is complete. This can lead to issues where the server disposes the memory pool during shutdown prior to the transport's input writer returning all blocks to the pool leading to test failures like the following:

```
2018-05-11 11:57:12.588 -07:00 [TestLifetime] [Information] Starting test "WritingToConnectionAfterUnobservedCloseTriggersRequestAbortedToken-http://127.0.0.1:0"
2018-05-11 11:57:12.596 -07:00 [Microsoft.AspNetCore.Server.Kestrel.FunctionalTests.ResponseTests] [Debug] Hosting starting
2018-05-11 11:57:12.597 -07:00 [Microsoft.AspNetCore.Server.Kestrel.FunctionalTests.ResponseTests] [Debug] Hosting started
2018-05-11 11:57:12.597 -07:00 [Microsoft.AspNetCore.Server.Kestrel.FunctionalTests.ResponseTests] [Debug] Loaded hosting startup assembly "Sockets.FunctionalTests, Version=2.2.0.0, Culture=neutral, PublicKeyToken=adb9793829ddae60"
2018-05-11 11:57:12.597 -07:00 [Microsoft.AspNetCore.Server.Kestrel.FunctionalTests.ResponseTests] [Debug] Connection id ""0HLDNE2IVBEIU"" started.
2018-05-11 11:57:12.598 -07:00 [Microsoft.AspNetCore.Server.Kestrel.FunctionalTests.ResponseTests] [Information] Request starting HTTP/1.1 POST http:///  8193
2018-05-11 11:57:12.598 -07:00 [Microsoft.AspNetCore.Server.Kestrel.FunctionalTests.ResponseTests] [Debug] Connection id ""0HLDNE2IVBEIU"" paused.
2018-05-11 11:57:12.598 -07:00 [Microsoft.AspNetCore.Server.Kestrel.FunctionalTests.ResponseTests] [Debug] Connection id ""0HLDNE2IVBEIU"" resumed.
2018-05-11 11:57:12.599 -07:00 [Microsoft.AspNetCore.Server.Kestrel.FunctionalTests.ResponseTests] [Debug] Connection id ""0HLDNE2IVBEIU"" paused.
[...]

PR #2566, opened 2018-05-11
Wait for input writer to complete before calling OnConnectionClosed

halter73:
#2565

### DIFF
--- a/src/Kestrel.Core/Internal/HttpConnectionMiddleware.cs
+++ b/src/Kestrel.Core/Internal/HttpConnectionMiddleware.cs
@@ -76,37 +76,40 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal
             var processingTask = connection.StartRequestProcessing(_application);
 
             connectionContext.Transport.Input.OnWriterCompleted(
-                (error, state) => CompletedCallback(error, state),
+                (error, state) => ((PipeCompletionState)state).CompletionCallback(error),
                 inputCompletionState);
 
             connectionContext.Transport.Output.OnReaderCompleted(
-                (error, state) => CompletedCallback(error, state),
+                (error, state) => ((PipeCompletionState)state).CompletionCallback(error),
                 outputCompletionState);
 
-            await inputCompletionState.CompletedTcs.Task;
-            await outputCompletionState.CompletedTcs.Task;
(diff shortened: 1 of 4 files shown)


halter73:
@davidfowl please take a look.

src/Kestrel.Core/Internal/HttpConnectionMiddleware.cs
```diff
@@ -69,22 +70,43 @@ public Task OnConnectionAsync(ConnectionContext connectionContext)
...
             {
-                ((HttpConnection)state).OnConnectionClosed(error);
-            },
-            connection);
+                Connection = connection;
+            }
 
-            return processingTask;
+            public HttpConnection Connection;
+            public TaskCompletionSource<object> CompletedTcs = new TaskCompletionSource<object>();
```
davidfowl: Do you want to run these continuations inline?

src/Kestrel.Core/Internal/HttpConnectionMiddleware.cs
```diff
@@ -69,22 +70,43 @@ public Task OnConnectionAsync(ConnectionContext connectionContext)
...
+            await inputCompletionState.CompletedTcs.Task;
+            await outputCompletionState.CompletedTcs.Task;
+
+            connection.OnConnectionClosed();
 
-            connectionContext.Transport.Output.OnReaderCompleted((error, state) =>
+            await processingTask;
+        }
+
+        private static void CompletedCallback(Exception error, object state)
```
pakrym: Move this into `PipeCompletionState`

pakrym: And only expose `Task` from it.

src/Kestrel.Core/Internal/HttpConnectionMiddleware.cs
halter73: The pipe completion callbacks already get dispatched by default. This is closer to calling Abort and OnConnectionClosed directly in the callback like before.

src/Kestrel.Core/Internal/HttpConnectionMiddleware.cs
halter73: But now I'll have to cast twice 😢

PR #2572, opened 2018-05-14
Reenable api check

JunTaoLuo:
Addresses https://github.com/aspnet/KestrelHttpServer/issues/2350. We have updated the runtimes of the buildtools which unblocked this issue. However won't we be broken again when something changes in the runtime future? Do we want to keep doing this or wait for the overhaul of the API checks at https://github.com/aspnet/BuildTools/issues/594?

Tested locally.


JunTaoLuo:
Looks like most of our tests have been failing for a while?

PR #2575, opened 2018-05-15
[Platform] Read and Advance once per parse loop

benaadams:
Rather than `Reader.AdvanceTo` for every request in the input; only Advance the Reader once all the input has been processed (e.g. 1 once per loop rather than 16 times per loop for pipelined)

Rather than `Reader.ReadAsync` for every request in the input; only Read for more data once all the input has been processed (e.g. 1 once per loop rather than 16 times per loop for pipelined) 

/cc @davidfowl @sebastienros

### DIFF
--- a/benchmarkapps/PlatformBenchmarks/BenchmarkApplication.HttpConnection.cs
+++ b/benchmarkapps/PlatformBenchmarks/BenchmarkApplication.HttpConnection.cs
@@ -50,26 +50,39 @@ namespace PlatformBenchmarks
                 }
 
                 var result = await task;
-                if (!ParseHttpRequest(ref result))
+                var buffer = result.Buffer;
+                while (true)
                 {
+                    if (!ParseHttpRequest(ref buffer, result.IsCompleted, out var examined))
+                    {
+                        return;
+                    }
+
+                    if (_state == State.Body)
+                    {
+                        await ProcessRequestAsync();
+
+                        _state = State.StartLine;
+
+                        if (!buffer.IsEmpty)
+                        {
+                            // More input data to parse
+                            continue;
+                        }
+                    }
+
+                    // No more input or incomplete data, Advance the Reader
+                    Reader.AdvanceTo(buffer.Start, examined);
                     break;
-                }
-
-                if (_state == State.Body)
-                {
-                    await ProcessRequestAsync();
-
-                    _state = State.StartLine;
                 }
             }
         }
 
-        // Should be `in` but ReadResult isn't readonly struct
-        private bool ParseHttpRequest(ref ReadResult result)
+        private bool ParseHttpRequest(ref ReadOnlySequence<byte> buffer, bool isCompleted, out SequencePosition examined)
         {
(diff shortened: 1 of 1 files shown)


benchmarkapps/PlatformBenchmarks/BenchmarkApplication.HttpConnection.cs
```diff
@@ -81,33 +94,31 @@ private bool ParseHttpRequest(ref ReadResult result)
                     {
                         state = State.Headers;
                     }
+
+                    buffer = buffer.Slice(consumed);
                 }
 
                 if (state == State.Headers)
                 {
-                    if (parsingStartLine)
```
pakrym: Remove the var

pakrym: And use switch?

benaadams: Removed

Falls through; can't do that with switch (in C#)

Is better in follow up PR https://github.com/aspnet/KestrelHttpServer/pull/2576

```csharp
private State ParseHttpRequest(State state, ref ReadOnlySequence<byte> buffer, ref SequencePosition examined)
{
    SequencePosition consumed = default;

    if (state == State.StartLine)
    {
        if (Parser.ParseRequestLine(new ParsingAdapter(this), buffer, out consumed, out examined))
        {
            state = State.Headers;
        }

        buffer = buffer.Slice(consumed);
    }

    if (state == State.Headers)
    {
        if (Parser.ParseHeaders(new ParsingAdapter(this), buffer, out consumed, out examined, out int consumedBytes))
        {
            state = State.Body;
        }

        buffer = buffer.Slice(consumed);
    }

    return state;
}
```

benchmarkapps/PlatformBenchmarks/BenchmarkApplication.HttpConnection.cs
```diff
@@ -50,64 +50,74 @@ public partial class BenchmarkApplication : IHttpConnection
...
-                var parsingStartLine = state == State.StartLine;
-                if (parsingStartLine)
+                if (state == State.StartLine)
                 {
                     if (Parser.ParseRequestLine(new ParsingAdapter(this), buffer, out consumed, out examined))
                     {
                         state = State.Headers;
                     }
+
+                    buffer = buffer.Slice(consumed);
```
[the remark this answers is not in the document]
benaadams: Then it would be a different pattern to ParseHeaders? 

Also `consumed` is the value for `buffer.Start` which it is then exchanged for in this PR. e.g.
```
Reader.AdvanceTo(consumed, examined);
```
becomes
```
Reader.AdvanceTo(buffer.Start, examined);
```
So its for maintaining the `AdvanceTo` behaviour

benaadams:
https://github.com/TechEmpower/FrameworkBenchmarks/pull/3745

PR #2576, opened 2018-05-15
[Platform] sync fast path

benaadams:
Is https://github.com/aspnet/KestrelHttpServer/pull/2575

> Rather than `Reader.AdvanceTo` for every request in the input; only Advance the Reader once all the input has been processed (e.g. 1 once per loop rather than 16 times per loop for pipelined)
>
> Rather than `Reader.ReadAsync` for every request in the input; only Read for more data once all the input has been processed (e.g. 1 once per loop rather than 16 times per loop for pipelined) 

Then adds a sync fast-path for the tight parsing/response loop

/cc @davidfowl @sebastienros

### DIFF
--- a/benchmarkapps/PlatformBenchmarks/BenchmarkApplication.HttpConnection.cs
+++ b/benchmarkapps/PlatformBenchmarks/BenchmarkApplication.HttpConnection.cs
@@ -50,26 +50,39 @@ namespace PlatformBenchmarks
                 }
 
                 var result = await task;
-                if (!ParseHttpRequest(ref result))
+                var buffer = result.Buffer;
+                while (true)
                 {
+                    if (!ParseHttpRequest(ref buffer, result.IsCompleted, out var examined))
+                    {
+                        return;
+                    }
+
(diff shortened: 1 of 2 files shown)


benaadams:
Hopefully now should have > +3%

benaadams:
ProcessRequests (568k calls) is fast-path entry point which then calls into ProcessRequestAsync (9M calls)

image

benaadams:
Also added tweaks for 2.2 libs which I saw along the way

ReadOnlySpan.CopyTo https://github.com/dotnet/coreclr/issues/18005 -> https://github.com/dotnet/coreclr/pull/18006

Split BufferReader.Advance with fast path https://github.com/aspnet/Common/pull/357
Buffer Reader.Advance move all exceptional cases to slow path https://github.com/aspnet/Common/pull/358
BufferReader .ctor inline fast-path in MoveNext https://github.com/aspnet/Common/pull/359

davidfowl:
Took this instead https://github.com/aspnet/KestrelHttpServer/pull/2575

Issue #2573, opened 2018-05-14
[PERF] Regression on Json scenarios


sebastienros:
Regression of approximately 10% on Json serialization scenarios between versions `2.2.0-preview1-34172` and `2.2.0-preview1-34173`, around May 11 12pm UTC.

Update: Apparently on Linux only

sebastienros:
This commit: https://github.com/aspnet/KestrelHttpServer/commit/b040e33b1bbcc9719dc8e71c707a7e3fe1ff99ca

Source: http://aspnetci/viewLog.html?buildId=465847&buildTypeId=AspNetCore_Publish&tab=buildChangesDiv

davidfowl:
Must have been the scheduling change

davidfowl:
This should only have affected Linux

sebastienros:
@davidfowl awesome, because this is exactly what I saw too and had already updated the first post

DamianEdwards:
Helpe understand which versions are impacted here please? The linked commit was a merge from 2.1 to 2.2 (dev) so are we saying this is a regression in 2.1 RTM from what was previously in dev?

benaadams:
> Must have been the scheduling change

`merged 2 commits into release/2.1 ... 4 days ago` https://github.com/aspnet/KestrelHttpServer/pull/2557 had the change

```diff
-            // On *nix platforms, Sockets already dispatches to the ThreadPool.
-            var awaiterScheduler = IsWindows ? _scheduler : PipeScheduler.Inline;
```

davidfowl:
Yes, it's that one. 

/cc @halter73

PR #2578, opened 2018-05-16
[2.1.1] Fix Json perf regression in Socket Transport

halter73:
#2573 

This PR is based on dev for now. Should we start preparing a patch branch?


davidfowl:
Yes

sebastienros:
Did you measure or should I?

halter73:
@natemcmaster Should I create a release/2.1.1 branch based off of release/2.1 for this?

sebastienros:
I built this branch locally and ran with and without this assembly in release on Linux. But it gets worse with this change at 261K vs 300K. We were almost at 320K before the regression.

natemcmaster:
> Should I create a release/2.1.1 branch based off of release/2.1 for this?

You can use a branch named `halter73/release/2.1.1` if you want to stage multiple 2.1.1 fixes, but don't add a new `release/*` branch. Those have special meaning to our CI, and are also the way we keep track of changes that have patch approval. Once we finish RTM, 2.1.1 fixes will go into `release/2.1`.

sebastienros:
I verified the performance numbers and the fix reverted the regression.

PR #2580, opened 2018-05-16
Http[Request|Response]Stream.ValidateState made inlineable

gfoidl:
Instead of the `switch` `if` is used and the code got tightened.
Now it is inlined in the methods, that other than the call to `ValidateState` just forward to the `Read/Write`-methods.

### DIFF
--- a/src/Kestrel.Core/Internal/Http/HttpResponseStream.cs
+++ b/src/Kestrel.Core/Internal/Http/HttpResponseStream.cs
@@ -7,6 +7,7 @@ using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Infrastructure;
 using Microsoft.AspNetCore.Http.Features;
+using System.Runtime.CompilerServices;
 
 namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
 {
@@ -145,25 +146,26 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
             }
         }
 
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
         private void ValidateState(CancellationToken cancellationToken)
         {
-            switch (_state)
+            var state = _state;
+            if (state == HttpStreamState.Open || state == HttpStreamState.Aborted)
             {
-                case HttpStreamState.Open:
-                    if (cancellationToken.IsCancellationRequested)
-                    {
-                        cancellationToken.ThrowIfCancellationRequested();
-                    }
-                    break;
-                case HttpStreamState.Closed:
-                    throw new ObjectDisposedException(nameof(HttpResponseStream), CoreStrings.WritingToResponseBodyAfterResponseCompleted);
-                case HttpStreamState.Aborted:
-                    if (cancellationToken.IsCancellationRequested)
-                    {
-                        // Aborted state only throws on write if cancellationToken requests it
-                        cancellationToken.ThrowIfCancellationRequested();
-                    }
-                    break;
+                if (cancellationToken.IsCancellationRequested)
+                {
+                    // Aborted state only throws on write if cancellationToken requests it
+                    cancellationToken.ThrowIfCancellationRequested();
+                }
+            }
+            else
+            {
+                ThrowObjectDisposedException();
+            }
+
+            void ThrowObjectDisposedException()
+            {
+                throw new ObjectDisposedException(nameof(HttpResponseStream), CoreStrings.WritingToResponseBodyAfterResponseCompleted);
             }
         }
     }
--- a/src/Kestrel.Core/Internal/Http/HttpResponseStream.cs
+++ b/src/Kestrel.Core/Internal/Http/HttpResponseStream.cs
@@ -152,11 +152,8 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
             var state = _state;
             if (state == HttpStreamState.Open || state == HttpStreamState.Aborted)
             {
-                if (cancellationToken.IsCancellationRequested)
-                {
-                    // Aborted state only throws on write if cancellationToken requests it
-                    cancellationToken.ThrowIfCancellationRequested();
-                }
+				// Aborted state only throws on write if cancellationToken requests it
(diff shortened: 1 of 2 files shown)


src/Kestrel.Core/Internal/Http/HttpResponseStream.cs
```diff
@@ -145,25 +146,23 @@ public void Abort()
...
-                    throw new ObjectDisposedException(nameof(HttpResponseStream), CoreStrings.WritingToResponseBodyAfterResponseCompleted);
-                case HttpStreamState.Aborted:
-                    if (cancellationToken.IsCancellationRequested)
-                    {
-                        // Aborted state only throws on write if cancellationToken requests it
-                        cancellationToken.ThrowIfCancellationRequested();
-                    }
-                    break;
+				// Aborted state only throws on write if cancellationToken requests it
+				cancellationToken.ThrowIfCancellationRequested();
```
halter73: Do we know if CancellationToken.ThrowIfCancellationRequested itself gets inlined?

src/Kestrel.Core/Internal/Http/HttpResponseStream.cs
```diff
@@ -145,25 +146,23 @@ public void Abort()
             }
         }
 
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
```
pakrym: Do you need the attribute for this method? It looks pretty small

src/Kestrel.Core/Internal/Http/HttpResponseStream.cs
gfoidl: Yes it gets inlined. Checked the dasm. And it's just the same property-access, so this duplication was removed. See https://github.com/dotnet/coreclr/blob/85374ceaed177f71472cc4c23c69daf7402e5048/src/System.Private.CoreLib/src/System/Threading/CancellationToken.cs#L305-L311

src/Kestrel.Core/Internal/Http/HttpResponseStream.cs
gfoidl: Unfortunately yes. At least on `linux-x64` where I checked the dasm.

pakrym: What's the reason for not inlining? Unprofitable?

pakrym:
What motivated this change? Any before-after results?

gfoidl:
> What motivated this change?

Stepped with the debugger through the source, spotted these methods and I knew that `switch` won't inline. Hence I submitted this PR.

> Any before-after results?

I'm just getting into the Kestrel-source, so can you give me any pointers to relevant benchmarks? Rolling out my own that cover these aspects may be not so easy (I believe).

src/Kestrel.Core/Internal/Http/HttpResponseStream.cs
gfoidl: Yeah `unprofitable inline`.

halter73:
I don't think we have any microbenchmarks for `Http[Request|Response]Stream` yet. Feel free to add one. For your other PR, there's the [StringUtilitesBenchmark](https://github.com/aspnet/KestrelHttpServer/blob/27799b5acb59523889a3a52a39d606e795b81c50/benchmarks/Kestrel.Performance/StringUtilitiesBenchmark.cs).

pakrym:
You can also try https://github.com/aspnet/KestrelHttpServer/blob/dev/benchmarks/Kestrel.Performance/InMemoryTransportBenchmark.cs that runs entire pipeline in memory and see if it shows any change.

halter73:
InMemoryTransportBenchmark is a good idea. I don't think that tests the Http*Request*Stream though.

gfoidl:
### InMemoryTransportBenchmark

Note I ran the benchmarks with `netcoreapp2.1`, not with `netcoreapp2.0` as in the `csproj`.

#### Current dev

|             Method |     Mean |     Error |    StdDev |      Op/s |  Gen 0 | Allocated |
|------------------- |---------:|----------:|----------:|----------:|-------:|----------:|
|          Plaintext | 7.886 us | 0.1449 us | 0.1285 us | 126,807.0 |      - |     371 B |
| PlaintextPipelined | 3.914 us | 0.0775 us | 0.1060 us | 255,511.0 | 0.1221 |     432 B |

#### This PR

|             Method |     Mean |     Error |    StdDev |   Median |      Op/s |  Gen 0 | Allocated |
|------------------- |---------:|----------:|----------:|---------:|----------:|-------:|----------:|
|          Plaintext | 7.137 us | 0.1814 us | 0.5292 us | 6.869 us | 140,111.2 |      - |     368 B |
| PlaintextPipelined | 3.794 us | 0.0711 us | 0.0665 us | 3.781 us | 263,607.0 | 0.1221 |     432 B |

So it got a bit faster.
`HttpRequestStream` isn't touched by this benchmark. I assume this will also be faster, but not in the same amount as `HttpResponseStream`, due to the more complex validation.

muratg:
@JunTaoLuo Could you take a look? Let's see if we can do this in 2.2 instead.

JunTaoLuo:
I just re-ran the benchmarks on netcoreapp2.2 and here are the results:

Before:
```
             Method |     Mean |     Error |    StdDev |      Op/s | Allocated |
------------------- |---------:|----------:|----------:|----------:|----------:|
          Plaintext | 7.962 us | 0.1587 us | 0.3980 us | 125,598.0 |     369 B |
 PlaintextPipelined | 5.247 us | 0.0906 us | 0.0757 us | 190,575.9 |     432 B |
```
After:
```
             Method |     Mean |     Error |    StdDev |      Op/s | Allocated |
------------------- |---------:|----------:|----------:|----------:|----------:|
          Plaintext | 7.738 us | 0.1529 us | 0.2425 us | 129,240.2 |     369 B |
 PlaintextPipelined | 4.784 us | 0.0421 us | 0.0329 us | 209,020.8 |     432 B |
```

The improvements are modest. I'll merge to release/2.2 and master.

JunTaoLuo:
Merged in https://github.com/aspnet/KestrelHttpServer/commit/773aa6e6ffa542b30165db4cde3ef3a167f28ffc. Thansk @gfoidl !

PR #2586, opened 2018-05-17
Improve reliability of MaxRequestBufferSizeTests.LargeUpload

mikeharder:
- Increase _dataLength from 20MB to 40MB to improve reliability of Sockets on Windows
- Addresses https://github.com/aspnet/KestrelHttpServer/issues/2225


mikeharder:

mikeharder:
@halter73, @Tratcher: Added pause data for MaxRequestBufferSize=1MB (the default).  I don't understand why, but the Libuv and Sockets behaviors are much closer compared to MaxRequestBufferSize=5MB.

mikeharder:
@halter73, @Tratcher: OK to merge?  I could file an issue to investigate why Sockets gets paused so much later than Libuv when `MaxRequestBufferSize=5MB`, but since the two transports have similar behavior for the default `MaxRequestBufferSize=1MB`, I don't think it's worth the time.

PR #2588, opened 2018-05-18
Enable test file logging on Jenkins

halter73:


build/buildpipeline/osx.groovy
```diff
@@ -5,6 +5,8 @@ simpleNode('OSX10.12','latest') {
         checkout scm
     }
     stage ('Build') {
-        sh "./build.sh --ci /p:Configuration=${params.Configuration}"
+        def logFolder = getLogFolder()
+        def environment = "ASPNETCORE_TEST_LOG_DIR=${WORKSPACE}\\${logFolder}"
+        sh "${environment} ./build.sh --ci /p:Configuration=${params.Configuration}"
```
pranavkm: As long as it works. On Windows the whitespace resulting in the directory having a trailing whitespace which the `getLogFolder()` archiver does not support.

JunTaoLuo:
I just added msbuild properties to configure logging in https://github.com/aspnet/Logging/pull/836/files#diff-a881d07eb9db902bdb25589d86a598acR3. I would recommend setting the `LoggingTestingFileLoggingDirectory` msbuild property instead.

PR #2592, opened 2018-05-20
Derive SocketAwaitable from SocketAsyncEventArgs

gfoidl:
Based on https://github.com/aspnet/KestrelHttpServer/pull/2147#discussion_r149276777

`SocketAwaitable` (renamed to `SocketAwaitableEventArgs`) derives from `SocketAsyncEventArgs` and overrides `OnCompleted`, so 
* no extra object for the awaitable is needed and two int-fields saved
* no need to register to the `Completed` event

By the way
* `GetResult` made inlineable
* `SocketReceiver` and `SocketSender` uses a common base class (refactored)

I'd like to run some RPS-benchmark similar to https://github.com/aspnet/KestrelHttpServer/pull/2147, but I don't know how. Can you please give me some advice for it?


halter73:
@gfoidl I like this because it looks like pretty good code cleanup. I don't think the per-connection allocations this avoids will make a significant impact in our plaintext or json benchmarks, but I can try running it tomorrow.

gfoidl:
Rebased due to

```
Conflicting files

src/Kestrel.Transport.Sockets/Internal/SocketReceiver.cs
src/Kestrel.Transport.Sockets/Internal/SocketSender.cs
```

davidfowl:
@halter73 do you want to take this in 2.2?

halter73:
Yep.

davidfowl:
@dotnet-bot test this

PR #2593, opened 2018-05-20
Interlocked.Read/Add in SocketConnection only for 32-bit systems

gfoidl:
...on 64-bit system `long` reads are already atomic,, so they can be made way faster than with an interlocked operation.

The JIT will eliminate the dead-branch.

----

PS: Hope that I don't get bothersome...

### DIFF
--- a/src/Kestrel.Transport.Sockets/Internal/SocketConnection.cs
+++ b/src/Kestrel.Transport.Sockets/Internal/SocketConnection.cs
@@ -61,7 +61,21 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Transport.Sockets.Internal
         public override MemoryPool<byte> MemoryPool { get; }
         public override PipeScheduler InputWriterScheduler => _scheduler;
         public override PipeScheduler OutputReaderScheduler => _scheduler;
-        public override long TotalBytesWritten => Interlocked.Read(ref _totalBytesWritten);
+
+        public override long TotalBytesWritten
+        {
+            get
+            {
+                if (IntPtr.Size == sizeof(long))
+                {
+                    return _totalBytesWritten;
+                }
+                else
+                {
+                    return Interlocked.Read(ref _totalBytesWritten);
+                }
+            }
+        }
 
         public async Task StartAsync()
         {
@@ -244,7 +258,14 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Transport.Sockets.Internal
 
                 // This is not interlocked because there could be a concurrent writer.
                 // Instead it's to prevent read tearing on 32-bit systems.
-                Interlocked.Add(ref _totalBytesWritten, buffer.Length);
+                if (IntPtr.Size == sizeof(long))
+                {
+                    _totalBytesWritten += buffer.Length;
+                }
+                else
+                {
+                    Interlocked.Add(ref _totalBytesWritten, buffer.Length);
+                }
 
                 Output.AdvanceTo(end);
 
--- a/src/Kestrel.Transport.Sockets/Internal/SocketConnection.cs
+++ b/src/Kestrel.Transport.Sockets/Internal/SocketConnection.cs
@@ -242,16 +242,12 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Transport.Sockets.Internal
                     await _sender.SendAsync(buffer);
                 }
 
-                // This is not interlocked because there could be a concurrent writer.
+                // This is not "interlocked" because there could be a concurrent writer.
                 // Instead it's to prevent read tearing on 32-bit systems.
-                if (IntPtr.Size == sizeof(long))
-                {
-                    _totalBytesWritten += buffer.Length;
-                }
-                else
-                {
-                    Interlocked.Add(ref _totalBytesWritten, buffer.Length);
-                }
+                // The implementation is a faster variant for Interlocked.Add (at least on 64-bit)
+                var totalBytesWritten = Volatile.Read(ref _totalBytesWritten);
+                totalBytesWritten += buffer.Length;
+                Volatile.Write(ref _totalBytesWritten, totalBytesWritten);
 
                 Output.AdvanceTo(end);
 
--- a/src/Kestrel.Transport.Sockets/Internal/SocketConnection.cs
+++ b/src/Kestrel.Transport.Sockets/Internal/SocketConnection.cs
@@ -61,21 +61,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Transport.Sockets.Internal
         public override MemoryPool<byte> MemoryPool { get; }
         public override PipeScheduler InputWriterScheduler => _scheduler;
         public override PipeScheduler OutputReaderScheduler => _scheduler;
-
-        public override long TotalBytesWritten
-        {
-            get
-            {
-                if (IntPtr.Size == sizeof(long))
-                {
-                    return _totalBytesWritten;
-                }
-                else
-                {
-                    return Interlocked.Read(ref _totalBytesWritten);
-                }
-            }
-        }
+        public override long TotalBytesWritten => Volatile.Read(ref _totalBytesWritten);
 
         public async Task StartAsync()
         {


src/Kestrel.Transport.Sockets/Internal/SocketConnection.cs
```diff
@@ -61,7 +61,21 @@ internal SocketConnection(Socket socket, MemoryPool<byte> memoryPool, PipeSchedu
...
         public override PipeScheduler OutputReaderScheduler => _scheduler;
-        public override long TotalBytesWritten => Interlocked.Read(ref _totalBytesWritten);
+
+        public override long TotalBytesWritten
+        {
+            get
+            {
+                if (IntPtr.Size == sizeof(long))
+                {
+                    return _totalBytesWritten;
```
halter73: Shouldn't Interlocked.Read already optimized to this on 64-bit systems? @benaadams

halter73:
> PS: Hope that I don't get bothersome...

Don't worry about that. Just don't be to surprised if not everything gets merged.

src/Kestrel.Transport.Sockets/Internal/SocketConnection.cs
benaadams: Nah, [`Interlocked.Read`](https://github.com/dotnet/coreclr/blob/d73a35065ec05a8179c8b812986bd9708001782c/src/System.Private.CoreLib/src/System/Threading/Interlocked.cs#L157-L160) always does what it says:
```csharp
public static partial class Interlocked
{
    public static long Read(ref long location)
    {
        return Interlocked.CompareExchange(ref location, 0, 0);
    }
}
```
You are thinking of [`Volatile.Read`](https://github.com/dotnet/coreclr/blob/d73a35065ec05a8179c8b812986bd9708001782c/src/System.Private.CoreLib/src/System/Threading/Volatile.cs#L110-L160)
```csharp
public static partial class Volatile
{
#if BIT64
    public static long Read(ref long location)
    {
        // The VM will replace this with a more efficient implementation.
        var value = location;
        Interlocked.MemoryBarrier();
        return value;
    }
#else
    public static long Read(ref long location)
    {
        // On 32-bit machines, we use this implementation, since an ordinary volatile read
        // would not be atomic.
        //
        // On 64-bit machines, the VM will replace this with a more efficient implementation.
        return Interlocked.CompareExchange(ref location, 0, 0);
    }
#endif
}
```

halter73: I probably should have just looked for myself. Thanks though 😄

benaadams: Though it should probably be just changed to `Volatile.Read` rather than using `Interlocked.Read`

halter73: Will that prevent tearing?

benaadams: ^--- has a `#if BIT64` switch that uses `Interlocked.CompareExchange` in the 32bit path

gfoidl:
> don't be too surprised if not everything gets merged.

Yeah, I'm aware of this :wink:

src/Kestrel.Transport.Sockets/Internal/SocketConnection.cs
gfoidl: For reference the following code produces 

```c#
public class Bench
{
    private long _value = 42;

    [Benchmark(Baseline = true)]
    public long WithInterlocked() => Interlocked.Read(ref _value);

    [Benchmark]
    public long DirectReadOnx64()
    {
        if (IntPtr.Size == sizeof(long))
        {
            return _value;
        }
        else
        {
            return Interlocked.Read(ref _value);
        }
    }

    [Benchmark]
    public long VolatileRead() => Volatile.Read(ref _value);
}
```

this dasm

```asm
; WithInterlocked
00007FFC0E951510  cmp         dword ptr [rcx],ecx
00007FFC0E951512  lea         rdx,[rcx+8]
00007FFC0E951516  xor         ecx,ecx
00007FFC0E951518  xor         eax,eax
00007FFC0E95151A  lock cmpxchg qword ptr [rdx],rcx
00007FFC0E95151F  ret
;------------------------------------------------------------------------------
; DirectReadOnx64
00007FFC0E951530  mov         rax,qword ptr [rcx+8]
00007FFC0E951534  ret
;------------------------------------------------------------------------------
; VolatileRead
00007FFC0E951550  lea         rax,[rcx+8]
00007FFC0E951554  mov         rax,qword ptr [rax]
00007FFC0E951557  ret
```

So `DirectReadOnx64` and `VolatileRead` is "the same" code.

For bench-results I get this:

|          Method |      Mean |     Error |    StdDev | Scaled |
|---------------- |----------:|----------:|----------:|-------:|
| WithInterlocked | 5.9178 ns | 0.1028 ns | 0.0912 ns |  1.000 |
| DirectReadOnx64 | 0.0260 ns | 0.0229 ns | 0.0215 ns |  0.004 |
|    VolatileRead | 0.0194 ns | 0.0191 ns | 0.0179 ns |  0.003 |

So quite as expected.

I'll update to `Volatile.Read`.

src/Kestrel.Transport.Sockets/Internal/SocketConnection.cs
```diff
@@ -244,7 +244,14 @@ public void Dispose()
...
                 // This is not interlocked because there could be a concurrent writer.
                 // Instead it's to prevent read tearing on 32-bit systems.
-                Interlocked.Add(ref _totalBytesWritten, buffer.Length);
+                if (IntPtr.Size == sizeof(long))
+                {
+                    _totalBytesWritten += buffer.Length;
+                }
+                else
+                {
+                    Interlocked.Add(ref _totalBytesWritten, buffer.Length);
```
gfoidl: Here a variant with `Volatile` isn't as efficient as this one.

```c#
long totalBytesWritten = Volatile.Read(ref _totalBytesWritten);
totalBytesWritten += buffer.Length;
Volatile.Write(ref _totalBytesWritten, totalBytesWritten);
```

produces

```asm
00007FFC0DCA1570  lea         rax,[rcx+8]  
00007FFC0DCA1574  mov         rcx,rax  
00007FFC0DCA1577  mov         rcx,qword ptr [rcx]  
00007FFC0DCA157A  add         rcx,rdx  
00007FFC0DCA157D  mov         qword ptr [rax],rcx  
```

instead of a simple

```asm
00007FFC0DCA1550  add         qword ptr [rcx+8],rdx  
```
(for the x64-path).

|         Method |     Mean |     Error |    StdDev | Scaled |
|--------------- |---------:|----------:|----------:|-------:|
| InterlockedAdd | 6.159 ns | 0.0545 ns | 0.0510 ns |   1.00 |
| DirectAddOnx64 | 1.010 ns | 0.0319 ns | 0.0298 ns |   0.16 |
|    VolatileAdd | 1.354 ns | 0.0282 ns | 0.0264 ns |   0.22 |

So this one should be kept as-is in this PR.

tmds:
I wonder if this is ok for some arm like architectures? I think it may cause the write to not be observed by other threads?

halter73:
@tmds Do the Interlocked methods not introduce a memory barrier on arm?

src/Kestrel.Transport.Sockets/Internal/SocketConnection.cs
```diff
@@ -244,7 +244,14 @@ public void Dispose()
 
                 // This is not interlocked because there could be a concurrent writer.
                 // Instead it's to prevent read tearing on 32-bit systems.
-                Interlocked.Add(ref _totalBytesWritten, buffer.Length);
+                if (IntPtr.Size == sizeof(long))
+                {
+                    _totalBytesWritten += buffer.Length;
```
tmds: @halter73 my comment/question is about removing `Interlocked` here.
Reads and writes have semantics about how observable they are by other threads. Maybe by removing the `Interlocked` here, the change to `_totalBytesWritten` can go unnoticed by other threads? Or perhaps the C#/.NET memory model requires such writes (e.g. to member fields) to always be visible? I'm not sure.
In my own code I play safe and when using `Volatile.Read` for reading, I use it for writing too.

gfoidl: Good point.

> the C#/.NET memory model requires such writes (e.g. to member fields) to always be visible?

To my understanding this guarantee isn't given by the ECMA standard.
x86/amd64 usually have a strong memory model, so it is fine (and I just had this architecture in mind when doing the PR).
arm has a weak memory model, so it must not be fine, and your concern seems correct.

To be on the safe side we should use https://github.com/aspnet/KestrelHttpServer/pull/2593#discussion_r189871592 for both 32/64 bit? Or fallback to `Interlocked` but it's 4x slower than the `Volatile`-variant.

tmds: I think we can use either 'Volatile' or 'Interlocked' and that will be correct on 32-bit and 64-bit. So, we can pick the fastest and have the same code for both architectures.

muratg:
We’re closing this PR because this code was removed in 2.2.

PR #2604, opened 2018-05-23
[Platform] Update to latest (add db routes)

benaadams:
/cc @sebastienros


sebastienros:
@davidfowl what do you think with having db code in this repos? Or should we keep this part only in TE and just sync changes between the two repos, as we do with aspnet/Benchmarks.

davidfowl:
I'm not a fan of putting this scenario here.

benaadams:
Probably should live somewhere other than the TE repo? aspnet/Benchmarks then?

Could split it into more partial classes; separate the routing and db then have it as a src share for the non-db...

Will work something out; and see what you think.

davidfowl:
Meh, it's fine to put it here. It's the lesser of all evils.

muratg:
We’re closing this PR because we feel like Benchmarks is a better place to keep this. 

Issue #2609, opened 2018-05-23
Flaky Test: HeartbeatTests on Travis


halter73:
E.g.:

```
Starting test group: UnitTests/netcoreapp2.2
  Microsoft (R) Test Execution Command Line Tool Version 15.7.0
  Copyright (c) Microsoft Corporation.  All rights reserved.
  
  Starting test execution, please wait...
  [xUnit.net 00:00:14.6094487]     BlockedHeartbeatDoesntCauseOverlapsAndIsLoggedAsError [FAIL]
  Failed   BlockedHeartbeatDoesntCauseOverlapsAndIsLoggedAsError
RUNDOTNET : error Message:  [/home/travis/.dotnet/buildtools/korebuild/2.2.0-preview1-17060/KoreBuild.proj]
   Assert.True() Failure
  Expected: True
  Actual:   False
  Stack Trace:
     at Microsoft.AspNetCore.Server.Kestrel.Core.Tests.HeartbeatTests.BlockedHeartbeatDoesntCauseOverlapsAndIsLoggedAsError() in /_/test/Kestrel.Core.Tests/HeartbeatTests.cs:line 43
  [xUnit.net 00:00:16.6720891]     BlockedHeartbeatIsNotLoggedAsErrorIfDebuggerAttached [FAIL]
  Failed   BlockedHeartbeatIsNotLoggedAsErrorIfDebuggerAttached
RUNDOTNET : error Message:  [/home/travis/.dotnet/buildtools/korebuild/2.2.0-preview1-17060/KoreBuild.proj]
   Moq.MockException : 
  Expected invocation on the mock once, but was 2 times: h => h.OnHeartbeat(5/23/18 5:14:58 PM +00:00)
  
  Configured setups:
  h => h.OnHeartbeat(5/23/18 5:14:58 PM +00:00), Times.Exactly(2)
  
  Performed invocations:
  IHeartbeatHandler.OnHeartbeat(5/23/18 5:14:58 PM +00:00)
  IHeartbeatHandler.OnHeartbeat(5/23/18 5:14:58 PM +00:00)
  Stack Trace:
     at Moq.Mock.ThrowVerifyException(MethodCall expected, IEnumerable`1 setups, IEnumerable`1 actualCalls, Expression expression, Times times, Int32 callCount)
     at Moq.Mock.VerifyCalls(Interceptor targetInterceptor, MethodCall expected, Expression expression, Times times)
[...]

PR #2610, opened 2018-05-23
Don't block so many threads in HeartbeatTests

halter73:
#2609


halter73:
This should make the HeartbeatTests more reliable on low-core CIs like travis.

Issue #2608, opened 2018-05-23
Don't ever throw when writing to a response unless passed a CancellationToken


halter73:
Certain errors, like a connection timeout or application aborting the connection or a timeout, can cause writes to fail even when no CancellationToken is passed in.

PR #2612, opened 2018-05-24
Handle client and server aborts differently from eachhother

halter73:
#2608

### DIFF
--- a/src/Connections.Abstractions/DefaultConnectionContext.cs
+++ b/src/Connections.Abstractions/DefaultConnectionContext.cs
@@ -67,7 +67,7 @@ namespace Microsoft.AspNetCore.Connections
 
         public void Abort() => Abort(abortReason: null);
 
-        public virtual void Abort(ConnectionAbortedException abortReason)
+        public override void Abort(ConnectionAbortedException abortReason)
         {
             ThreadPool.QueueUserWorkItem(cts => ((CancellationTokenSource)cts).Cancel(), _connectionClosedTokenSource);
         }
--- a/src/Connections.Abstractions/DefaultConnectionContext.cs
+++ b/src/Connections.Abstractions/DefaultConnectionContext.cs
@@ -67,7 +67,7 @@ namespace Microsoft.AspNetCore.Connections
 
         public void Abort() => Abort(abortReason: null);
 
-        public virtual void Abort(Exception abortReason)
+        public virtual void Abort(ConnectionAbortedException abortReason)
         {
             ThreadPool.QueueUserWorkItem(cts => ((CancellationTokenSource)cts).Cancel(), _connectionClosedTokenSource);
         }
--- a/src/Connections.Abstractions/DefaultConnectionContext.cs
+++ b/src/Connections.Abstractions/DefaultConnectionContext.cs
@@ -65,7 +65,9 @@ namespace Microsoft.AspNetCore.Connections
 
         public CancellationToken ConnectionClosed { get; set; }
 
-        public virtual void Abort()
+        public void Abort() => Abort(abortReason: null);
+
+        public virtual void Abort(Exception abortReason)
         {
             ThreadPool.QueueUserWorkItem(cts => ((CancellationTokenSource)cts).Cancel(), _connectionClosedTokenSource);
         }
--- a/src/Connections.Abstractions/Features/IConnectionLifetimeFeature.cs
+++ b/src/Connections.Abstractions/Features/IConnectionLifetimeFeature.cs
@@ -1,6 +1,7 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System;
 using System.Threading;
 
 namespace Microsoft.AspNetCore.Connections.Features
@@ -8,6 +9,12 @@ namespace Microsoft.AspNetCore.Connections.Features
     public interface IConnectionLifetimeFeature
     {
         CancellationToken ConnectionClosed { get; set; }
+
         void Abort();
+
+        // REVIEW: This is technically a breaking change for implementors of this interface,
+        // but realistically this only affects transport implementations which today must
+        // implement other pubternal interfaces or facades which I doubt exist.
+        void Abort(Exception abortReason);
     }
 }
--- a/src/Connections.Abstractions/Features/IConnectionLifetimeFeature.cs
+++ b/src/Connections.Abstractions/Features/IConnectionLifetimeFeature.cs
@@ -1,7 +1,6 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
-using System;
 using System.Threading;
 
 namespace Microsoft.AspNetCore.Connections.Features
@@ -11,10 +10,5 @@ namespace Microsoft.AspNetCore.Connections.Features
         CancellationToken ConnectionClosed { get; set; }
 
         void Abort();
-
-        // REVIEW: This is technically a breaking change for implementors of this interface,
-        // but realistically this only affects transport implementations which today must
-        // implement other pubternal interfaces or facades which I doubt exist.
-        void Abort(ConnectionAbortedException abortReason);
     }
 }
--- a/src/Connections.Abstractions/Features/IConnectionLifetimeFeature.cs
+++ b/src/Connections.Abstractions/Features/IConnectionLifetimeFeature.cs
@@ -15,6 +15,6 @@ namespace Microsoft.AspNetCore.Connections.Features
         // REVIEW: This is technically a breaking change for implementors of this interface,
         // but realistically this only affects transport implementations which today must
         // implement other pubternal interfaces or facades which I doubt exist.
-        void Abort(Exception abortReason);
+        void Abort(ConnectionAbortedException abortReason);
     }
 }
--- a/src/Kestrel.Core/Adapter/Internal/AdaptedPipeline.cs
+++ b/src/Kestrel.Core/Adapter/Internal/AdaptedPipeline.cs
@@ -6,7 +6,9 @@ using System.IO;
 using System.IO.Pipelines;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http;
+using Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Infrastructure;
 using Microsoft.AspNetCore.Server.Kestrel.Transport.Abstractions.Internal;
+using Microsoft.Extensions.Logging;
 
 namespace Microsoft.AspNetCore.Server.Kestrel.Core.Adapter.Internal
 {
@@ -15,22 +17,23 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Adapter.Internal
         private static readonly int MinAllocBufferSize = KestrelMemoryPool.MinimumSegmentSize / 2;
 
         private readonly IDuplexPipe _transport;
-        private readonly IDuplexPipe _application;
 
         public AdaptedPipeline(IDuplexPipe transport,
-                               IDuplexPipe application,
                                Pipe inputPipe,
-                               Pipe outputPipe)
+                               Pipe outputPipe,
+                               IKestrelTrace log)
         {
             _transport = transport;
-            _application = application;
             Input = inputPipe;
             Output = outputPipe;
+            Log = log;
         }
 
         public Pipe Input { get; }
 
         public Pipe Output { get; }
+
+        public IKestrelTrace Log { get; }
 
         PipeReader IDuplexPipe.Input => Input.Reader;
 
@@ -47,8 +50,6 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Adapter.Internal
 
         private async Task WriteOutputAsync(Stream stream)
         {
-            Exception error = null;
-
             try
             {
                 if (stream == null)
@@ -63,13 +64,6 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Adapter.Internal
 
                     try
                     {
-                        if (result.IsCanceled)
-                        {
-                            // Forward the cancellation to the transport pipe
-                            _application.Input.CancelPendingRead();
-                            break;
-                        }
-
                         if (buffer.IsEmpty)
                         {
                             if (result.IsCompleted)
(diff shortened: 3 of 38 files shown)


src/Kestrel.Core/CoreStrings.resx
```diff
@@ -506,4 +506,16 @@ For more information on configuring HTTPS see https://go.microsoft.com/fwlink/?l
   <data name="BadRequest_RequestBodyTimeout" xml:space="preserve">
     <value>Reading the request body timed out due to data arriving too slowly. See MinRequestBodyDataRate.</value>
   </data>
+  <data name="ConnectionAbortedByApplication" xml:space="preserve">
+    <value>The connection was aborted by the application.</value>
+  </data>
+  <data name="ConnectionAbortedDuringServerShutdown" xml:space="preserve">
+    <value>The connection was aborted because the server is shutting down and request processing didn't complete within the time specified by HostOptions.ShutdownTimeout.</value>
```
davidfowl: 👍

src/Connections.Abstractions/Features/IConnectionLifetimeFeature.cs
```diff
@@ -1,13 +1,20 @@
...
     public interface IConnectionLifetimeFeature
     {
         CancellationToken ConnectionClosed { get; set; }
+
         void Abort();
+
+        // REVIEW: This is technically a breaking change for implementors of this interface,
+        // but realistically this only affects transport implementations which today must
+        // implement other pubternal interfaces or facades which I doubt exist.
+        void Abort(ConnectionAbortedException abortReason);
```
halter73: @muratg

davidfowl: This seems a bit weird. Interface breaking change aside, why `ConnectionAbortedException`

halter73: The ConnectionAbortedException is to maintain breaking compatibility. Previously, when the output PipeWriter was completed with an error, it would be swallowed by the libuv transport and turned into a ConnectionAbortedException with the (oh so helpful) "The connection was aborted" message. Now at least the message is more helpful.

Now that we actually rethrow the exception passed to abort to any pending input readers, it's important to call Abort with a ConnectionAbortedException for back compat. When I first made this change some tests failed because they were expecting an OperationCanceledException (which a ConnectionAbortedException is) instead of a TimeoutException.

I changed the type in the method signature so the compiler makes it impossible to break compat unintentionally, but we could change it back to taking any exception and just be careful when calling it.

halter73: I still think it makes  moresense for pending readers to observe a `ConnectionAbortedException` when the connection is aborted due to a call to `IConnectionLifetimeFeature.Abort()`. It's nice when you're reading a request body to know that every exception will be either an `IOException` or an `OperationCanceledException`. Having to handle completely arbitrary exceptions like `TimeoutException`s isn't nice.

davidfowl: I don't have a problem with the raised exception to be `ConnectionAbortedException`, I don't think it needs to be the originating exception type though. You can wrap the exception passed into a ConnectionAbortedException

halter73: Also, you can set the inner exception to whatever you want. Having a call to Abort() result in a ConnectionAbortedException just seems logical.

halter73: > You can wrap the exception passed into a ConnectionAbortedException

This is a transport feature, so every transport would have to have the exact same wrapping logic, and there would be no compile time guarantees the transport gets it right. I don't see why we wouldn't just keep the common logic in the core assembly.

src/Kestrel.Core/Adapter/Internal/AdaptedPipeline.cs
```diff
@@ -63,13 +64,6 @@ public class AdaptedPipeline : IDuplexPipe
 
                     try
                     {
-                        if (result.IsCanceled)
```
davidfowl: Why did we stop checking this? Aren't we still cancelling on the transport side?

src/Kestrel.Core/Internal/Http/Http1OutputProducer.cs
```diff
@@ -181,17 +179,15 @@ public void Abort(Exception error)
...
+                _lifetimeFeature.Abort(error);
+
                 if (!_completed)
                 {
                     _log.ConnectionDisconnect(_connectionId);
                     _completed = true;
-
-                    _outputPipeReader.CancelPendingRead();
-                    _pipeWriter.Complete(error);
+                    _pipeWriter.Complete();
```
davidfowl: So we never complete the transport with an error.

davidfowl:

davidfowl:
@halter73 lets mark this as a breaking change and doc it (even though in theory nobody will hit it).

davidfowl:
```
[xUnit.net 00:00:16.1659949]     BadRequestLogsAreNotHigherThanInformation [FAIL]
  Failed   BadRequestLogsAreNotHigherThanInformation
RUNDOTNET : error Message:  [C:\Users\appveyor\.dotnet\buildtools\korebuild\2.1.1-rtm-15790\KoreBuild.proj]
   Assert.All() Failure: 1 out of 10 items in the collection did not pass.
  [9]: Item: Microsoft.Extensions.Logging.Testing.WriteContext
       Xunit.Sdk.InRangeException: Assert.InRange() Failure
       Range:  (Trace - Information)
       Actual: Error
          at Xunit.Assert.InRange[T](T actual, T low, T high, IComparer`1 comparer) in C:\Dev\xunit\xunit\src\xunit.assert\Asserts\RangeAsserts.cs:line 41
          at Xunit.Assert.All[T](IEnumerable`1 collection, Action`1 action) in C:\Dev\xunit\xunit\src\xunit.assert\Asserts\CollectionAsserts.cs:line 36
  Stack Trace:
     at 
```

src/Kestrel.Core/Adapter/Internal/AdaptedPipeline.cs
```diff
@@ -108,7 +102,7 @@ public class AdaptedPipeline : IDuplexPipe
             }
             catch (Exception ex)
             {
-                error = ex;
+                Log.LogError(0, ex, $"{nameof(AdaptedPipeline)}.{nameof(WriteOutputAsync)}");
```
davidfowl: This should only happen now if writing to the stream fails, yeh

src/Kestrel.Core/Internal/Http2/Http2OutputProducer.cs
```diff
@@ -25,7 +26,7 @@ public void Dispose()
         {
         }
 
-        public void Abort(Exception error)
+        public void Abort(ConnectionAbortedException error)
```
davidfowl: ConnectionAbortedException may not make sense. You're aborting the request stream.

src/Kestrel.Core/Internal/HttpConnection.cs
```diff
@@ -123,9 +125,9 @@ public Task StartRequestProcessing<TContext>(IHttpApplication<TContext> applicat
                 if (_context.ConnectionAdapters.Count > 0)
                 {
                     adaptedPipeline = new AdaptedPipeline(_adaptedTransport,
-                                                          application,
```
davidfowl: 👏

src/Kestrel.Core/Internal/HttpConnection.cs
```diff
@@ -622,13 +641,11 @@ void IConnectionTimeoutFeature.ResetTimeout(TimeSpan timeSpan)
...
 
-        private void CloseUninitializedConnection()
+        private void CloseUninitializedConnection(ConnectionAbortedException abortReason)
         {
             Debug.Assert(_adaptedTransport != null);
 
-            // CancelPendingRead signals the transport directly to close the connection
-            // without any potential interference from connection adapters.
-            _context.Application.Input.CancelPendingRead();
+            _context.ConnectionFeatures.Get<IConnectionLifetimeFeature>().Abort(abortReason);
```
davidfowl: We should store the ConnectionContext on the `HttpConnectionContext`.

src/Connections.Abstractions/DefaultConnectionContext.cs
```diff
@@ -65,7 +65,9 @@ public DefaultConnectionContext(string id, IDuplexPipe transport, IDuplexPipe ap
 
         public CancellationToken ConnectionClosed { get; set; }
 
-        public virtual void Abort()
+        public void Abort() => Abort(abortReason: null);
```
davidfowl: Ugh, if we're going to make a breaking change, I'd like this to be moved to ConnectionContext.

src/Kestrel.Core/Internal/HttpConnection.cs
davidfowl: Ugh, see https://github.com/aspnet/KestrelHttpServer/pull/2612/files#r191705652

src/Connections.Abstractions/DefaultConnectionContext.cs
halter73: Does SignalR ever directly implement ConnectionContext in its tests or anything like that? I want to know if, unlike the IConnectionLifetimeFeature breaking change, this could actually break anyone downstream. @anurse

src/Kestrel.Core/Adapter/Internal/AdaptedPipeline.cs
halter73: Correct. This would be really bizarre given that Stream.WriteAsync should never fail on the RawStream given that no CancellationToken is passed here, but not strictly impossible.

src/Kestrel.Core/Internal/Http/Http1OutputProducer.cs
halter73: Correct. The Abort method is now the one true way to fail the connection.

Before completing with an error was one way to tell the transport not to worry about any unflushed data (in fact the transport wouldn't even be able to read any more unconsumed data from output pipe after the pipe was completed with the failure), but now that calling IConnectionLifetimeFeature.Abort(ex) cancels any pending reads from the output pipe which breaks the send loop in both transport, this is no longer necessary.

src/Kestrel.Core/Internal/Http2/Http2OutputProducer.cs
halter73: I don't think I understand what you're trying to get at here. Since writing to the response stream no longer fails absent a CancellationToken, the ConnectionAbortedException is only observed via the RequestStream in HTTP/1 as well.

src/Kestrel.Core/Internal/HttpConnection.cs
halter73: Seems reasonable. Might as well get it right if we're going to break things. I agree that aborting a connection is common enough to deserve a top-level method on the ConnectionContext. Interestingly enough this will be the first non-property on the ConnectionContext.

I think if we made the Abort(ConnectionAbortedException) method virtual instead of abstract on ConnectionContext, it would technically be non-breaking. Do you think we should do that?

halter73: The risk of making Abort virtual is that transports might not implement it. On second thought, I would rather just break compilation.

src/Connections.Abstractions/DefaultConnectionContext.cs
davidfowl: It would be added as virtual so non-breaking.

halter73: Yuck. https://github.com/aspnet/KestrelHttpServer/pull/2612/files#r191886723

To continue with this "if we're going to make a breaking change" mindset, do you think we should remove both Abort methods from IConnectionLifetimeFeature and change IConnectionLifetimeFeature.ConnectionClosed to a Task since it always completes eventually so we don't have to worry about the transports waiting for the middleware to complete prior to disposing the ConnectionClosed CancellationToken?

halter73: If you're committed to making Abort virtual, should ConnectionContext.Abort throw a NotImplementedException? I don't see how that's better than a MethodMissingException, but at least you wouldn't end up silently leaking sockets.

anurse: We do have a `TestConnectionContext` that inherits from it but it should be fine, it should just inherit it.

halter73: FWIW, you could probably just use Microsoft.AspNetCore.Connections.DefaultConnectionContext instead of using your own TestConnectionContext.

anurse: Yeah, we do that in other places, not entirely sure why we do it there, I'll take a quick scan.

halter73:
@davidfowl Adding ConnectionContext.Abort(reason) even as virtual no-op method is unfortunately still breaking if we ever want to even try to call it (short of using reflection to see if the method actually there) 😢 . The problem is that if you have 2.1.1/2.2.0 Kestrel.Core assembly and a 2.1.0 Connections.Abstractions assembly you will get a MissingMethodException when attempting to call Abort.

halter73:
@davidfowl  Assuming it's impossible to have an app running with Kestrel.Core v2.2.0 and Connections.Abstractions v2.1.0, this change is no longer breaking.

src/Kestrel.Transport.Abstractions/Internal/TransportConnection.cs
```diff
@@ -63,8 +55,14 @@ public TransportConnection()
...
-        public virtual void Abort()
+        // DO NOT remove this override to ConnectionContext.Abort(). Doing so would cause
+        // any TransportConnection that does not override Abort() or calls base.Abort()
+        // to stack overflow when IConnectionLifetimeFeature.Abort() is called.
+        public override void Abort(ConnectionAbortedException abortReason)
         {
+            AbortImpl(abortReason);
         }
+
+        public abstract void AbortImpl(ConnectionAbortedException abortReason);
```
halter73: @davidfowl I know this name is bad. Any suggestions? At least it's a pubternal type.

davidfowl: Make it protected and rename it to `AbortCore`

halter73:

src/Kestrel.Core/Internal/Http2/Http2OutputProducer.cs
davidfowl: I'm describing the nature of aborting in Http2 and passing in a ConnectionAbortedException for a request abort when it has nothing to do with aborting the underlying "connection". It's just semantics but it's not a hill I would die on as part of this PR.

Issue #2605, opened 2018-05-23
SSL error after long time running


ClassicDarkChocolate:
I'm currently using Let's Encrypt as SSL provider on my AWS EC2 server. I found that after a few hours or days, the server starts to throwing the following exception.

    fail: Microsoft.AspNetCore.Server.Kestrel[0]
          Uncaught exception from the OnConnectionAsync method of an IConnectionAdapter.
    System.Security.Authentication.AuthenticationException: Authentication failed, see inner exception. ---> Interop+OpenSsl+SslException: SSL Handshake failed with OpenSSL error - SSL_ERROR_SSL. ---> Interop+Crypto+OpenSslCryptographicException: error:140760FC:SSL routines:SSL23_GET_CLIENT_HELLO:unknown protocol
       --- End of inner exception stack trace ---
       at Interop.OpenSsl.DoSslHandshake(SafeSslHandle context, Byte[] recvBuf, Int32 recvOffset, Int32 recvCount, Byte[]& sendBuf, Int32& sendCount)
       at System.Net.Security.SslStreamPal.HandshakeInternal(SafeFreeCredentials credential, SafeDeleteContext& context, SecurityBuffer inputBuffer, SecurityBuffer outputBuffer, SslAuthenticationOptions sslAuthenticationOptions)
       --- End of inner exception stack trace ---
       at System.Net.Security.SslState.StartSendAuthResetSignal(ProtocolToken message, AsyncProtocolRequest asyncRequest, ExceptionDispatchInfo exception)
       at System.Net.Security.SslState.CheckCompletionBeforeNextReceive(ProtocolToken message, AsyncProtocolRequest asyncRequest)
       at System.Net.Security.SslState.StartSendBlob(Byte[] incoming, Int32 count, AsyncProtocolRequest asyncRequest)
       at System.Net.Security.SslState.ProcessReceivedBlob(Byte[] buffer, Int32 count, AsyncProtocolRequest asyncRequest)
       at System.Net.Security.SslState.StartReadFrame(Byte[] buffer, Int32 readBytes, AsyncProtocolRequest asyncRequest)
       at System.Net.Security.SslState.PartialFrameCallback(AsyncProtocolRequest asyncRequest)
    --- End of stack trace from previous location where exception was thrown ---
       at System.Net.Security.SslState.ThrowIfExceptional()
       at System.Net.Security.SslState.InternalEndProcessAuthentication(LazyAsyncResult lazyResult)
[...]

PR #2614, opened 2018-05-24
[2.1.1] Lower severity of AuthenticationException logs from SslStream handshake

halter73:
#2605 

@bartonjs @stephentoub

### DIFF
--- a/src/Kestrel.Core/Internal/HttpsConnectionAdapter.cs
+++ b/src/Kestrel.Core/Internal/HttpsConnectionAdapter.cs
@@ -5,6 +5,7 @@ using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Net.Security;
+using System.Security.Authentication;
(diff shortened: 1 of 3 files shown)


halter73:
@muratg I think this should also be backported for 2.1.1.

src/Kestrel.Core/Internal/HttpsConnectionAdapter.cs
```diff
@@ -192,6 +193,12 @@ public Task<IAdaptedConnection> OnConnectionAsync(ConnectionAdapterContext conte
                 sslStream.Dispose();
                 return _closedAdaptedConnection;
             }
+            catch (AuthenticationException ex)
```
pakrym: Exception filter?

muratg:
@halter73 Agreed, and I marked the bug appropriately.  Let's send this one to the shiproom with the rest.

PR #2567, opened 2018-05-11
Use correct TaskCompletionSource ctor

halter73:

PR #2526, opened 2018-04-26
React to pool cleanup

pakrym:

PR #2618, opened 2018-05-29
[2.1.1] Use correct TaskCompletionSource ctor

halter73:
These are test only changes. It's the combination of the following PRs backported for release/2.1:

- https://github.com/aspnet/KestrelHttpServer/pull/2531
- https://github.com/aspnet/KestrelHttpServer/pull/2567
- Parts of https://github.com/aspnet/KestrelHttpServer/pull/2526
- Parts of https://github.com/aspnet/KestrelHttpServer/pull/2547

### DIFF
--- a/test/Kestrel.FunctionalTests/ConnectionLimitTests.cs
+++ b/test/Kestrel.FunctionalTests/ConnectionLimitTests.cs
@@ -22,9 +22,9 @@ namespace Microsoft.AspNetCore.Server.Kestrel.FunctionalTests
         [Fact]
         public async Task ResetsCountWhenConnectionClosed()
         {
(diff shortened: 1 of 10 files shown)


halter73:
These changes have been in dev for a while and have improved test reliability. [This comment](https://github.com/aspnet/KestrelHttpServer/pull/2531#issuecomment-385472844) from @pakrym explains why:

> The common issue I discovered in our tests is completing TCS inside the appfunc and then awaiting it just before disposing the server.
This causes continuation to be run inline in appfunc and prevents graceful shutdown making tests slower and possibly more flaky.
There is also a race it that if TCS get's completed before being awaited we get different behaviour then if was awaited first and then completed.

PR #2619, opened 2018-05-29
[2.1.3] Consistently handle connection aborts

halter73:
This is a backport of graceful server shutdown fixes which improve functional test reliability, and followup PRs that normalize the exceptions thrown by the libuv and socket transports while providing better logs and error messages.

The following PRs backported to 2.1 release branch:

- https://github.com/aspnet/KestrelHttpServer/pull/2566
- https://github.com/aspnet/KestrelHttpServer/pull/2574
- https://github.com/aspnet/KestrelHttpServer/pull/2547
- https://github.com/aspnet/KestrelHttpServer/pull/2612


src/Connections.Abstractions/ConnectionContext.cs
```diff
@@ -13,5 +17,13 @@ public abstract class ConnectionContext
         public abstract IDictionary<object, object> Items { get; set; }
 
         public abstract IDuplexPipe Transport { get; set; }
+
+        public virtual void Abort(ConnectionAbortedException abortReason)
```
davidfowl: I think we want the no arg overload as well, why remove it?

src/Connections.Abstractions/DefaultConnectionContext.cs
```diff
@@ -65,7 +65,9 @@ public DefaultConnectionContext(string id, IDuplexPipe transport, IDuplexPipe ap
 
         public CancellationToken ConnectionClosed { get; set; }
 
-        public virtual void Abort()
+        public void Abort() => Abort(abortReason: null);
```
davidfowl: Add this to ConnectionContext as well.

src/Kestrel.Core/Internal/HttpConnectionMiddleware.cs
```diff
@@ -72,19 +76,26 @@ public Task OnConnectionAsync(ConnectionContext connectionContext)
...
+
+            await AsTask(lifetimeFeature.ConnectionClosed);
+
+            connection.OnConnectionClosed();
 
-            return processingTask;
+            await processingTask;
+        }
+
+        private Task AsTask(CancellationToken token)
```
davidfowl: 😢

src/Kestrel.Core/Internal/HttpConnectionMiddleware.cs
```diff
@@ -54,6 +57,7 @@ public Task OnConnectionAsync(ConnectionContext connectionContext)
             };
 
             var connectionFeature = connectionContext.Features.Get<IHttpConnectionFeature>();
+            var lifetimeFeature = connectionContext.Features.Get<IConnectionLifetimeFeature>();
```
davidfowl: This can be null right?

src/Kestrel.Transport.Abstractions/Internal/TransportConnection.cs
```diff
@@ -8,7 +8,7 @@
 
 namespace Microsoft.AspNetCore.Server.Kestrel.Transport.Abstractions.Internal
 {
-    public partial class TransportConnection : ConnectionContext
+    public abstract partial class TransportConnection : ConnectionContext
```
davidfowl: Too breaking

src/Kestrel.Transport.Libuv/Internal/ListenerContext.cs
```diff
@@ -38,6 +41,43 @@ protected UvStreamHandle CreateAcceptSocket()
             }
         }
 
+        protected async Task HandleConnectionAsync(UvStreamHandle socket)
```
davidfowl: I'm not a fan of this being in the patch. Is this what's currently checked into dev?

Tratcher:
https://github.com/aspnet/KestrelHttpServer/issues/2636

src/Kestrel.Core/Internal/HttpConnectionMiddleware.cs
halter73: No. Look at https://github.com/aspnet/KestrelHttpServer/pull/2619/files#diff-a6a566e69107e6ac28bc8c57908bf1f5L53.

src/Kestrel.Transport.Libuv/Internal/ListenerContext.cs
halter73: This is what's in dev. Even though we don't need to dispose CancellationTokenSources, I think it important for anything implementing connection features including the transport to know when the app/middleware is done executing and no longer needs access to the connection features it implements.

Just relying on Pipe APIs to determine when to cleanup resources backing connection features is insufficient. The HttpMiddleware, for example, still should be able access connection features after HttpContext.Abort() is called.

All that said, having a Task returning IConnectionDispatcher.OnConnection isn't critical to this backport given that it's acceptable to never dispose ConnectionClosed CancellationTokenSources, so I reverted that aspect of this backport.

halter73:
@davidfowl ping.

src/Kestrel.Transport.Abstractions/Internal/TransportConnection.cs
```diff
@@ -8,7 +8,7 @@
 
 namespace Microsoft.AspNetCore.Server.Kestrel.Transport.Abstractions.Internal
 {
-    public abstract partial class TransportConnection : ConnectionContext
+    public partial class TransportConnection : ConnectionContext
```
davidfowl: Isn't this breaking?

halter73: It wasn't abstract before. I made it non-abstract so that it's not breaking.

davidfowl: I see

Issue #2620, opened 2018-05-29
HttpUpgradeStream doesn't have overloads for Memory<byte>


buybackoff:
[HttpUpgradeStream](https://github.com/aspnet/KestrelHttpServer/blob/dev/src/Kestrel.Core/Internal/Http/HttpUpgradeStream.cs) was omitted from https://github.com/aspnet/KestrelHttpServer/pull/2333, while it's request and response streams do have those methods. Found it while investigating https://github.com/dotnet/corefx/issues/29951. 

Also cannot understand yet if the request and response streams use non-allocating ValueTask from https://github.com/dotnet/corefx/issues/27445?

So far all performance benefits from 2.1 are not available for WebSockets via `Microsoft.AspNetCore.WebSockets` middleware.

davidfowl:
@muratg is this something we can patch in 2.1.1 as well? There's no public API change here fortunately.

buybackoff:
Finally found your commit https://github.com/dotnet/corefx/commit/8958ef9a89f882886b233900b53788f13e6b0999 to understand if `HttpUpgradeStream` is the only blocker. And indeed it looks like a couple of `#ifdefs` should unlock the new `IValueTaskSource` feature.

halter73:
> Also cannot understand yet if the request and response streams use non-allocating ValueTask from dotnet/corefx#27445?

For reads, we use the non-allocating ValueTask since it calls through to [MessageBody.ReadAsync](https://github.com/aspnet/KestrelHttpServer/blob/f80091a3280971977ec71711b9105f82d19e37f1/src/Kestrel.Core/Internal/Http/MessageBody.cs#L39) which awaits a `PipeReader` which in turn uses `ValueTask<TResult>` all the way through and saves an allocation in the event read data is already available at the time of the call.

[...]

PR #2622, opened 2018-05-30
[2.1.1] Add Memory<byte> overloads to HttpUpgradeStream

halter73:
#2620

### DIFF
--- a/src/Kestrel.Core/Internal/Http/HttpUpgradeStream.cs
+++ b/src/Kestrel.Core/Internal/Http/HttpUpgradeStream.cs
@@ -145,6 +145,13 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
             return _requestStream.ReadAsync(buffer, offset, count, cancellationToken);
         }
 
+#if NETCOREAPP2_1
+        public override ValueTask<int> ReadAsync(Memory<byte> destination, CancellationToken cancellationToken = default)
+        {
+            return _requestStream.ReadAsync(destination, cancellationToken);
+        }
+#endif
+
         public override Task CopyToAsync(Stream destination, int bufferSize, CancellationToken cancellationToken)
         {
             return _requestStream.CopyToAsync(destination, bufferSize, cancellationToken);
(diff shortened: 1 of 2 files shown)


src/Kestrel.Core/Internal/Http/HttpUpgradeStream.cs
```diff
@@ -145,6 +145,13 @@ public override Task<int> ReadAsync(byte[] buffer, int offset, int count, Cancel
             return _requestStream.ReadAsync(buffer, offset, count, cancellationToken);
         }
 
+#if NETCOREAPP2_1
+        public override ValueTask<int> ReadAsync(Memory<byte> destination, CancellationToken cancellationToken = default)
+        {
+            return _requestStream.ReadAsync(destination, cancellationToken);
+        }
+#endif
```
Tratcher: List all of the current TFMs so we know if this gets outdated. E.g. this breaks if we add 2.2.

halter73: I'm not sure exactly what you're asking for here. But if you want to change this, we should probably change the pattern [everywhere else](https://github.com/aspnet/KestrelHttpServer/search?q=%23if+NETCOREAPP2_1&unscoped_q=%23if+NETCOREAPP2_1) too.

Tratcher: This pattern:
https://github.com/aspnet/MusicStore/blob/d971bfc8b777ca4098a951727218aef32cb9ebd5/test/MusicStore.E2ETests/Common/Helpers.cs#L51-L57
This is used pretty broadly in other repos to avoid dead code and regressions when TFMs change.

davidfowl:

src/Kestrel.Core/Internal/Http/HttpUpgradeStream.cs
halter73: I see. Do you think we should change the pattern everywhere or just one place since it fails compilation?

Tratcher: Everywhere, but it could be a seperate PR.

PR #2646, opened 2018-06-05
Improve tracking of connection lifetime

halter73:
I haven't seen this cause any issues in practice, but the following sample shows why the way we're currently tracking connection lifetimes is at least theoretically bad. I know we also want to start tracking non-HTTP connections. This PR doesn't do that, but I think we should still take this change in the meantime.

```c#
using System;
using System.Collections.Concurrent;
using System.Threading;
using System.Threading.Tasks;

namespace ConnectionManagerRace
{
    public class Program
    {
        {
            var manager = new ConnectionManager();
            Task.Run(() => StartConnections(100, manager));
            Thread.Sleep(500);
            manager.CloseAllConnectionsAsync().Wait();
        }

        public static void StartConnections(int count, ConnectionManager manager)
        {
            for (var i = 0; i < count; i++)
            {
                var conn = new Connection(manager);
                conn.StartRequestProcessing();
            }
        }
    }

    public class ConnectionManager
    {
        private readonly ConcurrentDictionary<long, Connection> _connections = new ConcurrentDictionary<long, Connection>();
        private long _lastId = 0;

        public void Add(Connection conn)
        {
            _connections.TryAdd(Interlocked.Increment(ref _lastId), conn);
        }

        public async Task CloseAllConnectionsAsync()
        {
            var closeTasks = new List<Task>();

            foreach (var pair in _connections)
            {
                closeTasks.Add(pair.Value.StopProcessingNextRequestAsync());
            }

            await Task.WhenAll(closeTasks);
        }
    }

    public class Connection
    {
        private readonly ConnectionManager _manager;
        private Task _lifetimeTask;

        public Connection(ConnectionManager manager)
        {
            _manager = manager;
        }

        public Task StartRequestProcessing()
        {
            return _lifetimeTask = ProcessRequestsAsync();
        }

        public async Task ProcessRequestsAsync()
        {
            _manager.Add(this);
            Thread.Sleep(1000);
            await Task.Delay(1000);
        }

        public Task StopProcessingNextRequestAsync()
        {
            return _lifetimeTask;
        }
    }
}
```

Which results in:

```
ConnectionManagerRace> dotnet run

Unhandled Exception: System.AggregateException: One or more errors occurred. (The tasks argument included a null value.
Parameter name: tasks) ---> System.ArgumentException: The tasks argument included a null value.
Parameter name: tasks
   at System.Threading.Tasks.Task.WhenAll(IEnumerable`1 tasks)
   at ConnectionManagerRace.ConnectionManager.CloseAllConnectionsAsync() in G:\dev\halter73\ConnectionManagerRace\Program.cs:line 48
   --- End of inner exception stack trace ---
   at System.Threading.Tasks.Task.Wait(Int32 millisecondsTimeout, CancellationToken cancellationToken)
   at System.Threading.Tasks.Task.Wait()
   at ConnectionManagerRace.Program.Main(String[] args) in G:\dev\halter73\ConnectionManagerRace\Program.cs:line 16
```

### DIFF
--- a/src/Kestrel.Core/Internal/HttpConnection.cs
+++ b/src/Kestrel.Core/Internal/HttpConnection.cs
@@ -11,7 +11,6 @@ using System.Net;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Connections;
-using Microsoft.AspNetCore.Connections.Features;
 using Microsoft.AspNetCore.Hosting.Server;
 using Microsoft.AspNetCore.Http.Features;
 using Microsoft.AspNetCore.Server.Kestrel.Core.Adapter.Internal;
@@ -30,6 +29,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal
 
         private readonly HttpConnectionContext _context;
         private readonly TaskCompletionSource<object> _socketClosedTcs = new TaskCompletionSource<object>(TaskCreationOptions.RunContinuationsAsynchronously);
+        private readonly TaskCompletionSource<object> _lifetimeTcs = new TaskCompletionSource<object>(TaskCreationOptions.RunContinuationsAsynchronously);
 
         private IList<IAdaptedConnection> _adaptedConnections;
         private IDuplexPipe _adaptedTransport;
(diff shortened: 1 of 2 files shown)


halter73:
@davidfowl ping

Issue #2251, opened 2018-01-10
TLS restrictions for HTTP/2


Tratcher:
https://tools.ietf.org/html/rfc7540#section-9.2.1

Tratcher:
Also review the other TLS requirements to make sure they're in place.

Tratcher:
- MUST disable compression
- MUST disable renegotiation
- MUST support ephemeral key exchange sizes of at least
   2048 bits for cipher suites that use ephemeral finite field Diffie-
   Hellman (DHE) [TLS12] and 224 bits for cipher suites that use
   ephemeral elliptic curve Diffie-Hellman (ECDHE) [RFC4492]
- SHOULD NOT use any of the cipher
   suites that are listed in the cipher suite black list
- MUST support TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256 [TLS-ECDHE]
   with the P-256 elliptic curve [FIPS186].

Tratcher:
@blowdart renegotiation is the only thing on this list we can control via SslStream, the rest are controlled at the OS level. Are there any of these we should ask SslStream to expose (compression?), or do we fall back do documenting the recommended OS configurations.

Drawaes:
With ref

https://en.m.wikipedia.org/wiki/Comparison_of_TLS_implementations#Compression

Tratcher:
Blocked on https://github.com/aspnet/KestrelHttpServer/issues/2661
Things we need to surface:
- Tls version
- CipherAlgorithm 
- KeyExchangeStrength

PR #2650, opened 2018-06-05
Disable AllowRenegotiation for HTTP/2

Tratcher:
#2251 This is a spec requirement.

I can't directly test it because SslStream doesn't have a client API to trigger a renegotiate. I added a test to ensure this didn't regress 1.1 client certs since renegotiation is often used to request optional client certs. We'll have to figure out http2 functional tests before I can write a http2 client cert test.

### DIFF
--- a/src/Kestrel.Core/Internal/HttpsConnectionAdapter.cs
+++ b/src/Kestrel.Core/Internal/HttpsConnectionAdapter.cs
@@ -158,6 +158,8 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Https.Internal
                 if ((_options.HttpProtocols & HttpProtocols.Http2) != 0)
                 {
                     sslOptions.ApplicationProtocols.Add(SslApplicationProtocol.Http2);
+                    // https://tools.ietf.org/html/rfc7540#section-9.2.1
+                    sslOptions.AllowRenegotiation = false;
                 }
 
                 if ((_options.HttpProtocols & HttpProtocols.Http1) != 0)
--- a/test/Kestrel.FunctionalTests/HttpsConnectionAdapterTests.cs
+++ b/test/Kestrel.FunctionalTests/HttpsConnectionAdapterTests.cs
@@ -12,6 +12,7 @@ using System.Net.Sockets;
 using System.Security.Authentication;
 using System.Security.Cryptography.X509Certificates;
 using System.Text;
+using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.Http.Features;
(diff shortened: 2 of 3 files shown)


src/Kestrel.Core/Internal/HttpsConnectionAdapter.cs
```diff
@@ -158,6 +158,8 @@ public Task<IAdaptedConnection> OnConnectionAsync(ConnectionAdapterContext conte
                 if ((_options.HttpProtocols & HttpProtocols.Http2) != 0)
                 {
                     sslOptions.ApplicationProtocols.Add(SslApplicationProtocol.Http2);
+                    // https://tools.ietf.org/html/rfc7540#section-9.2.1
+                    sslOptions.AllowRenegotiation = false;
```
halter73: I'm halfway tempted to disable this for all https connections. Do you know of any clients that actually renegotiate for any reason?

test/Kestrel.FunctionalTests/HttpsConnectionAdapterTests.cs
```diff
@@ -375,7 +376,44 @@ public void ThrowsWhenNoServerCertificateIsProvided()
                 }
             }
         }
+#if NETCOREAPP2_2
```
halter73: nit: newlines before and after the ifdef

test/Kestrel.FunctionalTests/HttpsConnectionAdapterTests.cs
```diff
@@ -381,7 +382,44 @@ public void ThrowsWhenNoServerCertificateIsProvided()
...
+                Assert.NotNull(tlsFeature.ClientCertificate);
+                Assert.NotNull(context.Connection.ClientCertificate);
+                return context.Response.WriteAsync("hello world");
+            }, new TestServiceContext(LoggerFactory), listenOptions))
+            {
+                using (var client = new TcpClient())
+                {
+                    // SslStream is used to ensure the certificate is actually passed to the server
+                    // HttpClient might not send the certificate because it is invalid or it doesn't match any
+                    // of the certificate authorities sent by the server in the SSL handshake.
```
halter73: I don't understand this comment, the other tests in this class use HttpClientSlim which has a validateCertificate option.

src/Kestrel.Core/Internal/HttpsConnectionAdapter.cs
Tratcher: I've mainly seen it used by the server to request a client certificate, but I wouldn't risk changing it for existing users right now.

>   Note that disabling
  renegotiation can result in long-lived connections becoming unusable
   due to limits on the number of messages the underlying cipher suite
   can encipher.

test/Kestrel.FunctionalTests/HttpsConnectionAdapterTests.cs
```diff
@@ -381,7 +382,44 @@ public void ThrowsWhenNoServerCertificateIsProvided()
...
+                    var stream = await OpenSslStream(client, server, HttpProtocols.Http1);
+                    await AssertConnectionResult(stream, true);
```
halter73: Nit: assert the "hello world" response content.

test/Kestrel.FunctionalTests/HttpsConnectionAdapterTests.cs
Tratcher: Consolidated the test code

test/Kestrel.FunctionalTests/HttpsConnectionAdapterTests.cs
```diff
@@ -642,7 +646,6 @@ public void ThrowsForCertificatesMissingServerEku(string testCertName)
 
             return stream;
         }
-
```
JunTaoLuo: nit: undo new line removal.

Issue #2640, opened 2018-06-01
http/2: Create automation to run h2spec


shirhatti:
Initial goal is to create baseline automation and disable it on the CI.
Look at what Websockets has done with Autobahn tests

Tratcher:
https://github.com/summerwind/h2spec

Tratcher:
@muratg are you OK with checking in the pre-compiled binaries for h2spec?
https://github.com/summerwind/h2spec/releases

That seems preferable to having to build them on the fly.

shirhatti:
Can't we do what we've done for libuv? Or is that not worth the complexity?

Tratcher:
This is quite different from libuv, it's a test only tool, not something we want to ship. It also has additional dependencies if we try to build it from source like installing the Go compiler.

The tests run in seconds so I'd like to enable them as normal functional tests in every test pass. It'll be more reliable if the binaries were always available and you didn't have to install the Go compiler on every machine.

muratg:
@Tratcher let's sync-up when you're in office on that part.

Tratcher:
Offline notes: There's been no updates to h2spec for over a year now. We can hand role a nupkg, put it on myget, and reference it from the tests.

PR #2653, opened 2018-06-07
Automation for h2spec functional tests 

Tratcher:
#2640 These won't pass on the CI yet because we still need to work out the acquisition mechanism for the h2spec binaries. The test code should be reviewable though.

Depends on https://github.com/aspnet/Universe/pull/1222

### DIFF
--- a/test/Kestrel.FunctionalTests/Http2/H2SpecTests.cs
+++ b/test/Kestrel.FunctionalTests/Http2/H2SpecTests.cs
@@ -1,0 +1,135 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+#if NETCOREAPP2_2
+
+using System.Linq;
+using System.Net;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Builder;
+using Microsoft.AspNetCore.Hosting;
+using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Server.Kestrel.Core;
+using Microsoft.AspNetCore.Testing;
+using Microsoft.AspNetCore.Testing.xunit;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace Microsoft.AspNetCore.Server.Kestrel.FunctionalTests.Http2
+{
+    [OSSkipCondition(OperatingSystems.MacOSX, SkipReason = "Missing SslStream ALPN support: https://github.com/dotnet/corefx/issues/30492")]
+    [MinimumOSVersion(OperatingSystems.Windows, WindowsVersions.Win81,
+        SkipReason = "Missing Windows ALPN support: https://en.wikipedia.org/wiki/Application-Layer_Protocol_Negotiation#Support")]
+    public class H2SpecTests : TestApplicationErrorLoggerLoggedTest
+    {
+        [ConditionalTheory]
+        [MemberData(nameof(H2SpecTestCases))]
+        public async Task RunIndividualTestCase(H2SpecTestCase testCase)
+        {
+            var hostBuilder = TransportSelector.GetWebHostBuilder()
+                .UseKestrel(options =>
+                {
+                    options.Listen(IPAddress.Loopback, 0, listenOptions =>
+                    {
+                        listenOptions.Protocols = HttpProtocols.Http2;
+                        if (testCase.Https)
+                        {
+                            listenOptions.UseHttps(TestResources.TestCertificatePath, "testPassword");
+                        }
+                    });
+                })
+                .ConfigureServices(AddTestLogging)
+                .Configure(ConfigureHelloWorld);
+
+            using (var host = hostBuilder.Build())
+            {
+                await host.StartAsync();
+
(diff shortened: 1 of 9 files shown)


test/Kestrel.FunctionalTests/TestHelpers/H2SpecCommands.cs
```diff
@@ -0,0 +1,215 @@
...
+
+            using (var process = Process.Start(processOptions))
+            {
+                var data = process.StandardOutput.ReadToEnd();
+                logger.LogDebug(data);
+
+                var results = File.ReadAllText(tempFile);
+                File.Delete(tempFile);
+
+                var xml = new XmlDocument();
```
pakrym: XDocument?

halter73:
Is there a special meaning behind WIIP, or is that just a typo?

Tratcher:
😆 indefinite?

test/Kestrel.FunctionalTests/TestHelpers/H2SpecCommands.cs
Tratcher: Any specific reason?

pakrym: All cool kids use it :)

Tratcher: You should know better than to give into peer pressure.

Tratcher:
Ready for review. The CI failures look like unrelated flaky tests.

test/Kestrel.FunctionalTests/Http2/H2SpecTests.cs
```diff
@@ -0,0 +1,135 @@
...
+            {
+                return $"{Id}, HTTPS:{Https}, {Description}";
+            }
+        }
+
+        private void ConfigureHelloWorld(IApplicationBuilder app)
+        {
+            app.Run(context =>
+            {
+                return context.Request.Body.CopyToAsync(context.Response.Body);
```
halter73: Any specific reason for copying the request body rather than just responding with "Hello World"?

test/Kestrel.FunctionalTests/Http2/H2SpecTests.cs
```diff
@@ -0,0 +1,135 @@
...
+using Microsoft.AspNetCore.Testing;
+using Microsoft.AspNetCore.Testing.xunit;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace Microsoft.AspNetCore.Server.Kestrel.FunctionalTests.Http2
+{
+    [OSSkipCondition(OperatingSystems.MacOSX, SkipReason = "Missing SslStream ALPN support: https://github.com/dotnet/corefx/issues/30492")]
+    [MinimumOSVersion(OperatingSystems.Windows, WindowsVersions.Win81,
+        SkipReason = "Missing Windows ALPN support: https://en.wikipedia.org/wiki/Application-Layer_Protocol_Negotiation#Support")]
```
halter73: Could we add a condition based on being able to execute h2spec just in case we ever try to run tests on alpine or something like that?

test/Kestrel.FunctionalTests/TestHelpers/H2SpecCommands.cs
```diff
@@ -0,0 +1,222 @@
...
+                return Path.Combine(root, "linux", "h2spec");
+            }
+            else if (RuntimeInformation.IsOSPlatform(OSPlatform.OSX))
+            {
+                return Path.Combine(root, "darwin", "h2spec");
+            }
+            throw new NotImplementedException("Invalid OS");
+        }
+
+        public static IList<Tuple<string, string>> EnumerateTestCases()
```
halter73: 👍

test/Kestrel.FunctionalTests/Http2/H2SpecTests.cs
Tratcher: Some of the tests deal with invalid request frame sequences. If you don't wait for the request body and immediately respond with Hello World then the client may not observe any errors and the test fails. Echoing the body was the most reliable way to wait for request errors.

test/Kestrel.FunctionalTests/Http2/H2SpecTests.cs
Tratcher: You don't expect H2spec to be able to run on alpine? 

Actually, if we can't run h2spec then the test enumeration will fail and we'll never get to the skip conditions. This will fail the test run. That seems preferable until we at least know why and where it fails.

halter73: Will the other tests still run? If so, that seems fine. Alpine uses musl. According to the [musl faq](https://www.musl-libc.org/faq.html)

> At present, some glibc-linked shared libraries can be loaded with musl, but all but the simplest glibc-linked applications will fail if musl is dropped-in in place of /lib/ld-linux.so.2.

I know our libuv binaries segfault when used on Alpine, and that's a shared library.

Tratcher: Yes I'd expect the other tests to still run unless it was a catastrophic failure.

Issue #2651, opened 2018-06-05
H2Spec ArgumentOutOfRangeException 


Tratcher:
This happens on h2spec (v2.1) test http2/4.2 and the test hangs afterwards. Repros with and without TLS.
```
crit: Microsoft.AspNetCore.Server.Kestrel[0]
      Unexpected exception in HttpConnection.ProcessRequestsAsync.
System.ArgumentOutOfRangeException: Specified argument was out of the range of valid values.
   at System.Span`1..ctor(T[] array, Int32 start, Int32 length)
   at Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http2.Http2Frame.get_Raw() in D:\github\AspNet\KestrelHttpServer\src\Kestrel.Core\Internal\Http2\Http2Frame.cs:line 22
   at Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http2.Http2FrameReader.ReadFrame(ReadOnlySequence`1 readableBuffer, Http2Frame frame, SequencePosition& consumed, SequencePosition& examined) in D:\github\AspNet\KestrelHttpServer\src\Kestrel.Core\Internal\Http2\Http2FrameReader.cs:line 23
   at Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http2.Http2Connection.ProcessRequestsAsync[TContext](IHttpApplication`1 application) in D:\github\AspNet\KestrelHttpServer\src\Kestrel.Core\Internal\Http2\Http2Connection.cs:line 160
   at Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http2.Http2Connection.ProcessRequestsAsync[TContext](IHttpApplication`1 application) in D:\github\AspNet\KestrelHttpServer\src\Kestrel.Core\Internal\Http2\Http2Connection.cs:line 221
   at Microsoft.AspNetCore.Server.Kestrel.Core.Internal.HttpConnection.ProcessRequestsAsync[TContext](IHttpApplication`1 httpApplication) in D:\github\AspNet\KestrelHttpServer\src\Kestrel.Core\Internal\HttpConnection.cs:line 185
dbug: Microsoft.AspNetCore.Server.Kestrel[2]
      Connection id "0HLEB1F2PS7B8" stopped.
```

Note this is most visible when running the full h2spec test suite from the command line against the http/2 sample app. Running the individual test cases from VS does not hang.

PR #2656, opened 2018-06-08
Enforce max frame size 

Tratcher:
#2651 The MaxFrameSize setting wasn't being enforced so instead you'd get an ArgumentOutOfRangeException for frames larger than the buffer. The h2spec test 4.2 found this, but it wasn't actually failing because the exception was still handled and terminated the connection, only with the wrong http2 error.


Tratcher:
Correct, those failures are unrelated infrastructure issues.

PR #2660, opened 2018-06-11
React to aspnet/BuildTools#684

halter73:
This should fix the travis linux builds.

aspnet/BuildTools#684

### DIFF
--- a/korebuild-lock.txt
+++ b/korebuild-lock.txt
@@ -1,2 +1,2 @@
-version:2.2.0-preview1-17081
-commithash:73f09c256e2a54270951562ecc0ef4a953926c36
+version:2.2.0-preview1-17087
+commithash:596d40f67cbe774675e73c39d463b628bf453531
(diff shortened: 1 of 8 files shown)


natemcmaster:
It looks like it's still broken.

By the way, I think a better fix is to just run `build.cmd -update` to get the latest korebuild. The $HOME changes were reverted Friday, so the install path for dotnet should still be $PWD/.dotnet.

halter73:
@natemcmaster ⬆️ 📅 . You can ignore the last commit for the most part. I am trying to reduce the number of calls to ManualResetEvent.Wait() and similar to free up threads so Task.Run doesn't take over 10 seconds to run its callback on Travis. This helps the build to get to the point where it can run the docker systemd activation tests.

natemcmaster:
Manually re-queued https://dotnet.visualstudio.com/public/public%20Team/_build/results?buildId=1439&view=logs

PR #2666, opened 2018-06-13
Prevent double-close race condition in ListenHandleTests.CanListenToOpenTcpSocketHandle()

mikeharder:
- Assign managed socket to static field to prevent disposal
- Addresses #2597

### DIFF
--- a/test/Kestrel.Transport.Libuv.FunctionalTests/ListenHandleTests.cs
+++ b/test/Kestrel.Transport.Libuv.FunctionalTests/ListenHandleTests.cs
@@ -14,26 +14,30 @@ namespace Microsoft.AspNetCore.Server.Kestrel.FunctionalTests
     [OSSkipCondition(OperatingSystems.Windows, SkipReason = "Listening to open TCP socket and/or pipe handles is not supported on Windows.")]
     public class ListenHandleTests : LoggedTest
     {
+        // The Socket.Handle will be passed into libuv, which calls close() on the file descriptor when TestServer is disposed.  If
+        // the managed Socket is also disposed or finalized, it will try to call close() again on the file descriptor, which may lead to
+        // race condition bugs (test hangs) if the file descriptor has been re-used for another resource.  In .NET Core, objects
+        // assigned to static fields should never be disposed or finalized (even at process shutdown).
+        // https://github.com/aspnet/KestrelHttpServer/issues/2597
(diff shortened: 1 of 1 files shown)


test/Kestrel.Transport.Libuv.FunctionalTests/ListenHandleTests.cs
```diff
@@ -14,26 +14,30 @@ namespace Microsoft.AspNetCore.Server.Kestrel.FunctionalTests
     [OSSkipCondition(OperatingSystems.Windows, SkipReason = "Listening to open TCP socket and/or pipe handles is not supported on Windows.")]
     public class ListenHandleTests : LoggedTest
     {
+        // The Socket.Handle will be passed into libuv, which calls close() on the file descriptor when TestServer is disposed.  If
+        // the managed Socket is also disposed or finalized, it will try to call close() again on the file descriptor, which may lead to
+        // race condition bugs (test hangs) if the file descriptor has been re-used for another resource.  In .NET Core, objects
+        // assigned to static fields should never be disposed or finalized (even at process shutdown).
+        // https://github.com/aspnet/KestrelHttpServer/issues/2597
+        private static readonly Socket _listenSocket = new Socket(AddressFamily.InterNetwork, SocketType.Stream, ProtocolType.Tcp);
```
Tratcher: So this instance is usable for only one test?

mikeharder: Yes, assigning the Socket to a static is a hacky way to prevent double-dispose.  I believe the less-hacky way would be to write our own cross-plat p-invokes to directly create a native Socket, but it's a lot more work for just test code.

Tratcher: Name the field after the test then? Or embed it in the test?

mikeharder: I renamed the field, not sure what you mean by "embed it in the test" since it needs to be a static.

mikeharder:
@halter73: Ping.  The only change is moving the Socket from a local variable to a static.  Tests ran overnight on my Linux VM with no hangs.  Would like to merge today to fix the Linux hangs in CI.

mikeharder:

mikeharder:
Issues for flaky tests during build checks:
- https://github.com/aspnet/KestrelHttpServer/issues/2668
- https://github.com/aspnet/KestrelHttpServer/issues/2669

Issue #2657, opened 2018-06-08
Application code can block http2 frame processing loop


Tratcher:
Http2 connections support multiplexing with multiple parallel streams. There's one central loop per connection that reads incoming frames and dispatches them to the correct stream. However that dispatch does not force a thread hop, so if the app code blocks synchronously then the frame loop won't continue.

This is what happens when you put a break point in Run in the sample app:

```
 	Http2SampleApp.dll!Http2SampleApp.Startup.Configure.AnonymousMethod__1_0(Microsoft.AspNetCore.Http.HttpContext context) Line 19	C#
 	Microsoft.AspNetCore.Hosting.dll!Microsoft.AspNetCore.Hosting.Internal.HostingApplication.ProcessRequestAsync(Microsoft.AspNetCore.Hosting.Internal.HostingApplication.Context context)	Unknown
 	Microsoft.AspNetCore.Server.Kestrel.Core.dll!Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http.HttpProtocol.ProcessRequests<Microsoft.AspNetCore.Hosting.Internal.HostingApplication.Context>(Microsoft.AspNetCore.Hosting.Server.IHttpApplication<Microsoft.AspNetCore.Hosting.Internal.HostingApplication.Context> application) Line 553	C#
 	Microsoft.AspNetCore.Server.Kestrel.Core.dll!Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http.HttpProtocol.ProcessRequestsAsync<Microsoft.AspNetCore.Hosting.Internal.HostingApplication.Context>(Microsoft.AspNetCore.Hosting.Server.IHttpApplication<Microsoft.AspNetCore.Hosting.Internal.HostingApplication.Context> application) Line 464	C#
 	Microsoft.AspNetCore.Server.Kestrel.Core.dll!Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http2.Http2Connection.StartStream<Microsoft.AspNetCore.Hosting.Internal.HostingApplication.Context>(Microsoft.AspNetCore.Hosting.Server.IHttpApplication<Microsoft.AspNetCore.Hosting.Internal.HostingApplication.Context> application) Line 674	C#
[...]

Issue #2462, opened 2018-04-03
Investigate HTTP/2 behavior for RequestAborted and Abort


Tratcher:
For Http/1.x RequestAborted is used to detect early disconnects and Abort is used to force them.

For HTTP/2 requests can be multiplexed over a single connection and it's possible to terminate individual requests while leaving the connection active. Investigate how kestrel surfaces this behavior and if RequestAborted and Abort can be adapted accordingly. We may also need to introduce a lower level connection Abort API to kill the underlying connection.

Mirrors https://github.com/aspnet/IISIntegration/issues/753

muratg:
Backlogging until HTTP/2 support is back.

Tratcher:
Incoming RST_STREAM frames are used to signal abort on the associated request:
https://github.com/aspnet/KestrelHttpServer/blob/ee86e97b6ba561e9c4ee13e53d44551f7e759a6a/src/Kestrel.Core/Internal/Http2/Http2Connection.cs#L471-L474
I'll do some manual verification on this. Also, make sure this doesn't call user code that can block the main connection loop.

Followup: What about Reqeuest.Abort()? Does it terminate the stream or the connection?

Tratcher:
Call stack:
Http2Connection.ProcessRstStreamFrameAsync
HttpProtocol.Abort
HttpProtocol.OnInputOrOutputCompleted (Offloads to new thread)
HttpProtocol.CancelRequestAbortedToken

Problem: This offload depends on ApplicationSchedulingMode.

Tratcher:
On connection tear down all streams have RequestAborted called. Call stack:
Http2Connection.ProcessRequestsAsync - finally block
Http2Stream.Http2Abort
Http2Stream.OnInputOrOutputCompleted
HttpProtocol.OnInputOrOutputCompleted (Offloads to new thread)
HttpProtocol.CancelRequestAbortedToken

Problem: This offload depends on ApplicationSchedulingMode. Blocking abort on one stream could prevent it from being called on others.

Tratcher:
Bug: Response headers and data are still sent after receiving a RST frame for a given stream, only the final end stream data frame is not sent. These should all be discarded. There are even (incorrect) tests for this.

Tratcher:
Bug: The request stream is not unblocked when a RST frame is received.

Tratcher:
Feature:
[...]

PR #2670, opened 2018-06-15
Prevent app code from blocking the connection loop 

Tratcher:
#2657 Each connection is multiplexed and has a loop that reads and processes frames. That processing must not ever get blocked by application code.

I fixed one code path where the initial request is dispatched to the app synchronously and the connection loop may get blocked by app code.

The only other case I found is for RST frames. https://github.com/aspnet/KestrelHttpServer/issues/2462#issuecomment-397702247. RequestAborted is dispatched on a background thread but that can be affected by ApplicationSchedulingMode. What's the guidance for ApplicationSchedulingMode? Should Inline mode assume the developer is going to write ideal code and never block the app loop? Or are there some conditions where we don't trust the app enough and always offload to a background thread like with the initial request above?

### DIFF
--- a/src/Kestrel.Core/Internal/Http2/Http2Connection.cs
+++ b/src/Kestrel.Core/Internal/Http2/Http2Connection.cs
@@ -7,6 +7,7 @@ using System.Collections.Concurrent;
 using System.IO.Pipelines;
 using System.Security.Authentication;
 using System.Text;
+using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Connections;
 using Microsoft.AspNetCore.Connections.Features;
@@ -703,7 +704,13 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http2
             }
 
             _streams[_incomingFrame.StreamId] = _currentHeadersStream;
-            _ = _currentHeadersStream.ProcessRequestsAsync(application);
+            // Must not allow app code to block the connection handling loop.
+            ThreadPool.UnsafeQueueUserWorkItem(state =>
+            {
+                var (app, currentStream) = (Tuple<IHttpApplication<TContext>, Http2Stream>)state;
+                _ = currentStream.ProcessRequestsAsync(app);
+            },
+            new Tuple<IHttpApplication<TContext>, Http2Stream>(application, _currentHeadersStream));
         }
 
         private void ResetRequestHeaderParsingState()
--- a/test/Kestrel.Core.Tests/Http2ConnectionTests.cs
+++ b/test/Kestrel.Core.Tests/Http2ConnectionTests.cs
@@ -473,6 +473,72 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Tests
             Assert.Equal(stream3DataFrame2.DataPayload, _worldBytes);
         }
 
+        [Fact]
+        public async Task DATA_Received_Multiplexed_AppMustNotBlockOtherFrames()
+        {
+            var stream1Read = new ManualResetEvent(false);
+            var stream1ReadFinished = new ManualResetEvent(false);
+            var stream3Read = new ManualResetEvent(false);
+            var stream3ReadFinished = new ManualResetEvent(false);
+            await InitializeConnectionAsync(async context =>
+            {
+                var data = new byte[10];
+                var read = await context.Request.Body.ReadAsync(new byte[10], 0, 10);
+                if (context.Features.Get<IHttp2StreamIdFeature>().StreamId == 1)
+                {
+                    stream1Read.Set();
+                    Assert.True(stream1ReadFinished.WaitOne(TimeSpan.FromSeconds(10)));
+                }
+                else
+                {
+                    stream3Read.Set();
+                    Assert.True(stream3ReadFinished.WaitOne(TimeSpan.FromSeconds(10)));
+                }
+                await context.Response.Body.WriteAsync(data, 0, read);
+            });
+
+            await StartStreamAsync(1, _browserRequestHeaders, endStream: false);
+            await StartStreamAsync(3, _browserRequestHeaders, endStream: false);
+
+            await SendDataAsync(1, _helloBytes, endStream: false);
+            Assert.True(stream1Read.WaitOne(TimeSpan.FromSeconds(10)));
+
+            await SendDataAsync(3, _helloBytes, endStream: false);
+            Assert.True(stream3Read.WaitOne(TimeSpan.FromSeconds(10)));
+
+            stream3ReadFinished.Set();
+
+            await ExpectAsync(Http2FrameType.HEADERS,
+                withLength: 37,
+                withFlags: (byte)Http2HeadersFrameFlags.END_HEADERS,
+                withStreamId: 3);
+           await ExpectAsync(Http2FrameType.DATA,
+                withLength: 5,
+                withFlags: (byte)Http2DataFrameFlags.NONE,
+                withStreamId: 3);
+            await ExpectAsync(Http2FrameType.DATA,
+                withLength: 0,
+                withFlags: (byte)Http2DataFrameFlags.END_STREAM,
+                withStreamId: 3);
+
+            stream1ReadFinished.Set();
+
+            await ExpectAsync(Http2FrameType.HEADERS,
+                withLength: 37,
+                withFlags: (byte)Http2HeadersFrameFlags.END_HEADERS,
+                withStreamId: 1);
+            await ExpectAsync(Http2FrameType.DATA,
+                withLength: 5,
+                withFlags: (byte)Http2DataFrameFlags.NONE,
+                withStreamId: 1);
+            await ExpectAsync(Http2FrameType.DATA,
+                withLength: 0,
+                withFlags: (byte)Http2DataFrameFlags.END_STREAM,
+                withStreamId: 1);
+
+            await StopConnectionAsync(expectedLastStreamId: 3, ignoreNonGoAwayFrames: false);
+        }
+
         [Theory]
         [InlineData(0)]
         [InlineData(1)]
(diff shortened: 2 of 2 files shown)


halter73:
Looks like this is causing Http2ConnectionTests.ConnectionResetLoggedWithActiveStreams to fail

halter73:
> What's the guidance for ApplicationSchedulingMode? Should Inline mode assume the developer is going to write ideal code and never block the app loop?

Yes.

src/Kestrel.Core/Internal/Http/HttpProtocol.cs
```diff
@@ -1343,7 +1343,7 @@ private Pipe CreateRequestBodyPipe()
             (
                 pool: _context.MemoryPool,
                 readerScheduler: ServiceContext.Scheduler,
-                writerScheduler: PipeScheduler.Inline,
+                writerScheduler: PipeScheduler.ThreadPool,
```
halter73: This being inline is fine for HTTP/1.1 where there's no multiplexing. We should make this virtual, and only change the HTTP/2 behavior.

halter73: Also this affects the scheduling of server code, not app code, so I don't see how this could affect whether or not the app code could block the connection loop.

src/Kestrel.Core/Internal/Http2/Http2Connection.cs
```diff
@@ -672,7 +672,15 @@ private void StartStream<TContext>(IHttpApplication<TContext> application)
...
-            _ = _currentHeadersStream.ProcessRequestsAsync(application);
+            // Must not allow app code to block the connection handling loop.
+            System.Threading.ThreadPool.UnsafeQueueUserWorkItem(state =>
+                {
+                    var tuple = (Tuple<IHttpApplication<TContext>, Http2Stream>)state;
+                    var app = tuple.Item1;
+                    var currentStream = tuple.Item2;
+                    _ = currentStream.ProcessRequestsAsync(app);
+                },
+                new Tuple<IHttpApplication<TContext>,Http2Stream>(application, _currentHeadersStream));
```
halter73: Nit: For more @pranavkm points you can shorten this to

```C#
ThreadPool.UnsafeQueueUserWorkItem(state =>
{
    var (app, currentStream) = (Tuple<IHttpApplication<TContext>, Http2Stream>)state;
    _ = currentStream.ProcessRequestsAsync(app);
},
(application, _currentHeadersStream));
```

src/Kestrel.Core/Internal/Http2/Http2Connection.cs
```diff
@@ -672,7 +672,15 @@ private void StartStream<TContext>(IHttpApplication<TContext> application)
             }
 
             _streams[_incomingFrame.StreamId] = _currentHeadersStream;
-            _ = _currentHeadersStream.ProcessRequestsAsync(application);
+            // Must not allow app code to block the connection handling loop.
+            System.Threading.ThreadPool.UnsafeQueueUserWorkItem(state =>
```
halter73: Nit: Add using.

src/Kestrel.Core/Internal/Http2/Http2Connection.cs
```diff
@@ -672,7 +672,15 @@ private void StartStream<TContext>(IHttpApplication<TContext> application)
...
+                {
```
halter73: Nit: In Kestrel we consistently use the same level of indentation for the `{` after a `=>`.

test/Kestrel.Core.Tests/Http2ConnectionTests.cs
```diff
@@ -131,8 +131,8 @@ public Http2ConnectionTests()
         {
             var inlineSchedulingPipeOptions = new PipeOptions(
                 pool: _memoryPool,
-                readerScheduler: PipeScheduler.Inline,
-                writerScheduler: PipeScheduler.Inline,
+                readerScheduler: PipeScheduler.ThreadPool,
+                writerScheduler: PipeScheduler.ThreadPool,
```
halter73: This is probably responsible for the ConnectionResetLoggedWithActiveStreams failure. I don't think you should change this. If for some reason it's needed for a new test, use `threadPoolSchedulingPipeOptions` for just the new test.

test/Kestrel.Core.Tests/Http2ConnectionTests.cs
```diff
@@ -641,7 +707,15 @@ void IHttpHeadersHandler.OnHeader(Span<byte> name, Span<byte> value)
                 ignoreNonGoAwayFrames: false,
                 expectedLastStreamId: 1,
                 expectedErrorCode: Http2ErrorCode.STREAM_CLOSED,
-                expectedErrorMessage: CoreStrings.FormatHttp2ErrorStreamClosed(Http2FrameType.DATA, streamId: 1));
+                expectedErrorMessage: null);
+
+            // There's a race where either of these messages could be logged, depending on if the stream cleanup has finished yet.
```
halter73: Was this race there before or just after changing the scheduling mode? If it's the latter, revert this test change.

src/Kestrel.Core/Internal/Http2/Http2Connection.cs
gfoidl: With `UnsafeQueueUserWorkItem` the state-argument (valuetuple) gets boxed. This allocation can be avoided with the generic `QueueUserWorkItem`-overload:
```c#
ThreadPool.QueueUserWorkItem(state =>
{
    var (app, currentStream) = state;
    _ = currentStream.ProcessRequestsAsync(app);
},
(application, _currentHeadersStream), false);
```
Or is the "unsafe queuing" needed here?

Further the allocation for the delegate could be avoided by caching it.

Tratcher: Unsafe is more efficient as it avoids capturing the current thread context.

Tratcher: I'm not able to fully extract the delegate due to the method's generic type.

benaadams: > Unsafe is more efficient as it avoids capturing the current thread context.

Can always do this instead for a similar effect
```csharp
bool restoreFlow = false;
try
{
    if (!ExecutionContext.IsFlowSuppressed())
    {
        ExecutionContext.SuppressFlow();
        restoreFlow = true;
    }

    ThreadPool.QueueUserWorkItem(state =>
    {
        var (app, currentStream) = state;
            _ = currentStream.ProcessRequestsAsync(app);
    },
    (application, _currentHeadersStream), false);
}
finally
{
    // Restore the current ExecutionContext
    if (restoreFlow)
        ExecutionContext.RestoreFlow();
}
```
As per https://github.com/dotnet/corefx/pull/26077

Tratcher: What's the advantage to doing that over UnsafeQueueUserWorkItem?

benaadams: Depends if you can use the generic state overload (on netcoreapp2.1) to avoid the boxing of the state
```csharp
QueueUserWorkItem<TState>(Action<TState> callBack, TState state, bool preferLocal)
```

benaadams: Also is an `Action<T>` callback rather than a `WaitCallback`

Tratcher: We're trying to limit our use of netcoreapp2.1 APIs to only places where we can't avoid it. If this shows up in the profiles then we may try adding a fast-path.

benaadams: Is no advantage vs `UnsafeQueueUserWorkItem` on netstandard2.0

gfoidl: Besides "limit our use of netcoreapp2.1 APIs" would it make sense to make `UnsafeQueueUserWorkItem` generic similar to https://github.com/dotnet/coreclr/pull/16570?

halter73: I was suggesting the new syntax for terseness not for valuetuple. We don't want to ifdef to use netcoreapp2.1 APIs unless theres a functional improvement or significant perf improvement. Explicitly call ExecutionContext.Suppress/RestoreFlow is also pretty gross.

src/Kestrel.Core/Internal/Http2/Http2Connection.cs
```diff
@@ -672,7 +672,13 @@ private void StartStream<TContext>(IHttpApplication<TContext> application)
...
 
             _streams[_incomingFrame.StreamId] = _currentHeadersStream;
-            _ = _currentHeadersStream.ProcessRequestsAsync(application);
+            // Must not allow app code to block the connection handling loop.
+            ThreadPool.UnsafeQueueUserWorkItem(state =>
+            {
+                var (app, currentStream) = (Tuple<IHttpApplication<TContext>, Http2Stream>)state;
+                _ = currentStream.ProcessRequestsAsync(app);
+            },
+            new Tuple<IHttpApplication<TContext>,Http2Stream>(application, _currentHeadersStream));
```
halter73: Nit: missing space after `>,`.

Also if we really want to avoid an allocation here, we could just add a reference to the application in Http2Stream.

halter73:
It looks like dispatching might be causing some [flakiness](https://dotnet.visualstudio.com/public/public%20Team/_build/results?buildId=2362&view=logs) with Http2ConnectionTests.HEADERS_Received_StreamClosed_ConnectionError.

```
2018-06-18T19:36:23.5227430Z   [xUnit.net 00:00:05.9252730]     HEADERS_Received_StreamClosed_ConnectionError [FAIL]
2018-06-18T19:36:23.5244610Z   Failed   HEADERS_Received_StreamClosed_ConnectionError
2018-06-18T19:36:23.5368000Z RUNDOTNET : error Message:  [/Users/vsts/agent/2.134.2/work/.dotnet/buildtools/korebuild/2.2.0-preview1-17087/KoreBuild.proj]
2018-06-18T19:36:23.5385250Z    Assert.Contains() Failure
2018-06-18T19:36:23.5400530Z   Not found: The client sent a HEADERS frame to closed stream ID 1.
2018-06-18T19:36:23.5418860Z In value : HTTP/2 connection error (STREAM_CLOSED): The client sent a HEADERS frame to stream ID 1 which is in the "half-closed (remote) state". [/Users/vsts/agent/2.134.2/work/.dotnet/buildtools/korebuild/2.2.0-preview1-17087/KoreBuild.proj]
2018-06-18T19:36:23.5436600Z   Stack Trace:
2018-06-18T19:36:23.5452880Z      at Microsoft.AspNetCore.Server.Kestrel.Core.Tests.Http2ConnectionTests.WaitForConnectionErrorAsync[TException](Boolean ignoreNonGoAwayFrames, Int32 expectedLastStreamId, Http2ErrorCode expectedErrorCode, String expectedErrorMessage)
2018-06-18T19:36:23.5469770Z      at Microsoft.AspNetCore.Server.Kestrel.Core.Tests.Http2ConnectionTests.HEADERS_Received_StreamClosed_ConnectionError() in /_/test/Kestrel.Core.Tests/Http2ConnectionTests.cs:line 1014
2018-06-18T19:36:23.5487510Z   --- End of stack trace from previous location where exception was thrown ---
```

src/Kestrel.Core/Internal/Http2/Http2Connection.cs
Tratcher: That generic is very viral. Right now it's only passed through method signatures so it can't be stored in any member fields.

halter73: You could always cast from object in the callback, but I wouldn't bother.

gfoidl: > Nit: missing space after `>,`.

Reminder :wink:

PR #2675, opened 2018-06-19
Add TLS filter sample

Tratcher:
This is the last part of HTTP/2 TLS restrictions #2251 
- SHOULD NOT use any of the cipher suites that are listed in the cipher suite black list

Since this is a SHOULD we don't want to add a hardcoded prohibition for it. The list is also 12 pages long and would be expensive to check on each request. Users are better off configuring their TLS provider to refuse these ciphers, but if they need to check for some specific ones they can do that as shown in this sample. 

@blowdart does this look workable?

### DIFF
--- a/samples/Http2SampleApp/Program.cs
+++ b/samples/Http2SampleApp/Program.cs
@@ -1,8 +1,12 @@
 using System;
 using System.IO;
 using System.Net;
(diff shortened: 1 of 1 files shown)


blowdart:
Yup, good choice for the example too :)

PR #2686, opened 2018-06-26
Receive and send RST frames

Tratcher:
#2462 @halter73 I still need to finish the logging and error messages, but I wanted to show you the concept first.
Once the client has sent us a reset for a given stream we should stop sending headers or data on that stream. The request continues normally and OnStarting still triggers, you only see the error if you check RequestAborted or pass a cancellable token to WriteAsync.

### DIFF
--- a/src/Kestrel.Core/Internal/Http2/Http2OutputProducer.cs
+++ b/src/Kestrel.Core/Internal/Http2/Http2OutputProducer.cs
@@ -73,7 +73,6 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http2
 
         public void Abort(ConnectionAbortedException abortReason)
         {
-            // TODO: RST_STREAM?
             Dispose();
         }
 
@@ -128,13 +127,13 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http2
         {
             lock (_dataWriterLock)
             {
+                // The HPACK header compressor is stateful, if we compress headers for an aborted stream we must send them.
+                // Optimize for not compressing or sending them.
                 if (_completed)
                 {
                     return;
                 }
 
-                // The HPACK header compressor is stateful, if we compress headers for an aborted stream we must send them.
-                // Optimize for not compressing or sending them.
                 _frameWriter.WriteResponseHeaders(_streamId, statusCode, responseHeaders);
             }
         }
@@ -178,6 +177,18 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http2
 
                 _dataPipe.Writer.Complete();
                 return _dataWriteProcessingTask;
+            }
+        }
+
+        public Task WriteRstStreamAsync(Http2ErrorCode error)
+        {
+            lock (_dataWriterLock)
+            {
+                // Always send the reset even if the response body is _completed. The request body may not have completed yet.
+
+                Dispose();
+
+                return _frameWriter.WriteRstStreamAsync(_streamId, error);
             }
         }
 
--- a/src/Kestrel.Core/Internal/Http2/Http2Stream.cs
+++ b/src/Kestrel.Core/Internal/Http2/Http2Stream.cs
@@ -5,6 +5,7 @@ using System;
 using System.Buffers;
 using System.IO;
 using System.IO.Pipelines;
+using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Connections;
 using Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http;
@@ -16,15 +17,17 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http2
     public partial class Http2Stream : HttpProtocol
     {
         private readonly Http2StreamContext _context;
+        private readonly Http2OutputProducer _http2Output;
         private readonly Http2StreamOutputFlowControl _outputFlowControl;
+        private int _requestAborted;
 
         public Http2Stream(Http2StreamContext context)
             : base(context)
         {
             _context = context;
             _outputFlowControl = new Http2StreamOutputFlowControl(context.ConnectionOutputFlowControl, context.ClientPeerSettings.InitialWindowSize);
-
-            Output = new Http2OutputProducer(context.StreamId, context.FrameWriter, _outputFlowControl, context.TimeoutControl, context.MemoryPool);
+            _http2Output = new Http2OutputProducer(context.StreamId, context.FrameWriter, _outputFlowControl, context.TimeoutControl, context.MemoryPool);
+            Output = _http2Output;
         }
 
         public int StreamId => _context.StreamId;
@@ -144,17 +147,47 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http2
             }
         }
 
+        public bool TryUpdateOutputWindow(int bytes)
+        {
+            return _context.FrameWriter.TryUpdateStreamWindow(_outputFlowControl, bytes);
+        }
+
         public override void Abort(ConnectionAbortedException abortReason)
+        {
+            if (Interlocked.Exchange(ref _requestAborted, 1) != 0)
+            {
+                return;
+            }
+
+            AbortCore(abortReason);
+        }
+
+        protected override void ApplicationAbort()
+        {
+            Log.ApplicationAbortedConnection(ConnectionId, TraceIdentifier);
+            var abortReason = new ConnectionAbortedException(CoreStrings.ConnectionAbortedByApplication);
+            ResetAndAbort(abortReason, Http2ErrorCode.CANCEL);
+        }
+
+        private void ResetAndAbort(ConnectionAbortedException abortReason, Http2ErrorCode error)
+        {
+            if (Interlocked.Exchange(ref _requestAborted, 1) != 0)
+            {
+                return;
+            }
+
+            // Don't block on IO. This never faults.
+            _ = _http2Output.WriteRstStreamAsync(error);
+
+            AbortCore(abortReason);
+        }
+
+        private void AbortCore(ConnectionAbortedException abortReason)
         {
             base.Abort(abortReason);
 
             // Unblock the request body.
             RequestBodyPipe.Writer.Complete(new IOException(CoreStrings.Http2StreamAborted, abortReason));
-        }
-
-        public bool TryUpdateOutputWindow(int bytes)
-        {
-            return _context.FrameWriter.TryUpdateStreamWindow(_outputFlowControl, bytes);
         }
     }
 }
--- a/test/Kestrel.Core.Tests/Http2ConnectionTests.cs
+++ b/test/Kestrel.Core.Tests/Http2ConnectionTests.cs
@@ -1892,12 +1892,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Tests
             await WaitForAllStreamsAsync();
             Assert.Contains(1, _abortedStreamIds);
 
-            await SendGoAwayAsync();
-
-            // No data is received from the stream since it was aborted before writing anything
-            await WaitForConnectionStopAsync(expectedLastStreamId: 1, ignoreNonGoAwayFrames: false);
-
-            // TODO: Check logs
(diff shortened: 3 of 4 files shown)


Tratcher:
@halter73 getting the app to send RST when it calls Abort wasn't as bad as I thought. This should be additive on top of your changes.

Tratcher:
Updated: This finishes off receiving and sending RST frames.

test/Kestrel.Core.Tests/Http2ConnectionTests.cs
```diff
@@ -2312,6 +2306,104 @@ public Task HEADERS_Received_HeaderBlockContainsTEHeader_ValueIsNotTrailers_Stre
...
+                    _runningStreams[streamIdFeature.StreamId].TrySetException(ex);
+                }
+            });
+
+            await StartStreamAsync(1, _browserRequestHeaders, endStream: true);
+            await WaitForStreamErrorAsync(expectedStreamId: 1, Http2ErrorCode.CANCEL, expectedErrorMessage: null);
+            await WaitForAllStreamsAsync();
+            Assert.Contains(1, _abortedStreamIds);
+
+            await SendGoAwayAsync();
```
halter73: Why not just StopConnectionAsync? That should work, right?

test/Kestrel.Core.Tests/Http2ConnectionTests.cs
```diff
@@ -2312,6 +2306,104 @@ public Task HEADERS_Received_HeaderBlockContainsTEHeader_ValueIsNotTrailers_Stre
...
+            await ExpectAsync(Http2FrameType.DATA,
+                withLength: 10,
+                withFlags: 0,
+                withStreamId: 1);
+
+            await WaitForStreamErrorAsync(expectedStreamId: 1, Http2ErrorCode.CANCEL, expectedErrorMessage: null);
+            await WaitForAllStreamsAsync();
+            Assert.Contains(1, _abortedStreamIds);
+
+            await SendGoAwayAsync();
```
halter73: StopConnectionAsync

test/Kestrel.Core.Tests/Http2ConnectionTests.cs
```diff
@@ -1932,8 +1928,6 @@ public Task HEADERS_Received_HeaderBlockContainsTEHeader_ValueIsNotTrailers_Stre
 
             // No END_STREAM HEADERS or DATA frame is received since the stream was aborted
             await WaitForConnectionStopAsync(expectedLastStreamId: 1, ignoreNonGoAwayFrames: false);
-
-            // TODO: Check logs
```
halter73: I noticed these tests call `SendGoAwayAsync` too. Prior to these being added, only the `GOAWAY_` tests called this helper. Any other test that wanted to close the stream just closed the pipe. I think we should keep this consistent.

src/Kestrel.Core/Internal/Http2/Http2Stream.cs
```diff
@@ -144,17 +146,51 @@ protected override bool TryParseRequest(ReadResult result, out bool endConnectio
...
-        public bool TryUpdateOutputWindow(int bytes)
+        protected override void ApplicationAbort()
         {
-            return _context.FrameWriter.TryUpdateStreamWindow(_outputFlowControl, bytes);
+            Log.ApplicationAbortedConnection(ConnectionId, TraceIdentifier);
+            var abortReason = new ConnectionAbortedException(CoreStrings.ConnectionAbortedByApplication);
+            AbortAndReset(abortReason, Http2ErrorCode.CANCEL);
+        }
+
+        public void AbortAndReset(ConnectionAbortedException abortReason, Http2ErrorCode error)
```
halter73: private

src/Kestrel.Core/Internal/Http2/Http2OutputProducer.cs
```diff
@@ -46,6 +46,8 @@ public class Http2OutputProducer : IHttpOutputProducer
             _dataWriteProcessingTask = ProcessDataWrites();
         }
 
+        public bool IsCompleted => _completed;
```
halter73: It's not safe to check `_completed` outside the lock. I see that the way it's being used now is basically an optimization, but `HttpContext.Abort()` is not called frequently enough for this to be worth it. Remove this property.

src/Kestrel.Core/Internal/Http2/Http2Stream.cs
```diff
@@ -144,17 +146,51 @@ protected override bool TryParseRequest(ReadResult result, out bool endConnectio
...
+        {
+            if (Interlocked.Exchange(ref _requestAborted, 1) != 0)
+            {
+                return;
+            }
+
+            var outputCompleted = Output.IsCompleted;
```
halter73: Just put WriteRstStreamAsync on the OutputProducer to avoid any potential race conditions.

test/Kestrel.Core.Tests/Http2ConnectionTests.cs
Tratcher: GO_AWAY is relevant to the RST tests since it calls the Abort code path.

test/Kestrel.Core.Tests/Http2ConnectionTests.cs
```diff
@@ -529,10 +529,10 @@ void IHttpHeadersHandler.OnHeader(Span<byte> name, Span<byte> value)
             await StartStreamAsync(1, _browserRequestHeaders, endStream: false);
             await StartStreamAsync(3, _browserRequestHeaders, endStream: false);
 
-            await SendDataAsync(1, _helloBytes, endStream: false);
+            await SendDataAsync(1, _helloBytes, endStream: true);
             Assert.True(stream1Read.WaitOne(TimeSpan.FromSeconds(10)));
 
-            await SendDataAsync(3, _helloBytes, endStream: false);
+            await SendDataAsync(3, _helloBytes, endStream: true);
```
halter73: Was this change necessary for the test to pass?

Tratcher: Leftover from a prior version, reverted.

test/Kestrel.Core.Tests/Http2ConnectionTests.cs
halter73: I don't buy it. If it was relevant, these tests wouldn't still pass when removing the SendGoAwayAsync() call. 

The ProcessGoAwayFrameAsync() just cancels the pending read on the Http2Connection's Input PipeReader stopping the ProcessRequestsAsync() loop. Completing the PipeWriter on the other side via StopConnectionAsync() also stops the loop and aborts all the streams.

Tratcher:
Updated

src/Kestrel.Core/Internal/Http2/Http2OutputProducer.cs
```diff
@@ -181,6 +180,22 @@ public Task WriteStreamSuffixAsync()
             }
         }
 
+        public Task WriteRstStreamAsync(bool endReceived, Http2ErrorCode error)
+        {
+            lock (_dataWriterLock)
+            {
+                // Only reset if the connection is open or in a half closed state.
+                if (endReceived && _completed)
```
halter73: I don't like evaluating endReceived in the lock. It really implies that Http2Stream.EndStreamReceived is protected by this lock, but it's not.

src/Kestrel.Core/Internal/Http2/Http2OutputProducer.cs
```diff
@@ -181,6 +180,22 @@ public Task WriteStreamSuffixAsync()
...
+                {
+                    return Task.CompletedTask;
+                }
+
+                _completed = true;
```
halter73: If you set `_completed` to true, you're responsible for completing the data PipeWriter. I would probably just call Dispose() here.

src/Kestrel.Core/Internal/Http2/Http2Stream.cs
```diff
@@ -144,17 +147,45 @@ protected override bool TryParseRequest(ReadResult result, out bool endConnectio
...
 
             // Unblock the request body.
             RequestBodyPipe.Writer.Complete(new IOException(CoreStrings.Http2StreamAborted, abortReason));
         }
 
-        public bool TryUpdateOutputWindow(int bytes)
+        protected override void ApplicationAbort()
         {
-            return _context.FrameWriter.TryUpdateStreamWindow(_outputFlowControl, bytes);
+            Log.ApplicationAbortedConnection(ConnectionId, TraceIdentifier);
```
halter73: Should this be logged if the stream was already aborted? Why isn't this combined with ResetAndAbort()?

src/Kestrel.Core/Internal/Http2/Http2Stream.cs
```diff
@@ -144,17 +147,45 @@ protected override bool TryParseRequest(ReadResult result, out bool endConnectio
...
+
+        private void ResetAndAbort(ConnectionAbortedException abortReason, Http2ErrorCode error)
+        {
+            if (Interlocked.Exchange(ref _requestAborted, 1) != 0)
+            {
+                return;
+            }
+
+            // If this was an app initiated Abort then we need to inform the client.
+            _http2Output.WriteRstStreamAsync(endReceived: EndStreamReceived, error).GetAwaiter().GetResult();
```
halter73: Having `HttpContext.Abort()` block on I/O for any reason is new. I don't think we should do it. Fire and forget is better. This write can't fail since no cancellation token is passed in anyway.

src/Kestrel.Core/Internal/Http2/Http2Stream.cs
```diff
@@ -144,17 +147,45 @@ protected override bool TryParseRequest(ReadResult result, out bool endConnectio
...
+
+            base.Abort(abortReason);
+
+            // Unblock the request body.
+            RequestBodyPipe.Writer.Complete(new IOException(CoreStrings.Http2StreamAborted, abortReason));
```
halter73: Maybe put the above 2 lines in a shared AbortCore() method.

src/Kestrel.Core/Internal/Http2/Http2Stream.cs
Tratcher: I separated ResetAndAbort because I expect it to be called for other conditions like my next PR for invalid request data. There will be different log statements.

halter73: I'd combine the methods until there's another caller, but no big deal.

I'm not sure I get the point about there being different log statements. I'm just pointing out it seems wrong to log that app aborted the stream if the stream was already aborted.

Tratcher: True, ApplicationAbortedConnection could be logged multiple times, but only for calls to HttpContext.Abort, not for any other reason. That's actually the same as it used to be if you look at the base class (HTTP/1.1) behavior we're overriding. Redundant but harmless.

src/Kestrel.Core/Internal/Http2/Http2OutputProducer.cs
halter73: @Tratcher pointed out this is an `&&` not `||` offline, so now I see why it was done this way. We decided to send the RST regardless since that's allowed by spec. The server/app shouldn't be likely to try to abort a dead stream anyway.

halter73: Just add a comment saying that it was intentional that we don't take the lock or check completed in this method.

src/Kestrel.Core/Internal/Http2/Http2Stream.cs
halter73: OK

Tratcher:
Updated.

src/Kestrel.Core/Internal/Http2/Http2Stream.cs
```diff
@@ -144,17 +147,46 @@ protected override bool TryParseRequest(ReadResult result, out bool endConnectio
...
+        }
+
+        private void ResetAndAbort(ConnectionAbortedException abortReason, Http2ErrorCode error)
+        {
+            if (Interlocked.Exchange(ref _requestAborted, 1) != 0)
+            {
+                return;
+            }
+
+            var ignored = _http2Output.WriteRstStreamAsync(error);
```
halter73: Nit: `_ =`

Maybe comment saying the task can't fault.

PR #2690, opened 2018-06-28
Implement HTTP/2 output flow control

halter73:
I plan on doing input flow control (i.e. sending window update frames) in a separate PR since the work is pretty independent.


src/Kestrel.Core/Internal/Http2/Http2Connection.cs
```diff
@@ -533,7 +537,28 @@ private Task ProcessSettingsFrameAsync()
 
             try
             {
+                // ParseFrame will not parse an InitialWindowSize > int.MaxValue.
+                var previousInitialWindowSize = (int)_clientSettings.InitialWindowSize;
+
                 _clientSettings.ParseFrame(_incomingFrame);
```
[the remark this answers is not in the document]
halter73: No. This is the only place _clientSettings are even read. Any future reading would have to also happen on the frame processing loop.

src/Kestrel.Core/Internal/Http/HttpProtocol.cs
```diff
@@ -966,7 +966,7 @@ public void ProduceContinue()
                 RequestHeaders.TryGetValue("Expect", out var expect) &&
                 (expect.FirstOrDefault() ?? "").Equals("100-continue", StringComparison.OrdinalIgnoreCase))
             {
-                Output.Write100ContinueAsync(default(CancellationToken)).GetAwaiter().GetResult();
+                Output.Write100ContinueAsync().GetAwaiter().GetResult();
```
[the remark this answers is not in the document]
halter73: Was that way before. I agree, but doesn't need to be part of this PR.

src/Kestrel.Core/Internal/Http2/Http2Connection.cs
```diff
@@ -620,6 +645,25 @@ private Task ProcessWindowUpdateFrameAsync()
...
+            else if (_streams.TryGetValue(_incomingFrame.StreamId, out var stream))
+            {
+                if (!stream.TryUpdateOutputWindow(_incomingFrame.WindowUpdateSizeIncrement))
+                {
+                    throw new Http2StreamErrorException(_incomingFrame.StreamId, "TODO", Http2ErrorCode.FLOW_CONTROL_ERROR);
+                }
+            }
+            else
+            {
+                throw new Http2ConnectionErrorException(CoreStrings.FormatHttp2ErrorStreamClosed(_incomingFrame.Type, _incomingFrame.StreamId), Http2ErrorCode.STREAM_CLOSED);
```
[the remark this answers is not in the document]
halter73: Then it would only be half closed.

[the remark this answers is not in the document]
halter73: We talked about this offline. The client can still send window updates for an already half closed stream. Then once the app buffers all of it's output, the stream will become fully closed and removed from the dictionary even though a well behaving client could still be sending window updates. It's better just to ignore the update.

Tratcher:
Switch branches?

src/Kestrel.Core/Internal/Http2/Http2Connection.cs
```diff
@@ -620,6 +648,21 @@ private Task ProcessWindowUpdateFrameAsync()
...
+                    throw new Http2ConnectionErrorException(CoreStrings.Http2ErrorWindowUpdateSizeInvalid, Http2ErrorCode.FLOW_CONTROL_ERROR);
+                }
+            }
+            else if (_streams.TryGetValue(_incomingFrame.StreamId, out var stream))
+            {
+                if (!stream.TryUpdateOutputWindow(_incomingFrame.WindowUpdateSizeIncrement))
+                {
+                    throw new Http2StreamErrorException(_incomingFrame.StreamId, CoreStrings.Http2ErrorWindowUpdateSizeInvalid, Http2ErrorCode.FLOW_CONTROL_ERROR);
+                }
+            }
```
Tratcher: At least leave a comment that window updates for closed streams are ignored due to a race condition between the client receiving all the data an the steam closing.

src/Kestrel.Core/Internal/Http2/Http2FlowControl.cs
```diff
@@ -0,0 +1,86 @@
...
+            {
+                return false;
+            }
+
+            Available += bytes;
+
+            while (Available > 0 && _awaitableQueue.Count > 0)
+            {
+                var awaitable = _awaitableQueue.Dequeue();
+                awaitable.Complete();
```
Tratcher: Where are these callbacks? Do they offload? We never want to run response code on the request thread, even if it's server pumping code.

src/Kestrel.Core/Internal/Http2/Http2Stream.cs
```diff
@@ -143,15 +144,17 @@ protected override bool TryParseRequest(ReadResult result, out bool endConnectio
...
-        // OperationCanceledException. This is probably not a good idea.
-        public void Http2Abort(Exception abortReason)
+        public override void Abort(ConnectionAbortedException abortReason)
         {
-            _streams?.Abort(abortReason);
+            base.Abort(abortReason);
 
-            OnInputOrOutputCompleted();
+            // Unblock the request body.
+            RequestBodyPipe.Writer.Complete(new IOException("The request stream was aborted.", abortReason));
```
Tratcher: Resource string

src/Kestrel.Core/Internal/Infrastructure/SafePipeWriterFlusher.cs
```diff
@@ -0,0 +1,76 @@
...
+using System;
+using System.IO.Pipelines;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Connections;
+using Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http;
+
+namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Infrastructure
+{
+    public class SafePipeWriterFlusher
```
Tratcher: Safe from what? ThreadSafe?

src/Kestrel.Core/Internal/Http2/Http2FrameWriter.cs
```diff
@@ -109,51 +117,132 @@ public void WriteResponseHeaders(int streamId, int statusCode, IHeaderDictionary
...
-                    tasks.Add(WriteAsync(_outgoingFrame.Raw, cancellationToken));
+                // Zero-length data frames are allowed to be sent immediately even if there is no space available in the flow control window.
+                // https://httpwg.org/specs/rfc7540.html#rfc.section.6.9.1
+                if (dataLength != 0 && dataLength > flowControl.Available)
+                {
+                    return WriteDataAsyncAwaited(streamId, flowControl, data, dataLength, endStream);
                 }
 
-                _outgoingFrame.Length = data.Length;
+                flowControl.Advance(dataLength, out _);
```
Tratcher: Does this warrant an overload that verifies these assumptions internally? E.g. that does not a return an awaitable and Asserts Available?

src/Kestrel.Core/Internal/Http2/Http2FrameWriter.cs
```diff
@@ -109,51 +117,132 @@ public void WriteResponseHeaders(int streamId, int statusCode, IHeaderDictionary
...
+                {
+                    if (flowControl.IsAborted)
+                    {
+                        break;
+                    }
+
+                    var actual = flowControl.Advance(dataLength, out availabilityAwaitable);
 
-                data.CopyTo(_outgoingFrame.Payload);
+                    if (actual > 0)
```
Tratcher: I expect we'll want some minimum here greater than 0 to keep from writing one byte at a time.

src/Kestrel.Core/Internal/Http2/Http2StreamFlowControl.cs
```diff
@@ -0,0 +1,83 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Diagnostics;
+
+namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http2
+{
+    public class Http2StreamFlowControl
```
Tratcher: This looks like it's now response flow control specific and you'll need a different API for request flow control. Rename accordingly?

src/Kestrel.Core/Internal/Http2/Http2FrameWriter.cs
Tratcher: E.g. ResumeWriterThreshold

src/Kestrel.Core/Internal/Http2/Http2FlowControl.cs
halter73: We dispatch using ThreadPoolAwaitable before invoking user code.

I think we do want the response pumping code running in the request loop. It's just copying data into a PipeWriter before either ThreadPool dispatching to user code or awaiting again for the next window update which won't happen until more frames are processed from the request loop.

It also reduces contention with many other writes to the connection-level PipeWriter that happen in the same request loop while avoid the context switching cost in the cases when there's no need to dispatch to user code.

If you consider libuv threads, they do more or less that same thing when paired with multiple http/1.1 connections.

src/Kestrel.Core/Internal/Infrastructure/SafePipeWriterFlusher.cs
halter73: It is ThreadSafe, but that's only part of it. Unlike calling PipeWriter.FlushAync directly, you can have multiple callers await the same flush task. The following written directly agains a PipeWriter would throw an invalid operation exception whenever there was actually backpressure:

```c#
await Task.WhenAll(_writer.FlushAsync().AsTask(), _writer.FlushAsync.AsTask())
```

The same code written against SafePipeWriterFlusher would work. I can't really think of a great reason anyone would want to have multiple awaiters Response.Body.FlushAsync(), but it's something we support for HTTP/1.1 so I copied the code to make that work and reused it.

I'll call it StreamSafePipeFlusher to indicate that it's safe to use to implement Stream.FlushAsync.

src/Kestrel.Core/Internal/Http2/Http2Connection.cs
halter73: Done.

natemcmaster:
My automation detected this was targeting the dev branch and updated the PR to target master. We'll be deleting dev branches soon so please use master instead.

src/Kestrel.Core/Internal/Http/HttpProtocol.cs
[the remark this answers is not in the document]
Tratcher: https://github.com/aspnet/KestrelHttpServer/issues/2700

src/Kestrel.Core/Internal/Http2/Http2FrameWriter.cs
```diff
@@ -109,51 +117,133 @@ public void WriteResponseHeaders(int streamId, int statusCode, IHeaderDictionary
...
                 {
-                    data.Slice(0, _outgoingFrame.Length).CopyTo(_outgoingFrame.Payload);
-                    data = data.Slice(_outgoingFrame.Length);
+                    return Task.CompletedTask;
+                }
 
-                    tasks.Add(WriteAsync(_outgoingFrame.Raw, cancellationToken));
+                // Zero-length data frames are allowed to be sent immediately even if there is no space available in the flow control window.
+                // https://httpwg.org/specs/rfc7540.html#rfc.section.6.9.1
+                if (dataLength != 0 && dataLength > flowControl.Available)
```
Tratcher: Could this accidentally cause a zero length End-of-stream frame to be sent ahead of queued data in its own stream? Or do we guarantee at a higher layer only one active write per stream, even for the terminator?

halter73: There's only one write per stream being the reader of the per-stream response pipe.

src/Kestrel.Core/Internal/Http2/Http2FrameWriter.cs
```diff
@@ -109,51 +117,133 @@ public void WriteResponseHeaders(int streamId, int statusCode, IHeaderDictionary
...
+        private async Task WriteDataAsyncAwaited(int streamId, Http2StreamOutputFlowControl flowControl, ReadOnlySequence<byte> data, long dataLength, bool endStream)
+        {
+            while (dataLength > 0)
+            {
+                Http2OutputFlowControlAwaitable availabilityAwaitable;
+                var writeTask = Task.CompletedTask;
+
+                lock (_writeLock)
+                {
+                    if (flowControl.IsAborted)
```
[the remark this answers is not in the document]
halter73: The stream might be aborted even though the framewriter/connection isn't.

test/Kestrel.Core.Tests/Http2ConnectionTests.cs
```diff
@@ -2190,6 +2212,88 @@ public Task HEADERS_Received_HeaderBlockContainsTEHeader_ValueIsNotTrailers_Stre
             Assert.Contains(5, _abortedStreamIds);
         }
 
+        [Fact]
+        public async Task GOAWAY_Received_RelievesConnectionBackpressure()
```
Tratcher: While this is the current behavior, it's not the expected behavior. See https://github.com/aspnet/KestrelHttpServer/issues/2703. A connection can behave normally after a GoAway is received up until the connection is prematurely closed by the client. We may opt to differentiate by GoAway error code and terminate early for some of them.

src/Kestrel.Core/Internal/Infrastructure/StreamSafePipeFlusher.cs
```diff
@@ -0,0 +1,80 @@
...
+        {
+            _timeoutControl.StartTimingWrite(count);
+
+            try
+            {
+                await _lastFlushTask;
+            }
+            catch (OperationCanceledException ex)
+            {
+                outputProducer.Abort(new ConnectionAbortedException(CoreStrings.ConnectionAbortedByCancellationToken, ex));
```
Tratcher: For Http/2 it's the stream being aborted, correct? Not the whole connection?

test/Kestrel.Core.Tests/Http2ConnectionTests.cs
halter73: I think we just update this test when we change the GOAWAY behavior. I think this test might be fine since the client sends a GOAWAY frame with a Last-Stream-ID of 0 effectively aborting all three streams.

Issue #2684, opened 2018-06-25
Lowercase logic modifies non-letter characters in header names


JunTaoLuo:
The current logic applies the lowercase mask: https://github.com/aspnet/KestrelHttpServer/blob/dev/src/Kestrel.Core/Internal/Http2/HPack/HPackEncoder.cs#L144. But this logic does not check whether the character is a letter and can change non-letter symbols. For example, it would change `^` to `~` in header names. The fix would be to ensure that the lower case mask is only applied for upper case letters. We can also consider checking for invalid characters in the header here as well.

Tratcher:
The response header character set is validated as the headers are added to the collection. Are there any allowed characters being altered? E.g. `-`?

PR #2707, opened 2018-07-09
Consolidate HTTP charset validation logic

JunTaoLuo:
Addresses https://github.com/aspnet/KestrelHttpServer/issues/2684. The only functional change is to restrict the character set accepted for header names to `Token` characters. I'll add some tests to test for this difference.


src/Kestrel.Core/Internal/Infrastructure/HttpCharacters.cs
```diff
@@ -0,0 +1,105 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+
+namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Infrastructure
+{
+    public static class HttpCharacters
+    {
+        public static readonly int TableSize = 128;
+        public static readonly bool[] AlphaNumeric;
```
nil4: Exposing public arrays lets consumers mutate their contents. 
Could they be exposed as `ReadOnlySpan<bool>` instead?h

JunTaoLuo: Actually, these probably shouldn't be exposed as public. In fact, these definitions probably belong to HttpAbstractions. However, since it's only being used in Kestrel for now, I'll set the accessibility to internal.

halter73:
Can you run the RequestParsing micro benchmark?

JunTaoLuo:
Before:

src/Kestrel.Core/Internal/Http/HttpHeaders.cs
```diff
@@ -241,23 +242,37 @@ IEnumerator IEnumerable.GetEnumerator()
...
+                    {
+                        ThrowInvalidHeaderCharacter(ch);
+                    }
+                }
+            }
+        }
+
+        public static void ValidateHeaderNameCharacters(string headerCharacters)
+        {
+            if (headerCharacters != null)
```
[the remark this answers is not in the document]
JunTaoLuo: Ah right, we used to use the same logic for both names and values and I just copied it. Will update add a test for it as well.

src/Kestrel.Core/Internal/Http/HttpHeaders.Generated.cs
```diff
@@ -5862,7 +5862,12 @@ protected override bool TryGetValueFast(string key, out StringValues value)
 
         protected override void SetValueFast(string key, in StringValues value)
         {
-            ValidateHeaderCharacters(value);
+            if (string.IsNullOrEmpty(key))
+            {
+                throw new InvalidOperationException(CoreStrings.FormatInvalidEmptyHeaderName());
```
halter73: Nit: Why not `CoreStrings.InvalidEmptyHeaderName`?

src/Kestrel.Core/Internal/Infrastructure/HttpCharacters.cs
```diff
@@ -0,0 +1,105 @@
...
+
+            // field-value https://tools.ietf.org/html/rfc7230#section-3.2
+            FieldValue = new bool[TableSize];
+            for (var c = 0x20; c <= 0x7e; c++) // VCHAR and SP
+            {
+                FieldValue[c] = true;
+            }
+        }
+
+        public static bool IsAuthority(byte c)
```
halter73: What if `c >= TableSize`?

src/Kestrel.Core/Internal/Infrastructure/HttpCharacters.cs
```diff
@@ -0,0 +1,105 @@
...
+            => Authority[c];
+
+        public static bool IsToken(byte c)
+            => Token[c];
+
+        public static bool IsToken(char c)
+            => c <= 0x7f && Token[c];
+
+        public static bool IsFieldValue(char c)
```
halter73: Nit: IsHeaderFieldValue

src/Kestrel.Core/Internal/Infrastructure/HttpCharacters.cs
JunTaoLuo: good catch

src/Kestrel.Core/Internal/Infrastructure/HttpCharacters.cs
JunTaoLuo: the http spec calls it `field-value`. While it's true this is probably only used for headers, I'd like to keep the naming as consistent as possible.

Tratcher: Then put the spec terms and link in the method comments

src/Kestrel.Core/Internal/Infrastructure/HttpCharacters.cs
```diff
@@ -0,0 +1,105 @@
...
+        }
+
+        public static bool IsAuthority(byte c)
+            => Authority[c];
+
+        public static bool IsToken(byte c)
+            => Token[c];
+
+        public static bool IsToken(char c)
+            => c <= 0x7f && Token[c];
```
benaadams: Can this just bounds check the array rather than use `0x7f`?
```csharp
var tokens = Token;
return (uint)c < (uint)tokens.length ? tokens[c] : false;
```
Otherwise its going to do a second bounds check when accessing the array

src/Kestrel.Core/Internal/Infrastructure/HttpCharacters.cs
JunTaoLuo: in the array definiting.

src/Kestrel.Core/Internal/Http/Http1Connection.cs
```diff
@@ -274,13 +274,9 @@ private void OnAuthorityFormTarget(HttpMethod method, Span<byte> target)
 
             // This is not complete validation. It is just a quick scan for invalid characters
             // but doesn't check that the target fully matches the URI spec.
-            for (var i = 0; i < target.Length; i++)
+            if (HttpCharacters.IndexOfInvalidAuthorityChar(target) >= 0)
```
Tratcher: You don't use the index anywhere.
`if (HttpCharacters.ContainsInvalidAuthorityChar(target))`

src/Kestrel.Core/Internal/Infrastructure/HttpCharacters.cs
```diff
@@ -0,0 +1,145 @@
...
+
+using System;
+using System.Runtime.CompilerServices;
+
+namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Infrastructure
+{
+    internal static class HttpCharacters
+    {
+        public static readonly int TableSize = 128;
+        public static readonly bool[] AlphaNumeric;
```
Tratcher: Not used outside of the cctor, no need for the static field.

src/Kestrel.Core/Internal/Infrastructure/HttpCharacters.cs
```diff
@@ -0,0 +1,145 @@
...
+        public static readonly bool[] Authority;
```
Tratcher: private

src/Kestrel.Core/Internal/Infrastructure/HttpUtilities.cs
```diff
@@ -449,13 +418,13 @@ public static void ValidateHostHeader(string hostText)
                 }
 
                 // Enregister array
-                var hostCharValidity = HostCharValidity;
+                var hostCharValidity = HttpCharacters.Host;
                 for (var i = 0; i < hostText.Length; i++)
```
Tratcher: No IndexOf method for Host?

src/Kestrel.Core/Internal/Http/Http1Connection.cs
JunTaoLuo: Not in this case but it is used in the `HttpHeaders`: https://github.com/aspnet/KestrelHttpServer/blob/release/2.2/src/Kestrel.Core/Internal/Http/HttpHeaders.cs#L262. I wasn't sure if a mix of Contains and IndexOf is a good idea so I converted all of them into IndexOf which covers almost all the scenarios.

src/Kestrel.Core/Internal/Infrastructure/HttpCharacters.cs
JunTaoLuo: Some of these arrays may be used as is instead of via the IndexOf methods: https://github.com/aspnet/KestrelHttpServer/pull/2707/files#diff-d3b42025fb7322d45d6c332b7723177eR447

src/Kestrel.Core/Internal/Infrastructure/HttpCharacters.cs
JunTaoLuo: It's a static ctor so these need to be static as well

src/Kestrel.Core/Internal/Infrastructure/HttpUtilities.cs
JunTaoLuo: No good reason this was missed.

src/Kestrel.Core/Internal/Http/Http1Connection.cs
Tratcher: Creating an API best suited to the caller is better than creating standard APIs for standards sake.

src/Kestrel.Core/Internal/Infrastructure/HttpCharacters.cs
```diff
@@ -0,0 +1,165 @@
...
+{
+    internal static class HttpCharacters
+    {
+        public static readonly int TableSize = 128;
+        public static readonly ReadOnlyMemory<bool> Authority;
+        public static readonly ReadOnlyMemory<bool> Token;
+        public static readonly ReadOnlyMemory<bool> Host;
+        public static readonly ReadOnlyMemory<bool> FieldValue;
+
+        static HttpCharacters()
```
benaadams: Dump static constructors; move all static constants to a `Constants` class where the statics are initalized inline so the class is marked `.beforefieldinit`

e.g.
```csharp
public static readonly ReadOnlyMemory<bool> Authority = InitalizeAuthority();
```

Then access a single one of the readonly statics; outside of the fast-path, so they all initalize with the initialization checks embedded in a non-hot method.

The static constructor will cause the thread-safe initialisation check to be added to every access. 

`.beforefieldinit` will cause the thread-safe initialisation check to be added to the first access.

src/Kestrel.Core/Internal/Http/HttpHeaders.cs
```diff
@@ -45,6 +46,10 @@ public abstract class HttpHeaders : IHeaderDictionary
                 {
                     ThrowHeadersReadOnlyException();
                 }
+                if (string.IsNullOrEmpty(key))
+                {
+                    throw new InvalidOperationException(CoreStrings.InvalidEmptyHeaderName);
```
benaadams: Use a not returning `ThrowXxx` method; rather than loading string and throwing inline.

src/Kestrel.Core/Internal/Http/HttpHeaders.cs
```diff
@@ -170,6 +175,10 @@ protected virtual bool CopyToFast(KeyValuePair<string, StringValues>[] array, in
             {
                 ThrowHeadersReadOnlyException();
             }
+            if (string.IsNullOrEmpty(key))
+            {
+                throw new InvalidOperationException(CoreStrings.InvalidEmptyHeaderName);
```
benaadams: Use a not returning `ThrowXxx` method; rather than loading string and throwing inline.

benaadams: Also use same method as one above

src/Kestrel.Core/Internal/Infrastructure/HttpCharacters.cs
JunTaoLuo: Does this have to be in a different `Constants` class? Would it not work to initialize the statics inline in this class?

benaadams: Yes, but there several different sets of readonly statics; so it would be easiest to collate them all, then only needs a single access to initalize them all, rather than one access per class.

JunTaoLuo: I'm not too fond of creating a separate constants class so I just kept all of the statics here.

src/Kestrel.Core/KestrelServer.cs
```diff
@@ -57,6 +57,11 @@ internal KestrelServer(ITransportFactory transportFactory, ServiceContext servic
             Features = new FeatureCollection();
             _serverAddresses = new ServerAddressesFeature();
             Features.Set(_serverAddresses);
+
+            var initializeAuthority = HttpCharacters.Authority;
```
Tratcher: HttpCharacters.Initialize()?

halter73: I like it. Also, you don't to access all the static fields in HttpCharacters to initialize them. You only need to access one.

src/Kestrel.Core/Internal/Infrastructure/HttpCharacters.cs
```diff
@@ -0,0 +1,202 @@
...
+        private static readonly bool[] _alphaNumeric = InitializeAlphaNumeric();
+        private static readonly bool[] _authority = InitializeAuthority();
+        private static readonly bool[] _token = InitializeToken();
+        private static readonly bool[] _host = InitializeHost();
+        private static readonly bool[] _fieldValue = InitializeFieldValue();
+
+        internal static void Initialize()
+        {
+            // Access _alphaNumeric to initialize static fields
+            var initialize = _alphaNumeric;
```
halter73: I've checked, and this class is now indeed marked beforefieldinit. I'm a little worried the JIT could optimize this way further delaying initialization. I think that's unlikely though. @benaadams do you know if that's possible?

benaadams: Could always pass the variable out and no-inline it so the Jit definitely can't optimise it away?
```csharp
[MethodImpl(MethodImplOptions.NoInlining)]
internal static bool[] Initialize()
{
    // Access _alphaNumeric to initialize static fields
    return _alphaNumeric;
}
```

JunTaoLuo:
Before:
```
                                   Method |       Mean |     Error |    StdDev |        Op/s | Scaled | ScaledSD |   Gen 0 |   Allocated |
----------------------------------------- |-----------:|----------:|----------:|------------:|-------:|---------:|--------:|------------:|
                     PlaintextTechEmpower | 1,229.6 ns |  9.761 ns |  8.653 ns |   813,286.3 |   1.00 |     0.00 |       - |   168.13 KB |
                     PlaintextAbsoluteUri | 1,791.3 ns | 17.342 ns | 14.481 ns |   558,249.4 |   1.46 |     0.01 |  0.9766 |   280.13 KB |
            PipelinedPlaintextTechEmpower |   964.8 ns |  9.717 ns |  9.089 ns | 1,036,446.7 |   0.78 |     0.01 |       - |  2690.03 KB |
 PipelinedPlaintextTechEmpowerDrainBuffer |   690.5 ns |  5.162 ns |  4.030 ns | 1,448,132.1 |   0.56 |     0.00 |  7.8125 |  2689.01 KB |
                               LiveAspNet | 2,242.8 ns | 23.972 ns | 21.250 ns |   445,878.9 |   1.82 |     0.02 |  1.9531 |   504.25 KB |
                      PipelinedLiveAspNet | 2,048.0 ns | 16.700 ns | 15.621 ns |   488,290.4 |   1.67 |     0.02 | 31.2500 |  8084.07 KB |
                                  Unicode | 3,316.6 ns | 65.822 ns | 73.161 ns |   301,512.6 |   2.70 |     0.06 |  3.9063 |   880.25 KB |
                         UnicodePipelined | 3,114.6 ns | 26.191 ns | 23.218 ns |   321,069.1 |   2.53 |     0.02 | 62.5000 | 14248.08 KB |
```
After:
```

                                   Method |       Mean |     Error |    StdDev |        Op/s | Scaled | ScaledSD |   Gen 0 |   Allocated |
----------------------------------------- |-----------:|----------:|----------:|------------:|-------:|---------:|--------:|------------:|
                     PlaintextTechEmpower | 1,221.7 ns | 23.391 ns | 21.880 ns |   818,508.2 |   1.00 |     0.00 |       - |   168.13 KB |
                     PlaintextAbsoluteUri | 1,725.6 ns | 21.603 ns | 18.040 ns |   579,512.5 |   1.41 |     0.03 |  0.9766 |   280.13 KB |
            PipelinedPlaintextTechEmpower |   949.1 ns | 12.651 ns | 11.834 ns | 1,053,597.8 |   0.78 |     0.02 |       - |  2690.03 KB |
 PipelinedPlaintextTechEmpowerDrainBuffer |   689.0 ns |  6.988 ns |  6.537 ns | 1,451,311.1 |   0.56 |     0.01 |  7.8125 |  2689.01 KB |
                               LiveAspNet | 2,238.0 ns | 24.954 ns | 23.342 ns |   446,824.9 |   1.83 |     0.04 |  1.9531 |   504.25 KB |
                      PipelinedLiveAspNet | 2,088.0 ns |  8.057 ns |  5.329 ns |   478,918.9 |   1.71 |     0.03 | 31.2500 |  8084.07 KB |
                                  Unicode | 3,196.0 ns | 12.174 ns | 10.792 ns |   312,889.5 |   2.62 |     0.05 |  3.9063 |   880.25 KB |
                         UnicodePipelined | 3,128.6 ns | 18.365 ns | 15.336 ns |   319,630.5 |   2.56 |     0.05 | 62.5000 | 14248.08 KB |
```

Issue #2706, opened 2018-07-09
PortSupportedCondition is unreliable


Tratcher:
PortSupportedConditionAttribute is defined in AddressRegistrationTests and used to skip some test cases if a designated port is not available. Several of the tests that use it are flaky and fail with port already in use exceptions.

Problem: skip conditions are evaluated once at the start of a test pass when the list of tests is built. The port may be available at that point, but can become unavailable before conditional test actually runs.

This is a strong contributor to https://github.com/aspnet/KestrelHttpServer/issues/2685, especailly these tests:
AddressRegistrationTests.RegisterAddresses_IPv4Port5000Default_Success(addressInput: null, testUrl: "http://127.0.0.1:5000/") 
AddressRegistrationTests.RegisterAddresses_IPv6Port5000Default_Succes…Input: "", testUrls: ["http://127.0.0.1:5000/", "http://[::1]:5000/"])
http://aspnetci/viewType.html?buildTypeId=XPlat_Windows_Win10Universe&branch_XPlat=%3Cdefault%3E&tab=buildTypeStatusDiv

All of the recent failures have been for the Sockets version of these tests, not Libuv. It's not clear what other tests these are conflicting with. It's not a libuv conflict since those run independently. It must be a test in one of the other projects since Libuv never experiences the same conflict.

Secondary problem: PortSupportedConditionAttribute only checks if the IPv4 port is available, it doesn't check IPv6 even though some of the tests are specifically for IPv6. I haven't seen any test failures yet for this case, likely because IPv4 fails first.

@ryanbrandenburg

Tratcher:
Recommendation: PortSupportedCondition should be removed and those tests should check the port directly before use.

PR #2714, opened 2018-07-11
Replace PortSupportedCondition

Tratcher:
#2706 The PortSupportedCondition is unreliable because it runs when the tests are first discovered, not when they're executed. Move the check into the test.

### DIFF
--- a/test/Kestrel.FunctionalTests/AddressRegistrationTests.cs
+++ b/test/Kestrel.FunctionalTests/AddressRegistrationTests.cs
@@ -48,17 +48,25 @@ namespace Microsoft.AspNetCore.Server.Kestrel.FunctionalTests
 
         [ConditionalTheory]
         [MemberData(nameof(AddressRegistrationDataIPv4Port5000Default))]
-        [PortSupportedCondition(5000)]
         public async Task RegisterAddresses_IPv4Port5000Default_Success(string addressInput, string testUrl)
         {
+            if (!CanBindToEndpoint(IPAddress.Loopback, 5000))
+            {
+                return;
+            }
+
             await RegisterAddresses_Success(addressInput, testUrl, 5000);
         }
 
         [ConditionalTheory]
         [MemberData(nameof(AddressRegistrationDataIPv4Port80))]
-        [PortSupportedCondition(80)]
         public async Task RegisterAddresses_IPv4Port80_Success(string addressInput, string testUrl)
         {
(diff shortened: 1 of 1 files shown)


test/Kestrel.FunctionalTests/AddressRegistrationTests.cs
```diff
@@ -48,17 +48,23 @@ public class AddressRegistrationTests : TestApplicationErrorLoggerLoggedTest
 
         [ConditionalTheory]
         [MemberData(nameof(AddressRegistrationDataIPv4Port5000Default))]
-        [PortSupportedCondition(5000)]
         public async Task RegisterAddresses_IPv4Port5000Default_Success(string addressInput, string testUrl)
         {
+            if (!CanBindToEndpoint(IPAddress.Loopback, 5000))
+            {
+                return;
+            }
```
halter73: Nit: newlines after closing brackets.

halter73:
@Tratcher Have you figure out if it's really other tests that are using these ports?

test/Kestrel.FunctionalTests/AddressRegistrationTests.cs
```diff
@@ -98,7 +104,6 @@ public class AddressRegistrationTests : TestApplicationErrorLoggerLoggedTest
         [ConditionalTheory]
         [MemberData(nameof(IPEndPointRegistrationDataPort443))]
         [IPv6SupportedCondition]
-        [PortSupportedCondition(443)]
```
halter73: The tests are failing because you missed a call to CanBindToEndpoint in RegisterIPEndPoint_Port443_Success.

Tratcher:
We don't actually get AddressInUseExceptions, we get errors like this:
```
Message: System.AggregateException : An error occurred while writing to logger(s). (Unexpected critical error. Log Critical[0]: Unable to start Kestrel. System.IO.IOException: Failed to bind to address http://localhost:80. ---> System.AggregateException: One or more errors occurred. (An attempt was made to access a socket in a way forbidden by its access permissions.) (An attempt was made to access a socket in a way forbidden by its access permissions.) ---> System.Net.Sockets.SocketException: An attempt was made to access a socket in a way forbidden by its access permissions.
```

Tratcher:
I was able to reproduce the theory by staring another process after test discovery, but I wasn't able to find any case where another test started against port 5000. The most likely case is that a test binds to 0 and gets assigned 5000.

Issue #2659, opened 2018-06-11
Http2Stream.TryParseRequest needs to be able to respond with stream errors


Tratcher:
Http2Stream.TryParseRequest is where the basic request fields like method, scheme, path, etc. are validated. Per https://tools.ietf.org/html/rfc7540#section-8.1.2.6 and https://tools.ietf.org/html/rfc7540#section-5.4.2 if anythings wrong then the server should send back a RST_STREAM frame with PROTOCOL_ERROR.

Current failures are reported via BadHttpRequestException, which HttpProtocol.ProcessRequestsAsync catches and turns into a 400 response. This is allowed: "server MAY send an HTTP response prior to
   closing or resetting the stream", but this doesn't result in the expected RST_STREAM frame that would tell the client not to send the body.

Related:
- https://github.com/aspnet/KestrelHttpServer/issues/2154 h2spec failure http2/8.1.2.3/1

Tratcher:
The last place this kind of validation happens that can throw the right exception type is Http2Connection.StartStream.
https://github.com/aspnet/KestrelHttpServer/blob/a217206f1fe0d7a8abb67aaac54b3616e25d6f87/src/Kestrel.Core/Internal/Http2/Http2Connection.cs#L665-L671

Issue #2263, opened 2018-01-19
Http/2: Validate psudo-headers


Tratcher:
https://github.com/aspnet/KestrelHttpServer/blob/526dfdb332b30fd6042d8ef8d1a2ba99cc654010/src/Kestrel.Core/Internal/Http2/Http2Stream.cs#L55-L64

E.g.  Scheme should match the connection, http or https.

Related: https://github.com/aspnet/KestrelHttpServer/issues/2205 - Path un-escaping

Issue #2205, opened 2017-12-04
HTTP/2 Path is not un-escaped


Tratcher:
https://github.com/aspnet/KestrelHttpServer/blob/b8a1c04ffbeee6c60a74766142ff3e2e6779d701/src/Kestrel.Core/Internal/Http2/Http2Stream.cs#L59
IHttpRequestFeature.Path is expected to be in its un-escaped format, but the HTTP/2 code path is providing the raw escaped format. This will break request routing and processing.

Compare to the HTTP/1 code path:
https://github.com/aspnet/KestrelHttpServer/blob/b8a1c04ffbeee6c60a74766142ff3e2e6779d701/src/Kestrel.Core/Internal/Http/Http1Connection.cs#L214

Issue #2154, opened 2017-11-09
h2spec failures


anurse:
After running [h2spec](https://github.com/summerwind/h2spec) on Kestrel's HTTP/2 implementation, I encountered the following failures. This issue tracks fixing those failures and then enabling automated h2spec testing:

```
  * http2/6.5.3 Sends multiple values of SETTINGS_INITIAL_WINDOW_SIZE (duration: 0.01s)
    DATA Frame (length:1, flags:0x00, stream_id:1)
    DATA Frame (length:12, flags:0x00, stream_id:1)
  * http2/6.9.1 Sends SETTINGS frame to set the initial window size to 1 and sends HEADERS frame (duration: 0.01s)
    DATA Frame (length:1, flags:0x00, stream_id:1)
    DATA Frame (length:12, flags:0x00, stream_id:1)
  * http2/6.9.1 Sends multiple WINDOW_UPDATE frames increasing the flow control window to above 2^31-1 (duration: 2.00s)
    GOAWAY Frame (Error Code: FLOW_CONTROL_ERROR)
    SETTINGS Frame (length:0, flags:0x01, stream_id:0)
  * http2/6.9.1 Sends multiple WINDOW_UPDATE frames increasing the flow control window to above 2^31-1 on a stream (duration: 2.00s)
    RST_STREAM Frame (Error Code: FLOW_CONTROL_ERROR)
    DATA Frame (length:0, flags:0x01, stream_id:1)
  * http2/6.9.2 Changes SETTINGS_INITIAL_WINDOW_SIZE after sending HEADERS frame (duration: 0.01s)
    DATA Frame (length:1, flags:0x00, stream_id:1)
    DATA Frame (length:12, flags:0x00, stream_id:1)
  * http2/6.9.2 Sends a SETTINGS frame for window size to be negative (duration: 2.01s)
    DATA Frame (length:1, flags:0x00, stream_id:1)
    Timeout
  * hpack/4.2 Sends a dynamic table size update at the end of header block (duration: 2.01s)
    GOAWAY Frame (Error Code: COMPRESSION_ERROR)
    Connection closed
    DATA Frame (length:0, flags:0x01, stream_id:1)
```

anurse:
Also this one is flaky:

```
  * generic/3.1 Sends a DATA frame with padding (duration: 0.01s)
    HEADERS Frame (stream_id:1)
    Connection closed
```

Tratcher:
Also http2/5.4.1/2
[...]

PR #2731, opened 2018-07-17
Validate Http/2 pseudo headers

Tratcher:
#2205 #2263 #2659 https://github.com/aspnet/KestrelHttpServer/issues/2154#issuecomment-395504675

These fields were not being properly validated or decoded (path). When they're invalid send a RST frame. The remaining exception is Host validation since the shared code there would need to be re-written, it currently sends a 400 which is passible.

I split the Http2Connection tests into Http2Stream tests for behavior specific to the Http2Stream class.

### DIFF
--- a/src/Kestrel.Core/Internal/Http/PathNormalizer.cs
+++ b/src/Kestrel.Core/Internal/Http/PathNormalizer.cs
@@ -4,6 +4,9 @@
 using System;
 using System.Diagnostics;
 using System.Runtime.InteropServices;
+using System.Text;
+using Microsoft.AspNetCore.Connections.Abstractions;
+using Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Infrastructure;
 
 namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
 {
@@ -11,6 +14,53 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
     {
         private const byte ByteSlash = (byte)'/';
         private const byte ByteDot = (byte)'.';
+
+        public static string DecodePath(Span<byte> path, bool pathEncoded, string rawTarget, int queryLength)
+        {
+            int pathLength;
+            if (pathEncoded)
+            {
+                // URI was encoded, unescape and then parse as UTF-8
+                // Disabling warning temporary
+                pathLength = UrlDecoder.DecodeInPlace(path);
+
+                // Removing dot segments must be done after unescaping. From RFC 3986:
+                //
+                // URI producing applications should percent-encode data octets that
+                // correspond to characters in the reserved set unless these characters
+                // are specifically allowed by the URI scheme to represent data in that
+                // component.  If a reserved character is found in a URI component and
+                // no delimiting role is known for that character, then it must be
+                // interpreted as representing the data octet corresponding to that
+                // character's encoding in US-ASCII.
+                //
+                // https://tools.ietf.org/html/rfc3986#section-2.2
+                pathLength = RemoveDotSegments(path.Slice(0, pathLength));
+
+                return GetUtf8String(path.Slice(0, pathLength));
+            }
+
+            pathLength = RemoveDotSegments(path);
+
+            if (path.Length == pathLength && queryLength == 0)
+            {
+                // If no decoding was required, no dot segments were removed and
+                // there is no query, the request path is the same as the raw target
+                return rawTarget;
+            }
+
+            return path.Slice(0, pathLength).GetAsciiStringNonNullCharacters();
+        }
+
+        private static unsafe string GetUtf8String(Span<byte> path)
+        {
+            // .NET 451 doesn't have pointer overloads for Encoding.GetString so we
+            // copy to an array
+            fixed (byte* pointer = &MemoryMarshal.GetReference(path))
+            {
+                return Encoding.UTF8.GetString(pointer, path.Length);
+            }
+        }
 
         // In-place implementation of the algorithm from https://tools.ietf.org/html/rfc3986#section-5.2.4
         public static unsafe int RemoveDotSegments(Span<byte> input)
--- a/src/Kestrel.Core/Internal/Http2/Http2Stream.cs
+++ b/src/Kestrel.Core/Internal/Http2/Http2Stream.cs
@@ -3,13 +3,18 @@
 
 using System;
 using System.Buffers;
+using System.Diagnostics;
 using System.IO;
 using System.IO.Pipelines;
+using System.Runtime.InteropServices;
+using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Connections;
+using Microsoft.AspNetCore.Connections.Abstractions;
 using Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http;
 using Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Infrastructure;
+using Microsoft.Extensions.Primitives;
 using Microsoft.Net.Http.Headers;
 
 namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http2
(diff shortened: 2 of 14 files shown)


Tratcher:
Updated with better error messages, logging, and RST's for invalid Host/Authority headers.

src/Kestrel.Core/Internal/Http2/Http2Stream.cs
```diff
@@ -55,71 +60,221 @@ protected override string CreateRequestId()
...
+            // Must start with a leading slash
+            if (pathSegment.Length == 0 || pathSegment[0] != '/')
+            {
+                ResetAndAbort(new ConnectionAbortedException(CoreStrings.FormatHttp2StreamErrorPathInvalid(RawTarget)), Http2ErrorCode.PROTOCOL_ERROR);
+                return false;
+            }
+
+            var pathEncoded = pathSegment.IndexOf('%') >= 0;
+
+            // Borrowed from Http1Connection.OnOriginFormTarget
```
halter73: Can we share this somehow?

src/Kestrel.Core/Internal/Http2/Http2Stream.cs
```diff
@@ -55,71 +60,221 @@ protected override string CreateRequestId()
...
+                        // there is no query, the request path is the same as the raw target
+                        requestUrlPath = pathSegment.ToString();
+                    }
+                    else
+                    {
+                        requestUrlPath = pathBuffer.Slice(0, pathLength).GetAsciiStringNonNullCharacters();
+                    }
+                }
+            }
+            catch (InvalidOperationException)
```
halter73: It would be nice if we created a shared TryParseTarget method and didn't rely on catching an exception for invalid targets if possible.

src/Kestrel.Core/Internal/Http2/Http2Stream.cs
```diff
@@ -55,71 +60,221 @@ protected override string CreateRequestId()
...
+        // Compare to Http1Connection.OnStartLine
         protected override bool TryParseRequest(ReadResult result, out bool endConnection)
         {
             // We don't need any of the parameters because we don't implement BeginRead to actually
             // do the reading from a pipeline, nor do we use endConnection to report connection-level errors.
+            endConnection = !TryValidatePseudoHeaders();
+            return true;
+        }
+
+        private bool TryValidatePseudoHeaders()
```
davidfowl: Is there now way to do this before we turn them into strings?

Tratcher: Not without significant changes to the decoder and header collections.
1. Make these all Known headers https://github.com/aspnet/KestrelHttpServer/issues/2261
2. Make the decoder return raw bytes for Known headers
3. Do validation on the raw bytes, then decode
4. Lazy decode any remaining known headers

The :path is the only field this will make a perf difference for on valid requests, it's the only one transformed.

src/Kestrel.Core/Internal/Http2/Http2Stream.cs
Tratcher: Unifying the code here is reasonable. That exception only comes from the decoders, I'm not going to try re-writing them at the moment.

Tratcher:
Refactored.

src/Kestrel.Core/Internal/Http2/Http2Stream.cs
halter73: I could see this being not worth the effort to rewrite the entire DecodePath method not to throw InvalidOperationException internally. I still think it's clearer to move the `catch (InvalidOperationException)` inside the DecodePath method. Just rename it to TryDecodePath and return false in the the catch.

Tratcher: In the Http1 code path it's not just DecodePath inside the try/catch, it's also RawTarget and Query ASCII decoding

src/Kestrel.Core/Internal/Http2/Http2Stream.cs
```diff
@@ -191,6 +140,84 @@ private bool TryValidatePseudoHeaders()
...
+            if (host.Count > 1)
+            {
+                // RST replaces 400
+                ResetAndAbort(new ConnectionAbortedException(CoreStrings.FormatBadRequest_InvalidHostHeader_Detail(hostText)), Http2ErrorCode.PROTOCOL_ERROR);
+                return false;
+            }
+
+            try
+            {
+                HttpUtilities.ValidateHostHeader(hostText);
```
halter73: I'd still like to see this change to an IsHostHeaderValid(hostText) call. We can keep ValidateHostHeader for HTTP/1.x and just have that also call IsHostHeaderValid.

src/Kestrel.Core/Internal/Http2/Http2Stream.cs
halter73: https://github.com/aspnet/KestrelHttpServer/issues/2737

src/Kestrel.Core/Internal/Http2/Http2Stream.cs
Tratcher: Updated

src/Kestrel.Core/Internal/Http2/Http2Stream.cs
```diff
@@ -55,69 +60,190 @@ protected override string CreateRequestId()
...
-            // A server MUST respond with a 400 (Bad Request) status code to any
-            // HTTP/1.1 request message that lacks a Host header field and to any
-            // request message that contains more than one Host header field or a
-            // Host header field with an invalid field-value.
+            var pathSegment = queryIndex == -1 ? path.AsSpan() : path.AsSpan(0, queryIndex);
+
+            return TryValidatePath(pathSegment);
+        }
+
+        private bool TryValidateMethod()
```
JunTaoLuo: Curious why the `TryValidat*` naming. Might be more intuitive as `*IsValid`

Tratcher: Try was more traditionally used for methods and Is for properties (e.g. persistent state), but otherwise there's no reason one way or the other.

src/Kestrel.Core/Internal/Http2/Http2Stream.cs
```diff
@@ -55,69 +60,190 @@ protected override string CreateRequestId()
...
+
+            try
+            {
+                // The decoder operates only on raw bytes
+                var pathBuffer = new byte[pathSegment.Length].AsSpan();
+                for (int i = 0; i < pathSegment.Length; i++)
+                {
+                    var ch = pathSegment[i];
+                    // The header parser should already be checking this
+                    Debug.Assert(20 < ch && ch < 127);
```
JunTaoLuo: This looks odd. If the header parser is checking this then why bother with the debug.assert? if we want to double check, then I think we'll need to follow the RFC https://tools.ietf.org/html/rfc3986#section-3.3 and reject gen-delims for example (for example `[` and `]`). If we want to do a quick check I think you want to assert `32 < ch && ch < 127` 20-32 in dec are still control chars and space.

src/Kestrel.Core/Internal/Http2/Http2Stream.cs
Tratcher: Actually, the big difference is that these methods have side effects, they ResetAndAbort if it fails. Not planning to rename.

src/Kestrel.Core/Internal/Http2/Http2Stream.cs
halter73: The point of a Debug.Assert is that it should be impossible for it to fail. I don't mind double checking at all. What we shouldn't do is add a Debug.Assert that could fail because previous validation was less strict than the assert.

src/Kestrel.Core/Internal/Http/PathNormalizer.cs
```diff
@@ -12,6 +15,53 @@ public static class PathNormalizer
...
+                // there is no query, the request path is the same as the raw target
+                return rawTarget;
+            }
+
+            return path.Slice(0, pathLength).GetAsciiStringNonNullCharacters();
+        }
+
+        private static unsafe string GetUtf8String(Span<byte> path)
+        {
+            // .NET 451 doesn't have pointer overloads for Encoding.GetString so we
```
davidfowl: This comment is wrong

Issue #2760, opened 2018-07-27
Mitigate UseIIS conflict


Tratcher:
Today CreateDefaultBuilder calls UseKestrel and then it calls UseIIS for conditional in-proc support. Each of these register IServer and the last one wins.

The conflict comes when you call UseKestrel again after CreateDefaultBuilder to set some options. This works fine when you run self-host (e.g. on your Mac), but when switch to IIS in-proc (like publishing to Azure) Kestrel replaces IIS in-proc and the server fails to start.

There's a workaround today where instead of calling UseKestrel you can call
```
                builder.ConfigureServices(services =>
                {
                    services.Configure<KestrelServerOptions>(options =>
                    {
                        // ...
                    });
                })
```

Proposal: Promote the workaround to a top level API
```
                builder.ConfigureKestrel(options =>
                {
                    // ...
                })
```

This becomes a drop in replacement for anyone already calling UseKestrel. Discovery would be a problem still, but we could update our samples to help with that.

@shirhatti @halter73 @pakrym @jkotalik

halter73:
Would we deprecate the UseKestrel overload with an options configuration callback?

Tratcher:
That's pretty agressive. Maybe in 3.0 if we saw a lot of people run into this conflict?

shirhatti:
I'm with Chris on this. No need to deprecate it unless it proves to be a pit of failure

PR #2761, opened 2018-07-27
Add new ConfigureKestrel extensions

Tratcher:
#2760 To avoid conflicts with UseIIS.

### DIFF
--- a/samples/Http2SampleApp/Program.cs
+++ b/samples/Http2SampleApp/Program.cs
@@ -24,7 +24,8 @@ namespace Http2SampleApp
                     factory.SetMinimumLevel(LogLevel.Trace);
                     factory.AddConsole();
                 })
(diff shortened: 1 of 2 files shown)


samples/Http2SampleApp/Program.cs
```diff
@@ -24,7 +24,8 @@ public static void Main(string[] args)
                     factory.SetMinimumLevel(LogLevel.Trace);
                     factory.AddConsole();
                 })
-                .UseKestrel((context, options) =>
+                .UseKestrel()
+                .ConfigureKestrel((context, options) =>
```
halter73: Why not change the other samples?

Tratcher: In these samples it doesn't matter which you use, I only changed this one as a proof of concept.

Issue #2733, opened 2018-07-18
Verify request and response Content-Length


Tratcher:
https://tools.ietf.org/html/rfc7540#section-8.1.2.6

In http2 requests and responses use the chunk style data frames to transfer the body and indicate when it's complete. However the Content-Length header is still supported and MUST be cross-checked against the number of bytes received or sent. A mismatch is a stream level protocol error.

We can also use the Content-Length tracking to know when to set the End_Stream flag rather than waiting until the app pipeline exits. Compare with https://github.com/aspnet/KestrelHttpServer/issues/2678
(Q: How would that optimization interact with trailers? Related: https://github.com/aspnet/KestrelHttpServer/issues/2630)

PR #2769, opened 2018-07-31
Verify request Content-Length

Tratcher:
#2733 
I had to re-order a few things to make sure the content-length header had been received and parsed before we acted on the end-of-stream flag.

I'll do a seperate PR for the response length verification.

### DIFF
--- a/src/Kestrel.Core/Internal/Http2/Http2Connection.cs
+++ b/src/Kestrel.Core/Internal/Http2/Http2Connection.cs
@@ -73,6 +73,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http2
         private Http2Stream _currentHeadersStream;
         private RequestHeaderParsingState _requestHeaderParsingState;
         private PseudoHeaderFields _parsedPseudoHeaderFields;
+        private Http2HeadersFrameFlags _headerFlags;
         private bool _isMethodConnect;
         private int _highestOpenedStreamId;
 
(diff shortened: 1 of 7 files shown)


src/Kestrel.Core/Internal/Http2/Http2Connection.cs
```diff
@@ -372,6 +373,12 @@ private Task ProcessDataFrameAsync()
                     throw new Http2ConnectionErrorException(CoreStrings.FormatHttp2ErrorStreamHalfClosedRemote(_incomingFrame.Type, stream.StreamId), Http2ErrorCode.STREAM_CLOSED);
                 }
 
+                if (stream.IsAborted)
+                {
+                    // Ignore data frames for aborted streams.
```
halter73: This prevents data frames from being counted towards connection-level input flow control. I think this could lead to mismatched connection window size accounting by the client and server.

src/Kestrel.Core/Internal/Http2/Http2Stream.cs
```diff
@@ -279,8 +280,20 @@ public Task OnDataAsync(Http2Frame dataFrame)
             {
                 RequestBodyStarted = true;
 
+                if (InputRemaining.HasValue)
+                {
+                    // https://tools.ietf.org/html/rfc7540#section-8.1.2.6
+                    if (payload.Count > InputRemaining.Value)
+                    {
+                        throw new Http2StreamErrorException(StreamId, CoreStrings.Http2StreamErrorMoreDataThanLength, Http2ErrorCode.PROTOCOL_ERROR);
```
halter73: Does this abort the connection or just the stream? If the latter, I'm pretty sure the payload still needs to count towards the connection-level flow control window. Calling  `_inputFlowControl.Abort()` _then_ `_inputFlowControl.Advance(payload.Count)` should do the trick.

Tratcher: It's only for the stream. I'll take a look at flow connection flow control accounting.

src/Kestrel.Core/Internal/Http2/Http2Stream.cs
```diff
@@ -279,8 +280,20 @@ public Task OnDataAsync(Http2Frame dataFrame)
...
+                    }
+
+                    InputRemaining -= payload.Count;
+                }
+
                 if (endStream)
                 {
+
```
JunTaoLuo: nit: extra line

halter73:
Looks good!

Issue #2703, opened 2018-07-06
Re-implement received GOAWAY support


Tratcher:
https://tools.ietf.org/html/rfc7540#section-6.8

There's a rudimentary implementation for processing received GOAWAY frames that immediately aborts all streams and then closes the connection. GOAWAY is supposed to indicate a graceful shutdown, not an immediate abort. After a GOAWAY you're only guaranteed that the client will not open any new streams. They may continue to send you Window Updates, Pings, Data frames, etc., and the server may continue to send responses. Once all responses are complete then the server can send its own GOAWAY and close the connection. If the client really wants to abort the connection they will send a GOAWAY and then close the socket.

Issue #2671, opened 2018-06-15
Http/2 Implement graceful shutdown request draining


Tratcher:
On IServer.StopAsync the server should wait for the given timeout (5s) for requests to complete. This hasn't been implemented yet for the HTTP/2 connections, all active requests are aborted immediately.

PR #2771, opened 2018-08-01
Gracefully shutdown HTTP/2 connections on server shutdown

JunTaoLuo:
Addresses https://github.com/aspnet/KestrelHttpServer/issues/2703 and https://github.com/aspnet/KestrelHttpServer/issues/2671. 

Adding a FSM to Http2Connection:
1. Open - This is the initial state. During this state, we will process all incoming requests
2. Closing - This state is reached upon receiving a GOAWAY from the client or when the server receives the signal to shutdown (via ctrl-c for example). During this state we will still process incoming frames on streams that are open. Once all streams complete, the connection will be closed. During server initiated shutdown, we will send a GOAWAY to the client with max value for highest opened stream id as a notification to the client that shutdown is imminent but we are still processing existing requests.
3. Closed - This state is reached upon abortive shutdown, or when all remaining streams have been closed while in the Closing state. We send a final GOAWAY to the client that also include the actual highest opened stream ID we processed.

### DIFF
--- a/src/Kestrel.Core/Internal/Http2/Http2Connection.cs
+++ b/src/Kestrel.Core/Internal/Http2/Http2Connection.cs
@@ -75,9 +75,9 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http2
         private PseudoHeaderFields _parsedPseudoHeaderFields;
         private Http2HeadersFrameFlags _headerFlags;
         private bool _isMethodConnect;
+        private readonly object _stateLock = new object();
         private int _highestOpenedStreamId;
-
-        private bool _stopping;
+        private Http2ConnectionState _state = Http2ConnectionState.Open;
 
         private readonly ConcurrentDictionary<int, Http2Stream> _streams = new ConcurrentDictionary<int, Http2Stream>();
 
@@ -98,20 +98,60 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http2
 
         public void OnInputOrOutputCompleted()
         {
-            _stopping = true;
+            lock (_stateLock)
+            {
+                if (_state != Http2ConnectionState.Closed)
+                {
+                    _frameWriter.WriteGoAwayAsync(_highestOpenedStreamId, Http2ErrorCode.NO_ERROR);
+                    UpdateState(Http2ConnectionState.Closed);
+                }
+            }
+
             _frameWriter.Complete();
         }
 
         public void Abort(ConnectionAbortedException ex)
         {
-            _stopping = true;
+            lock (_stateLock)
+            {
+                if (_state != Http2ConnectionState.Closed)
+                {
+                    _frameWriter.WriteGoAwayAsync(_highestOpenedStreamId, Http2ErrorCode.INTERNAL_ERROR);
+                    UpdateState(Http2ConnectionState.Closed);
+                }
+            }
+
             _frameWriter.Abort(ex);
         }
 
         public void StopProcessingNextRequest()
+            => StopProcessingNextRequest(true);
+
+        public void StopProcessingNextRequest(bool sendGracefulGoAway = false)
         {
-            _stopping = true;
-            Input.CancelPendingRead();
+            lock (_stateLock)
+            {
+                if (_state == Http2ConnectionState.Open)
+                {
+                    if (_streams.IsEmpty)
+                    {
+                        _frameWriter.WriteGoAwayAsync(_highestOpenedStreamId, Http2ErrorCode.NO_ERROR);
+                        UpdateState(Http2ConnectionState.Closed);
+
+                        // Wake up request processing loop so the connection can complete if there are no pending requests
+                        Input.CancelPendingRead();
+                    }
+                    else
+                    {
+                        if (sendGracefulGoAway)
+                        {
+                            _frameWriter.WriteGoAwayAsync(Int32.MaxValue, Http2ErrorCode.NO_ERROR);
+                        }
+
+                        UpdateState(Http2ConnectionState.Closing);
+                    }
+                }
+            }
         }
 
         public async Task ProcessRequestsAsync<TContext>(IHttpApplication<TContext> application)
@@ -128,12 +168,12 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http2
                     return;
                 }
 
-                if (!_stopping)
+                if (_state != Http2ConnectionState.Closed)
                 {
                     await _frameWriter.WriteSettingsAsync(_serverSettings);
                 }
 
-                while (!_stopping)
+                while (_state != Http2ConnectionState.Closed)
                 {
                     var result = await Input.ReadAsync();
                     var readableBuffer = result.Buffer;
@@ -198,11 +238,20 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http2
             }
             finally
             {
-                 var connectionError = error as ConnectionAbortedException
+                var connectionError = error as ConnectionAbortedException
                     ?? new ConnectionAbortedException(CoreStrings.Http2ConnectionFaulted, error);
 
                 try
                 {
+                    lock (_stateLock)
+                    {
+                        if (_state != Http2ConnectionState.Closed)
+                        {
+                            _frameWriter.WriteGoAwayAsync(_highestOpenedStreamId, errorCode);
+                            UpdateState(Http2ConnectionState.Closed);
+                        }
+                    }
+
                     // Ensure aborting each stream doesn't result in unnecessary WINDOW_UPDATE frames being sent.
                     _inputFlowControl.StopWindowUpdates();
 
@@ -211,7 +260,6 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http2
                         stream.Abort(connectionError);
                     }
 
-                    await _frameWriter.WriteGoAwayAsync(_highestOpenedStreamId, errorCode);
                     _frameWriter.Complete();
                 }
                 catch
@@ -245,7 +293,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http2
 
         private async Task<bool> TryReadPrefaceAsync()
         {
-            while (!_stopping)
+            while (_state != Http2ConnectionState.Closed)
             {
                 var result = await Input.ReadAsync();
                 var readableBuffer = result.Buffer;
@@ -625,7 +673,8 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http2
                 throw new Http2ConnectionErrorException(CoreStrings.FormatHttp2ErrorStreamIdNotZero(_incomingFrame.Type), Http2ErrorCode.PROTOCOL_ERROR);
             }
 
-            StopProcessingNextRequest();
+            StopProcessingNextRequest(sendGracefulGoAway: false);
+
             return Task.CompletedTask;
         }
 
@@ -725,12 +774,20 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http2
         {
             try
             {
-                _hpackDecoder.Decode(payload, endHeaders, handler: this);
-
-                if (endHeaders)
+                lock (_stateLock)
                 {
-                    StartStream(application);
-                    ResetRequestHeaderParsingState();
+                    _highestOpenedStreamId = _currentHeadersStream.StreamId;
+                    _hpackDecoder.Decode(payload, endHeaders, handler: this);
+
+                    if (endHeaders)
+                    {
+                        if (_state != Http2ConnectionState.Closed)
+                        {
+                            StartStream(application);
+                        }
+
+                        ResetRequestHeaderParsingState();
+                    }
                 }
             }
             catch (Http2StreamErrorException)
@@ -786,11 +843,6 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http2
 
         private void ResetRequestHeaderParsingState()
         {
-            if (_requestHeaderParsingState != RequestHeaderParsingState.Trailers)
-            {
-                _highestOpenedStreamId = _currentHeadersStream.StreamId;
-            }
-
             _currentHeadersStream = null;
             _requestHeaderParsingState = RequestHeaderParsingState.Ready;
             _parsedPseudoHeaderFields = PseudoHeaderFields.None;
@@ -827,7 +879,19 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http2
 
         void IHttp2StreamLifetimeHandler.OnStreamCompleted(int streamId)
         {
-            _streams.TryRemove(streamId, out _);
+            lock (_stateLock)
+            {
+                _streams.TryRemove(streamId, out _);
+
+                if (_state == Http2ConnectionState.Closing && _streams.IsEmpty)
+                {
+                    _frameWriter.WriteGoAwayAsync(_highestOpenedStreamId, Http2ErrorCode.NO_ERROR);
+                    UpdateState(Http2ConnectionState.Closed);
+
+                    // Wake up request processing loop so the connection can complete if there are no pending requests
+                    Input.CancelPendingRead();
+                }
+            }
         }
 
         public void OnHeader(Span<byte> name, Span<byte> value)
@@ -953,6 +1017,19 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http2
         private static bool IsConnectionSpecificHeaderField(Span<byte> name, Span<byte> value)
         {
             return name.SequenceEqual(_connectionBytes) || (name.SequenceEqual(_teBytes) && !value.SequenceEqual(_trailersBytes));
+        }
+
+        private void UpdateState(Http2ConnectionState state)
+        {
+            _state = state;
+            if (state == Http2ConnectionState.Closing)
+            {
+                Log.Http2ConnectionClosing(_context.ConnectionId);
+            }
+            else if (state == Http2ConnectionState.Closed)
+            {
+                Log.Http2ConnectionClosed(_context.ConnectionId, _highestOpenedStreamId);
+            }
         }
 
         void ITimeoutControl.SetTimeout(long ticks, TimeoutAction timeoutAction)
--- a/src/Kestrel.Core/Internal/Http2/Http2Frame.GoAway.cs
+++ b/src/Kestrel.Core/Internal/Http2/Http2Frame.GoAway.cs
@@ -7,7 +7,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http2
     {
         public int GoAwayLastStreamId
         {
-            get => (Payload[0] << 24) | (Payload[1] << 16) | (Payload[2] << 16) | Payload[3];
+            get => (Payload[0] << 24) | (Payload[1] << 16) | (Payload[2] << 8) | Payload[3];
             set
             {
                 Payload[0] = (byte)((value >> 24) & 0xff);
@@ -19,7 +19,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http2
 
         public Http2ErrorCode GoAwayErrorCode
         {
-            get => (Http2ErrorCode)((Payload[4] << 24) | (Payload[5] << 16) | (Payload[6] << 16) | Payload[7]);
+            get => (Http2ErrorCode)((Payload[4] << 24) | (Payload[5] << 16) | (Payload[6] << 8) | Payload[7]);
             set
             {
                 Payload[4] = (byte)(((uint)value >> 24) & 0xff);
--- a/test/Kestrel.Core.Tests/Http2ConnectionTests.cs
+++ b/test/Kestrel.Core.Tests/Http2ConnectionTests.cs
@@ -11,8 +11,9 @@ using System.Linq;
 using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
-using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.Connections;
+using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Http.Features;
 using Microsoft.AspNetCore.Server.Kestrel.Core.Features;
 using Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http;
 using Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http2;
@@ -21,8 +22,8 @@ using Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Infrastructure;
 using Microsoft.AspNetCore.Server.Kestrel.Transport.Abstractions.Internal;
 using Microsoft.AspNetCore.Testing;
 using Microsoft.Net.Http.Headers;
+using Moq;
 using Xunit;
-using Microsoft.AspNetCore.Http.Features;
 
 namespace Microsoft.AspNetCore.Server.Kestrel.Core.Tests
 {
@@ -110,6 +111,8 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Tests
         private readonly Dictionary<string, string> _decodedHeaders = new Dictionary<string, string>(StringComparer.OrdinalIgnoreCase);
         private readonly HashSet<int> _abortedStreamIds = new HashSet<int>();
         private readonly object _abortedStreamIdsLock = new object();
+        private readonly TaskCompletionSource<object> _closingStateReached = new TaskCompletionSource<object>(TaskCreationOptions.RunContinuationsAsynchronously);
+        private readonly TaskCompletionSource<object> _closedStateReached = new TaskCompletionSource<object>(TaskCreationOptions.RunContinuationsAsynchronously);
 
         private readonly RequestDelegate _noopApplication;
         private readonly RequestDelegate _readHeadersApplication;
@@ -305,12 +308,23 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Tests
 
             _pair = DuplexPipe.CreateConnectionPair(inputPipeOptions, outputPipeOptions);
 
+            var mockKestrelTrace = new Mock<TestKestrelTrace>(_logger)
+            {
+                CallBase = true
+            };
+            mockKestrelTrace
+                .Setup(m => m.Http2ConnectionClosing(It.IsAny<string>()))
+                .Callback(() => _closingStateReached.SetResult(null));
+            mockKestrelTrace
+                .Setup(m => m.Http2ConnectionClosed(It.IsAny<string>(), It.IsAny<int>()))
+                .Callback(() => _closedStateReached.SetResult(null));
+
             _connectionContext = new Http2ConnectionContext
             {
                 ConnectionFeatures = new FeatureCollection(),
                 ServiceContext = new TestServiceContext()
                 {
-                    Log = new TestKestrelTrace(_logger)
+                    Log = mockKestrelTrace.Object
                 },
                 MemoryPool = _memoryPool,
                 Application = _pair.Application,
@@ -946,7 +960,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Tests
 
             await WaitForConnectionErrorAsync<Http2ConnectionErrorException>(
                 ignoreNonGoAwayFrames: false,
-                expectedLastStreamId: 0,
+                expectedLastStreamId: 1,
                 expectedErrorCode: Http2ErrorCode.PROTOCOL_ERROR,
                 expectedErrorMessage: CoreStrings.FormatHttp2ErrorHeadersInterleaved(Http2FrameType.DATA, streamId: 1, headersStreamId: 1));
         }
@@ -1585,7 +1599,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Tests
 
             await WaitForConnectionErrorAsync<Http2ConnectionErrorException>(
                 ignoreNonGoAwayFrames: false,
-                expectedLastStreamId: 0,
+                expectedLastStreamId: 1,
                 expectedErrorCode: Http2ErrorCode.PROTOCOL_ERROR,
                 expectedErrorMessage: CoreStrings.FormatHttp2ErrorHeadersInterleaved(Http2FrameType.HEADERS, streamId: 3, headersStreamId: 1));
         }
@@ -1613,7 +1627,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Tests
 
             await WaitForConnectionErrorAsync<HPackDecodingException>(
                 ignoreNonGoAwayFrames: false,
-                expectedLastStreamId: 0,
+                expectedLastStreamId: 1,
                 expectedErrorCode: Http2ErrorCode.COMPRESSION_ERROR,
                 expectedErrorMessage: CoreStrings.HPackErrorIncompleteHeaderBlock);
         }
@@ -1869,7 +1883,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Tests
 
             await WaitForConnectionErrorAsync<Http2ConnectionErrorException>(
                 ignoreNonGoAwayFrames: false,
-                expectedLastStreamId: 0,
+                expectedLastStreamId: 1,
                 expectedErrorCode: Http2ErrorCode.PROTOCOL_ERROR,
                 expectedErrorMessage: CoreStrings.FormatHttp2ErrorHeadersInterleaved(Http2FrameType.PRIORITY, streamId: 1, headersStreamId: 1));
         }
@@ -2190,7 +2204,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Tests
 
             await WaitForConnectionErrorAsync<Http2ConnectionErrorException>(
                 ignoreNonGoAwayFrames: false,
-                expectedLastStreamId: 0,
+                expectedLastStreamId: 1,
                 expectedErrorCode: Http2ErrorCode.PROTOCOL_ERROR,
                 expectedErrorMessage: CoreStrings.FormatHttp2ErrorHeadersInterleaved(Http2FrameType.RST_STREAM, streamId: 1, headersStreamId: 1));
         }
@@ -2262,7 +2276,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Tests
 
             await WaitForConnectionErrorAsync<Http2ConnectionErrorException>(
                 ignoreNonGoAwayFrames: false,
-                expectedLastStreamId: 0,
+                expectedLastStreamId: 1,
                 expectedErrorCode: Http2ErrorCode.PROTOCOL_ERROR,
                 expectedErrorMessage: CoreStrings.FormatHttp2ErrorHeadersInterleaved(Http2FrameType.SETTINGS, streamId: 0, headersStreamId: 1));
         }
@@ -2369,7 +2383,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Tests
 
             await WaitForConnectionErrorAsync<Http2ConnectionErrorException>(
                 ignoreNonGoAwayFrames: false,
-                expectedLastStreamId: 0,
+                expectedLastStreamId: 1,
                 expectedErrorCode: Http2ErrorCode.PROTOCOL_ERROR,
                 expectedErrorMessage: CoreStrings.FormatHttp2ErrorHeadersInterleaved(Http2FrameType.PING, streamId: 0, headersStreamId: 1));
         }
@@ -2417,23 +2431,47 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Tests
         }
 
         [Fact]
-        public async Task GOAWAY_Received_AbortsAllStreams()
+        public async Task GOAWAY_Received_SetsConnectionStateToClosingAndWaitForAllStreamsToComplete()
         {
-            await InitializeConnectionAsync(_waitForAbortApplication);
+            await InitializeConnectionAsync(_echoApplication);
 
             // Start some streams
-            await StartStreamAsync(1, _browserRequestHeaders, endStream: true);
-            await StartStreamAsync(3, _browserRequestHeaders, endStream: true);
-            await StartStreamAsync(5, _browserRequestHeaders, endStream: true);
+            await StartStreamAsync(1, _browserRequestHeaders, endStream: false);
+            await StartStreamAsync(3, _browserRequestHeaders, endStream: false);
 
             await SendGoAwayAsync();
 
-            await WaitForConnectionStopAsync(expectedLastStreamId: 5, ignoreNonGoAwayFrames: true);
+            await _closingStateReached.Task.DefaultTimeout();
 
-            await WaitForAllStreamsAsync();
-            Assert.Contains(1, _abortedStreamIds);
-            Assert.Contains(3, _abortedStreamIds);
-            Assert.Contains(5, _abortedStreamIds);
+            await SendDataAsync(1, _helloBytes, true);
+            await ExpectAsync(Http2FrameType.HEADERS,
+                withLength: 37,
+                withFlags: (byte)Http2HeadersFrameFlags.END_HEADERS,
+                withStreamId: 1);
+            await ExpectAsync(Http2FrameType.DATA,
+                withLength: 5,
+                withFlags: (byte)Http2DataFrameFlags.NONE,
+                withStreamId: 1);
+            await ExpectAsync(Http2FrameType.DATA,
+                withLength: 0,
+                withFlags: (byte)Http2DataFrameFlags.END_STREAM,
+                withStreamId: 1);
+            await SendDataAsync(3, _helloBytes, true);
+            await ExpectAsync(Http2FrameType.HEADERS,
+                withLength: 37,
+                withFlags: (byte)Http2HeadersFrameFlags.END_HEADERS,
+                withStreamId: 3);
+            await ExpectAsync(Http2FrameType.DATA,
+                withLength: 5,
+                withFlags: (byte)Http2DataFrameFlags.NONE,
+                withStreamId: 3);
+            await ExpectAsync(Http2FrameType.DATA,
+                withLength: 0,
+                withFlags: (byte)Http2DataFrameFlags.END_STREAM,
+                withStreamId: 3);
+
+            await WaitForConnectionStopAsync(expectedLastStreamId: 3, ignoreNonGoAwayFrames: false);
+            await _closedStateReached.Task.DefaultTimeout();
         }
 
         [Fact]
@@ -2526,7 +2564,10 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Tests
                 withFlags: (byte)Http2HeadersFrameFlags.END_HEADERS,
                 withStreamId: 5);
 
-            await SendGoAwayAsync();
+            // Close all pipes and wait for response to drain
+            _pair.Application.Output.Complete();
+            _pair.Transport.Input.Complete();
+            _pair.Transport.Output.Complete();
 
             await WaitForConnectionStopAsync(expectedLastStreamId: 5, ignoreNonGoAwayFrames: false);
 
@@ -2600,7 +2641,10 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Tests
             await VerifyStreamBackpressure(3);
             await VerifyStreamBackpressure(5);
 
-            await SendGoAwayAsync();
+            // Close all pipes and wait for response to drain
+            _pair.Application.Output.Complete();
+            _pair.Transport.Input.Complete();
+            _pair.Transport.Output.Complete();
 
             await WaitForConnectionStopAsync(expectedLastStreamId: 5, ignoreNonGoAwayFrames: false);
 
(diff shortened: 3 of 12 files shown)


src/Kestrel.Core/Internal/Http2/Http2Connection.cs
```diff
@@ -95,19 +139,34 @@ public Http2Connection(Http2ConnectionContext context)
...
+            }
             _frameWriter.Complete();
         }
 
         public void Abort(ConnectionAbortedException ex)
         {
-            _stopping = true;
+            if (_stateManager.Exchange(ConnectionState.Closed) != ConnectionState.Closed)
+            {
+                _frameWriter.WriteGoAwayAsync(_highestOpenedStreamId, Http2ErrorCode.INTERNAL_ERROR).ConfigureAwait(false).GetAwaiter().GetResult();
```
JunTaoLuo: Alternatively, we need to convert everything to async, which means Http1Connection needs to change as well.

src/Kestrel.Core/Internal/Http2/Http2Connection.cs
```diff
@@ -43,6 +43,32 @@ private enum PseudoHeaderFields
             Unknown = 0x40000000
         }
 
+        internal enum ConnectionState
```
JunTaoLuo: Internal for testing.

src/Kestrel.Core/Internal/Http2/Http2Connection.cs
```diff
@@ -771,7 +841,7 @@ private void StartStream<TContext>(IHttpApplication<TContext> application)
 
         private void ResetRequestHeaderParsingState()
         {
-            if (_requestHeaderParsingState != RequestHeaderParsingState.Trailers)
+            if (_requestHeaderParsingState != RequestHeaderParsingState.Trailers && _stateManager.State != ConnectionState.Closed)
             {
                 _highestOpenedStreamId = _currentHeadersStream.StreamId;
```
JunTaoLuo: Hmm there's still a race here.

JunTaoLuo: Might be easier to use locks instead of Interlocked since otherwise we need to combine state and highest stream id into a single field.

src/Kestrel.Core/Internal/Http2/Http2Connection.cs
```diff
@@ -751,6 +815,12 @@ private Task DecodeTrailersAsync(bool endHeaders, Span<byte> payload)
 
         private void StartStream<TContext>(IHttpApplication<TContext> application)
         {
+            if (_stateManager.State == ConnectionState.Closed)
```
JunTaoLuo: StartStream and ResetRequestHeaderParsingState must be completed atomically

test/Kestrel.Core.Tests/Http2ConnectionTests.cs
```diff
@@ -2061,6 +2061,17 @@ public Task HEADERS_Received_HeaderBlockContainsTEHeader_ValueIsNotTrailers_Stre
 
             await SendGoAwayAsync();
 
+            var spinLimit = 50; // 2.5 seconds
+            while (_connection.State == Http2Connection.ConnectionState.Open && spinLimit --> 0)
```
Tratcher: Is that `spinLimit-- > 0`?

test/Kestrel.Core.Tests/Http2ConnectionTests.cs
```diff
@@ -2061,6 +2061,17 @@ public Task HEADERS_Received_HeaderBlockContainsTEHeader_ValueIsNotTrailers_Stre
...
+            {
+                await Task.Delay(50);
+            }
+            Assert.Equal(Http2Connection.ConnectionState.Closing, _connection.State);
+
+            await SendRstStreamAsync(1);
```
Tratcher: Change this test to use graceful request completions. E.g. use a different app that waits for data before responding, then send a final data frame after the go away.

test/Kestrel.Core.Tests/Http2ConnectionTests.cs
```diff
@@ -2061,6 +2061,17 @@ public Task HEADERS_Received_HeaderBlockContainsTEHeader_ValueIsNotTrailers_Stre
...
+            await SendRstStreamAsync(3);
+            await SendRstStreamAsync(5);
+
             await WaitForConnectionStopAsync(expectedLastStreamId: 5, ignoreNonGoAwayFrames: true);
```
Tratcher: ignoreNonGoAwayFrames: false, read each expected response.

src/Kestrel.Core/Internal/Http2/Http2Connection.cs
halter73: I'm fine fire and forgetting the writing small frames. This isn't the only place we do it.

Nit: In Kestrel, we don't bother calling ConfigureAwait(false) before calling GetAwaiter().GetResult(). I don't see how it would do much since there's no continuation to post anyway.

Where we do use ConfigureAwait(false) is in the KestrelServer.StartAsync/StopAsync path because xunit's SynchronizationContext causes problems otherwise.

src/Kestrel.Core/Internal/Http2/Http2Connection.cs
```diff
@@ -125,12 +184,12 @@ public void StopProcessingNextRequest()
...
                 }
 
-                if (!_stopping)
+                if (_stateManager.State != ConnectionState.Closed)
                 {
                     await _frameWriter.WriteSettingsAsync(_serverSettings);
                 }
 
-                while (!_stopping)
+                while (_stateManager.State == ConnectionState.Open || (_stateManager.State == ConnectionState.Closing) && !_streams.IsEmpty)
```
halter73: Nit: add an extra parenthesis around the `&&` part of the condition for clarity.

src/Kestrel.Core/Internal/Http2/Http2Connection.cs
```diff
@@ -205,7 +265,10 @@ public void StopProcessingNextRequest()
                         stream.Abort(connectionError);
                     }
 
-                    await _frameWriter.WriteGoAwayAsync(_highestOpenedStreamId, errorCode);
+                    if (previousConnectionState != ConnectionState.Closed)
+                    {
+                        await _frameWriter.WriteGoAwayAsync(_highestOpenedStreamId, errorCode);
+                    }
```
halter73: Super nit: add extra newline after `}` unless it's followed by an else/catch/finally or another `}`. This applies to a lot of other lines in this PR too.

src/Kestrel.Core/Internal/Http2/Http2Connection.cs
halter73: Is it possible for this condition to be true for a well-behaved client? I'm guessing we've already sent a goaway frame at this point, so there isn't too much point in sending another. But if there's an unavoidable where even a well-behaved client can attempt to start a new stream while we're in the closed state, just not tracking the connection seems problematic. Any subsequent data frames sent by the client for that stream would cause the server to terminate the connection with an error immediately.

It also appears _streams could go from non-empty to empty back to non-empty all while the connection is the closing state. This would allow to connection to enter the closed state with a non-empty streams dictionary which seems problematic.

src/Kestrel.Core/Internal/Http2/Http2Connection.cs
halter73: :+1: for using locks for connection state transitions. I don't suspect these will be high contention.

test/Kestrel.Core.Tests/Http2ConnectionTests.cs
JunTaoLuo: yes but I liked the aesthetics of the "goes to" operator

src/Kestrel.Core/Internal/Http2/Http2Frame.GoAway.cs
```diff
@@ -7,7 +7,7 @@ public partial class Http2Frame
     {
         public int GoAwayLastStreamId
         {
-            get => (Payload[0] << 24) | (Payload[1] << 16) | (Payload[2] << 16) | Payload[3];
+            get => (Payload[0] << 24) | (Payload[1] << 16) | (Payload[2] << 8) | Payload[3];
```
halter73: This is the second time I've seen this bug. I did a search for all instances of `<< 16.+<< 16` in our code base in a single line, and it seems like there's a another instance of this bug in Http2Frame.RstStream.cs in RstStreamErrorCode. Might was well fix it there too.

test/Kestrel.Core.Tests/Http2ConnectionTests.cs
```diff
@@ -2061,6 +2061,17 @@ public Task HEADERS_Received_HeaderBlockContainsTEHeader_ValueIsNotTrailers_Stre
 
             await SendGoAwayAsync();
 
+            var spinLimit = 50; // 2.5 seconds
```
halter73: It might be slightly better/faster to hook to a state change event. In some other Kestrel tests, we just set a TCS when we see a specific log indicating a state transition. A test-only callback might also be acceptable.

test/Kestrel.Core.Tests/Http2ConnectionTests.cs
```diff
@@ -2161,6 +2208,10 @@ public Task HEADERS_Received_HeaderBlockContainsTEHeader_ValueIsNotTrailers_Stre
 
             await SendGoAwayAsync();
 
+            await SendRstStreamAsync(1);
+            await SendRstStreamAsync(3);
+            await SendRstStreamAsync(5);
```
halter73: The point of this test is that a connection-level abort _by itself_ is enough to complete any calls to Response.Body.WriteAsync that were waiting on the connection output flow control window to update that. This is no longer testing that with the extra stream reset frames.

test/Kestrel.Core.Tests/Http2ConnectionTests.cs
```diff
@@ -2904,7 +3114,7 @@ public Task HEADERS_Received_HeaderBlockContainsTEHeader_ValueIsNotTrailers_Stre
                 withStreamId: 0);
         }
 
-        private async Task StartStreamAsync(int streamId, IEnumerable<KeyValuePair<string, string>> headers, bool endStream)
+        private async Task StartStreamAsync(int streamId, IEnumerable<KeyValuePair<string, string>> headers, bool endStream, bool trackAsRunningStream = true)
```
halter73: Nit: new params seems to be unused.

src/Kestrel.Core/Internal/Http2/Http2Connection.cs
halter73: The way this is written, it seem like theres a race where _streams could go from non-empty to empty back to non-empty all while the connection is the closing state. This would allow to connection to enter the closed state with a non-empty streams dictionary which seems problematic.

test/Kestrel.Core.Tests/Http2ConnectionTests.cs
halter73: :+1: This was there before. I haven't double check why, but generally speaking you should have a very good reason to use `ignoreNonGoAwayFrames: true`. It's usually much better to assert exactly which frames you expect.

test/Kestrel.Core.Tests/Http2ConnectionTests.cs
JunTaoLuo: State changes currently are not logged and I don't think there's a good way of adding a test-only callback that doesn't at least add a tiny overhead. I'll see what it looks like.

test/Kestrel.Core.Tests/Http2ConnectionTests.cs
JunTaoLuo: Sending a GOAWAY isn't by itself enough enough to signify a shutdown of an Http2 connection. By the phrasing in the RFC, it's a signal that there's an intention to shutdown the connection in the near future. A connection-level abort seems to be simply closing the transport-level connection. Should I simulate that here instead?

Tratcher: Yeah, close the pipe.

src/Kestrel.Core/Internal/Http2/Http2Connection.cs
JunTaoLuo: This is possible. For example, the client is opening streams and sending new requests on them while the server is in the process of shutdown (via ctrl-c). Suppose the shutdown was completed relatively quickly on the server, there's a chance that by the time the client's request for new streams will arrive when the connection is already closed. This is a classic race condition described in the HTTP/2 spec. The mitigation is that when we close the connection, we need to send a final GOAWAY that has the highest frame number set. This way, the client that receives the GOAWAY will be able to tell which frames need to be retried.

As for the second situation, _streams should never be able to go from empty to non-empty. Once the connection reaches the Closing state and _streams become empty, the state will be changed to Closed. During the Closed state, no new streams can be started.

halter73: > As for the second situation, _streams should never be able to go from empty to non-empty. Once the connection reaches the Closing state and _streams become empty, the state will be changed to Closed. During the Closed state, no new streams can be started.

I don't see how you can guarantee this without a lock.

halter73: Never mind. I see this is now called with the lock acquired.

src/Kestrel.Core/Internal/Http2/Http2Connection.cs
```diff
@@ -95,19 +120,51 @@ public Http2Connection(Http2ConnectionContext context)
...
+            lock (_connectionStateLock)
+            {
+                if (_connectionState == ConnectionState.Open)
+                {
+                    if (sendGoAway)
+                    {
+                        _frameWriter.WriteGoAwayAsync(Int32.MaxValue, Http2ErrorCode.NO_ERROR);
+                    }
+
+                    _connectionState = ConnectionState.Closing;
```
halter73: So another goaway frame will be written once the last active stream completes, right Could you move directly to the closed state (and maybe write a second goaway frame immediately) in the case there aren't any active streams at this point?

src/Kestrel.Core/Internal/Http2/Http2Connection.cs
```diff
@@ -125,12 +182,12 @@ public async Task ProcessRequestsAsync<TContext>(IHttpApplication<TContext> appl
...
                 }
 
-                if (!_stopping)
+                if (_connectionState != ConnectionState.Closed)
                 {
                     await _frameWriter.WriteSettingsAsync(_serverSettings);
                 }
 
-                while (!_stopping)
+                while (_connectionState == ConnectionState.Open || (_connectionState == ConnectionState.Closing && !_streams.IsEmpty))
```
halter73: Since you are predicating this condition both on the connection state and the emptiness of the stream, you need to an acquire a lock in order to prevent a race where the stream count goes from zero to non-zero while the connection is in the closing state.

The two ways you could do this that I see is by either putting `_connectionState == ConnectionState.Closing && !_streams.IsEmpty` inside a property and acquiring the lock. That seems pretty heavy handed for something that's checked every frame though.

It's probably better to just move `_streams.TryRemove(streamId, out _);` from OnStreamCompleted inside the lock. I think this would make it impossible for the stream count goes from zero to non-zero while the connection is in the closing state.f

src/Kestrel.Core/Internal/Http2/Http2Connection.cs
JunTaoLuo: Yea before I switched to the implementation using locks, it there was a race here.

src/Kestrel.Core/Internal/Http2/Http2Connection.cs
JunTaoLuo: Good point.

src/Kestrel.Core/Internal/Http2/Http2Connection.cs
```diff
@@ -722,12 +788,28 @@ private Task DecodeHeadersAsync<TContext>(IHttpApplication<TContext> application
...
+
+                        ResetRequestHeaderParsingState();
+                    }
                 }
             }
             catch (Http2StreamErrorException)
             {
+                lock (_connectionStateLock)
+                {
+                    if (_connectionState != ConnectionState.Closed)
```
halter73: Why do you acquire the lock twice in this method? Wouldn't it read better to acquire a lock around the entire method body and go back to setting _highestOpenedStreamId in ResetRequestHeaderParsingState()? Is it even important to call ResetRequestHeaderParsingState() if the connection is already in the closed state?

src/Kestrel.Core/Internal/Http2/Http2Connection.cs
halter73: I just thought about this a bit more. Why not make this just `while (_connectionState != ConnectionState.Closed)` since we call CancelPendingRead in OnStreamCompleted anyway? I still like calling `_streams.TryRemove` inside the lock, but it shouldn't be relevant for this condition if we simplify it.

src/Kestrel.Core/Internal/Http2/Http2Connection.cs
JunTaoLuo: Trying to keep the locked operations as few as possible. I don't see why resetting the state would be necessary in a closed state but it seems like the safe thing to do.

src/Kestrel.Core/Internal/Http2/Http2Connection.cs
```diff
@@ -85,6 +92,24 @@ public Http2Connection(Http2ConnectionContext context)
             _hpackDecoder = new HPackDecoder((int)_serverSettings.HeaderTableSize);
         }
 
+        // internal for testing
+        internal ConnectionState State
```
Tratcher: => _connectionState;

src/Kestrel.Core/Internal/Http2/Http2Connection.cs
```diff
@@ -95,19 +120,59 @@ public Http2Connection(Http2ConnectionContext context)
 
         public void OnInputOrOutputCompleted()
         {
-            _stopping = true;
+            lock (_connectionStateLock)
+            {
+                if (_connectionState != ConnectionState.Closed)
+                {
+                    _frameWriter.WriteGoAwayAsync(_highestOpenedStreamId, Http2ErrorCode.NO_ERROR);
```
Tratcher: Do you really want to do IO under the lock? Or just change the state and set a bool to send the GoAway?

src/Kestrel.Core/Internal/Http2/Http2Connection.cs
```diff
@@ -95,19 +120,59 @@ public Http2Connection(Http2ConnectionContext context)
...
+        public void StopProcessingNextRequest(bool sendGoAway = false)
         {
-            _stopping = true;
+            lock (_connectionStateLock)
+            {
+                if (_connectionState == ConnectionState.Open)
+                {
+                    if (_streams.IsEmpty)
+                    {
+                        _frameWriter.WriteGoAwayAsync(_highestOpenedStreamId, Http2ErrorCode.NO_ERROR);
```
Tratcher: Why not honor sendGoAway = false on this code path?

JunTaoLuo: Only the graceful GOAWAY (with the highest opened stream id set to maxvalue) is optional. The final GOAWAY we send before closing the connection is a must.

src/Kestrel.Core/Internal/Http2/Http2Connection.cs
JunTaoLuo: These are fire and forget so it shouldn't cause any deadlocks. Was that your concern?

src/Kestrel.Core/Internal/Http2/Http2Connection.cs
Tratcher: so sendGoAway would be true in that case? For consistency in this method you should still check it.

src/Kestrel.Core/Internal/Http2/Http2Connection.cs
Tratcher: Wanting to hold the lock as briefly as possible to avoid contention or deadlocks. If they can never block then it's probably fine.

src/Kestrel.Core/Internal/Http2/Http2Connection.cs
halter73: I don't mind resetting the state in the closed state so much. Acquiring the lock twice in the same method seems unnecessary in this case though. Just lock around the whole thing.

src/Kestrel.Core/Internal/Http2/Http2Connection.cs
JunTaoLuo: Not quite, we set sendgoaway to false when processing a GOAWAY from the client. In this case, we don't want to respond by sending a graceful GOAWAY if there are streams still open. However, if there were no streams open, we want to close the connection right away and send the final GOAWAY with the highest opened stream id regardless of whether sendgoaway was set to false.

Tratcher: Then sendGoAway is misnamed and you should come up with something more specific.

JunTaoLuo:
:up::date: I think I've addressed all the feedback so far and fixed up all the tests.

test/Kestrel.Core.Tests/Http2ConnectionTests.cs
```diff
@@ -2050,23 +2059,76 @@ public async Task GOAWAY_Received_ConnectionStops()
...
+            await ExpectAsync(Http2FrameType.DATA,
+                withLength: 5,
+                withFlags: (byte)Http2DataFrameFlags.NONE,
+                withStreamId: 3);
+            await ExpectAsync(Http2FrameType.DATA,
+                withLength: 0,
+                withFlags: (byte)Http2DataFrameFlags.END_STREAM,
+                withStreamId: 3);
+
+            await WaitForConnectionStopAsync(expectedLastStreamId: 3, ignoreNonGoAwayFrames: false);
```
Tratcher: Assert Closed?

test/Kestrel.Core.Tests/Http2ConnectionTests.cs
```diff
@@ -2050,23 +2059,76 @@ public async Task GOAWAY_Received_ConnectionStops()
...
+            await InitializeConnectionAsync(_waitForAbortApplication);
+
+            _connection.OnInputOrOutputCompleted();
+
+            Assert.True(_connection.StateTransitionEvents[Http2Connection.ConnectionState.Closed].Wait(5000));
+
+            await SendGoAwayAsync();
+
+            Assert.Equal(Http2Connection.ConnectionState.Closed, _connection.State);
+            Assert.False(_connection.StateTransitionEvents[Http2Connection.ConnectionState.Closing].Wait(5000));
```
Tratcher: Does this test always take at least 5 seconds? Make this less than 1 second.

test/Kestrel.Core.Tests/Http2ConnectionTests.cs
```diff
@@ -2050,23 +2059,76 @@ public async Task GOAWAY_Received_ConnectionStops()
...
+            await InitializeConnectionAsync(_waitForAbortApplication);
+
+            _connection.StopProcessingNextRequest();
+
+            Assert.True(_connection.StateTransitionEvents[Http2Connection.ConnectionState.Closed].Wait(5000));
+
+            await SendGoAwayAsync();
+
+            Assert.Equal(Http2Connection.ConnectionState.Closed, _connection.State);
+            Assert.False(_connection.StateTransitionEvents[Http2Connection.ConnectionState.Closing].Wait(5000));
```
Tratcher: Shorter

test/Kestrel.Core.Tests/Http2ConnectionTests.cs
```diff
@@ -2886,6 +2954,158 @@ public async Task ConnectionResetNotLoggedWithNoActiveStreams()
...
+
+            Assert.Equal(Http2Connection.ConnectionState.Open, _connection.State);
+            Assert.Equal(1, _connection.HighestOpenedStreamId);
+
+            _connection.StopProcessingNextRequest();
+            Assert.True(_connection.StateTransitionEvents[Http2Connection.ConnectionState.Closing].Wait(5000));
+
+            VerifyGoAway(await ReceiveFrameAsync(), Int32.MaxValue, Http2ErrorCode.NO_ERROR);
+
+            await SendRstStreamAsync(1);
```
Tratcher: Complete gracefully

test/Kestrel.Core.Tests/Http2ConnectionTests.cs
```diff
@@ -2886,6 +2954,158 @@ public async Task ConnectionResetNotLoggedWithNoActiveStreams()
...
+            Assert.Equal(1, _connection.HighestOpenedStreamId);
+
+            await SendGoAwayAsync();
+
+            Assert.True(_connection.StateTransitionEvents[Http2Connection.ConnectionState.Closing].Wait(5000));
+
+            await StartStreamAsync(3, _browserRequestHeaders, endStream: false);
+
+            Assert.Equal(Http2Connection.ConnectionState.Closing, _connection.State);
+            Assert.False(_connection.StateTransitionEvents[Http2Connection.ConnectionState.Closed].Wait(5000));
```
Tratcher: Shorter

test/Kestrel.Core.Tests/Http2ConnectionTests.cs
```diff
@@ -2886,6 +2954,158 @@ public async Task ConnectionResetNotLoggedWithNoActiveStreams()
...
+
+        [Fact]
+        public async Task RejectNewStreamsDuringClosedConnection()
+        {
+            await InitializeConnectionAsync(_echoApplication);
+
+            await StartStreamAsync(1, _browserRequestHeaders, endStream: false);
+
+            Assert.Equal(Http2Connection.ConnectionState.Open, _connection.State);
+            Assert.Equal(1, _connection.HighestOpenedStreamId);
```
Tratcher: Is this racy?

test/Kestrel.Core.Tests/Http2ConnectionTests.cs
```diff
@@ -2886,6 +2954,158 @@ public async Task ConnectionResetNotLoggedWithNoActiveStreams()
...
+        public async Task AcceptNewStreamsDuringClosingConnection()
+        {
+            await InitializeConnectionAsync(_echoApplication);
+
+            await StartStreamAsync(1, _browserRequestHeaders, endStream: false);
+
+            Assert.Equal(Http2Connection.ConnectionState.Open, _connection.State);
+            Assert.Equal(1, _connection.HighestOpenedStreamId);
+
+            await SendGoAwayAsync();
```
Tratcher: It still seems bizzar for the client to start new streams after sending a GoAway. "The GOAWAY frame (type=0x7) is used to initiate shutdown of a connection or to signal serious error conditions.". You didn't do anything specific to allow it, correct? You just didn't disallow it?

test/Kestrel.Core.Tests/Http2ConnectionTests.cs
```diff
@@ -2886,6 +2954,158 @@ public async Task ConnectionResetNotLoggedWithNoActiveStreams()
...
+            await ExpectAsync(Http2FrameType.DATA,
+                withLength: 0,
+                withFlags: (byte)Http2DataFrameFlags.END_STREAM,
+                withStreamId: 3);
+
+            await WaitForConnectionStopAsync(expectedLastStreamId: 3, ignoreNonGoAwayFrames: false);
+        }
+
+        [Fact]
+        public async Task RejectNewStreamsDuringClosedConnection()
```
Tratcher: I don't see any Rejection here. Ignore?

test/Kestrel.FunctionalTests/Http2/ShutdownTests.cs
```diff
@@ -0,0 +1,197 @@
...
+            }))
+            {
+                // Setup a listening stream
+                var connection = server.CreateConnection();
+                var sslStream = new SslStream(connection.Stream);
+                await sslStream.AuthenticateAsClientAsync(new SslClientAuthenticationOptions()
+                {
+                    TargetHost = "localhost",
+                    RemoteCertificateValidationCallback = (_, __, ___, ____) => true,
+                    ApplicationProtocols = new List<SslApplicationProtocol>() { SslApplicationProtocol.Http2, SslApplicationProtocol.Http11 },
```
Tratcher: Remove Http11

test/Kestrel.FunctionalTests/Http2/ShutdownTests.cs
```diff
@@ -0,0 +1,197 @@
...
+                kestrelOptions.Listen(IPAddress.Loopback, 0, listenOptions =>
+                {
+                    listenOptions.Protocols = HttpProtocols.Http2;
+                    listenOptions.UseHttps(_x509Certificate2);
+                });
+            }))
+            {
+                // Setup a listening stream
+                var connection = server.CreateConnection();
+                var sslStream = new SslStream(connection.Stream);
```
Tratcher: Extract duplicate setup code

test/Kestrel.FunctionalTests/Http2/ShutdownTests.cs
```diff
@@ -0,0 +1,197 @@
...
+
+                var stopTask = server.StopAsync();
+
+                // Unblock the request
+                requestUnblocked.Set();
+
+                Assert.Equal("hello world HTTP/2", await requestTask);
+                Assert.Equal(stopTask, await Task.WhenAny(stopTask, Task.Delay(10000)));
+
+                await ReceiveFramesAsync(reader);
```
Tratcher: ? How are you intercepting a raw frame here? Shouldn't the server have read that?

test/Kestrel.FunctionalTests/Http2/ShutdownTests.cs
```diff
@@ -0,0 +1,197 @@
...
+                var reader = PipeReaderFactory.CreateFromStream(PipeOptions.Default, sslStream, CancellationToken.None);
+
+                var requestTask = Client.GetStringAsync($"https://localhost:{server.Port}/");
+                Assert.False(requestTask.IsCompleted);
+                Assert.True(requestStarted.Wait(5000), "timeout");
+
+                var stopTask = server.StopAsync();
+
+                // Keep the request blocked
+                Assert.False(requestUnblocked.Wait(6000), "request unblocked");
```
Tratcher: Remove, there's no reason to wait on this.

src/Kestrel.Core/Internal/Http2/Http2Connection.cs
```diff
@@ -195,7 +252,16 @@ public async Task ProcessRequestsAsync<TContext>(IHttpApplication<TContext> appl
             }
             finally
             {
-                 var connectionError = error as ConnectionAbortedException
+                lock (StateLock)
+                {
+                    if (State != ConnectionState.Closed)
+                    {
+                        _frameWriter.WriteGoAwayAsync(HighestOpenedStreamId, errorCode);
```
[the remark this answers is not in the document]
JunTaoLuo: Good catch, it can.

src/Kestrel.Core/Internal/Http2/Http2Connection.cs
```diff
@@ -85,6 +93,15 @@ public Http2Connection(Http2ConnectionContext context)
             _hpackDecoder = new HPackDecoder((int)_serverSettings.HeaderTableSize);
         }
 
+        // internal for testing
+        internal Dictionary<ConnectionState, ManualResetEventSlim> StateTransitionEvents { get; set; }
```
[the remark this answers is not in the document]
halter73: 👍 For the action over the dictionary. Better yet, just test behavior instead of verifying arbitrary internal state transitions happen.

I also almost always recommend using TaskCompletionSource with TaskCreationOptions.RunContinuationsAsynchronously instead of ManualResetEvents since the latter encourages blocking using Wait() which can lead to real problems on some CI agents with low core counts. It's better to use TCS.Task.TimeoutAfter()/DefaultTimeout(.)

src/Kestrel.Core/Internal/Http2/Http2Connection.cs
```diff
@@ -722,12 +788,28 @@ private Task DecodeHeadersAsync<TContext>(IHttpApplication<TContext> application
 
                 if (endHeaders)
                 {
-                    StartStream(application);
+                    lock (StateLock)
+                    {
+                        if (State != ConnectionState.Closed)
+                        {
+                            StartStream(application);
+                            HighestOpenedStreamId = _currentHeadersStream.StreamId;
```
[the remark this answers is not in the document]
halter73: That's what I've been trying to say. The two lock blocks are unnecessary if we just assign first.

test/Kestrel.Core.Tests/Http2ConnectionTests.cs
```diff
@@ -2886,6 +2954,158 @@ public async Task ConnectionResetNotLoggedWithNoActiveStreams()
...
+
+        [Fact]
+        public async Task GracefulGOAWAYSentByStopProcessingNextRequest()
+        {
+            await InitializeConnectionAsync(_waitForAbortApplication);
+
+            await StartStreamAsync(1, _browserRequestHeaders, endStream: true);
+
+            Assert.Equal(Http2Connection.ConnectionState.Open, _connection.State);
+            Assert.Equal(1, _connection.HighestOpenedStreamId);
```
halter73: Your's seem to be the only Http2ConnectionTests facts that access _connection directly. I don't think it's good to verify internal state. If it's important that certain state transitions happen before continuing the test to avoid races, I suggest hooking the logger for this purpose.

src/Kestrel.Core/Internal/Http2/Http2Connection.cs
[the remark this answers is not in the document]
JunTaoLuo: Ah fair enough, reordering would work, I must have misunderstood your previous comment.

test/Kestrel.Core.Tests/Http2ConnectionTests.cs
JunTaoLuo: Yea and it's unnecessary since it will be checked via goaway frames sent later on.

src/Kestrel.Core/Internal/Http2/Http2Connection.cs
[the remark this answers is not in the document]
JunTaoLuo: Now I recall why I couldn't do this. We need to assign the HighestOpenedStreamId when _hpackDecoder throws stream errors. In those cases, we still need need to atomically check the state and assign the stream id. I don't see how we can simplify since in the non-error case we need an atomic check between state, stream id and starting the stream.

src/Kestrel.Core/Internal/Http2/Http2Connection.cs
```diff
@@ -95,19 +112,59 @@ public Http2Connection(Http2ConnectionContext context)
...
+                            _frameWriter.WriteGoAwayAsync(Int32.MaxValue, Http2ErrorCode.NO_ERROR);
+                        }
+
+                        UpdateState(ConnectionState.Closing);
+                    }
+                }
+            }
+
+            // Wake up request processing loop so the connection can complete if there are no pending requests
             Input.CancelPendingRead();
```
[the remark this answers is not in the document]
JunTaoLuo: Why is that necessary? The request processing loop takes care of that.

test/Kestrel.Core.Tests/Http2ConnectionTests.cs
JunTaoLuo: The point here is to put the connection into a closing state. I did this via GOAWAY though I should have used StopProcessingNextRequest since that's more realistic. There's no functional difference though (other than an extra GOAWAY frame)

test/Kestrel.FunctionalTests/Http2/ShutdownTests.cs
JunTaoLuo: What do you mean the server should have read that? I open a stream and wait for the connection to send GOAWAYs as part of its lifecycle.

test/Kestrel.Core.Tests/Http2ConnectionTests.cs
JunTaoLuo: Added.

JunTaoLuo:
:up::date: updated the tests based on feedback

src/Kestrel.Core/Internal/Http2/Http2Connection.cs
```diff
@@ -75,9 +75,9 @@ private enum PseudoHeaderFields
         private PseudoHeaderFields _parsedPseudoHeaderFields;
         private Http2HeadersFrameFlags _headerFlags;
         private bool _isMethodConnect;
+        private object StateLock = new object();
```
halter73: Nit: `private readonly object _stateLock = new object();`

src/Kestrel.Core/Internal/Http2/Http2Connection.cs
```diff
@@ -828,6 +888,18 @@ private void AbortStream(int streamId, ConnectionAbortedException error)
         void IHttp2StreamLifetimeHandler.OnStreamCompleted(int streamId)
         {
             _streams.TryRemove(streamId, out _);
```
halter73: Move inside lock.

test/Kestrel.Core.Tests/Http2ConnectionTests.cs
```diff
@@ -311,6 +313,10 @@ private void InitializeConnectionFields(MemoryPool<byte> memoryPool)
                 ServiceContext = new TestServiceContext()
                 {
                     Log = new TestKestrelTrace(_logger)
```
halter73: I prefer using `new Mock<KestrelTrace>(Logger) { CallBase = true }` instead of adding a bunch of properties to TestKestrelTrace.

test/Kestrel.FunctionalTests/Http2/ShutdownTests.cs
```diff
@@ -0,0 +1,126 @@
...
+        {
+            var requestStarted = new ManualResetEventSlim();
+            var requestUnblocked = new ManualResetEventSlim();
+            using (var server = new TestServer(async context =>
+            {
+                requestStarted.Set();
+                Assert.True(requestUnblocked.Wait(5000), "request timeout");
+                await context.Response.WriteAsync("hello world " + context.Request.Protocol);
+            }, new TestServiceContext(LoggerFactory)
+            ,
```
halter73: Nit: Instead of having `,` on it's own line move `new TestServiceContext(LoggerFactory)` down to this line.

test/Kestrel.FunctionalTests/Http2/ShutdownTests.cs
```diff
@@ -0,0 +1,126 @@
...
+
+        [ConditionalFact]
+        public async Task GracefulShutdownWaitsForRequestsToFinish()
+        {
+            var requestStarted = new ManualResetEventSlim();
+            var requestUnblocked = new ManualResetEventSlim();
+            using (var server = new TestServer(async context =>
+            {
+                requestStarted.Set();
+                Assert.True(requestUnblocked.Wait(5000), "request timeout");
```
halter73: Don't use ManualResetEvents in tests if they can be avoided. Here using a TaskCompletionSource should work just as well without blocking a thread.

test/Kestrel.FunctionalTests/Http2/ShutdownTests.cs
```diff
@@ -0,0 +1,126 @@
...
+                {
+                    listenOptions.Protocols = HttpProtocols.Http2;
+                    listenOptions.UseHttps(_x509Certificate2);
+                });
+            }))
+            {
+                var requestTask = Client.GetStringAsync($"https://localhost:{server.Port}/");
+                Assert.False(requestTask.IsCompleted);
+
+                Assert.True(requestStarted.Wait(5000), "timeout");
```
halter73: ditto. also use DefaultTimeout() once this is switched to a TCS. 5 seconds is too short of a wait even for something as simple as queuing to the threadpool on some CI agents.

test/Kestrel.FunctionalTests/Http2/ShutdownTests.cs
```diff
@@ -0,0 +1,126 @@
...
+namespace Microsoft.AspNetCore.Server.Kestrel.FunctionalTests.Http2
+{
+    [OSSkipCondition(OperatingSystems.MacOSX, SkipReason = "Missing SslStream ALPN support: https://github.com/dotnet/corefx/issues/30492")]
+    [OSSkipCondition(OperatingSystems.Linux, SkipReason = "Curl requires a custom install to support HTTP/2, see https://askubuntu.com/questions/884899/how-do-i-install-curl-with-http2-support")]
+    [MinimumOSVersion(OperatingSystems.Windows, WindowsVersions.Win10)]
+    public class ShutdownTests : TestApplicationErrorLoggerLoggedTest
+    {
+        private static X509Certificate2 _x509Certificate2 = TestResources.GetTestCertificate();
+
+        private HttpClient Client { get; set; }
```
halter73: I think it would be better to model these tests after TlsTests which doesn't use HttpClient but instead is written against the raw connection using the Http2Frame type. I would even consider doing this test without SSL so the tests can run on all platforms. This will also make it easier to switch to an in-memory test.

src/Kestrel.Core/Internal/Http2/Http2Connection.cs
```diff
@@ -729,12 +778,28 @@ private Task DecodeHeadersAsync<TContext>(IHttpApplication<TContext> application
...
             catch (Http2StreamErrorException)
             {
+                lock (StateLock)
+                {
+                    if (_state != Http2ConnectionState.Closed)
+                    {
+                        _highestOpenedStreamId = _currentHeadersStream.StreamId;
+                    }
+                }
+
```
halter73: The above is getting removed, right?

test/Kestrel.Core.Tests/Http2ConnectionTests.cs
```diff
@@ -2417,23 +2423,73 @@ public async Task GOAWAY_Received_ConnectionStops()
...
+                withLength: 0,
+                withFlags: (byte)Http2DataFrameFlags.END_STREAM,
+                withStreamId: 3);
+
+            await WaitForConnectionStopAsync(expectedLastStreamId: 3, ignoreNonGoAwayFrames: false);
+            await _closedStateReached.Task.DefaultTimeout();
+        }
+
+        [Fact]
+        public async Task GOAWAY_Received_DoesNotTransitionFromClosedToClosing()
```
Tratcher: These negative tests that have to time out are of dubious value. Fine find a way to check the state without a negative timeout or remove them.

test/Kestrel.Core.Tests/Http2ConnectionTests.cs
```diff
@@ -3253,6 +3315,131 @@ public async Task ConnectionResetNotLoggedWithNoActiveStreams()
...
+            await StartStreamAsync(1, _browserRequestHeaders, endStream: false);
+
+            _connection.StopProcessingNextRequest();
+            VerifyGoAway(await ReceiveFrameAsync(), Int32.MaxValue, Http2ErrorCode.NO_ERROR);
+
+            await _closingStateReached.Task.DefaultTimeout();
+
+            await StartStreamAsync(3, _browserRequestHeaders, endStream: false);
+
+            await Assert.ThrowsAsync<TimeoutException>(async () => await _closedStateReached.Task.TimeoutAfter(TimeSpan.FromSeconds(1)));
```
Tratcher: Remove

test/Kestrel.FunctionalTests/Http2/ShutdownTests.cs
Tratcher: I don't mind having an integration test for a scenario like shutdown to make sure a real client can get a graceful response before the connection closes.

halter73: For what possible reason would a real client be different than a fake one in this scenario? What's important is how the server behaves when the app is still running when shutdown starts. I don't mind having _some_ integration tests specifically for making sure HTTP/2 works with a real client, but when we're testing specific things we should write the tests in a way that will allow them to run with our in-memory test transport.

Tratcher:
Everything looks good, just need to verify if ShutdownTests is going to be an integration test or not.

test/Kestrel.FunctionalTests/Http2/ShutdownTests.cs
Tratcher: This is one of those cases where we'd want to make sure our shutdown behavior matches the client's expectations. E.g. that sending multiple go-aways is handled as expected, and that the response gets fully flushed before the connection closes.

halter73: Fair enough 👍 . In the future we might want to think about moving tests that don't test the transport itself, but do require a real client, in a separate functional test project that doesn't run per transport.

test/Kestrel.FunctionalTests/Http2/ShutdownTests.cs
```diff
@@ -0,0 +1,125 @@
...
+
+                var stopTask = server.StopAsync();
+
+                // Keep the request blocked
+                Assert.Equal(stopTask, await Task.WhenAny(stopTask, Task.Delay(10000)));
+            }
+
+            Assert.Contains(TestApplicationErrorLogger.Messages, m => m.Message.Contains("is closing."));
+            Assert.Contains(TestApplicationErrorLogger.Messages, m => m.Message.Contains("is closed. The last processed stream ID was 1."));
+            Assert.DoesNotContain(TestApplicationErrorLogger.Messages, m => m.Message.Contains("Request finished in"));
```
halter73: Does "Some connections failed to close gracefully during server shutdown." get logged in this test?

test/Kestrel.FunctionalTests/Http2/ShutdownTests.cs
```diff
@@ -0,0 +1,125 @@
...
+            _ => { }))
+            {
+                var requestTask = Client.GetStringAsync($"https://localhost:{server.Port}/");
+                Assert.False(requestTask.IsCompleted);
+                await requestStarted.Task.DefaultTimeout();
+
+                var stopTask = server.StopAsync();
+
+                // Keep the request blocked
+                Assert.Equal(stopTask, await Task.WhenAny(stopTask, Task.Delay(10000)));
```
halter73: `await server.StopAsync().DefaultTimeout()`.

JunTaoLuo:
Going to merge now and we can address the functional tests if we decide we want them to be different in the future. It adds about 2 seconds to the overall run which I think is acceptable.

PR #2774, opened 2018-08-03
Send Resets for unhandled app exceptions

Tratcher:
This was all that was required to address #2733 for response content-length mismatches. The length was already being tracked and verified at the HttpProtocol layer, but the resulting exceptions were not triggering a Reset.

The broader problem was that no unhandled application exception was triggering a Reset. I've added a new override point for this.

### DIFF
--- a/src/Kestrel.Core/Internal/Http2/Http2Stream.cs
+++ b/src/Kestrel.Core/Internal/Http2/Http2Stream.cs
@@ -362,10 +362,18 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http2
             AbortCore(abortReason);
         }
 
+        protected override void ErrorAfterResponseStarted()
+        {
+            // We can no longer change the response, send a Reset instead.
+            base.ErrorAfterResponseStarted();
+            var abortReason = new ConnectionAbortedException(CoreStrings.Http2StreamErrorAfterHeaders);
+            ResetAndAbort(abortReason, Http2ErrorCode.INTERNAL_ERROR);
+        }
+
         protected override void ApplicationAbort()
         {
             var abortReason = new ConnectionAbortedException(CoreStrings.ConnectionAbortedByApplication);
-            ResetAndAbort(abortReason, Http2ErrorCode.CANCEL);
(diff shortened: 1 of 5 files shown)


src/Kestrel.Core/Internal/Http2/Http2Stream.cs
```diff
@@ -362,10 +362,18 @@ public override void Abort(ConnectionAbortedException abortReason)
...
+            base.ErrorAfterHeaders();
+            var abortReason = new ConnectionAbortedException(CoreStrings.Http2StreamErrorAfterHeaders);
+            ResetAndAbort(abortReason, Http2ErrorCode.INTERNAL_ERROR);
+        }
+
         protected override void ApplicationAbort()
         {
             var abortReason = new ConnectionAbortedException(CoreStrings.ConnectionAbortedByApplication);
-            ResetAndAbort(abortReason, Http2ErrorCode.CANCEL);
+            ResetAndAbort(abortReason, Http2ErrorCode.INTERNAL_ERROR);
```
JunTaoLuo: Why the change here?

Tratcher: I was reviewing the description and decided Cancel wasn't appropriate for the server to send. Cancel  is intended for the request initiator (the client in this case).

test/Kestrel.Core.Tests/Http2StreamTests.cs
```diff
@@ -1063,6 +1063,233 @@ public async Task ContentLength_Received_MultipleDataFramesUnderSize_Reset()
...
+            var headersFrame = await ExpectAsync(Http2FrameType.HEADERS,
+                withLength: 55,
+                withFlags: (byte)Http2HeadersFrameFlags.END_HEADERS,
+                withStreamId: 1);
+            await ExpectAsync(Http2FrameType.DATA,
+                withLength: 0,
+                withFlags: (byte)Http2DataFrameFlags.END_STREAM,
+                withStreamId: 1);
+
+            Assert.Contains(_logger.Messages, m => m.Exception?.Message.Contains("App Faulted") ?? false);
```
halter73: Nit: Assert error severity.

Issue #2772, opened 2018-08-03
Test failure: RunIndividualTestCase


ryanbrandenburg:
This test [fails](http://aspnetci/viewLog.html?buildId=523153&buildTypeId=Releases_22xPublic_UbuntuUniverse) occasionally with the following error:
```
System.Exception : Test failed: http2/6.9.2; 6.9.2. Initial Flow-Control Window Size
   at Microsoft.AspNetCore.Server.Kestrel.FunctionalTests.H2SpecCommands.RunTest(String testId, Int32 port, Boolean https, ILogger logger) in /_/test/Kestrel.FunctionalTests/TestHelpers/H2SpecCommands.cs:line 211
   at Microsoft.AspNetCore.Server.Kestrel.FunctionalTests.Http2.H2SpecTests.RunIndividualTestCase(H2SpecTestCase testCase) in /_/test/Kestrel.FunctionalTests/Http2/H2SpecTests.cs:line 51
--- End of stack trace from previous location where exception was thrown ---
------- Stdout: -------
| [0.001s] TestLifetime Information: Starting test RunIndividualTestCase-http2/6.9.2/1, HTTPS:False, Changes SETTINGS_INITIAL_WINDOW_SIZE after sending HEADERS frame at 2018-08-03T06:35:28
| [0.003s] Microsoft.AspNetCore.Hosting.Internal.WebHost Debug: Hosting starting
| [0.008s] Microsoft.AspNetCore.Hosting.Internal.WebHost Debug: Hosting started
| [0.009s] Microsoft.AspNetCore.Hosting.Internal.WebHost Debug: Loaded hosting startup assembly Libuv.FunctionalTests
| [0.060s] Microsoft.AspNetCore.Server.Kestrel Debug: Connection id "0HLFP1CUJ1D0Q" started.
| [0.061s] Microsoft.AspNetCore.Server.Kestrel Debug: Connection id "0HLFP1CUJ1D0R" started.
| [0.063s] Microsoft.AspNetCore.Server.Kestrel Trace: Connection id 0HLFP1CUJ1D0R received SETTINGS frame with flags 0x0 and length 6 for stream ID 0
| [0.064s] Microsoft.AspNetCore.Server.Kestrel Trace: Connection id 0HLFP1CUJ1D0R received SETTINGS frame with flags 0x1 and length 0 for stream ID 0
[...]

PR #2790, opened 2018-08-08
Logging and settings ack

Tratcher:
I Improved request frame logging and added response frame logging to help debug #2772. Here's a sample of the new logs:

```
      Connection id "0HLFT9VMIK352" started.
trce: Microsoft.AspNetCore.Server.Kestrel[37]
      Connection id "0HLFT9VMIK352" sending SETTINGS frame for stream ID 0 with length 0 and flags NONE
trce: Microsoft.AspNetCore.Server.Kestrel[37]
      Connection id "0HLFT9VMIK352" received SETTINGS frame for stream ID 0 with length 18 and flags NONE
trce: Microsoft.AspNetCore.Server.Kestrel[37]
      Connection id "0HLFT9VMIK352" sending SETTINGS frame for stream ID 0 with length 0 and flags ACK
trce: Microsoft.AspNetCore.Server.Kestrel[37]
      Connection id "0HLFT9VMIK352" received WINDOW_UPDATE frame for stream ID 0 with length 4 and flags 0x0
trce: Microsoft.AspNetCore.Server.Kestrel[37]
      Connection id "0HLFT9VMIK352" received HEADERS frame for stream ID 1 with length 235 and flags END_STREAM, END_HEADERS, PRIORITY
trce: Microsoft.AspNetCore.Server.Kestrel[37]
      Connection id "0HLFT9VMIK352" received SETTINGS frame for stream ID 0 with length 0 and flags ACK
info: Microsoft.AspNetCore.Hosting.Internal.WebHost[1]
      Request starting HTTP/2 GET https://localhost:5001/
trce: Microsoft.AspNetCore.Server.Kestrel[37]
      Connection id "0HLFT9VMIK352" sending HEADERS frame for stream ID 1 with length 53 and flags END_HEADERS
trce: Microsoft.AspNetCore.Server.Kestrel[37]
      Connection id "0HLFT9VMIK352" sending DATA frame for stream ID 1 with length 19 and flags NONE
trce: Microsoft.AspNetCore.Server.Kestrel[37]
      Connection id "0HLFT9VMIK352" sending DATA frame for stream ID 1 with length 0 and flags END_STREAM
info: Microsoft.AspNetCore.Hosting.Internal.WebHost[2]
      Request finished in 41.2045ms 200
```

As for #2772... This test was previously no-opping but we changed the test app in a prior PR so it started running flaky. The test sets the stream window size to 0, starts a request, and then sets it 1. It's expecting a settings ack and then a data frame. There's a race where sometimes it gets the data frame before the ack and the test times out. I've fixed it by sending the ack before we update any stream windows.

### DIFF
--- a/samples/Http2SampleApp/Program.cs
+++ b/samples/Http2SampleApp/Program.cs
@@ -44,7 +44,7 @@ namespace Http2SampleApp
                     {
                         listenOptions.Protocols = HttpProtocols.Http1AndHttp2;
                         listenOptions.UseHttps("testCert.pfx", "testPassword");
-                        listenOptions.UseConnectionLogging();
+                        // listenOptions.UseConnectionLogging();
                         listenOptions.ConnectionAdapters.Add(new TlsFilterAdapter());
                     });
 
@@ -53,7 +53,7 @@ namespace Http2SampleApp
                     options.Listen(IPAddress.Any, basePort + 5, listenOptions =>
                     {
                         listenOptions.Protocols = HttpProtocols.Http2;
-                        listenOptions.UseConnectionLogging();
+                        // listenOptions.UseConnectionLogging();
                     });
                 })
                 .UseContentRoot(Directory.GetCurrentDirectory())
--- a/samples/Http2SampleApp/Program.cs
+++ b/samples/Http2SampleApp/Program.cs
@@ -36,7 +36,6 @@ namespace Http2SampleApp
                     options.Listen(IPAddress.Any, basePort, listenOptions =>
                     {
                         listenOptions.Protocols = HttpProtocols.Http1;
-                        listenOptions.UseConnectionLogging();
                     });
 
                     // TLS Http/1.1 or HTTP/2 endpoint negotiated via ALPN
@@ -44,7 +43,6 @@ namespace Http2SampleApp
                     {
                         listenOptions.Protocols = HttpProtocols.Http1AndHttp2;
                         listenOptions.UseHttps("testCert.pfx", "testPassword");
-                        // listenOptions.UseConnectionLogging();
                         listenOptions.ConnectionAdapters.Add(new TlsFilterAdapter());
                     });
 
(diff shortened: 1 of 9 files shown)


samples/Http2SampleApp/Program.cs
```diff
@@ -44,7 +44,7 @@ public static void Main(string[] args)
                     {
                         listenOptions.Protocols = HttpProtocols.Http1AndHttp2;
                         listenOptions.UseHttps("testCert.pfx", "testPassword");
-                        listenOptions.UseConnectionLogging();
+                        // listenOptions.UseConnectionLogging();
```
JunTaoLuo: revert

src/Kestrel.Core/Internal/Http2/Http2Connection.cs
```diff
@@ -608,6 +608,8 @@ private Task ProcessSettingsFrameAsync()
 
                 _clientSettings.ParseFrame(_incomingFrame);
 
+                _frameWriter.WriteSettingsAckAsync(); // Ack before we update the windows, they could send data immediately.
```
JunTaoLuo: do we want this as fire and forget? Why not await it?

halter73: I'm on the fence on this. On one hand, there are several places already that we write small frames without awaiting them, and I think that's fine. On the other hand, other methods called by ProcessFrameAsync do return framewriter task, so it's more consistent to continue to do so.

I don't like introducing a new async state machine, but maybe that won't be to big of an issue in practice if settings frames are uncommon enough. Another alternative is just storing the Task returned by WriteSettingsAckAsync in a local and returning it later.

samples/Http2SampleApp/Program.cs
halter73: or remove

src/Kestrel.Core/Internal/Http2/Http2Connection.cs
Tratcher: Storing it and returning it later seems like a good compromise.

Issue #2793, opened 2018-08-08
Wait for each stream to complete before completing HTTP/2 connection even if the client already disconnected


halter73:
Even if a client already disconnected prior to shutdown, apps should have up to the graceful shutdown timeout to complete their business logic which might involve more than just writing a response.

I think we fix this by delaying the completion of Http2Connection.ProcessRequestsAsync() until all child Http2Stream.ProcessRequestsAsync() call complete even if the input pipe is completed.

Another reason for making this change is so we can enable memory pool validation in more of our tests. Here's what I was looking at.

```diff
diff --git a/test/Kestrel.Core.Tests/Http2ConnectionTests.cs b/test/Kestrel.Core.Tests/Http2ConnectionTests.cs
index c20c9ecf..dd1474e5 100644
--- a/test/Kestrel.Core.Tests/Http2ConnectionTests.cs
+++ b/test/Kestrel.Core.Tests/Http2ConnectionTests.cs
@@ -1065,12 +1065,6 @@ public async Task DATA_Received_StreamClosedImplicitly_ConnectionError()
         [Fact]
         public async Task DATA_Received_NoStreamWindowSpace_ConnectionError()
         {
-            // I hate doing this, but it avoids exceptions from MemoryPool.Dipose() in debug mode. The problem is since
-            // the stream's ProcessRequestsAsync loop is never awaited by the connection, it's not really possible to
-            // observe when all the blocks are returned. This can be removed after we implement graceful shutdown.
-            Dispose();
-            InitializeConnectionFields(new DiagnosticMemoryPool(KestrelMemoryPool.CreateSlabMemoryPool(), allowLateReturn: true));
-
             // _maxData should be 1/4th of the default initial window size + 1.
             Assert.Equal(Http2PeerSettings.DefaultInitialWindowSize + 1, (uint)_maxData.Length * 4);

@@ -1093,12 +1087,6 @@ public async Task DATA_Received_NoStreamWindowSpace_ConnectionError()
         [Fact]
         public async Task DATA_Received_NoConnectionWindowSpace_ConnectionError()
         {
-            // I hate doing this, but it avoids exceptions from MemoryPool.Dipose() in debug mode. The problem is since
-            // the stream's ProcessRequestsAsync loop is never awaited by the connection, it's not really possible to
-            // observe when all the blocks are returned. This can be removed after we implement graceful shutdown.
-            Dispose();
[...]

PR #2797, opened 2018-08-09
Gracefully wait for streams to complete when pipe completes

JunTaoLuo:
Addresses https://github.com/aspnet/KestrelHttpServer/issues/2793.

### DIFF
--- a/src/Kestrel.Core/Internal/Http2/Http2Connection.cs
+++ b/src/Kestrel.Core/Internal/Http2/Http2Connection.cs
@@ -79,6 +79,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http2
         private readonly object _stateLock = new object();
         private int _highestOpenedStreamId;
         private Http2ConnectionState _state = Http2ConnectionState.Open;
+        private readonly TaskCompletionSource<object> _streamsCompleted = new TaskCompletionSource<object>(TaskCreationOptions.RunContinuationsAsynchronously);
 
         private readonly ConcurrentDictionary<int, Http2Stream> _streams = new ConcurrentDictionary<int, Http2Stream>();
 
@@ -256,6 +257,11 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http2
                             _frameWriter.WriteGoAwayAsync(_highestOpenedStreamId, errorCode);
                             UpdateState(Http2ConnectionState.Closed);
                         }
+
+                        if (_streams.IsEmpty)
+                        {
+                            _streamsCompleted.TrySetResult(null);
+                        }
                     }
 
                     // Ensure aborting each stream doesn't result in unnecessary WINDOW_UPDATE frames being sent.
@@ -265,6 +271,8 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http2
                     {
                         stream.Abort(connectionError);
                     }
+
+                    await _streamsCompleted.Task;
 
                     _frameWriter.Complete();
                 }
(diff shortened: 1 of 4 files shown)


test/Kestrel.Core.Tests/Http2ConnectionTests.cs
```diff
@@ -1084,7 +1078,7 @@ public async Task DATA_Received_NoStreamWindowSpace_ConnectionError()
             await SendDataAsync(1, _maxData, endStream: false);
 
             await WaitForConnectionErrorAsync<Http2ConnectionErrorException>(
-                ignoreNonGoAwayFrames: false,
+                ignoreNonGoAwayFrames: true,
```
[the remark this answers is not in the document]
halter73: No headers frame would be written if you switched this test back to using `_waitForAbortApplication`, right?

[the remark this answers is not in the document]
JunTaoLuo: Yup, my bad.

src/Kestrel.Core/Internal/Http2/Http2Connection.cs
```diff
@@ -891,13 +899,19 @@ void IHttp2StreamLifetimeHandler.OnStreamCompleted(int streamId)
...
+                        UpdateState(Http2ConnectionState.Closed);
+
+                        // Wake up request processing loop so the connection can complete if there are no pending requests
+                        Input.CancelPendingRead();
+                    }
 
-                    // Wake up request processing loop so the connection can complete if there are no pending requests
-                    Input.CancelPendingRead();
+                    // Complete the task waiting on all streams to finish
+                    _streamsCompleted.TrySetResult(null);
```
Tratcher: What if the state is still Open?

src/Kestrel.Core/Internal/Http2/Http2Connection.cs
```diff
@@ -266,6 +272,8 @@ public async Task ProcessRequestsAsync<TContext>(IHttpApplication<TContext> appl
                         stream.Abort(connectionError);
                     }
 
+                    await _streamsCompleted.Task;
```
Tratcher: Timeout/cancellationToken?

test/Kestrel.Core.Tests/Http2ConnectionTests.cs
```diff
@@ -1065,12 +1065,6 @@ public async Task DATA_Received_StreamClosedImplicitly_ConnectionError()
         [Fact]
         public async Task DATA_Received_NoStreamWindowSpace_ConnectionError()
         {
-            // I hate doing this, but it avoids exceptions from MemoryPool.Dipose() in debug mode. The problem is since
```
Tratcher: There are more of these in Http2StreamTests https://github.com/aspnet/KestrelHttpServer/blob/bd3a8a3511238a87c61d9ed798db373c1b88dd78/test/Kestrel.Core.Tests/Http2StreamTests.cs#L967

src/Kestrel.Core/Internal/Http2/Http2Connection.cs
JunTaoLuo: true

src/Kestrel.Core/Internal/Http2/Http2Connection.cs
JunTaoLuo: Is the abortive shutdown path via Abort(...) not sufficient?

Tratcher: That's a good start, but right now you have no guarantee that the Task here will ever complete.

halter73: I don't think we need to have a timeout or a CancellationToken here. What we have now is in line with the behavior for HTTP/1. If a connection's ProcessRequestsAsync method doesn't complete before the shutdown timeout CancellationToken fires, then CloseAllConnectionsAsync will complete and AbortAllConnectionsAsync will be called.

It's think it's a good idea to complete this TCS in Http2Connection.Abort() though.

JunTaoLuo: Hmm looks like the only impact if the task doesn't complete is if there are outstanding streams. After the graceful timeout, the abortive path takes over and still completes/aborts the _frameWriter. The only difference is that Input.Complete() is not called. Is that the concern?

halter73: Now that I think about it more, I think it's a bad idea to complete this TCS in Abort. The socket get's closed in abort (if it isn't already), and that itself might unblock the application. AbortAllConnectionsAsync will timeout after a second if that's not the case anyway.

PR #2801, opened 2018-08-09
Fix bug in graceful shutdown

JunTaoLuo:
Oops, still had bugs in the last PR. Follow up to #2797

### DIFF
--- a/src/Kestrel.Core/Internal/Http2/Http2Connection.cs
+++ b/src/Kestrel.Core/Internal/Http2/Http2Connection.cs
@@ -910,8 +910,12 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http2
                         Input.CancelPendingRead();
                     }
 
-                    // Complete the task waiting on all streams to finish
-                    _streamsCompleted.TrySetResult(null);
+
+                    if (_state != Http2ConnectionState.Open)
+                    {
+                        // Complete the task waiting on all streams to finish
+                        _streamsCompleted.TrySetResult(null);
+                    }
                 }
             }
         }
--- a/test/Kestrel.Core.Tests/Http2ConnectionTests.cs
+++ b/test/Kestrel.Core.Tests/Http2ConnectionTests.cs
@@ -3325,6 +3325,54 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Tests
         }
 
         [Fact]
+        public async Task StopProcessingNextRequestSendsGracefulGOAWAYAndWaitsForStreamsToComplete()
+        {
+            var task = Task.CompletedTask;
+            await InitializeConnectionAsync(context => task);
+
+            // Send and receive an unblocked request
+            await StartStreamAsync(1, _browserRequestHeaders, endStream: true);
+
+            await ExpectAsync(Http2FrameType.HEADERS,
+                withLength: 55,
+                withFlags: (byte)Http2HeadersFrameFlags.END_HEADERS,
+                withStreamId: 1);
+            await ExpectAsync(Http2FrameType.DATA,
+                withLength: 0,
+                withFlags: (byte)Http2DataFrameFlags.END_STREAM,
+                withStreamId: 1);
+
+            // Send a blocked request
+            var tcs = new TaskCompletionSource<object>(TaskContinuationOptions.RunContinuationsAsynchronously);
+            task = tcs.Task;
+            await StartStreamAsync(3, _browserRequestHeaders, endStream: false);
+
+            // Close pipe
+            _pair.Application.Output.Complete();
+
+            // Assert connection closed
+            await _closedStateReached.Task.DefaultTimeout();
+            VerifyGoAway(await ReceiveFrameAsync(), 3, Http2ErrorCode.NO_ERROR);
+
+            // Assert connection shutdown is still blocked
+            // ProcessRequestsAsync completes the connection's Input pipe
+            var readTask = _pair.Application.Input.ReadAsync();
+            _pair.Application.Input.CancelPendingRead();
+            var result = await readTask;
+            Assert.False(result.IsCompleted);
+
+            // Unblock the request and ProcessRequestsAsync
+            tcs.TrySetResult(null);
+            await _connectionTask;
+
+            // Assert connection's Input pipe is completed
+            readTask = _pair.Application.Input.ReadAsync();
+            _pair.Application.Input.CancelPendingRead();
+            result = await readTask;
+            Assert.True(result.IsCompleted);
+        }
+
+        [Fact]
         public async Task StopProcessingNextRequestSendsGracefulGOAWAYThenFinalGOAWAYWhenAllStreamsComplete()
         {
             await InitializeConnectionAsync(_echoApplication);
--- a/test/Kestrel.Core.Tests/Http2StreamTests.cs
+++ b/test/Kestrel.Core.Tests/Http2StreamTests.cs
@@ -964,12 +964,6 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Tests
         [Fact]
         public async Task ContentLength_Received_SingleDataFrameUnderSize_Reset()
         {
-            // I hate doing this, but it avoids exceptions from MemoryPool.Dipose() in debug mode. The problem is since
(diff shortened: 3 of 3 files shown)


test/Kestrel.Core.Tests/Http2StreamTests.cs
```diff
@@ -964,12 +964,6 @@ public async Task ContentLength_Received_SingleDataFrameOverSize_Reset()
         [Fact]
         public async Task ContentLength_Received_SingleDataFrameUnderSize_Reset()
         {
-            // I hate doing this, but it avoids exceptions from MemoryPool.Dipose() in debug mode. The problem is since
-            // the stream's ProcessRequestsAsync loop is never awaited by the connection, it's not really possible to
-            // observe when all the blocks are returned. This can be removed after we implement graceful shutdown.
-            Dispose();
-            InitializeConnectionFields(new DiagnosticMemoryPool(KestrelMemoryPool.CreateSlabMemoryPool(), allowLateReturn: true));
```
halter73: This is no longer called in any of our unit tests, right?

src/Kestrel.Core/Internal/Http2/Http2Connection.cs
```diff
@@ -910,8 +910,12 @@ void IHttp2StreamLifetimeHandler.OnStreamCompleted(int streamId)
                         Input.CancelPendingRead();
                     }
 
-                    // Complete the task waiting on all streams to finish
-                    _streamsCompleted.TrySetResult(null);
+
+                    if (_state != Http2ConnectionState.Open)
```
halter73: Nice catch @Tratcher. Let's add a new regression unit test that:

1. Opens a stream where the application doesn't block.
2. Closes the stream.
3. Opens a new stream where the application blocks.
4. Closes the input pipe
5. Verifies that Http2Connection.ProcessRequestsAsync doesn't complete until the application for the second stream completes.

test/Kestrel.Core.Tests/Http2StreamTests.cs
JunTaoLuo: There are a few spots:
https://github.com/aspnet/KestrelHttpServer/blob/release/2.2/test/Kestrel.Core.Tests/DiagnosticMemoryPoolTests.cs#L176-L200. This is specifically testing those behaviours so we'll keep these.
https://github.com/aspnet/KestrelHttpServer/blob/release/2.2/test/Kestrel.Transport.FunctionalTests/Http2/ShutdownTests.cs#L100. This is the abortive test and I think it's unavoidable.
https://github.com/aspnet/KestrelHttpServer/blob/0e99235d59031106bfe6c593757b99ff99f261f7/test/Kestrel.Transport.FunctionalTests/MaxRequestBufferSizeTests.cs#L117. I'll see if these can be removed though.

halter73: Don't worry about the functional tests. I was only asking about unit tests.

The MaxRequestBufferSizeTests have been that way in a while. Even if it passes without allowing late returns on your machine, I would be very concerned about introducing test flakiness. You're also right about it being unavoidable for the abortive Shutdown tests.

JunTaoLuo:
Added a test

test/Kestrel.Core.Tests/Http2ConnectionTests.cs
```diff
@@ -3324,6 +3324,54 @@ public async Task CompletionSendsFinalGOAWAY()
             VerifyGoAway(await ReceiveFrameAsync(), 0, Http2ErrorCode.NO_ERROR);
         }
 
+        [Fact]
+        public async Task StopProcessingNextRequestSendsGracefulGOAWAYAndWaitsForStreamsToComplete()
+        {
+            var requestSemaphore = new SemaphoreSlim(1);
```
halter73: I think this could just be a TCS. No need to block a thread.

test/Kestrel.Core.Tests/Http2ConnectionTests.cs
```diff
@@ -3324,6 +3324,54 @@ public async Task CompletionSendsFinalGOAWAY()
...
+            _pair.Application.Input.CancelPendingRead();
+            var result = await readTask;
+            Assert.False(result.IsCompleted);
+
+            // Unblock the request and ProcessRequestsAsync
+            requestSemaphore.Release();
+            await _connectionTask;
+
+            // Assert connection's Input pipe is completed
+            result = await _pair.Application.Input.ReadAsync();
```
halter73: Nit: .DefaultTimeout()

halter73: Also, for consistency, use the same pattern you used above of calling ReadAsync, then CancelPendingRead, then awaiting the task. It should return a completed task before CancelPendingRead is even called if you use a TCS without the RunContinuationsAsync option. This is the one of the few times I would recommend not using that option as long as you add a comment explaining why.

JunTaoLuo: You can't use .DefaultTimeout() here since ReadAsync() returns a ValueTask

halter73: You could always call AsTask() first, but there shouldn't be much risk of a hanging test even if this regresses as long as the extra CancelPendingRead() is added anyway.

test/Kestrel.Core.Tests/Http2ConnectionTests.cs
```diff
@@ -3327,11 +3327,11 @@ public async Task CompletionSendsFinalGOAWAY()
...
         {
-            var requestSemaphore = new SemaphoreSlim(1);
-            await InitializeConnectionAsync(context =>
+            var task = Task.CompletedTask;
+            await InitializeConnectionAsync(async context =>
             {
-                requestSemaphore.Wait();
-                return Task.CompletedTask;
+                await task;
+                context.Response.StatusCode = 200;
```
halter73: Why is this here?

test/Kestrel.Core.Tests/Http2ConnectionTests.cs
```diff
@@ -3347,6 +3347,8 @@ public async Task StopProcessingNextRequestSendsGracefulGOAWAYAndWaitsForStreams
                 withStreamId: 1);
 
             // Send a blocked request
+            var tcs = new TaskCompletionSource<object>(); // Not running continuations as async since the task may be completed
```
halter73: Nit: Comment on own line. I would phrase it like so:

```
// No RunContinuationsAsynchronously so we can verify assertions immediately after calling SetResult.
```

test/Kestrel.Core.Tests/Http2ConnectionTests.cs
JunTaoLuo: oops test code I accidentally staged

test/Kestrel.Core.Tests/Http2ConnectionTests.cs
```diff
@@ -3324,6 +3324,54 @@ public async Task CompletionSendsFinalGOAWAY()
...
+            // Assert connection shutdown is still blocked
+            // ProcessRequestsAsync completes the connection's Input pipe
+            var readTask = _pair.Application.Input.ReadAsync();
+            _pair.Application.Input.CancelPendingRead();
+            var result = await readTask;
+            Assert.False(result.IsCompleted);
+
+            // Unblock the request and ProcessRequestsAsync
+            requestSemaphore.Release();
+            await _connectionTask;
```
halter73: Before you merge, add .DefaultTimeout() here though.

test/Kestrel.Core.Tests/Http2ConnectionTests.cs
```diff
@@ -3324,6 +3324,55 @@ public async Task CompletionSendsFinalGOAWAY()
...
+            // Assert connection shutdown is still blocked
+            // ProcessRequestsAsync completes the connection's Input pipe
+            var readTask = _pair.Application.Input.ReadAsync();
+            _pair.Application.Input.CancelPendingRead();
+            var result = await readTask;
+            Assert.False(result.IsCompleted);
+
+            // Unblock the request and ProcessRequestsAsync
+            tcs.TrySetResult(null);
+            await _connectionTask;
```
halter73: Before you merge, add .DefaultTimeout() here though.

PR #2819, opened 2018-08-11
Improve HTTP/2 stream abort logic

halter73:
- Fix race where headers frame could be written after an abort was observed
- Fix Http2StreamTests to verify expected abort-related exceptions

#2799

### DIFF
--- a/src/Kestrel.Core/Internal/Http2/Http2Stream.cs
+++ b/src/Kestrel.Core/Internal/Http2/Http2Stream.cs
@@ -352,7 +352,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http2
             return _context.FrameWriter.TryUpdateStreamWindow(_outputFlowControl, bytes);
         }
 
-        public override void Abort(ConnectionAbortedException abortReason)
+        public void Abort(IOException abortReason)
         {
             if (!TryApplyCompletionFlag(StreamCompletionFlags.Aborted))
             {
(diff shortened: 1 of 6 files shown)


src/Kestrel.Core/Internal/Http2/Http2Stream.cs
```diff
@@ -429,7 +431,7 @@ private bool TryApplyCompletionFlag(StreamCompletionFlags completionState)
             }
         }
 
-        private static bool ShoulStopTrackingStream(StreamCompletionFlags completionState)
+        private static bool ShouldStopTrackingStream(StreamCompletionFlags completionState)
```
halter73: Happy now @benaadams? Seriously though, thanks for spotting this.

Tratcher:
Hmm, interesting test failures.
Windows - Memory Pool errors
Linux - test hang
Mac - missing response

src/Kestrel.Core/Internal/Http2/Http2Stream.cs
```diff
@@ -391,12 +390,15 @@ private void ResetAndAbort(ConnectionAbortedException abortReason, Http2ErrorCod
...
+        private void AbortCore(Exception abortReason)
         {
-            base.Abort(abortReason);
+            // Call OnIOCompleted() which closes the output prior to poisoning the request body stream or pipe to
+            // ensure that an app that completes early due to the abort doesn't result in header frames being sent.
+            OnInputOrOutputCompleted();
 
             // Unblock the request body.
-            RequestBodyPipe.Writer.Complete(new IOException(CoreStrings.Http2StreamAborted, abortReason));
+            PoisonRequestBodyStream(abortReason);
```
JunTaoLuo: Is there a risk to calling this multiple times?

halter73: It could wind up changing the exception type thrown from the request body stream. AbortCore should only ever be called once per stream though.

Issue #2810, opened 2018-08-10
Implement MaxRequestBodySize for HTTP/2


Tratcher:
Meta issue: https://github.com/aspnet/KestrelHttpServer/issues/2805

Same semantics as HTTP/1.

PR #2822, opened 2018-08-13
Implement MaxRequestBodySize for HTTP/2 

Tratcher:
#2810 Reusing logic from HTTP/1.


test/Kestrel.Core.Tests/Http2StreamTests.cs
```diff
@@ -1725,7 +2007,7 @@ private async Task<Http2Frame> ReceiveFrameAsync()
 
         private async Task<Http2Frame> ExpectAsync(Http2FrameType type, int withLength, byte withFlags, int withStreamId)
         {
-            var frame = await ReceiveFrameAsync();
+            var frame = await ReceiveFrameAsync().DefaultTimeout();
```
halter73: Good call :+1:

Issue #2815, opened 2018-08-10
Limit concurrent HTTP/2 Streams per connection


Tratcher:
Meta issue: #2627

max streams per connection maps to SETTINGS_MAX_CONCURRENT_STREAMS.

The spec default is unlimited.

Tratcher:
Spec suggested minimum: 100
Nginx default: 128 http://nginx.org/en/docs/http/ngx_http_v2_module.html#http2_max_concurrent_streams
Apache default : 100 https://httpd.apache.org/docs/2.4/mod/mod_http2.html
IIS/Http.Sys default?

Issue #2773, opened 2018-08-03
Extract bitshift helper code


Tratcher:
RE: https://github.com/aspnet/KestrelHttpServer/pull/2771/files

The Http2Frame class has a lot of bitshift operations scattered throughout like `get => (Payload[0] << 24) | (Payload[1] << 16) | (Payload[2] << 8) | Payload[3];` We've already found several typos in these that caused to some very strange data corruption bugs.

Proposal: Extract these operations to specific helper APIs to allow testing them directly, reduce duplication, and make the Frame code easier to read. Session has a few examples: https://github.com/aspnet/Session/blob/92ee0bfe89c5136671d82a783799e1d5f7441b53/src/Microsoft.AspNetCore.Session/DistributedSession.cs#L360-L406

davidfowl:
OR use the new BinaryPrimitives helpers in System.Memory

Tratcher:
[Oooooooo](https://www.fuget.org/packages/System.Memory/4.5.1/netstandard2.0/lib/System.Memory.dll/System.Buffers.Binary/BinaryPrimitives#BinaryPrimitives)

PR #2825, opened 2018-08-14
Limit concurrent HTTP/2 Streams per connection

Tratcher:
#2815 Defaults to 100 which is similar to Nginx, Apache, and IIS/Http.Sys.

I needed to finish implementing the infrastructure for sending settings to the client.

I also experimented with the BinaryConverter APIs for #2773, they worked great.

### DIFF
--- a/src/Kestrel.Core/Internal/Http2/Http2Connection.cs
+++ b/src/Kestrel.Core/Internal/Http2/Http2Connection.cs
@@ -88,6 +88,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http2
             _context = context;
             _frameWriter = new Http2FrameWriter(context.Transport.Output, context.Application.Input, _outputFlowControl, this, context.ConnectionId, context.ServiceContext.Log);
             _hpackDecoder = new HPackDecoder((int)_serverSettings.HeaderTableSize);
+            _serverSettings.MaxConcurrentStreams = (uint)context.ServiceContext.ServerOptions.Limits.Http2.MaxStreamsPerConnection;
         }
 
         public string ConnectionId => _context.ConnectionId;
@@ -97,6 +98,8 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http2
         public IKestrelTrace Log => _context.ServiceContext.Log;
 
         public IFeatureCollection ConnectionFeatures => _context.ConnectionFeatures;
+
+        internal Http2PeerSettings ServerSettings => _serverSettings;
 
         public void OnInputOrOutputCompleted()
         {
@@ -172,7 +175,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http2
 
                 if (_state != Http2ConnectionState.Closed)
                 {
-                    await _frameWriter.WriteSettingsAsync(_serverSettings);
+                    await _frameWriter.WriteSettingsAsync(_serverSettings.GetNonProtocolDefaults());
                 }
 
                 while (_state != Http2ConnectionState.Closed)
@@ -617,10 +620,10 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http2
 
             try
             {
-                // ParseFrame will not parse an InitialWindowSize > int.MaxValue.
+                // int.MaxValue is the largest allowed windows size.
                 var previousInitialWindowSize = (int)_clientSettings.InitialWindowSize;
 
-                _clientSettings.ParseFrame(_incomingFrame);
+                _clientSettings.Update(_incomingFrame.GetSettings());
 
                 var ackTask = _frameWriter.WriteSettingsAckAsync(); // Ack before we update the windows, they could send data immediately.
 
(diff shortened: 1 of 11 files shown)


src/Kestrel.Core/Internal/Http2/Http2Frame.Settings.cs
```diff
@@ -1,42 +1,67 @@
...
+            var id = (Http2SettingsParameter)BinaryPrimitives.ReadUInt16BigEndian(payload);
+            var value = BinaryPrimitives.ReadUInt32BigEndian(payload.Slice(2));
+
+            return new Http2PeerSetting(id, value);
+        }
 
-            Length = 6 * settingCount;
+        public void PrepareSettings(Http2SettingsFrameFlags flags, IList<Http2PeerSetting> settings = null)
+        {
+            var settingCount = settings?.Count() ?? 0;
```
campersau: Can be `settings?.Count ?? 0` now.

Tratcher: Yes, for the Settings Ack code path no settings are sent.

src/Kestrel.Core/KestrelServerLimits.cs
```diff
@@ -251,6 +252,25 @@ public TimeSpan RequestHeadersTimeout
             }
         }
 
+        /// <summary>
+        /// Limits the number of concurrent request streams per HTTP/2 connection. Excess streams will be refused.
+        /// <para>
+        /// Defaults to 100
+        /// </para>
+        /// </summary>
+        public int MaxStreamsPerConnection
```
halter73: What do you think about making an Http2Limits class/property and adding this to that?

src/Kestrel.Core/KestrelServerLimits.cs
```diff
@@ -251,6 +252,25 @@ public TimeSpan RequestHeadersTimeout
...
+        {
+            get => _maxStreamsPerConnection;
+            set
+            {
+                if (value < 0)
```
halter73: `<=`

Tratcher: 0 is technically allowed. You can have the endpoint up for pings but not accept requests.

src/Kestrel.Core/KestrelServerLimits.cs
Tratcher: No, there won't be that many. Also most of the existing limits apply to Http/2.

src/Kestrel.Core/Internal/Http2/Http2Connection.cs
```diff
@@ -838,6 +841,11 @@ private void StartStream<TContext>(IHttpApplication<TContext> application)
                 throw new Http2StreamErrorException(_currentHeadersStream.StreamId, CoreStrings.Http2ErrorMissingMandatoryPseudoHeaderFields, Http2ErrorCode.PROTOCOL_ERROR);
             }
 
+            if (_streams.Count >= _serverSettings.MaxConcurrentStreams)
```
JunTaoLuo: Is there a potential race here? Between checking the _streams.Count and adding the _currentHeadersStream to _streams?

Tratcher: We're under the state lock, so there shouldn't be.

Tratcher: We also only ever process one frame at a time.

src/Kestrel.Core/Internal/Http2/Http2PeerSettings.cs
```diff
@@ -91,16 +85,42 @@ public void ParseFrame(Http2Frame frame)
...
+        internal IList<Http2PeerSetting> GetNonProtocolDefaults()
         {
-            yield return new Http2PeerSetting(Http2SettingsParameter.SETTINGS_HEADER_TABLE_SIZE, HeaderTableSize);
-            yield return new Http2PeerSetting(Http2SettingsParameter.SETTINGS_ENABLE_PUSH, EnablePush ? 1u : 0);
-            yield return new Http2PeerSetting(Http2SettingsParameter.SETTINGS_MAX_CONCURRENT_STREAMS, MaxConcurrentStreams);
-            yield return new Http2PeerSetting(Http2SettingsParameter.SETTINGS_INITIAL_WINDOW_SIZE, InitialWindowSize);
-            yield return new Http2PeerSetting(Http2SettingsParameter.SETTINGS_MAX_FRAME_SIZE, MaxFrameSize);
-            yield return new Http2PeerSetting(Http2SettingsParameter.SETTINGS_MAX_HEADER_LIST_SIZE, MaxHeaderListSize);
-        }
+            var list = new List<Http2PeerSetting>(0);
```
halter73: Given that `MaxConcurrentStreams != DefaultMaxConcurrentStreams` by default (ironic no?), shouldn't we at least pass 1 into the List ctor? Also, what's wrong with IEnumerable and yield return?

src/Kestrel.Core/Internal/Http2/Http2Connection.cs
JunTaoLuo: Oh yea, I added that state lock. I just forgot to check the callsite.

src/Kestrel.Core/Internal/Http2/Http2Connection.cs
```diff
@@ -620,7 +623,7 @@ private Task ProcessSettingsFrameAsync()
                 // ParseFrame will not parse an InitialWindowSize > int.MaxValue.
```
halter73: Nit: fix comment.

src/Kestrel.Core/KestrelServerLimits.cs
JunTaoLuo: But for Http/2 specific limits, maybe we should prefix them to something like `HTTP2MaxStreamsPerConnection`?

halter73: What all do we expect? I can already think of these potentially:

- MaxStreamsPerConnection
- TotalMaxStreams
- ConnectionWindowSize
- StreamWindowSize
- WindowUpdateInterval
- MaxFrameSize

Even if we only do a couple of these, I like having non-HTTP/1.x-related limits separated. In the future, we could add even more related to header table size and possibly other things we haven't even considered yet.

src/Kestrel.Core/KestrelServerLimits.cs
halter73: Why though? Even if if allowed by protocol, we don't need to allow a pit of failure.

src/Kestrel.Core/KestrelServerLimits.cs
Tratcher: I asked @shirhatti and he wants to keep it as a flat list.

Tratcher:
Minor updates

PR #2828, opened 2018-08-16
General connection management

davidfowl:
This is just a proof of concept PR that tries to make the handling of graceful shutdown work for more than just http scenarios. This should allow us to move TLS further out and should also allow us to start moving things to connection middleware instead of connection adapters.

Summary of the things added:
- Added IConnectionHeartbeatTickFeature that represents a per connection heartbeat with DateTimeOffset.UtcNow passed into the callback. This is trying to preserve the performance characteristics Kestrel had before on the HTTP path. Haven't measured anything here as yet. 
- Added IGracefulConnectionLifetimeFeature that represents an attempt to gracefully close the connection that isn't being aborted. This feels pretty awful but we may have to do it.
- Moved connection management to the ConnectionDispatcher and out of the HttpConnectionMiddleware

PS: I haven't even run all tests locally so I expect explosions all over.


davidfowl:
Opening a new PR

PR #2829, opened 2018-08-16
Extract bitshift helper code, Frame cleanup

Tratcher:
#2773 

Extracted all non-trivial bitwise operations from the Http2Frame classes. I also fixed several issues where it was reading 31 bits but writing 32 bits and stomping on adjacent (rarely used) fields.

There was an error in Data and Headers where reading the padding length or priority weight fields could read off the end of an invalid frame. I normalized everything to read from the restricted Payload Span so everything stays in bounds.

BinaryPrimitives are modestly faster and more consistent than the old code. Not to mention a lot cleaner.
```
             Method |     Mean |     Error |   StdDev |     Op/s | Scaled | ScaledSD | Allocated |
------------------- |---------:|----------:|---------:|---------:|-------:|---------:|----------:|
 GetUInt32AsBitwise | 21.80 us | 1.1883 us | 3.466 us | 45,861.5 |   1.00 |     0.00 |       0 B |
  GetUInt32AsBinary | 19.98 us | 0.6416 us | 1.656 us | 50,061.2 |   0.94 |     0.17 |       0 B |
```

### DIFF
--- a/src/Kestrel.Core/Internal/Http2/Http2Frame.Data.cs
+++ b/src/Kestrel.Core/Internal/Http2/Http2Frame.Data.cs
@@ -2,10 +2,18 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
-using Microsoft.Extensions.Logging;
 
 namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http2
 {
+    /*
+        +---------------+
+        |Pad Length? (8)|
+        +---------------+-----------------------------------------------+
+        |                            Data (*)                         ...
+        +---------------------------------------------------------------+
+        |                           Padding (*)                       ...
+        +---------------------------------------------------------------+
+    */
     public partial class Http2Frame
     {
         public Http2DataFrameFlags DataFlags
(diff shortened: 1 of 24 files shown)


src/Kestrel.Core/Internal/Http2/Http2Frame.Data.cs
```diff
@@ -14,64 +22,35 @@ public Http2DataFrameFlags DataFlags
...
         }
 
-        public ArraySegment<byte> DataPayload => DataHasPadding
-            ? new ArraySegment<byte>(_data, PayloadOffset + 1, Length - DataPadLength - 1)
-            : new ArraySegment<byte>(_data, PayloadOffset, Length);
+        public int DataPayloadOffset => DataHasPadding ? 1 : 0;
+
+        private int DataPayloadLength => PayloadLength - DataPayloadOffset - DataPadLength;
+
+        public Span<byte> DataPayload => Payload.Slice(DataPayloadOffset, HeadersPayloadLength);
```
JunTaoLuo: Why HeadersPayloadLength? Do you mean DataPayloadLength?

src/Kestrel.Core/Internal/Http2/Http2Frame.GoAway.cs
```diff
@@ -1,37 +1,38 @@
...
         {
-            get => (Payload[0] << 24) | (Payload[1] << 16) | (Payload[2] << 8) | Payload[3];
-            set
-            {
-                Payload[0] = (byte)((value >> 24) & 0xff);
-                Payload[1] = (byte)((value >> 16) & 0xff);
-                Payload[2] = (byte)((value >> 8) & 0xff);
-                Payload[3] = (byte)(value & 0xff);
-            }
+            get => (int)Bitshifter.ReadUInt31BigEndian(Payload);
```
JunTaoLuo: makes me wonder if we should just make the GoAwayLastStreamId uint. Stream ids are positive 31 bit integers according to spec.

Tratcher: int and uint are interchangeable for 31bit. We consistently use int for stream ids throughout the code base, and there's even a feature interface for it. That's a bigger change I don't want to make here, and it's not clear that we'd gain anything.

Tratcher:
Updated. Test failures are covered by #2831

PR #2830, opened 2018-08-17
Fixed a memory leak and flaky test

davidfowl:
- The memory leak was caused by not properly cleaning up the RequestBodyPipe for Http2Streams. Removing the _ioCompleted check before consuming the message body seems to work well.
- The flaky test didn't handle the fact that writing to the connection after a stream has been RSTed that it'll cause the entire connection to close.

### DIFF
--- a/src/Kestrel.Core/Internal/Http/HttpProtocol.cs
+++ b/src/Kestrel.Core/Internal/Http/HttpProtocol.cs
@@ -607,7 +607,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
                 application.DisposeContext(httpContext, _applicationException);
 
                 // Even for non-keep-alive requests, try to consume the entire body to avoid RSTs.
-                if (_ioCompleted == 0 && _requestRejectedException == null && !messageBody.IsEmpty)
+                if (_requestRejectedException == null && !messageBody.IsEmpty)
                 {
                     await messageBody.ConsumeAsync();
                 }
--- a/test/Kestrel.InMemory.FunctionalTests/Http2/Http2StreamTests.cs
+++ b/test/Kestrel.InMemory.FunctionalTests/Http2/Http2StreamTests.cs
@@ -836,18 +836,19 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Tests
             await StartStreamAsync(1, headers, endStream: false);
             await SendDataAsync(1, new byte[1].AsSpan(), endStream: false);
             await SendDataAsync(1, new byte[2].AsSpan(), endStream: false);
+            // This will cause the content length to go bad
             await SendDataAsync(1, new byte[10].AsSpan(), endStream: false);
-            await SendDataAsync(1, new byte[2].AsSpan(), endStream: true);
 
             await WaitForStreamErrorAsync(1, Http2ErrorCode.PROTOCOL_ERROR, CoreStrings.Http2StreamErrorMoreDataThanLength);
 
-            await StopConnectionAsync(expectedLastStreamId: 1, ignoreNonGoAwayFrames: false);
+            // This will cause trigger a go away frame to be sent
+            await SendDataAsync(1, new byte[2].AsSpan(), endStream: true);
 
-            var expectedError = new Http2StreamErrorException(1, CoreStrings.Http2StreamErrorMoreDataThanLength, Http2ErrorCode.PROTOCOL_ERROR);
-
-            Assert.NotNull(thrownEx);
-            Assert.Equal(expectedError.Message, thrownEx.Message);
-            Assert.IsType<Http2StreamErrorException>(thrownEx.InnerException);
+            await WaitForConnectionErrorAsync<Http2ConnectionErrorException>(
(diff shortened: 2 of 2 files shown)


src/Kestrel.Core/Internal/Http/HttpProtocol.cs
```diff
@@ -607,7 +607,7 @@ private async Task ProcessRequests<TContext>(IHttpApplication<TContext> applicat
                 application.DisposeContext(httpContext, _applicationException);
 
                 // Even for non-keep-alive requests, try to consume the entire body to avoid RSTs.
-                if (_ioCompleted == 0 && _requestRejectedException == null && !messageBody.IsEmpty)
```
Tratcher: I think this is trying to avoid a secondary exception if the connection has already closed. Don't merge this until @halter73 can take a look.

test/Kestrel.InMemory.FunctionalTests/Http2/Http2StreamTests.cs
```diff
@@ -836,18 +836,19 @@ public async Task ContentLength_Received_MultipleDataFramesOverSize_Reset()
             await StartStreamAsync(1, headers, endStream: false);
             await SendDataAsync(1, new byte[1].AsSpan(), endStream: false);
             await SendDataAsync(1, new byte[2].AsSpan(), endStream: false);
+            // This will cause the content length to go bad
             await SendDataAsync(1, new byte[10].AsSpan(), endStream: false);
-            await SendDataAsync(1, new byte[2].AsSpan(), endStream: true);
```
Tratcher: It's a product bug if we can't handle this race, not a test bug. The frame is likely already in flight when we send the RST and we need to handle it.

test/Kestrel.InMemory.FunctionalTests/Http2/Http2StreamTests.cs
```diff
@@ -836,18 +836,19 @@ public async Task ContentLength_Received_MultipleDataFramesOverSize_Reset()
...
+            await SendDataAsync(1, new byte[2].AsSpan(), endStream: true);
 
-            Assert.NotNull(thrownEx);
-            Assert.Equal(expectedError.Message, thrownEx.Message);
-            Assert.IsType<Http2StreamErrorException>(thrownEx.InnerException);
+            await WaitForConnectionErrorAsync<Http2ConnectionErrorException>(
+                ignoreNonGoAwayFrames: false,
+                expectedLastStreamId: 1,
+                expectedErrorCode: Http2ErrorCode.STREAM_CLOSED,
+                expectedErrorMessage: null);
```
Tratcher: Http2StreamErrorMoreDataThanLength

Tratcher:
#2799

test/Kestrel.InMemory.FunctionalTests/Http2/Http2StreamTests.cs
davidfowl: Can you elaborate? It's not a race the test is just broken. If you send to an aborted stream the entire connection is aborted. Is that not the expected behavior?

src/Kestrel.Core/Internal/Http/HttpProtocol.cs
davidfowl: Yep, I'll need to put this back and probably just add something special for HTTP2 here. It seems like it may have broken something else.

test/Kestrel.InMemory.FunctionalTests/Http2/Http2StreamTests.cs
Tratcher: No that's not acceptable behavior for the server. The client can have multiple frames in flight before the stream is aborted. Those frames need to be filtered out rather than causing a connection error. The only fatal error is if the client sends more frames to a stream after they've sent a reset (we don't handle this either, there's a skipped h2spec test for it).

davidfowl: So there’s a difference between in flight and sent after the abort? We don’t know the second frame is ok flight and it seems a bit odd to special case that no? It would mean we need to drain the socket before aborting any streams?

Tratcher: Sorry, I confused things by introducing the reverse scenario. Let me get that out of the way.

Sender abort: It's a connection level protocol error for the sender to send any frames on a stream after they've sent (not received) a reset. This applies to both client and server. There's an h2spec test for this we don't pass at the moment.

But that's not what's happening in this test. In this test the client is sending data frames and the server is sending the reset. The server has no idea when the client will receive the reset and stop sending frames. The reset is a polite, cooperative dismissal. There's nothing in the protocol that says how long the server should filter out frames from the client after a reset. If the client does not stop eventually then the server has no choice but to close the connection. I expect we'll need to introduce a new timeout here that tracks and filters aborted streams for a given time period.

src/Kestrel.Core/Internal/Http/HttpProtocol.cs
davidfowl: There's another way to fix it.

PR #2831, opened 2018-08-17
Fix and skip flaky tests

davidfowl:
Skipped 2 known test flakyness issues (they are actual product bugs that @Tratcher is working on) and fix the memory leak error that shows up in tests.

Wait on all aborted streams to complete before returning from the main processing loop.

### DIFF
--- a/src/Kestrel.Core/Internal/Http2/Http2Connection.cs
+++ b/src/Kestrel.Core/Internal/Http2/Http2Connection.cs
@@ -4,6 +4,7 @@
 using System;
 using System.Buffers;
 using System.Collections.Concurrent;
+using System.Collections.Generic;
 using System.IO;
 using System.IO.Pipelines;
 using System.Security.Authentication;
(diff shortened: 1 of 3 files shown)


test/Kestrel.InMemory.FunctionalTests/Http2/Http2StreamTests.cs
```diff
@@ -850,7 +850,7 @@ public async Task ContentLength_Received_MultipleDataFramesOverSize_Reset()
             Assert.IsType<Http2StreamErrorException>(thrownEx.InnerException);
         }
 
-        [Fact]
+        [Fact(Skip="Flaky test #2799")]
```
Tratcher: Stephen already fixed this one. Only the OverSize test is flaky now right?

src/Kestrel.Core/Internal/Http2/Http2Connection.cs
```diff
@@ -270,13 +271,19 @@ public async Task ProcessRequestsAsync<TContext>(IHttpApplication<TContext> appl
...
                     {
-                        stream.Abort(new IOException(CoreStrings.Http2StreamAborted, connectionError));
+                        stream.Value.Abort(new IOException(CoreStrings.Http2StreamAborted, connectionError));
+                        tasks.Add(stream.Value.ProcessingTask);
                     }
 
                     await _streamsCompleted.Task;
 
+                    // REVIEW: Do we need a timeout?
+                    await Task.WhenAll(tasks);
```
Tratcher: Yes, timeout. The app code could block these from finishing.

Any better name than tasks?

Tratcher: This is re-implementing _streamsCompleted where the only difference is the ordering inside OnRequestProcessingEnded. Can you re-order that method to get the same effect? Or can you remove _streamsCompleted as it appears redundant?

davidfowl: Sounds good

PR #2833, opened 2018-08-18
Fire OnStreamCompleted after all pipes are closed

JunTaoLuo:
Which flaky tests were these supposed to fix @davidfowl ?


Tratcher:
ContentLength_Received_MultipleDataFramesUnderSize_Reset failed on my PR with the memory pool error. The trouble is that it affects random tests.

Tratcher:
Better.

davidfowl:
Nice!

PR #2834, opened 2018-08-19
General connection management

### LINKED ISSUE #2335: Fix connection management with connection middleware
Today Kestrel exposes the `IConnectionBuilder` interface but connection management is only implemented in the HttpConnectionMiddleware. To properly port existing connection adapters to connection middleware, we need to move the connection management up a layer.

davidfowl:
This PR makes the handling of graceful shutdown work for more than just http scenarios. This should allow us to move TLS further down and should also allow us to start moving things to connection middleware instead of connection adapters.

Summary of the things changed/added:
- Added IConnectionHeartbeatTickFeature that represents a per connection heartbeat with DateTimeOffset.UtcNow passed into the callback. This is trying to preserve the performance characteristics Kestrel had before on the HTTP path. Haven't measured anything here as yet.
- Added IGracefulConnectionLifetimeFeature that represents an attempt to gracefully close the connection that isn't being aborted. This feels pretty awful but we may have to do it.
- Moved connection management to the ConnectionDispatcher and out of the HttpConnectionMiddleware
- Removed Http from the names of the ConnectionManager and Heartbeat

Fixes #2335


davidfowl:
The errors are due to https://github.com/aspnet/KestrelHttpServer/pull/2833 not being merged

src/Kestrel.Core/Internal/HttpConnection.cs
```diff
@@ -100,82 +99,98 @@ public async Task ProcessRequestsAsync<TContext>(IHttpApplication<TContext> http
...
+                                default:
+                                    // SelectProtocol() only returns Http1, Http2 or None.
+                                    throw new NotSupportedException($"{nameof(SelectProtocol)} returned something other than Http1, Http2 or None.");
+                            }
+
+                            _requestProcessor = requestProcessor;
                         }
+                    }
+
+                    _context.Transport.Input.OnWriterCompleted(
```
davidfowl: This may need to be moved earlier in the method just in case.

src/Kestrel.Core/Internal/HttpConnection.cs
```diff
@@ -381,7 +377,7 @@ private HttpProtocols SelectProtocol()
             return http2Enabled && (!hasTls || Http2Id.Span.SequenceEqual(applicationProtocol.Span)) ? HttpProtocols.Http2 : HttpProtocols.Http1;
         }
 
-        public void Tick(DateTimeOffset now)
+        public void Tick(in DateTimeOffset now)
```
davidfowl: Not sure this matters. I can change it back.

benaadams: `DateTimeOffset` is not a readonly struct; have added PR for it https://github.com/dotnet/coreclr/pull/19552

benaadams: Is a 16 byte struct, so will be copied and passed on stack without `in`; however also not a `readonly struct` 😢 so added PR for it  dotnet/coreclr#19552

davidfowl: Err, will change it back.

src/Connections.Abstractions/Features/IGracefulConnectionLifetimeFeature.cs
```diff
@@ -0,0 +1,14 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Threading;
+
+namespace Microsoft.AspNetCore.Connections.Features
+{
+    public interface IGracefulConnectionLifetimeFeature
```
davidfowl: Any naming suggestions?

benaadams:
```csharp
public interface IConnectionLifetimeNotificationFeature
{
    CancellationToken ConnectionCloseRequested { get; set; }
    void RequestClose();
}
```

davidfowl: Hmm, not bad.  ConnectionCloseRequested -> ConnectionClosing?

davidfowl: RequestClose -> Close?

benaadams: Close would definitately work as its parallel is `Abort` so its similar but less severe 
```csharp
public interface IConnectionLifetimeFeature
{
    CancellationToken ConnectionClosed { get; set; }
    void Abort();
}
```

Just wondering how the token use would look in code

```csharp
ConnectionClosingGracefully.Register( ... );

if (ConnectionClosingGracefully.IsCancellationRequested)
```

```csharp
ConnectionCloseRequested.Register( ... );

if (ConnectionCloseRequested.IsCancellationRequested)
```

```csharp
ConnectionClosing.Register( ... );

if (ConnectionClosing.IsCancellationRequested)
```

davidfowl:

src/Connections.Abstractions/Features/IConnectionHeartbeatTickFeature.cs
```diff
@@ -0,0 +1,12 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+
+namespace Microsoft.AspNetCore.Connections.Features
+{
+    public interface IConnectionHeartbeatTickFeature
```
davidfowl: cc @anurse This exists now.

src/Kestrel.Core/Internal/Infrastructure/ConnectionManagerShutdownExtensions.cs
```diff
@@ -8,29 +8,31 @@
...
+        public static async Task<bool> AbortAllConnectionsAsync(this ConnectionManager connectionManager)
         {
             var abortTasks = new List<Task>();
             var canceledException = new ConnectionAbortedException(CoreStrings.ConnectionAbortedDuringServerShutdown);
 
             connectionManager.Walk(connection =>
             {
-                abortTasks.Add(connection.AbortAsync(canceledException));
+                connection.TransportConnection.Abort(canceledException);
+                abortTasks.Add(connection.ExecutionTask);
```
halter73: This used to wait on just the socket being closed and not the middleware completing. On my machine, this is causing GracefulTurnsAbortiveIfRequestsDoNotFinish to consistently fail since NotAllConnectionsAborted now get logged instead of just NotAllConnectionsClosedGracefully. I'm not sure you you managed to get the CI green.

src/Kestrel.Core/Internal/Infrastructure/ConnectionReference.cs
```diff
@@ -0,0 +1,26 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using Microsoft.AspNetCore.Connections;
+
+namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Infrastructure
+{
+    public class ConnectionReference
```
halter73: I'm surprised git didn't treat this as a move. Anything change other than HttpConnection -> KestrelConnection?

src/Kestrel.Transport.Abstractions/Internal/TransportConnection.cs
```diff
@@ -55,6 +61,44 @@ public TransportConnection()
...
+                foreach (var (handler, state) in _heartbeatHandlers)
+                {
+                    handler(now, state);
+                }
+            }
+        }
+
+        public void OnHeartbeat(Action<object> action, object state)
+        {
+            // REVIEW: We could avoid this allocation with 2 lists
```
halter73: This would be solved pretty easily by the caching ISystemClock suggestion.

src/Kestrel.Transport.Sockets/Internal/SocketConnection.cs
```diff
@@ -236,6 +242,9 @@ private async Task DoSend()
 
                 // Complete the output after disposing the socket
                 Output.Complete(error);
+
+                // Cancel any pending flushes so that the input loop is un-paused
+                Input.CancelPendingFlush();
```
halter73: Either way it seems like a good idea to clean up the transport side of things as quickly as possible, but was not doing this before causing any major issues in practice? There's a lot of cleanup that's going to be delayed anyway until the app starts trying to consume the input and/or observes the connection closed.

src/Connections.Abstractions/Features/IConnectionLifetimeNotificationFeature.cs
```diff
@@ -0,0 +1,14 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Threading;
+
+namespace Microsoft.AspNetCore.Connections.Features
+{
+    public interface IConnectionLifetimeNotificationFeature
+    {
+        CancellationToken ConnectionClosing { get; set; }
```
halter73: Why is this better than ApplicationStopping?

halter73 (changes requested):

src/Connections.Abstractions/Features/IConnectionLifetimeNotificationFeature.cs
davidfowl: I don't want to plumb IApplicationLifetime through to every protocol (even though we have one now). This seems to fit more nicely with our existing feature model.

src/Kestrel.Core/Internal/HttpConnection.cs
```diff
@@ -235,12 +243,7 @@ private Http2Connection CreateHttp2Connection(IDuplexPipe transport)
             });
         }
 
-        public void OnConnectionClosed()
-        {
-            _socketClosedTcs.TrySetResult(null);
```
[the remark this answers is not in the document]
davidfowl: Nope, https://github.com/aspnet/KestrelHttpServer/blob/3122bdfc7c85781410d423929aff24c1f30136c3/src/Kestrel.Core/Internal/ConnectionDispatcher.cs#L82

It was pushed lower down the chain.

src/Kestrel.Core/Internal/Infrastructure/ConnectionManagerShutdownExtensions.cs
davidfowl: It waits on both https://github.com/aspnet/KestrelHttpServer/blob/3122bdfc7c85781410d423929aff24c1f30136c3/src/Kestrel.Core/Internal/ConnectionDispatcher.cs#L82. Not sure what you mean.

src/Kestrel.Core/Internal/Infrastructure/ConnectionReference.cs
davidfowl: Nope.

src/Kestrel.Transport.Sockets/Internal/SocketConnection.cs
davidfowl: Yes it was causing a test to fail. I'll find that test. Libuv works fine, sockets doesn't.

src/Kestrel.Transport.Abstractions/Internal/TransportConnection.cs
davidfowl: Yep, I like this!

src/Kestrel.Core/Internal/Infrastructure/KestrelEventSource.cs
```diff
@@ -37,6 +40,19 @@ public void ConnectionStart(HttpConnection connection)
             }
         }
 
+        [NonEvent]
+        public void ConnectionStart(TransportConnection connection)
```
halter73: Can't we remove the HttpConnection overloads of these methods now?

src/Kestrel.Core/Internal/HttpConnection.cs
```diff
@@ -381,6 +377,10 @@ private HttpProtocols SelectProtocol()
             return http2Enabled && (!hasTls || Http2Id.Span.SequenceEqual(applicationProtocol.Span)) ? HttpProtocols.Http2 : HttpProtocols.Http1;
         }
 
+        private void Tick()
+        {
+            Tick(_systemClock.UtcNow);
+        }
```
halter73: Nit: newline

src/Kestrel.Transport.Abstractions/Internal/TransportConnection.cs
```diff
@@ -14,10 +15,15 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Transport.Abstractions.Internal
     public abstract partial class TransportConnection : ConnectionContext
     {
         private IDictionary<object, object> _items;
+        private List<(Action<object> handler, object state)> _heartbeatHandlers;
+        private readonly object _heartbeatLock = new object();
+        private CancellationTokenSource _connectionClosingCts = new CancellationTokenSource();
```
halter73: Shouldn't this be disposed? Particularly because it might not fire. I know we dispose our registration, but other connection middleware may not.

halter73: Also nit: _connectionClosedRequestedCts

src/Kestrel.Transport.Abstractions/Internal/TransportConnection.FeatureCollection.cs
```diff
@@ -20,6 +20,8 @@ public partial class TransportConnection : IHttpConnectionFeature,
                                                IApplicationTransportFeature,
                                                ITransportSchedulerFeature,
                                                IConnectionLifetimeFeature,
+                                               IConnectionHeartbeatFeature,
+                                               IConnectionLifetimeNotificationFeature,
```
halter73: Nit: implement the methods for these interfaces explicitly in this file for consistency. It can be passthrough.

src/Kestrel.Core/Internal/Infrastructure/ConnectionManagerShutdownExtensions.cs
halter73: You're right. It's waiting longer than before causing an unexpected log failing the  GracefulTurnsAbortiveIfRequestsDoNotFinish test. This test isn't running on Jenkins since it only runs on Win 10 and above. The Jenkins Windows agent is running Win 8.

halter73 (approved):
Pending nits.

davidfowl:
As usual disposing cancellation tokens is the root of all evil.

Issue #2788, opened 2018-08-08
Test failure: ClientAbortingConnectionImmediatelyIsNotLoggedHigherThanDebug


ryanbrandenburg:
This test [fails](http://aspnetci/viewLog.html?buildId=525619&buildTypeId=XPlat_Windows_Win2008r2_Universe) occasionally with the following error:
```
Assert.Empty() Failure
Collection: [WriteContext { EventId = 0, Exception = Microsoft.AspNetCore.Connections.ConnectionAbortedException: The connection was aborted
   at System.IO.Pipelines.PipeCompletion.ThrowLatchedException()
   at System.IO.Pipelines.Pipe.GetReadResult(ReadResult& result)
   at System.IO.Pipelines.Pipe.GetReadAsyncResult()
   at Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http.HttpProtocol.ProcessRequests[TContext](IHttpApplication`1 application)
   at Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http.HttpProtocol.ProcessRequestsAsync[TContext](IHttpApplication`1 application) in /_/src/Kestrel.Core/Internal/Http/HttpProtocol.cs:line 463, Formatter = Func`3 { ... }, LoggerName = "Microsoft.AspNetCore.Server.Kestrel", LogLevel = Warning, ... }, WriteContext { EventId = 0, Exception = Microsoft.AspNetCore.Connections.ConnectionAbortedException: The connection was aborted
   at System.IO.Pipelines.PipeCompletion.ThrowLatchedException()
   at System.IO.Pipelines.Pipe.GetReadResult(ReadResult& result)
   at System.IO.Pipelines.Pipe.GetReadAsyncResult()
   at Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http.HttpProtocol.ProcessRequests[TContext](IHttpApplication`1 application)
   at Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http.HttpProtocol.ProcessRequestsAsync[TContext](IHttpApplication`1 application) in /_/src/Kestrel.Core/Internal/Http/HttpProtocol.cs:line 463, Formatter = Func`3 { ... }, LoggerName = "Microsoft.AspNetCore.Server.Kestrel", LogLevel = Warning, ... }]
[...]

PR #2844, opened 2018-08-23
Improve exception handling in SocketConnection

halter73:
- Avoid race where a connection reset observed by both DoSend() and DoReceive()
  resulted in a ConnectionAbortedException being thrown from the input Pipe
  instead of a ConnectionResetException.

Addresses #2788


halter73:
dotnet/corefx#31928 was filed for the Windows Release x64 Build failure.

PR #2847, opened 2018-08-24
Don't advance examined in Http1MessageBody.PumpAsync for canceled reads

halter73:
Addresses #2846


davidfowl:
This is happening now because of my change https://github.com/aspnet/KestrelHttpServer/blob/d318d7b94dd4e48c5718ca35487ab7e3e8b73951/src/Kestrel.Transport.Sockets/Internal/SocketConnection.cs#L243. I *think* it's a bug in pipelines.

halter73:
@davidfowl I think you may be right. I don't actually see how the initial examined value could be used in this case.

We might want to change this pattern everywhere to only advance examine once we've actually examined the data just to be safe. What do you think about doing that @davidfowl? I could submit a new PR for this.

halter73:
Reopening for investigative purposes.

halter73:
@pakrym

PR #2849, opened 2018-08-24
WIP: HttpsConnectionMiddleware

davidfowl:
Porting the HttpsConnectionAdapter to HttpConnectionMiddleware. There are a few things that are interesting:
- The HttpProtocol assumes that closing the output will close the input. I don't love that assumption but I preserved it here.
- Creating pipes is messy when you need to flow settings around (for back pressure and scheduling). It would be better if we just had a custom pipe implementation that decorates the SSLStream.

### DIFF
--- a/src/Kestrel.Core/Internal/HttpsConnectionMiddleware.cs
+++ b/src/Kestrel.Core/Internal/HttpsConnectionMiddleware.cs
@@ -1,0 +1,291 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Buffers;
+using System.Collections.Generic;
+using System.IO;
+using System.IO.Pipelines;
+using System.Net.Security;
+using System.Security.Authentication;
+using System.Security.Cryptography.X509Certificates;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Connections;
+using Microsoft.AspNetCore.Connections.Features;
+using Microsoft.AspNetCore.Http.Features;
+using Microsoft.AspNetCore.Server.Kestrel.Core;
+using Microsoft.AspNetCore.Server.Kestrel.Core.Adapter.Internal;
+using Microsoft.AspNetCore.Server.Kestrel.Core.Features;
+using Microsoft.AspNetCore.Server.Kestrel.Core.Internal;
+using Microsoft.AspNetCore.Server.Kestrel.Transport.Abstractions.Internal;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.AspNetCore.Server.Kestrel.Https.Internal
+{
+    public class HttpsConnectionMiddleware
+    {
+        private readonly ConnectionDelegate _next;
+        private readonly HttpsConnectionAdapterOptions _options;
+        private readonly X509Certificate2 _serverCertificate;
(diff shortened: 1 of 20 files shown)


src/Kestrel.Core/Internal/HttpsConnectionMiddleware.cs
```diff
@@ -0,0 +1,291 @@
...
+                            }
+                        }
+
+                        return true;
+                    });
+
+                certificateRequired = true;
+            }
+
+            using (var cancellationTokeSource = new CancellationTokenSource(_options.HandshakeTimeout))
```
davidfowl: This is no longer using `IConnectionTimeout` since it's added only added in HttpConnection.ProcessRequestAsync

src/Kestrel.Core/ListenOptionsHttpsExtensions.cs
```diff
@@ -210,10 +213,24 @@ internal static bool TryUseHttps(this ListenOptions listenOptions)
         /// <returns>The <see cref="ListenOptions"/>.</returns>
         public static ListenOptions UseHttps(this ListenOptions listenOptions, HttpsConnectionAdapterOptions httpsOptions)
         {
-            var loggerFactory = listenOptions.KestrelServerOptions.ApplicationServices.GetRequiredService<ILoggerFactory>();
+            var loggerFactory = listenOptions.KestrelServerOptions?.ApplicationServices.GetRequiredService<ILoggerFactory>() ?? NullLoggerFactory.Instance;
             // Set the list of protocols from listen options
             httpsOptions.HttpProtocols = listenOptions.Protocols;
-            listenOptions.ConnectionAdapters.Add(new HttpsConnectionAdapter(httpsOptions, loggerFactory));
+            httpsOptions.MaxInputBufferSize = listenOptions.KestrelServerOptions?.Limits.MaxRequestBufferSize;
```
davidfowl: This isn't great but it was the only way I could think of flowing the ServiceContext data.

davidfowl:

src/Kestrel.Core/Internal/HttpsConnectionMiddleware.cs
```diff
@@ -0,0 +1,291 @@
...
+using Microsoft.AspNetCore.Server.Kestrel.Core;
+using Microsoft.AspNetCore.Server.Kestrel.Core.Adapter.Internal;
+using Microsoft.AspNetCore.Server.Kestrel.Core.Features;
+using Microsoft.AspNetCore.Server.Kestrel.Core.Internal;
+using Microsoft.AspNetCore.Server.Kestrel.Transport.Abstractions.Internal;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.AspNetCore.Server.Kestrel.Https.Internal
+{
+    public class HttpsConnectionMiddleware
```
JanEggers: hi @davidfowl can I use this middleware to get tls support for the connection abstraction based mqtt handler I created?
(does this work with tcp or is http(s) required)

https://github.com/chkr1011/MQTTnet/blob/master/Source/MQTTnet.AspnetCore/MqttConnectionHandler.cs

davidfowl: Yes the idea is to have a TlsMiddleware that's separate from Kestrel that can be placed in front of any `ConnectionHandler`. This is the first go to do that within Kestrel itself to see what issues exist.

PR #2861, opened 2018-08-28
Don't preserve the ExecutionContext when dispatching callbacks

davidfowl:
- The layers up stack already preserve and restore the ExecutionContext, there's no need to capture and run for the scheduler callback and cancellation token callback

### DIFF
--- a/src/Connections.Abstractions/DefaultConnectionContext.cs
+++ b/src/Connections.Abstractions/DefaultConnectionContext.cs
@@ -67,7 +67,7 @@ namespace Microsoft.AspNetCore.Connections
 
         public override void Abort(ConnectionAbortedException abortReason)
         {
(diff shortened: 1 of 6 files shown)


Tratcher:
There's more: https://github.com/aspnet/KestrelHttpServer/search?q=QueueUserWorkItem&unscoped_q=QueueUserWorkItem

Why only change these?

davidfowl:
We can for consistency but:
- LibuvThread never captures the execution context (unlike the thread pool) so we're always on the default context when dispatching.
- ThreadPoolAwaitable - Might is likely ok to change since it's only used for async/await (the state machine captures the execution context)
- InMemoryTransort and DefaultConnectionContext don't really matter.

davidfowl:
Branch updated...

halter73:
@davidfowl Do you think we should backport this and/or the HostingLogScope scope change?

PR #2864, opened 2018-08-28
Go Slow

benaadams:
No merge!


sebastienros:
The issue was that I pasted `benaadams:patch-1` as the branch name when it should have been `patch-1`, but I tried it, and it was still fast ;) maybe because it's the platform benchmark.

PR #2865, opened 2018-08-29
Flow the cancellation tokens to ReadAsync and WriteAsync

davidfowl:


src/Kestrel.Core/Internal/Http/MessageBody.cs
```diff
@@ -43,7 +43,7 @@ public virtual async ValueTask<int> ReadAsync(Memory<byte> buffer, CancellationT
 
             while (true)
```
gfoidl: Should cancellation checked here too?

I.e.
```c#
while (!cancellationToken.IsCancellationRequested)
{
    // ....
}
```

Same below on line 84.

davidfowl: Eh, then I'd have to throw in if `cancellationToken.ThrowIfCancellationRequested()`. I'd also need to handle the case where the loop never ran and the token got triggered.

gfoidl: Makes sense -- I missed the throwing-behavior. 

In
https://github.com/aspnet/KestrelHttpServer/blob/105c395bac37186b789496a5656848bf318eb8db/src/Kestrel.Core/Internal/Http/MessageBody.cs#L104
the token could be passed in too. (Can't comment direct on that line here).

davidfowl: good catch

Tratcher:
Just because or did you run into a specific issue?

davidfowl:
I was trying things and it didn't work as expected then I realized that we don't pass in these tokens anywhere.

Tratcher:
Sounds like you need some new tests.

davidfowl:
Yea, I'll add some.

src/Kestrel.Core/Internal/Http/Http1OutputProducer.cs
```diff
@@ -89,7 +89,7 @@ public void Write<T>(Func<PipeWriter, T, long> callback, T state)
             }
         }
 
-        public Task WriteAsync<T>(Func<PipeWriter, T, long> callback, T state)
+        public Task WriteAsync<T>(Func<PipeWriter, T, long> callback, T state, CancellationToken cancellationToken)
```
halter73: This is only used for headers and therefore isn't used.

src/Kestrel.Core/Internal/Http/IHttpOutputProducer.cs
```diff
@@ -12,7 +12,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
     public interface IHttpOutputProducer : IDisposable
     {
         void Abort(ConnectionAbortedException abortReason);
-        Task WriteAsync<T>(Func<PipeWriter, T, long> callback, T state);
+        Task WriteAsync<T>(Func<PipeWriter, T, long> callback, T state, CancellationToken cancellationToken);
```
halter73: The caller never has a cancellationToken.

halter73 (changes requested):
We need at least 1 regression test.

src/Kestrel.Core/Internal/Http/Http1OutputProducer.cs
davidfowl: Why do you say that? Theoretically, you can change the response buffer to a small number and cause WriteAsync to hang on FlushAsync here https://github.com/aspnet/KestrelHttpServer/blob/80c8e83c689a8e2ed52e1c4b7b84c4c2d8ad15ee/src/Kestrel.Core/Internal/Http/Http1OutputProducer.cs#L107

src/Kestrel.Core/Internal/Http/IHttpOutputProducer.cs
davidfowl: Can you elaborate? WriteChunkedAsync calls this method today.

src/Kestrel.Core/Internal/Http/Http1OutputProducer.cs
halter73: My bad, for some reason I thought this was (only) called by WriteResponseHeaders which would never use a CancellationToken. Forgot this was used to write chunked response bodies.

Looking at the code now, I realize we should also just remove the `Write<T>` overload. That's what we used to use for header writing prior to the HTTP/2 refactoring.

It looks like we regressed CancellationToken handling for chunked response bodies in 2.1.0. Maybe we should backport this fix.

src/Kestrel.Core/Internal/Http/IHttpOutputProducer.cs
halter73: See above response. I was thinking about the old code.

davidfowl:
Bleh, flaky tests and leaks a result of the exception. I'll get back to this...

Tratcher (approved):

test/Kestrel.InMemory.FunctionalTests/RequestTests.cs
```diff
@@ -55,6 +56,73 @@ public async Task StreamsAreNotPersistedAcrossRequests()
...
+                    context.Response.ContentLength = 14;
+                    await context.Response.WriteAsync("Read cancelled");
+                }
+
+            }, new TestServiceContext(LoggerFactory)))
+            {
+                using (var connection = server.CreateConnection())
+                {
+                    await connection.Send(
+                        "GET / HTTP/1.1",
```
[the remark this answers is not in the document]
davidfowl: Hah, good catch.

halter73:
Can you remove the sync `Http1OutputProducer.Write<T>(Func<PipeWriter, T, long> callback, T state)` method as part of this change?

halter73 (approved):

davidfowl:

test/Kestrel.InMemory.FunctionalTests/ResponseTests.cs
```diff
@@ -110,12 +110,12 @@ public async Task ResponseBodyWriteAsyncCanBeCancelled()
...
             using (var server = new TestServer(async context =>
             {
-                await context.Response.WriteAsync("hello", cts.Token);
+                await context.Response.WriteAsync("hello", cts.Token).DefaultTimeout();
                 try
                 {
                     var task = context.Response.WriteAsync("world", cts.Token);
                     Assert.False(task.IsCompleted);
-                    await task;
+                    await task.DefaultTimeout();
```
halter73: 👍

davidfowl:

davidfowl:

Issue #2509, opened 2018-04-20
Investigate xUnit "non-serializable data" warnings


mikeharder:
Kestrel tests have hundrends of this scary-looking warning from xunit:

```
[12:02:22] : [Step 4/7]   [xUnit.net 00:00:00.3409135] Microsoft.AspNetCore.Server.Kestrel.Transport.Libuv.Tests: Non-serializable data
('System.Object[]') found for 'Microsoft.AspNetCore.Server.Kestrel.Transport.Libuv.Tests.LibuvTransportTests.TransportCanBindUnbindAndStop';
falling back to single test case.
```

http://aspnetci/viewLog.html?buildId=451476&tab=buildLog&buildTypeId=Lite_KestrelHttpServer&logTab=tree&filter=all&_focus=847#_state=350

They started when xUnit diagnostics messages were enabled (https://github.com/aspnet/KestrelHttpServer/pull/2505).  We should investigate what's causing these warnings, what is their implication (does it mean less tests are run than expected), and how to fix them.

JamesNK:
It's caused by non-serializable parameters in xunit tests. SignalR had the same thing - https://github.com/aspnet/SignalR/issues/1649

All tests are still run, you just don't see them as there own nodes in the VS test explorer.

mikeharder:
Partial fix at https://github.com/aspnet/KestrelHttpServer/pull/2520.  Decided the fix was getting too complicated and it would be better to just disable xUnit diagnostic messages.

PR #2505, opened 2018-04-18
Enable xUnit diagnostic messages

mikeharder:
- Required to log long-running tests
- Will help investigate https://github.com/aspnet/KestrelHttpServer/issues/2503

PR #2866, opened 2018-08-29
Add diagnostic messages to show hanging tests in the output

davidfowl:
Now test hangs will show up.

### DIFF
--- a/test/xunit.runner.json
+++ b/test/xunit.runner.json
@@ -3,5 +3,6 @@
   "appDomain": "denied",
   "methodDisplay": "method",
   "longRunningTestSeconds": 60,
+  "diagnosticMessages": true,
   "maxParallelThreads": -1
 }


halter73:
Now that we store artifacts are pretty much all our build servers, the technique I use to find the hanging test is to run `grep -rL "Finished test"` in the log directory. Only logs for incomplete tests will show up. Unfortunately this doesn't help for unit tests without logging enabled, but hopefully unit tests aren't the ones that are hanging. So maybe that an option if diagnosticMessages are a no-go.

davidfowl:
It's fine, SignalR has it turned on, it's not that bad.

mikeharder:
I was the last person to try enabling this, and I reverted it for the [following reasons](https://github.com/aspnet/KestrelHttpServer/pull/2522):

> - xUnit diagnostic messages were enabled in #2505 to help identify hung tests
> - However, this also enables log messages for non-serializable theory data (https://github.com/aspnet/KestrelHttpServer/issues/2509), which adds noise to logs
> - It doesn't seem possible to disable the "non-serializable data" messages but keep the "long-running test" messages
> - Converting all tests to serializable theory data is expensive and complicates test code
> - The best solution for now is to revert xUnit diagnostic messages to disabled, and re-enable when needed to identify hung tests

davidfowl:
The log isn't that much noiser though https://ci3.dot.net/job/aspnet_KestrelHttpServer/job/release_2.2/job/osx-Configuration_Release_prtest/178/consoleFull.

mikeharder:
The log now has 46 warnings like the following.  Note these add noise to local developer builds as well as CI builds.  We previously decided the noise of these warnings was not worth the benefit of identifying hung tests, and we didn't want to spend the time fixing the non-serializable parameters.

```
[xUnit.net 00:00:00.32] InMemory.FunctionalTests: Non-serializable data ('System.Object[]') found for
'Microsoft.AspNetCore.Server.Kestrel.Core.Tests.Http2ConnectionTests.HEADERS_Received_HeaderBlockContainsDuplicatePseudoHeaderField_StreamError';
falling back to single test case.
```

halter73:
@davidfowl Are you going to close this then?

davidfowl:
No I think this needs to be on.

Issue #2812, opened 2018-08-10
Implement MaxRequestHeadersTotalSize for HTTP/2


Tratcher:
Meta issue: #2805

Total size after decompression. Also count Continuation frames.

How does this apply to Trailers? https://github.com/aspnet/KestrelHttpServer/issues/2051

See SETTINGS_MAX_HEADER_LIST_SIZE

https://tools.ietf.org/html/rfc7540#section-10.5.1
"   A server that receives a larger header block than it is willing to
   handle can send an HTTP 431 (Request Header Fields Too Large) status
   code [RFC6585].  A client can discard responses that it cannot
   process.  The header block MUST be processed to ensure a consistent
   connection state, unless the connection is closed."

https://tools.ietf.org/html/rfc7540#section-6.5.2
"The value is based on the
      uncompressed size of header fields, including the length of the
      name and value in octets plus an overhead of 32 octets for each
      header field."

Http/1 cross reference:
It sends a 413 response and tears down the connection.
https://github.com/aspnet/KestrelHttpServer/blob/83488886e0ed32e946af2a25ae7f29ce9bcc4273/src/Kestrel.Core/Internal/Http/Http1Connection.cs#L165-L179

Tratcher:
Protocol default: unlimited
Nginx default: 16k http://nginx.org/en/docs/http/ngx_http_v2_module.html#http2_max_header_size

Tratcher:
Related: What if the response headers are larger than the client's header size limit?

Issue #2811, opened 2018-08-10
Implement MaxRequestHeaderCount for HTTP/2


Tratcher:
Meta issue: https://github.com/aspnet/KestrelHttpServer/issues/2805

How would this apply to trailers? https://github.com/aspnet/KestrelHttpServer/issues/2051

Tratcher:
This is already implemented in the base HttpProtocol
https://github.com/aspnet/KestrelHttpServer/blob/83488886e0ed32e946af2a25ae7f29ce9bcc4273/src/Kestrel.Core/Internal/Http/HttpProtocol.cs#L435-L441

Only needs tests.

Issue #2051, opened 2017-09-07
Put trailers in a separate collection


cesarbs:
Right now we merge them with request headers. They should go into their own collection. There should also be some way (likely a `Task`) to allow an application to wait until trailers have been received.

cesarbs:
@muratg Can you triage?

cc @Tratcher

Tratcher:
Ideas:
Add a new IHttpRequestTrailersFeature.
- bool ExpectTrailers - returns true if the request is chunked and the Trailers header has any values.
- bool TrailersAvailable - returns true if the Trailers collection has been populated. Redundant with Trailers != null?
- bool OnTrailers(Func<object, Task> callback, object state) - registers for a callback that gets invoked when trailers arrive. Compare to OnSendingHeaders. Multiple callbacks can be registered and they will be invoked in order. Returns false if unable to register, such as if ExpectTrailers is false, or if trailers have already arrived.
- IHeaderCollection Trailers - Null until trailers have arrived to avoid race conditions.

HttpRequest should have extension methods for each of the feature methods and properties.

Coss server notes:
- This likely can't be implemented by HttpSysServer or IIS as Http.Sys filters out request headers. Nor would it work when using IIS (ANCM) as a reverse proxy.

halter73:
> bool TrailersAvailable - returns true if the Trailers collection has been populated. Is this redundant with Trailers != null?

I think this bool is redundant. A Task instead of a bool might be more useful, but it would likely cause all the same concurrency problems as a callback but worse since there's no good way to observe the completion of continuations developers might attach to it. As you mention, the right way to wait is to simply consume the entire request body first.

> IHeaderCollection Trailers - Null until trailers have arrived to avoid race conditions

This feels gross. I think I'd rather throw an exception with a useful error message explaining to the developer they have to wait before accessing the trailers.

Tratcher:
>> IHeaderCollection Trailers - Null until trailers have arrived to avoid race conditions

> This feels gross. I think I'd rather throw an exception with a useful error message explaining to the developer they have to wait before accessing the trailers.

I wouldn't throw unless you had a bool to check if the property was safe. That's why we have HasStarted you can check before setting response headers. Throwing from a get seems even worse.

Issue #2813, opened 2018-08-10
Implement MaxRequestLineSize for HTTP/2


Tratcher:
Meta issue: #2805

The request line has no direct analog in HTTP/2. Proposal: sum up the the lengths of the required pseudo headers (:scheme, :authority, :path, :method).

PR #2870, opened 2018-08-30
Implement MaxRequestHeadersTotalSize for HTTP/2, improve error handling

Tratcher:
#2812, #2811

Counting the size of headers as they're received is pretty straight forward. Note this count is tracked separately for headers and trailers.

I've temporarily disabled http/2 request trailers to avoid multithreaded dictionary corruption. #2051 will re-implement them properly.

Most of the changes are around error handling. Because all streams share one header decompression state you can't stop processing a headers frame part way through without risking corruption to the decompression state for future streams. The safest action in the short term is to treat all header errors as connection errors. Most of these errors are for basic protocol correctness and should not happen in reality.

A few errors like header length limits will be more common. We'll want to overhaul the error handling mechanics to allow request validation errors to be handled on a per stream basis. To do this we would need to:
- Capture errors from request/header validation
- Continue to decompress Headers and Continuation frames to maintain the compression state while discarding the headers.
- Optionally send a 4XX response as appropriate for the error
- Reset the stream

I didn't want to start that refactor yet as having the limits working is more important. Well behaved clients won't be impacted.

### DIFF
--- a/src/Kestrel.Core/Internal/Http2/Http2Connection.cs
+++ b/src/Kestrel.Core/Internal/Http2/Http2Connection.cs
@@ -953,7 +953,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http2
         {
             // https://tools.ietf.org/html/rfc7540#section-6.5.2
             // "The value is based on the uncompressed size of header fields, including the length of the name and value in octets plus an overhead of 32 octets for each header field.";
-            _totalParsedHeaderSize += 32 + name.Length + value.Length;
+            _totalParsedHeaderSize += HeaderField.RfcOverhead + name.Length + value.Length;
             if (_totalParsedHeaderSize > _context.ServiceContext.ServerOptions.Limits.MaxRequestHeadersTotalSize)
             {
                 throw new Http2ConnectionErrorException(CoreStrings.BadRequest_HeadersExceedMaxTotalSize, Http2ErrorCode.PROTOCOL_ERROR);
--- a/src/Kestrel.Core/Internal/Http2/Http2Connection.cs
+++ b/src/Kestrel.Core/Internal/Http2/Http2Connection.cs
@@ -75,6 +75,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http2
         private RequestHeaderParsingState _requestHeaderParsingState;
         private PseudoHeaderFields _parsedPseudoHeaderFields;
         private Http2HeadersFrameFlags _headerFlags;
+        private int _totalParsedHeaderSize;
         private bool _isMethodConnect;
         private readonly object _stateLock = new object();
         private int _highestOpenedStreamId;
@@ -92,6 +93,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http2
             _serverSettings.HeaderTableSize = (uint)context.ServiceContext.ServerOptions.Limits.Http2.HeaderTableSize;
             _hpackDecoder = new HPackDecoder((int)_serverSettings.HeaderTableSize);
             _incomingFrame = new Http2Frame(_serverSettings.MaxFrameSize);
+            _serverSettings.MaxHeaderListSize = (uint)context.ServiceContext.ServerOptions.Limits.MaxRequestHeadersTotalSize;
         }
 
         public string ConnectionId => _context.ConnectionId;
@@ -888,6 +890,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http2
             _parsedPseudoHeaderFields = PseudoHeaderFields.None;
             _headerFlags = Http2HeadersFrameFlags.NONE;
             _isMethodConnect = false;
+            _totalParsedHeaderSize = 0;
         }
 
         private void ThrowIfIncomingFrameSentToIdleStream()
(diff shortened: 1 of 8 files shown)


Tratcher:
Added MaxRequestLineSize #2813. People won't hit this much in practice since each of these fields are headers and our HPack implementation limits headers to 4kb. See #2872

test/Kestrel.InMemory.FunctionalTests/HttpProtocolSelectionTests.cs
```diff
@@ -35,11 +35,34 @@ public Task Server_Http1Only_Cleartext_Success()
...
+            using (var server = new TestServer(context => Task.CompletedTask, testContext, listenOptions))
+            {
+                using (var connection = server.CreateConnection())
+                {
+                    await connection.Send(Encoding.ASCII.GetString(Http2Connection.ClientPreface));
+                    // Can't use Receive when expecting binary data
+                    var actual = new byte[expected.Length];
+                    var read = await connection.Stream.ReadAsync(actual, 0, actual.Length);
+                    Assert.Equal(expected.Length, read);
+                    Assert.Equal(expected, actual);
```
halter73: Nit: read again and assert read is 0.

test/Kestrel.InMemory.FunctionalTests/HttpProtocolSelectionTests.cs
```diff
@@ -35,11 +35,34 @@ public Task Server_Http1Only_Cleartext_Success()
...
         {
             // Expect a SETTINGS frame (type 0x4) with default settings
-            return TestSuccess(HttpProtocols.Http2, Encoding.ASCII.GetString(Http2Connection.ClientPreface),
-                "\x00\x00\x06\x04\x00\x00\x00\x00\x00\x00\x03\x00\x00\x00\x64");
+            var expected = new byte[]
+            {
+                0x00, 0x00, 0x0C, // Payload Length (6 * settings count)
+                0x04, 0x00, 0x00, 0x00, 0x00, 0x00, // SETTINGS frame (type 0x04)
+                0x00, 0x03, 0x00, 0x00, 0x00, 0x64, // Connection limit
+                0x00, 0x06, 0x00, 0x00, 0x80, 0x00 // Header size limit
```
halter73: Any particular reason for adding the header size limit to the settings frame here?

Tratcher: Kestrel sends all non-protocol-default settings at the start of every connection. This output is used here to confirm kestrel is treating the connection as HTTP/2.

halter73: Right. I had this confused with Http2Connection.ClientPreface.

halter73: It doesn't make much sense for the *client* to send these settings, but I didn't really think it through fully.

src/Kestrel.Core/Internal/Http2/Http2Connection.cs
```diff
@@ -944,23 +946,65 @@ void IHttp2StreamLifetimeHandler.OnStreamCompleted(int streamId)
...
         }
 
+        // We can't throw a Http2StreamErrorException here, it interrupts the header decompression state and may corrupt subsequent header frames on other streams.
+        // For now these either need to be connection errors or BadRequests. If we want to downgrade any of them to stream errors later then we need to
+        // rework the flow so that the remaining headers are drained and the decompression state is maintained.
         public void OnHeader(Span<byte> name, Span<byte> value)
         {
+            // https://tools.ietf.org/html/rfc7540#section-6.5.2
+            // "The value is based on the uncompressed size of header fields, including the length of the name and value in octets plus an overhead of 32 octets for each header field.";
+            _totalParsedHeaderSize += 32 + name.Length + value.Length;
```
JunTaoLuo: nit: `HeaderField.RfcOverhead`? I would prefer if this was a named constant instead.

test/Kestrel.InMemory.FunctionalTests/Http2/Http2ConnectionTests.cs
```diff
@@ -1562,7 +1566,62 @@ private async Task HEADERS_Received_InvalidHeaderFields_StreamError(IEnumerable<
...
+                new KeyValuePair<string, string>("h", _4kHeaderValue),
+            };
+
+            return HEADERS_Received_InvalidHeaderFields_ConnectionError(headers, CoreStrings.BadRequest_HeadersExceedMaxTotalSize);
+        }
+
+        [Fact]
+        public Task HEADERS_Received_TooManyHeaders_ConnectionError()
+        {
+            // > MaxRequestHeaderCount (100
```
JunTaoLuo: nit: missing closing parenthesis.

test/Kestrel.InMemory.FunctionalTests/Http2/Http2ConnectionTests.cs
```diff
@@ -2005,18 +2064,23 @@ public async Task SETTINGS_KestrelDefaults_Sent()
             await SendSettingsAsync();
 
             var frame = await ExpectAsync(Http2FrameType.SETTINGS,
-                withLength: 6,
+                withLength: 6 * 2,
```
JunTaoLuo: nit: This section looks like it can be extracted.

test/Kestrel.InMemory.FunctionalTests/HttpProtocolSelectionTests.cs
Tratcher: It blocks, the connection hasn't been closed from the other end.

halter73: I guess using a 1-byte larger expected array would be good enough. Or sending a go away frame, but that's probably needlessly complicated.

Issue #2874, opened 2018-09-01
SETTINGS_HEADER_TABLE_SIZE doesn't work with Chrome


Tratcher:
https://github.com/aspnet/KestrelHttpServer/pull/2838

```
Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http2.Http2SettingsParameterOutOfRangeException
  HResult=0x80131500
  Message=HTTP/2 SETTINGS parameter SETTINGS_HEADER_TABLE_SIZE must be set to a value between 0 and 4096
  Source=Microsoft.AspNetCore.Server.Kestrel.Core
  StackTrace:
   at Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http2.Http2PeerSettings.Update(IList`1 settings) in D:\github\AspNet\KestrelHttpServer\src\Kestrel.Core\Internal\Http2\Http2PeerSettings.cs:line 47
```
https://github.com/aspnet/KestrelHttpServer/blob/153020ef8a81aa6b4c6efebd83a8b04863e4d40d/src/Kestrel.Core/Internal/Http2/Http2PeerSettings.cs#L40-L46

Repro: The checked in Http2 sample app.

IE and Edge work. Chrome is sending 65536. This bug prevents any HTTP/2 connections from Chrome. There's no workaround. @JunTaoLuo @muratg I guess now we get to find out if anybody ever tries preview2...

https://tools.ietf.org/html/rfc7540#section-6.5.2
```
   SETTINGS_HEADER_TABLE_SIZE (0x1):  Allows the sender to inform the
      remote endpoint of the maximum size of the header compression
      table used to decode header blocks, in octets.  The encoder can
      select any size equal to or less than this value by using
      signaling specific to the header compression format inside a
      header block (see [COMPRESSION]).  The initial value is 4,096
      octets.
```

This was misinterpreted as to never allow a value over 4kb. What it really meant was that the side compressing headers must not grow the table beyond the size specified by the one decompressing the headers. There should be no limit enforced on the client's value, we should use the lesser of the client value and our own value, and our value should be allowed to be larger than 4kb.
https://github.com/aspnet/KestrelHttpServer/blob/153020ef8a81aa6b4c6efebd83a8b04863e4d40d/src/Kestrel.Core/Http2Limits.cs#L53-L56

PR #2877, opened 2018-09-02
Remove limits on SETTINGS_HEADER_TABLE_SIZE

Tratcher:
#2874 This fixes a regression that prevented Kestrel from accepting HTTP/2 connections from Chrome. 

Preview2 reset candidate.

### DIFF
--- a/src/Kestrel.Core/Http2Limits.cs
+++ b/src/Kestrel.Core/Http2Limits.cs
@@ -2,6 +2,7 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http2;
 
 namespace Microsoft.AspNetCore.Server.Kestrel.Core
 {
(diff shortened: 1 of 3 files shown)


Tratcher:
There's a few things going on. The request table and response table need to be treated separately. The server gets to limit how large the table is for requests, and the client gets to limit how large it is for responses. Each advertise that limit to the other. The recpient is under no obligation to allocate or use that much memory for the table, they can use as much or as little as they want up to given limit. We don't currently use a response compression table at all so the limit given to us by the client is meaningless. When we do start compressing response headers we should use the lesser table size of either the value provided by the client or the one specified in Http2Limits.

JunTaoLuo:
Right there are separate limits for each direction. But there's no upper
limit to this seeing other than the number of bits allocated to this
setting?

Tratcher:
Nothing in the spec, no, and no reason to impose a limit in either direction other than the one provided by Http2Limits.

JunTaoLuo:
I read the HTTP/2 and HPACK spec again and there weren't any additional limits. Seems like I understood the phrase `The encoder can select any size equal to or less than this value` wrong thinking it was referring to the initial value. This fix is correct.

halter73:
991cfa8

Issue #2858, opened 2018-08-27
Change how HTTP/2 frames are parsed and generated


Tratcher:
The Http2FrameReader reads the exact structure of the incoming frame into a memory blob before it begins to parse it, first reading the header to get the length and then reading the payload.
https://github.com/aspnet/KestrelHttpServer/blob/b039fb73c41468b6d6a28770285ba432d328023e/src/Kestrel.Core/Internal/Http2/Http2FrameReader.cs#L22-L36

This amounts to triple buffering of some frames and is unnecessary for parsing. The Http2FrameWriter does something similar composing the exact frame bytes into a whole frame before sending it.
https://github.com/aspnet/KestrelHttpServer/blob/b039fb73c41468b6d6a28770285ba432d328023e/src/Kestrel.Core/Internal/Http2/Http2FrameWriter.cs#L198-L223

Alternative: Frames can be parsed directly from the incoming pipe buffer. The logical fields like flags, length, frametype, etc. can still be written to a Http2Frame object, but it doesn't have to be stored in it's serialized state. 
- E.g. Make sure you have received the 9 byte frame header,
- parse the individual fields in order,
 - check the frame type and flags, 
 - make sure you have enough bytes for any supplementary frame fields, parse those, 
 - wait for the rest of the frame body to arrive,
 - then pass the pipe buffer directly to the associated handler. 
 - The header parser can parse directly from that buffer. The data frame handler can copy directly from that buffer to the stream's request body pipe.

The only place you need to allocate a large buffer is for the response header compression, the whole frame needs to be compressed before you can know the exact headers length. Interestingly this compression takes place under the write lock, so that buffer could be rented from the pipe itself to avoid double buffering.

@davidfowl

halter73:
Completely agreed. @davidfowl and I were discussing this last week.

PR #2893, opened 2018-09-06
Change how HTTP/2 frames are parsed and generated

Tratcher:
#2858 I was able to reduce the request buffer from 16kb to 15 bytes. The response buffer is still needed for compressing headers.

This helped eliminate several data copies for request headers, request and response data, and settings frames.

Converting Http2Frame to a POCO helped eliminate a lot of redundant de/serialization of properties that were accessed multiple times.

Measurements needed.

### DIFF
--- a/src/Kestrel.Core/Internal/Http2/HPack/HPackDecoder.cs
+++ b/src/Kestrel.Core/Internal/Http2/HPack/HPackDecoder.cs
@@ -2,6 +2,7 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.Buffers;
 using Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http;
 
 namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http2.HPack
@@ -109,11 +110,15 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http2.HPack
             _dynamicTable = dynamicTable;
         }
 
-        public void Decode(Span<byte> data, bool endHeaders, IHttpHeadersHandler handler)
+        public void Decode(ReadOnlySequence<byte> data, bool endHeaders, IHttpHeadersHandler handler)
         {
-            for (var i = 0; i < data.Length; i++)
+            foreach (var segment in data)
             {
-                OnByte(data[i], handler);
+                var span = segment.Span;
+                for (var i = 0; i < span.Length; i++)
+                {
+                    OnByte(span[i], handler);
+                }
             }
 
             if (endHeaders)
--- a/src/Kestrel.Core/Internal/Http2/HPack/HPackEncoder.cs
+++ b/src/Kestrel.Core/Internal/Http2/HPack/HPackEncoder.cs
@@ -24,7 +24,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http2.HPack
             _enumerator.MoveNext();
 
             var statusCodeLength = EncodeStatusCode(statusCode, buffer);
-            var done = Encode(buffer.Slice(statusCodeLength), out var headersLength);
+            var done = Encode(buffer.Slice(statusCodeLength), throwIfNoneEncoded: false, out var headersLength);
             length = statusCodeLength + headersLength;
 
             return done;
@@ -32,12 +32,21 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http2.HPack
 
         public bool Encode(Span<byte> buffer, out int length)
         {
+            return Encode(buffer, throwIfNoneEncoded: true, out length);
+        }
+
+        private bool Encode(Span<byte> buffer, bool throwIfNoneEncoded, out int length)
+        {
             length = 0;
 
             do
             {
                 if (!EncodeHeader(_enumerator.Current.Key, _enumerator.Current.Value, buffer.Slice(length), out var headerLength))
                 {
+                    if (length == 0 && throwIfNoneEncoded)
+                    {
+                        throw new HPackEncodingException(CoreStrings.HPackErrorNotEnoughBuffer);
+                    }
                     return false;
                 }
 
--- a/src/Kestrel.Core/Internal/Http2/Http2Connection.cs
+++ b/src/Kestrel.Core/Internal/Http2/Http2Connection.cs
@@ -69,7 +69,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http2
         private readonly Http2PeerSettings _serverSettings = new Http2PeerSettings();
         private readonly Http2PeerSettings _clientSettings = new Http2PeerSettings();
 
-        private readonly Http2Frame _incomingFrame;
+        private readonly Http2Frame _incomingFrame = new Http2Frame();
 
         private Http2Stream _currentHeadersStream;
         private RequestHeaderParsingState _requestHeaderParsingState;
@@ -92,7 +92,6 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http2
             _serverSettings.MaxFrameSize = (uint)context.ServiceContext.ServerOptions.Limits.Http2.MaxFrameSize;
             _serverSettings.HeaderTableSize = (uint)context.ServiceContext.ServerOptions.Limits.Http2.HeaderTableSize;
             _hpackDecoder = new HPackDecoder((int)_serverSettings.HeaderTableSize);
-            _incomingFrame = new Http2Frame(_serverSettings.MaxFrameSize);
             _serverSettings.MaxHeaderListSize = (uint)context.ServiceContext.ServerOptions.Limits.MaxRequestHeadersTotalSize;
         }
 
@@ -188,16 +187,21 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http2
                     var result = await Input.ReadAsync();
                     var readableBuffer = result.Buffer;
                     var consumed = readableBuffer.Start;
-                    var examined = readableBuffer.End;
+                    var examined = readableBuffer.Start;
 
                     try
                     {
                         if (!readableBuffer.IsEmpty)
                         {
-                            if (Http2FrameReader.ReadFrame(readableBuffer, _incomingFrame, _serverSettings.MaxFrameSize, out consumed, out examined))
+                            if (Http2FrameReader.ReadFrame(readableBuffer, _incomingFrame, _serverSettings.MaxFrameSize, out var framePayload))
                             {
                                 Log.Http2FrameReceived(ConnectionId, _incomingFrame);
-                                await ProcessFrameAsync(application);
+                                consumed = examined = framePayload.End;
+                                await ProcessFrameAsync(application, framePayload);
+                            }
+                            else
+                            {
+                                examined = readableBuffer.End;
                             }
                         }
 
@@ -372,7 +376,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http2
             return true;
         }
 
-        private Task ProcessFrameAsync<TContext>(IHttpApplication<TContext> application)
+        private Task ProcessFrameAsync<TContext>(IHttpApplication<TContext> application, ReadOnlySequence<byte> payload)
         {
             // http://httpwg.org/specs/rfc7540.html#rfc.section.5.1.1
             // Streams initiated by a client MUST use odd-numbered stream identifiers; ...
@@ -386,31 +390,31 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http2
             switch (_incomingFrame.Type)
             {
                 case Http2FrameType.DATA:
-                    return ProcessDataFrameAsync();
+                    return ProcessDataFrameAsync(payload);
                 case Http2FrameType.HEADERS:
-                    return ProcessHeadersFrameAsync(application);
+                    return ProcessHeadersFrameAsync(application, payload);
                 case Http2FrameType.PRIORITY:
                     return ProcessPriorityFrameAsync();
                 case Http2FrameType.RST_STREAM:
                     return ProcessRstStreamFrameAsync();
                 case Http2FrameType.SETTINGS:
-                    return ProcessSettingsFrameAsync();
+                    return ProcessSettingsFrameAsync(payload);
                 case Http2FrameType.PUSH_PROMISE:
                     throw new Http2ConnectionErrorException(CoreStrings.Http2ErrorPushPromiseReceived, Http2ErrorCode.PROTOCOL_ERROR);
                 case Http2FrameType.PING:
-                    return ProcessPingFrameAsync();
+                    return ProcessPingFrameAsync(payload);
                 case Http2FrameType.GOAWAY:
                     return ProcessGoAwayFrameAsync();
                 case Http2FrameType.WINDOW_UPDATE:
                     return ProcessWindowUpdateFrameAsync();
                 case Http2FrameType.CONTINUATION:
-                    return ProcessContinuationFrameAsync(application);
+                    return ProcessContinuationFrameAsync(application, payload);
                 default:
                     return ProcessUnknownFrameAsync();
             }
         }
 
-        private Task ProcessDataFrameAsync()
+        private Task ProcessDataFrameAsync(ReadOnlySequence<byte> payload)
         {
             if (_currentHeadersStream != null)
             {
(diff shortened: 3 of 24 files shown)


Tratcher:
@geoffkizer FYI, are you using the Http2Frame type at all?

geoffkizer:
Not currently, but thanks for the heads up

Tratcher:
I had @sebastienros run this in the lab and he saw no performance change. That's not unexpected since this should only improve per connection memory and the standard perf tests focus on stream density.

src/Kestrel.Core/Internal/Http2/Http2FrameReader.cs
```diff
@@ -3,40 +3,245 @@
...
+
+        private const int LengthOffset = 0;
+        private const int TypeOffset = 3;
+        private const int FlagsOffset = 4;
+        private const int StreamIdOffset = 5;
+
+        public const int SettingSize = 6; // 2 bytes for the id, 4 bytes for the value.
+
+        // No frame type needs more than 8 bytes of additional fields. Payload data beyond the known fields is not stored in this buffer.
+        private readonly byte[] _extendedHeader = new byte[HeaderLength + 8];
```
JunTaoLuo: Why not make this a static field instead? That would allow us to keep this class static and I believe static classes have a small performance benefit. Did a quick test on a local branch and it seems to work.

JunTaoLuo: Nevermind, we need a buffer per connection.

src/Kestrel.Core/Internal/Http2/Http2Connection.cs
```diff
@@ -606,7 +601,7 @@ private Task ProcessRstStreamFrameAsync()
             return Task.CompletedTask;
         }
 
-        private async Task ProcessSettingsFrameAsync()
+        private Task ProcessSettingsFrameAsync(ReadOnlySequence<byte> payload)
```
JunTaoLuo: part of this was done in https://github.com/aspnet/KestrelHttpServer/commit/de5ccb5c78db1cfd13b93579738be6d80e0ecb08

src/Kestrel.Core/Internal/Http2/Http2FrameWriter.cs
```diff
@@ -109,14 +113,34 @@ public Task Write100ContinueAsync(int streamId)
...
-                _continueBytes.CopyTo(_outgoingFrame.HeadersPayload);
-
-                return WriteFrameUnsynchronizedAsync();
+                WriteHeaderUnsynchronized();
+                _outputWriter.Write(_continueBytes);
+                return _flusher.FlushAsync();
             }
         }
 
+        // TODO: Why doesn't this method flush? Is that because we always write a data frame afterwards (even if it's empty)?
```
JunTaoLuo: I would assume so, but is there a penalty to flushing early?

src/Kestrel.Core/Internal/Http2/Http2FrameWriter.cs
```diff
@@ -182,45 +206,54 @@ public Task WriteDataAsync(int streamId, StreamOutputFlowControl flowControl, Re
...
+                foreach (var buffer in currentData)
                 {
-                    current.Span.CopyTo(payload);
-                    payload = payload.Slice(current.Length);
-                    unwrittenPayloadLength += current.Length;
+                    _outputWriter.Write(buffer.Span);
                 }
+
+                // Plus padding
+
```
JunTaoLuo: nit: extra line

src/Kestrel.Core/Internal/Http2/Http2FrameWriter.cs
```diff
@@ -271,71 +304,194 @@ private async Task WriteDataAsyncAwaited(int streamId, StreamOutputFlowControl f
...
             {
+                if (_completed)
+                {
+                    return Task.CompletedTask;
+                }
+
                 _outgoingFrame.PrepareRstStream(streamId, errorCode);
-                return WriteFrameUnsynchronizedAsync();
+                WriteHeaderUnsynchronized();
+                var buffer = _outputWriter.GetSpan(4);
```
JunTaoLuo: These three lines writing a uint/31-bit int/short could be extracted to extension methods.

src/Kestrel.Core/Internal/Http2/Http2FrameWriter.cs
```diff
@@ -271,71 +304,194 @@ private async Task WriteDataAsyncAwaited(int streamId, StreamOutputFlowControl f
...
+                {
+                    return Task.CompletedTask;
+                }
+
+                _outgoingFrame.PrepareSettings(Http2SettingsFrameFlags.NONE);
+                var settingsSize = settings.Count * Http2FrameReader.SettingSize;
+                _outgoingFrame.PayloadLength = settingsSize;
+                WriteHeaderUnsynchronized();
+
+                var buffer = _outputWriter.GetSpan(settingsSize).Slice(0, settingsSize); // GetSpan isn't precise
```
JunTaoLuo: How so? I'm wondering if it affects other places where GetSpan is called, especially on irregular numbers such as Http2FrameReader.HeaderLength

test/Kestrel.InMemory.FunctionalTests/Http2/Http2TestBase.cs
```diff
@@ -345,90 +351,158 @@ protected async Task StartStreamAsync(int streamId, IEnumerable<KeyValuePair<str
...
 
-            var frame = new Http2Frame(Http2Limits.MinAllowedMaxFrameSize);
+            var frame = new Http2Frame();
             frame.PrepareHeaders(Http2HeadersFrameFlags.END_HEADERS | Http2HeadersFrameFlags.PADDED | Http2HeadersFrameFlags.PRIORITY, streamId);
             frame.HeadersPadLength = padLength;
             frame.HeadersPriorityWeight = priority;
             frame.HeadersStreamDependency = streamDependency;
 
-            _hpackEncoder.BeginEncode(headers, frame.HeadersPayload, out var length);
+            var extendedHeaderLength = 6; // stream dependency + weight
```
JunTaoLuo: nit: padding + ..

test/Kestrel.InMemory.FunctionalTests/Http2/Http2TestBase.cs
```diff
@@ -345,90 +351,158 @@ protected async Task StartStreamAsync(int streamId, IEnumerable<KeyValuePair<str
...
+            var buffer = _headerEncodingBuffer.AsSpan();
+            var extendedHeader = buffer.Slice(0, extendedHeaderLength);
+            extendedHeader[0] = padLength;
+            var payload = buffer.Slice(extendedHeaderLength, buffer.Length - padLength - extendedHeaderLength);
+
+            _hpackEncoder.BeginEncode(headers, payload, out var length);
+            var padding = buffer.Slice(extendedHeaderLength + length, padLength);
+            padding.Fill(0);
 
             frame.PayloadLength = 1 + length + padLength;
```
JunTaoLuo: nit: 1 => extendedHeaderLength

src/Kestrel.Core/Internal/Http2/HPack/HPackDecoder.cs
```diff
@@ -108,11 +109,15 @@ internal HPackDecoder(int maxDynamicTableSize, DynamicTable dynamicTable)
             _dynamicTable = dynamicTable;
         }
 
-        public void Decode(Span<byte> data, bool endHeaders, IHttpHeadersHandler handler)
+        public void Decode(ReadOnlySequence<byte> data, bool endHeaders, IHttpHeadersHandler handler)
         {
-            for (var i = 0; i < data.Length; i++)
+            foreach (var segment in data)
```
davidfowl: This is going to be inefficient, but I'm not sure it's worth it to fix this until we remove OnByte

src/Kestrel.Core/Internal/Http2/Http2FrameReader.cs
```diff
@@ -3,40 +3,245 @@
...
+                case Http2FrameType.CONTINUATION: // None
+                case Http2FrameType.PUSH_PROMISE: // Not implemented frames are ignored at this phase
+                default:
+                    return 0;
+            }
+        }
+
+        public IList<Http2PeerSetting> ReadSettings(ReadOnlySequence<byte> payload)
+        {
+            var data = payload.ToArray().AsSpan();
```
davidfowl: This is super inefficient. There's a GetSpan helper in Kestrel that only allocates if it needs to. Use that.

src/Kestrel.Core/Internal/Http2/Http2FrameReader.cs
```diff
@@ -3,40 +3,245 @@
...
-            if (readableBuffer.Length < Http2Frame.HeaderLength)
+            if (readableBuffer.Length < HeaderLength)
             {
                 return false;
             }
 
-            var headerSlice = readableBuffer.Slice(0, Http2Frame.HeaderLength);
-            headerSlice.CopyTo(frame.Raw);
+            var headerSlice = readableBuffer.Slice(0, HeaderLength);
+            headerSlice.CopyTo(_extendedHeader);
```
davidfowl: Why do you copy at all?

Tratcher: Because parsing out of a `ReadOnlySequence<byte>` looked more complex than these 15 bytes were worth.

davidfowl: You don't need to copy though, you can just do what we do elsewhere.  Use the [ToSpan](https://github.com/aspnet/KestrelHttpServer/blob/d3d7c5519810ad97997f777c8a3683a9adbbe1e1/src/Kestrel.Core/Internal/Http/PipelineExtensions.cs#L20) extension method and 80% of the time it won't allocate. 😄

Tratcher: That's a question of odds then. One allocation per connection, or one allocation every time it ToSpan fails. If it fails more than once per connection then you're net negative.

Also we have to do it in two segments since the extended headers are variable length.

src/Kestrel.Core/Internal/Http2/Http2FrameWriter.cs
Tratcher: Flushing early creates more TLS & TCP packets.

src/Kestrel.Core/Internal/Http2/Http2FrameWriter.cs
Tratcher: If there were duplicates it would be worth extracting, but 
every usage of WriteUInt32BigEndian ends up being a little different.

src/Kestrel.Core/Internal/Http2/Http2FrameReader.cs
davidfowl: Does this logic run once per connection?

Tratcher: No, it's on every frame.

src/Kestrel.Core/Internal/Http2/Http2FrameWriter.cs
Tratcher: GetSpan gives you the current buffer if your hint size will fit in it, otherwise it gives you the next buffer. It doesn't slice it down to the size your asked for though. It's also limited to 4kb.

src/Kestrel.Core/Internal/Http2/Http2FrameReader.cs
davidfowl: Then it's worth changing.

Tratcher:
Rebased and updated

src/Kestrel.Core/Internal/Http2/Http2Connection.cs
```diff
@@ -639,14 +633,15 @@ private Task ProcessSettingsFrameAsync()
...
 
                 // Ack before we update the windows, they could send data immediately.
-                var ackTask =  _frameWriter.WriteSettingsAckAsync();
+                var ackTask = _frameWriter.WriteSettingsAckAsync();
 
                 if (_clientSettings.MaxFrameSize != previousMaxFrameSize)
                 {
-                    _frameWriter.UpdateMaxFrameSize(_clientSettings.MaxFrameSize);
+                    // Don't let the client choose an arbitrarily large size, this will be used for response buffers.
+                    _frameWriter.UpdateMaxFrameSize(Math.Min(_clientSettings.MaxFrameSize, _serverSettings.MaxFrameSize));
```
JunTaoLuo: Can we add a new test to ensure the minimum of the two values is always picked?

src/Kestrel.Core/Internal/Http2/HPack/HPackEncoder.cs
```diff
@@ -24,20 +24,29 @@ public bool BeginEncode(int statusCode, IEnumerable<KeyValuePair<string, string>
...
         {
             length = 0;
 
             do
             {
                 if (!EncodeHeader(_enumerator.Current.Key, _enumerator.Current.Value, buffer.Slice(length), out var headerLength))
                 {
+                    if (length == 0 && throwIfNoneEncoded)
+                    {
+                        throw new ArgumentOutOfRangeException("The given buffer was too small to encode any headers.");
```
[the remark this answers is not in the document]
Tratcher: Yes it does surface to the application on first write. I added it because I hung multiple tests while refactoring the max frame size handling. Without it this code spins in a loop.

I'm not as concerned about the application running into this. The client isn't allowed to reduce the max frame size bellow 16kb, and if your individual headers are over 16kb then you have other issues. I will change the type to InvalidOperationException.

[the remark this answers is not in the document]
halter73: How is it logged if the response doesn't contain a body?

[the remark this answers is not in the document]
Tratcher: It's logged as a general request error. I've added a little more error handling to abort the connection since the header compression table would now be considered corrupt. There's not a good way to run the normal Http2Connection cleanup code from here. I'm not concerned with making this code path more graceful at the moment, we'll see if users really run into this scenario.

src/Kestrel.Core/Internal/Http2/Http2Connection.cs
```diff
@@ -422,12 +426,6 @@ private Task ProcessDataFrameAsync()
                 throw new Http2ConnectionErrorException(CoreStrings.FormatHttp2ErrorStreamIdZero(_incomingFrame.Type), Http2ErrorCode.PROTOCOL_ERROR);
             }
 
-            // The Padding field is missing
-            if (_incomingFrame.DataPayloadOffset > _incomingFrame.PayloadLength)
-            {
-                throw new Http2ConnectionErrorException(CoreStrings.Http2FrameMissingFields, Http2ErrorCode.PROTOCOL_ERROR);
```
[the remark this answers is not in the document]
Tratcher: The frame reader checks it while it's reading those extra fields: https://github.com/aspnet/KestrelHttpServer/pull/2893/files#diff-feb5d4f5a5878a3a2366ddb77ef8d020R79

src/Kestrel.Core/Internal/Http2/Http2Connection.cs
Tratcher: Added

Tratcher:
Rebased and updated.

src/Kestrel.Core/Internal/Http2/HPack/HPackEncoder.cs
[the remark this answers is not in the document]
Tratcher: Filed https://github.com/aspnet/KestrelHttpServer/issues/2926 to address this by splitting header values acrross frames.

src/Kestrel.Core/Internal/Http2/HPack/HPackEncoder.cs
```diff
@@ -24,20 +24,29 @@ public bool BeginEncode(int statusCode, IEnumerable<KeyValuePair<string, string>
...
         {
             length = 0;
 
             do
             {
                 if (!EncodeHeader(_enumerator.Current.Key, _enumerator.Current.Value, buffer.Slice(length), out var headerLength))
                 {
+                    if (length == 0 && throwIfNoneEncoded)
+                    {
+                        throw new HPackEncodingException(CoreStrings.HPackErrorNotEnoughBuffer);
```
halter73: I know we don't plan for this to be permanent, but I would still like a test verifying this gets thrown from the first body write.

Tratcher: Added

davidfowl:
👍

Issue #2903, opened 2018-09-07
Map ListenOptions.Protocols from IConfiguration


shirhatti:
To allow enabling or disabling HTTP/2

PR #2908, opened 2018-09-10
Map ListenOptions.Protocols from IConfiguration

Tratcher:
#2903 @shirhatti @DamianEdwards requested this after a high level design review. This allows you to turn HTTP/2 (and HTTP/1.x) on/off server wide or per endpoint using only config.

New schemas:

Per endpoint there is a new `Protocols` field:
```
{
"Kestrel": {
  "EndPoints": {
    "Endpoint0": {
      "Url": "http://localhost:5000",
      "Protocols": "Http1AndHttp2"
    },
```

Globally there is a new `EndPointDefaults` section with a `Protocols` field:
```
{
"Kestrel": {
  "EndPointDefaults": {
      "Protocols": "Http1AndHttp2"
```
As an alternative we could collapse this to:
```
{
"Kestrel": {
      "DefaultProtocols": "Http1AndHttp2"
```

The value is any acceptable value of the HttpProtocols enum: Http1, Http2, or Http1AndHttp2.

The global default really is global, it will affect endpoints defined via any of the currently acceptable mechanics: The default http://localhost:5000, UseUrls, ASPNETCORE_URLS, Listen, Config, etc.).

Protocols specified in code will override the config values, including in ConfigureEndpointDefaults.

### DIFF
--- a/src/Kestrel.Core/Internal/ConfigurationReader.cs
+++ b/src/Kestrel.Core/Internal/ConfigurationReader.cs
@@ -9,9 +9,17 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal
 {
     internal class ConfigurationReader
     {
+        private const string ProtocolsKey = "Protocols";
+        private const string CertificatesKey = "Certificates";
+        private const string CertificateKey = "Certificate";
+        private const string EndpointDefaultsKey = "EndpointDefaults";
+        private const string EndpointsKey = "Endpoints";
+        private const string UrlKey = "Url";
+
         private IConfiguration _configuration;
         private IDictionary<string, CertificateConfig> _certificates;
         private IList<EndpointConfig> _endpoints;
+        private EndpointDefaults _endpointDefaults;
 
         public ConfigurationReader(IConfiguration configuration)
         {
@@ -28,6 +36,19 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal
                 }
 
                 return _certificates;
+            }
+        }
+
+        public EndpointDefaults EndpointDefaults
+        {
+            get
+            {
+                if (_endpointDefaults == null)
+                {
+                    ReadEndpointDefaults();
+                }
+
+                return _endpointDefaults;
             }
         }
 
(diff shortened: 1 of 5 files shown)


src/Kestrel.Core/Internal/ConfigurationReader.cs
```diff
@@ -80,6 +109,7 @@ private void ReadEndpoints()
                 {
                     Name = endpointConfig.Key,
                     Url = url,
+                    Protocols = endpointConfig[ProtocolsKey],
                     ConfigSection = endpointConfig,
                     Certificate = new CertificateConfig(endpointConfig.GetSection("Certificate")),
```
JunTaoLuo: Can we make "Certificate" and "Url" a private const string too

src/Kestrel.Core/Internal/ConfigurationReader.cs
```diff
@@ -80,6 +109,7 @@ private void ReadEndpoints()
                 {
                     Name = endpointConfig.Key,
                     Url = url,
+                    Protocols = endpointConfig[ProtocolsKey],
```
JunTaoLuo: update the comments above with the new endpoint config layout

src/Kestrel.Core/Internal/ConfigurationReader.cs
```diff
@@ -55,6 +71,19 @@ private void ReadCertificates()
...
         }
 
+        // "EndpointDefaults": {
+        //    "Protocols": "Http1AndHttp2",
+        // }
+        private void ReadEndpointDefaults()
+        {
+            _endpointDefaults = new EndpointDefaults()
+            {
+                ConfigSection = _configuration.GetSection("EndpointDefaults"),
```
JunTaoLuo: Why are we saving the config section?

src/Kestrel.Core/KestrelConfigurationLoader.cs
```diff
@@ -197,24 +201,41 @@ public KestrelConfigurationLoader HandleEndpoint(ulong handle, Action<ListenOpti
             return this;
         }
 
+        // Called from ApplyEndpointDefaults so it applies to even explicit Listen endpoints.
+        // Does not require a call to Load.
+        internal void ApplyConfigurationDefaults(ListenOptions listenOptions)
+        {
+            var defaults = ConfigurationReader.EndpointDefaults;
+
+            var protocols = ParseProtocols(defaults.Protocols);
```
JunTaoLuo: So the parsing occurs on loading each listenoption? Why not parse out the protocol in the ConfigurationReader in ReadEndpointDefaults and just do it once?

src/Kestrel.Core/KestrelServerOptions.cs
```diff
@@ -103,6 +103,7 @@ public void ConfigureEndpointDefaults(Action<ListenOptions> configureOptions)
         internal void ApplyEndpointDefaults(ListenOptions listenOptions)
         {
             listenOptions.KestrelServerOptions = this;
+            ConfigurationLoader?.ApplyConfigurationDefaults(listenOptions);
```
JunTaoLuo: When is ConfigurationLoader null?

Tratcher: It's null by default. It's only set by CreateDefaultBuilder:
https://github.com/aspnet/MetaPackages/blob/5605eabee6be4790f2fd22b27a2eb9c92070be23/src/Microsoft.AspNetCore/WebHost.cs#L163

Tratcher:
Updated

Issue #2750, opened 2018-07-25
Flaky test: WritingToConnectionAfterUnobservedCloseTriggersRequestAbortedToken


ryanbrandenburg:
This test [fails](http://aspnetci/viewLog.html?buildId=514027&buildTypeId=Releases_21Public_UbuntuUniverse) occasionally with the following error:
```
Moq.MockException : 
Expected invocation on the mock once, but was 0 times: t => t.ConnectionStop(It.IsAny<String>())
No setups configured.
Performed invocations:
KestrelTrace.IsEnabled(Critical)
KestrelTrace.BeginScope<ConnectionLogScope>([[ConnectionId, 0HLFHOOITD5M5]])
KestrelTrace.ConnectionStart("0HLFHOOITD5M5")
KestrelTrace.ConnectionDisconnect("0HLFHOOITD5M5")
KestrelTrace.ApplicationError("0HLFHOOITD5M5", "0HLFHOOITD5M5:00000001", System.OperationCanceledException: The operation was canceled.
   at System.Threading.CancellationToken.ThrowOperationCanceledException()
   at Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http.HttpResponseStream.ValidateState(CancellationToken cancellationToken) in /_/src/Kestrel.Core/Internal/Http/HttpResponseStream.cs:line 168
   at Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http.HttpResponseStream.WriteAsync(Byte[] buffer, Int32 offset, Int32 count, CancellationToken cancellationToken) in /_/src/Kestrel.Core/Internal/Http/HttpResponseStream.cs:line 111
   at Microsoft.AspNetCore.Server.Kestrel.FunctionalTests.ResponseTests.<>c__DisplayClass61_0.<<WritingToConnectionAfterUnobservedCloseTriggersRequestAbortedToken>b__6>d.MoveNext() in /_/test/Kestrel.FunctionalTests/ResponseTests.cs:line 2398
--- End of stack trace from previous location where exception was thrown ---
   at Xunit.Assert.RecordExceptionAsync(Func`1 testCode) in C:\Dev\xunit\xunit\src\xunit.assert\Asserts\Record.cs:line 82)
   at Moq.Mock.ThrowVerifyException(MethodCall expected, IEnumerable`1 setups, IEnumerable`1 actualCalls, Expression expression, Times times, Int32 callCount)
[...]

PR #2915, opened 2018-09-11
Relax connection stop checks in tests to reduce flakiness

JunTaoLuo:
Addresses https://github.com/aspnet/KestrelHttpServer/issues/2750. This will reduce test coverage and I'm modifying similar tests preemptively since they could probably fail in the same pattern.

### DIFF
--- a/test/Kestrel.FunctionalTests/RequestTests.cs
+++ b/test/Kestrel.FunctionalTests/RequestTests.cs
@@ -1302,7 +1302,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.FunctionalTests
                 await appFuncCompleted.Task.DefaultTimeout();
             }
 
-            mockKestrelTrace.Verify(t => t.ConnectionStop(It.IsAny<string>()), Times.Once());
+            mockKestrelTrace.Verify(t => t.ConnectionStop(It.IsAny<string>()), Times.AtMostOnce());
         }
 
         [Theory]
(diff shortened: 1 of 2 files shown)


test/Kestrel.FunctionalTests/RequestTests.cs
```diff
@@ -1302,7 +1302,7 @@ public async Task ServerCanAbortConnectionAfterUnobservedClose(ListenOptions lis
                 await appFuncCompleted.Task.DefaultTimeout();
             }
 
-            mockKestrelTrace.Verify(t => t.ConnectionStop(It.IsAny<string>()), Times.Once());
+            mockKestrelTrace.Verify(t => t.ConnectionStop(It.IsAny<string>()), Times.AtMostOnce());
```
JunTaoLuo: Is there a point in checking these at all if we allow 0 and 1? Maybe we should just remove them entirely. I suppose we are making sure ConnectionStop is never called more than once still.

Tratcher:
Background please. What is the problem and how does this fix it?

This is 2.1 only and won't be merged to 2.2 or master, correct?

JunTaoLuo:
Most of the information is in the issue. There are test failures in 2.1 which were fixed in 2.2 and master via changes in #2646. However, @Eilon has asked to fix the tests without back-porting src changes since product changes for 2.1 are unlikely to be approved by shiproom unless it was reported by customers.

Eilon:
Yup.

test/Kestrel.FunctionalTests/RequestTests.cs
halter73: There's almost no point other than to remind us that there was once a useful assertion here. Be extra careful not to apply these changes to release/2.2 or master when we merge.

Issue #2885, opened 2018-09-06
Test failure: AppCanHandleClientAbortingConnectionMidResponse


aspnet-hello:
This test [fails](http://aspnetci/viewLog.html?buildId=544982&buildTypeId=Releases_21Public_OsxUniverse) occasionally with the following error:
```
Assert.Empty() Failure
Collection: [WriteContext { EventId = ConnectionError, Exception = System.IO.IOException: Protocol wrong type for socket ---> System.Net.Sockets.SocketException: Protocol wrong type for socket
   at Microsoft.AspNetCore.Server.Kestrel.Transport.Sockets.Internal.SocketAwaitable.GetResult() in /_/src/Kestrel.Transport.Sockets/Internal/SocketAwaitable.cs:line 40
   at Microsoft.AspNetCore.Server.Kestrel.Transport.Sockets.Internal.SocketConnection.ProcessSends() in /_/src/Kestrel.Transport.Sockets/Internal/SocketConnection.cs:line 258
   at Microsoft.AspNetCore.Server.Kestrel.Transport.Sockets.Internal.SocketConnection.DoSend() in /_/src/Kestrel.Transport.Sockets/Internal/SocketConnection.cs:line 201
   --- End of inner exception stack trace ---, Formatter = Func`3 { ... }, LoggerName = "Microsoft.AspNetCore.Server.Kestrel.Transport.Sock"..., LogLevel = Information, ... }]
   at Microsoft.AspNetCore.Server.Kestrel.FunctionalTests.ResponseTests.AppCanHandleClientAbortingConnectionMidResponse(ListenOptions listenOptions) in /_/test/Kestrel.FunctionalTests/ResponseTests.cs:line 2505
--- End of stack trace from previous location where exception was thrown ---
```
Other tests within that build may have failed with a similar message, but they are not listed here. Check the link above for more info.

This test failed on 2.1.

CC @muratg

This issue was made automatically. If there is a problem contact ryanbrandenburg.

aspnet-hello:
Please use this workflow to address this flaky test issue, including checking applicable checkboxes and filling in the applicable "TODO" entries:

* Is this actually a flaky test?
  * No, this is a regular test failure, fix the test/product (TODO: Link to commit/PR)
  * Yes, proceed below...

* Is this test failure caused by product code flakiness? (Either this product, or another product this test depends on.)
  * [ ] File a bug against the product (TODO: Link to other bug)
[...]

PR #2916, opened 2018-09-11
Handle SocketError.ProtocolType as a connection reset on macOS

halter73:
Backport of #2845

Addresses #2885


Tratcher:
Didn't need this bit? https://github.com/aspnet/KestrelHttpServer/pull/2845/files#diff-a538265cafcdc8078989d06b998e6fb9R342

halter73:
Since is a 2.1 backport, and src changes are difficult to get approval for to begin with, I'm trying to make this as surgical as possible.

halter73:
@muratg @Eilon What do I need to do to get this approved for 2.1? It's a src change, but a very safe/minor one.

See https://github.com/aspnet/KestrelHttpServer/issues/2885#issuecomment-420431891 for context

Eilon:
What @muratg said.

halter73:
I agree that changing Sockets is a better change. I think the only way to avoid test failures in 2.1 without a product change such as this is to skip all of our connection reset functional tests on macOS in release/2.1. Should we do that instead?

muratg:
@halter73 yeah, I think that's the best we can do. We'd get coverage from Windows and Linux runs. And we _know_ the reason for the MacOS failures so let's just skip those.

Issue #2878, opened 2018-09-02
I broke the http/2 , preview 2.2


rene98c:
Where do i start debugging this?

My app was working great for 2 weeks on http2 but today i broke it.
It has to do something with browser, how it loads static resources, css and javascript. I optimized the static content delivery and finally figured that http2 is to blame.
It started to lock up with pending status at chrome browser while i  was refreshing the page few  times (ctrl - f5) . kestrel did not show any errors and database log was ok.
Checked the server resource usage, cpu memory  all looked fine like before, no change there.
Not a browser issue too, chrome, edge, firefox behaved the same, just waiting for the first byte.
And now it fails to load the page almost every time. just a minute ago i noticed app finally responded after 10minutes.
The app did load almost instantly and was fast before this issue started to occur.
I changed the kestrel to use only http1 and it started working fine on every server i deployed it. 
I have app running on 3 different virtual machines (debian 9.5), developing on windows and it locked up the same.

I'll post some updates as soon i find anything else.

halter73:
There's also #2874 which breaks HTTP/2 in Chrome specifically in preview 2.

rene98c:
Sure, repo here and issue is reproducible: https://github.com/rene98c/http2debug
http2debug
environment=Stagging if i set kestrel to use http2 (default):

i can access the image from app fine without locking (https://somesite/Content/img/app.png)
as soon as i open the page (https://somesite/login) it locks up, not always though. and if it works few time hitting ctrl+f5 locks it up eventually. environment=Development
Happens locally also but not that easily , i have to agressively hit ctrl+f5 to produce the same result
as soon as i add this line: o.ConfigureEndpointDefaults(x=>x.Protocols = HttpProtocols.Http1); everything works.

i have the app running 
for the demonstration purposes, this image https://bdec.ee/Content/img/app.png from the site is accessible.
If someone has opened the page here https://bdec.ee/login and locked the application the /img/app.png wont open either anymore.

rene98c:
@halter73 
did some debugging and located the cause that breaks http2 .

I found that css base64 encoded content is what causes the app to freeze up.
There are 2 base64 encoded resources in bundle.min.css.
I removed those lines and app seemed to run fine with http2.

background-image: url(data:image/png;base64, ...
src: url(data:application/font-woff2;charset=utf-8;base64 ...

Tratcher:
[...]

PR #2921, opened 2018-09-12
Dispatch Http/2 write callbacks

Tratcher:
See https://github.com/aspnet/KestrelHttpServer/issues/2878#issuecomment-420773020

### DIFF
--- a/src/Kestrel.Core/Internal/Http2/Http2OutputProducer.cs
+++ b/src/Kestrel.Core/Internal/Http2/Http2OutputProducer.cs
@@ -225,7 +225,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http2
             (
                 pool: pool,
                 readerScheduler: PipeScheduler.Inline,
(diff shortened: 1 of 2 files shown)


Tratcher:
@rene98c

halter73:
@benaadams I was also worried about the performance of this change initially, but I think this is OK since this only dispatches when relieving backpressure induced by the either the main connection-level Pipe or HTTP/2 flow control.

Since the reader scheduler is still inline, so long as the Http2OutputProducer.ProcessDataWrites() calls PipeReader.AdvanceTo() inline without awaiting an actual async completion of Http2FrameWriter.WriteDataAsync(), which is most of the time, a completed Task will be returned from HttpResponseStream.Write(). There's should be no performance penalty to using the ThreadPool PipeScheduler, which I don't think can be overridden, unless there's backpressure.

Also, the suggested workaround doesn't fix the bug since it's the WriteAsync prior to the blocking write that needs to dispatch to unblock the ProcessDataWrites() loop.

benaadams:
Fair 😄

PR #2923, opened 2018-09-13
Enable hpack/4.2 - Maximum Table Size test

JunTaoLuo:
We needed to ensure dynamic tables size updates occur at the beginning of the header block

Partially addresses https://github.com/aspnet/KestrelHttpServer/issues/2154.

### DIFF
--- a/src/Kestrel.Core/Internal/Http2/HPack/HPackDecoder.cs
+++ b/src/Kestrel.Core/Internal/Http2/HPack/HPackDecoder.cs
@@ -94,6 +94,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http2.HPack
         private int _headerValueLength;
         private bool _index;
         private bool _huffman;
+        private bool _headersObserved;
 
         public HPackDecoder(int maxDynamicTableSize)
             : this(maxDynamicTableSize, new DynamicTable(maxDynamicTableSize))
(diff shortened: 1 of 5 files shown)


src/Kestrel.Core/Internal/Http2/HPack/HPackDecoder.cs
```diff
@@ -110,9 +110,20 @@ internal HPackDecoder(int maxDynamicTableSize, DynamicTable dynamicTable)
 
         public void Decode(Span<byte> data, bool endHeaders, IHttpHeadersHandler handler)
         {
+            var headersObserved = false;
```
Tratcher: This isn't the right place, the state needs to be preserved across calls to Decode so it can handle Continuation frames.

src/Kestrel.Core/Internal/Http2/HPack/HPackDecoder.cs
```diff
@@ -197,8 +211,14 @@ public void OnByte(byte b, IHttpHeadersHandler handler)
                     {
```
Tratcher: I think the new error check should go here rather than in the Decode method. headersObserved needs to become a member field that's reset at the end of Decode if endHeaders is set. That way headersObserved persists across calls to Decode (e.g. Continuation frames).

JunTaoLuo: I see.

src/Kestrel.Core/Internal/Http2/HPack/HPackDecoder.cs
```diff
@@ -133,6 +139,7 @@ public void OnByte(byte b, IHttpHeadersHandler handler)
                         if (_integerDecoder.BeginDecode((byte)val, IndexedHeaderFieldPrefix))
                         {
                             OnIndexedHeaderField(_integerDecoder.Value, handler);
+                            _headersObserved = true;
```
Tratcher: You can consolidate all of this into the Ready state if you move this assignment above this `if` and copy it to the other header types (LiteralHeaderFieldWithIncrementalIndexingRepresentation, etc.). Then it can be enforced once in DynamicTableSizeUpdateRepresentation.

JunTaoLuo:
:up::date:

src/Kestrel.Core/Internal/Http2/HPack/HPackDecoder.cs
JunTaoLuo: Consolidated

src/Kestrel.Core/Internal/Http2/HPack/HPackDecoder.cs
```diff
@@ -402,5 +405,25 @@ private HeaderField GetHeader(int index)
...
+            {
+                throw new HPackDecodingException(
+                    CoreStrings.FormatHPackErrorDynamicTableSizeUpdateTooLarge(size, _maxDynamicTableSize));
+            }
+
+            // https://tools.ietf.org/html/rfc7541#section-4.2
+            // This dynamic table size
+            // update MUST occur at the beginning of the first header block
+            // following the change to the dynamic table size.
+            if (_headersObserved)
```
Tratcher: Want to move this to line 207 so it all stays consolidated in Ready?

PR #2924, opened 2018-09-13
WIP: Use the ArrayPool

davidfowl:
Use the ArrayPool instead of using the SlabMemoryPool. There are pros and cons:

Pros:
- The pool shrinks under memory pressure
- More than 4K sized buffers (This isn't using that right now)
- We'll probably get future advancements for free (possibly this https://github.com/dotnet/corefx/issues/31787)

Cons:
- Allocates per call (the object wrapping the array)
- No pinning, we'll end up paying the cost of pinning per operation (see https://github.com/dotnet/coreclr/issues/19936)

I need to run some performance tests.


muratg:
I like this. In addition to understanding perf impact, I'd also love to see memory usage shrink in a test too.

benaadams:
>  I'd also love to see memory usage shrink in a test too

Something doing large responses would probably be good for this? (And run GC before final measure; to find how much it will idle down to after being pushed high)

PR #2931, opened 2018-09-15
Remove calls to MemoryMarshal.GetReference

davidfowl:
- Spans can be pinned natively in C# 7.3

cc @jkotas @jaredpar (This is the moral equivalent right?)


davidfowl:

benaadams:
The slight behaviour difference is without `MemoryMarshal.GetReference` it will return a `null` pointer when `.Length` is zero; meaning if you did use the pointer returned you'd either get a null ref or access violation; which is probably safer than writing out of bounds. Though it only applies to zero lengths specifically.

Issue #2929, opened 2018-09-14
Category of some loggers are wrong - Name of Type instead Namespace+Type


tiago-soczek:
Hi,

While I was filtering log entries by category/source, I found these 2 loggers using just the Type name for the category of logging:

----

https://github.com/aspnet/KestrelHttpServer/blob/release/2.2/src/Kestrel.Core/Internal/HttpsConnectionAdapter.cs#L64

----

https://github.com/aspnet/KestrelHttpServer/blob/6fde01a825cffc09998d3f8a49464f7fbe40f9c4/src/Kestrel.Core/Adapter/ListenOptionsConnectionLoggingExtensions.cs#L33

----

In this case, would be better to use the full type name (Namespace+Type), right?

If needed, I'm available to work on this issue.

Thanks.

Tratcher:
You're right, those should have been `.CreateLogger<HttpsConnectionAdapter>()`. A PR would be welcome.

Tratcher:
Or like this:
https://github.com/aspnet/KestrelHttpServer/blob/6fde01a825cffc09998d3f8a49464f7fbe40f9c4/src/Kestrel.Transport.Sockets/SocketTransportFactory.cs#L41

Watch out for the tests: 
https://github.com/aspnet/KestrelHttpServer/blob/f70ba53253480a0c9b672537b8b13058b3e16160/test/Kestrel.InMemory.FunctionalTests/HttpsTests.cs#L386

tiago-soczek:
OK! I'll take care of this issue, please assign to me.

Tratcher:
Note: use the release/2.2 branch.

halter73:
Breaking change announcement: aspnet/Announcements#317

PR #2935, opened 2018-09-16
Fix category logger for HttpsConnectionAdapter and LoggingConnectionAdapter

tiago-soczek:
Addresses: #2929.

### DIFF
--- a/test/Kestrel.InMemory.FunctionalTests/HttpsTests.cs
+++ b/test/Kestrel.InMemory.FunctionalTests/HttpsTests.cs
@@ -385,7 +385,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.InMemory.FunctionalTests
 
             public ILogger CreateLogger(string categoryName)
             {
-                if (categoryName == nameof(HttpsConnectionAdapter))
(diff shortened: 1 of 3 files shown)


test/Kestrel.InMemory.FunctionalTests/HttpsTests.cs
```diff
@@ -385,7 +385,7 @@ private class HandshakeErrorLoggerProvider : ILoggerProvider
 
             public ILogger CreateLogger(string categoryName)
             {
-                if (categoryName == nameof(HttpsConnectionAdapter))
+                if (categoryName == TypeNameHelper.GetTypeDisplayName(typeof(HttpsConnectionAdapter)))
```
tiago-soczek: I preferred to use the `TypeNameHelper.GetTypeDisplayName` because it is used by the `LoggerFactory`. But `typeof(HttpsConnectionAdapter).FullName` would suffice for this case.

halter73:
@muratg Should we make a breaking change announcement for a logger namespace change?

muratg:
@halter73 good point! Could you write something up?

cc @Tratcher

Issue #2789, opened 2018-08-08
Accept Socket handle is inherited by child process


ppekrol:
Hi,

I've encountered an error with Kestrel. It appears that disposing the 'WebHost' in certain case is not stopping the server from listening and is causing incoming connections to hang.

I've created a repro for this:
https://github.com/ppekrol/KestrelIssue

Flow 1 (works):
```
dotnet restore
dotnet build
cd .\KestrelIssue\
dotnet run

Started server at: http://127.0.0.1:57741
Starting first request...
Finished first request...
No pingers!
Disposing server...
Disposed server...
NETSTAT:
  TCP    127.0.0.1:57741        127.0.0.1:57742        FIN_WAIT_2      22288
  TCP    127.0.0.1:57742        127.0.0.1:57741        CLOSE_WAIT      22288

Starting second request...

Unhandled Exception: System.Net.Http.HttpRequestException: No connection could be made because the target machine actively refused it ---> System.Net.Sockets.SocketException: No connection could be made because the target machine actively refused it
   at System.Net.Http.ConnectHelper.ConnectAsync(String host, Int32 port, CancellationToken cancellationToken)
   --- End of inner exception stack trace ---
   at System.Net.Http.ConnectHelper.ConnectAsync(String host, Int32 port, CancellationToken cancellationToken)
   at System.Threading.Tasks.ValueTask`1.get_Result()
   at System.Net.Http.HttpConnectionPool.CreateConnectionAsync(HttpRequestMessage request, CancellationToken cancellationToken)
   at System.Threading.Tasks.ValueTask`1.get_Result()
   at System.Net.Http.HttpConnectionPool.WaitForCreatedConnectionAsync(ValueTask`1 creationTask)
   at System.Threading.Tasks.ValueTask`1.get_Result()
   at System.Net.Http.HttpConnectionPool.SendWithRetryAsync(HttpRequestMessage request, Boolean doRequestAuth, CancellationToken cancellationToken)
   at System.Net.Http.RedirectHandler.SendAsync(HttpRequestMessage request, CancellationToken cancellationToken)
   at System.Net.Http.HttpClient.FinishSendAsyncBuffered(Task`1 sendTask, HttpRequestMessage request, CancellationTokenSource cts, Boolean disposeCts)
   at KestreLIssue.Program.Main(String[] args) in C:\workspaces\Temp\KestrelIssue\KestrelIssue\Program.cs:line 88
[...]

PR #2944, opened 2018-09-18
Do not inherit socket handles

Tratcher:
#2789 If the accept socket's handle gets inherited by child processes, then it's not fully closed until all of the child processes are closed.

Connected sockets don't seem to be affected.

### DIFF
--- a/src/Kestrel.Transport.Sockets/Internal/NativeMethods.cs
+++ b/src/Kestrel.Transport.Sockets/Internal/NativeMethods.cs
@@ -1,0 +1,31 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Net.Sockets;
+using System.Runtime.InteropServices;
+
+namespace Microsoft.AspNetCore.Server.Kestrel.Transport.Sockets.Internal
+{
+    internal static class NativeMethods
+    {
+        [DllImport("kernel32.dll", SetLastError = true)]
+        private static extern bool SetHandleInformation(IntPtr hObject, HANDLE_FLAGS dwMask, HANDLE_FLAGS dwFlags);
+
+        [Flags]
+        private enum HANDLE_FLAGS : uint
+        {
+            None = 0,
+            INHERIT = 1,
+            PROTECT_FROM_CLOSE = 2
+        }
+
+        internal static void DisableHandleInheritance(Socket socket)
+        {
+            if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
+            {
+                SetHandleInformation(socket.Handle, HANDLE_FLAGS.INHERIT, 0);
+            }
+        }
+    }
+}
--- a/src/Kestrel.Transport.Sockets/SocketTransport.cs
+++ b/src/Kestrel.Transport.Sockets/SocketTransport.cs
@@ -80,6 +80,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Transport.Sockets
             IPEndPoint endPoint = _endPointInformation.IPEndPoint;
 
             var listenSocket = new Socket(endPoint.AddressFamily, SocketType.Stream, ProtocolType.Tcp);
+            NativeMethods.DisableHandleInheritance(listenSocket);
 
             // Kestrel expects IPv6Any to bind to both IPv6 and IPv4
             if (endPoint.Address == IPAddress.IPv6Any)
--- a/test/Kestrel.Transport.FunctionalTests/HandleInheritanceTests.cs
+++ b/test/Kestrel.Transport.FunctionalTests/HandleInheritanceTests.cs
@@ -1,0 +1,60 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Diagnostics;
+using System.Net.Sockets;
+using System.Runtime.InteropServices;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Builder;
+using Microsoft.AspNetCore.Hosting;
+using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Testing;
+using Microsoft.AspNetCore.Testing.xunit;
+using Xunit;
+
+namespace Microsoft.AspNetCore.Server.Kestrel.FunctionalTests
+{
+    public class HandleInheritanceTests : TestApplicationErrorLoggerLoggedTest
+    {
+        [ConditionalFact]
+        [OSSkipCondition(OperatingSystems.MacOSX, SkipReason = "No fix available for Mac https://github.com/aspnet/KestrelHttpServer/pull/2944#issuecomment-426397600")]
+        public async Task SpawnChildProcess_DoesNotInheritListenHandle()
+        {
+            var hostBuilder = TransportSelector.GetWebHostBuilder()
+                .UseKestrel()
+                .ConfigureServices(AddTestLogging)
+                .UseUrls("http://127.0.0.1:0")
+                .Configure(app =>
+                {
+                    app.Run(context =>
+                    {
+                        return context.Response.WriteAsync("Hello World");
+                    });
+                });
+
+            using (var host = hostBuilder.Build())
+            {
+                await host.StartAsync();
+
+                var processInfo = new ProcessStartInfo
+                {
+                    FileName = RuntimeInformation.IsOSPlatform(OSPlatform.Windows) ? "cmd.exe" : "vi",
+                    CreateNoWindow = true,
+                };
+                using (var process = Process.Start(processInfo))
+                {
+                    var port = host.GetPort();
+                    await host.StopAsync();
+
+                    // We should not be able to connect if the handle was correctly closed and not inherited by the child process.
+                    using (var client = new TcpClient())
+                    {
+                        await Assert.ThrowsAnyAsync<SocketException>(() => client.ConnectAsync("127.0.0.1", port));
+                    }
+
+                    process.Kill();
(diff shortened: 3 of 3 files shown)


test/Kestrel.Transport.BindTests/HandleInheritanceTest.cs
```diff
@@ -0,0 +1,94 @@
...
+                    });
+                });
+
+            using (var host = hostBuilder.Build())
+            {
+                await host.StartAsync();
+
+                var processInfo = new ProcessStartInfo
+                {
+                    FileName = RuntimeInformation.IsOSPlatform(OSPlatform.Windows) ? "cmd.exe" : "bash",
```
halter73: Might as well using sh instead of bash since the former is required by POSIX even if it happens to be bash.

Tratcher:

Tratcher:
@dotnet-bot test OSX 10.12 Debug x64 Build
@dotnet-bot test OSX 10.12 Release x64 Build

test/Kestrel.Transport.BindTests/HandleInheritanceTest.cs
```diff
@@ -0,0 +1,105 @@
...
+                {
+                    var port = host.GetPort();
+                    await host.StopAsync();
+
+                    using (var client = new TcpClient())
+                    {
+                        await Assert.ThrowsAnyAsync<SocketException>(() => client.ConnectAsync("127.0.0.1", port));
+                    }
+
+                    process.Kill();
```
JunTaoLuo:
```
08:42:31   Error Message:
08:42:31    System.ComponentModel.Win32Exception : No such process
08:42:31   Stack Trace:
08:42:31      at System.Diagnostics.Process.Kill()
08:42:31      at Microsoft.AspNetCore.Server.Kestrel.FunctionalTests.HandleInheritanceTest.SpawnChildProcess_DoesNotInheritListenHandle() in /_/test/Kestrel.Transport.BindTests/HandleInheritanceTest.cs:line 53
08:42:31   --- End of stack trace from previous location where exception was thrown ---
08:42:31   Standard Output Messages:
08:42:31    | [0.010s] TestLifetime Information: Starting test SpawnChildProcess_DoesNotInheritListenHandle at 2018-09-20T15:42:30
08:42:31    | [0.075s] Microsoft.AspNetCore.Hosting.Internal.WebHost Debug: Hosting starting
08:42:31    | [0.173s] Microsoft.AspNetCore.Hosting.Internal.WebHost Debug: Hosting started
08:42:31    | [0.173s] Microsoft.AspNetCore.Hosting.Internal.WebHost Debug: Loaded hosting startup assembly Libuv.BindTests
08:42:31    | [0.184s] Microsoft.AspNetCore.Hosting.Internal.WebHost Debug: Hosting shutdown
08:42:31    | [0.238s] TestLifetime Information: Finished test SpawnChildProcess_DoesNotInheritListenHandle in 0.2340523s
```

We need something like this:
https://github.com/aspnet/DotNetTools/blob/60041b5b2bcb84a9a6b9eb460f6e395a270934c8/test/dotnet-watch.FunctionalTests/AwaitableProcess.cs#L146-L157

JunTaoLuo: Or maybe this would work? https://github.com/aspnet/Common/blob/e0f40110ff5e2101231ba54d48961bbd60a07727/shared/Microsoft.Extensions.Process.Sources/ProcessHelper.cs

JunTaoLuo: On my local run, I see a different error:
```
Failed   SpawnChildProcess_DoesNotInheritListenHandle
Error Message:
 Assert.Throws() Failure
Expected: typeof(System.Net.Sockets.SocketException)
Actual:   (No exception was thrown)
Stack Trace:
   at Microsoft.AspNetCore.Server.Kestrel.FunctionalTests.HandleInheritanceTest.SpawnChildProcess_DoesNotInheritListenHandle() in /Users/jtluo/Documents/workspace/juntaoluo/KestrelHttpServer/test/Kestrel.Transport.BindTests/HandleInheritanceTest.cs:line 50
Standard Output Messages:
 | [0.016s] TestLifetime Information: Starting test SpawnChildProcess_DoesNotInheritListenHandle at 2018-09-21T03:01:25
 | [0.085s] Microsoft.AspNetCore.Hosting.Internal.WebHost Debug: Hosting starting
 | [0.168s] Microsoft.AspNetCore.Hosting.Internal.WebHost Debug: Hosting started
 | [0.168s] Microsoft.AspNetCore.Hosting.Internal.WebHost Debug: Loaded hosting startup assembly Sockets.BindTests
 | [0.191s] Microsoft.AspNetCore.Hosting.Internal.WebHost Debug: Hosting shutdown
 | [0.241s] TestLifetime Information: Finished test SpawnChildProcess_DoesNotInheritListenHandle in 0.2341235s
```
I should take a closer look at what's actually happening in this test.

Tratcher:

test/Kestrel.Transport.BindTests/HandleInheritanceTest.cs
```diff
@@ -0,0 +1,109 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
```
halter73: This shouldn't go in BindTests. In order to avoid conflicts, I don't want any tests running in parallel to the AddressRegistrationTests even if they bind to port 0.

test/Kestrel.Transport.BindTests/HandleInheritanceTest.cs
```diff
@@ -0,0 +1,109 @@
...
+using Microsoft.AspNetCore.Builder;
+using Microsoft.AspNetCore.Hosting;
+using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Testing;
+using Microsoft.AspNetCore.Testing.xunit;
+using Xunit;
+
+namespace Microsoft.AspNetCore.Server.Kestrel.FunctionalTests
+{
+    [OSSkipCondition(OperatingSystems.MacOSX)] // Not applicable, and vi doesn't work there.
```
halter73: Are we sure handle inheritance isn't an issue on macOS?

test/Kestrel.Transport.BindTests/HandleInheritanceTest.cs
```diff
@@ -0,0 +1,109 @@
...
+    public class HandleInheritanceTest : TestApplicationErrorLoggerLoggedTest
+    {
+        [ConditionalFact]
+        public async Task SpawnChildProcess_DoesNotInheritListenHandle()
```
halter73: I tried running this test on my Windows 10 machine and it passed without your changes.

test/Kestrel.Transport.BindTests/HandleInheritanceTest.cs
Tratcher: Yes, the corefx issue was clear that this is only an issue for Windows.

test/Kestrel.Transport.BindTests/HandleInheritanceTest.cs
```diff
@@ -0,0 +1,109 @@
...
+using Microsoft.AspNetCore.Hosting;
+using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Testing;
+using Microsoft.AspNetCore.Testing.xunit;
+using Xunit;
+
+namespace Microsoft.AspNetCore.Server.Kestrel.FunctionalTests
+{
+    [OSSkipCondition(OperatingSystems.MacOSX)] // Not applicable, and vi doesn't work there.
+    public class HandleInheritanceTest : TestApplicationErrorLoggerLoggedTest
```
halter73: Nit: HandleInheritanceTest**s**

Tratcher:
Hmm, the Listen test fails on Mac sockets, and the Connection test fails on Windows Libuv.

Tratcher:
Ok, I tracked down the Mac issue.

The linux equivilent of what we did on windows is https://legacy.python.org/dev/peps/pep-0433/#socket-socket
But these APIs doen't exist on MacOS, or at least didn't exist a few years ago when other frameworks were dealing with this.
https://github.com/rust-lang/rust/issues/24237
https://github.com/libuv/libuv/blob/81c77427afb39b6fa6c0ecffdc1c1e2b3dde7aea/src/unix/core.c#L417-L424
Regardless, those workarounds primarily pass a flag to the socket constructor which we cannot do.

I'm going to go back to skipping this test on Mac since we don't have a way to fix it there. Still need to figure out what's flaky with Windows Libuv.

tmds:
On mac and Linux corefx does this by default: https://github.com/dotnet/corefx/blob/master/src/Native/Unix/System.Native/pal_networking.c#L1292-L1307

Perhaps corefx should do this for Windows sockets too?
The best implementation is when you can pass the flag when you create the socket. Otherwise there is a race-condition when someone creates a child between socket construction and setting the flag.

Tratcher:
@tmds what's the expected behavior of FD_CLOEXEC? From the name it sounds like close-on-process-exit. It isn't working in my Mac tests (currently disabled). Note my tests don't close the parent process, only the socket in the parent process.

tmds:
Cloexec causes the handle not to be inherited by child processes. Exec refers to the system call to execute a program. When the child process execs, that handle is closed for the child.

Tratcher:
> When the child process execs, that handle is closed for the child.

Ah, but the duplicated handle remains open in the parent process even if I close the original Socket? That would explain why it's not working on Mac, but doesn't explain why Linux is working.

tmds:
The listen socket is not duplicated to the children, so when the parent closes it, the server is no longer accepting connections. That explains why it works on Linux. I don't know why it doesn't work on Mac. Mac has limited cloexec apis compared to Linux. Maybe some things are not really supported.

For Windows, I think it makes sense that corefx sockets by default are not inherited by children.

Issue #2901, opened 2018-09-07
Add timeout for settings ack


JunTaoLuo:
Meta issue: #2627

See settings ack

Issue #2814, opened 2018-08-10
Make connection and stream windows configurable


Tratcher:
Meta issue: #2627

Today the connection and stream window sizes are hardcoded at the default 65kb. These need to be independently configurable.

This maps to SETTINGS_INITIAL_WINDOW_SIZE.

Tratcher:
Http.Sys/IIS default: 1mb (observed)

halter73:
Is this for both the connection and stream windows?

Tratcher:
The setting applies to both and the window update applies to the connection (no requests were sent in my test).

Tratcher:
@halter73 have you started on this or should I steal it?

halter73:
As discussed, by all means steal it. Thanks.

PR #2945, opened 2018-09-20
Make HTTP/2 connection and stream windows configurable

Tratcher:
#2814  Allowing the window size to be set larger than the protocol default of 64kb is easy and I think I've got that covered. I picked a kestrel default only modestly larger than the protocol default for two reasons:
1) For maximum interop testing for non-protocol defaults, especially sending that initial window update when the connection is first negotiated.
2) We don't want it to be too large because it directly correlates to server side buffer sizes.

Allowing the value to be set lower than the protocol default has some challenges and is the reason this is still WIP. I have a commented out test case covering this (SETTINGS_ServerInitialWindowSizeLowerThanDefault). The root of the problem is that the client is allowed to send frames immediately after opening the connection using the default limits. They don't have to wait for the server's settings frame that may lower the limits. This means the server can't enforce a limit lower than the protocol default until it gets a settings ack.

There are ways to deal with this:
A) Do nothing, disallow setting the initial window size to a value lower than the protocol default.
B) Use the lower limit, allow input flow control to go negative so long as it doesn't go past what the protocol default would have been, but leave the pipe limits at the protocol default size. This is almost the same as A in practice, the lower limits are not really enforced.
C) Use the protocol defaults for the window size and pipe buffers until we get a settings ack, then lower them (possibly negative). This would require implementing a settings ack timeout #2901, implementing temporary negative flow control, and optionally re-adjusting the pipe thresholds ([not currently possible](https://github.com/dotnet/corefx/issues/30689)). 

Option C has some value in letting you reduce your limit enforcement, but it's quite complicated. Option A is by far the simplest and easiest for the short term. @shirhatti @halter73 I'll check with you on this tomorrow.

[Edit: clarifications on pipes]

### DIFF
--- a/src/Kestrel.Core/Http2Limits.cs
+++ b/src/Kestrel.Core/Http2Limits.cs
@@ -15,6 +15,8 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core
         private int _headerTableSize = (int)Http2PeerSettings.DefaultHeaderTableSize;
         private int _maxFrameSize = (int)Http2PeerSettings.DefaultMaxFrameSize;
         private int _maxRequestHeaderFieldSize = 8192;
+        private int _initialConnectionWindowSize = 1024 * 128; // Larger than the default 64kb, and larger than any one single stream.
+        private int _initialStreamWindowSize = 1024 * 96; // Larger than the default 64kb
 
         /// <summary>
         /// Limits the number of concurrent request streams per HTTP/2 connection. Excess streams will be refused.
@@ -93,6 +95,50 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core
                 }
 
                 _maxRequestHeaderFieldSize = value;
+            }
+        }
+
+        /// <summary>
+        /// Indicates how much request body data the server is willing to receive and buffer at a time aggregated across all
+        /// requests (streams) per connection. Note requests are also limited by <see cref="InitialStreamWindowSize"/>
+        /// <para>
+        /// Value must be greater than or equal to 65,535 and less than 2^31, defaults to 128 kb.
+        /// </para>
+        /// </summary>
+        public int InitialConnectionWindowSize
+        {
+            get => _initialConnectionWindowSize;
+            set
+            {
+                if (value < Http2PeerSettings.DefaultInitialWindowSize || value > Http2PeerSettings.MaxWindowSize)
+                {
+                    throw new ArgumentOutOfRangeException(nameof(value), value,
+                        CoreStrings.FormatArgumentOutOfRange(Http2PeerSettings.DefaultInitialWindowSize, Http2PeerSettings.MaxWindowSize));
+                }
+
+                _initialConnectionWindowSize = value;
+            }
+        }
+
+        /// <summary>
+        /// Indicates how much request body data the server is willing to receive and buffer at a time per stream.
+        /// Note connections are also limited by <see cref="InitialConnectionWindowSize"/>
+        /// <para>
+        /// Value must be greater than or equal to 65,535 and less than 2^31, defaults to 96 kb.
+        /// </para>
+        /// </summary>
+        public int InitialStreamWindowSize
+        {
+            get => _initialStreamWindowSize;
+            set
+            {
+                if (value < Http2PeerSettings.DefaultInitialWindowSize || value > Http2PeerSettings.MaxWindowSize)
+                {
+                    throw new ArgumentOutOfRangeException(nameof(value), value,
+                        CoreStrings.FormatArgumentOutOfRange(Http2PeerSettings.DefaultInitialWindowSize, Http2PeerSettings.MaxWindowSize));
+                }
+
+                _initialStreamWindowSize = value;
             }
         }
     }
--- a/src/Kestrel.Core/Internal/Http2/Http2Connection.cs
+++ b/src/Kestrel.Core/Internal/Http2/Http2Connection.cs
@@ -63,7 +63,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http2
         private readonly Http2ConnectionContext _context;
         private readonly Http2FrameWriter _frameWriter;
         private readonly HPackDecoder _hpackDecoder;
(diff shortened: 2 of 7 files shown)


Tratcher:
I've gone with option A above, not allowing the window size to be set lower than the protocol default. I have some ideas on how to implement option C, but they are incremental additions and require implementing a few other features that aren't ready yet. I'll file a separate issue for follow up.

src/Kestrel.Core/Internal/Http2/Http2Connection.cs
```diff
@@ -93,6 +93,8 @@ public Http2Connection(Http2ConnectionContext context)
             _serverSettings.HeaderTableSize = (uint)context.ServiceContext.ServerOptions.Limits.Http2.HeaderTableSize;
             _hpackDecoder = new HPackDecoder((int)_serverSettings.HeaderTableSize);
             _serverSettings.MaxHeaderListSize = (uint)context.ServiceContext.ServerOptions.Limits.MaxRequestHeadersTotalSize;
+            _serverSettings.InitialWindowSize = (uint)context.ServiceContext.ServerOptions.Limits.Http2.InitialWindowSize;
+            _inputFlowControl = new InputFlowControl(_serverSettings.InitialWindowSize, _serverSettings.InitialWindowSize / 2);
```
JunTaoLuo: nit: so far, it looks like `minWindowSizeIncrement` is always just `initialWindowSize/2`. Why don't we just do that calculation in the constructor instead of passing in two variables.

src/Kestrel.Core/Http2Limits.cs
```diff
@@ -14,6 +14,7 @@ public class Http2Limits
         private int _maxStreamsPerConnection = 100;
         private int _headerTableSize = (int)Http2PeerSettings.DefaultHeaderTableSize;
         private int _maxFrameSize = (int)Http2PeerSettings.DefaultMaxFrameSize;
+        private int _initialWindowSize = 1024 * 96; // Larger than the default 64kb
```
JunTaoLuo: In which interop tests do we need this limit to be bigger?

Tratcher: It's not that we need it to be bigger, it's that we need it to be a value other than the default so that some of the extra logic runs like sending a window update at the start of the connection. I want that active at least for preview3 so we make sure it works with clients.

src/Kestrel.Core/Internal/Http2/Http2Connection.cs
Tratcher: We're pretty sure that will need to change once we start doing some testing. We'll need to play with it to find a reasonable balance.

src/Kestrel.Core/Http2Limits.cs
Tratcher: @halter73 because we don't have any interop tests to speak of, I changed it so it would get exercised while doing normal usage during preview3.

Note IIS & Http.Sys use a default of 1mb. That seemed a bit much for starters, but I expect we'll want to make this size larger when we start perf testing.

src/Kestrel.Core/Http2Limits.cs
```diff
@@ -74,5 +75,26 @@ public int MaxFrameSize
                 _maxFrameSize = value;
             }
         }
+
+        /// <summary>
+        /// Indicates how much request body data the server is willing to receive and buffer at a time per stream and per connection.
```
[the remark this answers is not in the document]
Tratcher: Added a seperate connection window setting. Cross checking against the stream window is problematic. You don't want to fail the setter based on the value of another property, that's too order dependent. Nor do we have any delayed validation of the server options that I see, this would be a new failure code path.

src/Kestrel.Core/Internal/Http2/Http2Stream.cs
```diff
@@ -446,14 +446,16 @@ private void AbortCore(Exception abortReason)
...
             (
                 pool: _context.MemoryPool,
                 readerScheduler: ServiceContext.Scheduler,
                 writerScheduler: PipeScheduler.Inline,
-                pauseWriterThreshold: Http2PeerSettings.DefaultInitialWindowSize,
-                resumeWriterThreshold: Http2PeerSettings.DefaultInitialWindowSize,
+                // Never pause within the window range. Flow control will prevent more data from being added.
+                // See the assert in OnDataAsync.
+                pauseWriterThreshold: windowSize + 1,
+                resumeWriterThreshold: windowSize + 1,
```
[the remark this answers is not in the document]
Tratcher: Yeah, that assert is what caught this bug. Some of my new tests pushed just a little too close to the limit and found this. They aren't reproducing it at the moment so I added a test just for this.

test/Kestrel.InMemory.FunctionalTests/Http2/Http2ConnectionTests.cs
```diff
@@ -215,48 +228,93 @@ public async Task DATA_Received_GreaterThanDefaultInitialWindowSize_ReadByStream
...
+            var initialConnectionWindowSize = _connectionContext.ServiceContext.ServerOptions.Limits.Http2.InitialConnectionWindowSize;
+            var framesInConnectionWindow = initialConnectionWindowSize / Http2PeerSettings.DefaultMaxFrameSize;
+
+            // Grow the client stream windows so no stream WINDOW_UPDATEs need to be sent.
+            _clientSettings.InitialWindowSize = int.MaxValue;
+
+            await InitializeConnectionAsync(_waitForAbortApplication);
+
+            // Grow the client connection windows so no connection WINDOW_UPDATEs need to be sent.
+            await SendWindowUpdateAsync(0, int.MaxValue - (int)Http2PeerSettings.DefaultInitialWindowSize);
```
halter73: Can we remove the code that grows the client and connection stream windows for tests that do not or don't need to send large response bodies?

Tratcher: It looks like I can remove it from 2 out of 5 tests

Tratcher:
@sebastienros window sizes are something we're going to want to experiment with. A) are our defaults performant, and B) do we send updates at a reasonable interval (50%)?

Issue #2872, opened 2018-08-31
Make HPACK max field sizes configurable


Tratcher:
HPackDecoder.MaxStringOctets is hardcoded to 4kb. This represents the maximum size of header fields, both compressed and uncompressed. Since the request PathAndQuery is now a single http/2 header this may be a problem for some apps that were used to the prior request line length limit of 8kb.

https://github.com/aspnet/KestrelHttpServer/issues/2813

Compare to nginx http://nginx.org/en/docs/http/ngx_http_v2_module.html#http2_max_header_size which also defaults to 4kb, but only for the compressed size.

PR #2949, opened 2018-09-21
Add configurability for max header field size in HPACK

JunTaoLuo:
Addresses https://github.com/aspnet/KestrelHttpServer/issues/2872

### DIFF
--- a/src/Kestrel.Core/Http2Limits.cs
+++ b/src/Kestrel.Core/Http2Limits.cs
@@ -14,11 +14,12 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core
         private int _maxStreamsPerConnection = 100;
         private int _headerTableSize = (int)Http2PeerSettings.DefaultHeaderTableSize;
         private int _maxFrameSize = (int)Http2PeerSettings.DefaultMaxFrameSize;
+        private int _maxRequestHeaderFieldSize = 8192;
 
         /// <summary>
         /// Limits the number of concurrent request streams per HTTP/2 connection. Excess streams will be refused.
         /// <para>
-        /// Defaults to 100
+        /// Value must be greater than 0, defaults to 100
         /// </para>
         /// </summary>
         public int MaxStreamsPerConnection
@@ -72,6 +73,26 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core
                 }
 
                 _maxFrameSize = value;
+            }
+        }
+
+        /// <summary>
+        /// Indicates the size of the maximum allowed size of a request header field sequence. This limit applies to both name and value sequences in their compressed and uncompressed representations.
+        /// <para>
+        /// Value must be greater than 0, defaults to 8192
+        /// </para>
+        /// </summary>
+        public int MaxRequestHeaderFieldSize
+        {
+            get => _maxRequestHeaderFieldSize;
+            set
+            {
+                if (value <= 0)
+                {
+                    throw new ArgumentOutOfRangeException(nameof(value), value, CoreStrings.GreaterThanZeroRequired);
+                }
+
+                _maxRequestHeaderFieldSize = value;
             }
         }
     }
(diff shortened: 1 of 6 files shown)


src/Kestrel.Core/Http2Limits.cs
```diff
@@ -74,5 +75,25 @@ public int MaxFrameSize
...
             }
         }
+
+        /// <summary>
+        /// Indicates the size of the maximum allowed size of a header field sequence. This limit applies to both name and value sequences in their compressed and uncompressed representations.
+        /// <para>
+        /// Value must be greater than 0, defaults to 4096
+        /// </para>
+        /// </summary>
+        public int MaxHeaderFieldSize
```
JunTaoLuo: Technically `header field` refers to the name-value pair according to the RFC. But they didn't have a specific term for the name or value portion of the header field and I'd rather keep the name as simple as possible.

test/Kestrel.Core.Tests/HPackDecoderTests.cs
```diff
@@ -442,10 +443,28 @@ public void DecodesStringLength_GreaterThanLimit_Error()
...
         }
 
+        [Fact]
+        public void DecodesStringLength_LimitConfigurable()
+        {
+            var decoder = new HPackDecoder(DynamicTableInitialMaxSize, 4097);
+            var string4097 = new string('a', 4097);
+
+            var encoded = _literalHeaderFieldWithoutIndexingNewName
+                .Concat(new byte[] { 0x7f, 0x82, 0x1f }) // 4097 encoded with 7-bit prefix, no Huffman encoding
```
halter73: The `DecodesStringLength_GreaterThanLimit_Error` test is different.

```
var encoded = _literalHeaderFieldWithoutIndexingNewName
    .Concat(new byte[] { 0xff, 0x82, 0x1f }) // 4097 encoded with 7-bit prefix
```

Shouldn't these match? I'm guessing the old test is wrong since this test actually verifies the dictionary.

test/Kestrel.Core.Tests/KestrelServerLimitsTests.cs
```diff
@@ -331,12 +331,27 @@ public void Http2HeaderTableSizeDefault()
         }
 
         [Theory]
-        [InlineData(4097)]
+        [InlineData(int.MinValue)]
         [InlineData(-1)]
         public void Http2HeaderTableSizeInvalid(int value)
         {
-            var ex = Assert.Throws<ArgumentOutOfRangeException>(() => new KestrelServerLimits().Http2.MaxFrameSize = value);
```
halter73: I see the old test was improperly named, but we shouldn't remove coverage for MaxFrameSize if this isn't tested elsewhere. The old InlineData values should still be invalid for MaxFrameSize.

test/Kestrel.Core.Tests/KestrelServerLimitsTests.cs
```diff
@@ -331,12 +331,27 @@ public void Http2HeaderTableSizeDefault()
         }
 
         [Theory]
-        [InlineData(4097)]
+        [InlineData(int.MinValue)]
         [InlineData(-1)]
```
halter73: Might as well include 0 as an edge case.

src/Kestrel.Core/Http2Limits.cs
halter73: This is only for request headers, right? We should probably make that clear. We should probably name this MaxRequestHeaderFieldSize.

JunTaoLuo:
Updated

src/Kestrel.Core/Http2Limits.cs
```diff
@@ -14,11 +14,12 @@ public class Http2Limits
         private int _maxStreamsPerConnection = 100;
         private int _headerTableSize = (int)Http2PeerSettings.DefaultHeaderTableSize;
         private int _maxFrameSize = (int)Http2PeerSettings.DefaultMaxFrameSize;
+        private int _maxRequestHeaderFieldSize = 4096;
```
Tratcher: It should be at least 8kb so it can handle the same request path & query values as HTTP/1

test/Kestrel.Core.Tests/KestrelServerLimitsTests.cs
halter73: I see. You might as well fix the inline data for the Http2MaxFrameSizeInvalid test. `1 << 14 - 1` should be `(1 << 14) - 1`.

Issue #2922, opened 2018-09-13
Impose limits on decoded integers via HPACK


JunTaoLuo:
The HPACK specification allows integers of any size and it's up to the implementation to pick reasonable limits.

https://tools.ietf.org/html/rfc7541#section-5.1
```
This integer representation allows for values of indefinite size. 
... 
Integer encodings that exceed implementation limits -- in
value or octet length -- MUST be treated as decoding errors.
Different limits can be set for each of the different uses of
integers, based on implementation constraints.
```

PR #2953, opened 2018-09-24
Impose limits on decoded integers via HPACK

JunTaoLuo:
Addresses: https://github.com/aspnet/KestrelHttpServer/issues/2922. Setting the limit to Int32.MaxValue for now since that's also the limit for request header field size.

### DIFF
--- a/src/Kestrel.Core/Internal/Http2/HPack/IntegerDecoder.cs
+++ b/src/Kestrel.Core/Internal/Http2/HPack/IntegerDecoder.cs
@@ -3,40 +3,65 @@
 
 namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http2.HPack
 {
+    /// <summary>
+    /// The maximum we will decode is Int32.MaxValue, which is also the maximum request header field size.
+    /// </summary>
     public class IntegerDecoder
     {
         private int _i;
         private int _m;
 
-        public int Value { get; private set; }
-
-        public bool BeginDecode(byte b, int prefixLength)
+        /// <summary>
+        /// Callers must ensure higher bits above the prefix are cleared before calling this method.
+        /// </summary>
+        /// <param name="b"></param>
+        /// <param name="prefixLength"></param>
+        /// <param name="result"></param>
+        /// <returns></returns>
+        public bool BeginTryDecode(byte b, int prefixLength, out int result)
         {
             if (b < ((1 << prefixLength) - 1))
             {
-                Value = b;
-                return true;
-            }
-            else
-            {
-                _i = b;
-                _m = 0;
-                return false;
-            }
-        }
-
-        public bool Decode(byte b)
-        {
-            _i = _i + (b & 127) * (1 << _m);
-            _m = _m + 7;
-
-            if ((b & 128) != 128)
-            {
-                Value = _i;
+                result = b;
                 return true;
             }
 
+            _i = b;
+            _m = 0;
+            result = 0;
             return false;
         }
+
+        public bool TryDecode(byte b, out int result)
+        {
+            var m = _m; // Enregister
+            var i = _i + ((b & 0x7f) << m); // Enregister
+
+            if ((b & 0x80) == 0)
+            {
+                // Int32.MaxValue only needs a maximum of 5 bytes to represent and the last byte cannot have any value set larger than 0x7
+                if ((m > 21 && b > 0x7) || i < 0)
+                {
+                    ThrowIntegerTooBigException();
+                }
+
+                result = i;
+                return true;
+            }
+            else if (m > 21)
+            {
+                // Int32.MaxValue only needs a maximum of 5 bytes to represent
+                ThrowIntegerTooBigException();
+            }
+
+            _m = m + 7;
+            _i = i;
+
+            result = 0;
+            return false;
+        }
+
+        public static void ThrowIntegerTooBigException()
+            => throw new HPackDecodingException(CoreStrings.HPackErrorIntegerTooBig);
     }
 }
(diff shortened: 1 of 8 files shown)


src/Kestrel.Core/Internal/Http2/HPack/IntegerDecoder.cs
```diff
@@ -17,24 +19,33 @@ public bool BeginDecode(byte b, int prefixLength)
...
+                // Int32.MaxValue only needs a maximum of 5 bytes to represent and the last byte cannot have any value set larger than 0x7
+                if ((_m > 28 && (b & 0xf8) > 0) || _i > int.MaxValue)
+                {
+                    throw new HPackDecodingException("Integer too big");
+                }
+
+                Value = unchecked((int)_i);
                 return true;
             }
+            else if (_m > 28)
```
JunTaoLuo: I'm explicitly excluding encodings that has extra empty continuation bytes. Technically it's possible to send an infinite amount of empty continuation bytes `0x80` without actually increasing the encoded integer's value.

src/Kestrel.Core/Internal/Http2/HPack/IntegerDecoder.cs
```diff
@@ -17,24 +19,33 @@ public bool BeginDecode(byte b, int prefixLength)
...
             _m = _m + 7;
 
-            if ((b & 128) != 128)
+            if ((b & 0x80) != 0x80)
             {
-                Value = _i;
+                // Int32.MaxValue only needs a maximum of 5 bytes to represent and the last byte cannot have any value set larger than 0x7
+                if ((_m > 28 && (b & 0xf8) > 0) || _i > int.MaxValue)
+                {
+                    throw new HPackDecodingException("Integer too big");
```
JunTaoLuo: I'll make this an entry in string.resx

test/Kestrel.Core.Tests/IntegerDecoderTests.cs
```diff
@@ -34,6 +34,21 @@ public void IntegerDecode(int i, int prefixLength, byte[] octets)
             Assert.Equal(i, decoder.Value);
         }
```
Tratcher: Can you add a round trip test? for 0 to max int, encode, decode, and verify?

JunTaoLuo:
Good idea

On Tue, Sep 25, 2018 at 9:40 AM Chris Ross <notifications@github.com<mailto:notifications@github.com>> wrote:

@Tratcher commented on this pull request.

________________________________

In test/Kestrel.Core.Tests/IntegerDecoderTests.cs<https://github.com/aspnet/KestrelHttpServer/pull/2953#discussion_r220265192>:

> @@ -34,6 +34,21 @@ public void IntegerDecode(int i, int prefixLength, byte[] octets)
             Assert.Equal(i, decoder.Value);
         }


Can you add a round trip test? for 0 to max int, encode, decode, and verify?

test/Kestrel.Core.Tests/IntegerDecoderTests.cs
JunTaoLuo: Turns out this might not be feasible. {0 .. int.maxint}X{1 .. 8 prefix length} by my estimate is going to take 2.5 hours.

src/Kestrel.Core/Internal/Http2/HPack/IntegerDecoder.cs
```diff
@@ -17,24 +19,33 @@ public bool BeginDecode(byte b, int prefixLength)
...
+                }
+
+                Value = unchecked((int)_i);
```
halter73: Isn't unchecked the default?

JunTaoLuo: I was playing around with unchecked to see if it made any difference to the benchmarks but it didn't so I suspect you are right.

test/Kestrel.Core.Tests/IntegerDecoderTests.cs
halter73: You should still do the round trip test with IntegerData. Maybe add a few more cases if you can think of anything interesting, but I agree every possible input seems like a bit much.

JunTaoLuo: I ended up opting for {0 .. 1 << 8} U {int.MaxValue - 1 << 8 .. int.MaxValue} I think that's a reasonable range.

JunTaoLuo:
Might want to take another look since I added a bit of optimization.

Before this PR:
```
                  Method |      Mean |     Error |    StdDev |          Op/s | Scaled | ScaledSD | Allocated |
------------------------ |----------:|----------:|----------:|--------------:|-------:|---------:|----------:|
 DecodeSingleByteInteger |  1.126 ns | 0.0273 ns | 0.0255 ns | 888,439,368.7 |   1.00 |     0.00 |       0 B |
  DecodeMultiByteInteger | 15.122 ns | 0.2951 ns | 0.3030 ns |  66,130,977.7 |  13.44 |     0.39 |       0 B |
```
After this PR:
```
                  Method |       Mean |     Error |    StdDev |            Op/s | Scaled | ScaledSD | Allocated |
------------------------ |-----------:|----------:|----------:|----------------:|-------:|---------:|----------:|
 DecodeSingleByteInteger |  0.8466 ns | 0.0057 ns | 0.0053 ns | 1,181,147,635.1 |   1.00 |     0.00 |       0 B |
  DecodeMultiByteInteger | 15.7103 ns | 0.0728 ns | 0.0568 ns |    63,652,630.6 |  18.56 |     0.13 |       0 B |
```

Note the multibyte scenario didn't improve since we limit checks which mostly balanced the gains.

src/Kestrel.Core/Internal/Http2/HPack/IntegerDecoder.cs
```diff
@@ -3,40 +3,65 @@
...
+            }
 
+            _m = m + 7;
+            _i = i;
+
+            result = 0;
             return false;
         }
+
+        public static void ThrowIntegerTooBigException()
```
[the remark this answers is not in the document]
JunTaoLuo: Added

Issue #2175, opened 2017-11-20
Enable ApiCheck for Socket Transport


halter73:
https://github.com/aspnet/KestrelHttpServer/pull/2173/files#r151965500

JunTaoLuo:
Looks like we have ApiCheck enabled for all of the Kestrel packages. However, it looks like the base lines are out of date since they were last updated to 2.0. We should set the baselines at 2.1 and then see whether anything changed for 2.2

PR #2959, opened 2018-09-27
Update API check baselines to 2.1.0

JunTaoLuo:
Addresses https://github.com/aspnet/KestrelHttpServer/issues/2175. I'm updating the baselines to 2.1.0


src/Connections.Abstractions/breakingchanges.netcore.json
```diff
@@ -0,0 +1,7 @@
+[
+  {
+    "TypeId": "public class Microsoft.AspNetCore.Connections.DefaultConnectionContext : Microsoft.AspNetCore.Connections.ConnectionContext, System.IDisposable, Microsoft.AspNetCore.Connections.Features.IConnectionIdFeature, Microsoft.AspNetCore.Connections.Features.IConnectionItemsFeature, Microsoft.AspNetCore.Connections.Features.IConnectionTransportFeature, Microsoft.AspNetCore.Connections.Features.IConnectionUserFeature, Microsoft.AspNetCore.Connections.Features.IConnectionLifetimeFeature",
```
JunTaoLuo: Adding this exception since we moved the Abort() method in https://github.com/aspnet/KestrelHttpServer/commit/ac31e5ab30e6f6023ddb12b24f76fa72e764a94a#diff-499c3a0185b3217a93e225bb80161218 but the API check tool can't figure out the method has been relocated and still exist on this type.

src/Kestrel.Https/baseline.netcore.json
```diff
@@ -1,5 +1,5 @@
 {
-  "AssemblyIdentity": "Microsoft.AspNetCore.Server.Kestrel.Https, Version=2.0.2.0, Culture=neutral, PublicKeyToken=adb9793829ddae60",
+  "AssemblyIdentity": "Microsoft.AspNetCore.Server.Kestrel.Https, Version=2.1.0.0, Culture=neutral, PublicKeyToken=adb9793829ddae60",
```
JunTaoLuo: @Eilon @ryanbrandenburg I'm using 2.1.0 as the baseline, though in other repos I'm seeing a different base versions. For example, Mvc uses 2.1.1: https://github.com/aspnet/Mvc/blob/release/2.2/src/Microsoft.AspNetCore.Mvc/baseline.netcore.json#L2. Which version should we be using?

src/Kestrel.Transport.Libuv/baseline.netcore.json
```diff
@@ -47,185 +47,6 @@
       ],
       "GenericParameters": []
     },
-    {
```
JunTaoLuo: FYI this is because of https://github.com/aspnet/KestrelHttpServer/pull/2426 which is expected.

src/Kestrel.Https/baseline.netcore.json
halter73: If we know the baseline is exactly the same for 2.1.0 and 2.1.3 as it theoretically should be, I would say using 2.1.0 as the baseline version is better so we're not inclined to update the baseline version with every patch.

Do we automatically verify that we don't add new APIs in patch releases? If not, I can see the argument for using the latest patch to ensure we don't break said new APIs in the future.

Eilon: Doesn't matter among 2.1.x because the APIs should be identical within those.  Do whatever is easiest for now.

src/Connections.Abstractions/breakingchanges.netcore.json
Tratcher: Hmm, that's technically a binary break. Putting Abort back on the child class and having it call the base class would be safer. Try it?

Eilon: Moving a method from a derived class to a base class is not a breaking change at all, is it?

Eilon: According to https://github.com/dotnet/corefx/blob/master/Documentation/coding-guidelines/breaking-change-rules.md#members:

> Members
> ✓ Allowed
> ...
> Moving a member onto a class higher in the hierarchy tree of the type from which it was removed

Tratcher: Ok, can we update the tool so we don't have this exception then?

JunTaoLuo: Filed https://github.com/aspnet/BuildTools/issues/767. However, I'm going to go with the breaking changes exception for now since I don't want to wait until that's resolved before checking this in.

src/Kestrel.Https/baseline.netcore.json
JunTaoLuo: I think 2.1.0 is sufficient for now then.

src/Connections.Abstractions/breakingchanges.netcore.json
JunTaoLuo: Apparently we are moving from ApiCheck to ApiCompat, so not sure if we'll update ApiCheck.

Eilon: That's right, there is no planned work on the tool.

PR #2965, opened 2018-09-28
typos, inline outs, redundant usings

SimonCropp:


davidfowl:
These "reformat the world" PRs are never great to review and cause conflicts with existing in flights PRs 87% of the time. That's not to say we won't take it but sometimes they do more harm than good. It's also pretty pointless to do this if there's no way to enforce this style in the future.

SimonCropp:
i dont think any of these are stylistic. it fixes spelling errors and makes some code more succinct. eg would you reject a PR that removes a few 100 redundant usings? that is not "style", it is "removing noise"?

perhaps the "inline outs" is stylistic. happy to revert those if u want

SimonCropp:
also i havnt actually applied any formatting tool on any of the files.

Tratcher:
Agreed on the merge conflict issue. However, I checked with the other devs and we're all in between work items so this should be safe to merge at the moment.

PR #2966, opened 2018-09-28
remove some possibly redundant asyncs

SimonCropp:
not sure on these ones??

### DIFF
--- a/src/Kestrel.Core/Internal/AddressBinder.cs
+++ b/src/Kestrel.Core/Internal/AddressBinder.cs
@@ -18,7 +18,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal
 {
     internal class AddressBinder
     {
-        public static async Task BindAsync(IServerAddressesFeature addresses,
+        public static Task BindAsync(IServerAddressesFeature addresses,
             KestrelServerOptions serverOptions,
             ILogger logger,
             Func<ListenOptions, Task> createBinding)
@@ -43,7 +43,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal
             listenOptions.Clear();
             addresses.Addresses.Clear();
 
-            await strategy.BindAsync(context).ConfigureAwait(false);
+            return strategy.BindAsync(context);
         }
 
         private static IStrategy CreateStrategy(ListenOptions[] listenOptions, string[] addresses, bool preferAddresses)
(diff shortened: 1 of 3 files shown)


src/Kestrel.Core/Internal/AddressBinder.cs
```diff
@@ -43,7 +43,7 @@ internal class AddressBinder
             listenOptions.Clear();
             addresses.Addresses.Clear();
 
-            await strategy.BindAsync(context).ConfigureAwait(false);
+            return strategy.BindAsync(context);
```
davidfowl: Nope, all code in the startup path needs ConfigureAwait. Also there's no point "optimizing" this code path since it's server startup.

test/shared/DummyApplication.cs
```diff
@@ -40,9 +40,9 @@ public void DisposeContext(HttpContext context, Exception exception)
             _httpContextFactory?.Dispose(context);
         }
 
-        public async Task ProcessRequestAsync(HttpContext context)
+        public Task ProcessRequestAsync(HttpContext context)
         {
-            await _requestDelegate(context);
+            return _requestDelegate(context);
```
davidfowl: This one is fine

src/Kestrel.Core/Internal/AddressBinder.cs
SimonCropp: isnt ConfigureAwait only relevant if the task is awaited in this context? if it is not awaited no state machine is created, and it is up to the caller to choose to ConfigureAwait or not?

davidfowl: Yes, that's true. Still I see no value in removing this one only. Usually, you'd make it so that you need to get off the sync context once and all lower calls can avoid the calls to ConfigureAwait, but this really doesn't save anything.

SimonCropp: a little confused. this is the recommended way of handling a single async call, it is a little bit faster (yes at startup but still), it is just as readable. So why not accept the PR?

Also why is this change bad but the one in the test is ok?

davidfowl: > this is the recommended way of handling a single async call

The recommendations for these types of things are extremely fuzzy. This changes the behavior (exceptions are now not faulted tasks but they can throw synchronously), it's of the method and there's no perf gain to be had. This is called once per application.

The removal of the ConfigureAwait when all of the calls underneath it have ConfigureAwait also doesn't sit well with me. Subtle changes like this that affect the behavior when there's no real performance gain is what I think there's no point making this change here.

davidfowl: > Also why is this change bad but the one in the test is ok?

The test method has a single line of code. This method has lots of code and ends with a tail call. They are different.

SimonCropp:
Fair enough :)

Issue #2832, opened 2018-08-17
Track aborted streams for a given grace period


Tratcher:
RE: https://github.com/aspnet/KestrelHttpServer/issues/2799 
ContentLength_Received_MultipleDataFramesOverSize_Reset exposes a race in the product that can cause unnessisary connection closures.

The client is sending multiple data frames for the stream. The server encounters an error for one of those frames and decides to reset the stream. It gets dropped from the active stream list. When the next frame is processed it does not belong to an active stream and it triggers a connection error. The client was not have a chance to observe that reset as the next frame is already in flight. 

Proposal: We need to track server aborted streams for a given grace period, drain their data, and even continue to decompress their headers to maintain the HPACK state. This gives the client a chance to observe the reset and stop sending frames. If the client continues to send frames  after the grace period expires then we can treat it as a connection error as before. This prevents unnecessary connection teardowns for stream errors.

Many of the limits we're implementing would trigger such resets. The code in Http2Connection.ValidateHeader today that throws stream exceptions similarly triggers this issue is corrupting the HPACK state by not processing all of the headers. Any error that is not fatal to the connection should put the stream into a graceful drain state.

One caveat: The client must not send additional streams if they have sent (rather than received) a reset. This is a connection error. Client aborted streams should not be given a graceful drain period. (#2154 h2spec http2/5.1/8)

Related:
#2756 - unread request body draining
#2811, #2812, #2813 header size limits

PR #2973, opened 2018-10-02
Disallow any frames after a reset is received

Tratcher:
#2154 This covers the last H2spec test. The spec says that the client must not send additional header or data frames on a stream after sending a reset. There was a race where additional frames would be accepted up until the request processing ended and the stream was removed from the active streams list. I added a new flag to prevent this. This is similar to the EndOfStream flag but with a different error message and it does not cause the request body to appear complete but rather aborted. Technically this could be treated as a stream error but I've opted to leave it as a connection error for now, we handle those better.

#2832 (not included) is related but covers an opposite scenario where the request is aborted from the server side but the client is still allowed to send frames for a given period. The current race there is that the frames are only accepted until the request processing ends and the stream is removed from the active streams list. The client may need more time to react so we'll want to do additional work there. I've not included any changes for that in this PR because there's no functional overlap, only conceptual.

### DIFF
--- a/src/Kestrel.Core/Internal/Http2/Http2Connection.cs
+++ b/src/Kestrel.Core/Internal/Http2/Http2Connection.cs
@@ -224,6 +224,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http2
                     catch (Http2StreamErrorException ex)
                     {
                         Log.Http2StreamError(ConnectionId, ex);
+                        // The client doesn't know this error is coming, allow draining additional frames for now.
                         AbortStream(_incomingFrame.StreamId, new IOException(ex.Message, ex));
                         await _frameWriter.WriteRstStreamAsync(ex.StreamId, ex.ErrorCode);
                     }
@@ -448,6 +449,11 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http2
 
             if (_streams.TryGetValue(_incomingFrame.StreamId, out var stream))
             {
+                if (stream.DoNotDrainRequest)
+                {
+                    // Hard abort, do not allow any more frames on this stream.
+                    throw new Http2ConnectionErrorException(CoreStrings.FormatHttp2ErrorStreamAborted(_incomingFrame.Type, stream.StreamId), Http2ErrorCode.STREAM_CLOSED);
+                }
                 if (stream.EndStreamReceived)
                 {
                     // http://httpwg.org/specs/rfc7540.html#rfc.section.5.1
@@ -501,6 +507,12 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http2
 
             if (_streams.TryGetValue(_incomingFrame.StreamId, out var stream))
             {
+                if (stream.DoNotDrainRequest)
+                {
+                    // Hard abort, do not allow any more frames on this stream.
+                    throw new Http2ConnectionErrorException(CoreStrings.FormatHttp2ErrorStreamAborted(_incomingFrame.Type, stream.StreamId), Http2ErrorCode.STREAM_CLOSED);
+                }
+
                 // http://httpwg.org/specs/rfc7540.html#rfc.section.5.1
                 //
                 // ...an endpoint that receives any frames after receiving a frame with the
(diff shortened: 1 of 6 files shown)


Tratcher:
@dotnet-bot test OSX 10.12 Debug

Tratcher:

test/Kestrel.Transport.FunctionalTests/Http2/H2SpecTests.cs
```diff
@@ -55,7 +55,7 @@ public static TheoryData<H2SpecTestCase> H2SpecTestCases
             get
             {
                 var dataset = new TheoryData<H2SpecTestCase>();
-                var toSkip = new[] { "http2/5.1/8" };
+                var toSkip = new string[] { /*"http2/5.1/8"*/ };
```
JunTaoLuo: just remove the entire variable.

Tratcher: I'd rather not, if we have to come back and skip these tests for other reasons later I don't want to re-invent this.

JunTaoLuo: That's what the commit history is for isn't it?

src/Kestrel.Core/Internal/Http2/Http2Connection.cs
```diff
@@ -609,7 +621,8 @@ private Task ProcessRstStreamFrameAsync()
             }
 
             ThrowIfIncomingFrameSentToIdleStream();
-            AbortStream(_incomingFrame.StreamId, new IOException(CoreStrings.Http2StreamResetByClient));
+            // No additional inbound header or data frames are allowed for this stream after receiving a reset.
```
JunTaoLuo: From the spec, it also seems like we should throw if we receive a second rst frame after receiving the first rst frame.

test/Kestrel.Transport.FunctionalTests/Http2/H2SpecTests.cs
Tratcher: Searching for something that was removed is much more difficult.

src/Kestrel.Core/Internal/Http2/Http2Connection.cs
JunTaoLuo: Same for window update.

test/Kestrel.Transport.FunctionalTests/Http2/H2SpecTests.cs
JunTaoLuo: I'm not a fan of dead code but fair enough. Not really too bad since this is only in tests.

src/Kestrel.Core/Internal/Http2/Http2Connection.cs
Tratcher: Hmm, those are going to be tough to enforce. I can check for them while the stream is still active but we ignore them if they arrive after the stream is removed from the collection. The window updates could be handled as part of a cool down, but we'd only planned on tracking server aborted streams for that.

JunTaoLuo: Checking while the streams are active should be fine for now. These are supposed to be stream errors anyway which would reset the stream that's already no longer active.

src/Kestrel.Core/Internal/Http2/Http2Connection.cs
```diff
@@ -771,6 +796,12 @@ private Task ProcessWindowUpdateFrameAsync()
             }
             else if (_streams.TryGetValue(_incomingFrame.StreamId, out var stream))
             {
+                if (stream.DoNotDrain)
```
JunTaoLuo: Add a test for these scenarios as well.

src/Kestrel.Core/Internal/Http2/Http2Stream.cs
```diff
@@ -53,6 +53,7 @@ public Http2Stream(Http2StreamContext context)
         public bool RequestBodyStarted { get; private set; }
         public bool EndStreamReceived => (_completionState & StreamCompletionFlags.EndStreamReceived) == StreamCompletionFlags.EndStreamReceived;
         private bool IsAborted => (_completionState & StreamCompletionFlags.Aborted) == StreamCompletionFlags.Aborted;
+        internal bool DoNotDrain => (_completionState & StreamCompletionFlags.DoNotDrain) == StreamCompletionFlags.DoNotDrain;
```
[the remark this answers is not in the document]
Tratcher: It matters if the abort was local or remote. If it was remote we don't want to accept any more frames. Today that's racey but this additional check makes it reliable. If it's a local abort we want to drain frames (which is still flaky, I'll work on that more in the next PR).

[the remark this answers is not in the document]
halter73: If the abort was local, you would only want to send frames buffered prior to the abort, no?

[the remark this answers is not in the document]
Tratcher: Ah, this is request specific. Does DoNotDrainRequest clarify?

src/Kestrel.Core/Internal/Http2/Http2Connection.cs
```diff
@@ -609,7 +621,20 @@ private Task ProcessRstStreamFrameAsync()
             }
 
             ThrowIfIncomingFrameSentToIdleStream();
-            AbortStream(_incomingFrame.StreamId, new IOException(CoreStrings.Http2StreamResetByClient));
+
+            if (_streams.TryGetValue(_incomingFrame.StreamId, out var stream))
+            {
+                // Second reset?
```
halter73: Remove the question mark since we're sure this is could only be caused by a second reset.

PR #2974, opened 2018-10-02
Suppress the execution context per operation

davidfowl:
Today we capture and restore the execution context per Socket operation. There's no need to do that because it already happens in the async state machine. 

I need to measure the performance since we're making an additional method call per operation to save it on the other side, it might be a wash. Ideally, we would just use https://github.com/dotnet/corefx/issues/32582

## BEFORE
image

## AFTER
image

Look at that beautiful flattened frame 😄. This is the code being avoided https://github.com/dotnet/coreclr/blob/302630ed5a3730470e9ffeeebcd38c737c03963d/src/System.Private.CoreLib/shared/System/Threading/ExecutionContext.cs#L126-L202

### DIFF
--- a/src/Kestrel.Transport.Sockets/Internal/SocketReceiver.cs
+++ b/src/Kestrel.Transport.Sockets/Internal/SocketReceiver.cs
@@ -4,6 +4,7 @@
 using System;
 using System.IO.Pipelines;
 using System.Net.Sockets;
+using System.Threading;
 
 namespace Microsoft.AspNetCore.Server.Kestrel.Transport.Sockets.Internal
 {
@@ -17,9 +18,12 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Transport.Sockets.Internal
         {
             _awaitableEventArgs.SetBuffer(Array.Empty<byte>(), 0, 0);
 
-            if (!_socket.ReceiveAsync(_awaitableEventArgs))
+            using (ExecutionContext.SuppressFlow())
             {
-                _awaitableEventArgs.Complete();
+                if (!_socket.ReceiveAsync(_awaitableEventArgs))
+                {
+                    _awaitableEventArgs.Complete();
+                }
             }
 
             return _awaitableEventArgs;
@@ -36,9 +40,12 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Transport.Sockets.Internal
 #else
 #error TFMs need to be updated
 #endif
-            if (!_socket.ReceiveAsync(_awaitableEventArgs))
+            using (ExecutionContext.SuppressFlow())
             {
-                _awaitableEventArgs.Complete();
+                if (!_socket.ReceiveAsync(_awaitableEventArgs))
+                {
+                    _awaitableEventArgs.Complete();
+                }
             }
 
             return _awaitableEventArgs;
--- a/src/Kestrel.Transport.Sockets/Internal/SocketReceiver.cs
+++ b/src/Kestrel.Transport.Sockets/Internal/SocketReceiver.cs
@@ -18,7 +18,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Transport.Sockets.Internal
         {
             _awaitableEventArgs.SetBuffer(Array.Empty<byte>(), 0, 0);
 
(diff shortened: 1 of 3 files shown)


muratg (commented):
Did you run the full test suite?

src/Kestrel.Transport.Sockets/Internal/SocketReceiver.cs
```diff
@@ -17,9 +18,12 @@ public SocketAwaitableEventArgs WaitForDataAsync()
         {
             _awaitableEventArgs.SetBuffer(Array.Empty<byte>(), 0, 0);
 
-            if (!_socket.ReceiveAsync(_awaitableEventArgs))
+            using (ExecutionContext.SuppressFlow())
```
benaadams: Can the follow already be suppressed when it gets here? (in which case this call would throw)

davidfowl: Unlikely but I can add the crazy code. Regardless, I’m sad this is per call instead of per construction

davidfowl:
> Did you run the full test suite?

Not sure what you mean? The regular tests? Yes they run as part of the PR and the execution context doesn't need to be captured here because it's already captured.

davidfowl:

src/Kestrel.Transport.Sockets/Internal/SocketReceiver.cs
davidfowl: The flow can't be suppressed here, this API is called from a single place in the code base.

halter73:
What's the primary motivation for this: perf or less stack frames in debug/exception output? Am I correct in understanding you expect the perf to stay the same?

davidfowl:
Looks like it's slightly faster on the plaintext benchmark:

## OLD  

|       RPS | CPU (%) | Memory (MB) | Avg. Latency (ms) | Startup (ms) | First Request (ms) | Latency (ms) |
| --------- | ------- | ----------- | ----------------- | ------------ | ------------------ | ------------ |
| 3,164,263 |      87 |         290 |                 2 |          192 |               39.9 |          0.5 |

## NEW 

|       RPS | CPU (%) | Memory (MB) | Avg. Latency (ms) | Startup (ms) | First Request (ms) | Latency (ms) |
| --------- | ------- | ----------- | ----------------- | ------------ | ------------------ | ------------ |
| 3,203,452 |      72 |         285 |               2.1 |          192 |                 47 |          0.6 |

src/Kestrel.Transport.Sockets/Internal/SocketSenderReceiverBase.cs
```diff
@@ -19,5 +20,10 @@ protected SocketSenderReceiverBase(Socket socket, PipeScheduler scheduler)
         }
 
         public void Dispose() => _awaitableEventArgs.Dispose();
+
+        protected AsyncFlowControl? SuppressExecutionContext()
+        {
+            return ExecutionContext.IsFlowSuppressed() ? (AsyncFlowControl?)null : ExecutionContext.SuppressFlow();
```
davidfowl: cc @benaadams

benaadams: You can `using` a null nullable? (Tested, yes you can)

That's nice and elegant :)

benaadams (approved):

davidfowl:
Running the performance numbers again to make sure they weren't all noise:

## Before

|       RPS | CPU (%) | Memory (MB) | Avg. Latency (ms) | Startup (ms) | First Request (ms) | Latency (ms) |
| --------- | ------- | ----------- | ----------------- | ------------ | ------------------ | ------------ |
| 3,170,415 |      83 |         323 |               1.8 |          191 |               43.3 |          0.5 |

## After

|       RPS | CPU (%) | Memory (MB) | Avg. Latency (ms) | Startup (ms) | First Request (ms) | Latency (ms) |
| --------- | ------- | ----------- | ----------------- | ------------ | ------------------ | ------------ |
| 3,175,771 |      76 |         306 |               1.8 |          185 |               43.1 |          0.5 |

Not, sure, it might be noise.

benaadams:
Should skip out this bit
image

Issue #2230, opened 2017-12-20
Consume BindingAddress in ServerAddress 


jkotalik:
[HttpAbstractions has a new type](https://github.com/aspnet/HttpAbstractions/commit/150bb3faf413226d053777bbc72618fb284a6970) that will be consumed for using the IServerAddressFeature. Though Kestrel cannot remove the ServerAddress type, we can deprecate it and then start using the BindingAddress internally. Kestrel will also have to throw if the PathBase is set before consuming the type.

PR #2975, opened 2018-10-02
Implement #2230 - Consume BindingAddress instead of ServerAddress

SonicGD:
Hello. I've tried to implement changes for #2230. I've done this:

1. Marked ServerAddress obsolete and suppress CS0618 errors in ServerAddressTests
2. Added reference to Microsoft.AspNetCore.Http in Kestrel.Core.csproj
3. Replaced usage of ServerAddress.FromUrl with BindingAddress.Parse
4. Update tests

I hope nothing is wrong =)

### DIFF
--- a/benchmarkapps/PlatformBenchmarks/BenchmarkConfigurationHelpers.cs
+++ b/benchmarkapps/PlatformBenchmarks/BenchmarkConfigurationHelpers.cs
@@ -4,6 +4,7 @@
 using System;
 using System.Net;
 using Microsoft.AspNetCore.Hosting;
+using Microsoft.AspNetCore.Http.Internal;
 using Microsoft.AspNetCore.Server.Kestrel.Core;
 using Microsoft.Extensions.Configuration;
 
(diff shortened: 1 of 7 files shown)


Tratcher:
@halter73 @shirhatti let's decide if we can take this in release/2.2. If so then in 3.0 it'd be easy to remove the obsolete type.

Tratcher:
@muratg @shirhatti thoughts on obsoleting [ServerAddress](https://github.com/aspnet/KestrelHttpServer/search?q=ServerAddress&unscoped_q=ServerAddress) in 2.2? It's officially a public type, but not one used in any public APIs.

muratg:
@shirhatti thoughts on this?

shirhatti:
If we take this in 2.2, we should probably update all servers. I'd like to see this https://github.com/aspnet/HttpSysServer/issues/422 go in as well

Tratcher:
@shirhatti this is an implementation detail for the servers, not a public API we expect people to use directly with the servers.

Tratcher:
Offline notes: This is fine to take in 2.2, and the HttpSys change would be orthogonal.

@SonicGD can you move this to the release/2.2 branch? If not then I can get to it next week.

SonicGD:
@Tratcher Sure, will do today

Tratcher:
Thanks

Issue #622, opened 2016-02-10
Allow Response Trailer Headers?


benaadams:
Now (shortly?) reading request header trailers #619; ability to add them to chunked encoding responses?

Currently supported by [`System.Web`](https://msdn.microsoft.com/sr-latn-rs/library/system.net.http.headers.httpresponseheaders.trailer(v=vs.110).aspx) 

Don't think browsers pay attention to them; but some discussion on https://github.com/whatwg/fetch/issues/34 as to their importance for CDNs etc; and whether to include them in the fetchapi

benaadams:
Would need design

halter73:
I don't think this is super important.

If you really want this, I think you should be willing to manually chunk your response which you can do if you explicitly set `Transfer-Encoding: chunked`. So nothing is topping you from sending response trailing headers today. We just don't make it easy.

madelson:
@halter73 how do you send response trailing headers using existing APIs?

halter73:
@madelson I'm not aware of any way to do this using HttpClient APIs. You would probably have to build something on raw sockets.

Tratcher:
Proposal: Request and response trailer headers belong in their own collections. These should get there own features. These features should also have a HasStarted type flag indicating if the trailers have been started or finished yet. The response trailers feature may even have a Flush method to end the body and send the trailers immediately.

Today kestrel puts request trailers into the normal request headers collection, but this may cause conflicts or concurrent modification errors.

NickCraver:
I hit a desire for these today, in MiniProfiler. There's now a [`Server-Timing` header](https://wicg.github.io/server-timing/), which really can't get its content until quite late in the pipe. While I can send the profiler ID (which is saved at the end of a request, but fetched in *another* request), putting the entirety of the profiling summary in a header send *before anything happens* has some fundamental issues with respect to our current linear understanding of space-time.

But `Trailers`! It's ideal here. At the *end* of the request (the very last step pretty much), we have a complete picture of what happened. In fact in [the examples of `Server-Timing`](https://wicg.github.io/server-timing/#examples), it's even recommended.  This is definitely a feature I'd use in MiniProfiler, and I totally agree with them being in their own collection (of the same type). 

There is a minor question open here, where in the initial headers collection needs knowledge a trailer exists. I'm not sure the best API for that - perhaps there's just a method (called before headers are sent) that takes a header name, which registers them for the `Trailers` header sent in the initial collection?

[...]

PR #2982, opened 2018-10-03
Http/2 response trailers

Tratcher:
#622, Depends on https://github.com/aspnet/HttpAbstractions/pull/1043.

If any headers have been added to the trailer collection then they will be sent at the end of the request where we would normally send an empty data frame with END_STREAM.

### DIFF
--- a/build/dependencies.props
+++ b/build/dependencies.props
@@ -11,13 +11,13 @@
     <MicrosoftAspNetCoreAllPackageVersion>2.2.0-preview3-35359</MicrosoftAspNetCoreAllPackageVersion>
     <MicrosoftAspNetCoreBenchmarkRunnerSourcesPackageVersion>2.2.0-preview3-35359</MicrosoftAspNetCoreBenchmarkRunnerSourcesPackageVersion>
     <MicrosoftAspNetCoreCertificatesGenerationSourcesPackageVersion>2.2.0-preview3-35359</MicrosoftAspNetCoreCertificatesGenerationSourcesPackageVersion>
-    <MicrosoftAspNetCoreHostingAbstractionsPackageVersion>2.2.0-preview3-35359</MicrosoftAspNetCoreHostingAbstractionsPackageVersion>
-    <MicrosoftAspNetCoreHostingPackageVersion>2.2.0-preview3-35359</MicrosoftAspNetCoreHostingPackageVersion>
-    <MicrosoftAspNetCoreHttpAbstractionsPackageVersion>2.2.0-preview3-35359</MicrosoftAspNetCoreHttpAbstractionsPackageVersion>
-    <MicrosoftAspNetCoreHttpFeaturesPackageVersion>2.2.0-preview3-35359</MicrosoftAspNetCoreHttpFeaturesPackageVersion>
-    <MicrosoftAspNetCoreHttpPackageVersion>2.2.0-preview3-35359</MicrosoftAspNetCoreHttpPackageVersion>
+    <MicrosoftAspNetCoreHostingAbstractionsPackageVersion>2.2.0-a-preview3-tratcher-trailers-17154</MicrosoftAspNetCoreHostingAbstractionsPackageVersion>
+    <MicrosoftAspNetCoreHostingPackageVersion>2.2.0-a-preview3-tratcher-trailers-17154</MicrosoftAspNetCoreHostingPackageVersion>
+    <MicrosoftAspNetCoreHttpAbstractionsPackageVersion>2.2.0-a-preview3-tratcher-trailers-16760</MicrosoftAspNetCoreHttpAbstractionsPackageVersion>
+    <MicrosoftAspNetCoreHttpFeaturesPackageVersion>2.2.0-a-preview3-tratcher-trailers-16760</MicrosoftAspNetCoreHttpFeaturesPackageVersion>
+    <MicrosoftAspNetCoreHttpPackageVersion>2.2.0-a-preview3-tratcher-trailers-16760</MicrosoftAspNetCoreHttpPackageVersion>
     <MicrosoftAspNetCoreTestingPackageVersion>2.2.0-preview3-35359</MicrosoftAspNetCoreTestingPackageVersion>
-    <MicrosoftAspNetCoreWebUtilitiesPackageVersion>2.2.0-preview3-35359</MicrosoftAspNetCoreWebUtilitiesPackageVersion>
+    <MicrosoftAspNetCoreWebUtilitiesPackageVersion>2.2.0-a-preview3-tratcher-trailers-16760</MicrosoftAspNetCoreWebUtilitiesPackageVersion>
     <MicrosoftExtensionsActivatorUtilitiesSourcesPackageVersion>2.2.0-preview3-35359</MicrosoftExtensionsActivatorUtilitiesSourcesPackageVersion>
     <MicrosoftExtensionsConfigurationBinderPackageVersion>2.2.0-preview3-35359</MicrosoftExtensionsConfigurationBinderPackageVersion>
     <MicrosoftExtensionsConfigurationCommandLinePackageVersion>2.2.0-preview3-35359</MicrosoftExtensionsConfigurationCommandLinePackageVersion>
--- a/src/Kestrel.Core/Internal/Http/HttpHeaders.Generated.cs
+++ b/src/Kestrel.Core/Internal/Http/HttpHeaders.Generated.cs
@@ -8998,4 +8998,235 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
             }
         }
     }
+
+    public partial class HttpResponseTrailers
+    {
+        private static byte[] _headerBytes = new byte[]
+        {
+            13,10,69,84,97,103,58,32,
+        };
+
+        private long _bits = 0;
+        private HeaderReferences _headers;
+
+
+        
+        public StringValues HeaderETag
+        {
+            get
+            {
+                StringValues value;
+                if ((_bits & 1L) != 0)
+                {
+                    value = _headers._ETag;
+                }
+                return value;
+            }
+            set
+            {
+                _bits |= 1L;
+                _headers._ETag = value; 
+            }
+        }
+
+        protected override int GetCountFast()
+        {
+            return (_contentLength.HasValue ? 1 : 0 ) + BitCount(_bits) + (MaybeUnknown?.Count ?? 0);
+        }
+
+        protected override bool TryGetValueFast(string key, out StringValues value)
+        {
+            switch (key.Length)
+            {
+                case 4:
+                    {
+                        if ("ETag".Equals(key, StringComparison.OrdinalIgnoreCase))
+                        {
+                            if ((_bits & 1L) != 0)
+                            {
+                                value = _headers._ETag;
+                                return true;
+                            }
+                            return false;
+                        }
+                    }
+                    break;
+            }
+
+            return MaybeUnknown?.TryGetValue(key, out value) ?? false;
+        }
+
+        protected override void SetValueFast(string key, in StringValues value)
+        {
+            ValidateHeaderValueCharacters(value);
+            switch (key.Length)
+            {
+                case 4:
+                    {
+                        if ("ETag".Equals(key, StringComparison.OrdinalIgnoreCase))
+                        {
+                            _bits |= 1L;
+                            _headers._ETag = value;
+                            return;
+                        }
+                    }
+                    break;
+            }
+
+            SetValueUnknown(key, value);
+        }
+
+        protected override bool AddValueFast(string key, in StringValues value)
+        {
+            ValidateHeaderValueCharacters(value);
+            switch (key.Length)
+            {
+                case 4:
+                    {
+                        if ("ETag".Equals(key, StringComparison.OrdinalIgnoreCase))
+                        {
+                            if ((_bits & 1L) == 0)
+                            {
+                                _bits |= 1L;
+                                _headers._ETag = value;
+                                return true;
+                            }
+                            return false;
+                        }
+                    }
+                    break;
+            }
+
+            Unknown.Add(key, value);
+            // Return true, above will throw and exit for false
+            return true;
+        }
+
+        protected override bool RemoveFast(string key)
+        {
+            switch (key.Length)
+            {
+                case 4:
+                    {
+                        if ("ETag".Equals(key, StringComparison.OrdinalIgnoreCase))
+                        {
+                            if ((_bits & 1L) != 0)
+                            {
+                                _bits &= ~1L;
+                                _headers._ETag = default(StringValues);
+                                return true;
+                            }
+                            return false;
+                        }
+                    }
+                    break;
+            }
+
+            return MaybeUnknown?.Remove(key) ?? false;
+        }
+
+        protected override void ClearFast()
+        {
+            MaybeUnknown?.Clear();
+            _contentLength = null;
+            var tempBits = _bits;
+            _bits = 0;
+            if(HttpHeaders.BitCount(tempBits) > 12)
+            {
+                _headers = default(HeaderReferences);
+                return;
+            }
+            
+            if ((tempBits & 1L) != 0)
+            {
+                _headers._ETag = default(StringValues);
+                if((tempBits & ~1L) == 0)
+                {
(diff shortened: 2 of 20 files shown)


src/Kestrel.Core/Internal/Http2/Http2OutputProducer.cs
```diff
@@ -200,7 +204,19 @@ private async Task ProcessDataWrites()
                 {
                     readResult = await _dataPipe.Reader.ReadAsync();
 
-                    await _frameWriter.WriteDataAsync(_streamId, _flowControl, readResult.Buffer, endStream: readResult.IsCompleted);
+                    if (readResult.IsCompleted && (_getTrailers()?.Any() ?? false))
```
davidfowl: LINQ? Use .Count

src/Kestrel.Core/Internal/Http2/Http2OutputProducer.cs
```diff
@@ -37,11 +39,13 @@ public class Http2OutputProducer : IHttpOutputProducer, IHttpOutputAborter
             Http2FrameWriter frameWriter,
             StreamOutputFlowControl flowControl,
             ITimeoutControl timeoutControl,
-            MemoryPool<byte> pool)
+            MemoryPool<byte> pool,
+            Func<HttpResponseTrailers> getTrailers)
```
davidfowl: This is a bit weird and allocatey

src/Kestrel.Core/Internal/Http2/Http2Stream.cs
```diff
@@ -40,7 +40,8 @@ public Http2Stream(Http2StreamContext context)
                 _context.ServerPeerSettings.InitialWindowSize / 2);
 
             _outputFlowControl = new StreamOutputFlowControl(context.ConnectionOutputFlowControl, context.ClientPeerSettings.InitialWindowSize);
-            _http2Output = new Http2OutputProducer(context.StreamId, context.FrameWriter, _outputFlowControl, context.TimeoutControl, context.MemoryPool);
+            _http2Output = new Http2OutputProducer(context.StreamId, context.FrameWriter, _outputFlowControl, context.TimeoutControl, context.MemoryPool,
+                () => Trailers);
```
davidfowl: This allocates. Make it not allocate 😄

src/Kestrel.Core/Internal/Http2/Http2OutputProducer.cs
Tratcher: Yeah, it was this or passing in the entire Http2Stream which this type seemed to be avoiding. I wasn't passing in the trailers directly since they're lazily allocated.

Tratcher:
Small update

src/Kestrel.Core/Internal/Http2/Http2Stream.FeatureCollection.cs
```diff
@@ -1,12 +1,34 @@
...
+    public partial class Http2Stream : IHttp2StreamIdFeature, IHttpResponseTrailersFeature
     {
+        internal HttpResponseTrailers Trailers { get; set; }
+        private IHeaderDictionary _userTrailers;
+
+        IHeaderDictionary IHttpResponseTrailersFeature.Trailers
+        {
+            get
+            {
+                if (Trailers == null)
```
JunTaoLuo: I suppose this is dependent on how _userTrailer works. But suppose the setter was called first and a non-null value was provided, this will still create a HttpResponseTrailer that will never be used.

src/Kestrel.Core/Internal/Http/HttpHeaders.Generated.cs
```diff
@@ -8998,4 +8998,235 @@ public bool MoveNext()
             }
         }
     }
+
+    public partial class HttpResponseTrailers
+    {
+        private static byte[] _headerBytes = new byte[]
+        {
+            13,10,69,84,97,103,58,32,
```
JunTaoLuo: What's this used for?

build/dependencies.props
```diff
@@ -11,13 +11,13 @@
     <MicrosoftAspNetCoreAllPackageVersion>2.2.0-preview3-35359</MicrosoftAspNetCoreAllPackageVersion>
     <MicrosoftAspNetCoreBenchmarkRunnerSourcesPackageVersion>2.2.0-preview3-35359</MicrosoftAspNetCoreBenchmarkRunnerSourcesPackageVersion>
     <MicrosoftAspNetCoreCertificatesGenerationSourcesPackageVersion>2.2.0-preview3-35359</MicrosoftAspNetCoreCertificatesGenerationSourcesPackageVersion>
-    <MicrosoftAspNetCoreHostingAbstractionsPackageVersion>2.2.0-preview3-35359</MicrosoftAspNetCoreHostingAbstractionsPackageVersion>
-    <MicrosoftAspNetCoreHostingPackageVersion>2.2.0-preview3-35359</MicrosoftAspNetCoreHostingPackageVersion>
-    <MicrosoftAspNetCoreHttpAbstractionsPackageVersion>2.2.0-preview3-35359</MicrosoftAspNetCoreHttpAbstractionsPackageVersion>
-    <MicrosoftAspNetCoreHttpFeaturesPackageVersion>2.2.0-preview3-35359</MicrosoftAspNetCoreHttpFeaturesPackageVersion>
-    <MicrosoftAspNetCoreHttpPackageVersion>2.2.0-preview3-35359</MicrosoftAspNetCoreHttpPackageVersion>
+    <MicrosoftAspNetCoreHostingAbstractionsPackageVersion>2.2.0-a-preview3-tratcher-trailers-17152</MicrosoftAspNetCoreHostingAbstractionsPackageVersion>
```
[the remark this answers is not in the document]
Tratcher: [Hosting](https://github.com/aspnet/Hosting/commit/3b857f2a00f4e7e44f5920c1202cf6f9c321bb25) is not interesting, it's only pass through dependencies to avoid conflicts. HttpAbstractions is linked above.

src/Kestrel.Core/Internal/Http2/Http2Stream.FeatureCollection.cs
```diff
@@ -1,12 +1,34 @@
...
+                {
+                    Trailers = new HttpResponseTrailers();
+                }
+                return _userTrailers ?? Trailers;
+            }
+            set
```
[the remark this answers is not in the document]
Tratcher: This is the implementation of the feature interface.

src/Kestrel.Core/Internal/Http/HttpHeaders.Generated.cs
Tratcher: Generated magic 😁 `\r\nETag: `. Not needed for HTTP/2 but would be needed for an eventual HTTP/1.1 implementation of trailers.

src/Kestrel.Core/Internal/Http2/Http2Stream.FeatureCollection.cs
halter73: This has caused me to notice how incredibly inconsistent we are about using the internal or user-replaceable request and response header dictionaries for internal logic.

#2984

tools/CodeGenerator/KnownHeaders.cs
```diff
@@ -269,6 +269,21 @@ public static string GeneratedFile()
                 PrimaryHeader = responsePrimaryHeaders.Contains("Content-Length")
             }})
             .ToArray();
+
+            var responseTrailers = new[]
+            {
+                "ETag",
```
halter73: Has anyone asked to send an ETag in the trailer? Do any clients support this? Adding "grpc-status" and "grpc-message" might be worthwhile.

Tratcher: ETag is a commonly discussed example for trailers, though it's not clear where it's supported. I added it here primarily as a placeholder to make the code generator work correctly. I hesitate to add GRPC specific headers, it's not clear how final that spec is.

src/Kestrel.Core/Internal/Http2/Http2FrameWriter.cs
```diff
@@ -193,6 +193,48 @@ public void WriteResponseHeaders(int streamId, int statusCode, IHeaderDictionary
...
+        public Task WriteResponseTrailers(int streamId, HttpResponseTrailers headers)
+        {
+            lock (_writeLock)
+            {
+                if (_completed)
+                {
+                    return Task.CompletedTask;
+                }
+
+                _outgoingFrame.PrepareHeaders(Http2HeadersFrameFlags.END_STREAM, streamId);
```
halter73: It looks to me that the bulk of this method could be shared between this and WriteResponseHeaders.

src/Kestrel.Core/Internal/Http2/Http2FrameWriter.cs
```diff
@@ -193,6 +193,48 @@ public void WriteResponseHeaders(int streamId, int statusCode, IHeaderDictionary
...
+                var buffer = _headerEncodingBuffer.AsSpan();
+                var done = _hpackEncoder.BeginEncode(EnumerateHeaders(headers), buffer, out var payloadLength);
```
halter73: What happens if a trailer is larger than the max response frame size?

Tratcher: It will throw that HPackEncodingException I added, same as response headers.

src/Kestrel.Core/Internal/Http2/Http2FrameWriter.cs
Tratcher: Consolidated

jtattermusch:
CC @jtattermusch

src/Kestrel.Core/Internal/Http/HttpResponseTrailers.cs
```diff
@@ -0,0 +1,65 @@
...
+            object IEnumerator.Current => _current;
+
+            public void Dispose()
+            {
+            }
+
+            public void Reset()
+            {
+                _state = 0;
+            }
```
halter73: Could most of this be moved into the HttpHeaders base class?

test/Kestrel.InMemory.FunctionalTests/Http2/Http2StreamTests.cs
```diff
@@ -1340,6 +1341,223 @@ public async Task MaxRequestBodySize_AppCanRaiseLimit(bool includeContentLength)
...
+                withLength: 11,
+                withFlags: (byte)Http2DataFrameFlags.NONE,
+                withStreamId: 1);
+
+            var goAway = await ExpectAsync(Http2FrameType.GOAWAY,
+                withLength: 8,
+                withFlags: (byte)Http2DataFrameFlags.NONE,
+                withStreamId: 0);
+
+            VerifyGoAway(goAway, 1, Http2ErrorCode.INTERNAL_ERROR);
```
halter73: Why not WaitForConnectionErrorAsync?

Tratcher: The ordering was wrong. The abort and GoAway happen outside of the normal connection loop and I needed to wait for those first. Also the transport Abort isn't fully implemented for our fake in-memory transport so I have to close the request pipe to unblock the connection loop. If I did that too early then I could get a standard GoAway rather than the error GoAway.

src/Kestrel.Core/Internal/Http2/Http2FrameWriter.cs
```diff
@@ -157,39 +160,69 @@ public void WriteResponseHeaders(int streamId, int statusCode, IHeaderDictionary
...
+                {
+                    _outgoingFrame.PrepareHeaders(Http2HeadersFrameFlags.END_STREAM, streamId);
+                    var buffer = _headerEncodingBuffer.AsSpan();
+                    var done = _hpackEncoder.BeginEncode(EnumerateHeaders(headers), buffer, out var payloadLength);
+                    FinishWritingHeaders(streamId, payloadLength, done);
+                }
+                catch (HPackEncodingException hex)
+                {
+                    _log.HPackEncodingError(_connectionId, streamId, hex);
+                    _http2Connection.Abort(new ConnectionAbortedException("", hex));
```
halter73: Can we just copy the inner exception message instead of using an empty string here?

src/Kestrel.Core/Internal/Http2/Http2FrameWriter.cs
```diff
@@ -157,39 +160,69 @@ public void WriteResponseHeaders(int streamId, int statusCode, IHeaderDictionary
                     _outgoingFrame.PrepareHeaders(Http2HeadersFrameFlags.NONE, streamId);
                     var buffer = _headerEncodingBuffer.AsSpan();
                     var done = _hpackEncoder.BeginEncode(statusCode, EnumerateHeaders(headers), buffer, out var payloadLength);
+                    FinishWritingHeaders(streamId, payloadLength, done);
+                }
+                catch (HPackEncodingException hex)
+                {
+                    _log.HPackEncodingError(_connectionId, streamId, hex);
+                    _http2Connection.Abort(new ConnectionAbortedException("", hex));
```
halter73: Ditto about copying the exception message

src/Kestrel.Core/Internal/Http2/Http2FrameWriter.cs
JunTaoLuo: Wouldn't that duplicate the exception message? I think currently, it will just show the hex's exception message once which is probably what we want.

src/Kestrel.Core/Internal/Http/HttpResponseTrailers.cs
Tratcher: The Enumerator needs to be here because it's a partial and half of it's implemented in the generated code. The remaining methods are either related to the enumerator or differ across the implementations.

src/Kestrel.Core/Internal/Http2/Http2FrameWriter.cs
Tratcher: I'd rather preserve the exception type and stack.

halter73: It's important that we don't use empty or default messages for ConnectionAbortedExceptions.

I've been on a mission to ensure that each ConnectionAbortedException message is unique and indicative of the reason for the abort. This message, and only the message not (Exception.ToString(), will now be printed by the transport's Abort() implementation to indicate the reason for closing the connection.

Tratcher: Updated

jtattermusch:
Is there an ETA for releasing a preview nuget package with trailers support?

JunTaoLuo:
Yes this will be available in 2.2.0-preview3 which will be available early next week.

davidfowl:
> Is there an ETA for releasing a preview nuget package with trailers support?

We have ci builds if you want to play before official previews. https://dotnet.myget.org/feed/dotnet-core/package/nuget/Microsoft.AspNetCore.Server.Kestrel

jtattermusch:
@davidfowl thanks. I tried that and got  
```
It was not possible to find any compatible framework version
The specified framework 'Microsoft.AspNetCore.All', version '2.2.0-preview3-35458' was not found.
  - Check application dependencies and target a framework version installed at:
      /usr/local/google/home/jtattermusch/DotnetSDK-2.2.100-preview2/
  - Installing .NET Core prerequisites might help resolve this problem:
      http://go.microsoft.com/fwlink/?LinkID=798306&clcid=0x409
  - The .NET Core framework and SDK can be installed from:
      https://aka.ms/dotnet-download
  - The following versions are installed:
      2.2.0-preview2-35157 at [/usr/local/google/home/jtattermusch/DotnetSDK-2.2.100-preview2/shared/Microsoft.AspNetCore.All]
```
because apparently in order to run AspNetCore apps, I need to have a corresponding SDK installed. I'm using the latest SDK from https://www.microsoft.com/net/download/dotnet-core/2.2, not sure where to find anything more recent.

JunTaoLuo:
You can find them at: https://github.com/dotnet/versions/tree/master/build-info/dotnet/product/cli/release/2.2. Note that it's usually easier to look for the newest SDK/Runtime first then find the corresponding packages on the myget feed if needed.

Tratcher:
FYI the nightly instructions are [here](https://github.com/aspnet/aspnetcore#daily-builds).

PR #2983, opened 2018-10-04
Use logical 'or' instead of bitwise in IsConnectionReset method.

flash2048:
Now in method **IsConnectionReset** used:
`return errno == ECONNRESET || errno == EPIPE || errno == ENOTCONN | errno == EINVAL;`

I think need use logical '**or**' for the last comparison too like this:
`return errno == ECONNRESET || errno == EPIPE || errno == ENOTCONN || errno == EINVAL;`

### DIFF
--- a/src/Kestrel.Transport.Libuv/Internal/LibuvConstants.cs
+++ b/src/Kestrel.Transport.Libuv/Internal/LibuvConstants.cs
@@ -22,7 +22,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Transport.Libuv.Internal
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
         public static bool IsConnectionReset(int errno)
         {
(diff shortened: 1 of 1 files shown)


davidfowl:
OOPS 😄

davidfowl:

davidfowl:
@flash2048 can you add a simple unit test that checks for those error codes?

flash2048:
@davidfowl It will difficult because the value for static variables gets with RuntimeInformation.IsOSPlatform. And other tests for IsConnectionReset method do not contain in the project.

Tratcher:
Hmm, looks like this was [added](https://github.com/aspnet/KestrelHttpServer/commit/a9d6cf13eb96b17b1cba15f80936516a02a31a32#diff-e5bbfa90617242cc0a8bbf5269407337) in 2.2.0-preview1.

halter73:
Thanks @flash2048

PR #2992, opened 2018-10-06
Add HTTP/2 keep-alive and request header timeouts

halter73:


src/Kestrel.Core/Internal/Http2/Http2Connection.cs
```diff
@@ -988,6 +1027,24 @@ void IHttp2StreamLifetimeHandler.OnStreamCompleted(int streamId)
             }
         }
 
+        private void StartKeepAliveTimeout()
```
JunTaoLuo: Looks like these methods should only be called under the _stateLock. We should add a comment just to remind ourselves.

JunTaoLuo: Actually seems like the only contention would be _currentTimeout having not updated yet. Why not expose the _timeoutReason on TimeoutControl?

src/Kestrel.Core/Internal/Http2/Http2Connection.cs
```diff
@@ -988,6 +1027,24 @@ void IHttp2StreamLifetimeHandler.OnStreamCompleted(int streamId)
...
+            _currentTimeout = TimeoutReason.KeepAlive;
+        }
+
+        private void StartRequestHeadersTimeout()
+        {
+            _context.TimeoutControl.SetTimeout(ServerOptions.Limits.RequestHeadersTimeout.Ticks, TimeoutReason.RequestHeaders);
+            _currentTimeout = TimeoutReason.RequestHeaders;
+        }
+
+        private void CancelNonRateTimeout()
```
Tratcher: Do you expect a matching CancelRateTimeout?

src/Kestrel.Core/Internal/Http2/Http2Connection.cs
```diff
@@ -174,6 +182,7 @@ public async Task ProcessRequestsAsync<TContext>(IHttpApplication<TContext> appl
             try
             {
                 ValidateTlsRequirements();
```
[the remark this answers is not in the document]
halter73: Yep. https://github.com/aspnet/KestrelHttpServer/blob/6164ec72253238f6b57e3ab71c9fa9d949dc67d1/src/Kestrel.Core/Internal/HttpsConnectionAdapter.cs#L127-L128

src/Kestrel.Core/Internal/Http2/Http2Connection.cs
halter73: Not sure actually. I think I'm going to take @JunTaoLuo's recommendation and just put a CurrentSetTimeoutReason on TimeoutControl and remove these methods for now.

src/Kestrel.Core/Internal/Http2/Http2Connection.cs
halter73: I like adding a CurrentSetTimeoutReason property to TimeoutControl, but it definitely won't remove the locking requirement.

PR #2994, opened 2018-10-07
removes unsafes from HttpHeaders

rehli:
Removes unsafe from class HttpHeaders
- Uses ReadOnlySpan<char> instead of Pointers

Addresses #2932

### DIFF
--- a/src/Kestrel.Core/Internal/Http/HttpHeaders.cs
+++ b/src/Kestrel.Core/Internal/Http/HttpHeaders.cs
@@ -277,101 +277,59 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
             }
         }
 
-        public static unsafe ConnectionOptions ParseConnection(in StringValues connection)
+        public static ConnectionOptions ParseConnection(in StringValues connection)
         {
             var connectionOptions = ConnectionOptions.None;
 
             var connectionCount = connection.Count;
             for (var i = 0; i < connectionCount; i++)
             {
-                var value = connection[i];
-                fixed (char* ptr = value)
+                var value = connection[i].AsSpan();
+                var currentPosition = 0;
+                while (currentPosition < value.Length)
                 {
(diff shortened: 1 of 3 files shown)


Tratcher:
FYI the IHttpResponseTrailersFeature errors will go away if you rebase on top of today's dependencies.props update.

Tratcher:
> "0HLHCS5QGRR50" bad request data: "The message body length cannot be determined because the final transfer coding was set to 'chunked' instead of 'chunked'."

I don't think this new parser is working quite yet.

rehlert95:
Now the Tests are running :)
I am not quite sure what is more performant in IsTokenEndValid(): Getting the new Offset with the ref, or returning a Tuple from the Method. 
And another Question, do you prefer more, smaller PRs like this one since it is just refactoring and not a Feature, or do you prefer bigger ones?

Tratcher:
By all means smaller PRs, so long as they are complete.

Have you tried our [benchmarks](https://github.com/aspnet/KestrelHttpServer/tree/release/2.2/benchmarks/Kestrel.Performance)?

src/Kestrel.Core/Internal/Http/HttpHeaders.cs
```diff
@@ -277,170 +277,137 @@ public static void ValidateHeaderNameCharacters(string headerCharacters)
...
+                    var offset = 0;
+                    if (token.Length >= 9 && (token[offset] | 0x20) == 'k')
                     {
-                        while (tokenEnd < end && *tokenEnd != ',')
-                        {
-                            tokenEnd++;
-                        }
-
-                        while (ch < tokenEnd && *ch == ' ')
+                        if ((token[++offset] | 0x20) == 'e' &&
```
gfoidl: This will introduce bounds-checks, because the JIT can't assume that `offset` is within the allowed bounds of `token`.

Other places as well.

muratg:

PR #2998, opened 2018-10-11
Use ValueTask instead of custom awaiter for SocketAwaitableEventArgs

davidfowl:
- This could improve the performance of async operations (reduce the number of allocations)

TODO: Measure performance, I may combine this with #2974

### DIFF
--- a/src/Kestrel.Transport.Sockets/Internal/SocketAwaitableEventArgs.cs
+++ b/src/Kestrel.Transport.Sockets/Internal/SocketAwaitableEventArgs.cs
@@ -106,19 +106,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Transport.Sockets.Internal
                 return new ValueTask<int>(this, 0);
             }
 
-            int bytesTransferred = BytesTransferred;
-            SocketError error = SocketError;
-
-            Reset();
-
-            return error == SocketError.Success ?
-                new ValueTask<int>(bytesTransferred) :
-                new ValueTask<int>(Task.FromException<int>(new SocketException((int)error)));
-        }
-
-        private void Reset()
-        {
-            Volatile.Write(ref _completion, null);
+            return CompleteSocketOperation();
         }
 
         /// <summary>Initiates a send operation on the associated socket.</summary>
@@ -130,6 +118,16 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Transport.Sockets.Internal
                 return new ValueTask<int>(this, 0);
             }
 
+            return CompleteSocketOperation();
+        }
+
+        private void Reset()
+        {
+            Volatile.Write(ref _completion, null);
+        }
+
+        private ValueTask<int> CompleteSocketOperation()
+        {
             int bytesTransferred = BytesTransferred;
             SocketError error = SocketError;
 
--- a/src/Kestrel.Transport.Sockets/Internal/SocketAwaitableEventArgs.cs
+++ b/src/Kestrel.Transport.Sockets/Internal/SocketAwaitableEventArgs.cs
@@ -27,12 +27,10 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Transport.Sockets.Internal
 
         protected override void OnCompleted(SocketAsyncEventArgs _)
         {
-            var callback = Interlocked.CompareExchange(ref _callback, _callbackCompleted, null);
+            var callback = Interlocked.Exchange(ref _callback, _callbackCompleted);
 
             if (callback != null)
             {
-                Debug.Assert(!ReferenceEquals(callback, _callbackCompleted));
-
                 object state = UserToken;
                 UserToken = null;
 
@@ -67,13 +65,13 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Transport.Sockets.Internal
 
             if (ReferenceEquals(awaitableState, _callbackCompleted))
             {
+                _callback = continuation;
                 ThreadPool.UnsafeQueueUserWorkItem(_waitCallback, this);
             }
             else if (awaitableState != null)
             {
                 throw new InvalidOperationException("Multiple continuations registered!");
             }
-
         }
 
         public int GetResult(short token)
@@ -132,8 +130,9 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Transport.Sockets.Internal
         {
             var args = (SocketAwaitableEventArgs)state;
             var callbackState = args.UserToken;
-            args.UserToken = null;
             var callback = args._callback;
+            args.UserToken = null;
+            args._callback = _callbackCompleted;
             callback(callbackState);
         }
     }
--- a/src/Kestrel.Transport.Sockets/Internal/SocketAwaitableEventArgs.cs
+++ b/src/Kestrel.Transport.Sockets/Internal/SocketAwaitableEventArgs.cs
@@ -72,7 +72,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Transport.Sockets.Internal
                 Task.Factory.StartNew(continuation, state);
 #endif
             }
-            else
+            else if (awaitableState != null)
             {
                 Debug.Fail("Multiple continuations registered!");
             }
--- a/src/Kestrel.Transport.Sockets/Internal/SocketAwaitableEventArgs.cs
+++ b/src/Kestrel.Transport.Sockets/Internal/SocketAwaitableEventArgs.cs
@@ -5,40 +5,93 @@ using System;
 using System.Diagnostics;
 using System.IO.Pipelines;
 using System.Net.Sockets;
-using System.Runtime.CompilerServices;
 using System.Threading;
 using System.Threading.Tasks;
+using System.Threading.Tasks.Sources;
 
 namespace Microsoft.AspNetCore.Server.Kestrel.Transport.Sockets.Internal
 {
-    public class SocketAwaitableEventArgs : SocketAsyncEventArgs, ICriticalNotifyCompletion
+    public class SocketAwaitableEventArgs : SocketAsyncEventArgs, IValueTaskSource<int>
     {
-        private static readonly Action _callbackCompleted = () => { };
+        private static readonly Action<object> _callbackCompleted = _ => { };
 
         private readonly PipeScheduler _ioScheduler;
 
-        private Action _callback;
+        private Action<object> _completion;
 
         public SocketAwaitableEventArgs(PipeScheduler ioScheduler)
         {
             _ioScheduler = ioScheduler;
         }
 
-        public SocketAwaitableEventArgs GetAwaiter() => this;
-        public bool IsCompleted => ReferenceEquals(_callback, _callbackCompleted);
-
-        public int GetResult()
+        protected override void OnCompleted(SocketAsyncEventArgs _)
         {
-            Debug.Assert(ReferenceEquals(_callback, _callbackCompleted));
+            Action<object> completion = _completion;
 
-            _callback = null;
-
-            if (SocketError != SocketError.Success)
+            if (completion != null || (completion = Interlocked.CompareExchange(ref _completion, _callbackCompleted, null)) != null)
             {
-                ThrowSocketException(SocketError);
+                Debug.Assert(!ReferenceEquals(completion, _callbackCompleted));
+
+                object state = UserToken;
+                UserToken = null;
+
+                _ioScheduler.Schedule(completion, state);
+            }
+        }
+
+        public ValueTaskSourceStatus GetStatus(short token)
+        {
+            if (ReferenceEquals(_completion, _callbackCompleted))
+            {
+                if (SocketError != SocketError.Success)
+                {
+                    return ValueTaskSourceStatus.Faulted;
+                }
+
+                return ValueTaskSourceStatus.Succeeded;
             }
 
-            return BytesTransferred;
+            return ValueTaskSourceStatus.Pending;
+        }
+
+        public void OnCompleted(Action<object> continuation, object state, short token, ValueTaskSourceOnCompletedFlags flags)
+        {
+            // We're ignoring ValueTaskSourceOnCompletedFlags and token since this is used in a single place in the code base
+            // so we don't need to handle capturing the SynchronizationContext and ExecutionContext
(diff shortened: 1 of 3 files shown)


davidfowl:

src/Kestrel.Transport.Sockets/Internal/SocketAwaitableEventArgs.cs
```diff
@@ -5,74 +5,137 @@
...
+#if NETCOREAPP2_1
+                ThreadPool.QueueUserWorkItem(continuation, state, preferLocal: false);
+#else
+                Task.Factory.StartNew(continuation, state);
+#endif
+            }
+            else if (awaitableState != null)
             {
-                ThrowSocketException(SocketError);
+                Debug.Fail("Multiple continuations registered!");
```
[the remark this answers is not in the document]
davidfowl: Thats why we run tests in both. I'm not worried about this though, this isn't general purpose. This is specifically used in a single place.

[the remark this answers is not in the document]
halter73: 👍 I see no reason to not just throw or even Environment.FailFast in this case.

src/Kestrel.Transport.Sockets/Internal/SocketAwaitableEventArgs.cs
```diff
@@ -5,74 +5,137 @@
...
+            }
+
+            return CompleteSocketOperation();
         }
 
-        public void Complete()
+        private void Reset()
         {
-            OnCompleted(this);
+            Volatile.Write(ref _completion, null);
```
halter73: Good catch. I guess theoretically without this volatile write, if `OnCompleted(SocketAsyncEventArgs)` was called from another thread before `OnCompleted(Action<object> continuation, ...`, the first OnCompleted call could observe a stale _completion callback and invoke it a second time.

src/Kestrel.Transport.Sockets/Internal/SocketAwaitableEventArgs.cs
```diff
@@ -5,74 +5,137 @@
...
             _ioScheduler = ioScheduler;
         }
 
-        public SocketAwaitableEventArgs GetAwaiter() => this;
-        public bool IsCompleted => ReferenceEquals(_callback, _callbackCompleted);
+        protected override void OnCompleted(SocketAsyncEventArgs _)
+        {
+            Action<object> completion = _completion;
+
+            if (completion != null || (completion = Interlocked.CompareExchange(ref _completion, _callbackCompleted, null)) != null)
```
halter73: I'm not sure `completion != null` is safe here. I think you need a memory barrier here for the same reason you need it in Reset(). For one, without a memory barrier, reading UserToken could be reordered to before reading _completion. There's also the issue with not transitioning into the _callbackCompleted state which would make GetStatus wrong momentarily, but I don't think that's very important.

src/Kestrel.Transport.Sockets/Internal/SocketAwaitableEventArgs.cs
```diff
@@ -5,74 +5,137 @@
...
-    public class SocketAwaitableEventArgs : SocketAsyncEventArgs, ICriticalNotifyCompletion
+    public class SocketAwaitableEventArgs : SocketAsyncEventArgs, IValueTaskSource<int>
     {
-        private static readonly Action _callbackCompleted = () => { };
+        private static readonly Action<object> _callbackCompleted = _ => { };
 
         private readonly PipeScheduler _ioScheduler;
 
-        private Action _callback;
+        private Action<object> _completion;
```
halter73: I think _callback or _continuation is a better name

halter73:
If this turns out to be valuable, OutputFlowControlAwaitable, ThreadPoolAwaitable and even LibuvAwaitable could probably use this IValueTaskSource treatment. As this gets more complicated, I'm wondering if at least some of this logic could be shared.

src/Kestrel.Transport.Sockets/Internal/SocketAwaitableEventArgs.cs
```diff
@@ -5,74 +5,137 @@
...
 
-        public void Complete()
+        private void Reset()
         {
-            OnCompleted(this);
+            Volatile.Write(ref _completion, null);
         }
 
-        protected override void OnCompleted(SocketAsyncEventArgs _)
+        private ValueTask<int> CompleteSocketOperation()
```
halter73: I prefer GetCompletedValueTask() since this isn't used to complete socket operations in the async case.

src/Kestrel.Transport.Sockets/Internal/SocketAwaitableEventArgs.cs
```diff
@@ -5,74 +5,137 @@
...
+
+            // Use UserToken to carry the continuation state around
+            UserToken = state;
+            Action<object> awaitableState = Interlocked.CompareExchange(ref _completion, continuation, null);
 
-            if (SocketError != SocketError.Success)
+            if (ReferenceEquals(awaitableState, _callbackCompleted))
+            {
+#if NETCOREAPP2_1
+                ThreadPool.QueueUserWorkItem(continuation, state, preferLocal: false);
```
halter73: Why not UQUWI?

src/Kestrel.Transport.Sockets/Internal/SocketAwaitableEventArgs.cs
```diff
@@ -5,74 +5,137 @@
...
+#else
+                Task.Factory.StartNew(continuation, state);
+#endif
```
halter73: Since you null out UserToken in `OnCompleted(SocketAsyncEventArgs` it's probably best to also do that here.

src/Kestrel.Transport.Sockets/Internal/SocketAwaitableEventArgs.cs
davidfowl: https://github.com/dotnet/corefx/issues/32547

src/Kestrel.Transport.Sockets/Internal/SocketAwaitableEventArgs.cs
```diff
@@ -5,74 +5,137 @@
...
         {
-            var continuation = Interlocked.Exchange(ref _callback, _callbackCompleted);
+            int bytesTransferred = BytesTransferred;
+            SocketError error = SocketError;
 
-            if (continuation != null)
-            {
-                _ioScheduler.Schedule(state => ((Action)state)(), continuation);
-            }
+            Reset();
```
halter73: Why call Reset() here? _completion must already be null, right? Which now that I think about it means GetStatus is wrong after sync completions. Again, I'm not sure how important that is.

src/Kestrel.Transport.Sockets/Internal/SocketAwaitableEventArgs.cs
halter73: The continuation is `Action<object>`, so it's not like we're boxing.

davidfowl: Actually, you're right, I could just queue `this` instead of using the state object directly since the UserToken is a field.

src/Kestrel.Transport.Sockets/Internal/SocketAwaitableEventArgs.cs
davidfowl: I'll null it out after the work item is invoked but before calling the continuation.

src/Kestrel.Transport.Sockets/Internal/SocketAwaitableEventArgs.cs
davidfowl: Sync completions don't use this code path. Unless I'm missing something...

src/Kestrel.Transport.Sockets/Internal/SocketAwaitableEventArgs.cs
halter73: You don't have to wait until after the work item invoked to null out UserToken for the same reason you don't need to in `OnCompleted(SocketAsyncEventArgs`. Just use the local state reference which isn't null. Technically you don't need to even set UserToken in this case, you could just set it in an else block at the end.

src/Kestrel.Transport.Sockets/Internal/SocketAwaitableEventArgs.cs
halter73: Why use the UserToken field at all in this case? Just use the state object directly? UserToken is only necessary if `OnCompleted(SocketAsyncEventArgs` is responsible to running the continuation.

src/Kestrel.Transport.Sockets/Internal/SocketAwaitableEventArgs.cs
halter73: ??? This is *only* called when ReceiveAsync/SendAsync return false, right?

I guess it's confusing because of the name. That's why I recommend renaming this to GetCompletedValueTask().

src/Kestrel.Transport.Sockets/Internal/SocketAwaitableEventArgs.cs
```diff
@@ -5,74 +5,136 @@
...
-
-            if (continuation != null)
-            {
-                _ioScheduler.Schedule(state => ((Action)state)(), continuation);
-            }
+            var args = (SocketAwaitableEventArgs)state;
+            var callbackState = args.UserToken;
+            args.UserToken = null;
+            var callback = args._callback;
+            callback(callbackState);
```
halter73: I see what was up now. I always seem to forget that you can't convert an `Action<object>` to a `WaitCallback` without allocating.

src/Kestrel.Transport.Sockets/Internal/SocketAwaitableEventArgs.cs
```diff
@@ -5,74 +5,136 @@
...
+            else if (awaitableState != null)
+            {
+                throw new InvalidOperationException("Multiple continuations registered!");
+            }
 
-            _callback = null;
+        }
 
-            if (SocketError != SocketError.Success)
+        public int GetResult(short token)
```
halter73:
```
The active test run was aborted. Reason: Unhandled Exception: System.InvalidOperationException: Multiple continuations registered!
```

You still need to null out `_callback` here. It's GetCompletedValueTask() where nulling out `_callback` is unnecessary because it will always be null there. Assuming you null it out here that is.

davidfowl: 😬 ooops

src/Kestrel.Transport.Sockets/Internal/SocketAwaitableEventArgs.cs
```diff
@@ -5,74 +5,136 @@
...
         }
 
-        public void Complete()
+        private ValueTask<int> GetCompletedValueTask()
         {
-            OnCompleted(this);
+            int bytesTransferred = BytesTransferred;
+            SocketError error = SocketError;
+
+            Volatile.Write(ref _callback, null);
```
halter73: _callback should always already be null here.

halter73:
Both Jenkins Windows failures look really scary. All sockets functional tests and logs like the following:

```
16:05:36   Log Critical[ApplicationNeverCompleted]: Connection id "0HLHFPHRAPFNN" application never completed 
16:05:36   Log Critical[ApplicationNeverCompleted]: Connection id "0HLHFPHRAPFOC" application never completed 
```

In one of the runs this was logged 7 times. 😱

src/Kestrel.Transport.Sockets/Internal/SocketAwaitableEventArgs.cs
```diff
@@ -5,74 +5,136 @@
...
         public SocketAwaitableEventArgs(PipeScheduler ioScheduler)
         {
             _ioScheduler = ioScheduler;
         }
 
-        public SocketAwaitableEventArgs GetAwaiter() => this;
-        public bool IsCompleted => ReferenceEquals(_callback, _callbackCompleted);
+        protected override void OnCompleted(SocketAsyncEventArgs _)
+        {
+            var callback = Interlocked.CompareExchange(ref _callback, _callbackCompleted, null);
```
halter73: Why did you change this from Interlocked.Exchange to Interlocked.CompareExchange? You want to move into the _callbackCompleted state no matter what. Here you're only transitioning it if a callback isn't registered yet.

davidfowl: It was always using CompareExchange.

src/Kestrel.Transport.Sockets/Internal/SocketAwaitableEventArgs.cs
davidfowl: I'm going to change this back, it's a tad bit complicated and not worth saving the allocation.

davidfowl:
Looks slightly worse:

|       RPS | CPU (%) | Memory (MB) | Avg. Latency (ms) | Startup (ms) | First Request (ms) | Latency (ms) |
| --------- | ------- | ----------- | ----------------- | ------------ | ------------------ | ------------ |
| 3,163,229 |      82 |         310 |               1.8 |          187 |               42.5 |          0.7 |

VS

|       RPS | CPU (%) | Memory (MB) | Avg. Latency (ms) | Startup (ms) | First Request (ms) | Latency (ms) |
| --------- | ------- | ----------- | ----------------- | ------------ | ------------------ | ------------ |
| 3,130,355 |      73 |         307 |               2.2 |          192 |               42.1 |          0.6 |

src/Kestrel.Transport.Sockets/Internal/SocketAwaitableEventArgs.cs
halter73: No it wasn't.

https://github.com/aspnet/KestrelHttpServer/blob/081cef0934c98b21ac03871d54561c54e9d2c7e1/src/Kestrel.Transport.Sockets/Internal/SocketAwaitableEventArgs.cs#L68-L70

I think the multiple OnCompleted methods make it easy to mix this kind of stuff up.

davidfowl:
> I guess we don't take it if there's no performance improvement and possibly even a regression. It doesn't seem to really simplify things either.

Yes that's the current conclusion. I need to see if there are debugging benefits here before I close it out (also future improvements that might be coming).

Issue #2807, opened 2018-08-10
Implement MinResponseDataRate for HTTP/2


Tratcher:
Meta issue: https://github.com/aspnet/KestrelHttpServer/issues/2805

We expect this to be a flow rate for the whole collection based on data in the transport pipe, not bound to any specific stream or frame type.

PR #3003, opened 2018-10-11
Implement MinResponseDataRate for HTTP/2

halter73:
#2807

### DIFF
--- a/src/Kestrel.Core/Internal/Http2/Http2FrameWriter.cs
+++ b/src/Kestrel.Core/Internal/Http2/Http2FrameWriter.cs
@@ -35,6 +35,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http2
         private readonly OutputFlowControl _connectionOutputFlowControl;
         private readonly string _connectionId;
         private readonly IKestrelTrace _log;
+        private readonly ITimeoutControl _timeoutControl;
         private readonly TimingPipeFlusher _flusher;
 
         private bool _completed;
@@ -54,6 +55,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http2
             _connectionOutputFlowControl = connectionOutputFlowControl;
             _connectionId = connectionId;
             _log = log;
+            _timeoutControl = timeoutControl;
             _flusher = new TimingPipeFlusher(_outputWriter, timeoutControl);
             _outgoingFrame = new Http2Frame();
             _headerEncodingBuffer = new byte[_maxFrameSize];
@@ -226,7 +228,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http2
             }
         }
 
-        public Task WriteDataAsync(int streamId, StreamOutputFlowControl flowControl, ReadOnlySequence<byte> data, bool endStream)
+        public Task WriteDataAsync(int streamId, StreamOutputFlowControl flowControl, MinDataRate minRate, ReadOnlySequence<byte> data, bool endStream)
         {
             // The Length property of a ReadOnlySequence can be expensive, so we cache the value.
             var dataLength = data.Length;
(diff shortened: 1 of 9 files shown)


src/Kestrel.Core/Internal/Http2/Http2OutputProducer.cs
```diff
@@ -74,7 +74,9 @@ public void Dispose()
             }
         }
 
-        public void Abort(ConnectionAbortedException abortReason)
+        // Review: This is called when a CancellationToken fires mid-write. In HTTP/1.x, this aborts the entire connection.
+        // Should we do that here?
```
JunTaoLuo: What's the drawback of aborting the stream instead?

src/Kestrel.Core/Internal/Http2/Http2OutputProducer.cs
```diff
@@ -74,7 +74,9 @@ public void Dispose()
...
+        void IHttpOutputAborter.Abort(ConnectionAbortedException abortReason)
```
Tratcher: Reminder: On user induced cancellation call ResetAndAbort to make sure a reset is sent to the client and no graceful EndOfStream is sent. Otherwise it looks like a graceful truncation to the client which is a ship blocker.

src/Kestrel.Core/Internal/Http2/Http2FrameWriter.cs
```diff
@@ -260,7 +262,7 @@ public Task WriteDataAsync(int streamId, StreamOutputFlowControl flowControl, Re
             |                           Padding (*)                       ...
             +---------------------------------------------------------------+
         */
-        private Task WriteDataUnsynchronizedAsync(int streamId, ReadOnlySequence<byte> data, bool endStream)
+        private Task WriteDataUnsynchronizedAsync(int streamId, MinDataRate minRate, ReadOnlySequence<byte> data, long dataLength, bool endStream)
```
JunTaoLuo: Looks like in all instances where this is called, dataLength is equal to data.Length?

halter73: True, but data.Length on a ReadOnlySequence iterates, so you might as well cache the result.

JunTaoLuo: I see

test/Kestrel.InMemory.FunctionalTests/Http2/Http2TimeoutTests.cs
```diff
@@ -166,5 +169,114 @@ public async Task ResponseDrain_SlowerThanMinimumDataRate_AbortsConnection()
...
+                default:
+                    throw new NotImplementedException(finalFrameType.ToString());
+            }
+
+            await WaitForConnectionErrorAsync<Http2ConnectionErrorException>(ignoreNonGoAwayFrames: false, expectedLastStreamId: 1, Http2ErrorCode.STREAM_CLOSED,
+                CoreStrings.FormatHttp2ErrorStreamClosed(finalFrameType, 1));
+        }
+
+        [Fact]
+        public async Task DATA_Sent_TooSlowlyDueToSocketBackPressureOnSmallWrite_AbortsConnectionAfterGracePeriod()
```
JunTaoLuo: Were you planning to add any more tests?

src/Kestrel.Core/Internal/Http2/Http2OutputProducer.cs
Tratcher: Right, otherwise the client has to depend on their own timeouts.

Tratcher:
Interesting indirectly related test failure:
```
ConnectionClosedWhenBothRequestAndResponseExperienceBackPressure [FAIL]
19:40:02   Failed   ConnectionClosedWhenBothRequestAndResponseExperienceBackPressure
19:40:02   Error Message:
19:40:02    Assert.Throws() Failure
19:40:02   Expected: typeof(System.OperationCanceledException)
19:40:02   Actual:   typeof(System.InvalidOperationException): Advancing examined to the end would cause pipe to deadlock because FlushAsync is waiting.
```

test/Kestrel.InMemory.FunctionalTests/Http2/Http2TimeoutTests.cs
Tratcher: You need a flow control back pressure variant of this.

halter73: Yep. I will be adding those.

halter73:
The ConnectionClosedWhenBothRequestAndResponseExperienceBackPressure failure isn't at all related at all to these HTTP/2-only changes. It is a reincarnation of #2846.

src/Kestrel.Core/Internal/Http2/Http2OutputProducer.cs
halter73: #3007

PR #3010, opened 2018-10-12
HTTP/2 Test stabalization

Tratcher:
#3009 These tests were flaky after we added resets for incomplete request bodies. There's a race between processing the data in the last frame, reacting to the error, and setting the end-of-stream flag. The simplest solution is to force the race by not sending the end-of-stream flag when where testing error conditions.


JunTaoLuo:
Why are the linux tests failing? I haven't seen them fail in this way before.

PR #3014, opened 2018-10-15
Use dev certs for samples and update tfms

JunTaoLuo:
Noticed these while I was running samples and benchmarks.


Directory.Build.targets
```diff
@@ -1,6 +1,5 @@
 <Project>
   <PropertyGroup>
-    <RuntimeFrameworkVersion Condition=" '$(TargetFramework)' == 'netcoreapp2.0' ">$(MicrosoftNETCoreApp20PackageVersion)</RuntimeFrameworkVersion>
```
JunTaoLuo: Not sure why we didn't update the TFM of a few apps along with the rest of the repo. I tried to remove all targets of netcoreapp2.0. We still need netcoreapp2.1 though.

benchmarks/Kestrel.Performance/Kestrel.Performance.csproj
```diff
@@ -4,7 +4,7 @@
     <!-- Using shorter assembly name instead of Microsoft.AspNetCore.Server.Kestrel.Performance because https://github.com/dotnet/BenchmarkDotNet/issues/498 -->
     <AssemblyName>Kestrel.Performance</AssemblyName>
     <RootNamespace>Microsoft.AspNetCore.Server.Kestrel.Performance</RootNamespace>
-    <TargetFramework>netcoreapp2.0</TargetFramework>
+    <TargetFramework>netcoreapp2.2</TargetFramework>
```
Tratcher: @sebastienros @davidfowl was there a reason these weren't updated? Something about running the benchmarks on old framework infrastructure?

Issue #2701, opened 2018-07-06
Flaky test: CanReadAndWriteWithHttpsConnectionAdapter


ryanbrandenburg:
This test [fails](http://aspnetci/viewLog.html?buildId=498343&buildTypeId=XPlat_Windows_Win10Universe) occasionally with the following error:
```
System.IO.IOException : Unable to read data from the transport connection: The I/O operation has been aborted because of either a thread exit or an application request..
---- System.Net.Sockets.SocketException : The I/O operation has been aborted because of either a thread exit or an application request.
   at System.Net.Sockets.Socket.AwaitableSocketAsyncEventArgs.ThrowException(SocketError error)
   at System.Net.Sockets.Socket.AwaitableSocketAsyncEventArgs.GetResult(Int16 token)
   at System.Net.Security.SslStreamInternal.<FillBufferAsync>g__InternalFillBufferAsync|38_0[TReadAdapter](TReadAdapter adap, ValueTask`1 task, Int32 min, Int32 initial)
   at System.Net.Security.SslStreamInternal.ReadAsyncInternal[TReadAdapter](TReadAdapter adapter, Memory`1 buffer)
   at System.IO.StreamReader.ReadBufferAsync()
   at System.IO.StreamReader.ReadToEndAsyncInternal()
   at Microsoft.AspNetCore.Testing.HttpClientSlim.ReadResponse(Stream stream)
   at Microsoft.AspNetCore.Testing.HttpClientSlim.PostAsync(Uri requestUri, HttpContent content, Boolean validateCertificate)
   at Microsoft.AspNetCore.Testing.HttpClientSlim.PostAsync(String requestUri, HttpContent content, Boolean validateCertificate)
   at Microsoft.AspNetCore.Server.Kestrel.FunctionalTests.HttpsConnectionAdapterTests.CanReadAndWriteWithHttpsConnectionAdapter() in /_/test/Kestrel.FunctionalTests/HttpsConnectionAdapterTests.cs:line 47
--- End of stack trace from previous location where exception was thrown ---
----- Inner Stack Trace -----
------- Stdout: -------
| [0.002s] TestLifetime Information: Starting test CanReadAndWriteWithHttpsConnectionAdapter at 2018-07-06T03:07:01
| [0.010s] Microsoft.AspNetCore.Hosting.Internal.WebHost Debug: Hosting starting
| [0.011s] Microsoft.AspNetCore.Hosting.Internal.WebHost Debug: Hosting started
[...]

PR #3016, opened 2018-10-16
Skip falky tests caused by https://github.com/dotnet/corefx/issues/30691

JunTaoLuo:
Addresses most of the tests flagged by https://github.com/aspnet/KestrelHttpServer/issues/2777 and https://github.com/aspnet/KestrelHttpServer/issues/2701. We'll start retrying these tests until the corefx issue is fixed.

Side note: Originally I was planning on adding a feature to the RetryTest attribute to limit this to Win10 since that's the only version of windows these tests are failing on but I ran into implementation issues in https://github.com/aspnet/Logging/tree/johluo/match-os-version where my Win10 machine was giving me a Win8 version prefix.

### DIFF
--- a/test/Kestrel.Transport.FunctionalTests/ResponseTests.cs
+++ b/test/Kestrel.Transport.FunctionalTests/ResponseTests.cs
@@ -17,12 +17,13 @@ using Microsoft.AspNetCore.Hosting;
 using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.Http.Features;
 using Microsoft.AspNetCore.Server.Kestrel.Core;
-using Microsoft.AspNetCore.Server.Kestrel.Core.Adapter.Internal;
 using Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Infrastructure;
 using Microsoft.AspNetCore.Server.Kestrel.Https;
 using Microsoft.AspNetCore.Server.Kestrel.Https.Internal;
 using Microsoft.AspNetCore.Testing;
+using Microsoft.AspNetCore.Testing.xunit;
 using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Logging.Testing;
 using Microsoft.Extensions.Primitives;
 using Moq;
 using Xunit;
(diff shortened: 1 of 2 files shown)


test/Kestrel.Transport.FunctionalTests/ResponseTests.cs
```diff
@@ -802,7 +803,14 @@ async Task App(HttpContext context)
...
         }
 
+        private bool ConnectionNotClosedWhenClientSatisfiesMinimumDataRateGivenLargeResponseHeadersRetryPredicate(Exception e)
+            => e is IOException && e.Message.Contains("Unable to read data from the transport connection: The I/O operation has been aborted because of either a thread exit or an application request");
+
         [Fact]
+        [RetryTest(nameof(ConnectionNotClosedWhenClientSatisfiesMinimumDataRateGivenLargeResponseHeadersRetryPredicate),
+            "Active investigation into potential corefx sockets bug: https://github.com/dotnet/corefx/issues/30691",
+            OperatingSystems.Windows,
+            5)]
```
JunTaoLuo: I'm increasing the retry limit here since when the tests fail, they fail quickly and it may take up to 3 runs total to pass. I'm setting it to 5 which should be a safer limit.

This is beginning to seem like a bad state somewhere not being cleared promptly.

test/Kestrel.Transport.FunctionalTests/ResponseTests.cs
```diff
@@ -802,7 +803,14 @@ async Task App(HttpContext context)
             Assert.False(requestAborted);
         }
 
+        private bool ConnectionNotClosedWhenClientSatisfiesMinimumDataRateGivenLargeResponseHeadersRetryPredicate(Exception e)
+            => e is IOException && e.Message.Contains("Unable to read data from the transport connection: The I/O operation has been aborted because of either a thread exit or an application request");
```
halter73: Should this be made static? Can we share predicates?

halter73:
Any plans to backport this to 2.1? With so many more functional tests, there are a lot more that need retrying there.

test/Kestrel.Transport.FunctionalTests/ResponseTests.cs
JunTaoLuo: It wouldn't make it that much better. Currently, we look for the predicate in the type that contains the actual type. I could move this to a base class or add a call to a static utility method elsewhere but I'm not sure it's worth it for just two tests.

JunTaoLuo:
We'll see about backporting to 2.1 since it will require the retry attribute in logging as well.

JunTaoLuo:
Ugh... I didn't squash.

PR #3023, opened 2018-10-17
Fix test flakiness caused by not reading in a loop

### LINKED ISSUE #3021: Test failure: RequestBodyReadAsyncCanBeCancelled
This issue was made automatically. If there is a problem contact ryanbrandenburg.

This test [failed](https://dnceng.visualstudio.com/9ee6d478-d288-47f7-aacc-f6e6d082ae6d/_build/results?buildId=31136) with the following error:

```
Assert.Equal() Failure
           ↓ (pos 1)
Expected: Hello 
Actual:   H
           ↑ (pos 1)
```

Other tests within that build may have failed with a similar message, but they are not listed here. Check the link above for more info.

This test failed on release/2.2.

CC muratg

halter73:
Fixes #3021

### DIFF
--- a/test/Kestrel.InMemory.FunctionalTests/RequestTests.cs
+++ b/test/Kestrel.InMemory.FunctionalTests/RequestTests.cs
@@ -67,10 +67,9 @@ namespace Microsoft.AspNetCore.Server.Kestrel.InMemory.FunctionalTests
                 var buffer = new byte[1024];
                 try
                 {
+                    await context.Request.Body.ReadUntilLengthAsync(buffer, 6, cts.Token).DefaultTimeout();
 
-                    int read = await context.Request.Body.ReadAsync(buffer, 0, buffer.Length, cts.Token);
-
-                    Assert.Equal("Hello ", Encoding.UTF8.GetString(buffer, 0, read));
+                    Assert.Equal("Hello ", Encoding.ASCII.GetString(buffer, 0, 6));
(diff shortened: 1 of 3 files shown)


test/Kestrel.InMemory.FunctionalTests/RequestTests.cs
```diff
@@ -67,10 +67,9 @@ public async Task RequestBodyReadAsyncCanBeCancelled()
                 var buffer = new byte[1024];
                 try
                 {
+                    await context.Request.Body.ReadUntilLengthAsync(buffer, 6, cts.Token).DefaultTimeout();
 
-                    int read = await context.Request.Body.ReadAsync(buffer, 0, buffer.Length, cts.Token);
-
-                    Assert.Equal("Hello ", Encoding.UTF8.GetString(buffer, 0, read));
+                    Assert.Equal("Hello ", Encoding.ASCII.GetString(buffer, 0, 6));
```
halter73: Only this change and the change to TraceIdentifierIsUnique() fix any real test bugs. The rest is minor test simplification.

The only other place I see where we read the request body without a loop is in Http2Stream/ConnectionTests. There's a lot of that there, but I think that's OK as long as we don't start adding data to the request body pipe for partial frames.

JunTaoLuo:
TraceIdentifierIsUnique looks broken.

Issue #3006, opened 2018-10-12
Both reader and writer has to be completed to be able to reset the pipe


halter73:
I'm seeing "Both reader and writer has to be completed to be able to reset the pipe" in some HTTP/2 tests where app doesn't fully consume request body. I've noticed this in AppDoesNotReadRequestBody_ResetsAndDrainsRequest and DATA_Received_ButNotConsumedByApp_CountsTowardsInputFlowControl.

```
[0.003s] [TestLifetime] [Information] Starting test AppDoesNotReadRequestBody_ResetsAndDrainsRequest-CONTINUATION at 2018-10-12T17:51:38
[0.015s] [Microsoft.AspNetCore.Server.Kestrel] [Verbose] Connection id "null" sending SETTINGS frame for stream ID 0 with length 18 and flags NONE
[0.015s] [Microsoft.AspNetCore.Server.Kestrel] [Verbose] Connection id "null" sending WINDOW_UPDATE frame for stream ID 0 with length 4 and flags 0x0
[0.016s] [Microsoft.AspNetCore.Server.Kestrel] [Verbose] Connection id "null" received SETTINGS frame for stream ID 0 with length 0 and flags NONE
[0.053s] [Microsoft.AspNetCore.Server.Kestrel] [Verbose] Connection id "null" sending SETTINGS frame for stream ID 0 with length 0 and flags ACK
[0.054s] [Microsoft.AspNetCore.Server.Kestrel] [Verbose] Connection id "null" received HEADERS frame for stream ID 1 with length 37 and flags END_HEADERS
[0.056s] [Microsoft.AspNetCore.Server.Kestrel] [Verbose] Connection id "null" sending HEADERS frame for stream ID 1 with length 55 and flags END_HEADERS
[0.056s] [Microsoft.AspNetCore.Server.Kestrel] [Verbose] Connection id "null" sending DATA frame for stream ID 1 with length 0 and flags END_STREAM
[0.072s] [Microsoft.AspNetCore.Server.Kestrel] [Warning] Connection processing ended abnormally.
System.InvalidOperationException: Both reader and writer has to be completed to be able to reset the pipe.
   at System.IO.Pipelines.ThrowHelper.ThrowInvalidOperationException_ResetIncompleteReaderWriter()
   at System.IO.Pipelines.Pipe.Reset()
[...]

PR #3028, opened 2018-10-17
Only reset the request body pipe for HTTP/1 

Tratcher:
#3006 This was causing an internal error due to the ordering of how we shut down the HTTP/2 streams. Resetting the pipes does not matter for HTTP/2 since we do not pool streams (yet).

I removed one test (StopAsyncPreventsFurtherDataConsumption) as it did not reflect how we were actually using StopAsync in HttpProtocol.

### DIFF
--- a/src/Kestrel.Core/Internal/Http/Http1MessageBody.cs
+++ b/src/Kestrel.Core/Internal/Http/Http1MessageBody.cs
@@ -137,9 +137,25 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
                 return Task.CompletedTask;
             }
 
+            // PumpTask catches all Exceptions internally.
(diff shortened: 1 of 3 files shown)


src/Kestrel.Core/Internal/Http/Http1MessageBody.cs
```diff
@@ -130,16 +130,19 @@ private async Task PumpAsync()
...
             if (!_context.HasStartedConsumingRequestBody)
             {
-                return Task.CompletedTask;
+                return;
             }
 
             _canceled = true;
             _context.Input.CancelPendingRead();
-            return _pumpTask;
+            await _pumpTask;
```
halter73: Let's use the two-method awaited pattern to avoid an extra async state machine in the normal case.

Tratcher:
Updated

PR #3030, opened 2018-10-18
Don't warn about SemVer

ryanbrandenburg:


Directory.Build.props
```diff
@@ -20,5 +20,6 @@
     <SignAssembly>true</SignAssembly>
     <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
     <LangVersion>latest</LangVersion>
+    <NoWarn>$(NoWarn);NU5105</NoWarn>
```
natemcmaster: Which project is producing this error? This warning should already be suppressed by Internal.AspNetCore.Sdk.

https://github.com/aspnet/BuildTools/blob/release/2.2/src/Internal.AspNetCore.Sdk/build/Common.targets#L15

ryanbrandenburg: [The error message](https://dnceng.visualstudio.com/public/_build/results?buildId=31698&view=logs) does not make clear which project is at fault, I'm just implementing [your suggestion](https://github.com/aspnet/AspNetCore-Internal/issues/1156#issuecomment-430800474) [as requested](https://github.com/aspnet/AspNetCore-Internal/issues/1156#issuecomment-431089293).

natemcmaster: > D:\a\1\s\.dotnet\x64\sdk\2.2.100-preview2-009404\Sdks\NuGet.Build.Tasks.Pack\build\NuGet.Build.Tasks.Pack.targets(199,5): error NU5105: The package version '3.0.0-alpha1-20181017.2' uses SemVer 2.0.0 or components of SemVer 1.0.0 that are not supported on legacy clients. Change the package version to a SemVer 1.0.0 string. If the version contains a release label it must start with a letter. This message can be ignored if the package is not intended for older clients. [D:\a\.dotnet\buildtools\korebuild\3.0.0-alpha1-20181011.3\modules\sharedsources\sharedsources.csproj]


Looks like we need to make a change to https://github.com/aspnet/BuildTools/blob/release/2.2/files/KoreBuild/modules/sharedsources/sharedsources.csproj

ryanbrandenburg: Is that really where the problem is? I don't see any NU5105 errors in the [BuildTools build](https://dnceng.visualstudio.com/public/_build/results?buildId=31810&view=logs). I don't think I really follow the thread of what's going on here and how you would like to fix it. Could you go into more detail?

ryanbrandenburg:
Closing this for now until we have additional clarity around what the real problem is here.

Issue #3007, opened 2018-10-12
RST HTTP/2 stream when a CancellationToken fires mid-write


halter73:
See https://github.com/aspnet/KestrelHttpServer/pull/3003/files#r224629533

PR #3032, opened 2018-10-18
Send RST for canceled HTTP/2 writes

Tratcher:
#3007 There are two scenarios:
1) The cancellation token is tripped before the write actually starts. WriteAsync will throw but no state changes take place. The app can handle the exception or let Kestrel handle it like any other app exception.
2) The cancellation token is tripped after the write starts. In this case the response is aborted and a RST is sent. Without the RST no response would be sent the client would not know anything was wrong until they eventually timed out.

Logs:
```
| [0.650s] Microsoft.AspNetCore.Server.Kestrel Trace: Connection id "(null)" sending HEADERS frame for stream ID 1 with length 37 and flags END_HEADERS
| [0.684s] Microsoft.AspNetCore.Server.Kestrel Debug: Trace id ":00000001": HTTP/2 stream error "INTERNAL_ERROR". A Reset is being sent to the stream.
| Microsoft.AspNetCore.Connections.ConnectionAbortedException: The connection or stream was aborted because a write operation was aborted with a CancellationToken. ---> System.OperationCanceledException: The operation was canceled.
|    at System.Threading.CancellationToken.ThrowOperationCanceledException()
|    at System.IO.Pipelines.PipeAwaitable.ObserveCancelation()
|    at System.IO.Pipelines.Pipe.GetFlushResult(FlushResult& result)
|    at System.IO.Pipelines.Pipe.GetFlushAsyncResult()
|    at System.IO.Pipelines.Pipe.DefaultPipeWriter.GetResult(Int16 token)
|    at System.Threading.Tasks.ValueTask`1.ValueTaskSourceAsTask.<>c.<.cctor>b__4_0(Object state)
| --- End of stack trace from previous location where exception was thrown ---
|    at Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Infrastructure.TimingPipeFlusher.TimeFlushAsync(MinDataRate minRate, Int64 count, IHttpOutputAborter outputAborter, CancellationToken cancellationToken) in d:\github\AspNet\KestrelHttpServer\src\Kestrel.Core\Internal\Infrastructure\TimingPipeFlusher.cs:line 82
|    --- End of inner exception stack trace ---
| [0.730s] Microsoft.AspNetCore.Server.Kestrel Trace: Connection id "(null)" sending RST_STREAM frame for stream ID 1 with length 4 and flags 0x0
```

### DIFF
--- a/src/Kestrel.Core/Internal/Http2/Http2OutputProducer.cs
+++ b/src/Kestrel.Core/Internal/Http2/Http2OutputProducer.cs
@@ -74,10 +74,11 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http2
             }
         }
 
-        // Review: This is called when a CancellationToken fires mid-write. In HTTP/1.x, this aborts the entire connection.
-        // Should we do that here?
+        // This is called when a CancellationToken fires mid-write. In HTTP/1.x, this aborts the entire connection.
+        // For HTTP/2 we abort the stream.
         void IHttpOutputAborter.Abort(ConnectionAbortedException abortReason)
         {
+            _stream.ResetAndAbort(abortReason, Http2ErrorCode.INTERNAL_ERROR);
             Dispose();
         }
 
--- a/src/Kestrel.Core/Internal/Http2/Http2Stream.cs
+++ b/src/Kestrel.Core/Internal/Http2/Http2Stream.cs
@@ -427,7 +427,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http2
             ResetAndAbort(abortReason, Http2ErrorCode.INTERNAL_ERROR);
         }
 
-        private void ResetAndAbort(ConnectionAbortedException abortReason, Http2ErrorCode error)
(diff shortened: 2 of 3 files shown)


src/Kestrel.Core/Internal/Http2/Http2OutputProducer.cs
```diff
@@ -74,10 +74,11 @@ public void Dispose()
             }
         }
 
-        // Review: This is called when a CancellationToken fires mid-write. In HTTP/1.x, this aborts the entire connection.
-        // Should we do that here?
+        // This is called when a CancellationToken fires mid-write. In HTTP/1.x, this aborts the entire connection.
+        // For HTTP/2 we abort the stream.
         void IHttpOutputAborter.Abort(ConnectionAbortedException abortReason)
         {
+            _stream.ResetAndAbort(abortReason, Http2ErrorCode.INTERNAL_ERROR);
```
halter73: You add a stream back reference for trailers, and all of the sudden we have tight coupling! 😆

Tratcher: But so convenient!

Issue #3031, opened 2018-10-18
Response Headers are blocked by flow control


Tratcher:
They get written into the buffer, but they wait for the first data frame to flush. If the data frames are blocked due to the flow control window then the headers do not get sent until the next window update is received.

Tratcher:
Repro:
```
        [Fact]
        public async Task WriteAsync_CancellationTokenTriggeredDueToFlowControl_SendRST()
        {
            var cts = new CancellationTokenSource();
            var headers = new[]
            {
                new KeyValuePair<string, string>(HeaderNames.Method, "GET"),
                new KeyValuePair<string, string>(HeaderNames.Path, "/"),
                new KeyValuePair<string, string>(HeaderNames.Scheme, "http"),
            };
            await InitializeConnectionAsync(async context =>
            {
                await Assert.ThrowsAsync<OperationCanceledException>(() => context.Response.WriteAsync("hello,", cts.Token));
            });

            _clientSettings.InitialWindowSize = 0;
            await SendSettingsAsync();
            await ExpectAsync(Http2FrameType.SETTINGS,
                withLength: 0,
                withFlags: (byte)Http2SettingsFrameFlags.ACK,
                withStreamId: 0);

            await StartStreamAsync(1, headers, endStream: true);

            var headersFrame = await ExpectAsync(Http2FrameType.HEADERS,
                withLength: 55,
                withFlags: (byte)Http2HeadersFrameFlags.END_HEADERS,
                withStreamId: 1);

            cts.Cancel();

            await WaitForStreamErrorAsync(1, Http2ErrorCode.INTERNAL_ERROR, null);

            await StopConnectionAsync(expectedLastStreamId: 1, ignoreNonGoAwayFrames: false);

            _hpackDecoder.Decode(headersFrame.PayloadSequence, endHeaders: false, handler: this);

            Assert.Equal(2, _decodedHeaders.Count);
            Assert.Contains("date", _decodedHeaders.Keys, StringComparer.OrdinalIgnoreCase);
            Assert.Equal("200", _decodedHeaders[HeaderNames.Status]);
        }
```
This times out waiting for the response headers.

Tratcher:
Reverted in https://github.com/aspnet/KestrelHttpServer/pull/3048 due to a major perf [regression](https://github.com/aspnet/Benchmarks/issues/541). Revisit in 3.0.

PR #3035, opened 2018-10-19
Flush HTTP/2 response headers

Tratcher:
#3031 Today they're only flushed by the first data frame, but if there's flow control back-pressure then they might not get flushed promptly. Flush them independently.

This is likely to case a small perf degradation due to more TCP packets getting sent.

### DIFF
--- a/src/Kestrel.Core/Internal/Http2/Http2FrameWriter.cs
+++ b/src/Kestrel.Core/Internal/Http2/Http2FrameWriter.cs
@@ -104,19 +104,6 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http2
             }
         }
 
-        public Task FlushAsync(IHttpOutputAborter outputAborter, CancellationToken cancellationToken)
-        {
-            lock (_writeLock)
-            {
-                if (_completed)
-                {
-                    return Task.CompletedTask;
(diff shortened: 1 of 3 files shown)


Tratcher:
This change seems to make these tests flaky:
DATA_Sent_TooSlowlyDueToSocketBackPressureOnLargeWrite_AbortsConnectionAfterRateTimeout
DATA_Sent_TooSlowlyDueToSocketBackPressureOnSmallWrite_AbortsConnectionAfterGracePeriod

src/Kestrel.Core/Internal/Http2/Http2FrameWriter.cs
```diff
@@ -171,9 +158,11 @@ public void WriteResponseHeaders(int streamId, int statusCode, IHeaderDictionary
                     throw new InvalidOperationException(hex.Message, hex); // Report the error to the user if this was the first write.
                 }
             }
+
+            _flusher.FlushAsync().GetAwaiter().GetResult();
```
halter73: Just fire and forget. The next flush will wait on backpressure if necessary.

Tratcher:
Updated. That made the tests happier too.

Tratcher:
@sebastienros FYI

Issue #2808, opened 2018-08-10
Implement MinRequestBodyDataRate for HTTP/2


Tratcher:
Meta issue: https://github.com/aspnet/KestrelHttpServer/issues/2805

Needs design.

PR #3051, opened 2018-10-24
 Add HTTP/2 request body data rate limit

halter73:
Addresses #2808

### DIFF
--- a/src/Kestrel.Core/Internal/Http/MessageBody.cs
+++ b/src/Kestrel.Core/Internal/Http/MessageBody.cs
@@ -4,6 +4,7 @@
 using System;
 using System.Buffers;
 using System.IO;
+using System.IO.Pipelines;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Infrastructure;
@@ -19,6 +20,8 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
 
         private bool _send100Continue = true;
         private long _consumedBytes;
+        private bool _timingReads;
+        private bool _backpressure;
 
         protected MessageBody(HttpProtocol context)
         {
@@ -43,7 +46,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
 
             while (true)
             {
-                var result = await _context.RequestBodyPipe.Reader.ReadAsync(cancellationToken);
+                var result = await TimedReadAsync(cancellationToken);
                 var readableBuffer = result.Buffer;
                 var consumed = readableBuffer.End;
                 var actual = 0;
@@ -68,11 +71,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
                 }
                 finally
                 {
-                    _context.RequestBodyPipe.Reader.AdvanceTo(consumed);
-
-                    // Update the flow-control window after advancing the pipe reader, so we don't risk overfilling
-                    // the pipe despite the client being well-behaved.
-                    OnDataRead(actual);
+                    CompleteTimedRead(consumed, actual);
                 }
             }
         }
@@ -83,7 +82,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
 
             while (true)
             {
-                var result = await _context.RequestBodyPipe.Reader.ReadAsync(cancellationToken);
+                var result = await TimedReadAsync(cancellationToken);
                 var readableBuffer = result.Buffer;
                 var consumed = readableBuffer.End;
                 var bytesRead = 0;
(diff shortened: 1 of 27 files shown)


Tratcher:
Can we get a high level summary? This is pretty complex.

halter73:
@Tratcher Sure.

This moves the request body read timing logic from Http1MessageBody.PumpAsync to the shared HttpMessageBody.ReadAsync/CopyToAsync.

The old PumpAsync code had inverted logic where it paused the timer when waiting on the application in PipeWriter.FlushAsync. The new logic runs the timer when waiting on data from the other side in PipeReader.ReadAsync.

The logic, while inverted, is functionally the same. In both cases, any tick where any time is spent awaiting a read, is counted as a full tick awaiting a read due to the broad granularity of the timer.

I renamed some of the ITimeoutControl methods to reflect this inverted logic and the fact that there now needs to be an explicit call after starting the request body reads to time said reads.

The TimeoutControl implementation needed some adaptation to support HTTP/2. This includes:

1. Supporting multiple concurrent timed reads that only need meat the min rate limit in aggregate.
2. Disabling the timer when the connection-level HTTP/2 input flow control window is small enough that we can deduce it's the server, not the client, that's responsible for limiting the request data rate.
3. Like before, the rate limit is enforced cumulatively for the entire request body. If there are multiple request bodies being read at once, this cumulative tracking is only reset once there a no request bodies being read at a point in time.

Since the rate limit is enforced in aggregate across all the streams in HTTP/2, any modifications to the per-request rate limit via the HttpContext are ignored for HTTP/2 requests.

Tratcher:
Should we take the IHttpMinRequestBodyDataRateFeature and IHttpMinResponseDataRateFeature off HttpProtocol and only implement them for HTTP/1?

src/Kestrel.Core/Internal/Http/MessageBody.cs
```diff
@@ -43,17 +48,27 @@ public virtual async ValueTask<int> ReadAsync(Memory<byte> buffer, CancellationT
...
                     {
-                        // buffer.Count is int
-                        actual = (int)Math.Min(readableBuffer.Length, buffer.Length);
+                        // buffer.Length is int
+                        actual = (int)Math.Min(readableBufferLength, buffer.Length);
+
+                        // Make sure we don't double-count bytes on the next read.
+                        if (readableBufferLength > actual)
+                        {
+                            _alreadyTimedBytes = readableBufferLength - actual;
```
Tratcher: Can you get rid of _alreadyTimedBytes by calculating actual before StopTimingRead?

halter73:
> Should we take the IHttpMinRequestBodyDataRateFeature and IHttpMinResponseDataRateFeature off HttpProtocol and only implement them for HTTP/1?

Seems reasonable. We already do this for IHttpUpgradeFeature. Technically it's a breaking change, even if it's not at the API level. At least the features are already absent on HttpSysServer, so we wouldn't break any apps that work cross-server.

src/Kestrel.Core/Internal/Http/MessageBody.cs
halter73: We could, but I think it's to give credit to the client for sending data as soon as possible. Otherwise, if the app uses really small buffers and processes the data slowly once it arrives, we could unfairly penalize high-latency connections.

src/Kestrel.Core/Internal/Http/MessageBody.cs
```diff
@@ -118,11 +131,12 @@ public virtual async Task CopyToAsync(Stream destination, CancellationToken canc
                 }
                 finally
                 {
-                    _context.RequestBodyPipe.Reader.AdvanceTo(consumed);
+
```
JunTaoLuo: nit: extra line

src/Kestrel.Core/Internal/Http2/FlowControl/InputFlowControl.cs
```diff
@@ -90,10 +92,7 @@ public bool TryUpdateWindow(int bytes, out int updateSize)
 
         public void StopWindowUpdates()
         {
-            lock (_flowLock)
-            {
-                _windowUpdatesDisabled = true;
-            }
+            _windowUpdatesDisabled = true;
```
JunTaoLuo: Why are we removing the lock here?

halter73: It didn't do anything

halter73: To be more specific, this is only used to stop sending window updates to the client after the associated stream or connection is aborted which is a terminal state. If we happen to send an unneeded window update it's not a big deal. This is simply an optimization.

This is only checked in one place, and doesn't need to be synchronized with any other state. _flow is updated whether or not window updates are disabled. I don't like unnecessary code that looks like it might be important but isn't.

Issue #3055, opened 2018-10-25
Use Chrome instead of Chromium on Ubuntu


JunTaoLuo:
I'm seeing failures on the VSTS test runs. https://dnceng.visualstudio.com/public/_build/results?buildId=35319&view=logs I think it's because there's no Chromium installed. I'll need to check the agents for whether Chrome is installed instead. We should be using Chromium instead of Chrome on linux.

cc @anurse

BrennanConroy:
We use chrome headless
```
Checking for ChromeHeadlessNoSandbox at /usr/bin/google-chrome...
2018-10-25T17:46:05.1832199Z   Located ChromeHeadlessNoSandbox at /usr/bin/google-chrome.
2018-10-25T17:46:05.1836971Z   Checking for ChromiumHeadless at undefined...
2018-10-25T17:46:05.1837332Z   Unable to locate ChromiumHeadless. Skipping.
2018-10-25T17:46:05.1841769Z   Checking for FirefoxHeadless at firefox...
2018-10-25T17:46:05.1847490Z   Located FirefoxHeadless at firefox.
2018-10-25T17:46:05.9776421Z   25 10 2018 17:46:05.976:INFO [karma]: Writing browser console to file: /home/vsts/work/1/s/artifacts/logs/browserlogs.console.2018-10-25T17-46-05-710Z
2018-10-25T17:46:06.0458241Z   25 10 2018 17:46:06.045:INFO [karma]: Karma v3.0.0 server started at http://0.0.0.0:9876/
2018-10-25T17:46:06.0461222Z   25 10 2018 17:46:06.045:INFO [launcher]: Launching browsers ChromeHeadlessNoSandbox, FirefoxHeadless with unlimited concurrency
2018-10-25T17:46:06.0502751Z   25 10 2018 17:46:06.049:INFO [launcher]: Starting browser ChromeHeadless
2018-10-25T17:46:06.0758674Z   25 10 2018 17:46:06.075:INFO [launcher]: Starting browser FirefoxHeadless
```

JunTaoLuo:
Spoke to @anurse and it turns out we don't have a good reason to have opted to install chromium on ubuntu instead of chrome. I'll just install chrome on our agents and switch the tests to use chrome instead of chromium

JunTaoLuo:
I've updated all of our ubuntu agents to have chrome installed.

PR #3056, opened 2018-10-25
Use chrome instead of chromium on Ubuntu

JunTaoLuo:
Addresses #3055


JunTaoLuo:
It's the only browser interop tests we have for HTTP/2 so we definitely want these to fail if chrome is not installed. Note that the tests are marked with a special attribute to fail on CI runs only. If the tests are run in a dev environment and chrome is not installed, we silently skip the tests.

JunTaoLuo:
Tested on the ubuntu aspnetci agents which should be the last place that doesn't have chrome installed and the tests ran successfully so I'm going to merge this.

ryanbrandenburg:
Ah, I misread the conditional. Good show then, carry on.

PR #3064, opened 2018-10-29
Add more chrome test logging and fix an issue with tests running as root on VSTS

JunTaoLuo:
Tests on VSTS run in root and on ubuntu that requires a --no-sandbox flag to be passed in. This addresses https://github.com/aspnet/AspNetCore/issues/3727.


JunTaoLuo:
ping @Tratcher @halter73

PR #3068, opened 2018-10-31
Make parser "safe"

JeremyKuhne:
@pakrym, @davidfowl This is the simple "convert to span" change. I'm porting the changes I made in CoreFxLab on top of this.


src/Kestrel.Core/Internal/Infrastructure/HttpUtilities.cs
```diff
@@ -549,5 +516,14 @@ private sealed class UTF8EncodingSealed : UTF8Encoding
 
             public override byte[] GetPreamble() => Array.Empty<byte>();
         }
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        internal unsafe static Span<byte> UnsafeAsSpan(this ref ReadOnlySpan<byte> span)
```
davidfowl: Just change the parser interface to take ROSpan

JeremyKuhne: I can, but I'll have to push this lower. Eventually the span gets rewritten in place. :/

JeremyKuhne:
Added the core of the BufferReader struct changes. Working on the actual meat of the parser changes now.

src/Kestrel.Core/Internal/Infrastructure/HttpUtilities.cs
```diff
@@ -549,5 +516,14 @@ private sealed class UTF8EncodingSealed : UTF8Encoding
...
+        {
+            fixed (byte* b = span)
+            {
+                return new Span<byte>(b, span.Length);
```
benaadams: Is this valid? The returned Span is now pointing at an untracked pointer and the data under it could be moved? (Especially for slow `Span<T>` as it would now have no rooting `object`)

/cc @jkotas

Slightly ironically you could do this more safely without `fixed` on .NET Core
```csharp
internal static Span<byte> UnsafeAsSpan(this ReadOnlySpan<byte> span)
{
    return MemoryMarshal.CreateSpan(ref MemoryMarshal.GetReference(span), span.Length);
}
```
But that doesn't help with the .NET Standard/slow Span

JeremyKuhne: It isn't. I'm going to move this down to where this actually gets modified.

JeremyKuhne:
I've changed the request line parser. As part of this change I'm changing tests to no longer expect carriage returns and line feeds in error messages. I presume this is ok?

src/Kestrel.Core/Internal/Infrastructure/HttpUtilities.cs
davidfowl: Yes this is unsafe 😄

src/Kestrel.Core/Internal/Http/Http1MessageBody.cs
```diff
@@ -362,7 +362,7 @@ private class ForChunkedEncoding : Http1MessageBody
             // byte consts don't have a data type annotation so we pre-cast it
             private const byte ByteCR = (byte)'\r';
             // "7FFFFFFF\r\n" is the largest chunk size that could be returned as an int.
-            private const int MaxChunkPrefixBytes = 10;
+            private const long MaxChunkPrefixBytes = 10;
```
davidfowl: Long?

JeremyKuhne: The reader's consumed is now `long`. I can keep this as `int` if you like.

src/Kestrel.Core/Internal/Infrastructure/HttpUtilities.cs
JeremyKuhne: Changed.

src/Kestrel.Core/Internal/Infrastructure/HttpUtilities.cs
JeremyKuhne: It's gone. :)

JeremyKuhne:
Okay, the header parsing is there too. Again, this change makes the detail reported on errors end before any CR/LF. It seems like that is the right answer, it would be difficult/costly to pass the context along to include up to the first LF as before.

The next step is to call straight into the methods that take the reader. That has a significant impact when parsing the entire request.

JeremyKuhne:
Fyi, I'm using this as a sanity check for getting `SequenceReader` into CoreFX.

src/Kestrel.Core/Internal/Http/HttpParser.cs
```diff
@@ -31,445 +32,285 @@ public HttpParser(bool showErrorDetails)
...
-                }
-                else if (ch == BytePercentage)
-                {
-                    if (pathStart == -1)
-                    {
-                        // Path starting with % is illegal
-                        RejectRequestLine(data, length);
-                    }
+            // Skip the space after the method
+            ReadOnlySpan<byte> target = data.Slice(++offset);
```
benaadams: Split this? Pre-increment can be confusing
```csharp
// Skip the space after the method
offset++;
ReadOnlySpan<byte> target = data.Slice(offset);
```

JeremyKuhne: Sure

JeremyKuhne:
Ok, I've updated the benchmarks to use the overloads that take the reader and done the initial microbenchmarks.

**Before**

```
               Method |     Mean |    Error |   StdDev |        Op/s | Scaled | ScaledSD | Allocated |
--------------------- |---------:|---------:|---------:|------------:|-------:|---------:|----------:|
 PlaintextTechEmpower | 266.6 ns | 1.689 ns | 1.580 ns | 3,751,042.0 |   1.00 |     0.00 |       0 B |
           LiveAspNet | 464.1 ns | 1.390 ns | 1.300 ns | 2,154,713.7 |   1.74 |     0.01 |       0 B |
              Unicode | 674.9 ns | 2.223 ns | 2.079 ns | 1,481,655.4 |   2.53 |     0.02 |       0 B |
```

**After**

```
               Method |     Mean |     Error |    StdDev |        Op/s | Scaled | ScaledSD | Allocated |
--------------------- |---------:|----------:|----------:|------------:|-------:|---------:|----------:|
 PlaintextTechEmpower | 178.0 ns | 1.3724 ns | 1.2837 ns | 5,617,275.3 |   1.00 |     0.00 |       0 B |
           LiveAspNet | 352.3 ns | 0.9841 ns | 0.8218 ns | 2,838,783.7 |   1.98 |     0.01 |       0 B |
              Unicode | 434.5 ns | 2.2884 ns | 2.0286 ns | 2,301,551.1 |   2.44 |     0.02 |       0 B |
```

Looking promising. 😄

joshfree:
> Ok, I've updated the benchmarks to use the overloads that take the reader and done the initial microbenchmarks.

Nice results @JeremyKuhne !

joshfree (approved):

src/Kestrel.Core/Internal/BufferReader.cs
```diff
@@ -1,153 +1,494 @@
...
+                        {
+                            first = first.Slice(startIndex, endIndex - startIndex);
+                        }
+                    }
+                    else
+                    {
+                        if (isMultiSegment)
+                            ThrowInvalidOperationException_EndPositionNotReached();
+
+                        first = new ReadOnlySpan<T>((T[])startObject, startIndex, (endIndex & IndexBitMask) - startIndex);
```
stephentoub: How do we know that startObject is an array?

src/Kestrel.Core/Internal/Http/IHttpParser.cs
```diff
@@ -11,5 +11,9 @@ public interface IHttpParser<TRequestHandler> where TRequestHandler : IHttpHeade
         bool ParseRequestLine(TRequestHandler handler, in ReadOnlySequence<byte> buffer, out SequencePosition consumed, out SequencePosition examined);
```
benaadams: Probably need to delete this method? Unless its getting called somewhere?

src/Kestrel.Core/Internal/Http/IHttpParser.cs
```diff
@@ -11,5 +11,9 @@ public interface IHttpParser<TRequestHandler> where TRequestHandler : IHttpHeade
...
 
         bool ParseHeaders(TRequestHandler handler, in ReadOnlySequence<byte> buffer, out SequencePosition consumed, out SequencePosition examined, out int consumedBytes);
```
benaadams: Probably need to delete this method? Unless its getting called somewhere?

src/Kestrel.Core/Internal/Http/IHttpParser.cs
```diff
@@ -11,5 +11,9 @@ public interface IHttpParser<TRequestHandler> where TRequestHandler : IHttpHeade
...
+
+        bool ParseRequestLine(TRequestHandler handler, ref BufferReader<byte> reader);
+
+        bool ParseHeaders(TRequestHandler handler, ref BufferReader<byte> reader);
```
benaadams: Should the interface have 2 `ParseHeaders` methods?

src/Kestrel.Core/Internal/Http/IHttpParser.cs
```diff
@@ -11,5 +11,9 @@ public interface IHttpParser<TRequestHandler> where TRequestHandler : IHttpHeade
         bool ParseRequestLine(TRequestHandler handler, in ReadOnlySequence<byte> buffer, out SequencePosition consumed, out SequencePosition examined);
 
         bool ParseHeaders(TRequestHandler handler, in ReadOnlySequence<byte> buffer, out SequencePosition consumed, out SequencePosition examined, out int consumedBytes);
+
+        bool ParseRequestLine(TRequestHandler handler, ref BufferReader<byte> reader);
```
benaadams: Should the interface have 2 `ParseRequestLine` methods?

test/Kestrel.Core.Tests/Http1ConnectionTests.cs
```diff
@@ -484,6 +484,12 @@ public async Task TakeStartLineThrowsOnNullCharInMethod(string method)
                 _http1Connection.TakeStartLine(readableBuffer, out _consumed, out _examined));
             _transport.Input.AdvanceTo(_consumed, _examined);
 
+            var index = requestLine.IndexOfAny(new char[] { '\r', '\n' });
```
pakrym: This one has an exact string too, use `$"{method} / HTTP/1.1" in assert.

src/Kestrel.Core/Internal/BufferReader.cs
```diff
@@ -1,153 +1,494 @@
...
+        /// <summary>
+        /// Check to see if the given <paramref name="next"/> value is next.
+        /// </summary>
+        /// <param name="advancePast">Move past the <paramref name="next"/> value if found.</param>
+        /// <param name="next">The value to look for.</param>
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public bool IsNext(T next, bool advancePast = false)
+        {
+            if (End)
+                return false;
```
stephentoub: Does this provide a perf benefit over doing:
```C#
if (!End)
{
   ...
}
return false;
```
?

src/Kestrel.Core/Internal/Http/HttpParser.cs
```diff
@@ -31,445 +31,281 @@ public HttpParser(bool showErrorDetails)
...
+        private static ReadOnlySpan<byte> Eol => new byte[] { ByteCR, ByteLF };
+
+        public bool ParseRequestLine(TRequestHandler handler, in ReadOnlySequence<byte> buffer, out SequencePosition consumed, out SequencePosition examined)
         {
             consumed = buffer.Start;
             examined = buffer.End;
+            var reader = new BufferReader<byte>(buffer);
+            bool success = ParseRequestLine(handler, ref reader);
+            consumed = reader.Position;
+            if (success) examined = reader.Position;
```
pakrym: Braces around single line blocks everywhere.

src/Kestrel.Core/Internal/Http/IHttpParser.cs
pakrym: I guess this is temporary to avoid breaking many existing usages of old methods.

src/Kestrel.Core/Internal/BufferReader.cs
pakrym: startIndex is positive, endIndex is negative which corresponds to array sequence type https://github.com/dotnet/corefx/blob/master/src/System.Memory/src/System/Buffers/ReadOnlySequence.cs#L551-L552

src/Kestrel.Core/Internal/Http/PathNormalizer.cs
```diff
@@ -14,32 +14,40 @@ public static class PathNormalizer
...
+                    return GetUtf8String(writablePath.Slice(0, pathLength));
+                }
             }
 
-            pathLength = RemoveDotSegments(path);
+            fixed (byte* p = path)
+            {
+                Span<byte> writablePath = new Span<byte>(p, path.Length);
+                pathLength = RemoveDotSegments(writablePath);
+            }
```
stephentoub: These changes seem like a bug waiting to happen: accepting a `ReadOnlySpan` and then mutating it. 😦

stephentoub: But the previous method accepted a `Span<byte>`, not a `ReadOnlySpan<byte>`.

src/Kestrel.Core/Internal/Http/HttpParser.cs
JeremyKuhne: Sure

src/Kestrel.Core/Internal/Http/PathNormalizer.cs
JeremyKuhne: > These changes seem like a bug waiting to happen: accepting a ReadOnlySpan and then mutating it. 

Yeah, though to be clear, I'm not adding this, just "deobfuscating" it.

src/Kestrel.Core/Internal/BufferReader.cs
JeremyKuhne: I don't recall, I'll change it corefxlab. The code here is temporary.

src/Kestrel.Core/Internal/Http/PathNormalizer.cs
pakrym: Before it was taking ReadOnlySequence, unpacking it into a pointer (https://github.com/aspnet/KestrelHttpServer/blob/release/2.2/src/Kestrel.Core/Internal/Http/HttpParser.cs#L66) higher up the stack and creating the Span out of the pointer(https://github.com/aspnet/KestrelHttpServer/blob/release/2.2/src/Kestrel.Core/Internal/Http/HttpParser.cs#L134) before calling this method. At leas not all unsafeness in concentrated in one place.

src/Kestrel.Core/Internal/Http/IHttpParser.cs
benaadams: Its versioning the interface by adding two new methods so its breaking them anyway

JeremyKuhne: I'm unclear on the ramifications of removing the old one- outside of work. :) Are there external dependencies on this interface?

benaadams: Its in the `.Internal` namespace, so its ok to break it :)

pakrym: @benaadams I meant inside kestrel repo, tests/benchmarks/etc.

@JeremyKuhne no, removing them is fine.

src/Kestrel.Core/Internal/Http/PathNormalizer.cs
JeremyKuhne: > But the previous method accepted a Span<byte>, not a ReadOnlySpan<byte>.

Upstream the data comes in as a ReadOnlySequence<byte> that was pinned into a Span. I've pushed that down to the one place where it was actually depending on Span.

test/Kestrel.Core.Tests/Http1ConnectionTests.cs
JeremyKuhne: Ok.

src/Kestrel.Core/Internal/Http/HttpParser.cs
```diff
@@ -31,445 +31,281 @@ public HttpParser(bool showErrorDetails)
...
+            if (headerLine.Length == 0 && value != ByteLF)
             {
-                RejectRequestHeader(headerLine, length);
+                BadHttpRequestException.Throw(RequestRejectionReason.InvalidRequestHeadersNoCRLF);
             }
-            if (headerLine[valueEnd + 1] != ByteCR)
+            else
             {
-                RejectRequestHeader(headerLine, length);
+                RejectRequestHeader(headerLine);
```
JeremyKuhne: I think this should change to a message that states that it found an unexpected CR or LF. Then it would be super clear that the detail message is followed by such. My change cuts the span before CR or LF, the previous code cut before any LF. I'd want to do that in a separate change though I think. Thoughts? @pakrym?

src/Kestrel.Core/Internal/Http/PathNormalizer.cs
stephentoub: > higher up the stack

That actually seems better to me.  With this new variant, code all throughout the stack thinks it's working with something that's read-only when in reality something it calls indirectly deep down may mutate it.  At least with the previous approach, it was something up at the top it seems that did the conversion, and everything down the stack from there knew "this thing might be mutated by something I call".

Why does it start off as `ReadOnly` in the first place?  Are we missing a mutable variant, and is that causing us problems?

Also, I've not tracked it through the whole PR, and am wondering why it's changing as part of this.  Is it just ancillary "clean up", or did this need to happen as part of the rewrite with `BufferReader`?  If it's the latter, and BufferReader is handing out a `ReadOnlySpan` that then needs to be mutated, it makes me wonder if we have a functionality gap, e.g. if the input actually came in as a `Span<T>` and was then cast to a `ReadOnlySpan<T>` as part of handing it off to the `BufferReader`, and if code then relies on the parsing of `BufferReader` and wants the slices it gets back to be mutable, how would we support that?

davidfowl: I think we should just potentially just stackalloc here, paths are super tiny and editing in place is a hack.

JeremyKuhne: > I think we should just potentially just stackalloc here, paths are super tiny and editing in place is a hack.

I'm going to do this right now to avoid this issue. I'll allocate via ArrayPool if strictly necessary (over 128?).

davidfowl: Sure

benchmarks/Kestrel.Performance/Http1ConnectionBenchmark.cs
```diff
@@ -78,14 +78,14 @@ private void InsertData(byte[] data)
 
         private void ParseData()
         {
-            if (!_parser.ParseRequestLine(new Adapter(this), _buffer, out var consumed, out var examined))
+            BufferReader<byte> reader = new BufferReader<byte>(_buffer);
```
davidfowl: var 😄

JeremyKuhne: I'm in a foreign land here. 😝

src/Kestrel.Core/Internal/Http/PathNormalizer.cs
JeremyKuhne: Ok, fixed. Strangely I'm getting even better performance numbers for 'LiveAspNet'.

src/Kestrel.Core/Internal/Http/IHttpParser.cs
JeremyKuhne: Ok, working my way through it.

benchmarks/Kestrel.Performance/HttpParserBenchmark.cs
```diff
@@ -51,24 +51,23 @@ private void InsertData(byte[] data)
 
         private void ParseData()
         {
-            if (!_parser.ParseRequestLine(new Adapter(this), _buffer, out var consumed, out var examined))
+            BufferReader<byte> reader = new BufferReader<byte>(_buffer);
```
danmosemsft: nit, var. (unless ASP.NET has other convention)

src/Kestrel.Core/Internal/Http/HttpParser.cs
```diff
@@ -31,445 +31,286 @@ public HttpParser(bool showErrorDetails)
...
             }
 
-            // Ignore end whitespace
-            var lengthChanged = false;
-            for (; valueEnd >= valueStart; valueEnd--)
+            // Move past the colon
+            int valueStart = nameEnd + 1;
+
+            // Trim any whitespace from the start and end of the value
```
danmosemsft: Is there no existing API to trim a ROS over UTF8?

benchmarks/Kestrel.Performance/HttpParserBenchmark.cs
davidfowl: VAR ALL THE THINGS!

src/Kestrel.Core/Internal/Http/IHttpParser.cs
JeremyKuhne: In the process of condensing to 1.

benchmarks/Kestrel.Performance/HttpParserBenchmark.cs
JeremyKuhne: 'var attemptCleanup = true'

JeremyKuhne:
@davidfowl, @pakrym I think this is now in a finished state. I've run all of the `Kestrel.Performance` benchmarks. The key inner measure is the `HttpParserBenchmark` which I shared initial performance numbers with above (in the 50% faster range):

**Before**

```
               Method |     Mean |    Error |   StdDev |        Op/s | Scaled | Allocated |
--------------------- |---------:|---------:|---------:|------------:|-------:|----------:|
 PlaintextTechEmpower | 266.3 ns | 1.138 ns | 1.064 ns | 3,755,168.1 |   1.00 |       0 B |
           LiveAspNet | 462.6 ns | 3.124 ns | 2.922 ns | 2,161,579.6 |   1.74 |       0 B |
              Unicode | 655.3 ns | 2.428 ns | 2.152 ns | 1,525,975.5 |   2.46 |       0 B |
```

**After**

```
               Method |     Mean |    Error |   StdDev |        Op/s | Scaled | ScaledSD | Allocated |
--------------------- |---------:|---------:|---------:|------------:|-------:|---------:|----------:|
 PlaintextTechEmpower | 175.9 ns | 1.138 ns | 1.064 ns | 5,684,818.2 |   1.00 |     0.00 |       0 B |
           LiveAspNet | 344.3 ns | 1.672 ns | 1.482 ns | 2,904,259.7 |   1.96 |     0.01 |       0 B |
              Unicode | 428.0 ns | 2.015 ns | 1.885 ns | 2,336,299.4 |   2.43 |     0.02 |       0 B |
```

`Http1ConnectionBenchmark` is faster too, in the ~10-15% range.

**Before**

```
               Method |       Mean |     Error |   StdDev |        Op/s | Scaled | ScaledSD |  Gen 0 | Allocated |
--------------------- |-----------:|----------:|---------:|------------:|-------:|---------:|-------:|----------:|
 PlaintextTechEmpower |   525.8 ns |  3.155 ns | 2.951 ns | 1,901,935.3 |   1.00 |     0.00 | 0.4883 |    172 KB |
           LiveAspNet | 1,362.6 ns | 10.312 ns | 9.141 ns |   733,874.6 |   2.59 |     0.02 | 1.9531 |    528 KB |
```

**After**

```
               Method |       Mean |    Error |   StdDev |        Op/s | Scaled | ScaledSD |  Gen 0 | Allocated |
--------------------- |-----------:|---------:|---------:|------------:|-------:|---------:|-------:|----------:|
 PlaintextTechEmpower |   458.2 ns | 2.341 ns | 2.190 ns | 2,182,667.0 |   1.00 |     0.00 | 0.7324 |    172 KB |
           LiveAspNet | 1,248.9 ns | 5.623 ns | 4.984 ns |   800,682.8 |   2.73 |     0.02 | 1.9531 |    528 KB |
```

`RequestParsingBenchmark` is slightly faster, but isn't written efficiently. It resets the connection between the request line and the rest of the header, requiring the reader to be reinitialized. There is one exception and that is `PipelinedPlaintextTechEmpowerDrainBuffer`, which uses the reader through the full request and measures the sort of improvement I'd expect (almost 20%).

**Before**

```
                                   Method |       Mean |     Error |    StdDev |        Op/s | Scaled | ScaledSD |   Gen 0 |   Allocated |
----------------------------------------- |-----------:|----------:|----------:|------------:|-------:|---------:|--------:|------------:|
                     PlaintextTechEmpower |   950.9 ns |  7.243 ns |  6.775 ns | 1,051,650.0 |   1.00 |     0.00 |  0.9766 |   172.13 KB |
                     PlaintextAbsoluteUri | 1,363.7 ns |  5.267 ns |  4.669 ns |   733,309.4 |   1.43 |     0.01 |  0.9766 |   284.13 KB |
            PipelinedPlaintextTechEmpower |   747.6 ns |  4.768 ns |  3.982 ns | 1,337,616.0 |   0.79 |     0.01 | 15.6250 |  2753.01 KB |
 PipelinedPlaintextTechEmpowerDrainBuffer |   540.9 ns |  1.284 ns |  1.072 ns | 1,848,742.9 |   0.57 |     0.00 |  7.8125 |  2753.01 KB |
                               LiveAspNet | 1,754.7 ns |  6.878 ns |  6.097 ns |   569,907.4 |   1.85 |     0.01 |  2.9297 |   528.13 KB |
                      PipelinedLiveAspNet | 1,672.5 ns |  6.553 ns |  5.809 ns |   597,891.3 |   1.76 |     0.01 | 31.2500 |  8466.04 KB |
                                  Unicode | 2,545.9 ns | 14.950 ns | 13.985 ns |   392,789.6 |   2.68 |     0.02 |  3.9063 |   916.25 KB |
                         UnicodePipelined | 2,550.1 ns | 12.750 ns | 11.926 ns |   392,146.8 |   2.68 |     0.02 | 62.5000 | 14824.08 KB |
```

**After**

```
                                   Method |       Mean |     Error |    StdDev |        Op/s | Scaled | ScaledSD |   Gen 0 |   Allocated |
----------------------------------------- |-----------:|----------:|----------:|------------:|-------:|---------:|--------:|------------:|
                     PlaintextTechEmpower |   922.9 ns |  6.138 ns |  5.742 ns | 1,083,502.7 |   1.00 |     0.00 |       - |   172.13 KB |
                     PlaintextAbsoluteUri | 1,262.7 ns |  6.163 ns |  5.765 ns |   791,967.3 |   1.37 |     0.01 |  0.9766 |   284.13 KB |
            PipelinedPlaintextTechEmpower |   713.1 ns |  5.334 ns |  4.728 ns | 1,402,313.1 |   0.77 |     0.01 | 15.6250 |  2753.01 KB |
 PipelinedPlaintextTechEmpowerDrainBuffer |   453.5 ns |  3.220 ns |  2.689 ns | 2,204,956.1 |   0.49 |     0.00 | 15.6250 |  2752.51 KB |
                               LiveAspNet | 1,707.9 ns | 13.422 ns | 11.898 ns |   585,520.3 |   1.85 |     0.02 |  1.9531 |   528.13 KB |
                      PipelinedLiveAspNet | 1,607.5 ns | 14.062 ns | 12.465 ns |   622,102.2 |   1.74 |     0.02 | 46.8750 |  8450.04 KB |
                                  Unicode | 2,404.5 ns | 11.966 ns | 10.608 ns |   415,885.8 |   2.61 |     0.02 |  5.8594 |   916.25 KB |
                         UnicodePipelined | 2,416.9 ns | 15.996 ns | 14.180 ns |   413,744.8 |   2.62 |     0.02 | 62.5000 | 14820.08 KB |
```

`KnownStringsBenchmark` is significantly faster (20% and up). `Http1ConnectionParsingOverheadBenchmark` is slower, but that is expected as we're creating a reader and not actually using it (this test uses a null parser). `Http1WritingBenchmark` is broken in master.

The rest of the `Kestrel.Performance` benchmarks showed no impact.

### Current state and next steps

This change removes essentially all of the unsafe code in HttpParser and makes it significantly faster. It relies on the `BufferReader` holding context to maximize the performance gains. There is a good chance you might find more headroom in other areas by leveraging a similar approach.

It also gets performance gains by changing the way we parse. It breaks up valid lines by CRLF and only passes the "good" line for further parsing, which simplifies the inner parsing logic. When iterating through the design in CoreFXlab I found that to be the source of about half of the performance gains.

The `BufferReader` code is a copy of what will go into CoreFX as `SequenceReader`. I'll be working on getting that through and then the `BufferReader` can be dropped from Kestrel.

benaadams:
Nice 😄

benchmarkapps/PlatformBenchmarks/BenchmarkApplication.HttpConnection.cs
```diff
@@ -72,39 +72,39 @@ private async Task ProcessRequestsAsync()
...
                 {
-                    if (Parser.ParseHeaders(new ParsingAdapter(this), buffer, out consumed, out examined, out int consumedBytes))
+                    if (Parser.ParseHeaders(new ParsingAdapter(this), ref reader))
                     {
                         state = State.Body;
+                        consumed = reader.Position;
+                        examined = consumed;
                     }
-
-                    buffer = buffer.Slice(consumed);
```
halter73: Removing these two slices of the ref buffer param always results in an infinite loop in BenchmarkApplication.ProcessRequestsAsync() since buffer.IsEmpty is never becomes true.

benaadams: Add test to catch this? (e.g. count iterations)

JeremyKuhne: > Add test to catch this? (e.g. count iterations)

Not sure where that would go.

benaadams: > Not sure where that would go.

[HttpParserTests.cs](https://github.com/aspnet/KestrelHttpServer/blob/master/test/Kestrel.Core.Tests/HttpParserTests.cs)?

Something based on [ProcessRequestsAsync](https://github.com/aspnet/KestrelHttpServer/blob/83488886e0ed32e946af2a25ae7f29ce9bcc4273/benchmarkapps/PlatformBenchmarks/BenchmarkApplication.HttpConnection.cs#L40-L79) ensuring its not stuck in an infinite loop?

JeremyKuhne: I don't see that we have a reference to it anywhere from tests. I'm a little hesitant to start adding them given my lack of experience in this codebase...

test/shared/HttpParsingData.cs
```diff
@@ -358,79 +358,79 @@ public static IEnumerable<string> TargetWithNullCharData
         public static IEnumerable<object[]> RequestHeaderInvalidData => new[]
         {
             // Missing CR
-            new[] { "Header: value\n\r\n", CoreStrings.FormatBadRequest_InvalidRequestHeader_Detail(@"Header: value\x0A") },
-            new[] { "Header-1: value1\nHeader-2: value2\r\n\r\n", CoreStrings.FormatBadRequest_InvalidRequestHeader_Detail(@"Header-1: value1\x0A") },
-            new[] { "Header-1: value1\r\nHeader-2: value2\n\r\n", CoreStrings.FormatBadRequest_InvalidRequestHeader_Detail(@"Header-2: value2\x0A") },
+            new[] { "Header: value\n\r\n", CoreStrings.FormatBadRequest_InvalidRequestHeader_Detail(@"Header: value") },
+            new[] { "Header-1: value1\nHeader-2: value2\r\n\r\n", CoreStrings.FormatBadRequest_InvalidRequestHeader_Detail(@"Header-1: value1") },
+            new[] { "Header-1: value1\r\nHeader-2: value2\n\r\n", CoreStrings.FormatBadRequest_InvalidRequestHeader_Detail(@"Header-2: value2") },
```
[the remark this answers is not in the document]
JeremyKuhne: https://github.com/aspnet/KestrelHttpServer/pull/3068#pullrequestreview-170870112

I think we should change the error message whenever we find a bad CR/LF and then it would be pretty easy to understand. Something like `Unexpected CR or LF in header: '{detail}'`.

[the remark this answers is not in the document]
JeremyKuhne: Could even put a separate message up for the two different cases:

`Unexpected CR in header: '{detail}'` or possibly `Unexpected CR without LF in header: '{detail}'` (although this sort of infers that the header is fine other than that, which isn't necessarily true).

Now that I'm looking at it, I could also rewind the reader and take the bad CR or LF into the detail message... Hmm. Maybe that is the right place to start... 

`Unexpected LF in header: '{detail}'`

src/Kestrel.Core/Internal/Http/Http1Connection.cs
```diff
@@ -152,64 +155,88 @@ public void ParseRequest(ReadOnlySequence<byte> buffer, out SequencePosition con
...
-            var result = _parser.ParseHeaders(new Http1ParsingHandler(this), buffer, out consumed, out examined, out var consumedBytes);
-            _remainingRequestHeadersBytesAllowed -= consumedBytes;
+            return result;
+        }
 
-            if (!result && overLength)
+        public bool TakeMessageHeaders(ref BufferReader<byte> reader, long length)
+        {
+            var consumed = reader.Consumed;
+            var remaining = length - consumed;
```
halter73: This annoying extra bookkeeping to keep track of the underlying ReadOnlySequence's length makes me think that BufferReader should offer a cached, lazily-initialized Length property. An efficient Remaining property could also be built on top of this.

JeremyKuhne: It is definitely worth considering. I'm being very cautious about growing the size of the reader, but at first blush it seems useful enough to take the extra field.

test/shared/HttpParsingData.cs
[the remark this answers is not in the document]
halter73: I like distinct error messages. Even just `Unexpected (CR|LF) in header: '{header up until the CR or LF}'` any time you see one without the other is enough for me.

davidfowl:
@JeremyKuhne  is this good to go? @sebastienros We want to run an end to end perf test here, where is the bottt 😄

pakrym:
I've run plaintext on this:

```
Before:
 
RequestsPerSecond:           1,913,912
Max CPU (%):                 93
WorkingSet (MB):             384
Avg. Latency (ms):           3.8
Startup (ms):                388
First Request (ms):          85.9
Latency (ms):                0.6
Total Requests:              19,309,615
Duration: (ms)               10,090
Socket Errors:               0
Bad Responses:               0
SDK:                         2.2.100-rtm-009578
Runtime:                     3.0.0-preview1-27101-02
ASP.NET Core:                3.0.0-alpha1-10655
 
After:
 
RequestsPerSecond:           1,913,873
Max CPU (%):                 92
WorkingSet (MB):             382
Avg. Latency (ms):           5.4
Startup (ms):                393
First Request (ms):          82.2
Latency (ms):                0.4
Total Requests:              19,305,979
Duration: (ms)               10,087
Socket Errors:               0
Bad Responses:               0
SDK:                         2.2.100-rtm-009578
Runtime:                     3.0.0-preview1-27101-02
ASP.NET Core:                3.0.0-alpha1-10655
```

Platform benchmark is unfortunately broken in master.

davidfowl:

JeremyKuhne:
@davidfowl, @pakrym Is there anything else you want done here or can this be merged? Also, how should we track flipping to `SequenceReader` from CoreFX?

davidfowl:
We can't merge this until we have a 3.0 baseline (which I'm hoping will be soon).

cc @sebastienros

davidfowl:
@sebastienros that's not good enough. I'd like us to do regular 3.0 runs against master so that we're not surprised anymore. Once we close down 2.2 and switch to 3.0 then we can take this change and see if it affects that baseline. I'm going to be super anal about changes we can't detect now 😄

muratg:

Issue #3015, opened 2018-10-16
BadHttpRequestException: Reading the request body timed out due to data arriving too slowly


clement911:
We have an asp.net core 2.1 app running on .net 4.7 in azure web app service.

Recently we started getting A LOT of the following error:

> Microsoft.AspNetCore.Server.Kestrel.Core.BadHttpRequestException: Reading the request body timed out due to data arriving too slowly. See MinRequestBodyDataRate.

Our app is under pretty constant heavy load (from incoming webhooks mostly). During big spikes, we tend to see more of these.

We used to run on asp.net core 2.0 / .net 4.6.1 for many months and we never saw that error before.
It seems to have started happening following our recent upgrade to asp.net core 2.1 / .net 4.7.

We'd like to get to the bottom of this issue, but we are not even sure where to start.
Any thoughts?

Tratcher:
@sebastienros have you worked up an FAQ wiki yet?

davidfowl:
I wonder if it's because if thread pool starvation... Is your application fully async? Do you have any other logs from Kestrel? Can you capture a dump or a profile? Is the application CPU bound or IO bound?

clement911:
It just happened a couple of times just earlier, usually it's happening for several successive requests in a short time.
I'm monitoring cpu, memory, thread count and a bunch of other metrics in Azure monitor and it all looks healthy.
So even when there is no apparent bottleneck, and the app is very responsive, it can still happen occasionally. 
Yes it is fully async.

> Do you have any other logs from Kestrel? 

No other warnings/errors. Probably info level but I'm not persisting those, although there is probably a way to get these.

> Can you capture a dump or a profile?

I should be able to capture a dump from the azure portal if you think this will help? 
Not sure what a profile is?

clement911:
Any thoughts on how to get to the bottom of this issue?

We're not sure if the errors are caused by legit requests.
Is it possible find out the url that was requested when this message gets generated?

We have been getting a couple of hundreds of these per day, even though all metrics look healthy.

Any particular log source that I can turn on in azure/asp.net core maybe?

The stack trace doesn't give us much.

```
Exception: Microsoft.AspNetCore.Server.Kestrel.Core.BadHttpRequestException: Reading the request body timed out due to data arriving too slowly. See MinRequestBodyDataRate.
   at Microsoft.AspNetCore.Server.Kestrel.Core.BadHttpRequestException.Throw(RequestRejectionReason reason)
[...]

PR #3070, opened 2018-11-01
 Don't count long tick intervals against rate measurements

halter73:
This might help address #3015

This only affects rate timeouts. Normal fixed timeouts might deserve the same treatment, but that would require some additional locking to ensure we don't modify the sentinel value.

### DIFF
--- a/src/Kestrel.Core/Internal/Infrastructure/TimeoutControl.cs
+++ b/src/Kestrel.Core/Internal/Infrastructure/TimeoutControl.cs
@@ -96,6 +96,8 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Infrastructure
                     return;
                 }
 
+                // Assume overly long tick intervals are the result of server resource starvation.
+                // Don't count extra time between ticks against the rate limit.
                 _readTimingElapsedTicks += Math.Min(timestamp - _lastTimestamp, Heartbeat.Interval.Ticks);
 
                 if (_minReadRate.BytesPerSecond > 0 && _readTimingElapsedTicks > _minReadRate.GracePeriod.Ticks)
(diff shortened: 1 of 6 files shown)


halter73:
@davidfowl Do you think we should something similar to this for fixed timeouts? (e.g. the request header timeout)

src/Kestrel.Core/Internal/Infrastructure/TimeoutControl.cs
```diff
@@ -124,6 +124,13 @@ private void CheckForWriteDataRateTimeout(long timestamp)
         {
             lock (_writeTimingLock)
             {
+                var extraTimeForTick = timestamp - _lastTimestamp - Heartbeat.Interval.Ticks;
```
JunTaoLuo: I was curious whether this could potentially be racy if multiple ticks end up running concurrently but I think it might be rare enough that it's not a problem. Essentially, we expect the _lastTimestamp to be updated after this value is calculated. However, there's no guarantee that CheckForWriteDataRateTimeout and Interlocked.Exchange(ref _lastTimestamp, timestamp) run atomically so concurrent ticks may add extra extraTimeForTick. Not only would this be rare, the effect would be that we allow a little extra time for write, which doesn't seem too bad.

src/Kestrel.Core/Internal/Infrastructure/TimeoutControl.cs
```diff
@@ -96,7 +96,7 @@ private void CheckForReadDataRateTimeout(long timestamp)
                     return;
                 }
 
-                _readTimingElapsedTicks += timestamp - _lastTimestamp;
```
Tratcher: Comments please

src/Kestrel.Core/Internal/Infrastructure/TimeoutControl.cs
halter73: Heartbeat.cs prevents overlapped ticks, so there shouldn't be a race here.

PR #3071, opened 2018-11-01
Ignore certificate errors

JunTaoLuo:
Some failures in https://github.com/aspnet/AspNetCore/issues/3763 will be addressed with this.


test/Kestrel.Transport.FunctionalTests/ChromeTests.cs
```diff
@@ -52,6 +52,7 @@ private void InitializeArgs()
                 $"--no-sandbox " +
                 $"--disable-gpu " +
                 $"--allow-insecure-localhost " +
+                $"--ignore-certificate-errors --enable-features=NetworkService " +
```
halter73: It seems that NetworkService is behind a feature flag because it's still experimental. If we see new issues, we should look at flipping this back off.

Issue #3081, opened 2018-11-08
Websocket sniffing doesn't work for Http2 compatible connections


BrennanConroy:
In SignalR we use the presence of the `IHttpWebSocketFeature` to determine if we should tell the client that websockets is "available". The `Http2Connection` no longer sets the `IHttpUpgradeFeature` which the websockets middleware uses and this makes websocket sniffing fail so clients will fallback to SSE even though websockets is supported over Http1.

BrennanConroy:
@Tratcher @halter73 @anurse @davidfowl

Tratcher:
@muratg @Eilon this is a potential 2.2 RTM bug.

Eilon:
Out of curiosity, could you argue that SignalR's sniffing is what's defective? It does seem a little odd to me to force the HTTP2 implementation to pretend like it *might* support upgrading, even though it absolutely never does and never will.

Eilon:
Have we anticipated any unintended side-effects of changing the HTTP2 implementation to pretend like it might support web sockets? I feel like in *theory* it would be safer to change SignalR, because that would by definition affect only SignalR. Like, have SignalR check:

```c#
bool ShouldSignalRTryWebSockets(Connection connection)
{
    if (connection is Http1)
        return ((HttpUpgrade)connection).IsUpgradable);
    else if (connection is Http2)
        return true;
    else
        return false;
}
```

And that way the ugliness is constrained to SignalR? That seems to be it would be a much smaller change and localized to only the affected area.

halter73:
Theoretically, there could be servers that support HTTP/2 but not WebSockets, so that's probably not the best pattern to promote.

Also, even if the IsUpgradableRequest is false, SignalR currently assumes WebSockets is supported just by the presence of the IHttpUpgradeFeature on a non-WebSocket request.

anurse:
I'd argue the current way WebSockets does this detection makes sense because there are three possible states:

1. The server is incapable of performing an Upgrade. (IIS/HttpSys on Windows 7)
1. The server is capable of Upgrade, but this request isn't an Upgrade request. (All HTTP/2 requests and HTTP requests without a `Connection: Upgrade` header)
1. The server is capable of Upgrade, and this request is an Upgrade request (HTTP/1 requests with a `Connection: Upgrade` header)

Those three states also map well to the behavior of the feature today:

1. The `IHttpUpgradeFeature` is not present.
[...]

PR #3082, opened 2018-11-08
Implement no-op HTTP/2 IHttpUpgradeFeature

halter73:
Addresses #3081 

@muratg We'd like to get this in for 2.2

### DIFF
--- a/test/Kestrel.Core.Tests/HttpProtocolFeatureCollectionTests.cs
+++ b/test/Kestrel.Core.Tests/HttpProtocolFeatureCollectionTests.cs
@@ -19,22 +19,30 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Tests
     public class HttpProtocolFeatureCollectionTests
     {
         private readonly TestHttp1Connection _http1Connection;
-        private readonly HttpConnectionContext _http1ConnectionContext;
+        private readonly HttpConnectionContext _httpConnectionContext;
         private readonly IFeatureCollection _collection;
+        private readonly IFeatureCollection _http2Collection;
 
         public HttpProtocolFeatureCollectionTests()
         {
-            _http1ConnectionContext = new HttpConnectionContext
+            var context = new Http2StreamContext
             {
                 ServiceContext = new TestServiceContext(),
                 ConnectionFeatures = new FeatureCollection(),
                 TimeoutControl = Mock.Of<ITimeoutControl>(),
                 Transport = Mock.Of<IDuplexPipe>(),
+                ServerPeerSettings = new Http2PeerSettings(),
+                ClientPeerSettings = new Http2PeerSettings(),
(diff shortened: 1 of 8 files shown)


test/Kestrel.Core.Tests/HttpProtocolFeatureCollectionTests.cs
```diff
@@ -19,22 +19,30 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Tests
...
                 ConnectionFeatures = new FeatureCollection(),
                 TimeoutControl = Mock.Of<ITimeoutControl>(),
                 Transport = Mock.Of<IDuplexPipe>(),
+                ServerPeerSettings = new Http2PeerSettings(),
+                ClientPeerSettings = new Http2PeerSettings(),
             };
 
-            _http1Connection = new TestHttp1Connection(_http1ConnectionContext);
+            _httpConnectionContext = context;
+            _http1Connection = new TestHttp1Connection(context);
```
JunTaoLuo: Http1Connection with a Http2StreamContext?

test/Kestrel.Core.Tests/HttpProtocolFeatureCollectionTests.cs
```diff
@@ -143,24 +151,22 @@ public void FeaturesSetByGenericSameAsByType()
...
-            Assert.Null(http2StreamCollection.Get<IHttpMinResponseDataRateFeature>());
+            Assert.Null(_http2Collection.Get<IHttpMinRequestBodyDataRateFeature>());
+            Assert.Null(_http2Collection.Get<IHttpMinResponseDataRateFeature>());
 
             Assert.NotNull(_collection.Get<IHttpMinRequestBodyDataRateFeature>());
             Assert.NotNull(_collection.Get<IHttpMinResponseDataRateFeature>());
         }
 
+        [Fact]
+        public void Http2StreamFeatureCollectionDoesIncludeUpgradeFeature()
```
JunTaoLuo: This is moved to HttpProtocol so is there an equivalent test to ensure it exist in Http1Connection?

test/Kestrel.Core.Tests/HttpProtocolFeatureCollectionTests.cs
halter73: Http2StreamContext derives from HttpConnectionContext and these tests need just enough properties set on the context to not null-ref in the constructor.

test/Kestrel.Core.Tests/HttpProtocolFeatureCollectionTests.cs
halter73: For Http1Connection, we have tests that actually use the feature.

Eilon:
Approved for 2.2.

Issue #3075, opened 2018-11-06
Interop tests cause chrome.exe console dialogs popups


Tratcher:
Run the tests from VS on your dev machine and you'll get a series of console dialog boxes opening. These should be hidden. @JunTaoLuo

JunTaoLuo:
I'm still seeing popups for https://github.com/aspnet/KestrelHttpServer/blob/ad3cba550955446cfa30db484c4160f569d16961/test/Kestrel.Transport.FunctionalTests/HandleInheritanceTests.cs#L40-L44 on full framework but we are just going to live with is as we plan to remove the full framework tests in 3.0

PR #3084, opened 2018-11-09
Use Selenium to test Chrome

JunTaoLuo:
Addresses https://github.com/aspnet/KestrelHttpServer/issues/3075
I still want to evaluate whether this version is more reliable. It ended up being a test only change so I was hoping to resolve the flakiness in 2.2 as well.


JunTaoLuo:
I've ran the converted Selenium tests 50 times without seeing any failures. Previously I was seeing a test failure with headless chrome 1 in 10-20 runs. It also turns out Selenium tests run faster in 8.6 seconds vs 11.4 seconds before the change.

test/Kestrel.Transport.Sockets.FunctionalTests/Kestrel.Transport.Sockets.FunctionalTests.csproj
```diff
@@ -7,6 +7,7 @@
     <DefineConstants Condition="$([MSBuild]::IsOSPlatform('OSX'))">$(DefineConstants);MACOS</DefineConstants>
     <ServerGarbageCollection>true</ServerGarbageCollection>
     <TestGroupName>Sockets.FunctionalTests</TestGroupName>
+    <NoWarn>CS8002</NoWarn>
```
JunTaoLuo: Unfortunately Selenium.WebDriver isn't strong named so I have to ignore this csc error. Feels strange that we don't do this anywhere else though.

build/dependencies.props
```diff
@@ -36,6 +36,8 @@
     <MoqPackageVersion>4.10.0</MoqPackageVersion>
     <NETStandardLibrary20PackageVersion>2.0.3</NETStandardLibrary20PackageVersion>
     <NewtonsoftJsonPackageVersion>11.0.2</NewtonsoftJsonPackageVersion>
+    <SeleniumWebDriverChromeDriverPackageVersion>2.43.0</SeleniumWebDriverChromeDriverPackageVersion>
```
JunTaoLuo: cc @Eilon, these are new dependencies on third party software. Can we get these through OSS?

test/Kestrel.Transport.Sockets.FunctionalTests/Kestrel.Transport.Sockets.FunctionalTests.csproj
```diff
@@ -7,6 +7,7 @@
     <DefineConstants Condition="$([MSBuild]::IsOSPlatform('OSX'))">$(DefineConstants);MACOS</DefineConstants>
     <ServerGarbageCollection>true</ServerGarbageCollection>
     <TestGroupName>Sockets.FunctionalTests</TestGroupName>
+    <WarningsNotAsErrors>CS8002;$(WarningsNotAsErrors)</WarningsNotAsErrors>
```
halter73: Does this mean one or both of the Selenium packages aren't strong named?

JunTaoLuo: Just Selenium.WebDriver

build/dependencies.props
Eilon: Good to go.

JunTaoLuo:
ping @halter73 @Tratcher

test/Kestrel.Transport.FunctionalTests/ChromeTests.cs
```diff
@@ -40,33 +42,33 @@ public class ChromeTests : LoggedTest
...
-                $"--virtual-time-budget=10000 " +
-                $"--log-net-log={NetLogPath} " +
-                $"--trace-startup --trace-startup-file={StartupLogPath} " +
-                $"--trace-shutdown --trace-shutdown-file={ShutdownLogPath}";
+            ChromeArgs = new [] {
+                $"--headless ",
+                $"--no-sandbox ",
+                $"--disable-gpu ",
+                $"--allow-insecure-localhost ",
+                $"--ignore-certificate-errors --enable-features=NetworkService ",
```
halter73: Why are you including multiple argument flags in a single array element?

test/Kestrel.Transport.FunctionalTests/ChromeTests.cs
```diff
@@ -7,6 +7,7 @@
 using System.Diagnostics;
```
halter73: You might want to move this to Kestrel.Transport.Sockets.FunctionalTests so it doesn't get run for the libuv transport functional tests now that #3091 has been merged.

test/Kestrel.Transport.FunctionalTests/ChromeTests.cs
JunTaoLuo: Good point, I didn't notice when I was updating the format.

JunTaoLuo:
cc @muratg this is the change we'd like to make in 2.2 to make headless chrome tests more stable.

build/dependencies.props
JunTaoLuo: This can now be down non-atomically, apparently.

PR #2502, opened 2018-04-18
Sockets: do 0 byte reads to avoid allocating memory for idle connections

davidfowl:
I cherry-picked 2283605b6b286ad105737547df0e6a769a90ae2d into a new branch on top of release/2.1.

PR #3089, opened 2018-11-10
WIP Revert https://github.com/aspnet/KestrelHttpServer/pull/2502

JunTaoLuo:
The performance regression may be too significant. See https://github.com/aspnet/Benchmarks/issues/569. cc @muratg @Eilon: @davidfowl wanted to potentially revert this for 2.2.


Eilon:
The benchmark issue is comparing 2.2 preview1 and 2.2 preview2. What's up with that? @sebastienros

JunTaoLuo:
The changes we made went into preview2. So the benchmarks are essentially testing without and with the 0-byte read changes.

sebastienros:
I don't know, maybe related to https://github.com/aspnet/Benchmarks/issues/569

JunTaoLuo:
FYI, this is still not committed to being reverted, but I figured I'll prepare the PR first in case we decide to take it. Looking into test failures now.

JunTaoLuo:
Preliminary numbers suggest we'd want to take https://github.com/aspnet/KestrelHttpServer/pull/3094 instead.

PR #3036, opened 2018-10-19
Reduce flakiness of LargeUpload tests

halter73:
- Move h2spec tests to own project

PR #3091, opened 2018-11-10
Stop skipping libuv functional tests

halter73:
I accidentally removed all the test classes shared between Sockets and libuv from the libuv functional test project as part of #3036. Fortunately, we haven't made any transport changes since #3036 was merged.


halter73:
@JunTaoLuo I wonder if we should make the Http2 chrome/selenium test Socket transport only like the h2spec tests.

davidfowl:
Um, do we have anything to catch this

JunTaoLuo:
@halter73 Yes. @davidfowl I think the only way to detect it would be if we add a notification system that keeps track of how many tests we run and alerts us when that number drops.

Drawaes:
It's pretty standard ci setups to have a build gate of test number drops by more than x percent and signoff for going lower of course test number watermark is from last successful build and test run.... My 2c

PR #3092, opened 2018-11-10
 Avoid possible tearing in HeartbeatManager.UtcNow

halter73:


src/Kestrel.Core/Internal/Infrastructure/HeartbeatManager.cs
```diff
@@ -10,18 +11,25 @@ public class HeartbeatManager : IHeartbeatHandler, ISystemClock
...
         public HeartbeatManager(ConnectionManager connectionManager)
         {
             _connectionManager = connectionManager;
             _walkCallback = WalkCallback;
         }
 
-        public DateTimeOffset UtcNow => _now;
+        public DateTimeOffset UtcNow => new DateTimeOffset(UtcNowTicks, TimeSpan.Zero);
+
+        public long UtcNowTicks => Interlocked.Read(ref _nowTicks);
```
benaadams: `Volatile.Read`? 

It does a `Interlocked.CompareExchange(ref location, 0, 0)` on 32 bit, but more lightweight on x64

Whereas `Interlocked.Read` always does an `Interlocked.CompareExchange(ref location, 0, 0)`

Issue #3042, opened 2018-10-20
Implement IThreadPoolWorkItem on IOQueue


davidfowl:
We'll be able to cut the allocations on UnsafeQueueUserWorkItem calls on the IOQueue. Depends on https://github.com/dotnet/corefx/issues/32485

PR #3093, opened 2018-11-11
Remove lock contention in IOQueue

benaadams:
Currently a lot of contention on the lock

image

Go through the lock if its not already doing work; otherwise skip the lock.

Submit work to threadpool outside the lock

After contention is much reduced

image

Also removes QUWIC allocations from the IOQueue on .NET Core 3.0

image

Resolves: https://github.com/aspnet/KestrelHttpServer/issues/3042

### DIFF
--- a/src/Kestrel.Transport.Sockets/Internal/IOQueue.cs
+++ b/src/Kestrel.Transport.Sockets/Internal/IOQueue.cs
@@ -8,9 +8,14 @@ using System.Threading;
 
 namespace Microsoft.AspNetCore.Server.Kestrel.Transport.Sockets.Internal
 {
+#if NETCOREAPP3_0
+    public class IOQueue : PipeScheduler, IThreadPoolWorkItem
+    {
+#else
     public class IOQueue : PipeScheduler
     {
-        private static readonly WaitCallback _doWorkCallback = s => ((IOQueue)s).DoWork();
+        private static readonly WaitCallback _doWorkCallback = s => ((IOQueue)s).Execute();
+#endif
 
         private readonly object _workSync = new object();
         private readonly ConcurrentQueue<Work> _workItems = new ConcurrentQueue<Work>();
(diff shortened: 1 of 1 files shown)


halter73:
Both of the windows Jenkins runs failed due to the Sockets functional tests timing out 😟

https://ci3.dot.net/job/aspnet_KestrelHttpServer/job/master/job/windows-Configuration_Release_prtest/117/consoleFull
https://ci3.dot.net/job/aspnet_KestrelHttpServer/job/master/job/windows-Configuration_Debug_prtest/117/consoleFull

src/Kestrel.Transport.Sockets/Internal/IOQueue.cs
```diff
@@ -45,12 +60,29 @@ private void DoWork()
...
                 {
                     if (_workItems.IsEmpty)
                     {
-                        _doingWork = false;
-                        return;
+                        // Nothing to do, exit.
+                        break;
+                    }
+
+                    if (!_doingWork)
```
halter73: It looks like you got this condition reversed.

benaadams:
> Are there any benchmarks results for this?

Added a PR for 2.2 branch for easy benchmarking https://github.com/aspnet/KestrelHttpServer/pull/3096

>  I see how in this reduces lock contention in Schedule() when DoWork() is in progress, but it does add a Volatile.Read() in Schedule() and the Volatile.Write() in DoWork() each time DoWork() needs to be restarted.

The `Volatile.Read` and `Volatile.Write` should be fairly free on x64 (as there are no register benefits); but does allow it to more often skip `Interlocked.CompareExchange`; spinning and sleeping from the lock.

Arm would have more costs from the `Volatile` calls

src/Kestrel.Transport.Sockets/Internal/IOQueue.cs
```diff
@@ -24,35 +28,69 @@ public override void Schedule(Action<object> action, object state)
...
                 }
 
-                lock (_workSync)
+                // Order is important here with Schedule.
+                // Set _doingWork prior to checking .IsEmpty
+                Volatile.Write(ref _doingWork, 0);
+
+                // Ensure ordering of write -> read is preserved, as order is reversed between Schedule and Execute,
+                // and they are two different memory locations.
+                Thread.MemoryBarrier();
```
benaadams: @stephentoub seem to have a race condition that only happens on netfx without the `Thread.MemoryBarrier` (added x2 to be sure); could this be due to the different implementations of `ConcurrentQueue<T>` on .NET Core and Full Framework and `workItems.IsEmpty` below only acting as a barrier on .NET Core?

stephentoub: > and workItems.IsEmpty below only acting as a barrier on .NET Core?

I doubt it.  Both implementations perform volatile reads, and if memory serves neither implementation has any volatile writes or full barriers.

benaadams: Probably just that am depending on the ordering of read and writes to two different memory locations then

benaadams: Probably just that am depending on the ordering of read and writes to two different memory locations then. Will drop `Volatile` since using `MemoryBarrier`

muratg:

PR #3096, opened 2018-11-12
[WIP][2.2] Remove lock contention in IOQueue

benaadams:
For testing https://github.com/aspnet/KestrelHttpServer/pull/3093

@sebastienros version for benchmarking; should effect pipelined, may effect non-pipelined

### DIFF
--- a/src/Kestrel.Transport.Sockets/Internal/IOQueue.cs
+++ b/src/Kestrel.Transport.Sockets/Internal/IOQueue.cs
@@ -10,7 +10,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Transport.Sockets.Internal
 {
     public class IOQueue : PipeScheduler
     {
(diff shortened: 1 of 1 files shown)


benaadams:
Seems to stick net45x in an infinite loop; am I missing something? /cc @halter73

davidfowl:
This should be against 3.0 (master)

benaadams:
> This should be against 3.0 (master)

Yeah, that's this one https://github.com/aspnet/KestrelHttpServer/pull/3093

This is only to benchmark more easily on 2.2; which is why I've flagged it as [WIP]

sebastienros:
I am not seeing any improvements on either pipelined or non-pipelined, windows or linux. Should it be an obvious win and my tests are wrong?

benaadams:
K, thanks for trying. I was seeing the contention against .NET Core 3.0 (built from current source); so it may be since that is faster, contention comes more into play. Will keep the other PR open.

PR #3103, opened 2018-11-15
Fix Selenium tests WIP

JunTaoLuo:
Trying to address https://github.com/aspnet/AspNetCore-Internal/issues/1363.

These tests pass locally and passes on the VSTS linux agents. I'm suspecting that there's something wrong with the setup of the VSTS Windows agents. I see that the version Chrome and Selenium Chrome driver is different from my local machine so I'm going to see if using the installed chrome driver fixes the issue.


JunTaoLuo:
Huh... this seems to work: https://dnceng.visualstudio.com/public/_build/results?buildId=44532&view=logs. I'm not sure why there's a requirement to match Chrome version and Chrome driver though.

